PR #65, opened 2012-12-15
New feature: Allow user to enter to the survey after registering.

Shnoulle:
New survey setting : directregister : if set to Yes : user have the link to enter to the survey just after registering
And some correction to registercontroller

3 possibility:
- No: default like now
- Yes: shown : show the url after register
- Yes: direct : directly goes to url just after registering

### DIFF
--- a/application/controllers/RegisterController.php
+++ b/application/controllers/RegisterController.php
@@ -118,7 +118,7 @@ class RegisterController extends LSYii_Controller {
         if ($sRegisterErrorMessage != "")
         {
             $_SESSION['survey_'.$surveyid]['register_errormsg']=$sRegisterErrorMessage;
-            Yii::app()->request->redirect(Yii::app()->createUrl("survey/index/sid/{$surveyid}/register/error"));
+            Yii::app()->request->redirect(Yii::app()->createUrl("/survey/index/sid/{$surveyid}/register/error"));
         }
 
         //Check if this email already exists in token database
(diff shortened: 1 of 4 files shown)


maziminke:
Denis, looks like this one was lost in space...?

Carsten, can we add this to 2.5? It is an often requested feature (start token based survey directly after registration).

danielgimeno:
Marcel is not completely forgoten, I've started a plugin to allow token to
log to LDAP server.

I'll keep you informed.

 Daniel

2016-02-19 9:02 GMT+01:00 Marcel Minke <notifications@github.com>:

Shnoulle:
Hi,

It's already in a plugin :
https://git.framasoft.org/SondagePro-LimeSurvey-plugin/replaceRegister/blob/develop/replaceRegister.php#L72

But yes : it can be interesting to have an option in core i think. Showing the token form too : You can register OR if you already have a token.

All 2.1 are lost.

Denis

danielgimeno:
Great, it's good information to not start from scratch.

Thank you

2016-02-19 9:52 GMT+01:00 Denis Chenu <notifications@github.com>:

Shnoulle:
@c-schmitz : my new idea with this : add a new little plugin : 

- Global setting : Redirect- to survey after successfull register by default : yes/no(default)
- Survey setting : Redirect- to survey after successfull register by default : yes/no/default(default)

Maybe we need too : "mandatory email" ?

maziminke:
Denis, I like the idea but I think it would be sufficient to always let the survey admin select which method to be used.

Email should always be mandatory if no redirect is enabled.

Shnoulle:
Yes : right

PR #78, opened 2013-05-03
RPC Routine to delete responses of particular token in a survey.

ravindrakhokharia:
Added delete_responses API routine, will delete responses of the participant, its files, response timings if available

This function will delete Responses for the token passed, 
It will delete all the response related files from the server,
Delete all responses from the table for the Token if available,
Delete Response timings, if available in the table.

### DIFF
--- a/application/controllers/admin/remotecontrol.php
+++ b/application/controllers/admin/remotecontrol.php
@@ -705,7 +705,7 @@ class remotecontrol_handle
 
 		return $oResponses;
 		
-	}
+   }
 	
     /**
      * RPC routine to get survey summary, regarding token usage and survey participation.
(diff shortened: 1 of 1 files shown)


bsdis:
Why did this excellent PR never get merged? :(

maziminke:
I also consider this a very useful API function.

@olleharstedt @gabrieljenik Is there much work involved getting this adjusted as needed and added to LS3 LTS and LS 4?

olleharstedt:
> I also consider this a very useful API function.
> 
> @olleharstedt @gabrieljenik Is there much work involved getting this adjusted as needed and added to LS3 LTS and LS 4?

No, seems straight-forward.

olleharstedt:
Well, I guess the code should use the existing model logic to delete responses.

gabrieljenik:
`Firstly you informed to move the delete files function to the Survey_Dynamic Model`
Not sure if agree with file handling inside a model. Theorically, agree. From practice, models usually handle DB representation. How is that doen now?

Maybe put that specifically on a helper method. Something like: removeFilesForResponse?

olleharstedt:
> Theorically, agree. From practice, models usually handle DB representation. How is that doen now?

IMO, everything should be purged from the ActiveRecord classes except database access. Everything else should be moved to service classes (also called components in Yii). Very long journey to get there, tho.

gabrieljenik:
So, helper method would be middle ground in that journey?

olleharstedt:
No, helper classes with sensible names and limitations. Instead of `activateSurvey` there should be `SurveyActivator` etc. I think Tonis already did that?

arnaudbore:
I know this PR has been closed and seems "dead" since June 2020 but I would like to add my thumb up :+1:  and push forward this feature. I'm also wondering why deleting a response depending on the token and not depending on the ID of the response.

arnaudbore:
Hi @olleharstedt ,
Thank you for reopening this PR. What would be the next step ? I'm afraid @ravindrakhokharia may have move to others projects and maybe not interested anymore in this PR. Just speculation here. Last question: what about deleting response depending on the id instead of a token? In my opinion, ID makes it more flexible.

olleharstedt:
Next step would be for someone to resolve the conflict. Should be easy, maybe just move the function to the bottom.

> Last question: what about deleting response depending on the id instead of a token? In my opinion, ID makes it more flexible.

Sure. Just have to find someone willing to code it. :)

arnaudbore:
Ok I'll give it a try. But definitively will need some help. :+1:

olleharstedt:
Feel free to join our Discord channel for questions.

c-schmitz:
Is this PR still needed? If yes, it needs an update.

arnaudbore:
This can be closed.

PR #181, opened 2014-03-18
#08863: Organize Question/Group functionality

ovillani:
Improved "Organize question groups/questions" functionality. All files
added now. Funtions view/hide marked questions and make marked questions
mandatory/optional still ToDo


c-schmitz:
So I take it that this feature is dead?

mfaber:
This work should really not be lost....Adds so much usibility to LS!!

maziminke:
I also think that this is a very valuable contribution to Limesurvey and will greatly improve the user interface.
Can we somehow get this integrated?!

c-schmitz:
Why did you close it?

c-schmitz:
Ok, finally dead - with the new 2.5 interface it will be harder to merge than to just rewrite it.

LouisGac:
If it's a great feature we could still implement it.
The question ordering functionality right now is not so good, and ovillani did a huge work.
About spaghetti code : he wrote it in the LS2.06 style, I personally can't blame him for that.

PR #206, opened 2014-06-24
Update tokens.php to acknowledge time difference setting

thomkaufmann:
A similar bug was reported and fixed here: http://bugs.limesurvey.org/view.php?id=8563
This bug caused emails to be sent for tokens which were not yet valid.


c-schmitz:
Closed due to missing feedback.

PR #216, opened 2014-09-08
LDAP auth & automatic user creation - AuthLDAP.php

Geeroc:
With the proper config/setting in config-defaults.php and with these changes in AuthLDAP.php, automatic user creation is made possible when using LDAP authentication plugin for limesurvey.


c-schmitz:
PR #225 also offers this feature. Closing this issue.

PR #217, opened 2014-09-10
API : get question properties to return default value

sitawit:
Function `get_question_properties` in RemoteControl 2 API now can be able to return `defaultvalue` upon request.

Users must add `defaultvalue` to `$aQuestionSettings` in order to get this value.
However, the default value is an _answer code_ in case users need assessment value they need to match the default value to data in `answeroptions` with this one. Because `answeroptions` also return `assessment_value`.


c-schmitz:
Merged in #286 instead.

PR #225, opened 2014-10-07
optionally autocreate users

wilberth:
This commit was requested as an update to pull request https://github.com/LimeSurvey/LimeSurvey/pull/176 in the first comment by mennodekker.

This commit offers the option to automatically create users in the user database. Any person that authenticates with 'search and bind' LDAP will be added if not present yet. Users will have their full name a
nd e-mail address set from the LDAP database at creation time. After automatic user creation, they will be able to log on with LDAP credentials or (after resetting their password) with internal database cred
entials.

IMO http://bugs.limesurvey.org/view.php?id=8865 is not an issue of AuthLDAP. It is intrinsic to AuthPluginBase or to authentication plugins in general. Solving this would require not using an authentication
plugin but a 'user database plugin'. As long as it is well understood that AuthLDAP and AuthWebserver offer an alternative authentication method for the existing user and not an additional user list, the cur
rent behaviour makes sense.

This commit does not use the newUserLogin event in an extra plugin as was proposed in the original code. I think user creation belongs in this plugin. Doing it in an extra plugin requires a second LDAP conne
ction in another plugin before this plugin closes the connection. The added complexity is IMO not worth it.

This commit does not produce informative error messages. This is intentional. No information should be given about login failure reasons for safety reasons. Getting all plugin settings correct can be  nuisan
ce because of this. It is therefore advisable to make sure the 'search and bind' LDAP is functional before activating this addition.

Possible future improvements:
- New users will have only 'survey create' permission. Make this configurable.
- Allowing creation of users after simple bind authentication.

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -6,14 +6,6 @@ class AuthLDAP extends AuthPluginBase
     static protected $description = 'Core: LDAP authentication';
     static protected $name = 'LDAP';
 
-    /**
-     * Can we autocreate users? For the moment this is disabled, will be moved 
-     * to a setting when we have more robust user creation system.
-     * 
-     * @var boolean
-     */
-    protected $autoCreate = false;
-
     protected $settings = array(
         'server' => array(
             'type' => 'string',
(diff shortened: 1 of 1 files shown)


Aestu:
I think users autocreation is a bad idea until http://bugs.limesurvey.org/view.php?id=8865 is resolved. This patch could be useful if we delegate users management in LDAP. However, when we create a new user we create also a new DB internal user. This can generate problems: For example, user deactivation in LDAP is not enough because the user can continue to user the Internal DB authentication method and can get a new password (as described in bug).

IMHO, we need an integral solution to bug 8865 and LDAP user creation before thinking in LDAP user autocreation. I hope to be able to merge this solution in 2.06 shortly (we should fix bug 9509).

Aestu:
Closing PR. I just pushed to repository this functionality reworking the code:

https://github.com/LimeSurvey/LimeSurvey/commit/ddda9de2769853b89ad808a61f4d9380102615de

PR #226, opened 2014-10-09
spell check of variable name + some rewrite

linagora-ossa:
spellcheck : oFomattingOptions -> oFormattingOptions

Fix function :
* export_responses
* export_responses_by_token


c-schmitz:
Fixed with #287 - thank you.

PR #234, opened 2014-11-15
set CHtmlPurifier options via setOptions()

rewindrepeat:
The API for CHtmlPurifier has changed. The former public attribute options is now private (and renamed to _options) and has to be set via CHtmlPurifier::setOptions($options).


c-schmitz:
This makes no sense to change. I can see that the current options do work as intended so I see no reason to change it.

PR #249, opened 2014-12-05
qanda_helper.php: remove doubled size for answer

PiNotEqual3:
The doubled size for answer columns in do_array_multiflexi() is wrong. 20% width are missing from the table and if you have the answer table at 100% width the last column will get the additional 20% and it get's to large.
Example:
$answerwidth=20:
$columnswidth=100-($answerwidth*2)=60;
with $numrows = 2;
$cellwidth=$columnswidth/$numrows=60/2 = 30;
=> columns: 20% answer text, 30% row, 30% row
=> 80% total, so the browser adds the missing 20% to the last column
Result: 20% answer text, 30% row, 50% row
I should have created an issue for this...

### DIFF
--- a/application/helpers/qanda_helper.php
+++ b/application/helpers/qanda_helper.php
@@ -5376,7 +5376,7 @@ function do_array_multiflexi($ia)
     {
         $answerwidth=20;
     }
-    $columnswidth=100-($answerwidth*2);
+    $columnswidth=100-($answerwidth);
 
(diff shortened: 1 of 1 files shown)


c-schmitz:
@PiNotEqual3 Thank you  - but I am sorry to say that the fix is not complete.
If you are using a semantic differential scale subquestion (try by adding a subquestion with the following text  'Good|Bad'), then on the right another column will be added.
This has to be taken into account for the width calculation.
Do you want to give it another try?  If you do also have a look at 

Let us know.

c-schmitz:
Closed due to missing feedback.

PR #254, opened 2014-12-08
Move column_style in attribute

Shnoulle:
New feature : allow different column style for each question (list and multiple)


c-schmitz:
Looks fine to me.

Shnoulle:
OK, some test seems strange, review and push today.

Maybe just an issue with config['colstyle']= null;

Shnoulle:
I think i didn't make it .... Don't remind it ....

PR #258, opened 2014-12-11
Replace EM hard coded javascript for event

Shnoulle:
- starting point for #08725: Replace Expression manager Hard coded ja...
- actually only relevance : relevanceOn and relevanceOff function sending 'relevance' event


Shnoulle:
... must be reviewed for actual master.

PR #270, opened 2015-01-13
New feature #9431: Create LDAP users taking email and full name from LDAP

Aestu:
A new type of users can be created:

*  Full name and email address are taken from LDAP at creation time (user and/or admins can change them later).
*  The only authentication method they can use is LDAP.
*  They can't modify their internal DB password
*  Admins can't modify their internal DB password
*  They can't reset their password using "forgotten password" funcionality
*  They don't receive any password in welcome email

### DIFF
--- a/application/controllers/admin/authentication.php
+++ b/application/controllers/admin/authentication.php
@@ -154,12 +154,24 @@ class Authentication extends Survey_Common_Action
                 $aData['errormsg'] = $this->getController()->lang->gT('User name and/or email not found!');
                 $aData['maxattempts'] = '';
                 $this->_renderWrappedTemplate('authentication', 'error', $aData);
+                return;
             }
-            else
+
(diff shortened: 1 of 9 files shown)


c-schmitz:
What if you want to mix LDAP and non-LDAP users?
I see a problem here with the  change in editusers.php - it should not test for the presence of the LDAP plugin - instead a new event be should be fired that if it returns false (set by any plugin), should prevent user editing. Your plugin should also still have the option to do it the old way (mixed mode).

Aestu:
This PR was obsolete. I created #288. The new code uses permission table to store the authentication methods allowed for every user. This makes possible mixed scenarios. For example, we can allow LDAP and Internal DB authentication to Alice but just LDAP to Joe.

Ok about firing an event instead of checking LDAP plugin availability.

PR #271, opened 2015-01-16
Add support for hook_get_auth_webserver_permissions

dani:
Just like hook_get_auth_webserver_profile, hook_get_auth_webserver_permissions can be used to customize the permissions of a newly imported users. This reintroduce the possibility to create users with diffrent permissions which has been removed when switching to the new permission table (see for example https://github.com/LimeSurvey/LimeSurvey/commit/a9b30bc9aa0ac422caa68ca8466cc05ef2dadee5)

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -206,6 +206,31 @@ $config['auth_webserver_autocreate_permissions'] = Array(
 //			'htmleditormode' => 'inline');
 //}
 
+//hook_get_auth_webserver_permissions
+// The optionnal 'hook_get_auth_webserver_permissions' function is for advanced user usage only.
+// It is used to customize the permissions of the imported user
+// If set, the this function will overwrite the auth_webserver_autocreate_permissions
+// defined above by its return value
+//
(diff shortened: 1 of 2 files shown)


Shnoulle:
Hi think it's better to add a new plugin event for this .

dani:
Can you elaborate ?

Shnoulle:
Quick:

        $event = new PluginEvent('createNewUser');
        $event->set('source','Authwebserver');
        $event->set('username',$username);
        $event->set('mail',$mail);
        $event->set('aPermission',$this->api->getConfigKey('auth_webserver_autocreate_permissions');
        $aPermission=>$event->get('aPermission');
See another for LDAP here : https://github.com/LimeSurvey/LimeSurvey/pull/270/files#diff-8382a1d5e1e1a958ce4f64185209acf4R98
We need same event for webserver, LDAP, LScore etc ....

dani:
Well, this is far above my PHP-foo

c-schmitz:
I agree that a plugin would here be the better choice. Sorry @dani , but thank you for submitting it, anyway - we appreciate the effort.

PR #272, opened 2015-01-16
New feature #8898 : New RC2 Routine delete_responses_by_attributes 

ravindrakhokharia:
delete_responses_by_attributes will allow to delete responses by passing attributes

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2401,6 +2401,52 @@ class remotecontrol_handle
 
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


ravindrakhokharia:
I haven't used the deleteAllByAttributes() as it doesnot call the event beforeDelete(), 

Also i have not implemented beforeDelete() event at model level, as it might conflict a routine in the response model which already used to delete the responses with a flag to delete the response files. 

While used the existing routine from the response model, which deletes single response with its files, when passed true as parameter

c-schmitz:
This is completely unacceptable. I won't repeat the same feedback we gave on a similar version of this patch.

PR #273, opened 2015-01-16
Updated feature #9030 : Included patch to get participant response stati...

ravindrakhokharia:
Included patch to get participant response statistics in get_participant_properties


c-schmitz:
Participants response statistics has nothing to do with the existing function get_participant_properties. You will need to create a new function for that - Also your function will crash if there is a token table but the survey is not active yet.

PR #274, opened 2015-01-21
Added alternate PDF font for Croatian language

nmisic:
Default one breaks Croatian specific letters.


Shnoulle:
It's with http://en.wikipedia.org/wiki/Gaj%27s_Latin_alphabet ? Then beter to use dejavusans : more complete. And use alphabetic order please :)

nmisic:
Yes dejavusans would be better. I'll make theese changes.

PR #276, opened 2015-02-07
Remote logout URL

fredyrouge:
Just now Limesurvey can start a session with remote auth but users can't close this remote session using the logout button. this PR try to solve it if you can close this remote session in a specific URL.

After to destroy the Limesurvey session it redirects to a remote URL when a remote webserver (ex: apache with  mod_mellon auth plugin) will destroy the remote session.


c-schmitz:
Hi,
this patch needs to go to the 2.06 branch as it is a new feature. Can you please resubmit it?

fredyrouge:
Done at https://github.com/LimeSurvey/LimeSurvey/pull/285

PR #279, opened 2015-02-20
Fixed Issue: On RemoteControl2, no message set when reordering not possible.

gabrieljenik:
RemoteControl2::set_question_properties() and RemoteControl2::set_group_properties().

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1495,12 +1495,12 @@ class remotecontrol_handle
                     $is_criteria_question=getQuestDepsForConditions($oQuestion->sid,$oQuestion->gid,"all",$iQuestionID,"by-targqid");
                     //We do not allow questions with dependencies in the same group to change order - that would lead to broken dependencies
 
-					if ((isset($dependencies) || isset($is_criteria_question)) && $sFieldName == 'question_order'){
-						$aResult[$sFieldName] = 'Questions with dependencies - Order cannot be changed';							
-						continue;
-					}
-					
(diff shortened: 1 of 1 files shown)


Aestu:
Thanks a lot for your contribution.

Some comments:
  * Pay attention to spaces, you use tabs. Take a look at https://manual.limesurvey.org/Coding_guidelines. *Indent using spaces instead of TAB*.
  * You use tabs or spaces at the end of lines. git diff, git show, etc, show this tabs in red, you can use them to detect it.
  * Some time ago, Carsten announced that merging changes into 2.06 is too difficult so you should generate another pull request for 2.06.

Thanks again,

Alfredo

gabrieljenik:
Hi, 
I think it is OK now.
Sorry for the missuse of tabs vs spaces.
Let me know if this is OK so I can make a similar PR for v2.06

Thanks!

Aestu:
Are you sure you can skip safely the try-catch code? (Just a question, not enough time to test)

This is not important but there are spaces at the end of some lines. If you are using command line, use git show <commit_id> and you will see them in red. Other suggestion: Generate just one commit or respect commit message rules in every commit message.

Thanks,

Alfredo

gabrieljenik:
Sorry for the commit message. I will pay more attention.
About the spaces, I have actually edited this using GitHub inner editor. I will review the ending spaces.

About the avoiding try-catch, yes, I it is OK to avoid it as it only happens for "order" field if order can't be modified because of dependencies.

Aestu:
Please, don't forget to write the 2.06 patch

Thanks!

gabrieljenik:
Nop... I won't... done :)

PR #288, opened 2015-02-27
New authentication method management for users (just to show the code, planned for 2.06)

Aestu:
This PR is an improved version of PR #270. This PR is just a way to show the code and receive comments, this functionality is planned for 2.06.

This PR includes:
- Users have new permissions: Use DB internal authentication, use LDAP authentication, use WS authentication
- When creating a user, we can choose LDAP or internal authentication. This configuration can be changed later. For example, we can change an user from internal to LDAP or we can allow both authentication methods for specific users.
- If we choose LDAP authentication method, full name and email address are taken from LDAP at creation time (user and/or admins can change them later).
- User not allowed to use internal DB authentication is not allowed to reset their password using "forgotten password" functionality
- User not allowed to use internal DB authentication doesn't receive any password in welcome email

TODO:
- Instead checking LDAP plugin presence, a new event should be fired that if it returns false (set by any plugin), should prevent user editing for the particular user.

### DIFF
--- a/application/libraries/PluginManager/PluginManager.php
+++ b/application/libraries/PluginManager/PluginManager.php
@@ -69,6 +69,26 @@
         }
         
         /**
+         * Return the status of plugin (true/active or false/desactive)
+         *
+         * @param sPluginName Plugin name
+         * @return boolean
+         */
+        public function isPluginActive($sPluginName)
+        {
+            $pluginModel = Plugin::model();
+            $record = $pluginModel->findByAttributes(array('name' => $sPluginName));
+            if ($record == false)
+            {
+                return false;
+            }
+            else
+            {
+                return $record->active;
+            }
+        }
+
+        /**
          * Returns the storage instance of type $storageClass.
          * If needed initializes the storage object.
          * @param string $storageClass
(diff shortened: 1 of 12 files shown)


Aestu:
permissions

application/libraries/PluginManager/PluginManager.php
```diff
@@ -69,6 +69,26 @@ public function getInstalledPlugins()
         }
         
         /**
+         * Return the status of plugin (true/active or false/desactive)
+         *
+         * @param sPluginName Plugin name
+         * @return boolean
+         */
+        public function isPluginActive($sPluginName)
```
Shnoulle: I like this function :)

SamMousa:
In LS3 authentication and authorization will be split. It will be possible, for example, to authenticate against LDAP and authorize against the local database.
Or authorize against LDAP (allowing user rights to be shared across LS installations for example).

What I do like about the screenshot however, is the ability to add new authorization items from a plugin! Definitely something we need!

Aestu:
@SamMousa I think your approach for LS3 is great but this PR is a good solution for 2.06. This fixes bug 8865. Futhermore, with this PR we can choose one or more authentication methods for any user. By the way, could you please take another look at bug 9509? Denis and me are able to reproduce this bug, I tried but I cannot find the problem. I doesn't find the difference in dependencies management between 2.05 and 2.06. Thanks!

SamMousa:
Will do Wednesday. Agreed BTW just wanted everyone to know the code has to
be changed in 3.0.0 again.

PR #289, opened 2015-03-02
Add CAS authentication plugin with user auto-creation capability

goyome:
This pull request intend to restore the CAS authentication capability which disappear with Limesurvey 2.05.
There's also an option to automatically create user based on LDAP.


Shnoulle:
Why put it in core and not in an external plugin ?

goyome:
I did that because of a lack of knowledge. I used the other authentication methods for inspiration and put the code in the same place.
I used the information on the following page to do it and didn't found any reference to external plugin.
 https://manual.limesurvey.org/Authentication_plugin_development

Do you have any pointer on the way to convert that code to an external plugin ?
Thanks in advance

goyome:
I think I understood what you meant. I moved the AuthCAS directory to LimeSurvey/plugins.
Is that OK ?

Shnoulle:
There are already 2 external authentification plugin : https://www.limesurvey.org/en/extensions/61-authentication.

You can take some code for example (WP does some "auth permission"). 

To load external library , i use:
third_party/myExternalLib/externallib.php (for example)
````
        $basedir=dirname(__FILE__); 
        Yii::setPathOfAlias('myplugin', $basedir);
        Yii::import('myplugin.third_party.myExternalLib.externallib');
````

I think you can do:
 ````
        $basedir=dirname(__FILE__); 
        Yii::setPathOfAlias('myplugin', $basedir);
        Yii::import('myplugin.third_party.CAS.CAS');
        phpCAS::client ....
````

Shnoulle:
We don't need CAS always ;) . 
If you found a way to move it to the plugin dir, it's really great :)

I think you can put it in our extension "dépot" after. Not sure we put it in default/demo plugin @c-schmitz ?

c-schmitz:
@goyome If you are willing to take care of maintenance in the future for this plugin we would be taking this into our repository. If you are not sure if you can maintain it, you should rather submit it to our extension directory. Anway, good work.
Please also have a look at 
and 
https://manual.limesurvey.org/Coding_guidelines

goyome:
OK. Thank you both a lot for your comments. 
I'm not completely sure to be able to maintain it. So, I'll take the option of the extension.
I close this PR et will add an entry in the extension repository as soon as possible.

goyome:
By the way, my PR included the deletion of old CAS parameters which are not used anymore. You could include that in a future modification of  application/config/config-defaults.php

c-schmitz:
Sure, will do. Thank you!

Shnoulle:
Can you provide a link to the plugin when it's made (maybe on github ?)

goyome:
Yes, no problem. But we may create an organization on github for my university to avoid linking this plugin only to me. If that doesn't happen I will create a new project containing only the plugin in my personal space. It may take a few days to happen but it will happen.

goyome:
Hi, 
I created an entry in the extension directory for this code a few days ago but nothing appeared since. Is it normal ? I can't follow the validation process, so, I was hoping you could tell me what are the next steps.
Thanks

Shnoulle:
Hi @goyome : is it now : https://www.limesurvey.org/en/extensions/85-authcas :)
I think @maziminke  ask at Carsten.
Great thanks

And sometimes, you have to call @c-schmitz , because sometime he takes care of his children ... 
I have too call him too for some of extensions . You can call him here , by email or even by irc ;)

goyome:
Wonderful. Thanks a lot for your quick and efficient answers. :smile:

PR #290, opened 2015-03-05
Fixed issue : Download files from response is shown even if there are no files

Shnoulle:
Dev: We don't have in 2.05, but clicking on the picture reset the filter


Shnoulle:
capture du 2015-03-05 18 24 40

c-schmitz:
Looks fine to me.

Shnoulle:
Yep, but icon are not at the same place . I test with btn class in a, but really too big ....

Maybe we have use only one time Response::model($surveyid)->findByPk($row['id']) (or Response::model($surveyid)->findAll($oCriteria) and a each Response::model($surveyid)->findAll($oCriteria)).

Some more clean dev ....

PR #291, opened 2015-03-12
Converted raw queries to use the RecordSet approach

intrixius:
This patch would make it easier to customize afterwards

### DIFF
--- a/application/controllers/admin/responses.php
+++ b/application/controllers/admin/responses.php
@@ -517,6 +517,15 @@ class responses extends Survey_Common_Action
                 $oCriteria->addCondition("submitdate IS NOT NULL");
             }
 
+            //if not admin, only show responses from participants that are owned by the current user
+            if (!Permission::model()->hasGlobalPermission('superadmin','read')) {
+                $oCriteria->select = "t.*, tokens.*";
+                $oCriteria->join = "LEFT JOIN {{tokens_" . $iSurveyID . "}} tokens ON t.token = tokens.token";
+                $oCriteria->join .= " LEFT JOIN {{participants}} p ON p.participant_id = tokens.participant_id ";
+                $iUserId = Yii::app()->session['loginID'];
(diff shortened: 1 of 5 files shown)


SamMousa:
I like the idea of not using raw queries!
Why not use AR instead? And if we stick with the createCommand i think $oDbCommand would be a better name for the variable since the object does not contain nor manage the records.

Won't something like Token::model($sid)->where(....)->findAll() work as well if not better?

SamMousa:
Also when using this approach there should be no need to use raw / manually escaped where clauses, instead use the helper functions to add column conditions then Yii will take care of escaping.

intrixius:
well, I am quite new to Yii, so I don't know all helpers yet.
I just looked at SurveyDao and used a similar approach...

SamMousa:
http://www.yiiframework.com/doc/api/1.1/CDbCriteria

Things like addInCondition, addColumnCondition stuff like that. 
Basically your code would be:
$criteria = new \CDbCriteria();
$criteria->addInCondition('color', array('blue','green','red');

$tokens = Token::model($surveyId)->findAll($criteria);

Are you be up for trying it that way?

intrixius:
looks even better, I will try :)
Though, I will have to find how to perform JOINs and conditions on joined fields, because that's what I need for my next patch...

SamMousa:
Feel free to add me on skype (sam_mousa) if you wish to discuss any future work.

Yii's activerecord allows you to mostly abstract away joins and just define relations (which then use joins when needed).

Cheers,

intrixius:
Well, I come from Laravel, so I know about ActiveRecord... will look into it

Shnoulle:
I like the [->compare](http://www.yiiframework.com/doc/api/1.1/CDbCriteria#compare-detail) system too. We don't need condition.
In quota replace addConsition by compare make system really more easy.

:)

c-schmitz:
See 8af6f174024d5d02071931ac23c8dc4afacaa367

PR #293, opened 2015-03-16
Fixed some notices appeared while in debug mode

gabrieljenik:


c-schmitz:
Hello gabriel,

you patch seems to fix issues which should not occur in the first place.
It looks like you have entries in your survey table without related entries in the survey_languagesettings table. If you run the integrity check it  should fix this automatically. Can you verify, please?

gabrieljenik:
I will review that. Strange as this happened with a survey which I had imported

Shnoulle:
I'm not sure we call function fixLanguageSetting after import a survey, maybe only for TSV mode.

And maybe some other fix : http://bugs.limesurvey.org/view.php?id=9559

gabrieljenik:
Actually, this appeared on an survey imported from the documentation sample surveys.

gabrieljenik:
Sorry, how shall I check integrity? Haven't done it before!

Shnoulle:
You have the icon at top, just next of 'backup DB'. What is the survey you imported from doc ?

gabrieljenik:
No errors in the DB are shown.
Have rolled back the fixed, tried to reproduce error, but I can't. :(

c-schmitz:
The check probably fixed it - not all fixes are reported.

Shnoulle:
Maybe we can fix notice even if we have to check integrity , and send an alert if user are super admin ?

PR #294, opened 2015-03-17
Fixed issue : unable to show different name than class name

Shnoulle:
Dev: $plugin->name is used to show to user, must be tranlatable for example
Dev: using get_class to use Class name

### DIFF
--- a/application/libraries/PluginManager/LimesurveyApi.php
+++ b/application/libraries/PluginManager/LimesurveyApi.php
@@ -26,7 +26,7 @@
          */
         protected function getTableName(iPlugin $plugin, $tableName)
         {
-            return App()->getDb()->tablePrefix . strtolower($plugin->getName()) . "_$tableName";
+            return App()->getDb()->tablePrefix . strtolower(get_class($plugin)) . "_$tableName";
         }
         /**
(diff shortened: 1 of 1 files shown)


Shnoulle:
$plugin->name are shown to the user, then we must use real class name. Maybe have to be fixed elsewhere too.
It's OK ?

SamMousa:
I will not fix / update this issue in 2.x since all plugins will need to be
changed for LS3 anyway.

Currently in LS3 plugin name is stored in a json metadata file. Personally
I'm against translating the plugin name, translating everything else should
be okay, but in my opinion the name should stay the same.

We don't translate "LimeSurvey" to "LimoenEnquete"  or "citron vert
enquete" either.

Shnoulle:
Maybe but "friendly name" can be different from DB name, no ?

SamMousa:
This will break existing plugins that already have data stored.
We can't break backwards compatibility for this in my opinion.

Shnoulle:
Plugin already need updating actually. It's seems unclear (for me) than name must be "unique id" name.

c-schmitz:
I agree that we need a "nice name" that should/could be different from the plugin name. However the PR will  break existing plugins, so closing it.

PR #275, opened 2015-02-03
Features/spss export fix token attributes

clemens-tolboom:
The export SPSS syntax file contains quite some errors.

This PR tries to fix these by getting rid of local variable and adds a map for token field definition.

It still needs some work but I'd love to get some feedback on how to make this PR land.

PR #296, opened 2015-03-23
Fix SPSS export bugs.

clemens-tolboom:
This PR replaces https://github.com/LimeSurvey/LimeSurvey/pull/275 per requested one commit.

It fixed lot's of SPSS error export definitions probably due to missing if/else construct.

We had
- same field definition in export on completely unrelated field types.
- token field definition were not specific enough.

The code
-  tries to align code for token fields versus data field.
- removed local variables into a new array on each loop thus garding for variable bleeding
- removed inline logic into the `$tokenMap` lookup

### DIFF
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -365,6 +365,73 @@ function SPSSFieldMap($iSurveyID, $prefix = 'V') {
 '*'=>Array('name'=>'Equation','size'=>1,'SPSStype'=>'A'),
 );
 
+    // NOTE: this is a partial map using _defaults + special token types only
+    $tokenMap = array(
+      '_defaults' => array(
+        'qid' => 0,
+        'code' => '',
+        'SPSStype' => 'A',
+        'LStype' => 'Undef',
+        'size' => '100',
+        'hide' => 0,
+        'scale' => '',
+      ),
+      'sent' => array(
+        'SPSStype' => 'DATETIME23.2',
+        'size' => '',
+      ),
+      'remindersent' => array(
+        'SPSStype' => 'DATETIME23.2',
+        'size' => '',
+      ),
+      'remindercount' => array(
+        'SPSStype' => 'F',
+        'size' => '4',
+      ),
+      'usesleft' => array(
+        'SPSStype' => 'F',
+        'size' => '4',
+      ),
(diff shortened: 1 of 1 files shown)


application/helpers/export_helper.php
```diff
@@ -365,6 +365,73 @@ function SPSSFieldMap($iSurveyID, $prefix = 'V') {
...
+        'SPSStype' => 'DATETIME23.2',
+        'size' => '',
+      ),
+      'startlanguage' => array(
+        'SPSStype' => 'A',
+        'size' => '19',
+      ),
+      'token' => array(
+        'SPSStype' => 'A',
+        'size' => '16',
```
Shnoulle: 36 is the max i think :)

mfaber:
Please check whether Stata export still works OK. It builds on the data structure that is created by the SPSS export code.
Thanks!


-------- Original Message --------

tammoterhark:
At Tools for Research we asked Clemens Tolboom to do this work for us, because we make use of SPSS (and PSPP).

We never use the Stata export, so it is impossible for us to check if this still is functioning well.

clemens-tolboom:
We should provide a list of fixed token / field mappings to help the developers.

tammoterhark:
mfaber may I ask your assistance in this? I do not know how to proceed, since I cannot test STATA.

application/helpers/export_helper.php
tammoterhark: Do you mean 36 tokens. Should we change line 375? Is that what you mean?

mfaber:
@tammoterhark Stata export produces XML-files. You could export some responses with and without the changes in your pull request and compare the files (using diff...).

tammoterhark:
OK, thank. Will try this tomorrow.

SamMousa:
@tammoterhark Are you still interested in looking into this, or should I close this PR?

tammoterhark:
Hi Sam, please close it. Mysteriously the faults disappeared after an
upgrade, but I don't remember which update....

Thanks for looking at it.....

Tammo
Op 17 nov. 2015 18:27 schreef "Sam" <notifications@github.com>:

clemens-tolboom:
Darn ... so many changes down the drain :-(

PR #297, opened 2015-03-27
Fixed issue : Plugin management don't set to rtl with rtl language

Shnoulle:
* fix some rtl part for admin, but not sure is the good way
* Did we fix it via one css file (adding .rtl class (:dir(rtl) not ready)
* or adding a -rtl.css for each extension ?
* We have to add lang and rtl for each ayout : need an helper class ?


Shnoulle:
This is still an issue in 2.06. Plugin management is not set to rtl.

PR #300, opened 2015-04-01
CPDB patch for sharing a participant with all the users

gmambro:
Dev public participants in CPDB: allows sharing a participant with all the users #9469
Fixed issue #09461: share a participant without selecting a user

### DIFF
--- a/application/controllers/admin/participantsaction.php
+++ b/application/controllers/admin/participantsaction.php
@@ -315,10 +315,17 @@ class participantsaction extends Survey_Common_Action
 
             foreach ($records as $row)
             {
-                $oShared = User::model()->getName($row['share_uid']); //for conversion of uid to human readable names
+                //for conversion of uid to human readable names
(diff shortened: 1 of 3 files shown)


c-schmitz:
- It not clear what 'all users' means - does it mean all current users in the user list? Does it also include future users?
- Please fix variable naming - see https://manual.limesurvey.org/Coding_guidelines

gmambro:
It was meant to enable future users to avoid the shortcomings of expanding it to the list of currently defined users. The idea to avoid changing the database schema was: when you share participant if uid = 0 means "any users" else share it with that specific user.  If you think it could be something useful (my idea wasabout implementing an import from the corporate address book to the CPDB, all the contacts in the address book are shared with all Limesurvey users) I can fix the variables and rebase it.

gmambro:
I've fixed the naming and rebased the patch on the current master

PR #304, opened 2015-04-01
Dev: Added extra fields returned for list_groups and list_questions API.

gabrieljenik:
- Returning more info will make listing more useful.
- Although the content returned may be longer, it may avoid lot of extra calls for group or question details.
- As for example, to list questions in the right order, before one needed to get question properties for every question listed, seeking the question_order.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1752,7 +1752,7 @@ class remotecontrol_handle
 
                 foreach ($oGroupList as $oGroup)
                 {
-                    $aData[]= array('id'=>$oGroup->primaryKey,'group_name'=>$oGroup->attributes['group_name']);
+                    $aData[]= array('id'=>$oGroup->primaryKey,'group_name'=>$oGroup->attributes['group_name'], 'group_order'=>$oGroup->attributes['group_order']);
                 }
                 return $aData;
             }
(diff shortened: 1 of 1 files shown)


Shnoulle:
Last time : please add "New feature : " to one of the commit , thanks.

gabrieljenik:
Sorry, I thought this was more under a dev comment not a nre feature.

On the other side, Is this a request or just a comment?
Are you asking me to do a new commit although this pull request has already been merged / closed?

Shnoulle:
More a comment :) . I think it's a new feature, but i didn't work a lot with API ....

If you interested to be included in our team :)

gabrieljenik:
Hi, I am interested.

Nevertheless, I think I still don't understand fully the issue.
Do you mean I haven't stick to the commit message guidelines?
I am sorry for that. I thought that I was doing so by starting the commit message with "DEV: ..."..

Shnoulle:
For the message guideline: @c-schmitz use an autmatic system to produce https://github.com/LimeSurvey/LimeSurvey/blob/master/docs/release_notes.txt .

Adding Fixed issuse or New feature (with 2 point after) add a line in the next release. BUt you'r rigth, except for the bug tracker issue, sometimes : "Did i really want to show this in changelog ?". But some user read changelog and can be interested ;)

gabrieljenik:
Have done new small commits to create an entry in the release_notes.
Is that what you were referring to?

Shnoulle:
Oh, it's not really needed. Sorry and thanks.

Shnoulle:
Hi @gabrieljenik . 

If you still interested to be in LS dev team (and push directly ;) ) . https://manual.limesurvey.org/How_to_join_the_LimeSurvey_project_team#First_steps Just the step to mail to LS dev mailing list :)

gabrieljenik:
I am already on that list...
About pushing directly, it would be OK! I like PRs because they let others see what I do. But I can push directly if you prefer.

Thanks!

PR #305, opened 2015-04-01
Dev: Added extra fields returned for list_groups and list_questions API.

gabrieljenik:
- Returning more info will make listing more useful.
- Although the content returned may be longer, it may avoid lot of extra calls for group or question details.
- As for example, to list questions in the right order, before one needed to get question properties for every question listed, seeking the question_order.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1748,7 +1748,7 @@ class remotecontrol_handle
 
                 foreach ($oGroupList as $oGroup)
                 {
-                    $aData[]= array('id'=>$oGroup->primaryKey,'group_name'=>$oGroup->attributes['group_name'], 'group_order'=>$oGroup->attributes['group_order']);
+                    $aData[]= array('id'=>$oGroup->primaryKey) + $oGroup->attributes;
                 }
                 return $aData;
             }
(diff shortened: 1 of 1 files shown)


Shnoulle:
And why not : $oGroup->attributes ? It return all the attribute.

gabrieljenik:
Well yes, that should work too and is simpler... :) Nice to see another view of the situation.
The same would work for questions, although it won't return all attributes, just the ones in questions table.

gabrieljenik:
Have done the changes.
Please, let me know how it looks now.

Shnoulle:
I don't use a lot API, it's just seeing the code. But $oGroup->primaryKey is not array(id,lang) ? We have ot already in attributes , no ?
Then : if we need at start:
array_replace() or arru_merge do the job ?

gabrieljenik:
If I understand correctly,  you mean we are sending some attributes duplicated by using the "+" operator?

Shnoulle:
I don't know , i didn't test ;) . But ->primaryKey is already in ->attributes .

gabrieljenik:
This is a print_r for a $aData row:
No duplicated attributes, as ID doesn't exist as an attribute itself and the primaryKey is set as an array.  So everything good I guess. :)

  Array
  (
    [id] => Array
    (
      [qid] => 8681
      [language] => en
    )

    [qid] => 8681
    [parent_qid] => 0
    [sid] => 299867
    [gid] => 369
    [type] => T
    [title] => xxxx
    [question] => Enter some text
    [preg] =>
    [help] =>
    [other] => N
    [mandatory] => N
    [question_order] => 1
    [language] => en
    [scale_id] => 0
    [same_default] => 0
    [relevance] => 1
  )

Shnoulle:
Seems very good request then ;) :+1:

PR #306, opened 2015-04-02
Fixed issue #9595: PHP error when deleting a condition

gmambro:


Shnoulle:
@gmambro : yep, this is needed . All system need fix here (we do a foreach BUT we need only to do for actual survey for exemple ...- one question => one survey). But ok, quick fix is here.

Can you:
* review the commit comment : "Fixed issue #9595 : Description of the issue"
*Do the commit for 2.06 too

Thanks

PR #309, opened 2015-04-13
Allow uploads of font files

Jan-E:
See https://bugs.limesurvey.org/view.php?id=9601


c-schmitz:
Thank you . Please review our rules for commit messages 
and resubmit. Thank you!

PR #314, opened 2015-04-23
Feature: make AuthLDAP support multiple Base DNs for "Search and Bind"

ecraven:
Dev: split $usersearchbase on semicolon (;)
Dev: loop over the values as DN when trying to authenticate via LDAP

With this patch, you can specify multiple Base DNs to authenticate
against in "Search and Bind" mode. LS will try to find the login in any
one of these DNs. They are tried in order.


ecraven:
would you mind commenting on *why* you closed this, and didn't merge it? conversations on IRC with several members seemed to indicate this would be useful and might be merged.

PR #315, opened 2015-04-28
Fixed issue : enforce the mandatory token field + show warning if the import csv file contains ignored columns  

sieutruc:
Hello,

Bug: 
- When check the mandatory option of token field, the csv file that does not contain that field can be imported. (these token fields are like firstname/lastname/field as the mandatory fields )
- When user import csv file that contains ignored columns, limesurvey should show the warning to the user.  

Patch:
- Generate a list containing all the mandatory fields and trigger an error if the csv file is not in the right format.
-  Show number of ignored columns to user ( %s collumns ignored )

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -1218,6 +1218,7 @@ class tokens extends Survey_Common_Action
             foreach ($languages as $language)
                 $captions[$language][$fieldname] = $_POST["caption_{$fieldname}_$language"];
         }
+
         Survey::model()->updateByPk($iSurveyId, array('attributedescriptions' => json_encode($fieldcontents)));
         foreach ($languages as $language)
         {
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -1218,7 +1218,6 @@ class tokens extends Survey_Common_Action
             foreach ($languages as $language)
                 $captions[$language][$fieldname] = $_POST["caption_{$fieldname}_$language"];
         }
-
         Survey::model()->updateByPk($iSurveyId, array('attributedescriptions' => json_encode($fieldcontents)));
         foreach ($languages as $language)
         {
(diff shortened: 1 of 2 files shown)


Shnoulle:
Hi,

I like to force import of a CSV file even without "mandatory" field. I always think CSV import don't have to test anything, just import (with a checkbox).
Myabe you can add a checkbox : 'force import' ? wher all linesis imported without any test ?

sieutruc:
Hi, thanks for your remark .

Actually in term of security, i think there are 2 issues that we need to consider :
- The token is a security tool used to control who can reply to a particular survey. If the owner has set the mandatory field besides the default ones (firstname/lastname/email), when importing a csv file, of course he has to specify clearly all the fields that must be. So a csv cannot be imported without testing anything unless he didn't check the mandatory fields of these fields.

- If there is any data loss, the user must be notified.  That's why all ignored columns should be explicitly shown as a warning.

<blockquote>Maybe you can add a checkbox : 'force import' ? where all linesis imported without any test ?<blockquote/>

I don't think that it is necessary. In my view, if the user sets a mandatory field, he considers that field acting like firstname field.

Shnoulle:
Hi : maybe it's your point of view.

BUT : i know a lot of user where "mandatory" is used ONLY in 'register' system. Import is used for CATI/CAWI system without another need of mandatory system.

Actually : you remove a possibility to add easily a lot of user . Token manager must have the ability to force import even if field is set to mandatory.

And for security : the attribute must be tested, for example : you add : Woman/Man : anybody can put "dog" if he want. Thne you don't really enforce security.

sieutruc:
So in your opinion, must there be a "force import" checkpoint for the first matter ?

And do you agree to merge the feature of showing number of ignored columns to user ?

sieutruc:
Can you tell me exactly what the "mandatory" field is used for ?

In my opinion,  firstname/lastname/email is the mandatory field. So if i don't declare it explicitely in the csv file, limesurvey will return back an error.(even when i do that as admin)

So if i check the mandatory field for an attribute, limesurvey should behave similarly like those fields above.

sieutruc:
@Shnoulle can you tell me what i can do in order that this patch will be accepted ?

For ex: as you suggested , add a checkbox : 'force import ?

Shnoulle:
I really don't know how to be sure to have this in LS core.  I just put my point of view :)

But actually : i use sometimes token without any information t1,t2,t3 etc .... for the code. Done for CATI : code is from CATI system.

sieutruc:
What does it mean "CATI system" ?

 I have an idea about this functionality. We can keep the old behavior and an option that allows a verification (or a check) of csv . If the use checks that option, any csv file that does not conform to the predefined format ( with the mandatory fields), after imported, limesurvey will show the warning as the patch does.

If the option is not cheked, it continues as it is supposed to do.

How do you think ?

Shnoulle:
CATI : https://en.wikipedia.org/wiki/Computer-assisted_telephone_interviewing :)

Think have an option to check/unchek is a good solution (I like to have 'don't force' by default, but really a point of view).

PR #316, opened 2015-05-15
Allow upload of txt and md files

Jan-E:
And fix the double comma while we are here anyway.


Shnoulle:
Can you do same fix for 2.06 please, we don't have already txt files ....

Jan-E:
See https://github.com/LimeSurvey/LimeSurvey/pull/317

PR #317, opened 2015-05-15
Dev: Allow upload of txt and md files

Jan-E:
Do https://github.com/LimeSurvey/LimeSurvey/pull/316 in the 2.06 branch as well.

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -82,7 +82,7 @@ $config['allowunblacklist']     =  'N';             // Allow participant to unbl
 $config['userideditable']     =  'N';               // Allow editing of user IDs
 $config['defaulttemplate']    =  'default';         // This setting specifys the default theme used for the 'public list' of surveys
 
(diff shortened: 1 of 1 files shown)


Shnoulle:
Great, but before merging, please : https://manual.limesurvey.org/Standard_for_Git_commit_messages for comit message.
Then here `Dev: Allow ...` ;)

Jan-E:
See changed commit message.

Shnoulle:
No, sorry, you update the title, but not the commit message, see https://github.com/LimeSurvey/LimeSurvey/commit/df6dacc733129482a3f458f8e4c998a842c7ef1b .

I think the easiest is to close this one and do a new commit with just the commit (and more : you can use edit this file at https://github.com/LimeSurvey/LimeSurvey/blob/master/application/config/config-defaults.php to do a new commit and push request :) )

Jan-E:
I had been looking for a way to change the title of a commit that is already pushed, but apparently there is no easy way. Here is a new one:
https://github.com/LimeSurvey/LimeSurvey/pull/318

Shnoulle:
Thanks :)

PR #321, opened 2015-06-01
Fixed issue: anomaly of slider

sieutruc:
Hello,

Bug:
- Create a survey with a group
- Create only one question that uses the slider as the attached image

slider

- Move the cursor A to the value 2
- Move the cursor B to the value 7
- Do not move the cursor C.

- Click "envoyer" to send the result to the server.

Result: the validation check fails because the value of the inputs is not good.

Debug: value of A is 2, value of B is 7 but the value of C is "null" in server side. It should normally be "0" in this case.

Patch: 
- Initialize all the value of cursors to real value.

I see in the code that you already did it but it was not enough because error occurs when "jsonOptions.slider_displaycallout" is set to "false". 

			if(havevalue || ( startvalue && jsonOptions.slider_displaycallout)){

### DIFF
--- a/scripts/numeric-slider.js
+++ b/scripts/numeric-slider.js
@@ -51,6 +51,8 @@ function doNumericSlider(qID,jsonOptions) {
 				step: parseFloat(jsonOptions.slider_step),
 				create: function() {
 					$('#slider-callout-'+myfname).appendTo($('#container-'+myfname+' .ui-slider-handle').get(0));
+					startvalue=''+startvalue;
+					startvalue=startvalue.replace(/\./,LSvar.sLEMradix);
 				},
 				slide: function( event, ui ) {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Here : the question is mandatory : then you need to SET default value. Actually is user don't "touch" the slider : it don't answer. Use default value to 0 if you want to set it : limesurvey.sondages.pro/535548/newtest/Y

It's not a anomaly, it's the desired behaviour. Mandatory : touch one times or more each slider.

see http://limesurvey.sondages.pro/535548/newtest/Y without manadatory

c-schmitz:
I agree -  no fix needed here. If the slider is untouched there was no response by the user and so no value is saved to indicate that the slider was not touched.

PR #322, opened 2015-06-04
Fixed issue 09608

scoops:
Selected language wasn't being inserted into the database. LS was pulling from nonexistent id on data entry page. Fixed the POST request, but it was giving a URL instead of the field, so I added code to split up the URL and assign the language code properly.


Shnoulle:
@scoops : seems you make a bug fix for 2.05 or 2.06 version and want tu merge in our develop version. develop version is a complete restart of admin page :) . Please : do it for 2.06 version

PR #323, opened 2015-06-04
Use onchange instead of onkeyup event

fmorote:
Use onchange instead of onkeyup event, now copy&paste token with mouse activates submit button


SamMousa:
Why not use oninput?
http://www.w3schools.com/tags/ev_oninput.asp

Shnoulle:
And better : why not using .on from jquery ?

SamMousa:
Well even if we use jquery we have to pick an event.

Shnoulle:
Yes, but with jquery : using multiple event is very easy :) . And here a blur seems better maybe ?

fmorote:
The oninput event is supported in Mozilla Firefox , chrome, safari, Opera, Internet Explorer from version 9. If LS does not suport <IE 9, not problem :smile:

Shnoulle:
We don't support IE < 9 for admin part :). Only for public part.

fmorote:
Works perfect for me, i apply the change

PR #324, opened 2015-06-04
Release/bugfix 09608

scoops:
Made a pull request for the wrong version of LS. Original comment: Selected language wasn't being inserted into the database. LS was pulling from nonexistent id on data entry page. Fixed the POST request, but it was giving a URL instead of the field, so I added code to split up the URL and assign the language code properly.


Shnoulle:
Hi, actualy : you try to merge your 2.05 fork on our 2.06 ;) .

Actually, we can not automerge between 2.05 to 2.06. You need to do a pull request for 2.05 and a different pull request for 2.06.

After make a pull request : if you find more than your commit in Commits : you have to review your pull request.
We need to see only your modification in Files changed.

Denis

PR #326, opened 2015-06-10
Extra forward slash in front of jstoolbox/date.js

kjw3:
This extra forward slash in front of the jstoolbox/date.js call was causing 500 internal server errors in our setup which is RHEL 6, Apache 2.2.15 and PHP 5.3.3

Requested URL was "/third_party//jstoolbox/date.js".

This was breaking the datepicker for date questions.


c-schmitz:
Thank you!

PR #327, opened 2015-06-11
Feature: make AuthLDAP support multiple Base DNs for "Search and Bind"

ecraven:
Dev: split $usersearchbase on semicolon (;)
Dev: loop over the values as DN when trying to authenticate via LDAP

With this patch, you can specify multiple Base DNs to authenticate
against in "Search and Bind" mode. LS will try to find the login in any
one of these DNs. They are tried in order.


c-schmitz:
Patch looks good - some changes needs to be done though.
Please have a look at https://manual.limesurvey.org/Standard_for_Git_commit_messages
and https://manual.limesurvey.org/Coding_guidelines

We also need the same patch for the branch "develop"

PR #335, opened 2015-06-30
Fixed issue 09706: Wrong minimum PHP version in installer

gmambro:
Array initialization fixed for PHP 5.3


c-schmitz:
Thank you!

PR #339, opened 2015-07-13
Fix #09776: Mail registered participants with RemoteControl API & Fix #09783: Use "adminname <adminemail>" as from-header in emailTokens

Jan-E:
PR for https://bugs.limesurvey.org/view.php?id=9776

mail_registered_participants is almost a direct copy of invite_participants, with one significant change: the default conditions can be overridden. Main use: do not mail all 'unsent' users, but select one.

For instance select one by tid:

    $overrideAllConditions = Array();
    $overrideAllConditions[] = 'tid = 12345';
    $response = $myJSONRPCClient->mail_registered_participants( $sessionKey, $survey_id, $overrideAllConditions );

### DIFF
--- a/application/helpers/admin/token_helper.php
+++ b/application/helpers/admin/token_helper.php
@@ -78,7 +78,7 @@ function emailTokens($iSurveyID,$aResultTokens,$sType)
 
 		$fieldsarray["{ADMINNAME}"] = $oSurvey['admin'];
 		$fieldsarray["{ADMINEMAIL}"] = $oSurvey['adminemail'];
-		$from =  $fieldsarray["{ADMINEMAIL}"];
+		$from = $fieldsarray["{ADMINNAME}"] . ' <' . $fieldsarray["{ADMINEMAIL}"] . '>';
 		if($from ==  '')
 			$from = Yii::app()->getConfig('siteadminemail');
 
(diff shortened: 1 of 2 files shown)


Jan-E:
@Shnoulle or @maziminke
Do i need to create PR's for other branches?

ktomk:
Two pull-requests should be more than enough. Perhaps it's already too much and you should close one of the two. Better ask in the Limesurvey Bugtracker itself against which branch you should create the pull-request. (I think this is commonly *develop* but I don't know for sure how this is handled with the Limesurvey project). And thank you for taking the time to make Limesurvey better.

Jan-E:
@Shnoulle already answered in the bugtracker that I should 'surely' create a PR for the develop branch as well. https://bugs.limesurvey.org/view.php?id=9776

Shnoulle:
@ktomk : in general : only master must be done, but actually develop is a complete rewrite of core code. And merging from actual master to actual develop is very difficult ...

Jan-E:
@ https://github.com/LimeSurvey/LimeSurvey/pull/339#issuecomment-120981487
Pulls to different branches will always be 2 PR's. It is up to you to pull one or the other or both.

Jan-E:
The other is this one BTW:
https://github.com/LimeSurvey/LimeSurvey/pull/340

Jan-E:
@c-schmitz I hope you had a good holiday. If you are not overwhelmed, please take a look at issues #09776 and #09783 
This is a combined PR for both issues

PR #341, opened 2015-07-13
Fixed issue #08030: Prefilled values for obviously not relevant questions lost

Shnoulle:
* use same system than default

I think all startingValues system mut be move to plugin, start in core for the 2 actual. But user can want to prefill by $_POST/$_SERVER or anything else.

Unsure we need actaul startingValues system in EM, adding 2 event in _validateQuestion make the fix, but i don't like 2 event in same function, one seem better .

Maybe one event in Question validation and one in updatedValues ? I think we need an event in updatedValues for Question plugin, and surely in question validation for Questions advanced settings too.


Shnoulle:
@c-schmitz I don't see any potential issue here. OK to merge ? Maybe just move isset($_SESSION[$LEM->sessid]['startingValues'][$sgqa]) test first

Shnoulle:
@SamMousa Why closing ?

SamMousa:
Hmm think by accident in mobile phone.. Wanted to comment that this all
changes in ls3... ;)

Shnoulle:
Oh yes, sure .... But system is same the "default" value. Maybe for LS3 find a "extendable" way at this point ? to move default and prefilled by url more easy to use in plugin (it can be used actualy but less easy).

PR #349, opened 2015-07-22
Fix #09790: Use conditions to get a subset of list_participants

Jan-E:
See https://bugs.limesurvey.org/view.php?id=9790

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1772,9 +1772,10 @@ class remotecontrol_handle
     * @param int  $iLimit Number of participants to return
     * @param bool $bUnused If you want unused tokens, set true
     * @param bool|array $aAttributes The extented attributes that we want
+    * @param array $aConditions Conditions to limit the list, e.g. with array(1 => 'email = "info@example.com"')
     * @return array The list of tokens
     */
-    public function list_participants($sSessionKey, $iSurveyID, $iStart=0, $iLimit=10, $bUnused=false,$aAttributes=false)
+    public function list_participants($sSessionKey, $iSurveyID, $iStart=0, $iLimit=10, $bUnused=false, $aAttributes=false, $aConditions=array() )
     {
         if ($this->_checkSessionKey($sSessionKey))
         {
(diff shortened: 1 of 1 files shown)


c-schmitz:
The way you are doing this would allow for SQL injections. You will need to find a more formal way of using conditions.

Jan-E:
The external application that calls the Remotecontrol API has to deal with possible SQL-injection attempts. Mine does: it strips slashes in the e-mail address and used filter_var to see if it is a valid e-mail address.

Many functions in the LSRC are not doing anything against SQL-injection attempts. Take update_response() for instance. As far as I can see there is no check against SQL Injections.

Shnoulle:
Not true.

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/remotecontrol/remotecontrol_handle.php#L2360

Using Yii with PDO. Here you don't use PDO, then can have SQL injection.

Jan-E:
OK. Is findByAttributes() safe for SQL Injections? If so, I could try to switch to something like Token::model($iSurveyID)->findByAttributes( ... )

Jan-E:
I suppose findAllByAttributes() is not vulnerable for SQL Injections, otherwise this would be vulnerable as well:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/remotecontrol/remotecontrol_handle.php#L2344

Using findAllByAttributes() I wote the following code for list_participants:

	$aAttributeValues = array();
	if (count($aConditions)) {
		foreach($aConditions as $sAttributeName => $value)
		{
			$aAttributeValues[$sAttributeName] = $value;
		}
		$aBasicDestinationFields = Tokens_dynamic::model($iSurveyID)->tableSchema->columnNames;
		$aAttributeValues = array_intersect_key($aAttributeValues, array_flip($aBasicDestinationFields));
	}

	if($bUnused)
		$oTokens = Token::model($iSurveyID)->incomplete()->findAllByAttributes($aAttributeValues, array('order' => 'tid', 'limit' => $iLimit, 'offset' => $iStart));
	else
		$oTokens = Token::model($iSurveyID)->findAllByAttributes($aAttributeValues, array('order' => 'tid', 'limit' => $iLimit, 'offset' => $iStart));

This is called by $aConditions = array('email' => 'info@example.com'). Would that be acceptable?

Jan-E:
Shortened the code a little bit, because part of it was just copying $aConditions to $aAttributeValues.

Jan-E:
@c-schmitz I have created new PR's for the master & develop branch with this code:

https://github.com/LimeSurvey/LimeSurvey/pull/355
https://github.com/LimeSurvey/LimeSurvey/pull/356

If findAllByAttributes() is vulnerable for SQL Injections, the LSRC has a lot of issues. So I must assume this use of findAllByAttributes is acceptable.

We need this for our custom registration app. Waiting for 6 seconds to get and check the list of all participants is a no-go;.

Jan-E:
Closing, because https://github.com/LimeSurvey/LimeSurvey/pull/355 and https://github.com/LimeSurvey/LimeSurvey/pull/356 were merged.

PR #352, opened 2015-07-27
Master statfilter 4browse

Shnoulle:
* Readded feature : stat filter for browse
* Readded feture : stat filter for export
* Unsure if all SQL injection is fixed here
* Please : review some new string before translation ;)

### DIFF
--- a/application/views/admin/export/exportresults_view.php
+++ b/application/views/admin/export/exportresults_view.php
@@ -4,13 +4,18 @@
     var sMsgColumnCount = '<?php eT("%s of %s columns selected",'js'); ?>';
 </script>
 <div class='header ui-widget-header'><?php eT("Export results");?>
-    <?php     if (isset($_POST['sql'])) {echo" - ".gT("Filtered from statistics script");}
-        if ($SingleResponse) {
+    <?php     
+        if (App()->request->getQuery('statfilter'))
+            echo" - ".gT("Filtered from statistics script");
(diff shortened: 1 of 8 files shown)


application/views/admin/export/exportresults_view.php
```diff
@@ -4,13 +4,18 @@
     var sMsgColumnCount = '<?php eT("%s of %s columns selected",'js'); ?>';
 </script>
 <div class='header ui-widget-header'><?php eT("Export results");?>
-    <?php     if (isset($_POST['sql'])) {echo" - ".gT("Filtered from statistics script");}
-        if ($SingleResponse) {
+    <?php     
+        if (App()->request->getQuery('statfilter'))
+            echo" - ".gT("Filtered from statistics script");
```
Shnoulle: @c-schmitz : this string must be updated maybe ? Seems unclear actually.

application/views/admin/responses/filterListResponses_view.php
```diff
@@ -0,0 +1,6 @@
+<p>
+    <?php eT("Showing Filtered Results"); ?>
+</p>
+<p>
+    <?php echo CHtml::link(gT("View without the filter."),array("admin/responses","sa"=>'browse','surveyid'=>$surveyid),array('class'=>'button btn-link')); ?>
```
Shnoulle: @c-schmitz : can you review this string ?

PR #353, opened 2015-07-27
Aligned cached JS to the patch for CSRF vulnerability in KCFinder

gmambro:
When third_party/kcfinder/cache/base.js cannot be updated (e.g. no write permission for webserver or same timestamp as other files in third_party/kcfinder/js) the old minified base.js which does not support csrf is used.


Shnoulle:
Only with autoupdate ? right ?
https://bugs.limesurvey.org/view.php?id=9726
(surely some line-break issue : we didn't see the 2 deletion ...)

c-schmitz:
Thank you - we solved it another way.

Shnoulle:
@c-schmitz : i think i do some test with cache dir not writable by webserver .

Shnoulle:
OK, cache is not needed : 
````
rm -Rf cache/* 
chmod a-w cache
````
No problem : cache still empty (and js/css is produced by PHP script only).

gmambro:
cache should removed from git too... https://github.com/LimeSurvey/LimeSurvey/tree/master/third_party/kcfinder/cache

Shnoulle:
You re right ... done ... totally right .....

PR #354, opened 2015-07-27
Fixed issue #9615

gmambro:
Issue #09615: KCFinder Image Browse Server not working when using db sessions 
This patch allows kcfinder to load sessions from DB (does not break default file based sessions).

### DIFF
--- a/third_party/kcfinder/core/bootstrap.php
+++ b/third_party/kcfinder/core/bootstrap.php
@@ -178,4 +178,196 @@ new SessionSaveHandler();
 
 // PUT YOUR ADDITIONAL CODE HERE
 
+
+/**
+ * A custom session handler to read session from Limesurvey database.
+ *
+ * @author Gabriele Mambrini
+ */
+class LSSessionSaveHandler 
+{
+    public $aDbConfig;
+    public $sSessionTableName;
+ 
+    /**
+     * @param array $aConfig Limesurvey configuration array
+     */
+    public function __construct($aConfig) 
+    {
+        $this->aDbConfig= $aConfig['components']['db'];
+
+	$sTblName = $aConfig['components']['session']['sessionTableName'];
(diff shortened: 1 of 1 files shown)


c-schmitz:
Hi Gabriele,

while the PR in general is fine it needs to be modified to adhere to our Coding guidelines, see https://manual.limesurvey.org/Coding_guidelines  and https://manual.limesurvey.org/How_to_document_your_source_code
- can you please change this accordingly? Thank you!

gmambro:
hope it should be better now

c-schmitz:
Great! Thank you!

Shnoulle:
\o/ great ! better than other fix ;)

Shnoulle:
Thanks again : https://github.com/LimeSurvey/LimeSurvey/commit/7d42ff6dcc7feeff98ab4efa99df5cb6cfa68510 :)

third_party/kcfinder/core/bootstrap.php
```diff
@@ -178,4 +178,196 @@ public function gc($maxlifetime) {
...
+  * checkLSSession() read Limesurvey configuration and if sessions are
+  * managed by CDbHttpSession create a new instance of
+  * LSSessionSaveHandler
+  *
+  * @author Gabriele Mambrini
+  */
+function checkLSSession() 
+{
+    //relative path calculated from the path where kcfinder is running
+    $sLimesurveyFolder = realpath( dirname(__FILE__) . "/../../../application");
```
olleharstedt: Might be easier to use Yii basePath here.

Shnoulle: No, we can't Yii is not set : script launched directly .

olleharstedt: OK.

Shnoulle: PS : i made the same error : asking why we don't use Yii model request .... but PDO

PR #362, opened 2015-08-27
Fixed issue #09845: Map question with OSM + combineJs error

Shnoulle:
Dev: this remove leaflet from combineJs
Dev: remove map.js too : used only one time for each survey, then not at all page
Dev: BUT tested only with apache (with/without subdir), unsure for IIS

If someone can test with different server ? Yii use a really more complex system, but maybe not needed ?


Shnoulle:
Replaced by : 5dea80a5f2be4f01ab7eed8cfc651d45cf4d4096

PR #364, opened 2015-08-28
Problem in update deletable files are not present

weberhofer:
As unlink expects a array an exception occurs in case the searched files are not present.


weberhofer:
This issue is still an issue. Have rebased the changes.

Shnoulle:
Thanks, for next commit/pull request : remind to use Standard for Git commit messages : https://manual.limesurvey.org/Standard_for_Git_commit_messages .
Better with an existing bug number , but not needed :)

PR #366, opened 2015-09-11
Feature/new plugin event admin homepage

KRav007:
I inserted a new plugin event which allows us to insert a custom view on the admin homepage, for non admin users!

Write in a plugin this:

´´´
    public function modifyStartpage(){
      $sViewPath = '/../../plugins/myplugin';
      $sViewName = 'myviewname';
      $event = $this->getEvent();
      $event->set('viewName', $sViewName);
      $event->set('viewUrl', $sViewPath);
}
´´´
Any suggestions for naming the event?

I look forward to your feedback.

Cheers Kai

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@
 .settings/org.eclipse.php.core.prefs
 .buildpath
 /application/config/config.php
+.idea
+.DS_Store
--- a/application/controllers/admin/index.php
+++ b/application/controllers/admin/index.php
@@ -35,7 +35,7 @@ class Index extends Survey_Common_Action
                     'message' => Yii::app()->session['loginsummary']
 
                 ));
-                $this->_renderWrappedTemplate('super', $aViewUrls);
+                $this->_renderWrappedTemplate('/admin/super', $aViewUrls);
             }
             unset(Yii::app()->session['just_logged_in'], Yii::app()->session['loginsummary']);
         }
@@ -46,7 +46,7 @@ class Index extends Survey_Common_Action
                 $this->_renderWrappedTemplate($event->get('viewUrl'), $event->get('viewName'));
 
             }else{
-                $this->_renderWrappedTemplate('super', 'firststeps');
+                $this->_renderWrappedTemplate('/admin/super', 'firststeps');
 
             }
 		}
--- a/application/controllers/admin/index.php
+++ b/application/controllers/admin/index.php
@@ -13,30 +13,46 @@
  */
 class Index extends Survey_Common_Action
 {
-
     public function run()
     {
-        
+
+        $event = new PluginEvent('modifyStartpage');
+
+        App()->getPluginManager()->dispatchEvent($event);
+
+        $event->set('viewName', $event->get('viewName'));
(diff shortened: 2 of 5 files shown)


Shnoulle:
Hi, i don't understand interest here ? We already have https://manual.limesurvey.org/Authentication_plugin_development#beforeLogin 

And in this event we can use:
````
Yii::setPathOfAlias('myPlugin', dirname(__FILE__));
Yii::import("myPlugin.myPluginFunction"); // Not needed here , but in case
Yii::app()->controller->layout='bare'; // bare don't have any HTML
App()->clientScript->registerCssFile(Yii::app()->assetManager->publish(dirname(__FILE__) . '/assets/mystyle.css')); // If needed, can register bootstrap too
$aData=array(
// Anything you want here for view
);
Yii::app()->controller->render("myPlugin.views.myVien",$this->aData);
````

Not tested , but it must work (and if it don't work due to plugin event : fix the event beforeLogin .

Denis

Shnoulle:
Oh, it to replace content for "just_logued in" only ?

I think it best if we can : 
- Add content anywhere
- Eventually replace content everywhere

I'm afraid of 1 000 plugin event to replace content to ... here and here and here ... etc ....

KRav007:
If I understand this "beforeLogin" right, it replaces a screen BEFORE login. The new event replaces the view on path index.php/admin - AFTER login
In the first version it only supports the screen for non super admin users, cause there are at least 2 views that manage the the display of this page for different user roles.

Shnoulle:
You're right .... I add another comment ;). Here you can replace any view ? With any params ? Idea is to do own "admin GUI" ?
If yes : maybe support addContent too ? (addContent have a system , not really used but here) to addContent after or before existing content.
And seems addContent in aplugin with a App()->end(); break next content (then allow to replace existing content).

Denis

KRav007:
> You're right .... I add another comment ;). Here you can replace any view ? With any params ? Idea is to do own "admin GUI" ?
All yes!

> addContent
Why not - will check

> 1000 events 
You are right - it was a starting point for me. 
What do you mean with "Eventually replace content everywhere". Is this a given "feature" or just a name for an general event?

If so, we could insert the event an different places and fire the event, if a certain key to a page is matching.
For my site perhabs "admin-home".

Right?

Shnoulle:
Hi, But after better look at code : you can hack all admin part. Then my sentence "it to replace content for "just_logued in" only ?" is outdated.

Actually : this event can replace any admin GUI. I don't like to replace admin GUI (I do a redirect to my own plugin with afterLogin, for example if user are in groupe 'onlyexportaccess' ). 
I like to see addContent, think it's better.

BUT : all must be redo for 3.0 ... Sam review all admin GUI system to be easily extendable by any plugin. Then i don't think i hack 2.06/2.07 admin GUI, but direct replace by my own actually (still used with 3.0).

LouisGac:
also, we need a component system (adding a custom menu, new pages with lists, forms, etc ) : an easy way to add entirely new component in admin gui respecting the MVC pattern (so using Yii modules).

Using an event system to that goal is not very suitable. But this will not change for LS3.

PR #368, opened 2015-09-24
2.051

apmuthu:
New branch to preserve v2.05+ with all backports till date. Branch forked as of 2015-06-06 LimeSurvey/master branch to my 2.051 branch.


Shnoulle:
What is the reason of backporting some fix but not security fix ?
I think just backporting security is better, no other fix.

apmuthu:
If v2.05 was hived off to another fork of that name and then v2.06 was made the master, all would have been fine. The fact that v2.06 no longer supports PHP v5.3.1 and needs a whole lot of changes breaks many installs if they inadvertantly synch with the master. v2.051 was created for just this reason. 

Now that LimeSurvey GmBH is taking over the project, and ComfortUpgrade by default is set to nag one to upgrade and new developments change all familiar path names like application and framework to protected and other folder names, those actually running v2.05+ code will be left out. The community is only as big as it can carry over it's users. v2.05+ has been in vogue for quite some time and it is in the interest of the project to retain it's installability and continued usability without needless upgradation of versions and technical gymnastics. v2.06 in each of it's recent releases has been installable only after commits have been made after the releases and that too at times with some error on the last step of the install page. 

When PHP version requirements change for newer versions of LimeSurvey, it would be prudent to hive it a separate fork before getting the new version into the master.

Many users still use XAMPP v1.7.3/7 on WinXP atleast as a hot synched local backup in case there is outage at the hosted linux instance and the project would be hard put to support them. Hence a legacy fork stump would help such users look after their needs forking it from there.

Shnoulle:
OK, then confirm : fix security issue before fix other issue ..... I do it with 1.92 during a while, but no time actualy. A backport is a system for security BEFORE fix.

SamMousa:
Hi,

First of all nice work, I'm always happy to see people contributing.
I do want to respond to some of your statements however since my opinion
strongly differs from yours.

1. If you run hosted software on a self-hosted server upgrading is not
optional, I think we should be even more aggressive with requiring newer
versions of base software. There is no excuse for running XAMPP on windows
XP in 2015.
2. If you want new features upgrade your base software, I don't see people
complaining that they can't put new CPUs in their 10 year old computers,
why do they think they have the right to complain about not being able to
run new software on a 10 year old stack?

3. Why do we need a legacy fork if we are not making any commits to it? The
tags / downloadable zips are still there if people want to use old
versions.
4. You mention "syncing" with master, I assume you are referring to
installs by GIT, which are not supported. If you want to install from GIT,
which is a fine approach (I use it myself for development) then why not run
decent (recent) other software as well?

PS. This is just my opinion and does not necessarily reflect that of the
team or LS.

Cheers,

LouisGac:
:+1: 
100% agree with sammoussa. 

I would had one point : 

If you're using LimeSurvey in a professional way, you should update it regularly. Many bugs are corrected from one version to the other. Many bugs have been corrected from 2.05 to 2.06.  As sammoussa said : "There is no excuse for running XAMPP on windows XP in 2015", to run an outdated version of LimeSurvey.

apmuthu:
I beg to disagree.

Anyone using any scriptset in a professional way should check each commit and see what breaks instead of blindly taking latest version is the greatest one and upgrade forthwith. Attempting to upgrade tried and tested versions and needlessly increasing code bloat on upgradation without benefitting from the new "features" that one may have no use for and new "security protections" that may have been otherwise solved or backported is not advisable.

Beware that many open source scripts have now been "corporatised" to serve narrow ends under the guise of speedy development and better sustainability... National and International intelligence agencies and crooks too have begun to benefit from such MBA dictated monopolistic scheming...

apmuthu:
Well here are some use cases:

1. Ability to run and maintain a hot working copy on a local machine running LimeSurvey on XAMPP would be assurance in running on a remote host to be able to survive outage and have it for offline reference.
2. If it ain't broken, don't fix it. The latest and greatest versions are often not justified in being used at all. Attempting to conform to some corporate vision of bloated code for compatibility with their other products is not my idea of progress. Besides, the learning curve will not be worth it as you age.
3. Most of the world's population lives in countries with unfair exchange rates and changing a perfectly working piece of hardware just to conform to someone's yardstick of "better" and "affordable" just does not cut it. A geometric increase in hardware requirements for a less than arithmetic increase in feature set is not a good reason to "upgrade" hardware and/or software.
4. Peer group pressure based on "we should be even more aggressive with requiring newer versions of base software" should not fool the diligent.
5. This legacy fork of mine - v2.051 - will have regular commits backporting what is possible and finding flaws and fixing it (which may not get fixed in the official master) as long as I choose to use it. Others may fork it and develop it and provide pulls back to me incase the official repo chooses not to inculcate it. Just look at the branch names in the existing repo how confusing it is!
6. All I am asking is that those in the community who have the right to create branches in the main repo, create one for v2.051 as of 2015-06-06 master before the merge of the v2.06 and vet my pull request and merge it therein so that those who choose to develop it the way they like will not be denied an opportunity of doing so.

SamMousa:
Well, undoubtedly I'm privileged.. But I for one am wasting enough time
supporting old systems in the current LS version in my spare time; I do not
at all feel even a little obliged to spend more of that time on ancient
systems.

Our project lead is on holiday so when he gets back he will probably get
back to you, regarding the fork.

That said, I do agree that our structure on Github and our versioning in
general is not ideal. That is why we will move to semantic version starting
from LS3.

Good luck with your backports!

apmuthu:
The v2.5 branch, though listed as v2.05+ is no longer so. It was started on 2015-09-10 and since then only LouisGac and Olle Haerstedt to extend bootstrap functionality with most of the v2.06+ in it. It's application/config/version.php states it to be v2.05+ with lower build number and far higher dbversionnumber and keeps it's min PHP version requirements as v5.3.0 whereas it needs v5.3.3 to install fully. The docs/release_notes.txt shows "Welcome to LimeSurvey v2.06+!". An extra table: lime_boxes was created along with some insert data in it. When it fails to install in the last step, the following sql will have to be manually entered (but the admin page still fails):
````
INSERT  INTO `lime_users`(`uid`,`users_name`,`password`,`full_name`,`parent_id`,`lang`,`email`,`htmleditormode`,`templateeditormode`,`questionselectormode`,`one_time_pw`,`dateformat`,`created`,`modified`) VALUES 
 (1,'admin','3838bd5806d32cd91144865aa822b9551417dd2796c163d390baa7074d3067a7','Administrator',0,'en','admin@example.net','default','default','default',NULL,1,'2015-09-30 20:23:31',NULL);
INSERT INTO `lime_plugins`(`id`,`name`,`active`) VALUES (1,'Authdb',1); 
INSERT  INTO `lime_permissions`(`id`,`entity`,`entity_id`,`uid`,`permission`,`create_p`,`read_p`,`update_p`,`delete_p`,`import_p`,`export_p`) VALUES 
 (1,'global',0,1,'superadmin',0,1,0,0,0,0);
INSERT  INTO `lime_settings_global`(`stg_name`,`stg_value`) VALUES 
 ('SessionName','iNat8El\"$<cyL_OsDLhS-1W/UhWJ!`n[?vwGpxz\"IaS1F_f)Zo1OG^wX])X+\\[8;')
,('sitename','LimeSurvey')
,('siteadminname','Administrator')
,('siteadminemail','admin@example.net')
,('siteadminbounce','admin@example.net')
,('defaultlang','en')
,('updateversions','false')
,('updateavailable','0')
,('updatelastcheck','2015-09-30 20:24:05');
````
Make appropriate replacements for SessionName value and admin password here is taken as 'pwd123'. 
The last 3 insertions of the last insert statement if present get wiped out on upgrade
The db upgrade gets performed the db version goes up from 181 to 184.

c-schmitz:
@apmuthu I am sorry but we will not support your efforts, there is simply not time for this legacy version.
2.06 works fine and after having a look through actual bug tracker issues there is no problem with stability at all. It is open source and if you want to continue on that path you are free to do, of course. 

But please don't expect any help from the core team - so please stop asking on every commit it is worth to be backported or not.
Anyway, if you want to help with the new version of LimeSurvey you are always welcome. Your input by your past bug reports was always appreciated and future input on bugs and features will also be.

PR #369, opened 2015-09-25
Google Analytics code not running: missing spaces

joaocc:
The original code was missing spaces before and after the curly brackets.
This caused Lime to try and expand it.

https://manual.limesurvey.org/Workarounds:_Manipulating_a_survey_at_runtime_using_Javascript#How_to_use_Script_.28eg._JavaScript_etc..29_in_LimeSurvey.3F
```
Usage of brackets ({ and })

Important.pngCaution with bracket ( { and }) : Expression manager use bracket ( { and }) to enclose expression. Then if you have to use bracket in your javascript, you must add space or line feed after starting bracket ({) and before closing bracket (})

```

### DIFF
--- a/application/helpers/replacements_helper.php
+++ b/application/helpers/replacements_helper.php
@@ -463,8 +463,8 @@ function templatereplace($line, $replacements = array(), &$redata = array(), $de
                 // Default Google Tracking
                 $_googleAnalyticsJavaScript = <<<EOD
 <script>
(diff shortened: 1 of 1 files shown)


Shnoulle:
Great catch ! To be complet : can you make a new Pull request BUT : with commit message respecting : https://manual.limesurvey.org/Standard_for_Git_commit_messages

Here : think it can be !
````
Fixed issue : Google Analytics code not running
Dev: missing space for EM
````
Then this string is added in our changelog (with your name)

PR #372, opened 2015-10-06
Feature/new plugin event after token table create

KRav007:
A new plugin event after a new token table was created.
For example: you could add attribute fields programmatically


Shnoulle:
Hi, think it can help, but same system for survey table creation can be a great idea ...
Not now, but rename the event to : afterTableCreate (or before, unsure) with something like this:

````
$event = new PluginEvent('afterTableCreate');
$event->set('surveyId', $iSurveyId);
$event->set('type', 'token');
$event->set('name', $tableName);
````

And this function MUST be moved to model : Token::createTable($iSurveyId); , else it don't happen for Remote control (for example) or with another plugin table creation.

SamMousa:
Let's not do that refactoring in 2.x. LS3 already has Token:: createTable
BTW.

KRav007:
What means "refactoring". Do you talk about the naming? I could need this kind of hook in 2.x.

Shnoulle:
@SamMousa LS3 don't have an event in Token::createTable ? Because LS2 already have Token::createTable function ...... then adding an event in 2 take 5 minutes to add it to 3 too ... same place, function i surely near the same.

SamMousa:
Feel free to add it @shnoulle

KRav007:
I put it in the model right now and make a test

Shnoulle:
@KRav007 : for event name / param : do you thinl it's a good idea to have same event for survey table/timing table (other table ? Why not plugin table ?) .

KRav007:
From my point of view it would be an big advantage to hook in everywhere. This offers us the possibility to build very special custom solutions for our clients. This is the future, I think. I get more and more requests for this.

c-schmitz:
Variable naming needs to be fixed.

c-schmitz:
...and should be based on the latest revision.

KRav007:
done

PR #373, opened 2015-10-08
Dev: Fixed issue : SQL error on update survey settings

Shnoulle:
Dev: Fixed issue : Update languageSettings when have it
Dev: unsure for ['surveyls_survey_id','required', 'on' => 'insert'], but think it's OK and better

### DIFF
--- a/protected/models/SurveyLanguageSetting.php
+++ b/protected/models/SurveyLanguageSetting.php
@@ -92,8 +92,17 @@ class SurveyLanguageSetting extends ActiveRecord
     public function rules()
     {
         return [
-            ['surveyls_survey_id','required', 'on' => 'insert'],
-            ['surveyls_language','required', 'on' => 'insert'],
+            [
+                'surveyls_survey_id','exist','except'=>'insert',
+                'allowEmpty'=>false,
+                'attributeName'=>'sid',
+                'className'=>Survey::class,
+            ],
+            [
+                'surveyls_language','in','except'=>'insert',
+                'allowEmpty'=>false,
+                'range'=>($this->survey) ? $this->survey->getAllLanguages() : array('en'),
+            ],
             ['surveyls_email_invite_subj', 'lsdefault'],
             ['surveyls_email_invite', 'lsdefault'],
             ['surveyls_email_remind_subj', 'lsdefault'],
--- a/protected/models/SurveyLanguageSetting.php
+++ b/protected/models/SurveyLanguageSetting.php
@@ -92,6 +92,8 @@ class SurveyLanguageSetting extends ActiveRecord
     public function rules()
     {
         return [
+            ['surveyls_survey_id','required', 'on' => 'insert'],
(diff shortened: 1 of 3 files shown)


protected/models/SurveyLanguageSetting.php
```diff
@@ -92,6 +92,8 @@ public function relations()
     public function rules()
     {
         return [
+            ['surveyls_survey_id','required', 'on' => 'insert'],
+            ['surveyls_language','required', 'on' => 'insert'],
```
SamMousa: You can use CRangeValidator instead:
['surveyls_language', 'in', 'range'=> array_keys($this->survey->allLanguages)]

protected/models/SurveyLanguageSetting.php
```diff
@@ -92,6 +92,8 @@ public function relations()
     public function rules()
     {
         return [
+            ['surveyls_survey_id','required', 'on' => 'insert'],
```
SamMousa: Use CExistValidator instead, since this is a foreign key.

Shnoulle: Good idea :) insert only or not ?

protected/models/SurveyLanguageSetting.php
Shnoulle: Same than https://github.com/LimeSurvey/LimeSurvey/pull/373/files#r41513558 . Needed on insert sure, but needed after too .....

protected/models/SurveyLanguageSetting.php
SamMousa: I'd say add it for all scenarios, can always relax it later.

Shnoulle: Yes, we always need it .... else we can remove it .... some borken system : must be fixed.

protected/models/SurveyLanguageSetting.php
Shnoulle: ... must be fixed : no ability to create SUrvey : $this->survey is null at create

SamMousa: Rules are not checked at creation. Just set survey Id after construction
and it should work.

Shnoulle:
Updated : no test on import done.

protected/models/SurveyLanguageSetting.php
Shnoulle: Rules are checked before survey is created : https://github.com/LimeSurvey/LimeSurvey/blob/develop/protected/controllers/SurveysController.php#L245 
2 solution : 
* we save survey BEFORE languagesettings (tested with : && $survey->save(false) && $languageSetting->validate() not working for language)
* we review languagesettings for 'on'=>'insert'

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/commit/301c4e0ac6196239d5ce3c8585641b37a2cc1d20

PR #374, opened 2015-10-24
Fixed issue #9995:  Internal Server Error when installing outside Doc…

gmambro:
…umentRoot

### DIFF
--- a/application/config/internal.php
+++ b/application/config/internal.php
@@ -37,6 +37,7 @@ $internalConfig = array(
             'combineCss'=>false,
             'compressCss'=>false,
             'combineJs'=>$userConfig['config']['debug']>0?false:true,
+            'basePath'=> dirname(dirname(dirname(dirname(__FILE__)))),
             'compressJs'=>false,
             'packages' => require('third_party.php'),
         ),


application/config/internal.php
```diff
@@ -37,6 +37,7 @@
             'combineCss'=>false,
             'compressCss'=>false,
             'combineJs'=>$userConfig['config']['debug']>0?false:true,
+            'basePath'=> dirname(dirname(dirname(dirname(__FILE__)))),
```
olleharstedt: What is this, really?

gmambro: Default doesn't work when installing outside document root https://github.com/LimeSurvey/LimeSurvey/blob/83eb591e7956bdc4d6d947be4543688296af0b95/application/extensions/ExtendedClientScript/ExtendedClientScript.php#L205

Shnoulle:
Hi,

Yes, this fix the issue BUT : i think we don't have to fix #9995 . Because here : it's really a misconfiguration. Set it "by default" seems a bad idea. 

This can be done without updating LS core : in config.php : you can set 'basePath'=>"what you want".

gmambro:
makes sense, however documentation should be updated for this case since using alias in web server is not a misconfiguration! :D

PR #375, opened 2015-10-28
Dev : Add tokenId param to invite and remind participants method, Onc…

syungkeaw:
Add tokenId param to invite and remind participants method, Once we want to send email to specific participant.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2116,9 +2116,10 @@ class remotecontrol_handle
     * @access public
     * @param string $sSessionKey Auth credentials
     * @param int $iSurveyID ID of the survey that participants belong
+    * @param array $aTokenIds Array of the token ID of participants
     * @return array Result of the action
     */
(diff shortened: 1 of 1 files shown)


syungkeaw:
hi, the $aTokenIds is array, however default is false because the first parameter of $oTokens->findUninvited (LINE:2141) and $oTokens->findUninvitedIDs(LINE:2142) method need to pass false for default value.

olleharstedt:
Then please use `array()` which also evaluates to false.

syungkeaw:
I have already changed to null.

olleharstedt:
OK, no problem.

olleharstedt:
Just make sure the other functions do not expect `=== false`.

syungkeaw:
Sure, it is just ... if ($aTokenIds).

syungkeaw:
Hi, When this pull request is merged?
Or what do you want me to do? 
I just want to know. thank you.

c-schmitz:
I just merged a similar PR. @syungkeaw I think it covers your feature completely. Can you check?

PR #377, opened 2015-10-29
Fixed issue: copyToCentral() fails to create entries in 'participants' table

odalecne:


c-schmitz:
Great catch - thank you very much!

PR #378, opened 2015-10-30
Fixed issue: 'token' column collation is not set in survey tables

odalecne:
In a SQL Server-based setup, this issue breaks any query joining survey table to token table on 'token' field raising a collation conflict error.


c-schmitz:
Thank you very much - good catch!

PR #384, opened 2015-11-11
Fixed issue #10025: Quotas : Terminate survey with warning sentence is unclear

Shnoulle:
Dev: in reality : this don't terminate survey: $_SESSION is leave, submit is not set and token is not closed.
Dev: this just show a warning
Dev: my english is really bad :)


Shnoulle:
English sentence to be checked

PR #385, opened 2015-11-12
Updated feature: Remote Control: list_users accepts user_id filter

gabrieljenik:
Added a user_id filter to be able to retrieve a single user data


gabrieljenik:
Not sure how, but too many things got in here!

PR #386, opened 2015-11-19
New feature: Plugin Event "after participant insert"

KRav007:
This lets you manipulate a CPD dataset after import


Shnoulle:
Think this https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/behaviors/PluginEventBehavior.php#L22 do the job : partcipant table name are always same (not like token and response).

PR #387, opened 2015-11-19
Fixed issue #10052: 0 as answer code in radio list causes problems when 'No answer…

adamwebb:
…' is enabled


Shnoulle:
Hi , goof fix, seems great. Just one think : https://manual.limesurvey.org/Standard_for_Git_commit_messages

Can you rewrite the commit message ?

adamwebb:
Great. Has that worked?

Shnoulle:
Yes, i just test your patch in live and merge :)

PR #388, opened 2015-11-23
Gqlist njgw

njgwinter:
Hi,
For my local installation I've made changes to make editing surveys easier. Specifically, this pull request modifies the edit survey screen to include a column on the left that shows all groups and questions. When a user clicks on a group it opens the group. When they click on a question it opens it for editing, and it opens automatically the advanced settings.
I am making a pull request in hopes that this might be integrated into LS.
This is my first try using git/github, so please be kind if I've messed this up somehow.  I am happy to try to fix if I have.
Also, I am happy to try to make these modifications to one or more of the development versions of LS, but I would appreciate any feedback or guidance before I get into that.


c-schmitz:
When we updated the interface this became obsolete.

PR #390, opened 2015-11-27
Fixed Issue #10062: SuperAdmin can't add users to not-owned UserGroups

gabrieljenik:


c-schmitz:
Hi, can you fix it so that the if statement uses its own {} block ?

gabrieljenik:
Fixed! Thanks

PR #392, opened 2015-11-30
New Feature: Added ADMINNAME and ADMINEMAIL as replacement field for survey texts

gabrieljenik:


Shnoulle:
Hi, not for this one specific. But really : i think we need a core replacement Plugin system :).
* Dev : https://bugs.limesurvey.org/view.php?id=9735
* I'm sure we have a Feature request for this , but don't find it.

PR #393, opened 2015-12-04
Fixed issue : enforce the mandatory token field + show warning if the import csv file contains ignored columns

sieutruc:
@Shnoulle , as specified in the last our discussion in the request #315 , i have created the new patch   that fixes the problem and also proposes an attributes check to client, so that he can import with or without an extra step of verification (in this case, we named it as "display warning attributes".)  

I put a photo of my implementation  as the following figure : 

displawarningattributes

Can you tell me if this solution is good and whether this functionality will be merged to the limesurvey ?
If you would like to add something more, please tell me.

Summary of the use cases : 
- an attribute is mandatory and it is missing in the import csv file, the  "display warning attributes" option is checked, a warning below will be shown to users, ex: `attribute_2 columns missing`
- an attribute is not declared but it is present in the import csv file, the  "display warning attributes" option is checked, a warning  below will be shown to users , ex : `attribute_4 columns ignored`
- If the  "display warning attributes" option is not checked by default, it works like before.

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -2081,7 +2081,7 @@ class tokens extends Survey_Common_Action
                         }
                         $aFirstLine = str_getcsv($buffer, $sSeparator, '"');
                         $aFirstLine = array_map('trim', $aFirstLine);
-                        $aIgnoredColumns = array();
+                        $aIgnoredColumns = $invalideAttrFieldName = $missingAttrFieldName= array();
                         // Now check the first line for invalid fields
                         foreach ($aFirstLine as $index => $sFieldname)
                         {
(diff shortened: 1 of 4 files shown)


sieutruc:
Can you tell me if this proposition will be accepted ? Or do i need to have some modification ?

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/commit/9820071ba8387f6af38c309c44a8cc09e3e10442#commitcomment-14899540

Hi , just asking to @c-schmitz  now :  
> (14:15:43) c_schmitz: shnoulle: 2.5 please
> (14:16:30) c_schmitz: shnoulle: gT("columns missing") needs to be fixed to use printf

Then : for 2.5 .... more work i think.

c-schmitz:
If you could create this patch for 2.5 this would be great.
I also note that your variable naming is not correct.
For explanations please have a look at https://manual.limesurvey.org/Coding_guidelines
and 
https://manual.limesurvey.org/Coding_guidelines#Localization

PR #395, opened 2015-12-08
[Fixed issue] reuse the master field map in case of activating a survey

sieutruc:
BUG: 
 - Create a new survey with the "Save IP address" value to "No", create 5 tokens, do one or two  survey.
 - Deactivate that survey and reactive it with the "Save IP address" value to "Yes", restore the initial token table, and do one survey from that table. 
-----> Error: surveydynamic.ipaddr does not exist.
Check the table of that survey in the database, " describe lime_survey_[surveynumber]", this table does not have the field  "ipaddr".

Reason: 

 - Limesurvey reuses the master field map when activating the survey,  if master field map does not contain ipaddr , the created lime_survey_[surveynumber] table will not contain that field. 

- It is mentioned in the end of the function  createFieldMap of common_helper.php

`// If the fieldmap was randomized, the master will contain the proper order.  Copy that fieldmap with the new language settings`
           ` if (isset(Yii::app()->session['survey_'.$surveyid]['fieldmap-' . $surveyid . '-randMaster']))`

Patch: 

 * unset that session above everytime we activate a survey.

I have also noticed that the argument "$force_refresh" seems not be used in that function ???

Maybe you forget to use it to unset some session variable ? like the one that i did.

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -340,6 +340,9 @@ class SurveyAdmin extends Survey_Common_Action
 
                 $aData['tnewtable'] = $tnewtable;
                 $aData['toldtable'] = $toldtable;
+
+                // Reset the session of the survey when deactivating it
+                killSurveySession($iSurveyID);
             }
 
             //Remove any survey_links to the CPDB
--- a/application/helpers/admin/activate_helper.php
+++ b/application/helpers/admin/activate_helper.php
@@ -258,9 +258,6 @@ function activateSurvey($iSurveyID, $simulate = false)
     //Check for any additional fields for this survey and create necessary fields (token and datestamp)
     $prow = Survey::model()->findByAttributes(array('sid' => $iSurveyID));
 
-	//reset the session data of the survey if exist
-    unset(Yii::app()->session['survey_'.$iSurveyID]);
-
     //Get list of questions for the base language
     $fieldmap = createFieldMap($iSurveyID,'full',true,false,getBaseLanguageFromSurveyID($iSurveyID));
 
--- a/application/helpers/admin/activate_helper.php
+++ b/application/helpers/admin/activate_helper.php
@@ -258,6 +258,9 @@ function activateSurvey($iSurveyID, $simulate = false)
     //Check for any additional fields for this survey and create necessary fields (token and datestamp)
     $prow = Survey::model()->findByAttributes(array('sid' => $iSurveyID));
 
+	//reset the session data of the survey if exist
+    unset(Yii::app()->session['survey_'.$iSurveyID]);
+
     //Get list of questions for the base language
     $fieldmap = createFieldMap($iSurveyID,'full',true,false,getBaseLanguageFromSurveyID($iSurveyID));
 


Shnoulle:
Hi,

Only with random group or not ?

I think https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/database.php#L123 LimeExpressionManager::SetDirtyFlag(); must be used .
Unsure it does the good work whith random group/question.

But here : it seems if we have a random group order ins session AND the survey is reset (for example : adding a group) broke the system.

sieutruc:
i have tested with and without randomgroup, in 2 cases they did not work if i used set LimeExpressionManager::SetDirtyFlag(); 

when i switched "Save IP address" from false to true, the "lime_survey_[idsurvey]" did not contain "ipaddr" field. 

But if i put the "unset" piece of code in the "createFieldMap" and tried to use its "force_refresh" parameter like the following : 

 ` if($force_refresh) {`
        `//unset the master field map session, so we can use the new field map`
       `     unset($_SESSION['survey_'.$surveyid]['fieldmap-' . $surveyid . '-randMaster']);`
    `}`

It seems function like my first patch.

Do you have any suggestion what i found ?

Shnoulle:
Maybe specific for ip adress and not for random group ....

I make some test and look :)

Maybe we can just reset Yii::app()->session['survey_'.$surveyid] directly ..... when activate. But i think EM:setDiretyFlag do it .....

sieutruc:
i looked at the function EM:setDiretyFlag, i do not see it clears the session data of a survey.

I propose the modification like you said, reset all session data of the survey before re/activating it: 

`//reset the session data of the survey if exist`
`unset(Yii::app()->session['survey_'.$iSurveyID]); `

Shnoulle:
Yes, this fix the issue .... but it's a bad way to fix it .... Can happen elsewhere (activate in RC because i know you use it ... (OK,OK RC don't have _SESSION ....)

Shnoulle:
Seems to happen only with group randomization. Can you make a bug report ?
I think this group_randomization is buggy : maybe fix it and not another function.

sieutruc:
after looking at the function "buildsurveysession", i propose another solution that corrects the function "deactivate" of survey, it means whenever user deactivates a survey, it has to clear all the session data of that survey. :) 

Because as you said, the "'fieldmap-' . $surveyid . '-randMaster'" session is only set when there is randomization group. When  user deactivates and activates the survey, that session data causes the malfunctioning of the function "createFieldMap".

Are you agree with me ? or should i create a bug report anyway ?

Shnoulle:
Hi,

1st : always create a bug report please :)
2nd : think we need to kill al $_SESSION in all browser when deactivate ... => yes seems a good idea to reset session when deactivate.

sieutruc:
yeap, it's done, here the link is : 

https://bugs.limesurvey.org/view.php?id=10102

Shnoulle:
Just one think before merge ... https://manual.limesurvey.org/Standard_for_Git_commit_messages

````
Fixed issue #10102: reuse the master field map in case of activating a survey with at least one randomization group
Dev: reset the session when deactivate fix the issue 
````

PR #399, opened 2015-12-18
Respect $config['updatenotification'] settings

wrenchpilot:
Uses the $config['updatenotification']  variable to properly show updates notifications.

Currently if $config['updatenotification']  is set to 'stable', the system shows 'unstable' updates anyway.


c-schmitz:
Setting has become obsolete.

PR #403, opened 2016-01-07
New feature: Add a default value to set save incomplete questions con…

danielgimeno:
New feature: Add a default value to set save incomplete questions configuration in config-defaults file.

Dev: Added value 'defaultAllowSave' in config-defaults file
Dev: Removed useless array_merge from newSurvey and editSurveySettings functions
Dev: fixed unused var dqata renamed to aData


danielgimeno:
This commit does not apply to LS3 Develop branch. The array_merge from newSurvey and editSurveySettings functions are not present in LS3 version, so is not applicable to it.

PR #405, opened 2016-01-08
New feature: Remote API, new User, UserGroup and Permissions handling…

gabrieljenik:
… routines

Added routines for managing Users, Permissions and UserGroups.
Source code was ported from the core controllers.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2704,4 +2704,896 @@ class remotecontrol_handle
         }
         return false;
     }
+    
+    /**
+     * RPC routine to create a new user
+     * @param string $sSessionKey
+     * @param array $aAttributes
+     * @return array Array with status, uid, password
+     */
+    public function add_user($sSessionKey, $aAttributes, $sendMail = true, $sendPassword = true)
+    {
+        // Check Session
+        if (!$this->_checkSessionKey($sSessionKey))
+        {
+            return array('status' => 'Invalid Session Key');
+        }
+
+        // Check Permission
+        if( !Permission::model()->hasGlobalPermission('users','create') ) {
+            return array('status' => 'No permission');
+        }
+
+        // Validate UserName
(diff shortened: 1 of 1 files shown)


c-schmitz:
The patch is duplicating code from core. Instead the original code should be properly refactored so it can be used both by core and API without the necessity to be duplicated.

gabrieljenik:
I agree with you.
Core functions should call API methods.
Or there should be a central lib.

But aren't all other API methods duplicating code already?

c-schmitz:
The never ones shouldn't - the old ones porbably do.

gabrieljenik:
hmm don't agree that much with that approach... I mean, it is avoiding having new features while following the same pattern of all other API methods...

But is fine, I will see what I can do...

Thanks!

Shnoulle:
Core LimeSurvey controllers + API must call model function , no ?

gabrieljenik:
HI Shnoulle,

I agree, bouth should call the model... but what about validations (I would call them business validations)... they should be inside the model, inside a central lib or duplicated in both the API and the Controller?

If I recall correctly, I am using models, but some validations are outside the model.

As per the pluginevent, I saw it... pretty nice.. Are you suggesting to put this in there?

c-schmitz:
Yes, model function please.

Shnoulle:
@gabrieljenik : no, not here ... i make a pull request, but you seems to use a lot RemoteCOntrol and fix a lot of issue. Then you can be our Lead Dev on the RemoteControl system ? ANd help me to move it to a core plugin too maybe.

gabrieljenik:
@Shnoulle: Thanks! It will be great to get that "title". But the issue is that I am not sure I will have the time, at least on the first half of 2016, to take care of that position. I will try to move your development to core...

maziminke:
@c-schmitz , I need these feature for an upcoming project soon. Is there any chance we can merge it into 2.5x? These are only new functions so there should be no side effects or the like.

c-schmitz:
I don't see this patch being accepted anytime soon if the mentioned problems are not fixed first.

PR #407, opened 2016-01-11
Fixed issue #10208: Rendering any PDF, CSV, HTML introducing numerical 

GuillaumeSmaha:
See bug declared on https://www.limesurvey.org/en/forum/french-forum/103021-affichage-r%C3%A9ponse-multiple-entr%C3%A9es-num%C3%A9riques


Aestu:
There is a bug in the bugtracker too:
https://bugs.limesurvey.org/view.php?id=10208

Shnoulle:
Hi,

Seems OK, just one think. Can you have a look at https://manual.limesurvey.org/Standard_for_Git_commit_messages and make a new commit message : 
Something like
````
Fixed issue #10208: Rendering any PDF, CSV, HTML introducing numerical
````

Then it goes to our changelog :) with appropriate author.

**Comment by GuillaumeSmaha** [comment]
I updated the title and the commit message. Is it good for you ?

**Comment by GuillaumeSmaha** [comment]
Also fix the issue : 
https://bugs.limesurvey.org/view.php?id=9648

**Comment by GuillaumeSmaha** [comment]
Do I need to implement "num_value_int_only" or it will be done in another patch ?

**Comment by Aestu** [comment]
Code enabling integer_only property for multiple numerical input was already in this pull request:

https://github.com/LimeSurvey/LimeSurvey/pull/413/files

By the way, both patches cover the same functionality if you add this functionality to yours. We have to choose one of them. @Shnoulle , @c-schmitz , what do you think?

**Comment by Shnoulle** [comment]
@Aestu : this one seems better. Other pull use getExtendedAnswer for "code" question type too.
For @GuillaumeSmaha : i like to have different patch for different bug/feature. BUT i'm not the lead dev.

**Comment by Aestu** [comment]
@Shnoulle : The problem is that this patch doesn't solve the problem when pdf is generated by the own user with "Print answers" functionality (after taking the survey) and doesn't solve a problem in javascript validation that @danielgimeno detected.

In fact, @danielgimeno was already modifying his patch to take into account your comment about getExtendedAnswer. In my opinion, modifying getExtendedAnswer use in @danielgimeno patch is easier than merging code into this patch.

**Comment by Shnoulle** [comment]
1st : i really think we musrt not use Global function. ALl this global function are from LS 1.87 or before . We need to remove it.
2nd : here we touch of answer : then must be in answer model for a new function.
3rd I don't see the  update/fix on the patch ?
4th : one patch ; one bug/feature

**Comment by GuillaumeSmaha** [comment]
@Aestu 
1) I checked for the PDF printed by the user, the values are correct because PrintanswersController.php calls function getFullResponseTable (from common_helper.php) which calls getExtendedAnswer. So https://github.com/LimeSurvey/LimeSurvey/pull/413/files#diff-48eb7edba3e2af076bba04777db710deR193 is not necessary. 
2) For the javascript, I don't need to fix it because I didn't implement num_value_int_only for multiple numerical. (See my previous question if I need to implement it or not in the same patch)

**Comment by Aestu** [comment]
Ok, let's go on with this PR. We close the other one.

@GuillaumeSmaha : As @Shnoulle say, we can complete functionality in others PR in the future so I don't think you have to implement num_value_int_only in your PR, but we can wait for @Shnoulle opinion anyway.

**Comment by Shnoulle** [comment]
num_value_int_only id another feature. Here it's PDF rendering : 1 feature => 1 commit => 1 PR ;)

**Comment by Aestu** [comment]
Great. If this PR is ready, could you merge it @Shnoulle ?

**Comment by Shnoulle** [comment]
Hi, i didn't do any test then can not merge .... ask to Carsten

**Comment by c-schmitz** [comment]
Please have a look at https://manual.limesurvey.org/Coding_guidelines
At least the variable naming needs to be corrected.

**Comment by GuillaumeSmaha** [comment]
@c-schmitz I apply the coding guidelines.

**Comment by Aestu** [comment]
@c-schmitz : There are variable names that don't follow coding guidelines but they were already in the code. I don't see any new variable that doesn't follow guidelines.

**Comment by GuillaumeSmaha** [comment]
@Aestu  I updated the variable names in the new function getShortAnswer

File: application/helpers/admin/export/SurveyObj.php [hunk 1]
```diff
@@ -366,9 +379,39 @@ public function getFullAnswer($fieldName, $answerCode, Translator $translator, $
...
+        switch ($sFieldType)
+        {
+            case 'K':
+            case 'N':
+                if (trim($sValue)!='') {
+                    if (strpos($sValue,".")!==false) {
+                        $sValue=rtrim(rtrim($sValue,"0"),".");
+                    }
+                    $aQidAttributes = getQuestionAttributeValues($aQuestion['qid']);
+                    if (isset($aQidAttributes['num_value_int_only']) && $aQidAttributes['num_value_int_only']) {
```
Shnoulle: Just a remark here : num_value_int_only can be set after survey activation. Then some user can have float value with num_value_int_only .
If the value is an int : double rtrim do the job. Force to be int can remove some information.

GuillaumeSmaha:
@Shnoulle Ok, but there are the same function in this file https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/common_helper.php#L1456

Shnoulle:
You're right , maybe a bug here too :). 
Exporting it's not really same than view . getExtendedAnswer is used only in response browse i think (and really old, lot of different fix , must be reworked some time).

GuillaumeSmaha:
By the way, number_format has an issue with high number.

GuillaumeSmaha:
@Shnoulle I remove number_format

Shnoulle:
Thanks, i don't say we must remove it. I just think of some potential issue ...

And:
If already set to integer before survey start : the double trim ,system fix it .

Thanks you

GuillaumeSmaha:
Arf bad rebase ><

GuillaumeSmaha:
Ok rebase fixed ..

GuillaumeSmaha:
So, are you agree to merge this pull request ?

GuillaumeSmaha:
@Shnoulle @c-schmitz @olleharstedt @Aestu  @SamMousa Can somebody confirm this fix ?

PR #409, opened 2016-01-11
fix issue #10213

Spycomb:
make the button send post request instead of get request


Spycomb:
are you sure about the "save" and "close" button on the top of "return to survey summary", because that seems weird to me, what do you think??

PR #413, opened 2016-01-11
Fixed issue #10208 render PDF, CVS, HTML in next situations:

danielgimeno:
Fixed issue #10208 render PDF, CVS, HTML in next situations:

- When render view of all survey answers
- When exported PDF, CVS, HTML, etc formats
- When is enabled print own answers and export them in a PDF etc [complement to pull request number #407]

Fixed issue #10208 when administrator exports a survey in PDF, CSV, HTML, etc format, numeric data is exported with 10 zeros, no matter if you selected only integer or not.
Dev: Added posibility to set integer or float values in multiple numeric questions in settings.
Dev: Added fixnum_checkconditions (Javascript) to multiple numeric answers to check if is allowed introduce decimal mark.

### DIFF
--- a/application/helpers/admin/export/Writer.php
+++ b/application/helpers/admin/export/Writer.php
@@ -321,6 +321,7 @@ abstract class Writer implements IWriter
                 $value = $response[$column];
                 if (isset($oSurvey->fieldMap[$column]) && $oSurvey->fieldMap[$column]['type']!='answer_time' && $oSurvey->fieldMap[$column]['type']!='page_time' && $oSurvey->fieldMap[$column]['type']!='interview_time')
                 {
+                    $value = getExtendedAnswer($oSurvey->id, $column, $value, $sLanguageCode);
                     switch ($oOptions->answerFormat) {
                         case 'long':
                             $elementArray[] = $this->getLongAnswer($oSurvey, $oOptions, $column,$value);
--- a/application/helpers/qanda_helper.php
+++ b/application/helpers/qanda_helper.php
@@ -3113,6 +3113,21 @@ function do_multiplenumeric($ia)
 
     $answer_main = '';
 
+    if (trim($aQuestionAttributes['num_value_int_only'])==1)
+    {
+        $acomma="";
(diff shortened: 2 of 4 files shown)


danielgimeno:
Pull request number #407 (by @GuillaumeSmaha) is incomplete. The case when user is allowed to print his own answers and export them in Pdf format is not covered.

application/helpers/admin/export/Writer.php
```diff
@@ -321,6 +321,7 @@ protected function transformResponseValue($value, $fieldType, FormattingOptions
                 $value = $response[$column];
                 if (isset($oSurvey->fieldMap[$column]) && $oSurvey->fieldMap[$column]['type']!='answer_time' && $oSurvey->fieldMap[$column]['type']!='page_time' && $oSurvey->fieldMap[$column]['type']!='interview_time')
                 {
+                    $value = getExtendedAnswer($oSurvey->id, $column, $value, $sLanguageCode);
```
Shnoulle: I don't think it's OK to return getExtendedAnswer for 'short' answers. And i'm sure we need another function. getExtendedAnswer do "Long answer [answer code]" for single choice

application/helpers/qanda_helper.php
```diff
@@ -3165,7 +3180,7 @@ function do_multiplenumeric($ia)
                     $answer_main .= $dispVal;
                 }
 
-                $answer_main .= '" onkeyup="'.$checkconditionFunction.'(this.value, this.name, this.type);" '." {$maxlength} />\n\t".$suffix."\n</span>{$sliderright}\n\t</li>\n";
+                $answer_main .= '" onkeyup="'.$checkconditionFunction.'(this.value, this.name, this.type, \'onchange\','.$integeronly.');" '." {$maxlength} />\n\t".$suffix."\n</span>{$sliderright}\n\t</li>\n";
```
Shnoulle: It's another issue here ? ANd i think we must remove all inline javascript ... too complicated to manage in template or in plugin.

PR #414, opened 2016-01-12
New plugin event "createStatistics" for pdf statistics

KRav007:
- insert event for statistics pdf creation.This event could also used in future for xls & html
- todo's insert for xls / html if make sense, insert layout features
- rename variable $surveyid to new naming rule

### DIFF
--- a/application/helpers/admin/statistics_helper.php
+++ b/application/helpers/admin/statistics_helper.php
@@ -2929,21 +2929,25 @@ class statistics_helper {
             // set default header data
             // Since png crashes some servers (and we can not try/catch that) we use .gif (or .jpg) instead
             $sHeaderlogo = 'statistics.gif';
+            $iHeaderLogoWidth = 10;
+
 
             /**
              * Plugin event 'createStatistics'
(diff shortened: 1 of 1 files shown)


Shnoulle:
It's more for filename, meta and header, right ?
Actauuly we can use it to do own stat, but need a App->end() ;)

KRav007:
>It's more for filename, meta and header, right ?

Yes that's right. In first step it used it for the header and filename. In future projects you could spread the features to layout of the pdf.

>...but need a App->end()

Sorry, but what you by that? Is something in code missing or wrong?

Shnoulle:
For the App->end : you can do "What you want" in your plugin rendering is not started (and header)

For example : you can replace the PDF stats by the doc stats (or anything you want)

It's right for any plugin event :)

LouisGac:
By the way, in 2.5, statistic is now using views. So it's much more easy to read and modify the rendering.

KRav007:
good to hear - but for my current projects not relevant cause of to much migration time. Next Project ... :-)

PR #416, opened 2016-01-15
Fixed issue #9644: Unable to preview survey when using database sessions - link bug 09592

GuillaumeSmaha:
Dev: Call toQuoteValue with PDO::PARAM_LOB to get directly a bytea value

I sent a pull request on Yii framework to fix this issue : https://github.com/yiisoft/yii/pull/4004
But there is also a modification on LimeSurvey code to fix it.

### DIFF
--- a/application/config/config-sample-dblib.php
+++ b/application/config/config-sample-dblib.php
@@ -35,7 +35,7 @@ return array(
 
         // Uncomment the following line if you need table-based sessions
         // 'session' => array (
-            // 'class' => 'system.web.CDbHttpSession',
+            // 'class' => 'application.core.web.DbHttpSession',
             // 'connectionID' => 'db',
             // 'sessionTableName' => '{{sessions}}',
         // ),
--- a/application/config/config-sample-mysql.php
+++ b/application/config/config-sample-mysql.php
@@ -36,7 +36,7 @@ return array(
 // Make sure MySQL max_allowed_packet setting is large enough, some surveys generate over 2 MB of session data.
 /*
         'session' => array (
-            'class' => 'system.web.CDbHttpSession',
+            'class' => 'application.core.web.DbHttpSession',
             'connectionID' => 'db',
             'sessionTableName' => '{{sessions}}',
         ),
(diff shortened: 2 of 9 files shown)


Shnoulle:
@c-schmitz we already have some patch for Yii. But we can update too Yii quoteValue in application.

GuillaumeSmaha:
OK, I will update my patch.

Shnoulle:
Hi, i don't think fix framework , but Yii use public function :).

The fix framework in application : https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/db

GuillaumeSmaha:
@Shnoulle Ok, I understand.
So, I edit  application/core/db/DbConnection.php instead of framework/db/CDbConnection.php
I create a file application/core/web/DbHttpSession.php

In config.php, I use :

			'class' => 'application.core.web.DbHttpSession',

instead of

			'class' => 'system.web.CDbHttpSession',


But how do you manage the next update to notify users to use  'application.core.web.DbHttpSession', instead of 'system.web.CDbHttpSession' ?


And I find in kcfinder a condition using system.web.CDbHttpSession :
https://github.com/LimeSurvey/LimeSurvey/blob/master/third_party/kcfinder/core/bootstrap.php#L366
May be something will be break ?

Shnoulle:
Hi,

But core/db/DbConnection extend framework/db/CDbConnection.php https://github.com/GuillaumeSmaha/LimeSurvey/blob/fix_session_expire_postgresql/application/core/db/DbConnection.php#L3 

Then https://github.com/GuillaumeSmaha/LimeSurvey/blob/fix_session_expire_postgresql/framework/db/CDbConnection.php#L573 can fixed without updating config ? No , I made an error.

Sorry : i want to do spome test : but https://bugs.limesurvey.org/view.php?id=9644#c34284 no bug with actual GIT and PostgreSQL 9.1.18

GuillaumeSmaha:
1) Yes, I need the function quoteValue in db connection, but I notice I should use quoteValue($str, $quoteParam) without optional parameter for compatibility.

2) I try without updating the configuration and the class is not called. For example, DbConnection is defined in the internal configuration :
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/config/internal.php#L77

3) I answered to the post and how reproduce the error

Shnoulle:
Oh yes , extend CDbHttpSession, DbConnection is alreay extended .

Think it's OK

PR #417, opened 2016-01-15
Re-added Yii framework License file.

cebe:
https://github.com/yiisoft/yii/blob/master/LICENSE

>  * Redistributions of source code must retain the above copyright
   notice, this list of conditions and the following disclaimer.

:)


cebe:
@c-schmitz @Shnoulle @SamMousa ping.

PR #420, opened 2016-01-19
unable to use Survey::model in Plugin cron : 2nd solution

Shnoulle:
Dev: Survey::model use Permission, and Permission use session
Dev: use alternative soilution : don't call getUserId when get permission


Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/pull/419 : this one seem more clear, but 419 allow to add self::getUserId for other function

Shnoulle:
Thanks @SamMousa ;)

PR #408, opened 2016-01-11
Fixed issue #10221: High values are wrong on rendering

GuillaumeSmaha:
Number_format returns a bad value with number greater than 9999999999999998.

Example of calls with number_format :

    echo "Number : ".$a."\n";
    echo "Format : ".number_format($a, 0, '', '')."\n";
    echo "Fixed  : ".number_format_double($a, 0, '', '')."\n";

Results :

	Number : 9999999999999998
	Format : 9999999999999998
	Fixed  : 9999999999999998

	Number : 9999999999999999
	Format : 10000000000000000
	Fixed  : 9999999999999999

	Number : 12345678901234567167
	Format : 12345678901234567168
	Fixed  : 12345678901234567167

	Number : 12345678901234567168
	Format : 12345678901234567168
	Fixed  : 12345678901234567168

	Number : 12345678901234567169
	Format : 12345678901234567168
	Fixed  : 12345678901234567169

	Number : 12345678901234567170
	Format : 12345678901234567168
	Fixed  : 12345678901234567170

	Number : 99999999999999999999
	Format : 100000000000000000000
	Fixed  : 99999999999999999999

PR #421, opened 2016-01-20
Update feature: Improvement priceFormat

GuillaumeSmaha:
Dev: Remove String.fromCharCode: Key map != Ascii map !
Dev: Allow Home & End key
Dev: Add option to allow CtrlKey: default: true
Dev: Add option to keep cursor at the same place except for the decimal part: default: true
Dev: Add option to shift cents into integer: default: false
Dev: Add option to start directly with the integer part: default: true

*It doesn't work on input with int only or thousand separator and int only because priceFormat is called twice in https://github.com/LimeSurvey/LimeSurvey/blob/master/scripts/numerical_input.js#L17

### DIFF
--- a/third_party/jquery-price-format/jquery_price_format.js
+++ b/third_party/jquery-price-format/jquery_price_format.js
@@ -28,7 +28,12 @@
             clearSufix: false,
 			allowNegative: false,
 			insertPlusSign: false,
-			clearOnEmpty:false
+			clearOnEmpty:false,
+			keepCursorPosition: true,
+			startWithInteger: true,
+			shiftCentsToInteger: false,
+			allowCtrlKey: true
+
 		};
 
 		var options = $.extend(defaults, options);
@@ -58,6 +63,10 @@
 			var allowNegative = options.allowNegative;
 			var insertPlusSign = options.insertPlusSign;
 			var clearOnEmpty = options.clearOnEmpty;
+			var keepCursorPosition = options.keepCursorPosition;
+			var startWithInteger = options.startWithInteger;
+			var shiftCentsToInteger = options.shiftCentsToInteger;
+			var allowCtrlKey = options.allowCtrlKey;
 			
 			// If insertPlusSign is on, it automatic turns on allowNegative, to work with Signs
 			if (insertPlusSign) allowNegative = true;
@@ -65,7 +74,30 @@
 			function set(nvalue)
 			{
 				if(obj.is('input'))
-					obj.val(nvalue);
+				{
+					var dObj = obj[0];
+					if(keepCursorPosition && obj.is('input') && dObj.selectionStart == dObj.selectionEnd)
+					{
+						var str = obj.val();
+						var pointPosition = dObj.selectionStart;
+
+						var diffThousandSeparator = 0;
+						if(str.indexOf(thousandsSeparator) !== -1 || nvalue.indexOf(thousandsSeparator) !== -1)
+						{
+							diffThousandSeparator = str.length - nvalue.length;
+							pointPosition -= diffThousandSeparator;
+						}
+
+						obj.val(nvalue);
+
+						dObj.selectionStart = pointPosition; 
+						dObj.selectionEnd = pointPosition;
+					}
+					else
+					{
+						obj.val(nvalue);
+					}
+				}
 				else
 					obj.html(nvalue);
 			}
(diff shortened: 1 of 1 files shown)


Shnoulle:
Update the jquery-price-format, then send it to Sam.

For twice call a .not('.integeronly') do the trick :)

GuillaumeSmaha:
@Shnoulle Twice priceFormat should be fixed in the PR https://github.com/LimeSurvey/LimeSurvey/pull/412

GuillaumeSmaha:
This also fix a bug with Previous button.

For example : 2 groups. On the first : 1 numerical input with thousand separator and on the second whatever.
Enter the number 200.00. Click Next
Click Previous: The number is convert into 2.00 due the previous version of priceFormat which always take the 2 last digits as the decimal part


May be https://bugs.limesurvey.org/view.php?id=10211

GuillaumeSmaha:
Bug found : Int only & thousand separator : Too much zeros.

GuillaumeSmaha:
Ok no bug ... Just my test case which doesn't work.
I had twice calls of priceFormat for the inputOnly integer.

maziminke:
Denis, can you look into this and merge if everything looks fine?
Thanks!

Shnoulle:
No ,

Because if i merge , last time there are a bug in it : i'm in cahrge of correction. And actually : i don't want charge on it.

maziminke:
Carsten, do you have balls :-)

Am 19.02.2016 um 10:12 schrieb Denis Chenu:

Shnoulle:
Assigned to @olleharstedt because he start to work on it. There are 2 other pull request for fix thousand_separator ;)

third_party/jquery-price-format/jquery_price_format.js
```diff
@@ -65,7 +74,30 @@
...
+						dObj.selectionStart = pointPosition; 
+						dObj.selectionEnd = pointPosition;
+					}
+					else
+					{
+						obj.val(nvalue);
+					}
+				}
 				else
 					obj.html(nvalue);
```
olleharstedt: Please always add curly braces around all if- and else-statements.

olleharstedt:
If I enter `12.12` into a number field and then click backspace to remove it, the field will be filled with zeros.

third_party/jquery-price-format/jquery_price_format.js
```diff
@@ -201,13 +261,21 @@
 				// Minus Sign, Plus Sign
 				if (allowNegative && (code == 189 || code == 109 || code == 173)) functional = true;
 				if (insertPlusSign && (code == 187 || code == 107 || code == 61)) functional = true;
+
+				// Home, End
+				if (code == 35) functional = true;
```
olleharstedt: As above, please add curly braces.

olleharstedt:
Another issue: If I enter 12 and then click tab to next numerical question and enter 12, the result will be 120.

olleharstedt:
I merged this by mistake, it should still be open. Sorry.

GuillaumeSmaha:
@olleharstedt  Ok, I am fixing it

GuillaumeSmaha:
@olleharstedt If you revert the merge, I can create a new pull request.

olleharstedt:
It's reverted.

There's a lot of work on this feature right now, and it has a lot of issues... Shnoulle just did this one: https://github.com/LimeSurvey/LimeSurvey/pull/445

I would suggest:

1. One fix should be one commit. Now it's many fixes in one, which makes it harder to follow.
2. Please please make sure to link to bugs in the bug-tracker. It's not clear what is bug fix and what is a new feature implementation. If you split your merge requests between bug fixes and new features we can prioritize the bug fixes.

Thanks for your effort.

GuillaumeSmaha:
Ok thanks
It's already splited in few commits :
- https://github.com/LimeSurvey/LimeSurvey/pull/421 This pull request to enhanced the input 
behavior and have a friendly input to use
Now, there is a new pull request : https://github.com/LimeSurvey/LimeSurvey/pull/446
- https://github.com/LimeSurvey/LimeSurvey/pull/407 This aims to fix the drawing of the result : Draw 9 instead of 9.00000 for multiple numerical input in CSV, PDF and HTML
- https://github.com/LimeSurvey/LimeSurvey/pull/412 This fixes the value saved in the database to remove correctly the thousand marks and to convert the decimal mark into a point.
- https://github.com/LimeSurvey/LimeSurvey/pull/408 This fixes an issue with big number (issue with number_format)

PR #422, opened 2016-01-20
Fixed #10198: Unable to import tokens due to missing global surveys -…

mennodekker:
…> import permission

dev: do i need to fix currently set permission in a db patch?


mennodekker:
Not really  used to working like this. It seems my pull was auto updated with another commit. If you want me to do two pulls let me know and I will change it.

PR #425, opened 2016-02-02
Fixed issue #10266: After update to 2.5 with updated default template

Shnoulle:
- validate template in getTemplateConfiguration
-  don't test validity of standard template :not needed
- revert to standard/default if not 2.5 compatible


Shnoulle:
@LouisGac : here i deactivate old template. Maybe add a default config.xml ? if template is outdated ?

PR #426, opened 2016-02-02
New feature: Avoid activate a survey expired or in the future

danielgimeno:
Dev: Added an warning meesage indicating that a survey is in the future or expired. Also added a new buttons to update survey settings or activate anyway

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -427,6 +427,31 @@ class SurveyAdmin extends Survey_Common_Action
             $aData['failedgroupcheck'] = $failedgroupcheck;
             $aData['aSurveysettings'] = getSurveyInfo($iSurveyID);
 
+            if (empty(Yii::app()->request->getPost('activate'))){
+                $survey = Survey::model()->findByAttributes(array('sid' => $iSurveyID));
+
+                $todayDate = new DateTime('now');
+                $surveyExpireDate = new DateTime($survey->expires);
+                $surveyStartDate = new DateTime($survey->startdate);                
+
+                if (!empty($survey->startdate) && $todayDate < $surveyStartDate) {
+                    
+                    $aViewUrls['output'] = "<div class='messagebox ui-corner-all'><div class='warningheader'>".gT("This Survey is not yet started. Would like to change the survey settings?")."</div>";                        
+                }
+                else if (!empty ($survey->expires) && $todayDate > $surveyExpireDate) {
+                    $aViewUrls['output'] = "<div class='messagebox ui-corner-all'><div class='warningheader'>".gT("This Survey is expired. Would like to change the survey settings?")."</div>";                        
+                }
+                else {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Hi,

In 2.05/06 : we see if survey is active : different icon. Think it's OK, no need an alert ?

PS: 
1st : must test if expires or startdate is set before showing anything
2nd:  $_POST : use App()->request()->getPost('activate');
3rd: i really think this feature can be in plugin if needed (think it's not needed).

Aestu:
@Shnoulle We discussed about this issue in a developer meeting. I thought we should prevent activation but the group suggested this more conservative solution. I thought there was a consensus, we have already made the effort.

Shnoulle:
I don't rememeber th dev meeting, maybe i'm not here :). I don't understand : you mean we need an alert or not ?

Aestu:
I don't remember the day of the meeting but I'm sure I give Daniel this task because I had the conversation with the team. IMHO having an alert is very positive and can prevent problems.

Shnoulle:
Then : must be fixed : Yy getPost + must test if expires or startdate is set

danielgimeno:
Of course! ASAP! 
I'm sure this additional message will help

danielgimeno:
Done it.

application/controllers/admin/surveyadmin.php
```diff
@@ -427,6 +427,31 @@ public function activate($iSurveyID)
             $aData['failedgroupcheck'] = $failedgroupcheck;
             $aData['aSurveysettings'] = getSurveyInfo($iSurveyID);
 
+            if (empty(Yii::app()->request->getPost('activate'))){
```
olleharstedt: Any reason this blob is not in a separate function?

application/controllers/admin/surveyadmin.php
```diff
@@ -427,6 +427,31 @@ public function activate($iSurveyID)
...
+                $survey = Survey::model()->findByAttributes(array('sid' => $iSurveyID));
+
+                $todayDate = new DateTime('now');
+                $surveyExpireDate = new DateTime($survey->expires);
+                $surveyStartDate = new DateTime($survey->startdate);                
+
+                if (!empty($survey->startdate) && $todayDate < $surveyStartDate) {
+                    
+                    $aViewUrls['output'] = "<div class='messagebox ui-corner-all'><div class='warningheader'>".gT("This Survey is not yet started. Would like to change the survey settings?")."</div>";
```
olleharstedt: Why not use flash message?

Shnoulle: Because : show at same place where we see "ANonymous : YES/NO" etc ...
After you can add questoion etc ...

Seems better here : show the alert BEFORE activate :)

Shnoulle:
Why close it ?

danielgimeno:
Because it was developed on a 2.06 version and I decided to start a new 2.5 patch with new features.

PR #430, opened 2016-02-04
Fixed Issue #10311: Question group list is unsorted by default

zerwalter:
Added default sort to active data provider


LouisGac:
thank you. Don't hesitate to participate more. 
I should create a grid view for label sets when I have time, if you're interested : feel free to do it. 
https://demo.limesurvey.org/index.php?r=admin/labels/sa/view

zerwalter:
@LouisGac 
Ok, I will work on it.

PR #431, opened 2016-02-05
Update remotecontrol_handle.php

BertHankes:
Added RPC Routine to copy survey

It uses all internal functions and procedures of limesurvey.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2699,4 +2699,71 @@ class remotecontrol_handle
         }
         return false;
     }
+
+    /**
+     * RPC Routine to copy survey.
+     *
+     * @access public
+     * @param string $sSessionKey Auth credentials
+     * @param int $iSurveyID_org Id of the source  Survey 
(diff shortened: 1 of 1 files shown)


BertHankes:
Thanks to Denis Chenu for making a plugin that extends the RemoteControl function. http://extensions.sondages.pro/extendremotecontrol/ [^]

 This made my day. The requested RPC is now alive

Shnoulle:
Hi @BertHankes : I think it can be great to move all extend remote control system to a core plugin in 3.0 : each function in there own plugin, the core system only add an event for remote control.

If you think it can be interesting : maybe you can open a mantis bug in "devloppement", and if you are ready to develop this when nwe go for 3.0 : we can assign it to you on github :)

BertHankes:
I don't know what my contribution would be. Making extended RPC's?
The array of current RPC's IMHO is sufficient.
It was nice to implement my CopySurvey exention this way. In 3.0 it will become easier I guess.

B.t.w. the export_responses RPC by extendRemoteControl died on me. I think due to rights issues.. So I call my CopySurvey now by the extendRemoteControl and the export_responses 'direct'.

PR #433, opened 2016-02-07
DB error don't show whole information

Shnoulle:
* Yii always return'CDbConnection failed to open the DB connection.' increatePdoInstance
* adding quicly a new dbTest
* this must be done only when installing (public user don't have to see exact error) then not extend DBconnection


Shnoulle:
With new system : 
capture du 2016-02-07 13-55-57

Shnoulle:
Unsure it's the best method here :)

c-schmitz:
looks good to me, lets see how it works out

Shnoulle:
Maybe Yii 2 have better solution, i take a look for 3.XX release ;)

PR #435, opened 2016-02-09
Sometimes it can be impossible to move a questionafter another one.

Shnoulle:
* put max-height of the li
* Put some basic style for organizer : no color : theme can do it
* css transition for animate
* hide button if not needed : did it break (more) big survey ?

## Before ##
capture du 2016-02-09 16-31-57
capture du 2016-02-09 16-50-19

## After ##
capture du 2016-02-09 18-20-14


Shnoulle:
@olleharstedt Can you check with a big survey : i'm upset of https://github.com/LimeSurvey/LimeSurvey/pull/435/files#diff-a502228e39e27a4de77e019846f10ea1R76 part

Shnoulle:
@LouisGac : you test with a big survey ? Because of JS .each() for each question. Maybe can be "deactivated" for survey with more than XX questions

PR #437, opened 2016-02-10
Re-enable search survey by sid, special casing pgsql db instances.

sickpig:


LouisGac:
Thanks for the job.

Shnoulle:
CAST can not do the job ? CAST seem more clear for me

sickpig:
sure we can use cast() 

the '::' operator is just a shortcut

application/models/Survey.php
```diff
@@ -746,6 +746,8 @@ public function search()
         $criteria->join .='LEFT JOIN {{users}} AS users ON ( users.uid = t.owner_id )';
         $criteria->compare('t.active', $this->active, true, 'AND');
         $criteria->compare('surveys_languagesettings.surveyls_title', $this->searched_value, true, 'AND');
+        $sid_reference = (Yii::app()->db->getDriverName() == 'pgsql' ?' t.sid::varchar' : 't.sid');
```
sickpig: using cast we would have: 
``
  $sid_reference = (Yii::app()->db->getDriverName() == 'pgsql' ? ' cast (t.sid as varchar) ' : 't.sid');
``

PR #439, opened 2016-02-13
Update fullpagebar_view.php

BertHankes:
Present 'Create a new survey' button only when user had create rights


LouisGac:
Thank you for fix

PR #443, opened 2016-02-16
Dev: this fix a check which will prevent "any user" shares in CPDB

gmambro:
The original check was too strict to allow the share_uid = 0 used by "any user" shares

### DIFF
--- a/application/controllers/admin/participantsaction.php
+++ b/application/controllers/admin/participantsaction.php
@@ -746,7 +746,7 @@ class participantsaction extends Survey_Common_Action
     */
     function exporttocsv()
     {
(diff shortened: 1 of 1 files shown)


application/controllers/admin/participantsaction.php
```diff
@@ -746,7 +746,7 @@ function getSearchIDs()
     */
     function exporttocsv()
     {
-        if (Yii::app()->request->getPost('searchcondition','') != '') // if there is a search condition then only the participants that match the search criteria are counted
+        if (Yii::app()->request->getPost('searchcondition','') !== '') // if there is a search condition then only the participants that match the search criteria are counted
```
olleharstedt: Could you please add a comment in the code that '' == 0 and why we need !==?

PR #444, opened 2016-02-22
Fixed #10550: multi-line group description flows out of box

mfavetti:
Caused by having the group description be a p tag which may only contain inline elements. Changed it to be div for all 3 included templates so that group descriptions containing block level elements will not break the layout.


Shnoulle:
Yep, have the same issue at start of Skeletonquest :)

mfavetti:
I have no idea if I just fixed those merge conflicts correctly or not :-/ I merged the current master to my branch and fixed the conflicts, then committed the changes. Is this the right thing to do?

LouisGac:
yep, you're doing it right. 
thank you again for all your contributions, they are great.

PR #445, opened 2016-02-23
Fixed issue #09876: Different bogus with thousand seperator

Shnoulle:
Dev: Fixed : can not have thousand and no thousand in same page
Dev: Allow some fix/option in template
Dev: Allow fix and option in plugin
Dev: Allow empty, but disallow 0 : option when send the function (user without XSS can call twice)

### DIFF
--- a/scripts/numerical_input.js
+++ b/scripts/numerical_input.js
@@ -1,54 +1,96 @@
-$(document).ready(function () {
+
+/**
+ * replace the LEMval function, do it at ready to be sure em_javascript is loaded
+ * better to move this to em_javascript directly, no ?
+ */
+window.orgLEMval = window.LEMval;
+window.LEMval = function (alias) {
+    var varName = LEMalias2varName[alias.split(".", 1)];
+    var attr = LEMvarNameAttr[varName];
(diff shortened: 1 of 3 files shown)


Shnoulle:
No PHP fix needed : js do the job.

scripts/numerical_input.js
```diff
@@ -1,54 +1,96 @@
-$(document).ready(function () {
-    if (typeof LEMradix === 'undefined') { return; }
 
-    if (LEMradix == ',')
+/**
+ * replace the LEMval function, do it at ready to be sure em_javascript is loaded
```
olleharstedt: > do it at ready

It's not in ready?

olleharstedt:
One fix should be one commit. Not four fixes in one. It's possible to have more than one commit in one pull request, right?

Which bug is this?

> Dev: Allow some fix/option in template

Shnoulle:
thousand_epeartor are totally buggy, hard to fix only one part .....

For allow/fix options in template : https://github.com/LimeSurvey/LimeSurvey/pull/445/files#diff-e39de42c4425c990bedb3025ae48f9a4R21 is' a function, the a template function can replace it.

scripts/numerical_input.js
Shnoulle: Yes, outdated comment ... sorry.

Shnoulle:
Last git update fix the issue . @olleharstedt fix it

PR #447, opened 2016-02-24
Fixed issue: 'Unknown database driver' error during schema upgrade

odalecne:
In a SQL Server-based setup using dblib driver, this issue breaks the schema upgrade process when a primary key has to be dropped.


c-schmitz:
Thank you very much!

PR #448, opened 2016-02-24
Fixed issue: array declaration syntax incompatibility

odalecne:
The short array syntax is not compatible with the minimum supported PHP version stated in the installation requirements.


c-schmitz:
Thank you - again!

PR #450, opened 2016-02-26
Fixed token validation for the Token model.

sitawit:
The same reason as commit 79e2ffe194668ce961ec9295ea7fe5721edb4658.


c-schmitz:
Thank you!

PR #452, opened 2016-02-29
Fixed issue #10536: Higher risk that the emails are rated as Spam

Shnoulle:
* update to phpmailer 5.2.14
* remove X_PRIORITY
* Use phpMailer msgHTML to do the altBody
* add html if not exist : maybe a better control
* New feature : allow embedded image in mail (remove the servername : to do : update ckeditor config)

Question is : i use composer version of phpmailer, maybe a lighter version ? Remove example and doc ?

I put the new feature when i fix ckeditor to don't add http://exemple.org to email img src.


Shnoulle:
Carsten : question is "did we put whole php-mailer : more easy for us (and other dev can add Oaut-google system or any other without readding phpmailer) or did we use a liogther version (maybe just remove doc & example ?)

c-schmitz:
I think removing test & example is fine

Shnoulle:
OK, i do this.

I test : 
- only ascii in subject and text : the charset is set to us-ascii
- some french characeter (é è) in text/subject : set to utf8
- some arabaic character : utf8 : seems OK

All with/without picture.

I'm unsure for the html tag added. And i don't test From with accent actually.
If you have time to do some test with German character just after merging (i remoive the subject update by LS : PHPMailer do it)

PR #457, opened 2016-03-10
change link color in the help box

waltonam:
Changed the link color in the help box to make it readable. bug #10578


c-schmitz:
The 'Save and close' button should have the behaviour of the current save button,
The normal 'save' button should save and come back to the same screen (this time in edit mode, not in new mode, of course).

Also please note this: https://manual.limesurvey.org/Standard_for_Git_commit_messages

PR #460, opened 2016-03-11
Improvements to plugin manager admin page

mfavetti:
- Changed the back button to say "Return to admin panel" instead of "Close" for consistency with other pages

- Hooked up the rows per page dropdown

- Added sorting to the grid view (default sort is by plugin name, can also sort by active status or description)

- Changed appearance to match the other admin gridviews

Note:
I really didn't think having to change the view for some of these things was a nice approach, but the only alternative was to rewrite the controller and the model

### DIFF
--- a/application/controllers/admin/pluginmanager.php
+++ b/application/controllers/admin/pluginmanager.php
@@ -60,13 +60,7 @@ class PluginManager extends Survey_Common_Action
             }
         }
 
-        if(!is_null(Yii::app()->request->getParam('pageSize')))
-        {
-            $pageSize = intval(Yii::app()->request->getParam('pageSize'));
-            if(in_array($pageSize, Yii::app()->params['pageSizeOptions'])) {
-                Yii::app()->user->setState('pageSize', $pageSize);
-            }
-        }
+        $pageSize = intval(Yii::app()->request->getParam('pageSize', Yii::app()->params['defaultPageSize']));
 
         $aData['fullpagebar']['returnbutton']['url'] = 'index';
         $aData['fullpagebar']['returnbutton']['text'] = gT('Return to admin panel');
--- a/application/controllers/admin/pluginmanager.php
+++ b/application/controllers/admin/pluginmanager.php
@@ -60,7 +60,10 @@ class PluginManager extends Survey_Common_Action
             }
         }
 
-        $pageSize = intval(Yii::app()->request->getParam('pageSize', Yii::app()->params['defaultPageSize']));
+        if(isset($_GET['pageSize']))
+        {
+            Yii::app()->user->setState('pageSize',intval($_GET['pageSize']));
+        }
 
         $aData['fullpagebar']['returnbutton']['url'] = 'index';
         $aData['fullpagebar']['returnbutton']['text'] = gT('Return to admin panel');
--- a/application/controllers/admin/pluginmanager.php
+++ b/application/controllers/admin/pluginmanager.php
@@ -60,9 +60,8 @@ class PluginManager extends Survey_Common_Action
             }
         }
 
-        if(isset($_GET['pageSize']))
-        {
-            Yii::app()->user->setState('pageSize',intval($_GET['pageSize']));
+        if(Yii::app()->request->getParam('pageSize')) {
+            Yii::app()->user->setState('pageSize', intval(Yii::app()->request->getParam('pageSize')));
         }
 
         $aData['fullpagebar']['returnbutton']['url'] = 'index';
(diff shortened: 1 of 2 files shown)


Shnoulle:
Hi,
It's a really  great idea \o/ 
Sometimes have 4 page of plugin ....


With debug=2 : you don't have a PHP notice here : https://github.com/LimeSurvey/LimeSurvey/pull/460/files#diff-2d7ac9c7ae4330e5cc004cd17589303eR41 ?

I think here : https://github.com/LimeSurvey/LimeSurvey/pull/460/files#diff-515490083748f2521f61a7348a246dceR65

You can use : 
`Yii::app()->user->setState('pageSize',int_val(Yii::app()->request->getParam('pageSize',10)));`
getParam for GET or POST

Aestu:
The order of plugin loading is an interesting issue. Currently they load randomly, such as they are returned by sql query. It was one of the causes of:

https://bugs.limesurvey.org/view.php?id=9885

@Shnoulle , @c-schmitz , @SamMousa : Do you think forcing a loading order (for example alphabetical) would be a good idea? It would make app behaviour predictable but we don't want plugins that depend on the order they are loaded.

SamMousa:
No, plugins should load independently.

If they have a dependency they should fire custom events instead.

Shnoulle:
Here, it's only plugin manager list : improve here don't update anything else.

@SamMousa : I think in a future plugin system, have a require/depend system can be a good idea, no ?

SamMousa:
Yes !

mfavetti:
@Shnoulle 

I have made the change you suggested. However, even after double checking my php error settings (E_ALL, display_errors=on, debug=2, etc) I see no errors on the page. So I have no idea what the actual error was. Can you paste the error and let me know what settings you use so I will catch them in the future? Hard to fix it if I can't see it :P

Thanks

Shnoulle:
Hi,

I put the ? : it's a question. Looking at code, i see a test before pageSIze, ,then thinking pageSize can not be set ....
I didn't test .... sorry for missunderstanding.

PS : why not $pageSize = intval(Yii::app()->request->getParam('pageSize',10)); seems more clear than isset or other ?

mfavetti:
@Shnoulle 

Is this waiting on me for something? As far as I know it's ready to be merged.

Shnoulle:
@mfavetti no, sorry .... i just test after merging : we can not update "pageSize" with Firefox : 

Open plugin manager , set 20 rows per page : reset to 10.

mfavetti:
Ok, fixed.

Shnoulle:
But : pagesize is not send on each request ? It's the case with jqgrid. If not with yii gridview we musrt fine a good way 
See https://github.com/LimeSurvey/LimeSurvey/blob/2.06lts/application/controllers/admin/responses.php#L657 

Ps : can we update easily to POST request ? 
PS2: It's a way to test Yii gridview too , see : https://manual.limesurvey.org/Replace_jqGrid

Shnoulle:
I can confirm : pagesize is set all the time in GET request

Shnoulle:
OK, i merge it. But next time : remind the GIT commit message : https://manual.limesurvey.org/Standard_for_Git_commit_messages

mfavetti:
Thanks for the reminder on the commit messages, will do for next time.

Where are you seeing the pageSize on every request? I'm only able to see it being sent when the pageSize is changed via the dropdown. (yiiGridView.update is called) Otherwise it seems absent from GET/POST.

Also, can you explain why you want a POST request instead? I believe all you have to do is set ajaxType to "POST" on the gridview options array.

Shnoulle:
No, actually is $_GET
capture du 2016-04-01 08-59-28


There are issue with sorting by description or by name actually. I report the bug in mantis : you can take it ?

mfavetti:
I know it is using $_GET. I asked why you wanted to switch to $_POST and I suggested how it might be done.

I also asked where you see the $_GET request other than when changing the pageSize dropdown as this is the only time I see it, yet you stated that it sends a $_GET request on each page load.

As for the sorting issues, I searched on the bug tracker and couldn't find any plugin manager issues. What are the numbers?

Shnoulle:
Hi, sorry for missunderstanding .... :(

I see the pagesize each time on the request ,  don't you ? But with your solution :

1. Plugin manager
2. set pagesize to 100
3. got to a survey
4. plugin manager : pagesize still 100

Seems great :)

For the mantis bug : i don't report it actually , i put the link here when it's done.

And again : :+1: thank you

Shnoulle:
For $_POST : All $_POST is controlled via CRSF, not $_GET, we need CRSF if we update DB (else it's a bug). Not needed if we don't update, but i think it's always better to do it by default (if we add a DB update next year for example ...).

PR #462, opened 2016-03-15
Update index.php

occelebi:
Recently came across with this error. Selinux may block installation page.


c-schmitz:
I am really sorry but we will not include instructions on every particular flavor of Linux.

occelebi:
Allright please update error message then. Cos It was very annoying and time consuming. I just wanted to share with all as they do not have to tackle it again and again!

Shnoulle:
Se linux have a lot of other issue. ANd actually other user can want a better label only on this directory.

You can create a SELIUNX PART IN https://manual.limesurvey.org/Installation OR https://manual.limesurvey.org/Troubleshooting OR https://manual.limesurvey.org/General_FAQ

Our wiki is a .... wiki ;)

PR #463, opened 2016-03-16
Fix: token check should be case insensitive

gmambro:
While tokens should be case insensitive this check is still case sensitive, causing erratic behaviour when users enter a token in a different case than the one entered when the session was created.

### DIFF
--- a/application/controllers/survey/index.php
+++ b/application/controllers/survey/index.php
@@ -658,7 +658,7 @@ class index extends CAction {
 
     function _isClientTokenDifferentFromSessionToken($clientToken, $surveyid)
     {
-        return $clientToken != '' && isset($_SESSION['survey_'.$surveyid]['token']) && $clientToken != $_SESSION['survey_'.$surveyid]['token'];
+        return $clientToken != '' && isset($_SESSION['survey_'.$surveyid]['token']) && strtolower($clientToken) != strtolower($_SESSION['survey_'.$surveyid]['token']);
     }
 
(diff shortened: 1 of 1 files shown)


c-schmitz:
Tokens are supposed to be case-sensitive, we recently changed database scheme to make sure this is obeyed by the different database types, too.

Shnoulle:
And, if i remind : we add this test because we have erratic behaviour with mysql. @gmambro if you find an issue : better is to report bug.

gmambro:
I had to add the comparison because of an erratic behaviour!

c-schmitz:
As said, if you have problems please open a bug report - with a detailed description of your system and how to reproduce. Thank you!

gmambro:
I think there was a misunderstanding: are tokens supposed to be case sensitive? If so the patch was wrong of course, and maybe I've spotted a case when a wrong case token is accepted but soon after the user is given the session error instead of a more user friendly warning.

c-schmitz:
Yes, tokens are case-sensitive.

gmambro:
Maybe this is something I wasn't suppose to see in my token_xxxx table:  `token` varchar(35) COLLATE utf8_unicode_ci DEFAULT NULL,

c-schmitz:
Yes, correct collaton would be utf8_bin.

gmambro:
So some error has occurred during an upgrade on that DB with upgradeTokenTables181. Now DBVersion is 184.

PR #465, opened 2016-03-21
Fixed issue #10806: A lot of white space lost in survey and question edit

Shnoulle:
* Use always 12 columns, force html editor to max width of 100%
* add a css file for all admin theme
* use an input-group inside the col fix a lot of issue
* Why sometimes lable 4 and sometimes label 3 ?

capture du 2016-03-21 16-34-23
capture du 2016-03-21 16-34-35
capture du 2016-03-21 16-35-02
capture du 2016-03-21 16-35-14

Screenshot after the fix, for before fix : see bug report

### DIFF
--- a/application/helpers/admin/htmleditor_helper.php
+++ b/application/helpers/admin/htmleditor_helper.php
@@ -200,6 +200,7 @@
 
     function getInlineEditor($fieldtype,$fieldname,$fieldtext, $surveyID=null,$gID=null,$qID=null,$action=null)
     {
+        App()->getClientScript()->registerCssFile(Yii::app()->getConfig('styleurl')."htmleditor.css");
         $htmlcode = '';
         $imgopts = '';
         $toolbarname = 'inline';
--- a/application/views/admin/survey/Question/question_subviews/_tabs.php
+++ b/application/views/admin/survey/Question/question_subviews/_tabs.php
@@ -31,35 +31,33 @@
 
         <!-- Question Code -->
         <div class="form-group">
-            <div class="col-lg-offset-1">
-                <label class="col-sm-2 control-label"  for='title'><?php eT("Code:"); ?></label>
-                <div class="col-sm-5">
+                <label class="col-sm-3 control-label"  for='title'><?php eT("Code:"); ?></label>
+                <div class="col-sm-9">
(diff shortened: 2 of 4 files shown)


LouisGac:
hey Denis,

are the strech actions still working ? for left menu, and accordions ?

Shnoulle:
Fior the left menu : i send 2 screenshot : one with , one without .... Look at HTML editor in the second : width is set to 650px , but real width is less due to https://github.com/LimeSurvey/LimeSurvey/pull/465/files#diff-abfe9dbbb3ffadbfded9a7fa21d77c5cR2 I already use this system in 2.05. It's why i added the div class='html' to have a max-width on the ckeditor iframe.

LouisGac:
Thank you for your screenshots.

The left menu can be closed (like in your screenshot), but it can be stretched too (my question was about the stretched state). The right accordion can be stretched too. I'm asking because I don't see any change to the JavaScript in your commit (maybe it's not needed?)

Here the JavaScript to open/close/stretch the left menu and the right accordions : 
https://github.com/LimeSurvey/LimeSurvey/blob/master/styles/Sea_Green/scripts/sidemenu.js

Also, we yet did quiet a lot of changes to this part last week with Carsten (he asked for some specific ratios) : 
https://github.com/LimeSurvey/LimeSurvey/commit/54efb3527fd00ca5958bd865d32f73bf7d633503

Shnoulle:
left menu have z-index, right menu is after. No change on this part was needed.

LouisGac:
Ok, I just read in detail your commit, seems ok. 

Just for you to know, as you can see here:
https://github.com/LimeSurvey/LimeSurvey/blob/master/styles/Sea_Green/scripts/sidemenu.js#l269-l288

The strech/unstrech of accordion is based on width animation. The unstretched action apply a width in percent corresponding to the number of bootstrap column of the accordion:
https://github.com/LimeSurvey/LimeSurvey/blob/master/styles/Sea_Green/scripts/sidemenu.js#l276

So when you change the accordion width, the javascript must be updated. That why the sidemenu.js is in Admin Themes, not in js core. So  integrator can create its own admin theme with specific ratios.

Shnoulle:
i made my test with Bay of many

Think you can : 
* update class span5 to span12
* use css animate only method

PR #466, opened 2016-03-21
New feature : newUnsecureRequest event : direct request without CRSF

Shnoulle:
* this allow extend of remote control : add a demo plugin
* this allow other system using POST request (exemple SOAP)
* The plugin must subscribe to newUnsecureRequest : think the event name is clear enough

@c-schmitz : you speak when see the forum post about extendRemoteControl of a way to do it without updating config.php 

With this : we can make a little remote control system just adding functions in plugin.
- https://github.com/LimeSurvey/LimeSurvey/pull/431
- I have another pull request, send it tomorrow here.


SamMousa:
Do we also have a feature to get the CSRF token via the `LimesurveyApi` object? Because if the request is not a REST request we should definitely add a `getCsrfToken()` to `LimeSurveyApi` object so people who make web interfaces that do `POST` can benefit from it.

Shnoulle:
With Yii 2.0 we can decativate CRSF in action, not in 1.X version. Only in config of the application. Here we deactivate CRSF for plugins/unsecure in internal.

Actually you mean : to do a REST api : 
- call getCRSF of the API
- send CRSF + data to directRequest

This solution offer to deactivate CRSF to unsecureRequest then 
- send data to unsecureRequest directly.

:)

Shnoulle:
@KRav007 and @Jan-E : you idea on this ? I think too we can offer plugin with
* remoteManageSurveys by plugin (create survey, create/update group and question)
* remoteManageResponses
* remoteManageTokens

etc ....

Then an user who just need to send email by remote control can deactivate remoteManageSurveys and remoteManageResponses for example : more security because less action is open.

SamMousa:
Rest api does not need csrf since it does not use a session.

Shnoulle:
Yes, right :) still JSON RPC or SOAP or ? do $_POST

gabrieljenik:
I think I understand the approach about extending remote control via plugin w/unsecure-direct-request.

Nevertheless, what if we could just create a method in the API which just triggers a directRequest event on the plugins? So, instead of generating a new entry point (extendedRemoteControl), we create a generic method on the existing entry points.

The ideas about moving API functions to plugins is nice too, but I guess a different evolution, right?

Shnoulle:
@gabrieljenik : you already have this event newDirectRequest BUT : crsf protection broke it. See https://git.framasoft.org/SondagePro-LimeSurvey-plugin/extendRemoteControl/blob/master/README.md#limesurvey-configuration .

You're right : another idea was to add a new event in remote control directly. But here : a new unsecureRequest allow

1. extend remotecontrol
2. create another API
3. allow to prefill survey by $_POST (for an example)
4. any other idea other dev can have.

Actually : less event : more option. With adding a new event to remotecontrol allow only the 1st point.

And : yes : moving API function is really another idea for 3.X release. But LS core can offer "getSessionKey/releaseSessionKey" only, and a plugin all other function. It's more a project idea :)

gabrieljenik:
About "_Actually : less event : more option._ " Sorry, didn't get what you were trying to say.

About "_With adding a new event to remotecontrol allow only the 1st point._", I parly disagree.
- If you can extend the remotecontrol, you are very close to being able to create a new API too.
- "_Any other idea other dev can have_". can be implemented too.
- "_Allow to prefill survey by $_POST (for an example)_", what you mean?

If I understand correctly, adding the unsecure event could be usefull too, not only for complementing the API, but to other potential developments.

Nevertheless, to extend the API, I beleive it is easier to use a new "pluginDirectRequest" method in the API, as that wouldn´t require to create code for handling requests, as the extendRemoteControl does (https://github.com/LimeSurvey/LimeSurvey/pull/466/files#diff-67d1116d08a448054c07a7f6647044e6R47).

So, concluding, I would split the request in 2...
- One side, for the unsecure request event
- Other side, pluginDirectRequest api method

What do you think?

Shnoulle:
I didn't understand really ?

We already have newDirectRequest . The remote control PLUGIN can add an event using unsecureDirectRequest.

You mean you like a newRemoteControlFunction event, right ?

gabrieljenik:
Sorry.

I meant:
- One side, for the unsecure request event in the plugin
- Other side, create new pluginDirectRequest method on the API, that would call newDirectRequest on the plugins.

If "_https://github.com/LimeSurvey/LimeSurvey/pull/466/files#diff-67d1116d08a448054c07a7f6647044e6R51 not needed_", how the plugin will invoke the methods in the custom RemoteControlHandler (which extends remotecontrol_handle) ?

Shnoulle:
Line 51, not 50 . MLine 51 is here to try to deactivate CRSF in plugin directly. Can be done in Yii2 (maybe) but not in Yii1.

In don't understand what you mean in real for pluginDirectRequest : you mean remotecontrol_handler call newDirectRequest ? In LS code or in the external tool code ?

gabrieljenik:
"_In don't understand what you mean in real for pluginDirectRequest : you mean remotecontrol_handler call newDirectRequest ? In LS code or in the external tool code ?_"

Yes, I mean creating a new api method in remotecontrol_handler. When that new method is called, it would call/trigger newDirectRequest event on the plugins.

Just to clarify. 
"_One side, for the unsecure request event in the plugin
Other side, create new pluginDirectRequest method on the API, that would call newDirectRequest on the plugins._"
I meant, 
- implementing the Unsecure event in the plugins, to make that an option for "_any other idea other dev can have_"
- implementing the new pluginDirectRequest method on the API, for allow developers to extend the API.

Does it make sense?

Shnoulle:
Removed the "demo" plugin. Re add it when have a better API

gabrieljenik:
Nice.

PR #472, opened 2016-04-06
New feature [statistics]: add answer ID toggle

smipth:
Dev: suggested in ticket [#10796](https://bugs.limesurvey.org/view.php?id=10796)

### DIFF
--- a/application/helpers/admin/statistics_helper.php
+++ b/application/helpers/admin/statistics_helper.php
@@ -2383,10 +2383,11 @@ class statistics_helper {
     * @param mixed $surveyid
     * @param mixed $sql
     * @param mixed $usegraph
+    * @param mixed $ansID
      *
      *
     */
-    protected function displayResults($outputs, $results, $rt, $outputType, $surveyid, $sql, $usegraph, $browse, $sLanguage)
+    protected function displayResults($outputs, $results, $rt, $outputType, $surveyid, $sql, $usegraph, $ansID=1, $browse, $sLanguage)
     {
 
         /* Set up required variables */
@@ -2739,7 +2740,13 @@ class statistics_helper {
                 }
 
                 //text for answer column is always needed
-                $fname="$al[1] ($al[0])";
+                //choice whether to show answer id
+                if($ansID==0){
+                    $fname="$al[1]";
+                }
+                else{
+                    $fname="$al[1] ($al[0])";
+                }
 
             }    //end if -> show aggregated data
 
@@ -3757,7 +3764,7 @@ class statistics_helper {
     /**
      * Generates statistics with subviews
      */
-    public function generate_html_chartjs_statistics($surveyid, $allfields, $q2show='all', $usegraph=0, $outputType='pdf', $pdfOutput='I',$sLanguageCode=null, $browse = true)
+    public function generate_html_chartjs_statistics($surveyid, $allfields, $q2show='all', $usegraph=0, $ansID=1, $outputType='pdf', $pdfOutput='I',$sLanguageCode=null, $browse = true)
     {
         $aStatisticsData=array();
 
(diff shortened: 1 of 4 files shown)


LouisGac:
Hi smipth,
Statistics is still a very messy code, hard to read and to maintain : so, congrats for the great work ;-)
Sadly, when testing your code, it generates a notice.
Please, test your code with "debug mode" set to 2 before submitting. 
As soon as you removed the notices I will merge your PR.

smipth:
Always plenty to do :) .  Hopefully this one loads without errors, I checked it with debug mode 2.

smipth:
Hey just to check back and see if we can get this merged in.  Thanks.

application/helpers/admin/statistics_helper.php
```diff
@@ -2383,10 +2383,11 @@ protected function displaySimpleResults($outputs, $results, $rt, $outputType, $s
     * @param mixed $surveyid
     * @param mixed $sql
     * @param mixed $usegraph
+    * @param mixed $ansID
      *
      *
     */
-    protected function displayResults($outputs, $results, $rt, $outputType, $surveyid, $sql, $usegraph, $browse, $sLanguage)
+    protected function displayResults($outputs, $results, $rt, $outputType, $surveyid, $sql, $usegraph, $ansID=1, $browse, $sLanguage)
```
olleharstedt: I am sceptical about adding yet another argument to an already too big function signature.

application/helpers/admin/statistics_helper.php
```diff
@@ -2383,10 +2383,11 @@ protected function displaySimpleResults($outputs, $results, $rt, $outputType, $s
     * @param mixed $surveyid
     * @param mixed $sql
     * @param mixed $usegraph
+    * @param mixed $ansID
```
olleharstedt: `mixed` or `boolean`? Please describe what the argument does.

Shnoulle:
Looking at stats for another prject ... this really must be reworked a lot ....

Some idea
[ ] Remove this awfull first letter system
[ ] Use a real oCriteria and not this sql string
[ ] Add getQuestionStats($iQid) function : return an array for the stats, depend on question type. iQid can be Question OR subQuestion (this allow , for example to have Y/N/null for multiple in one array OR all the multiple checkbox like actually.
[ ] Add displayQuestionStat($iQid,$type='html') : to add the render , depend of export type
[ ] A lot work to be done .... after

LouisGac:
[ ] Remove this awfull first letter system : 100% agree, cause of 95% of statistics bugs (still in 2.06)
[ ] Use a real oCriteria and not this sql string : 100% agree, will make EASY to build advanced stats (just adding filter to a search method...)
[ ] Add getQuestionStats($iQid) function: will be a method of the question object, absolutely
[ ] Add displayQuestionStat($iQid,$type='html') : same

Shnoulle:
I think future getQuestionStats($iQid) call a Question object method, no ?
Then can call Question Object method and set some option after (option by $_POST).

The , for example : Question object method return always same array (with null value for example), but statiticics->getQuestionStats remove the null part.

It's just the simpliest example i found.

LouisGac:
Still 100% agree. 

We were also thinking about adding the possibility to the question object to provide its own graph if needed. For example, if a totally new question type need a specific graph to make sense, statistics will fist look if the Question Object has its own method to render the graph.

Eg: let say someone write a question type, saying: "please click on this kind of things in this picture". The "making sense graph" for such a question type would be a Heat Map. 

Of course, the Question Objects itself can implement the graph rendering on the base of interfaces, so different question object can implement the same graph rendering.

Shnoulle:
And , if at a time we have "newStats" like we have newExport.
Example : for export in impress or odt or in json or on XML or ....

How the question object can know what to do ?

I think more rendering is a graph class part, not a question part :) . We are not 100% OK ;)

LouisGac:
we're just saying the same thing. whatever : it will be much more clear when we'll start to design the UML diagrams. So not for now :-D

LouisGac:
Not saying we're doing to do it exactly like that, but that give an idea of the typical way of resolving the problem you raised:
 
 https://en.wikipedia.org/wiki/Strategy_pattern

c-schmitz:
I don't see this being merged anytime soon and it has also become obsolete.

PR #473, opened 2016-04-07
Fixed issue #10850: Questions attributes "Display columns:" wrong def…

andmemasin:
…ault value 6->1

Changed the default value to 1.


c-schmitz:
Thank you very much!

PR #477, opened 2016-04-18
Auditlog events conf page

danielgimeno:
New feature: configuration AuditLog config page + new log survey settings change event

           Dev: Added AuditLog config page + default values for each operation type
           Dev: Log any survey setting change
           Dev: Fix missing dispatch event for update & delete tokens & partipants.


application/config/config-defaults.php
```diff
@@ -646,6 +646,20 @@
 // Use alias notation, we should move to this format everywhere.
 $config['plugindir']               = 'webroot.plugins';
 
+// AuditLog Default values
+$config['auditLog'] = array(
```
Shnoulle: auditlog is a plugin : please don't add system for specific plugin in core.

Each plugin setting allow a default value too

plugins/AuditLog/AuditLog.php
```diff
@@ -1,31 +1,100 @@
...
+			),
+			'AuditLog_Log_SurveySettings' => array(
+				'type' => 'checkbox',
+				'label' => 'Log if a user changes survey settings',
+			),
+		);
+
+
+		/**
+		* Get default values from config-defaults
```
Shnoulle: See previous comment.

PR #478, opened 2016-04-18
New feature #10954 : always add font-awesome in public and admin

Shnoulle:
* move top fa for public part
* admin part have a lot of glyphicon
* I think i fix whole public : next part : start the layout/public.html to replace getHeader en footer .....
* Before we have 5 times fa fonts ..... with different version .....


Shnoulle:
Tested and merged. Maybe some fix elsewhere , but tested complmete survey. See https://bugs.limesurvey.org/view.php?id=10954#c37428

PR #480, opened 2016-04-19
New feature: configuration AuditLog config page + new log survey sett…

danielgimeno:
…ings change event

Dev: Added AuditLog config page + default values for each operation type
Dev: Log any survey setting change
Dev: Fix missing dispatch event for update & delete tokens & partipants.

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -625,6 +625,12 @@ class tokens extends Survey_Common_Action
 
             foreach ($aData as $k => $v)
                 $token->$k = $v;
+
+            $beforeParticipantSave = new PluginEvent('beforeParticipantSave');
+            $beforeParticipantSave->set('model',$token );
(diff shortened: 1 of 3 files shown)


application/controllers/admin/tokens.php
```diff
@@ -902,6 +908,13 @@ function delete($iSurveyID)
             self::_newtokentable($iSurveyID);
         }
 
+        $token = Token::model($iSurveyID)->find('tid=' . $sTokenIDs);
```
olleharstedt: This breaks when `$sTokenIDs` is a list of ids (mark more than one token in token list and click delete). Please fix.

danielgimeno: Fixed. PR made.

Daniel.

2016-05-04 14:58 GMT+02:00 Olle Härstedt <notifications@github.com>:

PR #483, opened 2016-04-28
Fix some part of new datetime 

Shnoulle:
- Use fontawesome for datetime question
- Add a label for accessibilty : must test whole system with screen reader
- registerScript in replace of included script. Tested : seem not needed ?

### DIFF
--- a/application/views/survey/questions/date/selector/answer.php
+++ b/application/views/survey/questions/date/selector/answer.php
@@ -41,9 +41,16 @@
         */
         ?>
 
-        <?php $this->widget('yiiwheels.widgets.datetimepicker.WhDateTimePicker', array(
+        <?php
+            echo CHtml::tag("div",array('id'=>"answer{$name}_datetimepicker",'class'=>"input-group date"),"",false);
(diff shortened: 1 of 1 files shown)


c-schmitz:
See also https://eonasdan.github.io/bootstrap-datetimepicker/Options/#icons

Shnoulle:
No, because yiiwheels update html in PHP not only in JS : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/yiiwheels/widgets/datetimepicker/WhDateTimePicker.php#L78

You're right : alternate solution can be : 
Add our input only, ,and update it in JS only to add the input-group and icon.

The 2 solution are valid, and we have a third
- Use yiiwheels : actual solution
- Add an input and do it with some JS file and option inside HTML
- Create a LS extension for inpu (near same than SettingsWidget but with views etc ....)

c-schmitz:
You can update the YiiWheels widget directly for the surrounding HTML. I updated it anyway to accomodate the new datetimepicker library.

Shnoulle:
Then : it's not the official yiiwheel ? A sort of LS for of yiiwheel ?

c-schmitz:
The official yiiwheel is inactive, we use the latest version and I modified datetimepicker to get the bugs out.

PR #485, opened 2016-05-02
Labels with listradio answers where closed too early

michielvaneerd:
The LABEL tags for listradio answers where closed too soon, so the DIV tags with the answers fell outside and did not look right. This change fixes this.


LouisGac:
The HTML must remain outside of the label tag. It can contain advanced HTML (such as other div, p, images, arrays, etc. ), and break the HTML validity (it's explained in the comment itself). 

If you're experiencing display problems, try to correct them rather with CSS or js.

PR #488, opened 2016-05-08
New feature #09816: Using quotas to prevent new entries

Shnoulle:
Dev: readd the system like done in 2.06 less than 150831
Dev: little issue : don't happen at survey welcome page (plugin can do the job more easily)

- Since we have "active/unactiv" have this solution is more better than before
- But : a plugin can do the job really better : break BEFORE welcome page with less php file


Shnoulle:
Need https://github.com/LimeSurvey/LimeSurvey/commit/bda940c5c6d0ed70bbb6a05aae0c2c8a2470b5f7

PR #489, opened 2016-05-11
Fixed issue #10323: Rollback AuthLDAP to a version that supports user creation

georg-rath:
I backported the version from commit f630da9943340178d67edfbb55ae732f352af21b, which fixes the problem seen in the bug report.


emmarichardson:
So, do I need apply this patch to an older version of the AuthLDAP file or does it work on the new AuthLDAP?

c-schmitz:
Please check out https://bugs.limesurvey.org/view.php?id=10323 where a patched file was attached for the latest 2.50 version and report there if that works for you.

PR #490, opened 2016-05-12
Fixed issue #11015: Default Admin surveylist order  - Latest first

TonisOrmisson:


Shnoulle:
We have a "last view" survey . It can be really great if we can have "order by last viewed"

TonisOrmisson:
Displaying surveys by oldest first just got me frustrated a bit :)
Order by last viewed might be good also. Although - personally i would keep it simple and display order simply desc. by creation date. If the list is too dynamic then it might cause additional problems. E.g. if I have multiple people working on different projects (being a usual case for me) then they start to flip positions while being worked on simultaneously. 

Also - I see that there is the new active-surveys list, but even there Intuitively I look for top of the list looking for the latest things.

olleharstedt:
> Order by last viewed might be good also

Use the last viewed surveys plugin for this: https://github.com/olleharstedt/LastVisitedSurveys

Shnoulle:
@olleharstedt : yes but LastVisitedSurveys is only for the last 5. Here : if you need same for all survey list : you need a model for whole surveys .
And an event for "SurveyListing" or something for extension in SurveyListing.

This fix a very old feature request : order by default by last created.

LouisGac:
sorry Tonis, I did it without merging your PR.

TonisOrmisson:
no prob.

PR #491, opened 2016-05-13
better compare of EM value

Shnoulle:
- don't find real issue in it but
- before true =>$bNumericArg=false, and before this fix strval(floatval(true))===strval(true)
- if arg type is NUMBER : always set to bNumericArg
- @olleharstedt  Maybe linked to bug https://bugs.limesurvey.org/view.php?id=11163 for 2.06 lts (i make a test, manually revert this part)

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -270,8 +270,8 @@ class ExpressionManager {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


Shnoulle:
Fixed in https://github.com/LimeSurvey/LimeSurvey/commit/d3fd417e986fb9b10b69a6bcaae4fb0eda8f571c

PR #492, opened 2016-05-13
insert no-more-tables class to array-by-columns-table

KRav007:
in this question the class was missing

### DIFF
--- a/application/views/survey/questions/arrays/column/answer.php
+++ b/application/views/survey/questions/arrays/column/answer.php
@@ -8,7 +8,7 @@
  */
  ?>
 
-<table class="array-by-columns-table table-array-by-column table question subquestion-list questions-list">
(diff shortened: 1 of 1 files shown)


olleharstedt:
No. Array-by-column is rebuilt using Javascript.

KRav007:
hi ole - ok. is it come from core? or from theme?

olleharstedt:
Will check.

olleharstedt:
In template, e.g. `default/scripts/template.js`, function `replaceColumnWithDiv`.

KRav007:
ok thanks

olleharstedt:
Nice, you're in Hamburg. You should join on after-work sometime. :)

KRav007:
yeah.. do you have a babysitter :-) where are you go to after work?

olleharstedt:
We work near Kampnagel but go to different places.

PR #497, opened 2016-05-27
equalize the answer view

KRav007:
rewrite of the answer view in case of bootstrap classes and insert given new variables

### DIFF
--- a/application/helpers/qanda_helper.php
+++ b/application/helpers/qanda_helper.php
@@ -3263,6 +3263,10 @@ function do_numerical($ia)
         'integeronly'            => $integeronly,
         'maxlength'              => $maxlength,
         'suffix'                 => $suffix,
+        'sm_col'                 => decide_sm_col($prefix, $suffix),
+        'kpclass'                => $kpclass,
+        'name'                   => $ia[1],
+        'dispVal'                => $dispVal = '',
     ), true);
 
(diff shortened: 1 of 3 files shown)


c-schmitz:
Why does this include a change to the token bar ?

What is the patch supposed to fix? Is there a visual change? If yes, which one? Can you maybe attach a pictures to demo the effect?

KRav007:
bildschirmfoto 2016-05-27 um 15 31 56

Yes it's a visual change as you see the given html uses a nested colum without a row between. from point of view, this is not bootstrap like an avoid the padding/margin collapsing witch is working everywhere. In this case we have an other padding to the left side as other answers. In addition to that it's not useful to wrap a textfield into a p-tag.

token bar .. a git accident :-)

c-schmitz:
I cannot merge this due to the token include. PLease fix this and the commit message - see https://manual.limesurvey.org/Standard_for_Git_commit_messages

KRav007:
Done :-)

Shnoulle:
Please : don't merge, propose a patch for answers_html .

See https://github.com/LimeSurvey/LimeSurvey/commit/9ad97a65d98bb31242d93139e5bae88088f376c1#diff-73a6a2567eefb211495b1563055ddb59L17 for the new HTML flow.

PR #501, opened 2016-06-10
Make LimeSurvey compatible with phing.

Jelle-S:
We use phing + jenkins to do our deployments. Because phing, by default excludes files that it assumes are hidden files, the files third_party/kcfinder/js/000._jquery.js and third_party/kcfinder/js/002._jqueryui.js are not deployed with it (see https://github.com/phingofficial/phing/blob/master/classes/phing/util/DirectoryScanner.php#L120). I renamed the files (removing the underscore) and tested with a default installation, creating a survey and uploading an image using kcfinder to the description of the survey.


c-schmitz:
With all due respect but it is a minor problem to configure Phing (that scanner) accordingly so the files are not excluded. We cannot bend around every default setting of another software. On top of it this kcfinder is a third party library so the change will be lost on next update of the library.

daften:
This is not a minor problem. It is hard-coded in phing itself, which means we'd need to patch phing. For some items, that is the correct approach. However in this case, the convention to name files with ._ as start of the extension is very bizarre. The files just need to be included first, they are not hidden files. ._ files are (as far as I know) only used by OS X as metadata files. So changing it at the root seems like the best solution.
Seeing as the latest commit in kcfinder is from 2 years ago and it's essentially dead in the water, I don't think this fixed would be reversed that easily.

There is no question here to bend around third-party software, but a question to keep in line with filenaming conventions.

I'd ask you to reconsider this PR. There are no downsides to it as far as I can see.

c-schmitz:
I am sorry but I don't see this as hardocded in Phing.
Check out the documentation on this: http://www.phing.info/docs/stable/hlhtml/index.html#FileSet
It can be easily configured in your Phing configuration.

daften:
And like I said, that is besides the point. The way the files are named goes against standard filenaming conventions, as far as I can see. Maybe we are the first to report this, but I am 100% we will not be the only ones. And as this has no disadvantages, I don't see why it shouldn't be merged ...

Shnoulle:
What is the standard file naming you talk ? I know only : start by dot for hidden files.

But maybe you can give standard : https://en.wikipedia.org/wiki/Category:Filenames don't show any standard.

Jelle-S:
@Shnoulle See http://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them and https://www.diigo.com/annotated/0c936fda5da4aa1159c189cea227e174

These files will be created when, for examle, you use a FAT formatted hard drive on OSX

Shnoulle:
And ? I didn't see any standard here ? 
.Thumbs is not a standard : a lot used but not a standard

Jelle-S:
From https://www.diigo.com/annotated/0c936fda5da4aa1159c189cea227e174:

> Second, the other ._ files:
>
> Suppose that "MyFolder" is not on an HFS disk, but is instead on an
older MS-DOS or (even better) a UFS disk - one that doesn't support a
lot of file metadata.  Then further support that some program on your
Mac OS X system sets filesystem feature that your disk does not support.
 This could be anything in the HFS Plus set: not *just* a resource fork,
but also file type, creator type, any unsupported date (creation date,
modification date, last accessed date), file label, ACL or other
arbitrary xattr metadata (in Mac OS X 10.4 or later), old-style file
comments, old-style Finder info, and even for some file systems, Unix
permissions (owner ID, group ID, and owner/group/world permissions).
>
> It is not acceptable for a Mac program to set these values and have them
vanish into the ether - it can make programs break rather badly, in
fact.  The OS wants all programs to work as well on all volume formats
as possible.  So, if a program sets *any* piece of file data on a file
system that does not natively support it, the Mac OS X code to implement
that file system stores it in an invisible "companion" file for the
file.  In our example, that now gives us this:
>
> MyFolder:
    .DS_Store
    MyFile
    ._MyFile
>
> The name for the companion file is the same as the original name
preceded by "._" and, if the file system supports it, set to be
invisible.  This is so that the OS can always find the companion file
based on the original file's name.  The name starts with "." because
many operating systems (most Unix and Linux variants) normally treat
such names as invisible, even if they don't support an explicit
"invisible" flag in metadata.

Jelle-S:
Also see the last sentence of the "The Macintosh file system" paragraph here: https://en.wikipedia.org/wiki/Resource_fork#The_Macintosh_file_system

c-schmitz:
Anyway, there is no need to discuss this anymore. It can be easily configured in Phing despite @daften's claim that it is 'hardcoded'. Just configure it properly and you are good. You are of course free to talk to the KCFinder guys and ask them to change this - as said it is a third party library we just use. I consider this matter to be closed. Thank you!

PR #506, opened 2016-06-15
Fixed issue #09828: Ranking question : update allowed can broke Surve…

Shnoulle:
…y DB

- add a new QuestionAttribute maxDBanswers, not updatable if survey is activated
- fix max_answers according to this attribute
- Have only maxDBanswers in survey runtime, then other are not set
- Sub question by answer number OR attribute in createFieldMap
- todo : some fix specific to EM, but maybe just in qanda and JS 
- todo: set maxDBanswers when update and when import LSA


Shnoulle:
For EM : i think i remove this from EM and do it only in qanda/viewer : EM don't need it : EM don't know the columns after maxanswers .

Shnoulle:
Havbe some issue : i fix it and redo the pull request

PR #507, opened 2016-06-15
Master 09828 ranking fix

Shnoulle:
- add a new QuestionAttribute maxDBanswers, not updatable if survey is activated
- fix max_answers according to this attribute
- Have only maxDBanswers in survey runtime, then other are not set
- Sub question by answer number OR attribute in createFieldMap
- fix specific to EM, max_answers set with min(max_answers,maxDBanswers)
- todo: set maxDBanswers when update and when import LSA : actually : done just before updating Answer parts

### DIFF
--- a/application/controllers/admin/database.php
+++ b/application/controllers/admin/database.php
@@ -154,7 +154,7 @@ class database extends Survey_Common_Action
             if($sQuestionType=="R" && Survey::model()->findByPk($iSurveyID)->active=="Y")
             {
                 $oQuestionAttributeMaxDBanswers = QuestionAttribute::model()->find(
-                    "qid = :qid AND attribute = 'maxDBanswers'",
+                    "qid = :qid AND attribute = 'max_subquestions'",
                     array(':qid' => $iQuestionID)
                 );
                 if (empty($oQuestionAttribute))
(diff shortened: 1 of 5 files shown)


Shnoulle:
OK : think have all except the DB update part (but fix it before deleting answers if i don't make errors : i retest to be sure).

application/controllers/admin/database.php
```diff
@@ -149,30 +149,31 @@ function index($sa = null)
...
-            $__max_db_answers = QuestionAttribute::model()->findByAttributes(array(
-                'qid' => $iQuestionID,
-                'attribute' => '__max_db_answers'
-            ));
+            /* for already activated survey and rank question type : fix the maxDbAnswer before deleting answers */
+            /* @todo : add it to upgrage DB system, and see for the lsa */
+            if($sQuestionType=="R" && Survey::model()->findByPk($iSurveyID)->active=="Y")
+            {
+                $oQuestionAttributeMaxDBanswers = QuestionAttribute::model()->find(
+                    "qid = :qid AND attribute = 'maxDBanswers'",
```
olleharstedt: You are mixing naming conventions. "Maximum answers" is "max_anwsers", so I'd expect "maxDBanswers" to be "max_db_answers".

olleharstedt:
What does this PR do that my commit didn't?

application/controllers/admin/database.php
Shnoulle: Yes, you're right, send the commit after.

Shnoulle:
Allow adding more answer than the answer number at start.

Or : have 100 answer to chose but only 3 columns in DB. I can send a little lss file but must be tested with this commit.

Shnoulle:
Naming conventyion fixed , but wait : i think there are a bug in LS2.50 with min_answers and max_answers

In 2.06 : you can use EM in same page. I think now : it don't work. Let me make a lss file + a bug report and fix it here.

Exemple survey : 

* MAX: Numeric question : "How many fruit do you love" 
* RANK : Rank your {MAX} loved fruit in order

Shnoulle:
OK : 2.06 is better for such error, but still some issue. I send bug report today, but completely different issue

Shnoulle:
Upload a lsa done with this pull merged
https://framadrop.org/r/ePiquT9z9v#TZk3FuCgqmd3QvDHOLZ/iawgbJFouCGJSrBAmO3cq78=

PR #511, opened 2016-06-17
Fixed #11404: EM tips are always empty after reloading page via browser

Shnoulle:
- use JumpTo to actual step without save value
- Same fix for previous from browser


Shnoulle:
Oups, pushed ... https://github.com/LimeSurvey/LimeSurvey/commit/92051fb070fae4d1a10bc342a262719a37429746

PR #516, opened 2016-07-05
Fix crash with "Undefined variable: esrow" when creating a label set

rloomans:
Fixes #11429

I'm not sure that this is the *right* fix... but it allowed me to create new label sets.

### DIFF
--- a/application/views/admin/labels/editlabel_view.php
+++ b/application/views/admin/labels/editlabel_view.php
@@ -51,7 +51,7 @@
                                 }
                                 Yii::app()->getController()->widget('yiiwheels.widgets.select2.WhSelect2', array(
                                     'asDropDownList' => true,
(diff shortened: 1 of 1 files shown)


Shnoulle:
:+1:  Thanks,
Just : we need 1 language minimum : can you add 'required' attribute the the input ?

````
'htmlOptions'=>array('multiple'=>'multiple','style'=>"width: 80%",'required'=>'required'),
````

And in fact : i think you can remove whole $row test : it's really not needed :)

rloomans:
@Shnoulle do you mean that the `unset($aAllLanguages[$esrow['language']])` never needs to happen at all?

Shnoulle:
In fact : yes : i don't think $row is sent for label here. And i don't understand wgy we need to remove it :)

PR #519, opened 2016-07-06
Fixed Issue #11435: Sometimes importing errors are not shown correctly

gabrieljenik:
Sometimes, the error message shown is just "Array".


gabrieljenik:
Commit message is wrong. Will be placing another PR.

PR #520, opened 2016-07-06
Fixed Issue #11435: Sometimes importing errors are not shown correctly

gabrieljenik:
Survey Import: Sometimes, the error message shown is just "Array".

### DIFF
--- a/application/views/admin/survey/importSurvey_view.php
+++ b/application/views/admin/survey/importSurvey_view.php
@@ -16,7 +16,12 @@
 
         <!-- error message -->
         <p>
-            <?php echo $sErrorMessage; ?>
+			<?php
+            if (is_string($sErrorMessage))
+                echo $sErrorMessage;
+            else
+                print_r($sErrorMessage);
+            ?>
         </p>
 
         <!-- buttons -->
--- a/application/views/admin/survey/importSurvey_view.php
+++ b/application/views/admin/survey/importSurvey_view.php
@@ -17,13 +17,13 @@
         <!-- error message -->
         <p>
 		<?php
- -        	if(is_array($aImportResults['error']))
+ -        	if(is_array($sErrorMessage))
 			{
-				foreach($aImportResults['error'] as $error) 
+				foreach($sErrorMessage as $error) 
 					echo $error."<br/>";
 			} else 
 			{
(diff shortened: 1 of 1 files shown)


Shnoulle:
Strange, i think we already fixed this , but only in 2.06 ? https://github.com/LimeSurvey/LimeSurvey/commit/2c61f969b626e0eaa52d3225252bbda1e95a3a99
in don-like the print_r, maybe : an ul/li is better :)

gabrieljenik:
How about now? :)
Used the sime from the commit you sent me.

Shnoulle:
Sorry @gabrieljenik : i don't take time to test it really (and don't see other code) .... in fact must be tested with all import system combinaison .... and find all way the specific import system add error ...

gabrieljenik:
What I was mentioning is that I ported the commit you mentioned me from 2.06 to the master branch.

Shnoulle:
Yes, saw this :). 
This last one surely fix the issue. Must rework the system i think :)

Shnoulle:
Here : its' strange : https://github.com/LimeSurvey/LimeSurvey/commit/2c61f969b626e0eaa52d3225252bbda1e95a3a99#diff-caa320c4f7a39345c091b01e20977165L17

We can not have : globally sucess , just some error ?
Maybe (i don't remind) $aImportResults['error'] is always set in $sErrorMessage in controller.

Don't know :)

gabrieljenik:
Actually, yes, that's how it is working...
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/surveyadmin.php#L1034

Not sure if it is the best as it is, but at least this patch will help showing the errors.

application/views/admin/survey/importSurvey_view.php
```diff
@@ -16,7 +16,16 @@
 
         <!-- error message -->
         <p>
-            <?php echo $sErrorMessage; ?>
+		<?php
+ -        	if(is_array($sErrorMessage))
```
Shnoulle: Didn't see this .... syntax error
Fixed in https://github.com/LimeSurvey/LimeSurvey/commit/44d3a784efd920d834fe80103ffe990e0801eff4

PR #521, opened 2016-07-15
New feature : allow adding new question attribute in Plugin

Shnoulle:
- 1st solution : easy way for plugin : just need to put the array in protected $questionAttributes array
- translation is done by LS core after
- no event added.
- Think the second one is better
- Actually i use "Survey settings" to add specific attribute OR use cssclass ...
- Have a cleanest way seems really needed

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -3998,6 +3998,26 @@ function questionAttributes($returnByName=false)
         'default'=>0,
         "help"=>gT('Use button group or radio list'),
         "caption"=>gT('Display type'));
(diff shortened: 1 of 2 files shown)


Shnoulle:
Usage example : https://gist.github.com/Shnoulle/a2cc37ed527e3ddab1547c5ed90dfe8b

Shnoulle:
Not the good way. Adding attribute with event (in fact event happen only one time for each load). Maybe can use cache for this.

exemple
````
		'cacheSettings'=>array(
			'class'=>'system.caching.DummyCache',
		),
````
in internal. Admin can use the best cache he add.

PR #522, opened 2016-07-15
New feature : allow adding new question attribute in Plugin

Shnoulle:
- 2nd solution : using event newQuestionAttributes
- Usage $this->getEvent()->append('questionAttributes', $questionAttributes)
- Think this one is better (see https://github.com/LimeSurvey/LimeSurvey/pull/521 )
- Actually i use "Survey settings" to add specific attribute OR use cssclass ...
- Have a cleanest way seems really needed


Shnoulle:
Actually : 
- Usage of Survey settings : https://github.com/SondagesPro/LS-arrayTextAdapt
- USage of cssclass or another attribute : https://github.com/SondagesPro/LS-radioToStarRating

We already have all other think : then we just need this.
And after we can remove time_limit question attribute from core , and put it in a core plugin ;)

application/helpers/common_helper.php
```diff
@@ -3999,6 +3999,26 @@ function questionAttributes($returnByName=false)
         "help"=>gT('Use button group or radio list'),
         "caption"=>gT('Display type'));
 
+        /**
+         * New event to allow plugin to add own question attribute (settings)
+         * Using $event->append('questionAttributes', $questionAttributes);
+         * $questionAttributes=[
```
olleharstedt: I think question attribute should be an object, to enforce that all attributes are defined.

olleharstedt: Also, this function (questionAttributes) should really be called `getQuestionAttributes` and put into its own file.

olleharstedt:
Are you on vacation, Denis? For the plugin I'm working on I will need plugin localization. Want to design a system for that?

application/helpers/common_helper.php
Shnoulle: Actual solution : all existing attributes are defined by an array

````
        $qattributes["display_type"]=array(
        "types"=>"YG",
        'category'=>gT('Display'),
        'sortorder'=>90,
        'inputtype'=>'buttongroup',
        'options'=>array(0=>gT('Button group','unescaped'),
        1=>gT('Radio list','unescaped')),
        'default'=>0,
        "help"=>gT('Use button group or radio list'),
        "caption"=>gT('Display type'));
````

Then i just add a way to allow plugin to add a new attributes, using same system.

See https://gist.github.com/Shnoulle/f959ffba582a304bf13c110094dad996

Shnoulle:
Vacation : http://www.sondages.pro/a-propos/absences.html

I want to create a new clean template before (but have some issue from HTML core : https://bugs.limesurvey.org/view.php?id=11448

If core is not fixed : i do a plugin ... with is own view.

Translation come after.
For plugin : think priority is Question attribute before translation. But i can do some test/fix/improve if you want next week.

Shnoulle:
@olleharstedt about translation

1. Did we allow different system or not http://www.yiiframework.com/doc/api/1.1/CMessageSource 
2. Did we want plugin allowed to replace string by core (or not ?)
3. If not : maybe use http://www.yiiframework.com/doc/api/1.1/CMessageSource#onMissingTranslation-detail
4. I don't know if we can have multiple http://www.yiiframework.com/doc/api/1.1/CMessageSource with different basePath/method etc ... My idea is here

application/helpers/common_helper.php
olleharstedt: Sure, I just wanted you to consider creating a class to contain the data and behaviour instead of using arrays and loose functions. See this: https://github.com/LimeSurvey/LimeSurvey/blob/master/application/libraries/MenuObjects/MenuItem.php

olleharstedt:
I created a [wiki page](https://manual.limesurvey.org/Plugin_localization) where we can dump some ideas. I might finish during you vacation together with Carsten.

application/helpers/common_helper.php
Shnoulle: In Fact, i always sthink questionattribute definition must be in QuestionAttribute model (extend QuestionAttribute model with own rules etc ...)

Shnoulle:
@olleharstedt in fact : its' a settings, a definition of Question Attribute : i don't think it must be an Object. 

Settings definition seems better in array and not in object

Shnoulle:
.... Think i made an error .... the best is , maybe, add a newQuestionSettings like we do for newSurveySettings .
I make an,pother pull request (the third ;) )

Shnoulle:
After looking and looking again .... hard to find a good way to use newQuestionSettings. Yes, seems to be a good idea, but in fact : core attribute and plugin attribute must be "near same"

Todo : fix TSV export to include all attribute

Shnoulle:
I revert the last commit. Put it in QuestionAttribute are nit the best way (i think).

Have a libraries ls/questionseetings maybe. I see for this.

Still unsure for object vs array for settings.

PR #524, opened 2016-07-26
Refactor: Use Views for Token/Captcha handling at the beginning of a survey

lacrioque:
Refactored the Token/Captcha Site


Shnoulle:
I don't know if http://www.yiiframework.com/doc/api/1.1/CCaptcha is usuable. But maybe it's the best solution ?

c-schmitz:
Good idea!

PR #525, opened 2016-07-27
Fix RPC method: import_survey always throws error

tomasgvivo:
When there is no error, ```$aImportResults['error']``` equals ```false```.
This should fix it.


Shnoulle:
Maybe : we must set $aImportResults['error']=null; (and not false) in the import system. Then it fix this.
But right : this fix is needed :) 

see https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/admin/survey/importSurvey_view.php#L27

Shnoulle:
Thanks you

PR #528, opened 2016-08-26
Seems the only way to allow time in Sub Question

Shnoulle:
Fixed issue #09909 : Empty colums in export file while data visible i…n GUI

Dev: this fix allow to use time for export
Dev: only tested GUI, not remotecontrol
Dev: don't know for other export than CSV
Dev: don't really like this way ....


Shnoulle:
But i have another fix for allowing Time (it's OK) disallow time (Not OK)

Shnoulle:
Really dirty way , then  remove ?

PR #529, opened 2016-09-01
Template editor: PHP compatibility again: use xpath instead of '->' syntax

hermannschwaerzlerUIBK:
This is related to #518 
With PHP 5.3.x the '->' syntax of SimpleXML is not always working as expected.
Apart from that I consider the use of xpath() clearer than a chain of (differently overloaded) '->'s.

See also:
https://www.limesurvey.org/forum/installation-a-update-issues/104215-css-pages-not-available-not-visible

### DIFF
--- a/application/controllers/admin/templates.php
+++ b/application/controllers/admin/templates.php
@@ -857,14 +857,15 @@ class templates extends Survey_Common_Action
     protected function _initcssfiles(TemplateConfiguration $oEditedTemplate, $editable=false)
     {
         // If editable CSS files are required, and if they are defined in the template config file
-        if($editable && is_object($oEditedTemplate->config->files_editable->css))
+        $editable_css = $oEditedTemplate->config->xpath('files_editable/css');
+        if($editable && is_object($editable_css[0]))
         {
-            $aCssFiles = (array) $oEditedTemplate->config->files_editable->css->filename;
+            $aCssFiles = $editable_css[0]->xpath('filename');
         }
         // Else we get all the CSS files
         else
         {
-            $aCssFiles = (array) $oEditedTemplate->config->files->css->filename;
+            $aCssFiles = (array) $oEditedTemplate->config->xpath('files/css/filename');
         }
         return $aCssFiles;
     }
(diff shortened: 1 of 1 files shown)


olleharstedt:
We should really switch from XML to JSON for template config.

olleharstedt:
Wait. We don't support 5.3 anymore. Please update to PHP 5.4 at least.

hermannschwaerzlerUIBK:
OK, I see:
https://manual.limesurvey.org/index.php?title=Installation&type=revision&diff=68624&oldid=67972
Just like that? No announcement?

c-schmitz:
Some user edited the minimum req in the manual without our approval - just reverted it.

olleharstedt:
This breaks template editor for me. Will revert.

LouisGac:
yep xpath makes problems with PHP 5.3
I commented the code to explain the fix we did.

LouisGac:
see:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/TemplateConfiguration.php#L123-L125

hermannschwaerzlerUIBK:
Sorry for the problems caused. It worked for me. :-(
Regarding the json_encode -> json_decode workaround: 
In line 125 you are calling
`json_encode ( ( array ) simplexml_load_string($sXMLConfigFile), 1)`
Why are you using `1` as value for the `options` parameter of `json_encode` and not the equivalent `JSON_HEX_TAG` constant?

LouisGac:
hermann: don't worry. LimeSurvey works on thousands of various servers, with thousands of different configurations, so it's always extremely hard to manage those kind of PHP 5.3 bugs that varies from an installation to another....

Concerning your question, there is no particular reason (that just the common snippet you'll find to solve this problem). So if you think we should rather use JSON_HEX_TAG, let us know why.

hermannschwaerzlerUIBK:
Hi Louis,
well using a constant instead of a magic number is imho better because

- for anyone reading the code it's clearer what you want to do (looking up what `1` means in this context isn't exactly easy).
- you enhance compatibility - if the value of the constant changes for some reason in future releases of PHP you still get what you want.

PR #532, opened 2016-09-07
Fix and improve beforeTokenEmail

Shnoulle:
- This fix the issue : beforeTokenEMail don't happen for confirm (this MUST be, between a bug and a feature)
- Adding survey id (as survey) in the event. Really more easy
- Adding model : it's for db attribute. Not sure the best name. But without : plugin must rename $type for invitation and reminder.


Shnoulle:
Just for LS version , it's a new feature right, but an "hidden" feature and near a bug fix. Then maybe just minor 2.51.3)

PR #533, opened 2016-09-08
Validation answer without EM, before put it in session

Shnoulle:
- Staring point, because ALL question can be validated this way : see https://github.com/Shnoulle/LimeSurvey-Shnoulle/commit/b98d5ee2a534b39544562435bed41af481db81ba#diff-796aba4d06254b9caea9d461e4ea80c9R8756
- we can not use array filtering for this case : we filter by real answer in DB : but this can be done with _validateQuestion (if someone found a way to find the answer filtered in EM for single choice :/ )
- Here use a Class var, another idea was to use a App->user->(s|g)estState system (in session) : i'm not sure about the good choice ..... Class var seems cleaner BUT : All LEM is put in session, then need to be resetted. https://github.com/Shnoulle/LimeSurvey-Shnoulle/commit/b98d5ee2a534b39544562435bed41af481db81ba#diff-796aba4d06254b9caea9d461e4ea80c9R649
- move the numeric test type at same place : we show the error to the user, really better

Something great with session : Exemple of situation

1. Put something bad in a question
2. Move previous : the answer is set to null, but you can proceed. No alert is shown (must do ? or not , think better is not , see point 3)
3. Move next : you see again the question, but here with the error message : you know the answer is not saved.

In fact : except for decimal(30,10) : i think user must hack HTML to see the error. I think the class alert alert-danger is OK.
capture du 2016-09-08 09-24-43

I send a surey for testing : https://bugs.limesurvey.org/view.php?id=11611 . Another survey can be tested : https://bugs.limesurvey.org/view.php?id=10827


Shnoulle:
Shit ... must review something with "no answer"

application/helpers/expressions/em_manager_helper.php
```diff
@@ -6607,11 +6606,41 @@ function _ValidateQuestion($questionSeq,$force=false)
                     }
                 }
             }
-
+            /**
+             * Control value against value from survey : see #11611
+             */
+            if ($qrel)
```
olleharstedt: Please don't just insert more code, but break out functions instead and call them.

Shnoulle: At start , i completely add a new function to make validation but .... in fact ProcessCurrentResponse is done for this : validation responses.

I can create a new function again with an hundred  var .... (here 

````
foreach($sgqas as $sgqa)
{
  if($this->getValidityPost($qid,$sqga,$qhidden,....);
  {
    $stringToParse .= Yii::app()->getController()->renderPartial('/survey/system/questionhelp/tips', array('qid'=>$qid,'vclass'=>'dberror alert alert-danger','vtip'  =>$this->invalidAnswerCore[$sgqa]), true);
    $qvalid=false;
  }
}
````
because actual function need $stringToParse and $qvalid to be updated.

And thid need to be done in $sgqa.

You ask me to rewrite completely _ValidateQuestion .

Comment (olleharstedt) [inline application/helpers/expressions/em_manager_helper.php]: No, just that new code should preferably be put in separate functions.

File: application/helpers/expressions/em_manager_helper.php [hunk 2]
```diff
@@ -10246,6 +10188,97 @@ public static function convertNonLatinNumerics($str, $lang)
...
+                 * Array number with checkbox Only 1 or 0 (maybe reset 0 to '')
+                 * And surely a lot more
+                 */
+            }
+            return true;
+        }
+
+        /**
+         * return the actual validity string , and reset the variable used ($_SESSION)
+         * @param $sgqa : the SGQ (answer name)
```
olleharstedt: Should be

    @param string $sgqa

Shnoulle:
@c-schmitz : big update : only one function. All this test must be move to Question Object when ready. Leave it for now.  We can add some validation after (exemple array number / checkbox : only 1/0 and set 0 to '')

Shnoulle:
Can we merge it to develop version ? (i redo the pull request , no issue, but only if we merge it).

I think of another solution too : only filter "possible code" : 5 alphanumeric caracter (by default), but allow it out of the list + add a "only in list" like "mandatory" system (in fact : i first started with this). And we can use arry_filter too more easily in this step.

Shnoulle:
Move to 3.0.

BUT : need to be reviewed 2 different control
- User entered value : OK must fix it completely and send alert
- Hidden question (for example) not posted but set with Equation : must only filter silently .... and don't fix it completely ....

PR #534, opened 2016-09-10
Fixed issue #11764: Search box in main survey list is case sensitive

Aestu:
Dev: Also fixing the same problem for Token list


Aestu:
@c-schmitz : This is my suggestion for bug #11764

Aestu:
Wrong bug id

PR #535, opened 2016-09-10
Fixed issue #11574: Search box in main survey list is case sensitive

Aestu:
Dev: Also fixing the same problem in tokens list

### DIFF
--- a/application/core/LSDbCriteria.php
+++ b/application/core/LSDbCriteria.php
@@ -1,0 +1,40 @@
+<?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
+/*
+ * LimeSurvey
+ * Copyright (C) 2007-2011 The LimeSurvey Project Team / Carsten Schmitz
+ * All rights reserved.
+ * License: GNU/GPL License v2 or later, see LICENSE.php
+ * LimeSurvey is free software. This version may have been modified pursuant
+ * to the GNU General Public License, and as distributed it includes or
+ * is derivative of works licensed under the GNU General Public License or
+ * other free or open source software licenses.
+ * See COPYRIGHT.php for copyright notices and details.
+ *
+ */
+
+class LSDbCriteria extends CDbCriteria
+{
+    /**
+     * Basic initialiser to the base controller class
+     *
+     * @access public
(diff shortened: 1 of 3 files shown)


Aestu:
@c-schmitz : This is my suggestion for bug #11574

Shnoulle:
We must , then , add a different function : one case sensitive , one insensitive. No ?

Aestu:
I'm not sure about another function. With postgres, you can declare the object as "CDbCriteria" and you would get case sensitive comparisons. Of course, you can't use case sensitive and insensitive comparisons in the same query but, do you think we really need it?

Shnoulle:
Clearly

1. we **must** have same system on any DB
2. In search : we must use **case insensitive** criteria (and more accent-insensitive)
3. There are some request where we **need sensitive** (+ accent) criteria : token (when user put a token in input to enter a survey for example)

Then , i'm not sure for the solution, but 2 way to do it (i think)

- In the LSDbCriteria->compare : add a $sensitive param
- In the LSDbCriteria : add 2 function compareInsensitive compareSensitive.

Maybe :)

LouisGac:
I think it's a good PR, cause little by little, we're filling the models with exceptions for pgsql and mssql. 
Also: may be you could do a PR to the Yii core?

Aestu:
IMHO this PR is a clear improvement on current behaviour.

@c_schmitz: My opinion about your points.

1.    Having different systems for different DB is much better than filters behaving differently for different DB.
2.    I agree but we don't have to develop it all now
3.    Token verification doesn't use CDbCriteria. If it did, token verification would be case insensitive for mysql.

I can't find any use of a compareSensitive function of LSDbCriteria. If I'm wrong, could you please give me a specific example? if this function is needed, why doesn't it already exist for mysql?

By the way, sorry for not attending dev meetings. Very busy at work these last few months. I tried to make it today and I will try again next week.

Shnoulle:
Sorry : yes : i think it's a good PR :).

For token : i'm sure we have bug report and fix somewhere about : enter with TokenValue with tokenvalue : this send a "The token is not the same before ... etc ...." error. Think it's fixed now.

PR #536, opened 2016-09-20
Better i18n management for QuestionAttribute

Shnoulle:
Actually : it's hard to move some attribute to i18n if it's not true before : we lost all existing part, and this can broke admin part with debug=>1.

- [x] Fixed issue #11682 : Possible illegal string offset with some attribute
- [x] New feature #11683: EM validations tip must be i18n
- [ ] Only one way to search value (Question::model()->getAdvancedSettingsWithValues must use QuestionAttribute::model()->getQuestionAttributes


Shnoulle:
Better fix in https://github.com/LimeSurvey/LimeSurvey/commit/305cb4c662cbb3fd598dff25e9fbb79680f16071

PR #537, opened 2016-09-24
support for beforeSurveyBarRender 

yuriybabenko:
- minor formatting changes to be more consistent with surrounding code
- implementation of beforeSurveyBarRender plugin event; passes entire to-be-rendered data structure to the plugin for use/alter as needed (suggest similar solutions for all data structures prior to being rendered)
- view updated to check for presence of `$surveybar['extraMenuItems']` and rendering the items present there as new buttons in the SurveyBar menu


application/core/Survey_Common_Action.php
```diff
@@ -850,23 +850,19 @@ function _surveybar($aData)
 
             $aData['gid'] = $gid; // = $this->input->post('gid');
 
-            if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveycontent', 'read'))
-            {
+            if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveycontent', 'read')) {
```
olleharstedt: Our convention is to have brackets on a new line, like
```php
if ()
{
}
```

LouisGac: It's not "our" convention.
It's PHP convention:
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md

c-schmitz: @olleharstedt See https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md - which says that in control structures the { on the same line is fine. Next line brackets applie only to functions/methods and classes.

LouisGac: well, that's right, indeed they say:

> Opening braces for control structures MUST go on the same line, and closing braces MUST go on the next line after the body.

so @yuriybabenko is right.

Shnoulle: https://manual.limesurvey.org/Coding_guidelines#Structures when it was updated ? Sorry o follow the old coding guideline since ..... the old coding guidelins of LimeSurvey (3 years old or less ? https://manual.limesurvey.org/index.php?title=Coding_guidelines&oldid=43745#Structures )

olleharstedt: Probably inherit from C: https://www.kernel.org/doc/Documentation/CodingStyle

PR #540, opened 2016-09-26
Fixed Issue #11710: Add 'search box' on Home page survey list

mfavetti:


LouisGac:
I prepared the functionality and set it to false on purpose.
Could you please rather add a user control for it on the homepage setting ? 
By the way: thank you again for your commits, they are useful.

Shnoulle:
@LouisGac : i put the feature request : why deacvtivate it by default ? I think if a user want the list, he want the search box too ? Maybe default to true, global setting or user setting to set to false. But default to true seem really better (for me).

LouisGac:
because I wanted to first develop a user control for it. It's not that hard or long. 

The home page is thought to be customized as much as possible, targeting the non-advanced users, not the super admin. If a super admin want to create a very simple home page for his team, with one or two boxes, and a very short survey with only one or two surveys (so  without the search box), he can. He will have a super simple home page. 

If someone want the full survey list as home page, he just have to try to access to "admin/survey/sa/listsurveys" to login. So today, you yet can go to => ""admin/survey/sa/listsurveys"", bookmark it as your admin root page, and tadaaa!

So, forcing the complex listing on home page will remove a possibility, and will not add anyone. 

If you want to improve it: you must add a user control on the home page setting so admin can choose if yes or no the widget shows the search box.

Shnoulle:
Yes, but the list is not shown by default : this is OK. But if super admin want to show it (then it's not a simple home page) : seems OK to have the search box added.

The question is not "Survey list by default", it's "If survey list : search box by default" :)

LouisGac:
Ok, I know you're going to tell I'm repeating my self, because that's what I'm going to do: repeating myself. 

Today: if an admin want a direct access to the complete survey list: he can (login to : "admin/survey/sa/listsurveys", he can even configure it as default landing page in config file). If he want to show a very simple home page with a very short list of survey **without search box**: he can. 

**So this PR remove a possibility, and doesn't add a new one.** 

Please, just add a user control for the search box display on home page setting (I'd say it's less than one hour of coding).  If you don't want to do it: I will do it when I'll have time.

Shnoulle:
Maybe repeating yourself, but don't answer to the original issue ....

Home page with "BOX" + "Survey list" is great BUT survey list without search box is useless

LouisGac:
Use case:

A admin have a team of 10 users, each one having only two surveys. 
He want a simple landing page for them: just 2 boxes, and the two surveys et the bottom. He doesn't want to bother them with a complete CGridView instance (with searchbox, filters, pagination, etc.) for only two surveys. 

Why are we still debating about that?  In the time we spent debating about it, we could have implemented the user control in the home page settings already.

PR #545, opened 2016-09-28
New feature: NUMBEROFGROUPS replacement field on welcome template (#11716)

mfavetti:


Shnoulle:
See : https://bugs.limesurvey.org/view.php?id=11432

You're right : it can be interesting to have the total groupo number, but : why only in first page ? And wgy not adding STEP too (to show easily something like : you do {STEP} on {NUMBEROFGROUPS} ) or better : TOTALSTEP for exemple.

mfavetti:
Sorry, I didn't know there were pending issues related to this one. I just saw [11716](https://bugs.limesurvey.org/view.php?id=11716) and thought that it would be quick to implement. Since that issue made the reference to the NUMBEROFQUESTIONS replacement, I just did something similar to that. And it was only available on welcome template page. (In the survey helper first page function)

Shnoulle:
Yes : you did right :) . But i'm afraid of the number of core replacement each time an user want a new place.

I think i make a pull request quickly for [11432](https://bugs.limesurvey.org/view.php?id=11432)

Thanks for the mantis link :)

LouisGac:
very good job.
adding replacement keywords is never a problem.
also, avoid doing it with the plugin system for now. It's not correctly implemented (not MVC, not inheriting any Yii class), so they would  make the code more dirty, and harder to maintain.

Shnoulle:
>  adding replacement keywords is never a problem.
I can not be OK with this : all core replacement is done X times, each times it take some ms more, some are reallly used a lot, and some are never used. In fact a lot of our user's don't know the purpose of each replacement : for me some are totally unclear actually.

Adding again and again core replacement make whomle system more complicated ....

LouisGac:
For sure, you're right, we should avoid to add thousands of keyword. 

He just added one keyword. We'll probably add other ones (some have been asked in the forum). That makes custom templating easier for everyone. 

Also, if we speak about performance, be sure that adding more and more dirty core plugin will make more performances problem.

Shnoulle:
PS : and remind all replacement can go to $_SESSION big file : i'm unsure here but must be tested ....

jexller:
Any reason why this PR isn't merged?

olleharstedt:
This PR has a bug. Please always test with debug = 2.

LouisGac:
a notice?

mfavetti:
Sorry, my fault. I do have debug set to 2, but I just didn't happen to see the notice. I made #649 to fix this.

application/helpers/replacements_helper.php
```diff
@@ -683,6 +684,7 @@ function templatereplace($line, $replacements = array(), &$redata = array(), $de
     $coreReplacements['MOVENEXTBUTTON'] = isset($movenextbutton) ? $movenextbutton : '';    // global
     $coreReplacements['NOSURVEYID'] = (isset($surveylist))?$surveylist['nosid']:'';
     $coreReplacements['NUMBEROFQUESTIONS'] = $_totalquestionsAsked;
+    $coreReplacements['NUMBEROFGROUPS'] = $totalgroups;
```
maziminke: Where are such placeholders currently documented? I can't find NUMBEROFGROUPS at the manual.

As so often, a feature not documented doesn't exist for the common user.

PR #546, opened 2016-09-28
Fixed issue #11690: Mandatory indication is missing in question list

mfavetti:
Just a simple copy pasted function where not all parts where changed :o

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -816,7 +816,7 @@ class Question extends LSActiveRecord
     {
         if ($this->type != "X"  && $this->type != "|")
         {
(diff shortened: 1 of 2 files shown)


application/models/Question.php
```diff
@@ -816,7 +816,7 @@ public function getMandatoryIcon()
     {
         if ($this->type != "X"  && $this->type != "|")
         {
-            $sIcon = ($this->other=="Y")?'<span class="fa fa-asterisk text-danger"></span>':'<span></span>';
+            $sIcon = ($this->mandatory=="Y")?'<span class="fa fa-asterisk text-danger"></span>':'<span></span>';
```
LouisGac: wo! nice catch

PR #547, opened 2016-09-28
Updated feature: User can now choose whether or not to create sample question and group when a creating survey (#11605)

mfavetti:
See [11605](https://bugs.limesurvey.org/view.php?id=11605)

Implementation is up for discussion:
- Should default be to create or not create (bias towards inexperienced or experienced users?)
- Should this be moved to a global setting so it's not an extra click each time a survey is created

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -1666,33 +1666,12 @@ class SurveyAdmin extends Survey_Common_Action
             // Update survey permissions
             Permission::model()->giveAllSurveyPermissions(Yii::app()->session['loginID'], $iNewSurveyid);
 
-            $createSample = ((int) App()->request->getPost('createsample', 0) === 1) ? true : false;
+            $createSample = (int)App()->request->getPost('createsample', 0) === 1;
 
             if($createSample) {
-                // Now create a new dummy group
-                $aInsertData=array();
-                $aInsertData[Survey::model()->findByPk($iNewSurveyid)->language]=array(
-                    'sid' => $iNewSurveyid,
-                    'group_name' => gt('My first question group','html',Survey::model()->findByPk($iNewSurveyid)->language),
-                    'description' => '',
-                    'group_order' => 1,
-                    'language' => Survey::model()->findByPk($iNewSurveyid)->language,
-                    'grelevance' => '1');
-                $iNewGroupID=QuestionGroup::model()->insertNewGroup($aInsertData);
-                // Now create a new dummy question
-                $oQuestion= new Question;
-                $oQuestion->sid = $iNewSurveyid;
-                $oQuestion->gid = $iNewGroupID;
-                $oQuestion->type = 'T';
-                $oQuestion->title = 'Q00';
-                $oQuestion->question = gt('A first example question. Please answer this question:','html',Survey::model()->findByPk($iNewSurveyid)->language);
-                $oQuestion->help = gt('This is a question help text.','html',Survey::model()->findByPk($iNewSurveyid)->language);
-                $oQuestion->mandatory = 'N';
-                $oQuestion->relevance = '1';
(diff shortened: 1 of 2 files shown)


application/controllers/admin/surveyadmin.php
```diff
@@ -1665,36 +1665,49 @@ function insert($iSurveyID=null)
...
-            $oQuestion->mandatory = 'N';
-            $oQuestion->relevance = '1';
-            $oQuestion->question_order = 1;
-            $oQuestion->language = Survey::model()->findByPk($iNewSurveyid)->language;
-            $oQuestion->save();
-            $iNewQuestionID=$oQuestion->qid;
-
-            Yii::app()->setFlashMessage($warning.gT("Your new survey was created. We also created a first question group and an example question for you."),'info');
 
+            $createSample = ((int) App()->request->getPost('createsample', 0) === 1) ? true : false;
```
olleharstedt: There's never any reason to return true/false. `$something === 1` is enough.

application/controllers/admin/surveyadmin.php
```diff
@@ -1665,36 +1665,49 @@ function insert($iSurveyID=null)
...
+
+            if($createSample) {
+                // Now create a new dummy group
+                $aInsertData=array();
```
olleharstedt: This code below could be moved to a separate function.

mfavetti:
Thanks for the feedback

Shnoulle:
Think the default must be "checked" : https://github.com/LimeSurvey/LimeSurvey/pull/547/files#diff-9f28f3223a9fe9550530dbc272f2080eR51 then with a new installation : a new user have it.

You can leave like this for 1st step, we can add a "user setting" when someone can do it. @olleharstedt user setting table are in develop version ?

olleharstedt:
@Shnoulle It's in some branch

application/controllers/admin/surveyadmin.php
```diff
@@ -1705,6 +1697,51 @@ function insert($iSurveyID=null)
     }
 
     /**
+    * This private function creates a sample group
+    *
+    * @param mixed $iSurveyID  The survey ID that the sample group will belong to
```
olleharstedt: Nice work! Don't forget the `@return int` phpdoc.

Shnoulle:
@olleharstedt Think it can be in develop no ? I don't know for master :)

PS : i don't find te Reply.... after your answer .....

LouisGac:
I'd say we can merge it asap.
Let's see what Carsten will say.

maziminke:
Since adding the new dummy group was done at the current stable version I think we can also push this one to the stable branch if it was tested in depth.

I think there has to be a global setting to globally turn this feature off!

It is nice for new users to have that already added but every more experienced user (I can only speak for myself) may be pretty annoyed to always have to delete the dummy group.

Thanks for your continuous great work!

tammoterhark:
Maybe an idea: just use the previous value. So if user checked it in the previous survey, keep it checked. If the user unchecked it in a previous survey, keep it unchecked.

LouisGac:
a good idea indeed.

tammoterhark:
@LouisGac: this could also be an interaction design principle where applicable: for example in the sorting order of the survey list, in the question type( by default add a question in the same type as the previous one, when designing a survey).

I think that often it is easier to use a previous value than one that is just on top of the list.

tammoterhark:
You could call this: sticky defaults

LouisGac:
Tammo: it's called _adaptive user interface_, it's the new trend in UX. We're yet using it a little bit (for the quick action states for example).

tammoterhark:
Well, great minds think alike...

LouisGac:
Yep: the biggest issue in UX design in the different expectations between advanced users and newbees.
'_adaptive user interface_' is the solution to that problem.

PR #548, opened 2016-10-03
Fixed issue #11736: Uncaught ReferenceError: Decimal is not defined

mfavetti:
https://bugs.limesurvey.org/view.php?id=11736


LouisGac:
Thank you again for your great contributions

PR #550, opened 2016-10-06
Update question.pstpl

BertHankes:
Add a class  question-valid-message to the Question valid message.
This is for styling this specific element.

### DIFF
--- a/templates/default/views/question.pstpl
+++ b/templates/default/views/question.pstpl
@@ -12,7 +12,7 @@
         </div>
 
         <!-- Question valid message -->
-        <div class="row question-wrapper">
+        <div class="row question-wrapper question-valid-message">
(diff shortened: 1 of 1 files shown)


Shnoulle:
Template issue only.

Else : please for https://github.com/LimeSurvey/LimeSurvey/tree/answers_html

Shnoulle:
Sorry : but not when working on complet new template + HTML views

BertHankes:
Sorry, Keep up the good work!

Shnoulle:
You xan work on the new default template : in fact i don't want really work on it : i make it "new html" compatible only.

But if you think something can/must be in HTML produced by LimeSurvey (and not in template) : it's the time :).

We break totally the 2.50 template compatibility.

PR #551, opened 2016-10-06
Update template.css

BertHankes:
Added class question-valid-message 
to css.


Shnoulle:
Same than other https://github.com/LimeSurvey/LimeSurvey/tree/answers_html

PR #552, opened 2016-10-06
Added .row to gender q-view buttons

tammoterhark:
Now a PR to answers_html after several other attempts...

### DIFF
--- a/application/views/survey/questions/date/dropdown/answer.php
+++ b/application/views/survey/questions/date/dropdown/answer.php
@@ -15,7 +15,7 @@
 <!-- Date, dropdown layout -->
 
 <!-- answer -->
-<div class="question date answer-item dropdown-item date-item">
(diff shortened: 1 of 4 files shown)


Shnoulle:
OK for this time, but next one : remind [Git commit messages](https://manual.limesurvey.org/Standard_for_Git_commit_messages)

Here : Dev: Added .row ...

Seem OK

Shnoulle:
And : PS : i remove the cold-sm-6 from HTML actually : leave it for template. There are no reason to add one if we can not control it via attributes.

tammoterhark:
Great! Also great about col-sm-6. That is indeed better in media statement in css. Is design decision, not developers decision.

tammoterhark:
I learn day by day the tricks of developers. I might even become one... Thanks!

PR #553, opened 2016-10-07
Dev: Radio padding because without/with comments are differently aligned

tammoterhark:


Shnoulle:
Think you make a "branch error" : try to edit
capture du 2016-10-07 10-45-10

tammoterhark:
;-)

Shnoulle:
This don't break other list ? Not needed ? Right ?

Seems OK for me because i don't have it in awesome and base and seems OK.

PR #554, opened 2016-10-07
Dev: undo negative margin on list inline, because it mis-aligns gende…

tammoterhark:
…r buttons with other questions.

### DIFF
--- a/templates/default/css/flat_and_modern.css
+++ b/templates/default/css/flat_and_modern.css
@@ -1479,7 +1479,7 @@ li {
 .list-inline {
   padding-left: 0;
   list-style: none;
-  margin-left: -5px;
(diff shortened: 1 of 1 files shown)


Shnoulle:
No : i lost this one : form-inline is OK : i moved all other in https://github.com/LimeSurvey/LimeSurvey/commit/ff98c94296825fe65f0cc17df3bd3ed04b84e86f

And remove the css fix in https://github.com/LimeSurvey/LimeSurvey/commit/36f0dd35dae40fab595d3d631dbb1f0b5d48632b

Shnoulle:
See : http://testing.sondages.pro/answershtml/alternate/awesome/index.php/97738/lang/fr/newtest/Y

PR #556, opened 2016-10-11
Answers html + master to develop

Shnoulle:
All is near done. Think for 3.0 need:
- [x] Review HTML for save/reload and ? https://bugs.limesurvey.org/view.php?id=11771
- [ ] Fix index view : https://bugs.limesurvey.org/view.php?id=11754 : not clean actually . But if template use it and use included class : we broke it too
- [x] #11755: Tool link : no default action 
- [ ] Improve default template
- [x] Improve and rename base_awesome (basic ? )
- [ ] Review 5 start emoji + star : https://bugs.limesurvey.org/view.php?id=11752 yes, we broke existing question using it, but for a better view ;)
- [ ] Add news_paper
- [ ] Add ubuntu_orange
- [ ] Hard testing by our hardcore user
- [ ] **Move it to dev alpha**
- [ ] Add a 'no col' system : https://bugs.limesurvey.org/view.php?id=11743
- [ ] Default column width and table header width : https://bugs.limesurvey.org/view.php?id=11726 if we don't do it now, we break (a very little) template API (the template API is HTML produced)
- [ ] Add an input_size attribute : https://bugs.limesurvey.org/view.php?id=11734 : allow form-inline OR ls-input-size{width:auto} system
- [ ] Hard testing by our hardcore user n°2
- [ ] Send alpha public / public beta

Question about:

Think the other are fix or improvement and then don't break template API : [answers_html tag in mantis](http://tinyurl.com/z4fjkt2)

PS : lesser hard to see commits history : https://github.com/Shnoulle/LimeSurvey-Shnoulle/commits/answers_html-mergedMaster


Shnoulle:
For information : this pull request is not needed now. Need to be merged to update some other part. Then : merged this morning.

Leave it for task list

PR #557, opened 2016-10-11
Fixed #11772: Expression manager convert_value does not obey strict setting

mennodekker:
Please apply patch to 2.06lts branch too when acception.

See https://bugs.limesurvey.org/view.php?id=11772 for more info and test survey


mennodekker:
Issue resolved, closing pull request

olleharstedt:
Thanks for reporting.

PR #558, opened 2016-10-11
Dev: Added coreClass in order to have same classes in all array questions.

tammoterhark:
Also deleted question class (2x), this seems superfluous.

### DIFF
--- a/application/views/survey/questions/arrays/dualscale/answer_dropdown.php
+++ b/application/views/survey/questions/arrays/dualscale/answer_dropdown.php
@@ -11,7 +11,7 @@
 
 ?>
 
-<table class="table question table-bordered table-dual-scale question subquestion-list questions-list dropdown-list">
+<table class="<?php echo $coreClass; ?> table table-bordered table-hover table-dual-scale subquestion-list questions-list dropdown-list">
 
 <!-- Column groups -->
 <colgroup>


Shnoulle:
Thanks : i lost this one :) you done my task :heart_eyes:

tammoterhark:
Array dual scale with drowpdown:

Is there any reason for: 
ls-answers subquestion-list questions-list array-dropdown-list table table-bordered table-hover table-dual-scale subquestion-list questions-list dropdown-list

and then I mean: subquestion-list (2x) and question-list (2x) dropdown-list (next to array-drowpdown-list ?

These classes are not used in other views.

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/commit/8dc488d2856c6f8c87c51231241c463f0d529dbd

For  subquestion-list questions-list ; used in other view : to allow template manager to set specific js or css to all of this list(array)

tammoterhark:
OK, but they where added by coreClass also. 2 x same class is not necessary. Or does it have some function?

Shnoulle:
Yes, but already removed here : https://github.com/LimeSurvey/LimeSurvey/commit/8dc488d2856c6f8c87c51231241c463f0d529dbd#diff-01d2ad2a779fecfe153432d78501b182R14 :)

Shnoulle:
In fact : finally : i wantto move all "no boostrap class" to qanda (or questionobject) : view just add "decorative" class

tammoterhark:
OK, I will stay away from question views now, since you seem to have covered it all. I also think that default template look as it should apart from some things that I cannot solve. I added them to Mantis, asking  higher gods to look into this.

tammoterhark:
Your last remark makes extra sense when user can override views easier than now. All functional classes are added by default. All styling classes are in the user editable views. That would be great, also for updating...

Shnoulle:
Yes : for overriding : i think template manager don't have to add "answer-item" etc part. Just have to add %coreClass% or something like that, to be sure to have the needed class for the API.

If he don't add it : he break API : we not fix his issue ;)

For default : if you already remove uneeded part and class and fix some part : it's already a good start.

Don't do save/load actually : i must fix HTML before.

PR #559, opened 2016-10-11
Dev: Also deleted classes that are injected via coreClass

tammoterhark:


Shnoulle:
I already fix double class : but not the ranking div .....

For table-bordered : if admin user add 

````
<table class="table-bordered">....</table>
````

To a question : he surely want it in question text. Then : please only apply to ls-answers part (or remove it if you prefer :) )

tammoterhark:
I understand your hack, but I find it not elegant.

Shnoulle:
I don't think it's a hack.

1. core HTML add table-bordered : because it's more clean to see the border by default (understand where is your mouse)
2. A template designer can prefer to hide this border : he can do it in template.css
3. The best way is to don't remove border, but hide it with transparent : border still here but not shown.

PS : in fact : i think "core list" must be bordered too (you know SkeletonQuest ;) ) : better see "where is my mouse : i i click here : it check THIS radio (and not another one).

Maybe the best is to leave border in default ?

Shnoulle:
extra div done in https://github.com/LimeSurvey/LimeSurvey/commit/d38306bf57ff73174e32d02f974da16ffa330b09 thanks :)

LouisGac:
Hack is a noble concept, it's the base of the OpenSource mentality.
https://en.wikipedia.org/wiki/Hacker_culture

You should not use it with this bad meaning, typical of Microsoft coders. 
https://en.wikipedia.org/wiki/Kludge#Computer_science

"_kludge_" is the word you choose use to speak about quick and dirty solutions.

Shnoulle:
I like to be a hacker, not an issue, but here i think tammo think like a "bidouillage" in french. Here : i think it's the best way :)

tammoterhark:
I just thought it was not elegant not to use table-bordered and to switch it off via ls-answers... 
http://en.bab.la/dictionary/french-english/bidouillage
I assume that you are right...

PR #560, opened 2016-10-12
Add afterQuestionTypesBuilt event

yuriybabenko:
Event allows plugins to alter the possible question types.

Use-case involves removing question types deemed not necessary.

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -644,6 +644,12 @@ class Question extends LSActiveRecord
          */
         asort($questionTypes);
 
+        // allow plugins to alter the Question Type list
+        $event = new PluginEvent('afterQuestionTypesBuilt');
+        $event->set('questionTypes', $questionTypes);
+        App()->getPluginManager()->dispatchEvent($event);
+        $questionTypes = $event->get('questionTypes');
+
         return $questionTypes;
     }
 


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/tree/questionobject_prototype1

yuriybabenko:
@Shnoulle - Can you explain what you mean by that link?

olleharstedt:
Why would you want to remove question types? Not sure I understand your use-case.

We are working with question types etc, there's a lot of information about it here: https://manual.limesurvey.org/Question_object_types

yuriybabenko:
@olleharstedt - Our use-case is rather proprietary. Certain types need to be removed as they're not needed and will cause problems with the downstream processing (outside of LS) of survey data.

This same event can be used to add/rename, or otherwise alter existing question types, which would be a more common use-case.

olleharstedt:
Hm, OK. Why don't you fork the repo and remove question types instead?

Shnoulle:
@yuriybabenko : in fact : Olle already start to rework completely Question Object system. Then when we are ready : you can deactivate some Question Object directly in the GUI (maybe) and add yourself some.

Then : i don't think we need adding an event on this purpose actually.

PR #564, opened 2016-10-18
Added wrapping div with class $coreClass for better behaviour

tammoterhark:
Denis: when adding coreClass (ls-answers) to a table, the behaviour is different than when it is added to a wrapping div. I think it is easier to always add the coreClass to the same element, so to a div. Agree? Then I will do all arrays.

### DIFF
--- a/application/views/survey/questions/arrays/texts/answer.php
+++ b/application/views/survey/questions/arrays/texts/answer.php
@@ -25,65 +25,66 @@
 <!-- Multi Text -->
 
 <!-- answer -->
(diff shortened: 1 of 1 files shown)


Shnoulle:
Oh no .... i don't like it : why adding a element not needed. The answer element is a table : must stille at table.

For me : it's more to have a better separation between another html part .

An,d more we have already a wrapper in question.pstpl

Shnoulle:
The real reason is : we already have question.pstpl for this. Then no need another wrapper. Add it only if needed (.row for exemple)

PR #566, opened 2016-10-24
Update AuthLDAP.php

TimShaw61:
The existing code assumes that the user will enter a mail attribute in lower case.  This will often be the 'mail' attribute which causes no problems.  Where the desired email attribute is in U/L case however, as in for example; alternateEMail, the user creation will fail because the ldap_get_entries array only contains lower case values.


Shnoulle:
Need to be in option and default to NO. User lowercase/uppercase login is important. And some LDAP login can have UpperCase

TimShaw61:
I don't understand why this was closed.  The code is still broken for anyone trying to use an attribute whose name is not entirely in lower-case.  The comment by Shnoulle above makes no sense as it does not appear relate to this issue

Shnoulle:
It's authentifiction and when authenticate Shnoulle@example.org !== shnoulle@example.org … except if admin say it…

TimShaw61:
It's **NOT** about authentication!!!

It's about the assumption that the email attribute name will be in lower case and when it isn't - eg. an emailattribute called 'alternateEMail' the creation of the user will fail because the attribute name will not be recognised unless it is changed to lower case.  This is correctly done in the code for the fullName attribute on the following line and should also be done for the email attribute.  Whoever coded this simply assumed that the email attribute would be 'mail' as in many LDAP implementations and didn't bother with the necessary strtolower()...

Hence my merge request

LouisGac:
1st: sorry for closing it without saying why. I personally can't judge about LDAP PR because I don't use it. I agreed with Denis comment: not everybody wants case insensitive login, and we can't push such a change to Master, affecting dozens of thousands running installations. It could be a security issue. 

I reopen the PR so we can still discuss it.

Shnoulle:
Right : it's not about Auth , sorry … . All ldap return lower string for key ? Even if attribute have lower and upper case ? I only find this : <https://php.net/manual/en/function.ldap-search.php#28991> about this issue … 
serach need lowerAndUpper but key returned are lowerandupper … MS directory or openLdap ?

TimShaw61:
This is well known to users of the PHP LDAP functions.  See [http://php.net/manual/en/function.ldap-get-entries.php](http://php.net/manual/en/function.ldap-get-entries.php) for confirmation of this behaviour:

_"...The structure of the array is as follows. The attribute index is converted to lowercase. (Attributes are case-insensitive for directory servers, but not when used as array indices.)..."_ 
 
The original code writer has understood this when using strtolower in the following line in order to correctly change the fullName attribute to lower case.  The assumption was made - because the writer had little experience of other possible LDAP attribute names - that the emailattribute would be in lower case ('mail') and thus needed no further munging.

I have run this single line alteration for nearly 12 months now and I'm getting tired of having to fix it every time I do an upgrade...

LouisGac:
well why did you waited such a long time to contest the closure of your PR? ;-) 
as you can see, as soon as you contested it I reopened it.

Shnoulle:
/me understand why i don't like to work with ldap and php ;)

TimShaw61:
I'm busy and this is only a small part of my job.  Today I had some time.  I'm grateful for your assistance and I'm not criticising anyone - just hoping to improve the code!

LouisGac:
I did the merge on GitHub, I'm not 100% sure it's correct. I'll have a closer look tomorrow

c-schmitz:
Thanks. This pull request has been integrated manually.

PR #567, opened 2016-10-25
Added new default as a testing theme

tammoterhark:
May be adjusted. May be used as a starting point for new templates. Would love to see the inheritance working (Louis?)


LouisGac:
yes, I'd love it too. :-D

Shnoulle:
To better keep track i do it in 3 commit : 

1. rename default https://github.com/LimeSurvey/LimeSurvey/commit/962e82b6d61700fe7b69829b043e1c71c9e52372
2. add your's as default https://github.com/LimeSurvey/LimeSurvey/commit/fcd3d9641a32d94e62ccdbd501bad488c52a2f4f
3. fix your for package https://github.com/LimeSurvey/LimeSurvey/commit/3bd93da9f1b8e9410efd3c063c6fde961da4b92e
4. remove a bunch of JS : https://github.com/LimeSurvey/LimeSurvey/commit/ca7b2ba31e5859e7e55fdc7c59533424657bdded

One question : this morninbg , working on index , i see the div id="group-{GID}" class='row' .

I want to remove the row class, i think this part must be only used with id.
It's easy to add it in startgroup/endgroup.pstpl : but you really need it ? It's better to have col-xs-12 on each part ?

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/commit/13ca4bbc02e98084f124db5dfef27e2618b5c24c

PR #569, opened 2016-10-25
Dev added margin for survey-expiry-wrapper

tammoterhark:
Denis, please look, because I am unsure if this is the right approach. Doesn't javascript also play a role, because the menu bar (fixed) has a variable height?


Shnoulle:
In http://getbootstrap.com/examples/navbar-fixed-top/css : 
````
body {
  min-height: 2000px;
  padding-top: 70px;
}
````

But you'r right : this padding need to be updated in js after.

Shnoulle:
I remove some other class tomorrow :) dontread + th-X
- dontread : replace by aria-hidden in the line
- th-X : undocumented, don't understand , must be reviewed

tammoterhark:
Thanks!

PR #570, opened 2016-10-26
Several dev commits, small things

tammoterhark:
@shnoulle: Looks like we are nearly there with this template.

Still I find the alignment of input/label combination in radio buttons and checkboxes (especially when the label is multiple lines) not nice, but don't know how to solve it.

What do you think: shall I put back the original darker blue, so users get more or less the same feeling?

I think that the color combination now is ultimately very dull, but a good basis for further work. I do not want to make the default template too flashy...

### DIFF
--- a/templates/default/css/ls-functional.css
+++ b/templates/default/css/ls-functional.css
@@ -19,12 +19,6 @@
   .ls-label-xs-visibility {
     display: normal; } }
 
-.question-help-container .questionhelp {
-  padding-bottom: 5px; }
-
-.ls-em-tip {
-  padding-top: 5px; }
-
 div.question-help:empty {
   display: none; }
 
--- a/templates/default/css/ls-functional.css
+++ b/templates/default/css/ls-functional.css
@@ -15,6 +15,9 @@
   margin-top: 15px;
   margin-bottom: 0; }
 
+.ls-label-xs-visibility {
+  display: none; }
+
 @media (max-width: 768px) {
   .ls-label-xs-visibility {
     display: normal; } }
@@ -37,14 +40,8 @@ ul.ls-answers {
 .ls-js-hidden-sr {
   display: none; }
 
-.ls-choice {
-  background: white; }
-
 .checkbox, .radio {
   display: table-cell; }
 
-.survey-expiry-wrapper {
-  margin-top: 75px; }
-
-.ls-label-xs-visibility {
-  display: none; }
+.slider.slider-horizontal {
+  width: 100%; }
--- a/templates/default/css/ls-functional.css
+++ b/templates/default/css/ls-functional.css
@@ -43,12 +43,6 @@ ul.ls-answers {
 .checkbox, .radio {
   display: table-cell; }
 
-.text-left {
-  float: left; }
-
-.text-right {
-  float: right; }
-
 .survey-expiry-wrapper {
   margin-top: 75px; }
 
--- a/templates/default/css/styling.css
+++ b/templates/default/css/styling.css
@@ -96,3 +96,12 @@ label.radio-label, label.checkbox-label {
 .multiple-short-txt .form-group,
 .numeric-multi .form-group {
   margin-bottom: 10px; }
+
+.well {
+  background-color: #fff;
+  border: 1px solid #999;
+  border-radius: 0px;
+  box-shadow: none;
+  margin-bottom: 20px;
+  min-height: 20px;
+  padding: 15px; }
--- a/templates/default/css/styling.css
+++ b/templates/default/css/styling.css
@@ -114,3 +114,6 @@ label.radio-label, label.checkbox-label {
 
 .survey-expiry-wrapper {
   margin-top: 75px; }
+
+.btn-default {
+  border-color: #999; }
--- a/templates/default/css/styling.css
+++ b/templates/default/css/styling.css
@@ -5,6 +5,13 @@ div[id^="question"] {
   background-color: #ededed;
   margin-bottom: 1em; }
 
+.question-help-container .questionhelp {
+  padding-bottom: 5px;
+  padding-top: 5px; }
+
+.ls-em-tip {
+  padding-top: 5px; }
+
 .group-description-container {
   margin-bottom: 15px; }
 
@@ -16,7 +23,9 @@ div[id^="question"] {
 .question-answer-container {
   border-left: 1px solid #999;
   border-right: 1px solid #999;
-  border-bottom: 1px solid #999;
+  border-bottom: 1px solid #999; }
+
+.answer-container {
   padding-top: 10px;
   padding-bottom: 10px; }
 
@@ -69,3 +78,21 @@ td.radio-item, td.checkbox-item {
 
 tr.ls-heading th {
   text-align: center; }
+
+.answer-container .form-group {
+  margin-bottom: 0; }
+
+label.radio-label, label.checkbox-label {
+  display: inline; }
+
+.choice-5-pt-radio .answer-item {
+  margin-right: 10px; }
+
+@media (max-width: 768px) {
+  .choice-5-pt-radio .answer-item {
+    display: block; } }
+
+.multiple-opt-comments .form-group,
+.multiple-short-txt .form-group,
+.numeric-multi .form-group {
+  margin-bottom: 10px; }
--- a/templates/default/css/styling.css
+++ b/templates/default/css/styling.css
@@ -105,3 +105,12 @@ label.radio-label, label.checkbox-label {
   margin-bottom: 20px;
   min-height: 20px;
   padding: 15px; }
+
+.multiple-opt .ls-answers li {
+  margin-bottom: 10px; }
+
+.ls-choice {
+  background: white; }
+
+.survey-expiry-wrapper {
+  margin-top: 75px; }
--- a/templates/default/views/navigator.pstpl
+++ b/templates/default/views/navigator.pstpl
@@ -1,14 +1,14 @@
 <!-- Navigator hidden-xs-->
 
-<div class="row" id="navigator-container">
-    <div class="save-all hidden">
+<div class="rows" id="navigator-container">
+    <div class="row save-all hidden">
         {SAVE}
         {CLEARALL}
     </div>
 
-    <div class="save-all clearfix">
-      <div class="text-left">{MOVEPREVBUTTON}</div>
-      <div class="text-right">{MOVENEXTBUTTON}</div>
+    <div class="row save-all clearfix">
+      <div class="col-sm-6 text-left">{MOVEPREVBUTTON}</div>
+      <div class="col-sm-6 text-right">{MOVENEXTBUTTON}</div>
     </div>
 </div>
 


tpartner:
@tammoterhark, the input alignment is messed up because bootstrap is forcing the display of the label to be inline-block. If you want to dummy-down the template, I would remove that.

Something like:

`.list-radio .control-label,
.multiple-opt .control-label {
	display: inline;
}`

Shnoulle:
Tony have right, i only do big test with awesome, and old default have awesome. I do some fix in test/template-core.css now.
@tpartner : i think this part must be 'template' specific : i always think label must be block and input absolute (SkeletonQuest based), boostrap choose inline block and awesome inline-block too. Default from brower is inline.
I didn't think we must fix it limesurvey, but only in template : template specific. OK with this ?

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/commit/ac9a9877651e90801ce1263b956d47a1c1a0a42b#diff-ef769dae026d1f69fdee4cb5dfe5ba77R28 .

tammoterhark:
I agree, fix this in templates. What shall I do for default? I think that leaving it as it is, is not a show stopper, but it looks ugly... Is being ugly an argument to change things (I'm ugly...)?

Shnoulle:
Maybe best is reset flatandmodern to default seems better

tammoterhark:
You must be kidding!

tpartner:
Regarding the positioning, I also agree that it should be template-specific. 

If you want the template to be truly "basic", display both elements inline as the browser would. Otherwise, sure, an approach like SkeletonQuest works.

Shnoulle:
@tparner : yes ... basic is a bad naming ....

Propose : 

- test => basic : no awesome + no table to block in little screen + nothing ....
- basic : clear ? awesome checkbox + table to block + some little improvment that can be interesting

@tammoterhark sorry, but actually default is ... ugly ;) .

tammoterhark:
I agree that it is ugly. It was ugly before, but there were many things in it that I felt were giving a bad user experience: too wide margins and paddings, little contrast.

I just re-introduced the ugly colors and the original Lato font, in order to make people more comfortable. It will be in the next PR. Please evaluate.

I do my best, but, as said, I am not a coder. I am a project manager with a background in ergonomics...

tpartner:
I'll jump back in when you guys determine which template(s) we're focusing on. :)

tammoterhark:
You two, as gurus: please guide me. Where do I re-introduce the older table behaviour that made tables break down at smaller screens?

tammoterhark:
Denis: please decide which templates we are going to build for the new version that will be based on answers_html. I thought it was default, so I put several hours in it.

I don't mind shifting to another, but I thought we had agreed on this one.

tpartner:
@tammoterhark, as an example, for the flatmodern template, most of the "no tables" stuff seems to be in a media query starting at line 81 of template-core.css.

tammoterhark:
In the default as it is now in answers_html the awecheckboxes do not work, because the HTML does not have the triggering class="checkbox", as it was in the currently distributed default (2.55 -).

Denis: is that by design? If so, how do we trigger the awesome checkbox?

tammoterhark:
Thanks Tony, I will investigate.

c-schmitz:
We will not name the default template 'default' anymore. It is misleading and leads to misunderstanding. Whatever we choose as default template it should have a normal catchy name.

tammoterhark:
Denis: in the previous HTML for arrays classes like "no-more-tables no-more-tables-array-no-dropdown" were generated, which made tables break down in small screens. These classes are gone now. What did you have in mind for generating the behaviour as before?

tammoterhark:
@c-schmitz : I agree that default is a bad name, but should't there be a continuation of the name for backward compatibility? I would not mind changing it though....

Shnoulle:
About "In the default as it is now in answers_html the awecheckboxes" https://github.com/LimeSurvey/LimeSurvey/commit/2d65afb60d43a71475d598a2ac5cbc3bf205dc90 (issue was on my repo, not official)
About " in the previous HTML for arrays classes like "no-more-tables" : not needed table.ls-answers is OK https://github.com/LimeSurvey/LimeSurvey/blob/answers_html/templates/flatmodern/css/template-core.css#L84

no-more-tables muist be applyed (by default) to whole ls-answers table.

tpartner:
Yes, default should be an assignment, not a name. And, given that the templates are completely new, I don't think we need to worry about naming continuity.

I suggest simple names with an indication of the template feel:
- basic-blue
- basic-green
...
- flat-modern-blue
- flat-modern-green
...

tammoterhark:
@shnoulle: that means we have to edit the template-core.css? Or am I wrong?

tammoterhark:
Agree with Tony: I vote for basic-blue, basic-green, since flat and moders refers to bootswatch.

Shnoulle:
@c-schmitz : doof idea , but .... must have a 'fixed' default somewhere https://github.com/LimeSurvey/LimeSurvey/blob/answers_html/application/models/Template.php#L93 or a "updatable by php file only" default, and must be in standard template list : https://github.com/LimeSurvey/LimeSurvey/blob/answers_html/application/models/Template.php#L329

@tammoterhark : template distribution is always done via FTP/upload. Then css file not 'editable' by admin GUI can be updated.
template-core is part of template

c-schmitz:
@tammoterhark Backward compatibility is not an issue here - it is not depending on the name, but on the default being set in the code.

tammoterhark:
OK, renaming would also show users that things have changed. I think that is a good idea.

tammoterhark:
OK, I will look in older files for the ancient "no-more-tables-behaviour".

tpartner:
@tammoterhark, no, all you need to do is add something like the media query I referenced above to styling.css. It, among other things, applies the "no-table" styles to table.ls-answers elements.

And, while on the topic of styling.css, why isn't template.css included in the XML file? (and not included in the DOM)

tammoterhark:
I wanted to start "clean" so I left template.css out. This sets me back, but also it enforces me to think what is really needed.

tpartner:
Okay, but in the end, for consistency, I think it needs to be included - folks are used to making there changes in template.css and template.js.

tammoterhark:
@tpartner : I find your remark a little confusing, after discussing that changing the nem of the template is no issue...

Templates now have a different structure:
bootstrap.min.css (do not touch ...)
other-helpers.css (do not touch)
ls-functional.css (you may touch this, but be careful)
styling.css (change whatever you like)

In stead of the previous template.css which had (more or less) all of the above.

tpartner:
Changing the template names has nothing to do with the work flow for the user. As I said, for years they have made style changes in a file called template.css. Why would we change that? 

They will also find tons of references to that filename in the docs and the forum so may be mislead.

tammoterhark:
OK, I agree that references to that in de documentation (can be changed, but is a pain...) and the forum (cannot be changed) is a good argument.

Shall I rename styling.css to template.css?

tammoterhark:
I renamed styling.css to template.css, but I will leave the work for now. I think we are not working coherently towards a common goal. 

Denis, please, what is the common goal for developing a template: is it the default as what I was working on, or is the common goal the template that you are working on? I want to work together, not apart. Please tell me what to work on.

Shnoulle:
I already said : i didn't want to touch to default template. BUT : i think it's better if default template is not complex and heavily css/js hacked + must have modern look and color.
It's what i think , but it's MY opinion.

After : i surely distribute different template.

PS : color.css/template-core.css/anything.css is only a template manager system : this is not fixed, anyone can use own name. Tony have rights about template.css : i think too is must be the "preferable way" to update template, but i think i add a color.css in MY distributed template.

@tpartner : about templatename-gree templatename-blue etc ... i think more of something like templatename-cerulean / [templatename-superhero](http://bootswatch.com/superhero/) : just replacing the boostrap.css must/can do the trick. It's a readon why i didn't force color in [multinum-slider](https://github.com/LimeSurvey/LimeSurvey/blob/answers_html/scripts/numeric-slider.js#L39) but try to use a boostrap color.

tammoterhark:
Hi Denis,

it is OK that you do not want to touch the default template. I now recreated it, using the system as stated above:

bootstrap.min.css (do not touch ...)
other-helpers.css (do not touch)
ls-functional.css (you may touch this, but be careful)
template.css (change whatever you like)

I put all your code of template-core in ls-functional. But how to proceed now? I feel like I am finished now re-creating the default template, just as (ugly) as it was.... (please test).

How can we maintain the two templates, can we somehow share the ls-functional.css (in any name), so we do not have to focus both on the same thing?

And what do you wish me to do now? I would love to have that inheritance in place, since then we can really work on one template as the mother of new templates.

tammoterhark:
I recreated it because I think that current users that use the default template, want something back that looks and behave (more or less) the same.

I agree that we can do much better and more modern.

tammoterhark:
I am not a designer, maybe we can have a good designer do a good "give away" template for introduction of the new LimeSurvey version?

Shnoulle:
Actaully  (3.0) => fix html to be more easily updatable and fixable + fix by core.

An example : before 'action link' javascript was in template.js : it must not be a template related system.

I think (like yourself, i think) : default template must be improved BUT more : must be cleaned .

Inheritance etc ... it's not the purpose of 3.0 (for me), take another time and it's [another](https://en.wikipedia.org/wiki/Agile_software_development) development project (agile dev).  
Old templates are made by one user where fix something must be applied to the 3 included template. We can not fix whole in package/html but i want to be allowed to fix something without breaking user template.

Some part can/must be template choice (see label discussion, or alert() discussion for example). But some other must be core related (index for example).

tammoterhark:
I think we have achieved some of the things you stated:

>  fix html to be more easily updatable and fixable + fix by core.
done

> An example : before 'action link' javascript was in template.js : it must not be a template related system.

Don't know

> default template must be improved BUT more : must be cleaned .

Default template is cleaned and improved, but still same colors/fonts, in order not to interrupt many users.

> Inheritance etc ... it's not the purpose of 3.0 (for me), take another time and it's another development project (agile dev).

I agree, I just would like this to be in place, so we can more easily share work.

> Old templates are made by one user where fix something must be applied to the 3 included template. We can not fix whole in package/html but i want to be allowed to fix something without breaking user template.

That is why I want this inheritance so much... Even multi-layer... Like in Drupal. Saves me lightyears of work.

> Some part can/must be template choice (see label discussion, or alert() discussion for example). But some other must be core related (index for example).

Be careful, because even the index is something PeterEmil tried to change. He ran into many difficulties.

Shnoulle:
action link : fixed in package
index : fixed in this night commit

tammoterhark:
Denis, can we Skype?

Shnoulle:
No, don't have time : have to work

tammoterhark:
OK, later...

tammoterhark:
Just ping me when you want me to do something. I am on hold now.

Have a great day! I think we are doing a good job.

maziminke:
Just to contribute my 2 cents:

- Template name: I don't really care, a different name should do no harm. Maybe we could add something like "shipped" as a kind of prefix to indicate which templates come with Limesurvey directly?
- CSS file name: Please keep the "template.css" file name for the file which users can/should edit.
- Template details: I second the suggestion by @tpartner to have less margins and paddings at the shipped templates. That is something I heard from lots of users and customers. They often asked how to create a more compact version/layout.

Thanks a lot for the great work so far!

tammoterhark:
For now I thinkl keeping the default just as it is, would be the least interruption to users and their surveys.

I would very much like to work on (after we have tested the default) a new, more mobile first template, with much smaller margins/paddings, because on a mobile screen, with the keyboard on 9for text questions) we can do much better...

Shnoulle:
@maziminke : 

1. name : i don't care
2. 
  - template.css , yes we never talk to remove it like template.js. 
  - And better i put some information at start of boostrap.css ('and renaming it clearly) to explain how to make a new 'global theme'. Leave this file without any update (except the font link update) see https://github.com/LimeSurvey/LimeSurvey/commit/e2e98dcb4ebdbdd3d4b033e46c570457cd0d8ca0#diff-abfac071c9f82273ee02498f93265e23 
  - Use boostrap-rtl : https://github.com/LimeSurvey/LimeSurvey/blob/answers_html/templates/default/css/bootstrap-rtl.css : need reviewing. except this one + awesome-bootstrap-checkbox-rtl.css : rtl use same css file
3. .... already done since some week .... see screenshot
capture du 2016-11-07 16-13-20

Shnoulle:
PS ! maybe time to test .....

tammoterhark:
Denis< would it be an idea to make margins/paddings smaller:
@media only screen and (max-width: 760px)

BTW: why difference between 760 and 768 in template.css? Just curious (want to learn..)

tammoterhark:
@shnoulle: If you want that, I can make a first proposal.

Shnoulle:
I saw issue for table. But : tou want less margin/padding in question part ? Look at screenshot. Point Of View

Shnoulle:
760 come from no-more-tables https://css-tricks.com/responsive-data-tables/ .

768 is bootstrap choice , see old comment of Louis in 2.50 default template.

I think i use 760px in my template. But surely use 80em or something similar and not px.

tammoterhark:
@shnoulle: I would like to invest some time in making all margins/padding for mobile a little smaller. There is a lot of space that is not needed now. Also: on collapsed tables not all border are needed: border also take up space, because they need space next to them. Please see image:

mobile

Shnoulle:
Ah : you mean margin-top/bottom : go go

tammoterhark:
Maybe not start with mobile first in this template, but it would be great to start (later, I know...) a development of a really mobile first template. I need it.

PR #571, opened 2016-10-27
Answers html index

Shnoulle:


Shnoulle:
All done : code must be reviewed, but don't update (a lot ) html.

Todo : 
1. control if we can use LimeExpressionManager::GetStepIndexInfo() in group system too
2. if yes : move to a psudo singmleton : only one call is needed to construc whole index for a page
3. Some code improvement

PS : allow multiple index in (link or button) : this can be realy great
PS2: remove the hard included index : must update manual about this ...

PR #572, opened 2016-10-27
answers_html official develop

Shnoulle:
Actually :
1. default template <s>must</s>can be reviewed (boostrap file use 'default flatly file))
2. rtl not fully tested but done


Shnoulle:
OK to merge it tomorrow ? More clear if it's develop branch i think.

tammoterhark:
@Shnoulle: you also change at least one file in the default template (template-core.css). This conflicts with the work that I am doing. 

Since I think that you do more advanced things than I am able to do, I now sit back and wait for instructions. I can do testing and write bug reports. Working on the same template does not work, especially since I work with sass. Or we have to both work with sass and take the sass files into GIT.

https://github.com/LimeSurvey/LimeSurvey/pull/572/commits/14fae5df17b4ad6a013ae31b206fc0b4ac5e33be

I may still not understand correctly what you guys are doing: I hope that template-core.css only addresses stuff that is in the question views and thus can be used independently of the template files in the temple (.pstpl files). If so, we could use template-core as a resource for all templates. If tags need to be overrideden, it is possiblke, but having a good template-core.css as an extension of bootstrap.min.css sounds like a good start, until w can have inheritance in place.

LouisGac:
Tammo: the task was to clean the HTML and the CSS, keeping the current 2.5 default template working. That what you should focus on. When this will be done, the CSS/HTML should be stable, and then it should be possible to create new ones.

tpartner:
I, like Tammo, am also waiting for instructions on which template should be tested/developed. It seems to be a moving target.

LouisGac:
Old 2.5 default template. It never moved.

Shnoulle:
This commit : 

Was just 'better' separation between 'template vs core' i think text-info his more a template system than a core system.

Unsure for pull-left.

For me : purpose : have a 'near' 2.5X default template with near same functionality + one another template to better see what can/must be in template and what must be in core.

@tpartner : i know you know how to update question the way you want with some js. I hope adding class/html to allow more easily update HTML in JS with 'ALL' template. In 2.00,2.0x ... : you must ask each time 'what template'. I hope we can do some workaround working in any template (excpet for some color for example).

PR #573, opened 2016-10-27
Dev: reintroduced old default colors and fonts.

tammoterhark:
I really think that the margins and paddings in the old (currently distributed) default are too wide, especially on mobile screens. What would be best: make them wide on wide screens and small on mobiles, or leave it small (like now) or re-introduce the older, wider margins/paddings? Please discuss.

Making everything as much as the currenntly distributed version, will making adaptation better, but I feel bad distributing a template that goes against many usability rules...


Shnoulle:
At tammo : i merge this one or not ?

tammoterhark:
Yes, please merge. And let's phone first!

tammoterhark:
Please also review, you are a better developer than I am, so I need feedback...

Shnoulle:
I take a look after : maybe some part can be removed now :)

Shnoulle:
When i put phone first : i think of @ media system : no media : do a template for phone, use only media with min-width.

We can not really do like this for table.ls-answers : table is table .....

tammoterhark:
The more that can be removed, the better. As little css as possible...

And I hate arrays on mobile phones. We now have something like a system, but it is still very bad user experience.

PR #576, opened 2016-11-01
Extend AuthLDAP with optional group-membership restriction

fredcy:
This adds an optional restriction to LDAP authentication intended for checking that the user is a member of a particular group in LDAP. It adds two new setting parameters -- a base DN and a filter -- and if set searches those and rejects the authentication attempt if there is no match.

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -99,6 +99,16 @@ class AuthLDAP extends ls\pluginmanager\AuthPluginBase
         'automaticsurveycreation' => array(
                 'type' => 'checkbox',
                 'label' => 'Grant survey creation permission to automatically created users'
+        	),
+        'groupsearchbase' => array(
+                'type' => 'string',
+                'label' => 'Optional base DN for group restriction',
+                'help' => 'E.g., ou=Groups,dc=example,dc=com'
+                ),
+        'groupsearchfilter' => array(
+                'type' => 'string',
+                'label' => 'Optional filter for group restriction',
+                'help' => 'Required if group search base set. E.g. (&(cn=limesurvey)(memberUid=$username))'
                 )
     );
 
@@ -434,6 +444,8 @@ class AuthLDAP extends ls\pluginmanager\AuthPluginBase
         $usersearchbase		= $this->get('usersearchbase');
         $binddn     		= $this->get('binddn');
         $bindpwd     		= $this->get('bindpwd');
+        $groupsearchbase        = $this->get('groupsearchbase');
+        $groupsearchfilter      = $this->get('groupsearchfilter');
 
         // Try to connect
         $ldapconn = $this->createConnection();
@@ -492,6 +504,21 @@ class AuthLDAP extends ls\pluginmanager\AuthPluginBase
                 $this->setAuthFailure(self::ERROR_USERNAME_INVALID);
                 ldap_close($ldapconn); // all done? close connection
                 return;
+            }
+
+            // If specifed, check group membership
+            if ($groupsearchbase != '' && $groupsearchfilter != '')
+            {
+                $filter = str_replace('$username', $username, $groupsearchfilter);
+                $groupsearchres = ldap_search($ldapconn, $groupsearchbase, $filter);
+                $grouprescount = ldap_count_entries($ldapconn, $groupsearchres);
+                if ($grouprescount < 1)
+                {
+                    $this->setAuthFailure(self::ERROR_USERNAME_INVALID,
+                    gT('Valid username but not authorized by group restriction'));
+                    ldap_close($ldapconn); // all done? close connection
+                    return;
+                }
             }
 
             // binding to ldap server with the userDN and privided credentials


Shnoulle:
I like the option of restriction to [simplebind](https://github.com/fredcy/LimeSurvey/blob/17b4395839d0a53a3aceb099ff8049a5e4cc5eec/application/core/plugins/AuthLDAP/AuthLDAP.php#L461). 

But really think we can use LDAP filter for same purpose if using filter .

fredcy:
Not sure I understand your comment, @Shnoulle. But I think I disagree with both of your points.

First, in my case we have to use "search and bind" as the user entries are not in one OU but rather have to be searched a subtree rooted at an OU. So simple bind would not work for us.

Second, I can find no way to use a filter on the user search to do what we need. We do not have the "memberOf  overlay" installed in our LDAP system and our sys admins are very reluctant to add it. Since LDAP searches go entry-by-entry and do not allow for join-like expressions, I have no way to specify a filter that both searches for uid=$username and also tests that that same uid/username is a `memberUid` value in a completely separate group entry (in a different OU subtree, to boot).

Shnoulle:
for LDAP filter : you don't have to have a memberOf.

fredcy:
I have found no way to construct such a filter. Please tell me how you would do it. I would be more than happy to use a filter if that would work.

We have a dozen or so applications in house that use LDAP groups to control authorization. All use a secondary LDAP search over the group OU to do that. None get by with a filter on the user search.

fredcy:
I'm currently using a plug-in that I built yesterday from a copy of AuthLDAP.php. The plugin is basically the same as the result of this PR (except for its name). If this PR goes nowhere I'll just use my plugin. I just wanted to contribute back and offer something of potential value to other limesurvey users.

Aestu:
@fredcy : Thanks for sharing your code. However, I don't understand why a filter can not do the job. Perhaps you can explain us your problem with an example.

fredcy:
Our user accounts are within ou=People,dc=imsa,dc=edu in sub-OUs like ou=Staff. To authenticate accounts we search ou=People and its sub-OUs for uid=$username and then bind with $password against the matching DN if found.

To control authorization we use posixGroups in LDAP. In this case we have the entry cn=wwwlimesurvey,ou=Groups,dc=imsa,dc=edu which comprises memberUid attributes listing the uid values of user accounts that have authorization to use LimeSurvey. 

So we need to verify that the user trying to login has their uid as one of the memberUid attribute values in this cn=wwwlimesurvey,ou=Groups,dc=imsa,dc=edu entry. I don't see any way to do that using a filter on the user search.

fredcy:
Here's an ldif that gives a greatly simplified example of our LDAP structure. The scenario is for username "fred" to login. We need to verify that "fred" is in the "wwwlimesurvey" group.

```
dn: dc=imsa,dc=edu
objectclass: organization
dc: imsa

dn: ou=People,dc=imsa,dc=edu
ou: People
objectClass: organizationalUnit

ou=Staff,ou=People,dc=imsa,dc=edu
ou: Staff
objectClass: organizationalUnit

dn: uid=fred,ou=Staff,ou=People,dc=imsa,dc=edu
objectClass: person
uid=fred

dn: ou=Groups,dc=imsa,dc=edu
objectClass: organizationalUnit
ou: Groups

dn: cn=wwwlimesurvey,ou=Groups,dc=imsa,dc=edu
objectClass: top
cn: wwwlimesurvey
memberUid: foo
memberUid: bar
memberUid: fred
```

Shnoulle:
But why not adding it to simplebing ? I really think we need some simple bind restriction.

I'm not a LDAP specialist , maybe allowing "user params" in filter is sufficient/adequate ?

LouisGac:
a decision about this PR would be a good thing. 
If I understood well, it doens't break anything, it only provides new possibilities right?

fredcy:
It adds an optional new constraint for LDAP authentication and authorization. With the defaults it acts exactly like the base LDAP authentication scheme. So it's backward compatible. 

One drawback is that the new (optional) configuration parameters could be confusing to site administrators that have no need for them.

Shnoulle:
General issue about plugin : in my opinion.

I prefer to have 3 lights simple plugin than a big complex plugin . Here 
- AuthLDAPSimpleBind
- AuthLDAPSearchAndBind
- AuthLDAPByGroup

Clearly 2.63 simplebind and searchandbind must be separated : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/plugins/AuthLDAP/AuthLDAP.php#L365

c-schmitz:
@fredcy Would you be so kind and create a new Wiki page about the LDAP authentication plugin in our manual at https://manual.limesurvey.org ? Nobody at our team has LDAP knowledge und you would probably help lots of other users with thath.

fredcy:
Thanks.

I've just created this section in the manual: https://manual.limesurvey.org/Authentication_plugins#Authentication_with_OpenLDAP.2C_uid_attribute.2C_and_group_restriction_.5B2.62.2B.5D

I put `2.62+` as the version note for now, assuming that the commit will be included in that release.

PR #577, opened 2016-11-01
Fixed issue #10025: Quotas : Terminate survey with warning allow submit

Shnoulle:
- fix the issueDev: same message is shown twice, deactivate url the 1st time
- i like to use directly movenext and send only 'quota' input to EM : but broken (see comment)
- then update move button value to confirmquota, and redirect in SurveyRuntimeHelper

### DIFF
--- a/application/controllers/admin/quotas.php
+++ b/application/controllers/admin/quotas.php
@@ -82,6 +82,11 @@ class quotas extends Survey_Common_Action
         }
     }
 
+    /**
+     *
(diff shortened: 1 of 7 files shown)


Shnoulle:
Arg ... bad commit here : https://github.com/LimeSurvey/LimeSurvey/commit/9c261a6a4b1af8280d8c27dc7334d55b4f6d08fe commited my test ....

PR #579, opened 2016-11-03
Dev: typos

tammoterhark:


Shnoulle:
Abut thsi class : we surely need a better class name .... more fixed ...

tammoterhark:
In other cases you go from general to specific
ls-group-xx

Would that not imply that here it would have to be: column-separator-right, column-separator-left?

PR #580, opened 2016-11-05
Dev: proposal for new remaining/total box

tammoterhark:
@shnoulle: please review. This way the box with remaining/total can be positioned below the sliders, optionally right, left, centered...

Some classes may need adjustment, but I cannot fully understand your idea about what classes belong where... I can go further, but I want your opinion first.

### DIFF
--- a/application/views/survey/questions/multiplenumeric/rows/dynamic.php
+++ b/application/views/survey/questions/multiplenumeric/rows/dynamic.php
@@ -10,48 +10,56 @@
  * @var $suffix
  */
 ?>
-<?php if($sumRemainingEqn):?>
-    <li class="form-group ls-group-remaining">
-        <div class="control-label col-xs-12 col-sm-<?php echo $sLabelWidth; ?>">
-            <?php eT('Remaining: ');?>
-        </div>
-        <div class="ls-input-group col-xs-12 col-sm-<?php echo $sInputContainerWidth; ?>">
-            <?php if ($prefix != ''): ?>
+<?php if($sumRemainingEqn || $sumEqn):?>
+  <div class="col-sm-push-<?php echo $sLabelWidth; ?> col-sm-<?php echo $sInputContainerWidth; ?>">
+    <div class="ls-group-remaining-total">
+      <table>    
+        <?php if($sumRemainingEqn):?>
+          <tr class="form-group ls-group-remaining">
+            <td class="control-label">
+              <?php eT('Remaining: ');?>
+            </td>
+            
+            <td class="ls-input-group">
+              <?php if ($prefix != ''): ?>
                 <div class="ls-input-group-extra prefix-text prefix text-right">
-                    <?php echo $prefix; ?>
+                  <?php echo $prefix; ?>
                 </div>
-            <?php endif; ?>
-            <div id="remainingvalue_<?php echo $id; ?>" class="form-control-static numeric dynamic-remaining" data-number="1"><!-- alternative class : form-control : display like an input:text -->
-                {<?php echo $sumRemainingEqn;?>}
-            </div>
-            <?php if ($suffix != ''): ?>
+              <?php endif; ?>
+              <div id="remainingvalue_<?php echo $id; ?>" class="form-control-static numeric dynamic-remaining" data-number="1"><!-- alternative class : form-control : display like an input:text -->
+                      {<?php echo $sumRemainingEqn;?>}
+              </div>
+              <?php if ($suffix != ''): ?>
                 <div class="ls-input-group-extra suffix-text suffix text-left">
-                    <?php echo $suffix; ?>
+                  <?php echo $suffix; ?>
                 </div>
-            <?php endif; ?>
-        </div>
-    </li>
-<?php endif; ?>
+              <?php endif; ?>
+            </td>
+          </tr>
+        <?php endif; ?>
 
-<?php if($sumEqn):?>
-    <li class="form-group ls-group-total">
-        <div class="control-label col-xs-12 col-sm-<?php echo $sLabelWidth; ?>">
-            <?php eT('Total: ');?>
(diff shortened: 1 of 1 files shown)


tammoterhark:
You have a more systematic approach than I have, I guess. I would be inclined to delete many classes, since now the label and the value are not aligned. But you know if these classes add something else to the system...

Shnoulle:
@tammoterhark in fact : i don't understand your issue. Actual solution give you any ability to put remaining the way you want. What is your issue with actual somution ?

It's a template issue : 
````
.ls-group-remaining .ls-input-group-extra{display:none;}
.ls-group-remaining .control-label{padding-right:0;padding-top: 7px;}
.ls-group-remaining .ls-input-group{padding-left:0}
.slider-list .ls-group-remaining .form-control-static{padding-left:0;text-align:left}
.ls-group-total .ls-input-group-extra{display:none;}
.ls-group-total .control-label{padding-right:0;padding-top: 7px;}
.ls-group-total .ls-input-group{padding-left:0}
.slider-list .ls-group-total .form-control-static{padding-left:0;text-align:left}
````
capture du 2016-11-05 17-57-12

**Comment by Shnoulle** [comment]
table is not needed , table is always a bad idea.

**Comment by Shnoulle** [comment]
Another one
```
.ls-group-remaining .ls-input-group-extra{display:none;}
.ls-group-remaining .control-label{display:inline-block;width:auto;padding-right:0;padding-top: 7px;}
.ls-group-remaining .ls-input-group{padding-left:0;display:inline-block;width:auto;}
.slider-list .ls-group-remaining .form-control-static{padding-left:0;text-align:left;display:inline-block;width:auto;}
.ls-group-total .ls-input-group-extra{display:none;}
.ls-group-total .control-label{padding-right:0;padding-top: 7px;display:inline-block;width:auto;}
.ls-group-total .ls-input-group{padding-left:0;display:inline-block;width:auto;}
.slider-list .ls-group-total .form-control-static{padding-left:0;text-align:left}
```
capture du 2016-11-05 18-05-09

tammoterhark:
I thought that because of:
<div class="ls-input-group col-xs-12 col-sm-<?php echo $sInputContainerWidth; ?>">

the label will always be left of a "certain point and the value will always be to the right of that point due to:

<div class="ls-input-group col-xs-12 col-sm-<?php echo $sInputContainerWidth; ?>">

sliderremaining

tammoterhark:
I what I proposed the table (or any other construction) would be plced under the sliders, not the label left and the value right of the invisible line.

Shnoulle:
It was centered in last GIT version http://testing.sondages.pro/answershtml/index.php/33965?lang=fr

https://github.com/LimeSurvey/LimeSurvey/commit/0f3e723e8ba0784c0e283b80b8cfce44075367cf#diff-eba34bc02d2cbd778d86a42b0397731fR15

Shnoulle:
With some css : you can easy set some padding to .ls-group-remaining 

capture du 2016-11-06 02-46-44

Again : template issue : not core issue

tammoterhark:
Denis, are we looking at the same thing? As it is now, it looks like this:

https://limegit.toolsforresearch.com/index.php/survey/index/sid/243255/newtest/Y/lang/en

I can hardly believe that this is what you want. I think we cannot do this to the users of the default template.

tammoterhark:
Git pulled 10 minutes ago...

Shnoulle:
It's actual situation with default template. 
Personally : i think it's OK. 

**Important think** : other can be done with some css only in template.css .

tammoterhark:
Currently it looks like this:
https://survey.toolsforresearch.com/index.php/243255?lang=en

I think that looks a lot better.

Shnoulle:
Your template.css are outdated, not a big difference but. Set config to 1. 

Else : It's a point of view. And : we can do same than 2.50 with actual HTML, or same than 2.06 too : http://demonstration.sondages.pro/33965?newtest=Y

In fact : i think have something like http://testing.sondages.pro/answershtml/index.php/68284?lang=fr (input:deactivated) is the best. Easily done with actual template.

tammoterhark:
config = debug, I guess?

tammoterhark:
Debug = 1, emptied /tmp/assets, but still there. How can it be that I do not have the correct template.css? I pull from the LimeSurvey git, not from your GIT. Is the correct template.css at LimeSurvey?

Shnoulle:
Oups : missunderstanding i think : screenshots as made adding css at end of template.css OR update with firebug .... quick test

tammoterhark:
I go to get some food first, hungry.

jexller:
lstotalremaining

Tammo: The total remaining in the left corner will be overlooked by many respondents. Most harmonic layout not always the best for surveys where people start scanning quickly.
But the styling with the boxes is nice and both values are nearer together. So it might be nice to move your styled valueboxes in the middle of the sliders.

tammoterhark:
I was not promoting to put it in the left corner, I was promoting to keep the label and the value close together and to be able (as a template builder) to position the block with both labels and values anywhere on the screen.

Shnoulle:
Actual HTML allow with ~ 10 lines to put label+pseudo value anywhere on the screen

There are 3 questions : 
1. What default template must do 
2. What a "only boostrap" template must do
3. What HTML must be serve : See my comment here : https://bugs.limesurvey.org/view.php?id=11883#c41789 I think have the same HTML allow more easy workaround (same workaround for the 2 situation).

tammoterhark:
If you can get it done with the HTML as it is now, please do it. I thought that you would need to overrule col-sm-xx statements and I think that that is a bit hackish. But if it gets the business done, OK by me...

tammoterhark:
I would love to see how you do this: I like to learn.

tammoterhark:
Ah, I guess I see.... :before!

PR #581, opened 2016-11-05
Update surveysettings.js

BertHankes:
Strip urlparams of unwanted params. 
See: https://bugs.limesurvey.org/view.php?id=11866

This will prevent the passing of potential XSS content.
BTW: The stripped elements (act, id and title) are stripped later on in /application/controllers/admin/database.php. So now it is not posted and therefore save regarding XSS scanners.


c-schmitz:
Obsolete.

PR #583, opened 2016-11-08
Dev: Extra tags in media query for better styling on mobile

tammoterhark:
@shnoulle please reviw the latest commit. The others: I'm sorry; they do something and after that revert it. I am afraid to loose work if I delete these...

### DIFF
--- a/templates/default/css/template.css
+++ b/templates/default/css/template.css
@@ -62,10 +62,12 @@ padding-top:1em;padding-bottom:1em;margin-bottom:2em;
 .question-container {
   margin-bottom: 2em;
   background:white;
+  border-bottom: 1px solid #dbdbdb;
 }
 
 .question-title-container {
-  padding-top: 2em;padding-bottom:1em;
+  padding-top: 1em;
+  padding-bottom:1em;
 }
 .question-number,.question-code{float:left;margin:0 0.3em}
 .question-number:empty,.question-code:empty{display:none}
@@ -73,7 +75,12 @@ padding-top:1em;padding-bottom:1em;margin-bottom:2em;
 .question-valid-container {
   position: relative;
   text-align: center;
-  padding-bottom: 1em;
+  //padding-bottom: 1em;
+}
+
+.ls-question-help,
+.ls-questionhelp{
+  margin-bottom:5px;
 }
 
 .question-valid-container .text-danger{
@@ -89,12 +96,18 @@ padding-top:1em;padding-bottom:1em;margin-bottom:2em;
 .answer-container {
   border-width: 0px 1px 0px 1px;
   border-style: solid;
-  padding-top: 3em;
-  padding-bottom:1.5em;
+  //padding-top: 3em;
+  //padding-bottom:1.5em;
+}
+
+.ls-answers{
+  padding-top:15px;
+  padding-bottom:15px;
+  margin-bottom:0;
 }
 
 .question-help-container {
-  border-width: 0px 1px 1px 1px;
+  border-width: 0px 1px 0px 1px;
   border-style: solid;
 }
 /* little icon near the question-help */
@@ -125,7 +138,6 @@ padding-top:1em;padding-bottom:1em;margin-bottom:2em;
     right:0;
 }
 
-
 /**
  * Unsure part
  */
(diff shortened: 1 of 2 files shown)


templates/default/css/template.css
```diff
@@ -622,3 +624,47 @@ div.yes-no .btn-group label {
...
+    border:none;
+  }
+  
+  .ls-answers tr:last-child{
+    margin-bottom:0;
+  }
+  
+  .ls-answers > tbody > tr > td.radio-item ,
+  .ls-answers > tbody > tr > td.checkbox-item{
+    min-height:44px; /*minimal height for touching with finger */
```
Shnoulle: In this case : max-device-width: and not width for media

templates/default/css/template.css
```diff
@@ -622,3 +624,47 @@ div.yes-no .btn-group label {
...
 }
+
+@media only screen and (max-width: 760px){
+  h1, .h3{
+    font-size:18px;
+  }
+  
+  .h4{
+    font-size:16px;
+  }
```
Shnoulle: Unsure for this part :/

templates/default/css/template.css
```diff
@@ -622,3 +624,47 @@ div.yes-no .btn-group label {
...
+  
+  .ls-answers tr:last-child{
+    margin-bottom:0;
+  }
+  
+  .ls-answers > tbody > tr > td.radio-item ,
+  .ls-answers > tbody > tr > td.checkbox-item{
+    min-height:44px; /*minimal height for touching with finger */
+    padding-top:10px;
+    padding-left:28px;
```
Shnoulle: OK for padding top , but padding-left must be same than li in big screen, no ?

templates/default/css/template.css
tammoterhark: Sorry, do not understand your comment.

templates/default/css/template.css
```diff
@@ -89,12 +96,18 @@ padding-top:1em;padding-bottom:1em;margin-bottom:2em;
 .answer-container {
   border-width: 0px 1px 0px 1px;
   border-style: solid;
-  padding-top: 3em;
-  padding-bottom:1.5em;
+  //padding-top: 3em;
+  //padding-bottom:1.5em;
```
Shnoulle: Really NEED a margin here !

templates/default/css/template.css
Shnoulle: I don't like it ..... to little.

templates/default/css/template.css
Shnoulle: width : big screen with little windows max-device-width : phone or really little screen

templates/default/css/template.css
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/commit/040db0d38c48e2403f87adb0f1a16825995ea74f

templates/default/css/template.css
tammoterhark: You are right, should be the same in big screen. Shal I change that? Please accept the current changes and I will do it again. (I do not know how do retract commits...)

templates/default/css/template.css
tammoterhark: Please look, I added it in another class.

Shnoulle: before my commit : no space before/after the ls-answers (already merged)

templates/default/css/template.css
tammoterhark: Do you mean that I should change the mediaquery?

templates/default/css/template.css
tammoterhark: Will you do that or shall I do it?

Shnoulle: Like you want :) i think i try to fix the bug for array by dual scale (https://bugs.limesurvey.org/view.php?id=11863) before

templates/default/css/template.css
tammoterhark: You put in margin, while it should have been padding. Look at box around answers..

Shnoulle: Or, margin on ls-answers ? But littlest than before.

PR #585, opened 2016-11-09
Dev: edit on answers-container

tammoterhark:
@Shnoulle : due to earlier development this was erroneously stated as margin, while it should have been padding.


Shnoulle:
@tammoterhark : trye tu set develop for merging ;)

PR #587, opened 2016-11-15
Dev: some smaller margins for collapsed arrays on mobile.

tammoterhark:


tammoterhark:
@shnoulle: why is the difference between: 

@media only screen and (max-device-width: 760px){ 

and

@media only screen and (max-width: 760px){ 

I prefer the last one: easier for testing on local, but I do not see the reason to add the "device". Please tell me.

templates/default/css/template.css
```diff
@@ -654,15 +655,24 @@ div.yes-no .btn-group label {
   .table-bordered{
     border:none;
   }
-
-  .ls-answers tr:last-child{
+  
+  .ls-answers tbody:last-child tr:last-child{
```
Shnoulle: Nee last child ? Why not all ?

Shnoulle:
max-device-width : only for 'REAL screen can not be up to that'
max-width : real screen can be , but window of brownser is small

You'r right : we need max-width even after

templates/default/css/template.css
tammoterhark: There are multiple tbody statements in the the question (due to repeated headers?)

Shnoulle: Yes: repeat header. But all need margin-bottom:0; no ? (and the last one can be great with  border-radius ;) )

tpartner:
Also note that in some iPhones, max-device-width gives a wrong result when in landscape mode - it actually gives the portrait width. Best to stick to device-width.

tpartner:
Oops - best to stick to max-width.

Shnoulle:
Thanks @tpartner :) . 

I think it's better to use padding (like boostrap seems to do with btn) than min-height: what is your opinion here ?

tpartner:
For responsive height? Yes, I use padding-top because you can use a percentage (of the parent) value:
height: 0;
min-height: 0;
padding-top: x%;

Shnoulle:
? height: 0; no, i just mean : for a minimal height for 'pseudo-button' : something where user must put his finger to select :).

Missunderstanding ;)

tpartner:
Oh, yeah, there padding is probably better than min-height.

PR #591, opened 2016-11-18

BertHankes:
Dev: Because of revoke of jqgrid and participantdisplay.js should be done in all the templates!


BertHankes:
Maybe this willl resubmit too!

PR #593, opened 2016-11-21
Dev: saving some extra space on small screens

tammoterhark:
Reduced margins and font-size on question help items

### DIFF
--- a/templates/default/css/template.css
+++ b/templates/default/css/template.css
@@ -674,5 +674,14 @@ div.yes-no .btn-group label {
   
   .ls-answers.table > tbody + tbody {
     border-top: none; /*overriding Bootstrap */
-}
+  }
+  
+  .help-block:last-child{
+    margin-bottom:0;
+  }
+  
+  .ls-questionhelp,
+  .help-block{
+    font-size:90%;
+  }
 }


tammoterhark:
Reason: wrestling with very limited space on mobile screens, especially when using text questions (with on screen keyboard)

Shnoulle:
I really don't like the 90% font-size

tammoterhark:
You do not like 90% or don't you like reducing the font-size at all?

Shnoulle:
Reducing font-size , except with small tag :)

tpartner:
I don't agree with the small tag - no guarantee how different browsers will treat it (unless you define a rule for it).

Shnoulle:
@tpartner boostrap is here : http://getbootstrap.com/css/#small-text

Resume : if you want a "little" font-size : put .small class on 'help', for workstation and phone.
There are no reason to show little font on phone and not in workstation. It's my Point of view.

The major issue is : why you want a little font on phone and NOT on workstation ? Same with margin : non margin on phone : why not on workstation to ?

tpartner:
@Shnoulle, as I said "unless you define a rule for it"!

tammoterhark:
@tpartner: please explain what you mean by rule.

I wanted to reduce the screen real estate on mobile phones, while keeping the appearance on larger screens not too cramped. That was my reason, but, as said, I like to learn from people that know than me...

Shnoulle:
Why reduce on phone and not on workstation ? 
@tpartner : yes , i know . It's the reset.css reason :)

tammoterhark:
@shnoulle; the whole idea of responsive design is to adapt to screen size, or what am I missing?

tammoterhark:
Because on phone the screen space is very limited. Many sites also reduce h1 size and other sizes..

tpartner:
@Shnoulle, thanks for explaining CSS to me.

@tammoterhark, I agree, different font sizes may be warranted for different screen sizes.

Shnoulle:
@tpartner : you start  :smiley_cat: 

 :angel:

LouisGac:
Reducing the size of characters and inputs on small screens can be dangerous. 

The main goal is not that everything fit in the small screen. Rhe main goal is that everything remain readable on small screens, and, more important, that user can interact with inputs on small screens. 

If the characters and the labels are too small, participant could click on the wrong item without noticing it. So the line height should be of the size of an average finger. Not less: what ever is the screen size. 

People having very small screens are used to have scroll bar in the website they visit. So it's not a big deal.

LouisGac:
@see: https://www.smashingmagazine.com/2012/02/finger-friendly-design-ideal-mobile-touchscreen-target-sizes/

tammoterhark:
@LouisGac : I was aiming at reducing the size of help items. They cannot be clicked on (can they?). I thought reducing them would not be a problem. I will never try to reduce the size of something that needs to be "clickable".

I still have another idea for reducing screen real estate, but that has to do with the menu bar on top. Shall I make another Mantis request for that?

LouisGac:
ok, so not a probem for me. 
Just, like said shnoulle, I agree that it should affect both Desktop and phone version.

Shnoulle:
For h1/.h1 h2/.h2 : it show bigger. Not an issue if we show it 'lower' on phone. But need to be bigger than other.
My POV

tammoterhark:
OK, I understand both of you (I guess). If it needs to be uniform on large/small screen, I would like to propose to make the help texts a little smaller, just to make them a little less important than the question text and the label texts. Could we agree on that? Or shall we just leave that for template builders?

LouisGac:
seems good to me.

tammoterhark:
@LouisGac 

> seems good to me.

Does this refer to:

1. Could we agree on that? (Making all help text a little smaller)
2. Or shall we just leave that for template builders?

LouisGac:
1.

Shnoulle:
1. : yes. Or lighter (in color)
2. : in template question.pstpl. Don't think we have to force size/color in core except using success/error/warning color.

tammoterhark:
Great, I think we have an agreement! ;-)

PR #595, opened 2016-11-21
ga('send', 'pageview') is generated 2 times

autopoiesis:
ga('send', 'pageview') is generated 2 times in the Google Analytics tracking code when "Survey-SID/Group" is selected as GA style in the survey settings.


c-schmitz:
Thank you - for future PRs please use correct Git commit messages. See https://manual.limesurvey.org/Standard_for_Git_commit_messages

PR #596, opened 2016-11-23
Develop ckeditor46

BertHankes:


c-schmitz:
That does not look right - into master from develop?

BertHankes:
You are right: see https://bugs.limesurvey.org/view.php?id=11925

BertHankes:
Wrong action

PR #597, opened 2016-11-23
FIxed issue #11947: Update Session.php

BertHankes:
uploading this file on unix system guarded by ASL and clamav sees this file as a virus. It responds to chr(hexdec()) (line 72). Working with chr( hexdec( ) ) is doing the trick

### DIFF
--- a/application/models/Session.php
+++ b/application/models/Session.php
@@ -69,7 +69,7 @@ class Session extends CActiveRecord
     private function hexToStr($hex){
         $string='';
         for ($i=0; $i < strlen($hex)-1; $i+=2){
-            $string .= chr(hexdec($hex[$i].$hex[$i+1]));
+            $string .= chr( hexdec( $hex[$i].$hex[$i+1] ) );
         }
         return $string;
     }


Shnoulle:
> Working with chr( hexdec( ) ) is doing the trick

ASV + ClamAV bug here : this must not update their control :)))

BertHankes:
ASL (Atomic Secured Linux) + clamav... Clearly a false positive, but a bug on their behalf? 
This is the only catch! So why not update the source!?

Shnoulle:
I say : if any virus maker can just add a space to become good : ASL/ClamAV are buggy :) .

I don't say any word on this commit :)

BertHankes:
ASL is very capable. Buggy? Far from it. It is doing a great job on our server keeping the Sh*%$t out! :-)

Shnoulle:
:wink: Yes but update `chr(hexdec(` to `chr( hexdec( ` can be made by you, then can be made by any virus maker :smirk:

BertHankes:
I know... but a I'm capable of doing something about it. So why not?
I will put this case to the ASL team (no direct access) so it will take a while.
In the maintime... commit?

c-schmitz:
Bert, we need would proper commit message, otherwise we cannot accept the PR.
You just renamed the pull request in othe cases which has no effect on the commit messages.
Can you please fix that and create new pull requests? For the obsolete files you could also bundle the commits in one.
Thank you for your work!

PR #599, opened 2016-11-24

BertHankes:
Because of revoke of jqgrid


LouisGac:
the commit message is not good:

But the commit itself seems good to me.

PR #602, opened 2016-11-24

BertHankes:
Fixed issue #11923: Obsolete files in 161118 (update styles/*/css/fonts.css)
Dev: Because of revoke of jqgrid and participantdisplay.js should be done in all the templates!
Dev: remove jqueryui fa replacement


Shnoulle:
Hi @BertHankes : the title of the pull request is OK, but not the commit comment. Here : you commit comment are `Update fonts.css` , and (think) , it must be `Dev: remove jqueryui fa replacement ` .

For the `Dev: ` part : i'm sure it was :) .

See https://github.com/LimeSurvey/LimeSurvey/pull/597#issuecomment-262713781 : **commit** message :)

BertHankes:
Hi Denis... I'm lost...  The fonts.css needs cleaning and belongs to issue 11923.
The comment: Dev: Because of revoke of jqgrid and participantdisplay.js should be done in all the templates! IMHO is correct... I can't do them all

Shnoulle:
It's not about commit , no issue here : it's good. It's about **commit message** : https://manual.limesurvey.org/Standard_for_Git_commit_messages#Syntax

PR #607, opened 2016-12-06
Feature/user stats view

KRav007:
I saw this ugly template. I added the bootstrap classes and using the bem class naming.
If you confirm the naming, I could write a path for the templates and delete the old classes and ids

### DIFF
--- a/application/views/statistics_user_view.php
+++ b/application/views/statistics_user_view.php
@@ -1,12 +1,20 @@
 <script type='text/javascript'>
-    var graphUrl="<?php echo Yii::app()->getController()->createUrl("admin/statistics/sa/graph"); ?>";
+    var graphUrl = "<?php echo Yii::app()->getController()->createUrl("admin/statistics/sa/graph"); ?>";
 </script>
 
-<div id='statsContainer'>
-    <div id='statsHeader'>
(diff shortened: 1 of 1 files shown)


Shnoulle:
Hi,

Think this one must be in [develop](https://github.com/LimeSurvey/LimeSurvey/tree/develop) + in the good directory : https://github.com/LimeSurvey/LimeSurvey/tree/develop/application/views/survey/frontpage + use more views ($statisticsoutput must use view too).

Something different : i really didn't understand why using row > col-sm-12. Using row > col-sm-X if you need : diferenet width for one row OR if you nhave multiple rows.

boostrpa don't NEED always row/col-sm-X

Shnoulle:
See http://testing.sondages.pro/answershtml/index.php/statistics_user/282267?language=fr , started reworking on 'table' class. Don't remind when ... (and must use some views here ...)

LouisGac:
I agree with Shnoulle, but this PR still improves the actual rendering of master.

Shnoulle:
@LouisGac : you make the twig files ;) , need merging master in develop and develop in twig :/

LouisGac:
I know... I will assume :'(

PR #610, opened 2016-12-07
Dev: boostrap package management

Shnoulle:
Purpose : find a way to move boostrap-rtl to package and not to template only

- boostrap-rtl need to be after boostrap
- currently : we don't 'add' boostrap even if we tell it here : https://github.com/LimeSurvey/LimeSurvey/blob/develop/templates/default/config.xml#L89
- Maybe good solution can be (if framework is bs) : 
  - add bootstrap from template is exist , else add default bootstrap
  - add boostrap-rtl if we are in rtl mode

Proposed solution need fix

1. Add boostrap package if framework is boostrap
2. Add boostrap.css from template (BUT : need testing existence and good dir)
3. Add boostrat-rtl package if needed.

The BUT is the question, any idea ? What it's seems to be the best way to kwow if 'boostrap.css' exist and the complete url of this file


Shnoulle:
OK, think i got an idea :) 

**1**

Use a real XML (revert https://github.com/LimeSurvey/LimeSurvey/commit/867f830c71ffff588e41177ebc2d66a049607dcf)

**2**

Use attribute for some part , exemple : 
````
        <css>
            <filename>css/template.css</filename>
            <filename dir='ltr'>css/template-ltr.css</filename>
            <filename dir='rtl'>css/template-rtl.css</filename>
        </css>
````

**3**

For package : allow dir too 

**4** 

For cssframework : allow replacing specific files from package, exemple with boostrap.
````
<cssframework css:boostrap="css/bootstrap.css" css:yiustrap="">bootstrap</cssframework>
````

Then a 'cssframework' extension/plugin/module (in future) propose a complete package + adapted views.
The template can easily update one css file only.

Exemple with jquery-ui last version 
````
<cssframework css:theme="css/ui.theme.css">jqueryui</cssframework>
````
because jqueryui cssfrmaework package offer 2 css : [core.css + theme.css](https://api.jqueryui.com/theming/css-framework/) (I don't say i do a jqueryui package : it's just because THIS framework have a "theme.css" for the example.

Shnoulle:
PS : i replace whole boostrap.css by the last one, then a template without replacing it include default boostrap.css.

For old version it's not really an issue because "theme" css files is loaded after. Just 2 files for the same prupose : don't break but can be improved.

Shnoulle:
Last version : don't use dir attribute, but use replace attribute : allow to remove some css (or some js ) from a framework.

Starting code not 'framework' dependant : external system can add other framework (plugin or config can do it currently)

PR #611, opened 2016-12-07
Dev: use mandatory for errors (test in radio list)

Shnoulle:
Issue in EM file 👎
capture du 2016-12-07 16-38-50

I try to fix (and understand the WTF)


Shnoulle:
OK, master too : https://bugs.limesurvey.org/view.php?id=11867

LouisGac:
I'll fix it tomorrow

PR #613, opened 2016-12-13
develop : separe settings in survey and question

Shnoulle:
# More accordion #
Seems REALLY better and needed for questions advanced settings

## survey settings: ##
separate_survey_plugins_settings

## question settings: ##
question_settings
**and**
question_settings2

Real accordion in questions settings (show one hide other)

### DIFF
--- a/application/views/admin/survey/Question/advanced_settings_view.php
+++ b/application/views/admin/survey/Question/advanced_settings_view.php
@@ -3,23 +3,37 @@
  * This view generate the advanced question attributes
  */
 $currentfieldset='';
+$categoryNum=0;
 ?>
 <!-- Advanced Settings -->
 <?php foreach ($attributedata as $index=>$aAttribute):?>
 
     <!-- Fieldsets -->
     <?php if ($currentfieldset!=$aAttribute['category']): ?>
+        <?php $categoryNum++; ?>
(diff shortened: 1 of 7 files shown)


LouisGac:
+1

Shnoulle:
@LouisGac : about "survey settings" any reason why it's not a "real" accordion ?

LouisGac:
what you mean by "real" ?

Shnoulle:
See http://getbootstrap.com/javascript/#collapse-example-accordion : open a collapse close all other.

I'm never sure what is the best between the 2 (auto-close or not): open one only, or close only when user really want it.

In general : i like to allow user to do 'what he want', but for accordion , i'm unsure of the good behavior : auto close or not.

LouisGac:
Well, I can tell you that you're the one who specifically asked for removing the behavior "open one collapse other" :-D  I think it's the best behavior. 

You also asked to keep opened the opened accordions on page reload: I never did it (not that hard, but would take time, and to do it everywhere, best would to move the accordion with those specific LimeSurvey features in a widget)

Shnoulle:
Arf ...... it's about "open the one already opened" more after submit .

You'r right : "open one collapse other" is better. I fix it.

Shnoulle:
Must fix : switch are broken currently

PR #614, opened 2016-12-26
Update to PHPmailer 5.2.19

Jan-E:
@c-schmitz @tammoterhark See https://bugs.limesurvey.org/view.php?id=12020
AFAIK this is the necessary upgrade.

### DIFF
--- a/application/extensions/SettingsWidget/SettingsWidget.php
+++ b/application/extensions/SettingsWidget/SettingsWidget.php
@@ -360,7 +360,7 @@ class SettingsWidget extends CWidget
         $editorOptions = array_merge(array(
             'html' => true,
         ), isset($metaData['editorOptions']) ? $metaData['editorOptions'] : array());
-        return Chtml::tag('div', array("style"=>'height:auto;width:100%'),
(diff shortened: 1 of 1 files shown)


Jan-E:
@Shnoulle could you check if this update is consistent with the previous PHPmailer update? See https://github.com/LimeSurvey/LimeSurvey/pull/452

Jan-E:
BTW: I took the release tarball for 5.2.19 from https://github.com/PHPMailer/PHPMailer/releases, deleted the /examples, but updated the MD files from git head at https://github.com/PHPMailer/PHPMailer

git head is on par with 5.2.19, so feel free to use that in stead of the release tarball.

lacrioque:
Thank you very much for your work.
Can you please use our correct commit message formatiing?
In this case the correct commit message would be "Fixed issue #12020: Critical Vulnerability in PHPMailer" 

We need othis formatting so sort the commit to the correct error report and to have our internal search and filter methods working. Thank you!

Jan-E:
New PR: https://github.com/LimeSurvey/LimeSurvey/pull/615

PR #619, opened 2017-01-13
Hide IFrame embedding security option when XSS protection is active.

Cheatha:
IFrames [seems to be filtered](https://www.limesurvey.org/forum/can-i-do-this-with-limesurvey/108110-embedding-youtube-via-iframe-with-active-xss-protection) when XSS protection is active. Showing the »IFrame embedding allowed« option is suggesting it is valid to use iFrames with active XSS protection. This PR hides the »IFrame embedding allowed« option when XSS protection is active.

### DIFF
--- a/application/views/admin/global_settings/_security.php
+++ b/application/views/admin/global_settings/_security.php
@@ -48,23 +48,25 @@
     </div>
 </div>
 
-<div class="form-group">
-    <label class="col-sm-5 control-label"  for="force_ssl">
-    <?php if (Yii::app()->getConfig("demoMode")==true){ ?>
-    <span class="text-danger asterisk"></span>
-    <?php }; ?>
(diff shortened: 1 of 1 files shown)


c-schmitz:
Thank you for submitting a pull request but unfortnuately the patch has several issues before it can be accepted:

- The IFRAME option should not vanish but be disabled. We only hide options if the user has no permissions to use them. Also your patch does not consider the editing period. It only becomes applicable after the user saved once. Disabling/enabling should be happening during edit. 
- The commit message is wrong. Please refer to our documentation at https://manual.limesurvey.org/Standard_for_Git_commit_messages for the correct syntax on any commit messages.

Shnoulle:
This pull request can not be related to the forum post, the issue are still there : but 'IFrame embedding allowed' global settings are totally unrelated with XSS protection.

The only action done with this settings are set the header : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/common_helper.php#L1415

PR #621, opened 2017-01-19
New feature #12066: set config in console + update by console

Shnoulle:
- assetManager are not set for console, then updatedb only


LouisGac:
I like the feature, but I don't get the relation with the assetManager?

Shnoulle:
Because : when updating via GUI : [we update the time of some directory](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/UpdateForm.php#L371) : you made it now ? ;)

We can redo it without using assetManager

LouisGac:
yep, but it's not related to updatedb. 

It's called from here:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/Survey_Common_Action.php#L386

Shnoulle:
1st i name UpdateCommand, i renamed it UpdateDBCommand to be more clear. And republishAssets is called during the update process no ?

Shnoulle:
I must review it about runtimePath and 3.0 : do it

Shnoulle:
Reviewed : https://github.com/LimeSurvey/LimeSurvey/pull/628/files

- console and web use same logic 
- console get whole config
- update command cleaned

PR #622, opened 2017-01-20
Fix timeline remote

evently-nl:
Fixes the Remote Control Timeline function


evently-nl:
Hi, 

Hope i got all the git standards right, this fixes the remote control timeline function it was missing the "Permission::model()->" before the hasSurveyPermissions check.

If I need to change anything, please let me know!

LouisGac:
seems good to me

PR #627, opened 2017-01-20
New feature: New plugin event afterSurveyQuestionAssessment

gabrieljenik:
Dev: Allows the plugin to recalculate the assessment value for a question with standard assessment as well as to calculate assessment value for any question type.

### DIFF
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -1906,12 +1906,19 @@ function doAssessment($surveyid, $returndataonly=false)
                     $event = new PluginEvent('afterSurveyQuestionAssessment');
                     $event->set('surveyId', $surveyid);
                     $event->set('lang', $_SESSION['survey_'.$surveyid]['s_lang']);
-                    $event->set('gid', $field);
-                    $event->set('qid', $field['gid']);
+                    $event->set('gid', $field['gid']);
+                    $event->set('qid', $field['qid']);
 
                     if (array_key_exists('sqid', $field))
                     {
+
                         $event->set('sqid', $field['sqid']);
+                    }
+
+                    if (array_key_exists('aid', $field))
+                    {
+
+                        $event->set('aid', $field['aid']);
                     }
 
                     $event->set('assessmentValue', $assessmentValue);
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -1897,7 +1897,7 @@ function doAssessment($surveyid, $returndataonly=false)
                 }
 
                 // If this is a question (and not a survey field, like ID), save asessment value
-                if (!empty($field['title']))
+                if ($field['qid'] > 0)
                 {
                     /**
                      * Allow Plugin to update assessment value
@@ -1905,12 +1905,22 @@ function doAssessment($surveyid, $returndataonly=false)
                     // Prepare Event Info
                     $event = new PluginEvent('afterSurveyQuestionAssessment');
                     $event->set('surveyId', $surveyid);
-                    $event->set('question', $field);
+                    $event->set('lang', $_SESSION['survey_'.$surveyid]['s_lang']);
+                    $event->set('gid', $field);
+                    $event->set('qid', $field['gid']);
+
+                    if (array_key_exists('sqid', $field))
+                    {
+                        $event->set('sqid', $field['sqid']);
+                    }
+
                     $event->set('assessmentValue', $assessmentValue);
+
                     if (isset($_SESSION['survey_'.$surveyid][$field['fieldname']]))
                     {
                         $event->set('response', $_SESSION['survey_'.$surveyid][$field['fieldname']]);
                     }
+
                     // Dispatch Event and Get new assessment value
                     App()->getPluginManager()->dispatchEvent($event);
                     $updatedAssessmentValue=$event->get('assessmentValue', $assessmentValue);
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -1865,6 +1865,10 @@ function doAssessment($surveyid, $returndataonly=false)
             $groups=array();
             foreach($fieldmap as $field)
             {
+                // Init Assessment Value
+                $assessmentValue = NULL;
+
+                // Default Assessment Value Calculation
                 if (in_array($field['type'],array('1','F','H','W','Z','L','!','M','O','P')))
                 {
                     $fieldmap[$field['fieldname']]['assessment_value']=0;
@@ -1875,8 +1879,7 @@ function doAssessment($surveyid, $returndataonly=false)
                             if ($_SESSION['survey_'.$surveyid][$field['fieldname']] == "Y")
                             {
                                 $aAttributes=getQuestionAttributeValues($field['qid']);
-                                $fieldmap[$field['fieldname']]['assessment_value']=(int)$aAttributes['assessment_value'];
-                                $total=$total+(int)$aAttributes['assessment_value'];
+                                $assessmentValue = (int)$aAttributes['assessment_value'];
                             }
                         }
                         else  // Single choice question
@@ -1886,13 +1889,39 @@ function doAssessment($surveyid, $returndataonly=false)
                             if ($usresult)
                             {
                                 $usrow = $usresult->read();
-                                $fieldmap[$field['fieldname']]['assessment_value']=$usrow['assessment_value'];
-                                $total=$total+$usrow['assessment_value'];
+                                $assessmentValue = $usrow['assessment_value'];
                             }
                         }
                     }
                     $groups[]=$field['gid'];
                 }
+
+                // If this is a question (and not a survey field, like ID), save asessment value
+                if (!empty($field['title']))
+                {
+                    /**
+                     * Allow Plugin to update assessment value
+                     */
+                    // Prepare Event Info
+                    $event = new PluginEvent('afterSurveyQuestionAssessment');
+                    $event->set('surveyId', $surveyid);
+                    $event->set('question', $field);
+                    $event->set('assessmentValue', $assessmentValue);
+                    if (isset($_SESSION['survey_'.$surveyid][$field['fieldname']]))
+                    {
+                        $event->set('response', $_SESSION['survey_'.$surveyid][$field['fieldname']]);
+                    }
+                    // Dispatch Event and Get new assessment value
+                    App()->getPluginManager()->dispatchEvent($event);
+                    $updatedAssessmentValue=$event->get('assessmentValue', $assessmentValue);
+
+                    /**
+                     * Save assessment value on the response
+                     */
+                    $fieldmap[$field['fieldname']]['assessment_value']=$updatedAssessmentValue;
+                    $total=$total+$updatedAssessmentValue;
+                }
+
                 $i++;
             }
 


LouisGac:
seems good to me

Shnoulle:
I didn't understand : Equation question type + [Expression Manager](https://manual.limesurvey.org/Expression_Manager#Tailored_Questions.2C_Answers.2C_and_Reports) (for end message) replace assessment with a really more easy way.

gabrieljenik:
I think of this because some people still use Assessment, as it doesn't require deep development skills.

LouisGac:
In the forum, a lot of people seems to use assessments

Shnoulle:
But here : there are 2 solution 
- Use Equation question type to make calculation directly inside LimeSurvey core
- Create a plugin (then PHP dev) to make own assesments

ANd you say "it doesn't require deep development skills". You think PHP is more easy than Equation ?

gabrieljenik:
> ANd you say "it doesn't require deep development skills". You think PHP is more easy than Equation ?

No, of course not :)

It is just that for those people, some plugins can be developed.

application/helpers/frontend_helper.php
```diff
@@ -1886,13 +1889,39 @@ function doAssessment($surveyid, $returndataonly=false)
...
-                                $total=$total+$usrow['assessment_value'];
+                                $assessmentValue = $usrow['assessment_value'];
                             }
                         }
                     }
                     $groups[]=$field['gid'];
                 }
+
+                // If this is a question (and not a survey field, like ID), save asessment value
+                if (!empty($field['title']))
```
Shnoulle: There are no other way to see if question can be a real question ?

LouisGac:
and I bet that the plugin Gabriel is writting is easier to use than PHP or equation for end user. 

But your solution using equation seems interesting Denis.

application/helpers/frontend_helper.php
```diff
@@ -1886,13 +1889,39 @@ function doAssessment($surveyid, $returndataonly=false)
...
+                {
+                    /**
+                     * Allow Plugin to update assessment value
+                     */
+                    // Prepare Event Info
+                    $event = new PluginEvent('afterSurveyQuestionAssessment');
+                    $event->set('surveyId', $surveyid);
+                    $event->set('question', $field);
```
Shnoulle: No : must send qid (the question id) + sub(s) question code

application/helpers/frontend_helper.php
gabrieljenik: hmm, I bet yes.
but... Is there anything bad with this one?
Maybe I am missing something...

application/helpers/frontend_helper.php
```diff
@@ -1865,6 +1865,10 @@ function doAssessment($surveyid, $returndataonly=false)
             $groups=array();
             foreach($fieldmap as $field)
             {
+                // Init Assessment Value
+                $assessmentValue = NULL;
```
Shnoulle: Maybe do this only if assesment are enabled for the survey. Less code produced for a lot of page with potentially a lot of question.

application/helpers/frontend_helper.php
gabrieljenik: Can I ask why?

application/helpers/frontend_helper.php
```diff
@@ -1886,13 +1889,39 @@ function doAssessment($surveyid, $returndataonly=false)
...
                 }
+
+                // If this is a question (and not a survey field, like ID), save asessment value
+                if (!empty($field['title']))
+                {
+                    /**
+                     * Allow Plugin to update assessment value
+                     */
+                    // Prepare Event Info
+                    $event = new PluginEvent('afterSurveyQuestionAssessment');
```
Shnoulle: Than for an array text 5x5 we have 25 for one question, right ?

application/helpers/frontend_helper.php
gabrieljenik: I just worked on the standard code. Added the event on the same level / code block as the standard assessment process.

application/helpers/frontend_helper.php
gabrieljenik: Yes, but that's how the current assessment process is currently working

application/helpers/frontend_helper.php
Shnoulle: qid is the id of the question : we send qid for 'this question' : the parent question. Send same information for plugin dev please.
Here : we have to send qid (parent qid) + other informations for sub(s)-question(s).

here field come from fielmap : we are REALLY unsure if we kee fieldmap in 4/5 ... release

application/helpers/frontend_helper.php
Shnoulle: I ask if there a better way, when we update/remove/deprecate fieldmap ....

gabrieljenik: Well, when that happens, this should be updated too. In the current process, fieldmap is a core thing, so this will not be an extraordinary situation. It will be just couple of more lines to update.

As far as now, I don't find other ways without involving fieldmap.

application/helpers/frontend_helper.php
gabrieljenik: I understand. By sending fieldmap, we avoid having to requery the DB for getting informaiton about the question. That's why I like sending more info than IDs.

I can send instead an array that extracts info from fieldmap. So if fieldmap changes, the array interface keeps fixed and then we only have to update one side of the interface.

What do you think?

Shnoulle: Thinking, for question/subquestion (and answer)

- Allways send `surveyid` as `surveyid` 
- Seems best to send `qid` as `qid` for PARENT qid (the one you can edit directly in admin gui)
- For subquestion : didn't know for sure : title or qid + scale_id ? 
- For 'database column'/'answer' : `sgq` seems good (else we need to [construct](https://framagit.org/SondagePro-LimeSurvey-plugin/arrayTextAdapt/blob/master/arrayTextAdapt.php#L156) each time)
- Seems interseting to have question type (but not mandatory)

It's the first plugin event for SGQ, then we must choose seriously. I really want a plugin event in [ExpressionManager _validateQuestion](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/expressions/em_manager_helper.php#L6095) : then this plugin fix some part we send in next pluginb for SGQA

application/helpers/frontend_helper.php
Shnoulle: Yes : it's an issue of old code, but old code didn't send an event + didn't search in DB for all question type.

application/helpers/frontend_helper.php
Shnoulle: In Array text ? No i don't think : array text => no assesment

Shnoulle:
PS : for a new plugin using this event : you add an Question Attribute ? Right ? Question Attribute : qid (parent_qid)

application/helpers/frontend_helper.php
gabrieljenik: It did searched db for all subquestions.
This covers all question types, which gives lot of felxibility.

application/helpers/frontend_helper.php
gabrieljenik: I was referring that the current process each subquestion of  question.
This has more scope ==> more flexibility.

gabrieljenik:
That's open for the plugin developer to decide.

application/helpers/frontend_helper.php
gabrieljenik: For a case I am working on, I need the question text. Also the code could be usefull if developers want to apply code for only some specific questions. All that info is available. If we don't send it, people may have to requery.

gabrieljenik: > For 'database column'/'answer' : sgq seems good
What do you mean? Sorry.

> It's the first plugin event for SGQ, then we must choose seriously
What do you mean? Sorry.

application/helpers/frontend_helper.php
gabrieljenik: I was reviewing code. This code is inside doAssessment function.
If we only want this to execute when assessment is enabled, I guess the best is to look for those places and check the settings there.
Example: 
- replacements_helper.php around line 545

It is infact cached on SurveyRuntimeHelper.php around line 708, 757 and 783

application/helpers/frontend_helper.php
Shnoulle: database/column : it's the SGQA : currently `surveyidXgroupidXquestionidSubQuestionY_SubQuestionY`

It's the first plugin who send information for Sub question (the final answer X times for array/multiple choice XxY times for array text/number : if we send ABC information in this plugin, we must send ABC information in next plugin too ....

Shnoulle: For the question text : please : get it for model. Each time you send an information : you puit some micro memory more in PHP. But micro memory + micro memory + .... make 512 minimum to lauch a little survey.

gabrieljenik: > It's the first plugin who send information for Sub question (the final answer X times for array/multiple choice XxY times for array text/number : if we send ABC information in this plugin, we must send ABC information in next plugin too ....

Partly agree. I am more into saying some basic information shall be shared among plugins and that there can be specific info for each plugin.

> For the question text: get it for model.
You mean, get it from the Model (from DB), right?

>  But micro memory + micro memory + .... make 512 minimum to lauch a little survey.
I understand... I agree for the question text. Still, sending the question title (limited string) could be a good idea, as it is much more basic info. Don't you agree?

Thinking forward, maybe it is just easier to use a singleton to store fieldmap (or similar) as a app-wide cache.

Shnoulle: > Partly agree. I am more into saying some basic information shall be shared among plugins and that there can be specific info for each plugin.

Nothing must be specific for plugin, only for event. And here SGQA/qid/surveyid seems to be widely used after in any plugin.

We already have : surveyid or survey for survey id :/ . I added gid because seems to be a common var for beforeQuestionRender and some other (i don't need it specially , i can play without) , but try to get the middle betwwen : needed and lot used information and more information than needed. .

gabrieljenik: OK, will update it and let you know.

application/helpers/frontend_helper.php
gabrieljenik: Will update this to checking $field['qid']

Shnoulle:
Yes, but seems to be the easiest way :). I have a plugin where using subqid, but more an hack .

gabrieljenik:
Updated plugin interface. Now seems more similar to what you expected.
PS: Where have all our previous conversations gone? After my code-updates seem to be gone.

application/helpers/frontend_helper.php
```diff
@@ -1897,20 +1897,30 @@ function doAssessment($surveyid, $returndataonly=false)
...
-                    $event->set('question', $field);
+                    $event->set('lang', $_SESSION['survey_'.$surveyid]['s_lang']);
+                    $event->set('gid', $field);
+                    $event->set('qid', $field['gid']);
+
+                    if (array_key_exists('sqid', $field))
+                    {
+                        $event->set('sqid', $field['sqid']);
+                    }
+
```
Shnoulle: OK for subquestion with one choice but not for Array/text, Array/number where [createFieldMap](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/common_helper.php#L2064) are silly .... 

Why not add [fieldname](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/common_helper.php#L2087) ? It's the DB name + session name

application/helpers/frontend_helper.php
```diff
@@ -1897,20 +1897,30 @@ function doAssessment($surveyid, $returndataonly=false)
...
+                if ($field['qid'] > 0)
                 {
                     /**
                      * Allow Plugin to update assessment value
                      */
                     // Prepare Event Info
                     $event = new PluginEvent('afterSurveyQuestionAssessment');
                     $event->set('surveyId', $surveyid);
-                    $event->set('question', $field);
+                    $event->set('lang', $_SESSION['survey_'.$surveyid]['s_lang']);
```
Shnoulle: I think it's better to use App()->language in plugin except for admin part where we can have different lang.

gabrieljenik: It is the same used couple of lines before. I just wanted to keep consistency.

application/helpers/frontend_helper.php
gabrieljenik: Adding fieldname would require parsing that all time (memory + cpu + not that comfortable). :)
Why this is not OK for "Array/text, Array/number"? I don't get it, sorry.

Thanks

Shnoulle: It's not in field ? 

Arry/text + Array/number have 2 sub question : one for X and one for Y. Here you get only the Y subquestion. The X are in aid (maybe ????) 

field are really a broken thing : i really think we must not use it .....

application/helpers/frontend_helper.php
Shnoulle: You're right : but old code are sometimes broken. 
PS remind : i'm not the one who decide here ;)

application/helpers/frontend_helper.php
gabrieljenik: > Adding fieldname would require parsing that all time (memory + cpu + not that comfortable). :)

I meant it would require parsig all time from the plugin side.

> The X are in aid (maybe ????)
Ohh I see what you mean. I will also send it.

application/helpers/frontend_helper.php
gabrieljenik: > PS remind : i'm not the one who decide here ;)
I know, but your review is important.

> You're right : but old code are sometimes broken.
As to minimize new bugs and keep consistency, I just prefer to keep it that way.

application/helpers/frontend_helper.php
gabrieljenik: I added aid to the interface. That should cover all options.

gabrieljenik:
Hi Guys, Let me know what you think at final.
If OK, would be great, as I would update my installation accordingly and start using the plugin.

olleharstedt:
I guess we can bring it up at the next team meeting?

gabrieljenik:
Great! Thanks

olleharstedt:
@gabrieljenik Please also add a manual page for this event.

gabrieljenik:
https://manual.limesurvey.org/AfterSurveyQuestionAssessment

olleharstedt:
Great, thanks!

PR #628, opened 2017-01-20
2 console feature + better runtimePath management

Shnoulle:
- Usage of same logic in console and web
- User config can replace tmp (via `config['config']['tempdir']`) or only runtimePath (Yii way)
- Document information for runtimePath in config.php


Shnoulle:
@c-schmitz I quickly test install via console, but not install until enbd. Can you check ?

PR #629, opened 2017-01-23
Dev: constants for question types

jiripavlicek:
I made small improvement for better code readability/manageability - there is used so much strings like 'A', 'B', 'C', ... in source in conditions for question types. I replaces those string by constants. Why? Modern IDE's can show you where constant/variable/class/whatever else is used in source, but searching for string like 'A' can't return the same comfortable result. I made it fo question types, but there are other string for graph types and in import/export. may be next time I'll improve other source code part this way.

This isn't any bug/issue report. This is created as announce of my pull request for improve code readability/quality. It doesn't fixes any bug/issue or add new feature. This is only code development improvement.

https://bugs.limesurvey.org/view.php?id=12074


maziminke:
In general I like this improvement. You do not have to look up the codes for different question types when adjusitng detals for a certain QT. It makes the code much more readable.

olleharstedt:
Good idea.

olleharstedt:
Is this PR complete? Did you update all switches on question type, even in the expression manager?

olleharstedt:
Also switches in em_javascript.js.

Shnoulle:
For js : we already have https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Question.php#L653 no ? Or add it in data-qtype

LouisGac:
if you replaced the question code by those constants everywhere: it's a very food change.
It's just a pity you did it in master rather rather than in develop....

jiripavlicek:
Hi LouisGac, I red https://manual.limesurvey.org/How_to_contribute_new_features and there is "1. Use the latest Github version for your development", but nothing about branch. So I used latest version. Please add it to this page to better instruct peole about contributting.

LouisGac:
Don't worry, we're still going to use your work. 
Thank you for that great contribution.

jiripavlicek:
Hi olleharstedt, I know, that in em_javascript.js file are question type strings too, but this is static JavaScript file not generated by PHP, so I can't replace constants in it without change this file to dynamic generated by PHP.

machaven:
This is a really good change, that makes the code much more readable.

For the em_javascript.js you can achieve the same goal, by creating an object that is frozen.

```
var QuestionTypes = Object.freeze({
    EXCLAMATION_LIST_DROPDOWN: '!',
    L_LIST_DROPDOWN: 'L',
    O_LIST_WITH_COMMENT: 'O',
   ......
});
```

then changing case statements and other usages in the js file, like so:

```
                switch (attr.type) {
                    case QuestionTypes.EXCLAMATION_LIST_DROPDOWN:
                    case QuestionTypes.L_LIST_DROPDOWN:
                    case QuestionTypes.O_LIST_WITH_COMMENT: 
```

TonisOrmisson:
I also like ti very much and would like to see it implemented

LouisGac:
well, feel free to convert the commit for develop branch ;-)

c-schmitz:
Thanks. This pull request has been integrated manually.

PR #632, opened 2017-02-01
New feature #12102: Allow loading desired template from URL params

TonisOrmisson:
Use case:

I am using LimeSurvey as data collector for mixed mode surveys. EG. Its the data collection engine for CATI as well as online filling. A typical scenario is when a respondent is contacted via telephone and interview is started and then respondent wishes to complete the interview on its own online.

Now - the problem is that CATI template has some specific custom UI elements - some extra buttons & info fields that are CATI specific and not needed for online direct filling. The easy fix is to allow loading a different template depending on the mode (CATI vs online).

https://bugs.limesurvey.org/view.php?id=12102

### DIFF
--- a/application/controllers/survey/index.php
+++ b/application/controllers/survey/index.php
@@ -36,7 +36,15 @@ class index extends CAction {
         $param = $this->_getParameters(func_get_args(), $_POST);
         $surveyid = $param['sid'];
 
-        $oTemplate = Template::model()->getInstance('', $surveyid);
+        // load target template by name if specified && exists
+        if(isset($param['templatename']) && in_array($param['templatename'],array_keys(Template::getTemplateList())) ){
+            $templateName = $param['templatename'];
+        }else{
+            $templateName = '';
+        }
+
+        $oTemplate = Template::model()->getInstance($templateName, $surveyid);
+
         $this->oTemplate = $oTemplate;
         App()->clientScript->registerScript('sLSJavascriptVar',$sLSJavascriptVar,CClientScript::POS_HEAD);
         App()->clientScript->registerScript('setJsVar',"setJsVar();",CClientScript::POS_BEGIN);// Ensure all js var is set before rendering the page (User can click before $.ready)
@@ -144,9 +152,9 @@ class index extends CAction {
         {
             $sReloadUrl=$this->getController()->createUrl("/survey/index/sid/{$surveyid}",array('token'=>$clienttoken,'lang'=>App()->language,'newtest'=>'Y'));
             $asMessage = array(
-            gT('Token mismatch'),
-            gT('The token you provided doesn\'t match the one in your session.'),
-            "<a class='reloadlink newsurvey' href={$sReloadUrl}>".gT("Click here to start the survey.")."</a>"
+                gT('Token mismatch'),
+                gT('The token you provided doesn\'t match the one in your session.'),
+                "<a class='reloadlink newsurvey' href={$sReloadUrl}>".gT("Click here to start the survey.")."</a>"
             );
             $this->_createNewUserSessionAndRedirect($surveyid, $redata, __LINE__, $asMessage);
         }
@@ -157,9 +165,9 @@ class index extends CAction {
             if($clienttoken){$aReloadUrlParam['token']=$clienttoken;}
             $sReloadUrl=$this->getController()->createUrl("/survey/index/sid/{$surveyid}",$aReloadUrlParam);
             $asMessage = array(
-            gT('Previous session is set to be finished.'),
-            gT('Your browser reports that it was used previously to answer this survey. We are resetting the session so that you can start from the beginning.'),
-            "<a class='reloadlink newsurvey' href={$sReloadUrl}>".gT("Click here to start the survey.")."</a>"
+                gT('Previous session is set to be finished.'),
+                gT('Your browser reports that it was used previously to answer this survey. We are resetting the session so that you can start from the beginning.'),
+                "<a class='reloadlink newsurvey' href={$sReloadUrl}>".gT("Click here to start the survey.")."</a>"
             );
             $this->_createNewUserSessionAndRedirect($surveyid, $redata, __LINE__, $asMessage);
         }
@@ -189,9 +197,9 @@ class index extends CAction {
             if ($bPreviewRight === false)
             {
                 $asMessage = array(
-                gT("Error"),
-                gT("We are sorry but you don't have permissions to do this."),
-                sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
+                    gT("Error"),
+                    gT("We are sorry but you don't have permissions to do this."),
+                    sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
                 );
                 $this->_niceExit($redata, __LINE__, null, $asMessage);
             }
@@ -277,9 +285,9 @@ class index extends CAction {
         {
             $redata = compact(array_keys(get_defined_vars()));
             $asMessage = array(
-            gT("Error"),
-            gT("We are sorry but the survey is expired and no longer available."),
-            sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
+                gT("Error"),
+                gT("We are sorry but the survey is expired and no longer available."),
+                sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
             );
 
             $this->_niceExit($redata, __LINE__, $thissurvey['templatedir'], $asMessage);
@@ -290,9 +298,9 @@ class index extends CAction {
         {
             $redata = compact(array_keys(get_defined_vars()));
             $asMessage = array(
-            gT("Error"),
-            gT("This survey is not yet started."),
-            sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
+                gT("Error"),
+                gT("This survey is not yet started."),
+                sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
             );
             $this->_niceExit($redata, __LINE__, $thissurvey['templatedir'], $asMessage);
         }
@@ -304,9 +312,9 @@ class index extends CAction {
         {
             $redata = compact(array_keys(get_defined_vars()));
             $asMessage = array(
-            gT("Error"),
-            gT("You have already completed this survey."),
-            sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
+                gT("Error"),
+                gT("You have already completed this survey."),
+                sprintf(gT("Please contact %s ( %s ) for further assistance."),$thissurvey['adminname'],$thissurvey['adminemail'])
             );
 
             $this->_niceExit($redata, __LINE__, $thissurvey['templatedir'], $asMessage);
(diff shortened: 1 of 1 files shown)


Shnoulle:
Currently : can easily be done in plugin : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L251

I i remind : original event was done to use a different template for mobile and workstation.

TonisOrmisson:
Thanks, I will look into this option!

TonisOrmisson:
I need a bit help. 
what is the value I need to pass there. I can't get the template to change passing template name nor the full path. Eg like that:
```
    public function __construct(PluginManager $manager, $id) {
        parent::__construct($manager, $id);
        $this->subscribe('beforeSurveyPage', 'processTest');
    }

    public function processTest(){
        $path = Template::getTemplatePath('default1');
        $this->event->set('template',$path);
    }
```

maziminke:
I'd love to see that feature in the next major version. 

Is there any additional information available about the plugin event that can be used?
If I remember correctly there had been problems in the past because sometimes either the CSS or the pstpl files were not loaded correctly.

@LouisGac Maybe you can contribute to what extend we have to adjust things for the planned usage of twig based templates!?

TonisOrmisson:
As far as I understand it should work by passing template name like that:
````    
    public function __construct(PluginManager $manager, $id) {
        parent::__construct($manager, $id);
        $this->subscribe('beforeSurveyPage', 'processTest');
    }

    public function processTest(){
        $this->event->set('template','default1');
    }
````
but this does not seem to work. In production mode nothing changes. In debugging I see that css file is not accessible. Anyway - seems to be problematic via plugin right now.

On the other hand - getting it directly from url & changing core works like a charm :)

**Comment by Shnoulle** [comment]
Any one can use any template on any survey without any control here ! Really against such feature in core.

@TonisOrmisson : yes : seems the original event are broken currently

**Comment by TonisOrmisson** [comment]
OK good point. Going via plugin seems a more reasonable way to go to implement some sort of authorization. 

And thank you for confirming that what I tried is not working!

I would definitely work on a plugin for that if it would actually work that way.

I would not oppose closing this since it seems not reasonable to do it in the suggested way.

**Comment by Shnoulle** [comment]
@TonisOrmisson : remind : i'm not the lead dev here :), it's not because i'm not for such system in core.

But if i want an event to allowing update a survey : maybe i look directly on model currently.

1. https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Survey.php#L630 
2. https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Survey.php#L672

I'm unsure of the situation of getSurveyInfo at long term .... :/ 

Maybe best are : 

1. Allow update more part of survey array in https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L251 , i think there are templatedir + templatename in this array
2. Maybe add some extra part (didn't know exactly which currently, but why not 'show question code' for example : i really like to have a way to show question code to admin, but not for public. Or survey format : allinone for a 'preview complete survey', only if user are admin :). It's just some example.

Currenltly : i thing if we use `$oTemplate = Template::model()->getInstance($thissurvey['templatedir'],'');` at  https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L271 : it can work for your purpose ? Can you do a quick test (for example force always your template ;).

**Comment by maziminke** [comment]
Denis is right about the security issues (anyone can load any template). Maybe we should just add a check if the survey owner is allowed to access the requested template and if not we load the assigned template!?

Still, I consider this a very useful feature and would like to see it being added to the core or at least as a plugin.

**Comment by TonisOrmisson** [comment]
My use case in principal means that I need to use different template not for admin users (having permissions). But for respondents (participants). Since I would be directing respondents from an interface having Remote Control to LimeSurvey - I would think setting a template over RC to specific respondent. Or getting any authorization via the RC somehow. Anyway the point is to be able to switch the template to Participant, not User.

**Comment by TonisOrmisson** [comment]
@Shnoulle Thanks. I was looking at the core code and I got a bit confused how the template was set. Seemed a bit messy & confusing. I did not succeed the way you suggested. I am hoping some (core?) developers could help to get this working as it seems it should be working. Then I would start on working a plugin.

**Comment by Shnoulle** [comment]
@TonisOrmisson : i think (but unsure) : 

Try to replace https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L271 by `$oTemplate = Template::model()->getInstance('$thissurvey['templatedir']', $surveyid);`

Test with your plugin https://github.com/LimeSurvey/LimeSurvey/pull/632#issuecomment-276637270 : if it's OK : you fix the bug :)

**Comment by Shnoulle** [comment]
Or better : 
remove this https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L271

And `$oTemplate = Template::model()->getInstance($event->get('template'), $surveyid); `

Can you make a try ?

**Comment by TonisOrmisson** [comment]
Thanks for trying to help. 
Tried what you suggested and does not seem to work for me. I get errors where CSS is being looked up in wrong folder. 

I get confused by template being initiated here: 
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L39

And then later again here: 
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/survey/index.php#L271

I do not understand the codebase enough to understand why this works this way and whether this is logical or not. Anyway this I get confused by. Tried multiple ways you suggested as well as other ways :(

Anyway. Chaniging the template in first initiation (inaccessible via plugin) seems to work. Have not managed to get the template changed via the second initialization.

**Comment by Shnoulle** [comment]
I take a quick look , but monday more ....

**Comment by LouisGac** [comment]
IMHO: you'd rather put your plugin event in templateConfiguration model

**Comment by LouisGac** [comment]
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/TemplateConfiguration.php#L67

**Comment by TonisOrmisson** [comment]
@LouisGac I'm sorry I do not follow. How would I access templateConfiguration from plugin?
You mean there could be a new event created to plugin events to access that?

**Comment by LouisGac** [comment]
yep that's what I meant.

**Comment by Shnoulle** [comment]
Think it's more linked to Survey : its' the template used by survey, not a functionality of the template.

Functionnality of the template can be alternatives views or css or .... but no link with survey.

**Comment by LouisGac** [comment]
I'm just saying that over ridding once the value of sTemplateName  in templateConfiguration would be enough. then any call to that instance will have the correct value.

**Comment by Shnoulle** [comment]
If adding a new Plugin event : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Survey.php#L248 seems the best place no ?

Because happen at the origin : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Survey.php#L49 of survey.

**Comment by TonisOrmisson** [comment]
Well. I tested both ways. 
Putting a new event to templateConfiguration will still fail to load css (eg. flat_and_modern.css).
Putting a new event to here:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Survey.php#L248
seems to work great in practice. 

Now I'm not sure which way is more logical or more in line for overall plugin events policy so far. 

Having an event afterFixSurveyAttribute could have the perspective of changing a lot more than just template. I'm not familiar with the broader policy of plugin events, but having that could add quite a lot of options in theory. 

And how can we go from here. Who decides if adding an event there is OK or not? If I knew it would be OK, I could add the event and see if there could be any other attributes that could be fixed via this?

**Comment by Shnoulle** [comment]
I'm not the man who decide ;)

Else : onAfterFindSurvey ? for the name ;) 

Maybe allowing some attribute update for staring ? Send whole attribute in a array ?

    $event->set('attributes',$this->attributes);
    $this->template=Template::templateNameFilter($event->get('template',$this->template));
    $this->format=$event->get('format',$this->format);
    $this->showqnumcode=$event->get('showqnumcode',$this->showqnumcode);

In plugin : remind to test if you are on Surveys controller : 

    if(Yii::app()->getController()=='survey'){
        // Public action
    }


BUT : remind : it's MY point of view

**Comment by Shnoulle** [comment]
PS about css : if i remind : we use templatename from session : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/replacements_helper.php#L104 .... surely set in loader.

**Comment by TonisOrmisson** [comment]
I scetched it like that:

```
        $event = new PluginEvent('onAfterFindSurvey');
        // set the attributes we won't allow to fix
        $deniedAttributes = array('datecreated');
        $allowedAttributes = array_diff(array_keys($this->attributes),$deniedAttributes);
        foreach ($allowedAttributes as $attribute){
            if (!is_null($event->get($attribute)))
            {
                $this->{$attribute} = $event->get($attribute);
            }
        }
```
Idea is to allow fixing most of the attributes and identify some that do not seem OK to be changed, eg. datecreated - probably some more, but I didn't go though it in detail

Shnoulle:
Thik there are more disallowed attribute than allowed.

Example : 

- must be set before activate (in beforeActivateSurvey event)
  - active
  - anonymized
  - datestamp
  - refurl
  - Some other (language , etc ... a lot on fact)
- Can be done with another more specific plugin
  - autoredirect
  -  htmlemail
  - sendconfirmation
  - ...
- Not used in find (single survey), but only in findAll (the nevent not happen), or used in findAll + in find
  - expires
  - publicstatistics

Can be strange is user update it  
- owner
- ...

It's the reason why i start with minimal risk :)

TonisOrmisson:
My point was to view this a bit broader than my issue at hand regarding the template :)

The attributes that are set before activate - I thought so to. 

The attributes that can be done with another plugin - I am not able to identify. Possibly could be accessed via multiple events? No?

The attributes "not used in single survey" - to be honest - I do not follow - Im not that familiar with the code. Testig resulted the "expires" to work as expected - but probably here some logic I do not follow right now. 

The "strange to be modified" attributes - I agree. 

Bottom line - full identification of possible disAllowedAttributes seems to be out of my scope right now. On the other hand It would be kind of a waste to put only the "template" here? There might be some some valuable use cases here for some other guys. 

Any help to identify reasoneble allowedAttributes would be appreciated. I have created an initial list to start with:

`
        $allowedAttributes = array(
            'usecookie',
            'allowregister',
            'allowsave',
            'autoredirect',
            'allowprev',
            'sendconfirmation',
            'showxquestions',
            'shownoanswer',
            'showwelcome',
            'showprogress',
            'questionindex',
            'nokeyboard',
            'alloweditaftercompletion',
            'usecaptcha',
            'showgroupinfo',
            'showqnumcode',
            'navigationdelay',
        );
`
Most of the attributes are regarding visual components which seems logical in this context.

Shnoulle:
About 'single survey' : if we add an event here : it's call only when using Survey::model()->find(), not with Survey::model()->findAll(). When we search for a lot of survey (the survey listing at home page, the survey list on admin page) : we don't use find, but findAll (or findAllByAttribute or ...) : then there are no reason to add 'attribute used only when findAll', and can be strange to set 'attribute used when findAll and when find'

For a better example expires : user can see the survey in the survey listing, but if he click on it and a plugin set expires to 2000-01-01 : he can not access to it :).

About : "Can be done in another plugin event' : i think some attribute is not 'whole survey view' but more "specific survey part' , for example 'autoredirect' seems very linked to afterSurveySubmitted, no ? 

In fact: if we put an `$allowedAttributes` we can set a minimal list for starting, and adding some good part after. I really think we can start by `'usecookie',   'allowprev', 'showxquestions', 'shownoanswer',  'showprogress', 'questionindex',   'usecaptcha', 'showgroupinfo', 'showqnumcode', 'navigationdelay'`

Clearly in afterSurveySubmitted event : `'autoredirect','sendconfirmation',`

Think's it's more related to anbothger event, but unclear (i'm totally unsure) `'alloweditaftercompletion','allowsave', 'allowregister',`

About `'nokeyboard',` : this settings just add a javascript system. I think 1. This settings is outdated (all phone have cleaner way to show keyboard) + a javascript like this is more a template issue.

About `showwelcome` : unsure if we must/can add it : why not ? But why too ;) .

TonisOrmisson:
@Shnoulle Thanks for the help! 

I created a new PR #634

TonisOrmisson:
@maziminke plugin: https://github.com/TonisOrmisson/limesurvey-url-templates
needed changes to the core will be with next release

PR #634, opened 2017-02-07
New feature: add Plugin event afterFindSurvey

TonisOrmisson:
This is a further development of a PR started here #632:

The initial idea was to allow template to be changed to participants via plugin events. Existing events did not work to effectively change the template.

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -251,10 +251,10 @@ class Survey extends LSActiveRecord
 
         $event = new PluginEvent('afterFindSurvey');
         App()->getPluginManager()->dispatchEvent($event);
+        // set the attributes we allow to be fixed
         $allowedAttributes = array( 'template','usecookie', 'allowprev',
             'showxquestions', 'shownoanswer', 'showprogress', 'questionindex',
             'usecaptcha', 'showgroupinfo', 'showqnumcode', 'navigationdelay');
-        // set the attributes we won't allow to fix
         foreach ($allowedAttributes as $attribute){
             if (!is_null($event->get($attribute)))
             {
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -248,15 +248,14 @@ class Survey extends LSActiveRecord
     public function fixSurveyAttribute($event)
     {
         $event = new PluginEvent('afterFindSurvey');
-        $event->set('surveyid',$this->sid) ;
+        $event->set('surveyid',$this->sid);
         App()->getPluginManager()->dispatchEvent($event);
         // set the attributes we allow to be fixed
         $allowedAttributes = array( 'template','usecookie', 'allowprev',
             'showxquestions', 'shownoanswer', 'showprogress', 'questionindex',
             'usecaptcha', 'showgroupinfo', 'showqnumcode', 'navigationdelay');
-        foreach ($allowedAttributes as $attribute){
-            if (!is_null($event->get($attribute)))
-            {
+        foreach ($allowedAttributes as $attribute) {
+            if (!is_null($event->get($attribute))) {
                 $this->{$attribute} = $event->get($attribute);
             }
         }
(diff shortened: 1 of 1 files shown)


Shnoulle:
Just a question : maybe add some ->set ?

Minimal : `$event->set('surveyid',$this->sid)`

Shnoulle:
Oh, and move `$this->template=Template::templateNameFilter($this->template);` after the event get.

Because : it's here to FIX : if template didn't exist anymore : set it to default. Then seems it's needed.

TonisOrmisson:
@Shnoulle added, thanks!

Shnoulle:
OK, we have dev metting this afternoon :) i put this in topics of dev metting :) Thanks a lot, a great pull request (IMO)

application/models/Survey.php
```diff
@@ -247,6 +247,19 @@ public function rules()
...
     {
+        $event = new PluginEvent('afterFindSurvey');
+        $event->set('surveyid',$this->sid) ;
+        App()->getPluginManager()->dispatchEvent($event);
+        // set the attributes we allow to be fixed
+        $allowedAttributes = array( 'template','usecookie', 'allowprev',
+            'showxquestions', 'shownoanswer', 'showprogress', 'questionindex',
+            'usecaptcha', 'showgroupinfo', 'showqnumcode', 'navigationdelay');
+        foreach ($allowedAttributes as $attribute){
+            if (!is_null($event->get($attribute)))
```
olleharstedt: Nit picky, but `{` should be on same line as `if`. Also, space between `)` and `{` above.

application/models/Survey.php
```diff
@@ -247,6 +247,19 @@ public function rules()
     */
     public function fixSurveyAttribute($event)
     {
+        $event = new PluginEvent('afterFindSurvey');
+        $event->set('surveyid',$this->sid) ;
```
olleharstedt: No space between `)` and `;`.

olleharstedt:
:+1:

TonisOrmisson:
Thanks to all!

olleharstedt:
@TonisOrmisson Please also add a manual page for this event.

TonisOrmisson:
@olleharstedt Yes, was already looking into it, but have no build number yet. Anyway, will do that.

olleharstedt:
I guess it will be next minor version: 2.63.0.

TonisOrmisson:
does this seem ok:
https://manual.limesurvey.org/AfterFindSurvey

Shnoulle:
Version is OK now, not need buildnumber ? Right ?

olleharstedt:
Yes, version is enough.

olleharstedt:
@TonisOrmisson Great!

TonisOrmisson:
and an initial version of a plugin that enables to set configured templates over url parameter

https://github.com/TonisOrmisson/limesurvey-url-templates

just needs a small additional LS change access Template list via api instead of referring directly to Template:
https://github.com/TonisOrmisson/LimeSurvey/commit/e7be7baec8f1015076ceb5673d864043904b6b17

maziminke:
@TonisOrmisson Did you test the plugin with the latest Limesurvey 3.1x version?

TonisOrmisson:
@maziminke Im not sure. I am not actively using this plugin at this point so I can not comment on the latest releases (I have a plan for it for the future). But I can check what the status is. Feel free to note any issues here
https://github.com/TonisOrmisson/limesurvey-url-templates/issues

maziminke:
Thanks, we will test with the latest version and will fix things as needed and send you a pull request.

TonisOrmisson:
@maziminke my first check shows there seems to be a problem with saving the json data for plugins. Seems its a Lime issue, I'll see where it lies

TonisOrmisson:
Works until 3.15.1
broken since 3.15.2 Some issue with LS saving json type plugin settings. 
I'll find the cause and report it

TonisOrmisson:
this one breaks it
https://github.com/LimeSurvey/LimeSurvey/commit/2f87a0fb3271579a23c7e6a9fcd7027335bb7e5f

PR #635, opened 2017-02-07
Fixed issue #12116: Change input class from input-lg to input

tammoterhark:


tammoterhark:
@lacrioque In discussion with schnoulle we saw that changing this to input-sm may even be better. I leave it to you to decide on this. It is only a visual change.

Shnoulle:
I ask : you test and see :) 

I test now , with a big screen

Actual : really too big ....
actual-bigscreen

Normal : better
normal-bigscreen

sm : OK , why not 
sm-bigscreen

olleharstedt:
:+1:

tammoterhark:
Is there something I need to do?

PR #636, opened 2017-02-08
dev: add getTemplateList to LimeSurveyApi

TonisOrmisson:
Add getTemplateList to LimeSurveyApi to enable getting template list to plugins via API. 

This is to enable working of this plugin:
https://github.com/TonisOrmisson/limesurvey-url-templates
Which is a result of having #634

### DIFF
--- a/application/libraries/PluginManager/LimesurveyApi.php
+++ b/application/libraries/PluginManager/LimesurveyApi.php
@@ -4,6 +4,8 @@ use Yii;
 use User;
 use PluginDynamic;
 use SurveyDynamic;
+use Template;
+
     /**
     * Class exposing a Limesurvey API to plugins.
     * This class is instantiated by the plugin manager,
@@ -168,11 +170,19 @@ use SurveyDynamic;
         /**
          * Get the current request object
          *
-         * @return LSHttpRequest
+         * @return \LSHttpRequest
          */
         public function getRequest()
         {
             return App()->getRequest();
+        }
+
+        /**
+         * Returns an array of all available template names - does a basic check if the template might be valid
+         * @return array
+         */
+        public function getTemplateList(){
+            return Template::getTemplateList();
         }
 
         /**
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -251,10 +251,10 @@ class Survey extends LSActiveRecord
 
         $event = new PluginEvent('afterFindSurvey');
         App()->getPluginManager()->dispatchEvent($event);
+        // set the attributes we allow to be fixed
         $allowedAttributes = array( 'template','usecookie', 'allowprev',
             'showxquestions', 'shownoanswer', 'showprogress', 'questionindex',
             'usecaptcha', 'showgroupinfo', 'showqnumcode', 'navigationdelay');
-        // set the attributes we won't allow to fix
         foreach ($allowedAttributes as $attribute){
             if (!is_null($event->get($attribute)))
             {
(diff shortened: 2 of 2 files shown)


Shnoulle:
Not specific to this issue, i'm always upset about adding simple part to api : did we need to add a lot lot of think in API, or find a better way to expose model ?

TonisOrmisson:
@Shnoulle I did it due to my interpretation of Plugins manual that says

> Plugins should only extend LimeSurvey via its "public" API. This means that directly using classes found in the source code is a bad practice. Though we can't force you not to, you risk having a broken plugin with every minor update we do. 

I have no problem using the Template model directly, but it seemed it was not a preferred way ?

Shnoulle:
Yes, i know : according to manual : we need this in LimeSurveyApi :) . But not my preferred way :)

TonisOrmisson:
ok, thanks :)

LouisGac:
sure TonisOrmisson, not your fault. 

The person who wrote the plugin system thought that little by little all the LimeSurvey application would be rewrote as plugins, pushing LimeSurvey architecture outside of Yii, in an event/hook based architecture (a little bit like in Drupal). 

As you could see, the plugin system works. But we're now trying to use Yii as much as we can, trying to push the LimeSurvey architecture to a MVC modules logic respecting a tree of inheritance (+ behaviors to manage interfaces). So in longer term, LimeSurvey code will be rather organised like that:
https://github.com/humhub/humhub/tree/master/protected/humhub/modules

Basically: we're going to an architecture of the code that is closer to Joomla than Drupal, more respectful of the Yii Framework, more respectful of the modern way of writing web applications. But that's a long term goal. For now, we don't provide any way to add Yii Modules to LimeSurvey, and the plugin system is still the one to be used for everything... 

About the differences between an event/hook architecture and a MVC Modules architecture, you can read this article showing the differences between Drupal and Joomla in the code point of view (not on the enduser/interface pov):
https://www.butterfly.com.au/blog/cms/joomla-vs-drupal-a-technical-comparison-of-the-best-open-source-cms

Shnoulle:
@LouisGac : then the sentence 

> Plugins should only extend LimeSurvey via its "public" API. This means that directly using classes found in the source code is a bad practice. Though we can't force you not to, you risk having a broken plugin with every minor update we do.

Are not true now ? If yes maybe we can update manual (and i revert this commit)

LouisGac:
Plugins should be used only to manage silent events. If now, people are trying write plugins which need changes the "public API" or need to edit the source code: it's because we're trying to do too much with plugins. 

In a MVC module approach, it will be extremely easy for any module to overload any core class/method:
http://www.yiiframework.com/doc/guide/1.1/en/basics.module

Else, if you think about it, after years and years, the whole Yii framework + The whole LimeSurvey classes and methods would finished in the "Plugin public API" (and that's why it has been called "LimeSurvey Public API": it was clearly the goal of it, to become "framework agnostic").

TonisOrmisson:
I do not remember where - I somewhere read that instead of plugins LimeSurvey should use Yii Modules instead. Is this a realistic future approach?

For me personally it would be more logical to do a Yii Module instead of a plugin. Im not sure if this is realistic right now.

LouisGac:
Yes, it's much MORE logical. 

But we have so many things to do, and right now, we prioritize all the developments that makes a difference for end user, in term of functionalities. So, we're not going to develop a LimeSurvey Yii Modules system right now, for LS3. Maybe for LS4... 

But, I guess now you understand better all the matter around the plugin "LimeSurvey public API".

TonisOrmisson:
Thanks for feedback!

maziminke:
Thanks for this great contribution!

PR #638, opened 2017-02-09
New feature #12120: direct action by php-cli

Shnoulle:
- Use index : default action
- param target mandatory
- 2 optionnal parameters function and option

### DIFF
--- a/application/commands/PluginCommand.php
+++ b/application/commands/PluginCommand.php
@@ -1,9 +1,9 @@
 <?php
     /*
     * LimeSurvey (tm)
-    * Copyright (C) 2011 The LimeSurvey Project Team / Carsten Schmitz
+    * Copyright (C) 2011-2017 The LimeSurvey Project Team / Carsten Schmitz
     * All rights reserved.
-    * License: GNU/GPL License v2 or later, see LICENSE.php
+    * License: GNU/GPL License v3 or later, see LICENSE.php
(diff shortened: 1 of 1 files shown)


Shnoulle:
Usage  : cron event don't have target. If we add target now : old cron event don't use it => we break all old plugin using cron event.

Then need another plugin event with mandatory target type. Plugin can use target or not (like newDirectRequest : any plugin can die even if not target).

Add 2 extra parameter function + option : Yii don't accept any number of parameters with -- , user must use $_SERVER['argv'] for all other parameters (or CConsoleCommand->resolveRequest )

Shnoulle:
I merge it monday if there are no issues

application/commands/PluginCommand.php
```diff
@@ -15,13 +15,36 @@ class PluginCommand extends CConsoleCommand
...
         {
             $pm = \Yii::app()->pluginManager;
             $event = new PluginEvent('cron');
             $event->set('interval', $interval);
             $pm->dispatchEvent($event);
         }
+
+        /**
+         * Call direct by command (it's default)
+         * @plugin string : the plugin name, target of action
```
olleharstedt: What's up with the PHPdoc block? Should be
```
@param string $target
```
etc.

Shnoulle: yes, should be :)

PR #639, opened 2017-02-14
New feature: Export printable questionnaires

TonisOrmisson:
opened here: 
https://bugs.limesurvey.org/view.php?id=12130

Basic idea is to enable an easy way of archiving questionnaires in a standardized format that can be read in any computer by anybody.

Surveys are programmed in LimeSurvey usually by people who do not create the questionnaires. The Questionnaires are often created by people who do not manage LimeSuevey. So it would be reasonable to archive the survey in a format that is not related to LimeSurvey nor QueXML. The idea is manage archiving so I would should also work as mass-action. Select a number of surveys, and then generate archive formats (plugin-maybe?).

1. Create an option to export all language printable questionnaires zipped.
2. Find a way to loop this through surveys (plugin maybe?)


application/controllers/admin/export.php
```diff
@@ -27,6 +27,10 @@ function __construct($controller, $id)
         parent::__construct($controller, $id);
 
         Yii::app()->loadHelper('export');
+
+        Yii::import('application.controllers.admin.*');
+        require_once('printablesurvey.php');
```
olleharstedt: Why use require_once instead of Yii system?

application/controllers/admin/export.php
```diff
@@ -1327,6 +1335,79 @@ public function quexml($iSurveyID)
...
+     *
+     * @param integer $iSurveyID
+     */
+    private function _exportPrintableHtmls($iSurveyID){
+        $oSurvey = Survey::model()->findByPk($iSurveyID);
+        $oTemplate = Template::model()->getInstance($oSurvey->template);
+        $assetsDir = Template::getTemplateURL($oSurvey->template);
+        $fullAssetsDir = Template::getTemplatePath($oSurvey->template);
+        $aLanguages = $oSurvey->getAllLanguages();
+        if(!empty($aLanguages)){
```
olleharstedt: What to do if languages is empty?

olleharstedt:
Looks good. Do we need to check permissions?

application/controllers/admin/export.php
Shnoulle: We load ALL admin controllers ?????

Shnoulle: http://www.yiiframework.com/doc/api/1.1/YiiBase#import-detail Yii::import('application.controllers.admin.printablesurvey',1);

Shnoulle:
Question : you really think a lot of user want this ? LimeSurvey GUI doesn't have enough buttons and functions ?

application/controllers/admin/export.php
TonisOrmisson: this works better, yes - thanks. 
I was using an example from here: 
http://www.yiiframework.com/doc/guide/1.1/en/extension.integration

application/controllers/admin/export.php
TonisOrmisson: Isn't there always at least baselanguage?
```
    public function getAllLanguages()
    {
        $sLanguages = self::getAdditionalLanguages();
        $baselang=$this->language;
        array_unshift($sLanguages,$baselang);
        return $sLanguages;
    }
```

application/controllers/admin/export.php
Shnoulle: Using **3rd-Party** Libraries , and there are better solution , and [this](http://www.yiiframework.com/doc/guide/1.1/en/extension.integration#using-namespaced-3rd-party-libraries) work better for 3rd-party

LouisGac:
I agree it's a good functionality, it makes it easy to make an overview of a questionnaire, and it adds only a single button in the export ones.

Thank you for the great work.

TonisOrmisson:
Just an additional comment on why users should need that. 
It's a very typical case that original questionnaires are created between end-client and a research manager type of position in doc format. 

Now after the QN is agreed it goes to some IT related person to programming in LS. Then the Questionnaire often gets slightly changed during programming & testing. 

Now if we need to check what the actual used QN was - the only right place to look for is in LS. But Research manager usually does not even access that. And it rarely makes sense duplicating all the final changes to doc fromats. So I would say the only reasonable solution is to archive the Questionnaire from LS to non LS users as html format, that can be used eg for follow-up survey base QN.

Anyway - this is the problem I need to solve with that.

TonisOrmisson:
... and about making it a loop option. 
I have currently a I would say at least 3 years of projects I would like to archive this way. I have hundreds of LS surveys. And I really do not want to do that manually :)

Additionally. I am thinking of making an archive package containing the html as well as the limesurvey archive as one zipped package. so that you could archive both LS technical stuff and a common readable html in one action.

TonisOrmisson:
The looping most probably isn't for everyone - so I would think pligin direction there.

LouisGac:
just add an export massive action in the survey grid footer

LouisGac:
see:
https://github.com/LimeSurvey/LimeSurvey/tree/master/application/extensions/admin/survey/ListSurveysWidget

massive actions are in the view _selector:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/admin/survey/ListSurveysWidget/views/massive_actions/_selector.php

To understand how it works, see the massive action widget:
https://github.com/LimeSurvey/LimeSurvey/tree/master/application/extensions/admin/grid/MassiveActionsWidget

If you need any help just ask me in the IRC

TonisOrmisson:
@LouisGac Thanks! I'm a bit familiar with that actually. I'm a bit working on switching Quotas view to grid & adding massive actions there.

https://bugs.limesurvey.org/view.php?id=12117

LouisGac:
Just to help you a bit, in: 
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/admin/survey/ListSurveysWidget/views/massive_actions/_selector.php

in the 'aActions'  array, you should add something like:

`
                // Export multiple printable
                array(
                    // li element
                    'type'        => 'action',
                    'action'      => 'export',
                    'url'         =>  App()->createUrl('/admin/export/sa/exportMultiplePrintableSurveys/'),
                    'iconClasses' => 'icon-export',
                    'text'        =>  gT("Printable survey (.html)"),
                    // modal
                    'actionType'  => 'modal',
                    'modalType'   => 'yes-no',
                    'keepopen'    => 'yes',
                    'sModalTitle'   => gT('Export printable survey'),
                    'htmlModalBody' => gT('This will export printable survey.').' '.gT('Continue?'),
                ),`

Then just create the method /admin/export/sa/exportMultiplePrintableSurveys/, knowing that what it renders will be shown in the modal. 

So indeed, when you understood how it works, it's pretty easy to use.

LouisGac:
also, here you'll see the method for exportMultipleStructureSurveys:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/export.php#L910

and its related view:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/admin/survey/ListSurveysWidget/views/massive_actions/_export_archive_results.php

I suggest you to use the same view (or to create a very similar one in /application/extensions/admin/survey/ListSurveysWidget/views/massive_actions/_export_archive_results.php) to keep a coherence in the rendering of the different massive export actions. 

You'll notice that renderPartial use an alias instead of a path to find the view, that's because the views are outside application/views.

application/controllers/admin/export.php
olleharstedt: In that case, why check it? You can use `assert` for mental checks/internal errors, but it's not really used in the LS code-base, might look weird.

Shnoulle:
https://bugs.limesurvey.org/view.php?id=12130#c43015

TonisOrmisson:
@LouisGac Thanks for help!
I posted the code using the bits you contributed

LouisGac:
did you tested it?

TonisOrmisson:
@LouisGac Yes tested in terms of real life clicks. No automated testing. Seems to work for me. Both massive and single exports

maziminke:
Great new feature, I love it.

Once done please add  some documentation at the manual, thanks!

TonisOrmisson:
Please let me know if there is anything I should do on this that I am missing still

olleharstedt:
Feel free to join us on IRC #limesurvey-team Tuesday at 14:30, when we have our team meeting. You can remind people of your PR there, if no one got the chance to merge it yet. Ping us on #limesurvey to get the password to #limesurvey-team.

PR #640, opened 2017-02-15
Fix the Remote Control Timeline function

evently-nl:
Fix the remote control export_timeline function by adding the missing 'Permission::model()->' to the permissions check


LouisGac:
Thank you, that's a good one. 
But please, could you submit it again with a correct commit message?

evently-nl:
Sure! I'll do that tonight. I've made a similar commit on the master branch, was that one correct?

LouisGac:
I just merged it

PR #642, opened 2017-02-23
New feature: API method to get all uploaded files for a given token

brammeleman:
as discussed here: https://www.limesurvey.org/forum/can-i-do-this-with-limesurvey/98457-remotecontrol-api-download-response-to-file-upload-question-type

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2530,6 +2530,48 @@ class remotecontrol_handle
     }
 
 
+    /**
+    * RPC to obtain all uploaded files for a single response
+    *
+    * @access public
+    *
+    * @param string  $sSessionKey  Auth credentials
+    * @param int     $iSurveyID    ID of the Survey
+    * @param int     $sToken       Response token
+    *
+    * @return array On success: array containing all uploads of the specified response
+    *               On failure: array with error information
+    */
+    public function get_uploaded_files($sSessionKey, $iSurveyID, $sToken)
+    {
+        if (!$this->_checkSessionKey($sSessionKey)) return array('status' => 'Invalid session key');
+
+        if (!tableExists('{{survey_' . $iSurveyID . '}}')) return array('status' => 'No Data, survey table does not exist.');
+
+        if(!Permission::model()->hasSurveyPermission($iSurveyID, 'responses', 'read')) return array('status' => 'No permission');
+
+        $oResponse = Response::model($iSurveyID)->findByAttributes(array('token' => $sToken));
+
+        if (!($oResponse instanceof Response)) {
+            return array('status' => 'Could not find response for given token');
+        }
+
+        $uploaded_files = array();
+        foreach ($oResponse->getFiles() as $aFile)
+        {
+            $sFileRealName = Yii::app()->getConfig('uploaddir') . "/surveys/" . $iSurveyID . "/files/" . $aFile['filename'];
+            
+            if (!file_exists($sFileRealName)) return array('status' => 'Could not find uploaded files');
+
+            $uploaded_files[$aFile['filename']] = array(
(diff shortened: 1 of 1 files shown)


Shnoulle:
One file only here ?

If you need all files from token : must be zipped
If you need one file : must restrict to filename ? 

Else : see https://manual.limesurvey.org/Standard_for_Git_commit_messages

brammeleman:
> One file only here ?

Sorry, I don't understand this question.

> If you need all files from token : must be zipped

Should it return base64 encoded zip files? It should be smaller. Most likely, you would need to unzip on the client side anyway for further processing/storing on disk.

> If you need one file : must restrict to filename ?

It returns the uploads referenced by the name of the file on disk. Like:

```
 [uploads] => Array
 (
   [fu_wdmygs38jjueuti] => Array
   (
      [meta] => Array
      (
         [title] => My CV
         [comment] => no comment
         [size] => 469.188
         [name] => CV.pdf
         [filename] => fu_wdmygs38jjueuti
         [ext] => pdf
      )

      [content] => JVBERx0xLjQKJaqrrK0KNCAwIG9iago8PAovVGl0bGUgKHVua25vd24pCi9DcmVhdG9yIChBcGFjaGUg
....
etc, etc
```
I could give it an extra optional parameter to specify the file you'd like to get.

I'm sorry for not paying attention to the commit message guide lines. I don't think I can change the commit message for this commit anymore...

Bests,

Bram

Shnoulle:
Oups, yes OK : return a json with base64 encode file content. No need to zip.

About commit message : http://stackoverflow.com/a/179147/2239406

````
git commit --amend -m "New feature : ....."
git push  brammeleman rpc-get-uploaded-files-feature -f
````

And now , my Point Of View about this feature : can be in a plugin ;)

olleharstedt:
@Shnoulle Why would it be in a plugin?

My view: Every API command should be its own command class, instead of add methods to the existent class indefinitely. But that would take some refactoring.

brammeleman:
Thanks for the git hint! It worked!

> And now , my Point Of View about this feature : can be in a plugin ;)

To my opinion, it should be part of the LS remote API. It does not make sense to be able to retrieve the survey responses, but not the uploads (which are also responses). I know you can construct an url to the file and download it, but only if the uploads directory is in the webservers document root. Which is not an option in our setup (confidential documents are being uploaded).

Shnoulle:
Why in plugin : default ls behaviour when donwload file : send a zip to browser , then  default behaviour from RPC : send a zip :).

@olleharstedt : _Every API command should be its own command class_ +1 and adding API command can be done via extension/plugin

Shnoulle:
And i say : **can** not must 

I fact : i think ALL RPC command must be extension, except the connexion key.

olleharstedt:
Yes, adding a new command should as easy as adding a new class to a command namespace, from plugin or where ever.

brammeleman:
@Shnoulle I get the point. At this point, I'm not too interested in a LS architecture discussion ;-)

Shnoulle:
> If you browse the responses and click an uploaded document, you'll get the document as uploaded. Not wrapped in a zip.

Not for whole reponses (1 response line) if i don't make error :)

Shnoulle:
> @Shnoulle I get the point. At this point, I'm not too interested in a LS architecture discussion ;-)

Then : maybe best is to do a plugin : http://extensions.sondages.pro/development-and-example/extend-remotecontrol-api/

Because next time we update LS architecture, we must update your code if it was integrated in LS

LouisGac:
this PR seems good to me. I don't understand all this debate. We must encourage people to do PR. This PR make the RPC better, bringing a functionality that should be core: RPC should provide a way to get relevant datas for all question types, even for file upload.

c-schmitz:
+1

brammeleman:
thanks!

PR #643, opened 2017-02-23
New feature # Massive actions fror quotas #12117

TonisOrmisson:
I started with asking Massive activation & deactivations of quotas here:
https://bugs.limesurvey.org/view.php?id=12117

Since quotas page did not use gridview it was difficult to do that without basically doing that page from scratch. 

There is a number of changes related to technical issues. 

Mainly: the GridView CheckboxColumn gets wrong selected id-s if any of the grid row content has some additional `<tr>` elements inside that. I had a number of versions of that before I reached where I am.
because of that:
- the Quota-members view must be on the same `<tr>` row as the main quota row and
- the Quota members view itself must not use `<tr>` elements. 
otherwise the CheckboxColumns will not work :(

Due to that the overall horizontal space had to be optimized. 
The Status column is removed. The status is indicated by the color of the Panel containing the quota members. 
Also the Quota action column is removed. The Action is indicated by an icon in members' panel heading: different icons with an explanation in tooltip title.

Additionally there is an additional icon - that has main language Autoload URL in the tooltip. 

All ideas are of course welcome :) 

TODO: since the quota members needed to be done without using `<tr>` elements, it probably needs some css classes to have some padding & structure for visual - but I was not sure where exactly this css should go.


TonisOrmisson:
And an image:
lime-quota-grid

Shnoulle:
I like the screenshot 👏 👏

LouisGac:
seems to be a very good job.

olleharstedt:
Looks good!

TonisOrmisson:
Thanks for the feedback. 

I went some further to also get to change the Urls & texts bulk. 

Since the Forms were duplicated I started from optimizing the initial forms. I used Yii CForm & model components instead of old static forms. This also gives use the validations & errrorSummaries. 

In order to have use of the validations, I kept all submissions in the same controller action (previously separate actions). Skipped the validation (done in model).

Anyway now there is only actual form in one place that can be used for both creating / updating the models. And the LanguageSetting form I used to put into MassiveActions Modal to be able to change all selected quotas at the same time. 

Also. The Languagesetting texts are now in the correct language. 

// TODO:
We need some visual design for the Yii's form errorSummary to be red somehow. Those classes are missing from CSS currently

TonisOrmisson:
The modal view to change bulk LangiageSettings. Note the validation errors.
screenshot from 2017-02-26 17 15 42

And the Quota form with validation result errorSummary
screenshot from 2017-02-26 17 18 56

PR #645, opened 2017-02-28
 Fixed issue #12132: Survey List Status Filter 'Active and running' does not list expiring surveys

TonisOrmisson:
Fixed issue #12132: Survey List Status Filter 'Active and running' does not list expiring surveys


LouisGac:
could you please explain a little bit what was the problem?

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12132

selecting "active And running" in survey list did not display surveys that were active & running, but had an (future) expiry date set

LouisGac:
sorry, I just read the code, I didn't notice the bug report

PR #649, opened 2017-03-02
New feature: NUMBEROFGROUPS replacement field on welcome template (#11716) Updated

mfavetti:
This broke the admin panel because it had no default value. Just added one like totalquestions has.

### DIFF
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -1050,7 +1050,6 @@ function buildsurveysession($surveyid,$preview=false)
     $iNumberofQuestions = Question::model()->getNumberOfQuestions($surveyid);
 
     $_SESSION['survey_'.$surveyid]['totalquestions'] = $totalquestions - (int) reset($iNumberofQuestions);
-    $_SESSION['survey_'.$surveyid]['totalgroups'] = QuestionGroup::model()->getTotalGroupsWithQuestions($surveyid);
 
     // 2. SESSION VARIABLE: totalsteps
     setTotalSteps($surveyid, $thissurvey, $totalquestions);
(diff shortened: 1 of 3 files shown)


mfavetti:
Original: https://github.com/LimeSurvey/LimeSurvey/pull/545

Shnoulle:
Think it's better to do , in template_replace

    $oSurvey=Survey::model()->findByPk($_surveyid);
    if($oSurvey){
      $totalgroups=$oSurvey->->getNumberOfGroups(); /* function to construct */
      /* Can add some other variable here */
    } else {
      $totalgroups="";
    }
    $coreReplacements['NUMBEROFGROUPS']=$totalgroups;

Because : i really think we must remove &$redata = array() and just use $replacements step by step. _Currently : think we must use only $this->varname or $this->areplacemet, but @LouisGac have already work on it in twig version_

LouisGac:
thank you for watching the merge of your commits

mfavetti:
@Shnoulle I'm a bit confused what you want. So currently, the count is pulled from:

QuestionGroup::model()->getTotalGroupsWithQuestions($surveyid) on line 1043 of frontend_helper.php
and is added to the session. (I just duplicated the functionality of the totalquestions replacement.)

The problem is that if that variable is not set in the session, there is an undefined variable notice for line 689 of replacements_helper.php so I added a default zero for totalgroups just like totalquestions has, which fixes the notice.

Instead you would like the totalgroups to be set in templatereplace() of replacements_helper.php (rather than in buildsurveysession() of frontend_helper.php)?

 I added another commit can you see if it's what you meant?

Shnoulle:
PS : i prefer like this : yes : i think

1. remove this awfull &$redata must be done , and each var in &$redata was to be removed one by one.
2. filling session again and again seems a false bad idea on my POV.

prefer this one.

I test and merge

PR #650, opened 2017-03-02
Fixed issue #12156: get_question_properties ordering

mfavetti:
https://bugs.limesurvey.org/view.php?id=12156

Not sure if you guys actually wanted to implement this or not, but here it is.


c-schmitz:
Thank you!!

PR #651, opened 2017-03-03
Dev: fix missed changed variables

TonisOrmisson:


c-schmitz:
Thank you for catching that.

PR #657, opened 2017-03-08
Fixed issue #12183: New quota : autoload_url to Yes by default

TonisOrmisson:
Also, changed the "Status" to "Active"

### DIFF
--- a/application/controllers/admin/quotas.php
+++ b/application/controllers/admin/quotas.php
@@ -460,9 +460,6 @@ class quotas extends Survey_Common_Action
 
         $oQuota = new Quota();
         $oQuota->sid = $oSurvey->primaryKey;
-        $oQuota->active = 1;
-        $oQuota->autoload_url = 1;
-        $oQuota->action = 1;
 
 
         if(isset($_POST['Quota'])) {
--- a/application/models/Quota.php
+++ b/application/models/Quota.php
@@ -33,8 +33,12 @@
 class Quota extends LSActiveRecord
 {
 
+    const ACTION_TERMINATE = 1;
+    const ACTION_TERMINATE_WITH_WARNING = 2;
+
     /* Default attributes */
     public $active=1;
+    public $action = self::ACTION_TERMINATE;
 
     /**
      * Returns the static model of Settings table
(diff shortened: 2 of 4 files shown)


TonisOrmisson:
Ah, and autoload_url=1 is just how I personally use quotas by default :)
I had my hands full of changing the whole grid thing so I missed what it was default before :|

Shnoulle:
Argl, didn't see the autoload forced in newquota ... / 

I didn't think `ACTION_TERMINATE` const was necessary, 

About : set attribute in mdel by default: a good reason is to set it for ALL action (even with remotecontrol and plugin for example).

Are you OK if i close this one and fix it without const ?

Shnoulle:
Just 2 thinks 

1. `const ACTION_TERMINATE_WITH_WARNING` rename to `const ACTION_CONFIRM_TERMINATE` (see [mantis](https://bugs.limesurvey.org/view.php?id=10025))
2. Replace the string on gT by the new one used on [develop](https://github.com/LimeSurvey/LimeSurvey/commit/9c261a6a4b1af8280d8c27dc7334d55b4f6d08fe#diff-be52815e99279e0f049439292653f26cR27) _"Allow user to update his last answers."_ or something like that.

Shnoulle:
Sorry @TonisOrmisson about constant : maybe it's a good idea :) . More clear, but need to be 'for real' ;).

See my last comment

Shnoulle:
Thanks a lot :) I take a look and merge

TonisOrmisson:
Well yes. Let's say I got a bit confused by the resistance towards constants. In my opinion much more readable code. 
eg here: 
https://github.com/LimeSurvey/LimeSurvey/pull/657/files?diff=split#diff-7bd08e34637eda111db9e8d3de69d709L10

Shnoulle:
Yes, you're right, good named constant is really better. I use it (tomorrow) in frontend_helepr where quota was tested : i can remove some comment.

Maybe i fix the sentence, because in master : with 2 : survey is never really terminated (set to Q). I don't breally remind what happen with ACTION_CONFIRM_TERMINATE + autoload_url .... :/

Shnoulle:
Hey : remind : if you're not OK about my idea : you can answer :).

PR #659, opened 2017-03-10
Dev: review stat management screen

Shnoulle:
Dev: remove a potential XSS (echo $oQuestion->question) or broken HTML

Before 
capture du 2017-03-10 15-06-09

After
capture du 2017-03-10 16-06-48

### DIFF
--- a/application/views/admin/quotas/viewquotas_quota_members.php
+++ b/application/views/admin/quotas/viewquotas_quota_members.php
@@ -21,10 +21,10 @@ if ($oQuota->action == Quota::ACTION_TERMINATE){
         <div class="pull-right small">
             <span><span class="fa <?php echo $icon?>"></span> <?php echo $tooltip;?></span>
             <?php if($oQuota->autoload_url == 1):?>
-                <span><span class="fa fa-external-link"></span> <?php echo eT('Autoload URL:').' '.$oQuota->mainLanguagesetting->quotals_url;?>"</span>
+                <span><span class="fa fa-external-link"></span> <?php echo gT('Autoload URL:').' '.htmlentities($oQuota->mainLanguagesetting->quotals_url);?>"</span>
             <?php endif;?>
         </div>
-        <?php echo $oQuota->name;?>
+        <?php echo viewHelper::flatEllipsizeText($oQuota->name) ;?>
     </div>
     <table class="table table-striped table-condensed" >
         <thead>
--- a/application/views/admin/quotas/viewquotas_quota_members.php
+++ b/application/views/admin/quotas/viewquotas_quota_members.php
@@ -26,7 +26,7 @@ if ($oQuota->action == Quota::ACTION_TERMINATE){
         </div>
         <?php echo viewHelper::flatEllipsizeText($oQuota->name) ;?>
     </div>
-    <table class="table table-striped table-condensed" >
+    <table class="table table-quota-items table-striped table-condensed" >
         <thead>
             <tr>
                 <th><?php eT('Question');?></th>
--- a/application/views/admin/quotas/viewquotas_quota_members.php
+++ b/application/views/admin/quotas/viewquotas_quota_members.php
@@ -14,56 +14,45 @@ if ($oQuota->action == Quota::ACTION_TERMINATE){
     $tooltip = gT("Allow user to modify his last answers before terminate survey.");
     $icon = 'fa-exclamation-triangle';
 }
-
-
 ?>
 
 <div class="panel panel-<?php echo ($oQuota->active==1 ? 'primary' : 'default') ?>">
     <div class="panel-heading">
-        <div class="h4" >
-            <?php echo $oQuota->name;?>
+        <div class="pull-right small">
+            <span><span class="fa <?php echo $icon?>"></span> <?php echo $tooltip;?></span>
+            <?php if($oQuota->autoload_url == 1):?>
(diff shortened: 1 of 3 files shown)


Shnoulle:
@TonisOrmisson : what do you think of this new screen ?
Maybe btn-small for the Add answer and Remove (answer) button 

Usage of real table is really better than usage of style ;) (in fact : always use tag and class, never use stymle is the best solution)

Shnoulle:
I think we need bigger completed and limit too ....

TonisOrmisson:
see my original post regarding why I did not use <table> in this case:
https://github.com/LimeSurvey/LimeSurvey/pull/643

is your massive actions submitting correct ids with <tables>??
did not work for me and seems also for others:
https://github.com/yiisoft/yii/issues/3241

Shnoulle:
Oh shit ... / i look at it :)

Shnoulle:
Just tested : after this commit , with decativate/actvate and delete. Seems OK for me.

I test with more answers on quota and more quotas too

Tested on firefox

TonisOrmisson:
I had some serious trouble getting the right ids posted. Any `<tr>` items inside the gridview's `<tr>` did not work correctly for me. I suggest varying different selections.

Shnoulle:
I test again : really no issue for me , didn't update anything else than this commit.

Browser specific issue ?

Shnoulle:
I remove an extra div too : maybe this was the original bug

TonisOrmisson:
didn't seem to be browser related to me. 
this issue:
https://github.com/yiisoft/yii/issues/3241
coming from:
https://github.com/yiisoft/yii/blob/ce3fdd21a90cd85a7ac0cb0610977e5d0d2c0209/framework/zii/widgets/assets/gridview/jquery.yiigridview.js#L392
related to how gridview gets the checked ids. I played around for some time with that. This is why the funny format instead of table.

anyway. of course tables would be better if it works. If you say it works, I can pull your version on test this also to confirm

Shnoulle:
Please : test and confirm : i fix 4 XSS security issue (2 in last commit ,and 2 previously).

Shnoulle:
About bthe Yii issue : think it can be fixed with a good settings.tableClass , but really : can not reproduce here ....

Shnoulle:
jqgrid class : .table-striped.table-condensed.table

I don't add .table-striped class. Just let me add a more specific class 'table-quota' and 'table-quota-items' seems goo, no ?

TonisOrmisson:
got your version and at the beginning it seemed to work fine. 
I was about to say its ok, but then again funny stuff started. Its a weird problem. I'll post a video. and try to see how I can reproduce.

Shnoulle:
Maybe other things then because https://github.com/LimeSurvey/LimeSurvey/pull/659/commits/9c6fe8b97d1a655f8720a9f7688cebcddff7e83f surely fix the issue with https://github.com/yiisoft/yii/blob/ce3fdd21a90cd85a7ac0cb0610977e5d0d2c0209/framework/zii/widgets/assets/gridview/jquery.yiigridview.js#L392 :)

TonisOrmisson:
Sent you a short video on email

Shnoulle:
Please, i merge because of the XSS, if i can fix the issue you have i revert.

TonisOrmisson:
I would suggest to fix the XSS on the current version.

Shnoulle:
Too late, already released

TonisOrmisson:
posted the video here also
https://youtu.be/CO637fYgIfc

Shnoulle:
Trying to reproduce currently ... exactly same situation than your : 4 quota without answers 
????

TonisOrmisson:
Its a weird thing. at first looks like its working. Ill try to get it on video from beginning.

TonisOrmisson:
a bit longer video. 
first actions look ok, then errors. 
https://youtu.be/EswfAM1L7ww

Shnoulle:
Not needed a video, just a clean way of reproduce , i put some console.log currently in listActions.js , and i alrady try to activate/deactivate, check item, open a new tab, look at validation etc .....

TonisOrmisson:
I can not say clean way of reproducing. 
As I said originally - removing any nested table elements from inside the grid row makes the thing work flawlessly for me. Adding any table elements made this error re-occur.

Shnoulle:
Yes : but looking at code  : i don't see how this can happen ...

Shnoulle:
If i find any thing i fix it monday morning (and test with different browser too).
Can you take time on paginations ?

TonisOrmisson:
I can look at paginations yes. 
are you able to reproduce at least?

Shnoulle:
Yep : in fact check 1/2/3 -> console.log here https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/admin/grid/MassiveActionsWidget/assets/listActions.js#L26 receive the 3 first id : seems OK. But after come back : 4 items are updated.

Thanks for reproduce, no, i try to understand what happen after ;)

Shnoulle:
OK,

Can not reproduce now, and understand what happen : the bug surely can happen elsewhere : https://bugs.limesurvey.org/view.php?id=12193

Shnoulle:
Fixed in https://github.com/LimeSurvey/LimeSurvey/commit/4ae6395e9701d05649f609c826cedf62d1b36958

PR #660, opened 2017-03-10
Fixed issue #12192: Quotas page missing pagination - only 10 quotas v…

TonisOrmisson:
…isible


TonisOrmisson:
maybe I should add the pagination size selection also like survey grid?

LouisGac:
thank you for that fix. Yes, you could add the pagination size selector, would be nice.

PR #661, opened 2017-03-14
Fixed issue #12201: Quotas not triggered for secondary languages

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12201


TonisOrmisson:
While doing the change here
https://github.com/LimeSurvey/LimeSurvey/commit/59f51956825ee3ba1b8a6dbd5baa5a89b6d851fa#diff-a9dbc6157f3066d12e04af89a4bca062R4368

I did not realize that here
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/common_helper.php#L4355
... the Quota was linked to one language and the rest of the languages would not be loaded to 
`$oQuota->languagesettings`

I would say it is much more logical and less confusing to get the `$oQuota->languagesettings` directly from the model relations. Instead of making a custom query and confusing the idea of HAS_MANY languagesettings as described in model.

Shnoulle:
In fact, we don't need quota language setting to test it, we need only quota.

If we have 2 table/model
- Quota
- QuotaLanguage

It's better, but we don't have it, and fix it for Quota/Question/QuestionGroup .... is an hard task

TonisOrmisson:
I would take the `$oQuota->mainLanguagesetting` instead and rewrite this
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Quota.php#L157
without loop (which is reasonable anyway)

this should make it same as old one but more logical. 
ok?

Shnoulle:
Take default language or actual language : but take only one language please (see mantis).

If i don't make error : with 10 quotas and 10 languages : you make 100 loop where you need only 10 currently.

Shnoulle:
And why not $oQuotas restricted by language and survey ?

TonisOrmisson:
> Take default language or actual language : but take only one language please (see mantis).

Yes, my point exactly. We have this in model already only we need to kill the loop :)

> And why not $oQuotas restricted by language and survey ?

Because it is confusing and may cause further errors. This specific error is related to exactly me understanding that the Quota has `HAS_MANY` relations with Languagesettings according to the definition in Quota model relations. I think it is not very wise to mess with the model definitions. It would be ok if it would not be called an `$oQuota`. if it would be `$oQuotaWithMainLanguageOnly` or something .. then it would make pay attention that it is not standard model defined object.

application/helpers/common_helper.php
```diff
@@ -4362,7 +4362,7 @@ function getQuotaInformation($surveyid,$language,$iQuotaID=null)
         foreach ($oSurvey->quotas as $oQuota)
```
Shnoulle: <s>You still have quota X language foreach here , no ?</s>
Sorry : it's OK

application/models/Quota.php
```diff
@@ -83,7 +83,6 @@ public function primaryKey()
      */
     public function relations()
     {
-        $alias = $this->getTableAlias();
```
Shnoulle: 👍

Shnoulle:
Still an issue 

Testing survey 
[survey_archive_checkQuota.lsa.zip](https://github.com/LimeSurvey/LimeSurvey/files/844407/survey_archive_checkQuota.lsa.zip)
capture du 2017-03-15 13-14-43

* [fr (base lang)](git pull https://github.com/TonisOrmisson/LimeSurvey.git fix12201) 
* [en (added after set quota)](https://limesurvey.sondages.pro/index.php/67142?newtest=Y&lang=en#)

I look further

TonisOrmisson:
Thanks! 
I started to look into it, it seemed to work for me originally.

Shnoulle:
Yep, sorry. Their still an issue : seem it's not the good text returned : i look at it and confirm it after.

PR #662, opened 2017-03-14
DEV: add pagination size selector to quotas grid

TonisOrmisson:


application/controllers/admin/quotas.php
```diff
@@ -161,13 +161,20 @@ function index($iSurveyId, $quickreport = false)
         $totalcompleted = 0;
         $csvoutput = array();
 
+        // Set number of page
+        if (isset($_GET['pageSize']))
```
Shnoulle: I prefer `Yii::app()->getRequest()->getQuery('pageSize',20);` for example

TonisOrmisson:
used the surveyGrid as example:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/admin/survey/ListSurveysWidget/ListSurveysWidget.php#L49

Shnoulle:
Yes, and this part is buggy too : why pageSize set on Survey must set for whole list

Shnoulle:
@LouisGac about $_GET['pageSIze'] and `Yii::app()->user->setState('pageSize',(int)$_GET['pageSize']);`

1. Usage $_GET seems a bad idea for me
2. pageSize is set for all list ?

TonisOrmisson:
thanks. 
Default is set here:
https://github.com/LimeSurvey/LimeSurvey/pull/662/files#diff-83d39c85b95471b32a718b8be123acf3R169
coming from app params

Shnoulle:
Thanks :)

LouisGac:
great work again.

PR #664, opened 2017-03-15
Fixed issue : Quota end message use base language only

Shnoulle:


Shnoulle:
@TonisOrmisson : i can not set you as reviewer, but bad language for end page of quota

capture du 2017-03-15 14-08-56

[survey_archive_quotaBadLang.lsa.zip](https://github.com/LimeSurvey/LimeSurvey/files/844558/survey_archive_quotaBadLang.lsa.zip)

Can you have a look ?

Shnoulle:
Alternative : `public function getLanguagesetting($language=null)` in model. 

Because : common_helper have $language in the function (and not used currently)

TonisOrmisson:
a typo here:
https://github.com/LimeSurvey/LimeSurvey/pull/664/files#diff-a9dbc6157f3066d12e04af89a4bca062R4362
current not curent

TonisOrmisson:
I would prefer getCurrentLanguagesetting. 
its a convenient way of accessing properties like `$quota->currentLanguagesetting` instead of `$quota->getCurrentLanguagesetting()`. If you have an input parameter additionally for that, it gets more complex. For me at least.

Shnoulle:
Yes, it's what i see : must deprecated $language at same time

Shnoulle:
Thanks for the typo

TonisOrmisson:
ah .. and case sensitive also if accessd like that
https://github.com/LimeSurvey/LimeSurvey/pull/664/files/e0519866896ce4fc9e7a4f470f80d3e3d464ab1a..2b36d5634b0e2a47013feef6841cfab36f72fb0f#diff-a9dbc6157f3066d12e04af89a4bca062R4362
Should be ->currentLanguagesetting

Shnoulle:
It work for me currently : https://limesurvey.sondages.pro/index.php/67142?newtest=Y&lang=en ? You are sure ?

TonisOrmisson:
... and the 
` * @property QuotaLanguageSetting $currentLanguagesetting`
to class docs

TonisOrmisson:
seems its not case sensitive
http://www.yiiframework.com/wiki/167/understanding-virtual-attributes-and-get-set-methods/
Have not done it myself, I have a habbit of doing camelCase for such cases usually. makes it easier to read and also IDE to warn less on typos

Shnoulle:
@TonisOrmisson i didn't find the good way for unused/deprecated param ... maybe @ void ? Don't find  in phpdocumentor

TonisOrmisson:
Have not used myself, can't suggest  :(

Shnoulle:
`@param null : $deprecated not used` :)

PR #666, opened 2017-03-17
Update cpd_importParticipants in remotecontrol_handle.php

tostern:
Add language key to $aData array in cpd_importParticipants.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2699,6 +2699,7 @@ class remotecontrol_handle
                 'lastname' => $aParticipantData['lastname'],
                 'email' => $aParticipantData['email'],
                 'owner_uid' => Yii::app()->session['loginID'], // ToDo is this working?
+                'language' => $aParticipantData['language'], 
             );
 
             //Check for duplicate participants


application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -2699,6 +2699,7 @@ public function cpd_importParticipants($sSessionKey, $aParticipants)
                 'lastname' => $aParticipantData['lastname'],
                 'email' => $aParticipantData['email'],
                 'owner_uid' => Yii::app()->session['loginID'], // ToDo is this working?
+                'language' => $aParticipantData['language'], 
             );
```
KaiDream: Should participant_id be a field that is allowable so that added Survey {participant_id}s map to the cpd?

LouisGac:
I'd say that for retro compatibility  "language" should be optional, eg something like:
`$aData['language'] = (!empty($aParticipantData['language']))?$aParticipantData['language']:'';`

tostern:
Yep, the language should be optional. Without that string i'm not able to set the language for the cpdb, because the default language is hardcoded with 'en'.

c-schmitz:
@tostern Can you update the patch please? Otherwise we have to close it due to lack of feedback.

c-schmitz:
Seems to abandoned. Closing.

PR #670, opened 2017-03-22
DEV: clean survey model VOL1

TonisOrmisson:
Started to clean up the Survey model a bit. 
I'll push this here and work take a different angle meanwhile.
Found some broken things again I did not understand, see //FIXME tags


application/config/internal.php
```diff
@@ -118,7 +118,7 @@
         // These are defaults and are later overwritten in LSYii_Application by a path based on config tempdir/tempurl
         'assetManager' => array(
             'excludeFiles' => array("config.xml", "assessment.pstpl", "clearall.pstpl",  "completed.pstpl",  "endgroup.pstpl",  "endpage.pstpl",  "groupdescription.pstpl",  "load.pstpl",  "navigator.pstpl",  "printanswers.pstpl",  "print_group.pstpl",  "print_question.pstpl",  "print_survey.pstpl",  "privacy.pstpl",  "question.pstpl",  "register.pstpl",  "save.pstpl",  "startgroup.pstpl",  "startpage.pstpl",  "surveylist.pstpl",  "survey.pstpl",  "welcome.pstpl" ),
-            'baseUrl' => '/tmp/assets',
+            'baseUrl' => '/lime-dev/tmp/assets',
```
Shnoulle: asseturl is resetted in App , if update (good idea , because this one is false) set it to NULL ?

Shnoulle:
Why move frim survey id tro oSurvey in createFieldMap ?

Think it's best to use an int parameters, no ?

TonisOrmisson:
I would use objects instead of IDs. Why?

1. In most cases we already the objects created before calling a method. So we will be re-calling the object unnecessarily. 
2. Input is very clearly specified. Int could be ID of anything. If wrong entity id goes in - no possibly no errors produced. Using objects make debugging more easy.
3. Passing objects in many cases free us from passing some other object related parameters. Eg there is a lot of stuff passed to views related to survey objects that are all passed as separate items. Instead of just passing the object and accessing the object properties directly from view.  
4. It will simplify code. In the long run - A LOT of unnecessary code rows  can be deleted.

at least that's how it looks to me :)

Shnoulle:
Personnaly : i prefer to use objectId.

Here : i like to have it in 3.0, not in 2.XX because some plugin can use `createFieldMap` even if not in LS api.

application/config/internal.php
Shnoulle: thanks :)

TonisOrmisson: Sorry... this was not intended to push here, I reverted that.

Im not sure whether its an issue, but the default settings did not find my asset-path in a lime Sub-folder. Did not want to deal with that so I just manualluy adjusted the path :)

Shnoulle: Resetted here : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSYii_Application.php#L90 in 2.XX

Maybe i lost some part from internal (set asseturl to null here : https://github.com/LimeSurvey/LimeSurvey/commit/1878a118eccee076e9b986ae871a1cafc1973c42#diff-f5392477bcb7ace25c65e8bd7abd39e9R67)

Shnoulle: Fixed in https://github.com/LimeSurvey/LimeSurvey/commit/71685ac145af96adfbe85d9eb22d36488f80516e , thanks for report

lacrioque:
+1 for moving to object based params.
Far easier to debug and a lot faster on the long run, due to less database connections.

Shnoulle:
@TonisOrmisson think others merge fix this too ? Maybe we can close thgis one ? (Sorry for bad ordering of merging :( )

TonisOrmisson:
Yes. I Close this. Too much conflicts to resolve and I am not sure I can adequately do this at this point.

PR #672, opened 2017-03-24
Fixed issue #12226: [security] Improper Handling of Insufficient Privileges

mfavetti:
https://bugs.limesurvey.org/view.php?id=12226


c-schmitz:
Great - thank you!

PR #674, opened 2017-03-28
Dev: add sgqa property to Question

TonisOrmisson:
... to reduce code duplication in the future

### DIFF
--- a/application/models/Condition.php
+++ b/application/models/Condition.php
@@ -111,11 +111,13 @@ class Condition extends LSActiveRecord
     */
     public function updateCFieldName($iSurveyID, $iQuestionID, $iOldGroupID, $iNewGroupID)
     {
+        /** @var Condition[] $oResults */
         $oResults=$this->findAllByAttributes(array('cqid'=>$iQuestionID));
         foreach ($oResults as $oRow) {
             $cfnregs='';
-            if (preg_match('/(\S*?)'.$iSurveyID."X".$iOldGroupID."X".$iQuestionID."(.*)/", $oRow->cfieldname, $cfnregs) > 0) {
-                $sNewCfn=$cfnregs[1].$iSurveyID."X".$iNewGroupID."X".$iQuestionID.$cfnregs[2];
+            $sgqa = $oRow->questions->SGQA;
+            if (preg_match('/(\S*?)'.$sgqa."(.*)/", $oRow->cfieldname, $cfnregs) > 0) {
+                $sNewCfn=$cfnregs[1].$sgqa.$cfnregs[2];
                 Yii::app()->db->createCommand()
                     ->update($this->tableName(), array('cfieldname' => $sNewCfn),
                     'cid=:cid',array(':cid'=>$oRow->cid));
--- a/application/models/Condition.php
+++ b/application/models/Condition.php
@@ -111,13 +111,11 @@ class Condition extends LSActiveRecord
     */
     public function updateCFieldName($iSurveyID, $iQuestionID, $iOldGroupID, $iNewGroupID)
     {
-        /** @var Condition[] $oResults */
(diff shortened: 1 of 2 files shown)


Shnoulle:
1.

It's 
`$this->survey->primaryKey.'X'.$this->groups->primaryKey.'X'.$this->primaryKey;`

And :
If parent_is is set : it can be 
`$this->survey->primaryKey.'X'.$this->groups->primaryKey.'X'.$this->parent_id.$this->title;`

or array() for array/text array/number and array/multiple

Must control  : ranking, : think it's array too.

Shnoulle:
But i like the idea ....

Possible fix : 

1. Return sub question SGQA if available
2. Return array of sgqa ( see other and comments too ...)

TonisOrmisson:
Wouldn't it be simpler to keep like that and if has parent get it rather as `$oQuestion->parents->SQQA`?
more logical no?

Shnoulle:
No : i have a $qid , i use an function : i just want to have the SGQAs, not to have do test myself ..... else : i just use my own function .....

And more : if we have QuestionObject : yhis question object set SGQA : see ranking for example.

I think we must have a function to get ALL the SGQAd of a specific QID. And sometimes this SGQA didn't exist ( show text question type for example).

Shnoulle:
Here : if i ask a sub question qid , get the sgqa : i have a false sgqa, i can not use it .... why i use this function then ?

TonisOrmisson:
Can you refer code where you actually would need to get it for sub-question? I would think it is rather unused case. 

>I think we must have a function to get ALL the SGQAd of a specific QID. And sometimes this SGQA didn't exist ( show text question type for example).

Well. Actually even show text has an SQQA. I am assuming it does not get used that much but it clearly has a SID, GID & QID - so it also has an SGQA. 

I generally do not like things that get more complicated than they have to. I am not really convinced right now that getting a parent SGQA for a subquestion would be logical. How would you then get a sub-question's own SGQA then? Think about the cases where big arrays get split up in views on small screens. They probably work as separate questions rather than sub-questions in some places. 

I really have problems understanding why it can not be that simple. Any examples from current code?

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/commit/b3ef63e7f749dcf0e2999a987b72c40db09f9a31#diff-5f262794cfea9770eb1146a559f9eaa2R92

ANd in fact : all place where we use $field['fieldname'] currently

Shnoulle:
And see this one too : https://github.com/LimeSurvey/LimeSurvey/commit/b3ef63e7f749dcf0e2999a987b72c40db09f9a31#diff-5f262794cfea9770eb1146a559f9eaa2R97

Currently we always use fieldmap for such situation : SQA come from fieldmap then from [createFieldMap](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/common_helper.php#L1780) function

TonisOrmisson:
like that:
```
$sFieldName=$this->question->SGQA.$this->code;
$sFieldName=$this->question->SGQA.$temp[0];
```

TonisOrmisson:
I also was looking at the createFieldMap. 
It seems reasonable to me to look for a way to use question objects some way in createFieldMap also. And there we could access it similarly.
Btw I have a feeling FieldMap could be a separate model/class/object.

Shnoulle:
About createFieldMap function : currently we use a lot (more than needed) array returned by createFieldMap.

Then have a way to get real question information directly in Question::model seems a really great idea.

But then : get the real information , not false or badly set information.

`$sFieldName=$this->question->SGQA.$this->code;`

Yes : but still : need to do test .... i really don't want to have a function to just return something unusuable.

It's my Point Of View, but with this function we don't really 

> reduce code duplication

Because : we must test always what question type is the question before using this function.

PR #675, opened 2017-03-28
Yii usage about CRSF

Shnoulle:
Yii already have a good CRSF system : we just have to control if it's a POST request

An,d jquery ajax already have CRSF token adden in whole ajax request : no need to make a new hack again and again.

Question : what is the usage of _getPostOrParam : i really think we must just use getPost (if we need POST) or getParam if we allow GET or POST

### DIFF
--- a/application/controllers/admin/useraction.php
+++ b/application/controllers/admin/useraction.php
@@ -41,6 +41,9 @@ class UserAction extends Survey_Common_Action
         $value = Yii::app()->request->getPost($param);
         if(!$value)
         {
+            /* This already return GET or POST : : http://www.yiiframework.com/doc/api/1.1/CHttpRequest#getParam-detail
+             * DB update need $_POST, then only Yii::app()->request->getPost or control Yii::app()->request->getIsPostRequest()
+             **/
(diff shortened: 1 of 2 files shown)


Shnoulle:
Child of https://github.com/LimeSurvey/LimeSurvey/commit/21955942ad58dfb3907af76033070bd1ca3cf53b

application/controllers/admin/useraction.php
```diff
@@ -177,93 +180,91 @@ public function adduser()
     */
     public function deluser()
     {
+        if(Yii::app()->request->getIsPostRequest()){ /* DB action : need post request */
```
Shnoulle: I just add this test : all other red/green line is about space

LouisGac:
you can (must ^^) use the Yii Crsf in Ajax context

Shnoulle:
1. Yii::app()->request->getIsPostRequest() force to be $_POST
2. We already use CRSF if we use ajax/post : no need extra solution here, we do it in 2.00 : but need to be a POST

Shnoulle:
@lacrioque can we merge this one ?

PR #676, opened 2017-03-29
LDAP auth group filtering: support for $userdn

epol:
In the LDAP authentication module the group filtering can use the $userdn variable that contains the user DN.
This feature is useful in a Active Directory environment where the groups contain the user DNs as members.

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -108,7 +108,7 @@ class AuthLDAP extends ls\pluginmanager\AuthPluginBase
         'groupsearchfilter' => array(
                 'type' => 'string',
                 'label' => 'Optional filter for group restriction',
-                'help' => 'Required if group search base set. E.g. (&(cn=limesurvey)(memberUid=$username))'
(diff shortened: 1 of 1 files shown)


Shnoulle:
And next time ? We add more and more var ?

Personally : i don't want to _improve_ LDAPAuth like ths because

1. It's not a global system _Allow any filter_
2. LDAPAuth can be easily cloned and update for each situation purpose.

epol:
Hi Denis and thank you for your reply.

**tl;dr:** I think that many users are using Active Directory as the authentication and authorization method, so merging this edit would make their life easier configuring LimeSurvey.

**Long answer**
I agree with you that adding a lot of vars to cover every possible use would be clumsy and potentially harmful on the performance point of view (you may have to do more LDAP queries).
But, on the other hand, I think that my extension covers enough use cases to deserve some thinking. Indeed I think that there are a lot of users using Active Directory as their LDAP database and they may want to use groups (where members are pointed using the DN); adding the _$userdn_ attribute doesn't add any additional LDAP query (the variable is already available from a previous query), so the effect on performances should be negligible.

PR #678, opened 2017-04-03
Logging management, usage of Yii

Shnoulle:
- log warning and error to web without broke totally page with 0<debug<2
- allow to use tracevar without debug
- unsure for CFileLogRoute by default : can be interesting, but runtime path have web direct access issue

Maybe add a way to show php notice at the same way with  1<debug<2 with [ErrorHandler](http://www.yiiframework.com/doc/api/1.1/CErrorHandler) ?


Shnoulle:
Because you seem interested, maybe another way ? We can use PHP functin, but then replace the handler.

olleharstedt:
I'm not sure... Shouldn't this be up to the dev to configure?

Shnoulle:
It's already the case today, but we don't offer an easiest way . We already use log/trace for tracevar : we make a choice for the user :). 

We can examplain on forum how to update config.php to set debug=1, maybe add a system to log it when user make manually the choice (something in session for this user etc ....) . But exaplain how to put this in config.php it's more hard (on forum).

Here : the user can update it in config.php, but with a clean example . Maybe enable limesurveyErrorAndWarning (by default) only if debug >0 too ? The config.php file replace internal.php config part.

LouisGac:
I think we should merge it.

Shnoulle:
Maybe in develop only but still 

> but runtime path have web direct access issue

I'm really not sure here , we can disable direct access with .htacess , but only for apache then ?

Shnoulle:
http://www.yiiframework.com/extension/syslogroute/ and set it by default seems the best solution. no ?

c-schmitz:
@Shnoulle Can you update it so we can merge it to dev?

Shnoulle:
Oups, don't see the update. But don't log it `CFileLogRoute` else a lot of user can show error log publicly …

maziminke:
Denis, does this support logging any warnings or errors to a log file at e.g. /tmp?

Shnoulle:
See https://manual.limesurvey.org/Optional_settings#Logging_settings 

And if you wait 5 minutes i update it to don't log 404 error (there are a lot …)

Shnoulle:
The commit is here : https://github.com/LimeSurvey/LimeSurvey/commit/e08fbc018f89bf55173fecd5b86bc56686d42c39

But, manual explain it : https://manual.limesurvey.org/Optional_settings#Logging_settings

maziminke:
Thanks, Denis. The documentation at the manual is a little vague: Does the sample code log to a file and if so, where is it found?

Shnoulle:
https://manual.limesurvey.org/Optional_settings#Introduction + https://manual.limesurvey.org/Optional_settings#Yii_settings … i really think it can be more clear …

PR #679, opened 2017-04-05
Fixed issue #12237: Tips are missing in "General options".

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12237
Also fixed some missing closing `</p>` tags in the bottom.


Shnoulle:
No ;) .

If you have a 2.06/2.6lts : tips are different ;). And more : i don't think we need tooltip like actually but maybe boostrap-help :)

TonisOrmisson:
So it's not actually an issue that should be fixed?
The tips are done the same way as the tips that are already there in section below. Doesn't seem reasonable to implement help & tips in the same section. 

Texts - didn't check 2.0x. Used some texts that were already in translations. 
In general I also don't think the tips are essential there. I was just took a simple issue to fix to end my day :)

LouisGac:
I merge it for the unclosed </p>

PR #680, opened 2017-04-10
DEV: Add 'files' type to Plugin Settings

TonisOrmisson:
I am working on a plugin to allow admins to upload some survey related files to server (eg image files used in questionnaire).

plugin: https://github.com/TonisOrmisson/limesurvey-admin-file-upload


Shnoulle:
I think there are a feature request (since the start of all this) , else, some remark.

1. `$htmlOptions['multiple'] = 'multiple'` must be default but can be updatable in setting [*]
2. basePath and folder must be set by deafult and updatable by setting. Ddidn't know where for global settings, else for survey default : upload/survey/XXXX/ no ?
3. File type restriction ? Updatable by setting ?


for 1 : `$htmlOptions = $this->htmlOptions($metaData,$form,array('multiple' => 'multiple'));` do the trick.

TonisOrmisson:
The path & filetype, max uploads I intended to have in the plugin. 
If all of this is added to the core there will not be much to do with the plugin :)

TonisOrmisson:
this?
https://bugs.limesurvey.org/view.php?id=8113

Shnoulle:
Not this one : i search but don't found :).

About  filetype, max uploads : but we need a easy way to set it in 

```
protected $settings = array(
  'upload'=>array(
    'htmlOptions'=array(),
    'max'=>3,
    'accept'=>'csv'
    'options'=>$options
  ),
);
```

Think have path (where is saved the file at end) by default is cool too. If plugin didn't set it : where is saved ?

Shnoulle:
Thanks for the last one. Remind : i'm not the one who make the final decision here.

Whit filetype=>null : you allow all types ?

TonisOrmisson:
the plugin is not in working condition (yet)

TonisOrmisson:
I also wanted to make the PluginSettingData as Model, but cleaned it out of this one. 
In the light of that I would put all custom (not related to other types) options under 'options'

Shnoulle:
options seems used by js http://www.yiiframework.com/doc/api/1.1/CMultiFileUpload#options-detail

In another : i think we use pluginOptions or something like that.

Shnoulle:
We lust have a [accept null](https://github.com/yiisoft/yii/blob/1.1.17/framework/web/widgets/CMultiFileUpload.php#L125) to allow "any files type' or we force restriction ?
Maybe default can be allowed template upload ?

TonisOrmisson:
options as in:
```
class PluginSettingData extends CModel
{
    public $type;
    public $label;
    public $help;
    public $default;
    public $content;
    public $options;
    public $items;
```

TonisOrmisson:
I do not have strong opinion about the `->accept` setting. Since we are talking plugin should be without big restrictions & customizeable. It could also be audio/video to be embedded. 
Or it could be any type of file for participant to download. So I'm not sure what the default should be.

Shnoulle:
php file allowed , why not, twig surely for after ;) and scss too.
And remind : it's always the default, plugin can allow .mpg if needed ;)

About PluginSettingData::model() : a totally new concept for settings ?

TonisOrmisson:
Please have a look at the default file types. Should we add something else?

PluginSettingData::model() -> Will discuss that in separate PR if I manage to make it work logically. I have a general preference towards objects (that can be properly documented) vs arrays (undocumented, ambiguous) :)

TonisOrmisson:
About the path:

The SettingsWidget in general is about views, UI. It's not about controller & input handling. So With files - we actually will not be saving any files via the PluginSetting. Only thing we can save is the fileNames- and even that probably is useless since in real life we should be reading the path files if we want so see what files we have there. So in terms of path - the only reasonable thing we can add is a simple defaultpath - something that we can use as default value  in plugin. Seem logical?

Now setting default value for the path would not be locigal in a widget (about UI). In general what I would do is define all the input types as a separate model. Then the PluginSettingData model with the type 'files' could have the defaultPath option. that we can use both in plugIn as well ase the PluginSetting widget. 

If this seems reasonable then I would close this for now and do a separate task on how to make it into an object reasonably.

c-schmitz:
What is 
https://github.com/TonisOrmisson/limesurvey-admin-file-upload 
for ?

You can already upload and download files in a survey using the file manager, so I am not sure what you are trying to achieve here?

TonisOrmisson:
@c-schmitz Thanks!
Honestly - I had no idea that we had the file manager :) 
I have never really used the WYSIWYG editor except for getting insertans. 
As far as I am concerned - you can close this

Shnoulle:
Because : with a specific file upload : you can have easiest way to link a file with a plugin part.

See exemple with sliderEmote and 'img link'  : 
1. [image name](https://framagit.org/SondagePro-LimeSurvey-plugin/sliderEmote/blob/master/sliderEmote.php#L244) are fixed by default (OK if uploaded by ressources)
2. must do a [function to get the url](https://framagit.org/SondagePro-LimeSurvey-plugin/sliderEmote/blob/master/sliderEmote.php#L244) and allow use another image
3. must [try to explain](https://framagit.org/SondagePro-LimeSurvey-plugin/sliderEmote/blob/master/sliderEmote.php#L178) the how to use

With such plugin settings : replaced directly by one clean solution

TonisOrmisson:
In terms of what I intended to accomplish - the current file manager basically does what I meant. I understand there might be cases where it could make life a bit easier, but I can live easily without it.

PR #681, opened 2017-04-11
DEV: Add massive set-expiry to survey-list

TonisOrmisson:
also
massive_actions/_delete_results refactored to generic
massive_actions/_action_results

### DIFF
--- a/application/controllers/admin/questions.php
+++ b/application/controllers/admin/questions.php
@@ -1423,7 +1423,7 @@ class questions extends Survey_Common_Action
             }
         }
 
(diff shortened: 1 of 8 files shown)


TonisOrmisson:
For some reason the datepicker dropdown works only first time after refreshing the page.

LouisGac:
about grid refresh and date picker, have a look to what I did for similar problem for switch buttons:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/admin/grid/MassiveActionsWidget/assets/listActions.js#L209

You can even add the needed JS to that file.

LouisGac:
thank you for the great work.

PR #683, opened 2017-04-19
Issue12270

mfavetti:
https://bugs.limesurvey.org/view.php?id=12270

### DIFF
--- a/application/helpers/qanda_helper.php
+++ b/application/helpers/qanda_helper.php
@@ -3200,6 +3200,7 @@ function do_numerical($ia)
         $fValue = rtrim(rtrim($fValue,"0"),".");
     }
     $fValue = str_replace('.',$sSeparator,$fValue);
+    $fValue = htmlspecialchars($fValue);
 
     if ($thissurvey['nokeyboard']=='Y')
     {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Think yes , all can be `htmlspecialchars` but : maybe we must do it in view only and not in qanda (for develop , @LouisGac )

I see you add it too to do_array_texts : are you sure we don't do it again after ?

c-schmitz:
I agree that this should rather be done in the view using CHtml functions.

Shnoulle:
@c-schmitz : yes, but only for develop here, right ?

And : if i remind : we can not use directly because qanda already `htmlentities` some value.

mfavetti:
Can you link me where you see \CHtml in the view for array texts? I don't see it. I'm happy to change the fixes to be in the view, but I need an example.

By the way, the issue is mostly with the values being filled in the hidden fields and not the visible question. These hidden fields are filled directly from the session data. So either we have to sanitize before display in each view, or it should be sanitized before its ever stored in the session and then it could be trusted. Right now, it's treated as trusted even though it's never sanitized.

Shnoulle:
About array text : i think, because we already fix single text , we already fix it after.

About `\CHtml` : it's more a Dev mantis , I didn't think we must fix it now in views before develop/3.0

LouisGac:
the fix doesn't need to be in the view, it's ok like that for me.

Shnoulle:
> the fix doesn't need to be in the view, it's ok like that for me.

I think the issue is 'when display', then we must 'fix it' only when displaying ? No ?

But : right, currently : i think this pull request must be merged, because it must be reworked in develop :)

PR #685, opened 2017-04-25
fontawesome now being used correctly to show the asterisk

brainadam24:
also corrected the grammar in the comment above it

### DIFF
--- a/application/views/survey/question_help/asterisk.php
+++ b/application/views/survey/question_help/asterisk.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * This file render the asterisk for mandatory questions
+ * This file renders the asterisk for mandatory questions
  */
 ?>
-<span class="text-danger asterisk"></span>
+<span class="text-danger fa fa-asterisk"></span>


LouisGac:
We need the asterisk class for compatibility between font system. We don't want to be bound to font awesome. A third party developer may want to use another icon system.

tammoterhark:
@LouisGac: If so, then another developer can adapt the templates? Isn't making that possible that what you are doing right now?

brainadam24:
I just downloaded and installed LimeSurvey and it didn't show any icon, so i searched and found FA in the header, being included by default. How can you say you don't want to be bound to FA?
Just did a grep; the Fontawesome files are definitely a part of Limesurvey, that is being used in many files.

Shnoulle:
All this part must be rewritten (and are already rewtritted partiallly). Didn't show is easily fixed by css, but show it to screen reader is the real task done in develop version https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/views/survey/questions/question_help/asterisk.php

LouisGac:
ok let's be clear: Myself, I used the fa classes. But, in the team meeting, Shnoulle and Tammo asked for the simple asterisk class. Whatever, in develop branch, class can be edited in twig file.

Shnoulle:
To be clear : we use asterisk only since a lot of time, currently some template (maybe) didn't use `fa fa-asterisk` to shown asterisk but fa-danger or (i don't care). Then updating it now break limesurvey API ....

Then : for 3.0 only , and it's already done (by default , because 3.0 allow twig file to update this part (i think)).

LouisGac:
I agree.

brainadam24:
so if it can show the askterisk by default, how do I do that?

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/blob/master/templates/default/css/template.css#L953

PR #686, opened 2017-04-25
fixed accidental linebreaks on radio button lists

brainadam24:


Shnoulle:
No : user can add paragraph or table in radio button list : then need to be a div, see w3c.

brainadam24:
so the two divs should always break into different lines?

TonisOrmisson:
You can use template css to customize the elements. 
`display: inline;`

PR #687, opened 2017-04-26
Dev: boostrap to bootstrap

tammoterhark:
Common typing error by one of the developers.


tammoterhark:
I wonder what the reference to Mantis is worth in the heading of this file. I would suggest: either make a fuull reference to a usefull page or leave it away.

tammoterhark:
This file =  bootstrap.css

Shnoulle:
Gné, gné , gné .... no it's not french , it's me ;)

PR #688, opened 2017-04-26
New feature: Copy survey from remotecontrol

evently-nl:
Adds the option to copy surveys with the remote control. I don't now if this is a feature you want into the core remotecontrol? If so, i'll make one for 2.06lts as well.


LouisGac:
seems good to me.

Shnoulle:
This function, didn't exist currently ? Yes, it's a must have in core !

tammoterhark:
I would like this very much!

maziminke:
This is a MUST-HAVE! At some older projects we dealt with loading an LSS file to copy a survey so I would be very happy to have this included at 2.06 LTS as well.

Thanks for the great contribution.

BertHankes:
See also https://github.com/LimeSurvey/LimeSurvey/pull/431

evently-nl:
@BertHankes thanks! was wondering where I got the code from, found it on an old limesurvey server I was migrating to 2.5.*, could not remember for the life of it how it got there. Still working great on 2.5 (and 2.06lts as well by the way),  why did yours never got merged?

PR #692, opened 2017-05-08
corrected a few grammar mistakes

brainadam24:


TonisOrmisson:
translations go via
https://translate.limesurvey.org/projects/

c-schmitz:
As @TonisOrmisson already said, please use the translation website, instead. Thank you.

PR #694, opened 2017-05-15
corrected fontawesome icon usage

brainadam24:
that way the nbsp will displace the icon in an unwanted way


Shnoulle:
PS : correct usage is to add sr-only for accessibility (had to move it to todo list ...)

PR #695, opened 2017-05-15
#12398: Deleted input-sm for more consistancy

tammoterhark:
and corrected obvious spelling mistake: "checkox"

https://bugs.limesurvey.org/view.php?id=12398

### DIFF
--- a/application/views/survey/questions/answer/multiplechoice_with_comments/rows/answer_row_other.php
+++ b/application/views/survey/questions/answer/multiplechoice_with_comments/rows/answer_row_other.php
@@ -31,7 +31,7 @@
         <div class="form-group answer-item text-item other-text-item ls-input-group"><!-- input-group from BS seems OK too -->
             <label for="<?php echo $id;?>" class="label-text control-label ls-input-group-extra" id="label-<?php echo $id;?>" ><?php echo $labeltext;?></label>
             <input
-               class="other-text form-control input-sm multipleco-other-topic <?php echo $classes; echo $kpclass;?>"
+               class="other-text form-control input multipleco-other-topic <?php echo $classes; echo $kpclass;?>"
                type="text"
                name="<?php echo $name; ?>"
                id="<?php echo $id;?>"
--- a/question_templates/image_select/survey/questions/answer/multiplechoice/rows/answer_row_other.php
+++ b/question_templates/image_select/survey/questions/answer/multiplechoice/rows/answer_row_other.php
@@ -37,7 +37,7 @@
         <!-- comment -->
         <div class="form-group text-item other-text-item">
             <input
-                class="form-control input <?php echo $kpclass; ?>"
+                class="form-control input-sm <?php echo $kpclass; ?>"
                 type="text"
                 name="<?php echo $myfname; ?>"
                 id="answer<?php echo $myfname; ?>"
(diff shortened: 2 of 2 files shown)


LouisGac:
Hey Tammo, thank you for that PR. 

But, for any bugs related to question rendering: see that before with Denis.  (unless it's validation problems, JS, etc)

In dev branch, I didn't  touched the HTML of the questions. So it's probably a merge problem that will be solved in a single step.

tammoterhark:
@Shnoulle Could you please review and merge (when OK?)

Shnoulle:
Didn't see ant issue here. About https://github.com/LimeSurvey/LimeSurvey/pull/695#issuecomment-301535243 , maybe related to https://bugs.limesurvey.org/view.php?id=12360 .

Currently : i'm unsure what views is loaded in develop ...

LouisGac:
I must say that I'm a little bit worry about question HTML (Debugging 2.5 HTML took a very long time because of the switch to Bootstrap 😥 ). I didn't touched the question views at all in the develop branch.  I thought that the new answer_html consisting in changing very few things and was bug free. So, I thought the current rendering errors I've noticed while twigging was due to a merge problem... even if I couldn't find any difference between the last commit on answer_html and the one one develop regarding question views. 

**Indeed, I start to understand that it's not a merge problem, right? The new HTML for question views is very different from the one of 2.5, and should be entirely retested and fixed?  Isn't it?**

tammoterhark:
What directions for a solution? What can we do, what can we ask to others?

LouisGac:
well... we should see that with Denis during the team meeting, but one thing that can be done for now is to hard reset your installation to the last commit to the answer_html branch and see if you face the same kind of bugs regarding question rendering. If yes: it means that all the question rendering must be tested and fixed.

tammoterhark:
OK, let's discuss

Shnoulle:
New HTML before https://bugs.limesurvey.org/view.php?id=12360 was hardly tested and discussed. Here : didn't think if it was or not a merging problem.

LouisGac:
about: https://bugs.limesurvey.org/view.php?id=12360 

Survey list is not a question/answer view, but a template view, so no prob with this one. My question is about  question/answer's html: I've seen a lot of small bugs while twigging. 

I bet that if you merged Tammo's PR, it's because this bug also affected answer_html branch, right? 

It's not a huge problem, it's just to organise the test/debug.

Shnoulle:
About 12360 : qanda must be fixed for Question answer part too : https://testing.sondages.pro/develop/index.php/65265

Maybe small bugs is due to this (and here : it's a big bug). You can bet anything, here : [the -sm come from 2.50](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/survey/questions/multiplechoice_with_comments/rows/answer_row_other.php#L45) or not updated in 3.0 (no major issue). Same for [othertext-label-checkbox-container ](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/survey/questions/multiplechoice/rows/answer_row_other.php#L25) more an extra helper for template dev not used by template.

When i say : OK, all done : i test whole question in https://testing.sondages.pro/answershtml/ . I send a mantis bug for the `{{ aSurveyInfo.name }}` issue just some day after you make the partial merge in develop.

tammoterhark:
So, there seems to be no big problem, not something we cannot solve! Let's go for it, quarreling doesn't solve bugs.

tammoterhark:
BTW: what is "FUD"?

Shnoulle:
[Fear, uncertainty and doubt](https://en.wikipedia.org/wiki/Fear,_uncertainty_and_doubt)

LouisGac:
That FUD thing remind me the 2.5 release :-D 

So what ever, I think we must planed a full test of question/answer HTML

tammoterhark:
Let's discuss that. I am not sure if I know what a full test is. If this is what I have done before, it is OK.

Shnoulle:
Must fix 12360 before anything else

PR #696, opened 2017-05-16
Fixed issue #12394: Order questions in quota->addQuestion

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12394


application/models/Survey.php
```diff
@@ -1256,4 +1257,31 @@ public function getsSurveyUrl()
...
+     */
+    public function getQuotableQuestions()
+    {
+        $criteria=new CDbCriteria;
+        $criteria->select = 't.*';
+        $criteria->with=array(
+            'survey.groups',
+        );
+
+        $criteria->order ='`groups`.`group_order`, `t`.`question_order`';
```
Shnoulle: Just a question : this was tested on pgsql and mssql ? Maybe must use quoteColumnName ?  
````
$criteria->order ='`groups`.`group_order`, `t`.`question_order`';
````
 seems strange and only related to mysql ?

TonisOrmisson: @Shnoulle Thanks!
No, tested ONLY on Msql. 
I added the quoteColumnName as you suggested. 
I do not have the mssql testing option right now.

Shnoulle: Thenks to you ;) . Looking at code : seems to be the best way : http://www.yiiframework.com/doc/api/1.1/CDbSchema#quoteColumnName-detail

PR #697, opened 2017-05-16
Fixed isue #12377: "Add token entry" not translated on an error screen

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12377

text echoed without eT()


LouisGac:
Thanks

PR #698, opened 2017-05-17
Fixed issue #12387: Unecessary rendering of .answer-container element…

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12387

… in text-display questions

### DIFF
--- a/assets/packages/template-default/template-core.css
+++ b/assets/packages/template-default/template-core.css
@@ -91,7 +91,7 @@ table.ls-answers input[type="checkbox"],table.ls-answers input[type="radio"]{mar
 .dir-ltr .radio-item .ls-label-xs-visibility,.dir-ltr .checkbox-item .ls-label-xs-visibility{left:auto;margin-left:-20px;}
 .dir-rtl .radio-item .ls-label-xs-visibility,.dir-rtl .checkbox-item .ls-label-xs-visibility{right:auto;margin-right:-20px;}
 
-/* no argin on label : no margiun after and before too */
+/* no margin on label : no margin after and before too */
 .dir-ltr .radio-item label.ls-label-xs-visibility::before,.dir-ltr .checkbox-item label.ls-label-xs-visibility::before{margin-left:0;}
 .dir-rtl .radio-item label.ls-label-xs-visibility::before,.dir-rtl .checkbox-item label.ls-label-xs-visibility::before{margin-right:0;}
 .dir-ltr .radio-item label.ls-label-xs-visibility::after,.dir-ltr .checkbox-item label.ls-label-xs-visibility::after {margin-left:0;padding-left:0;}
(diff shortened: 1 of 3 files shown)


tpartner:
Please see my note in the bug report - https://bugs.limesurvey.org/view.php?id=12387

Shnoulle:
Please : wait for https://bugs.limesurvey.org/view.php?id=12360 to be fixed before testing HTML part.

And here : maybe better in CSS ? `.boilerplate .answer-container {display:none}` ?

tpartner:
As I said in my comment in the bug - is there no way to prevent rendering server-side?

TonisOrmisson:
I was also looking for a place to avoid rendering at first bu looking at the twig:
https://github.com/LimeSurvey/LimeSurvey/blob/develop/templates/default/views/subviews/survey/question.twig#L79
It seemed more logical to just hide it. 
I do not think there will be a twig layout per question type (or?)

About JS vs CSS. 
I put it there because it kind of was a logic place where also some other hiding occurs triggered from 
https://github.com/LimeSurvey/LimeSurvey/blob/develop/templates/default/views/layout_main.twig#L145

Why would you rather have CSS? My argument for JS is that we already have a JS function for  this type of purpose - and I like to have things in logical places (rather than maybe the best in terms of performance).

tpartner:
As Denis knows, I am a JS fan but my argument for CSS is: 

Any time it can be done with pure CSS, it is preferable because the JS isn't fired until full page load so there may be a brief flash of the visible element before the class is applied. Also, arguably, the JS adds an unnecessary load.

Shnoulle:
Currently : there are 2 part
- `{{ aQuestion.answer | raw }}` : is not empty : I didn't know if remove the hidden input can be done without broking EM. Can make a test BUT : i think too all this hidden input must be added by EM and not by view
- Hiding unneeded answers part : it's hiding in all condition, even with no-js : use css

But the major question (my Point Of View) : it's a template issue or a core issue ?

TonisOrmisson:
@Shnoulle Im not sure I understood you fully :)

Anyway I changed it to use CSS instead of JS as you suggested. Thanks!

Shnoulle:
@TonisOrmisson normal, i don't understand me too (always) ;). Let me try to explain (about 2 only).

The question is more : did we must hide this part 

1. [for all template](https://github.com/LimeSurvey/LimeSurvey/tree/develop/assets/packages/limesurvey)
2. [for all default based template](https://github.com/LimeSurvey/LimeSurvey/tree/develop/assets/packages/template-default) 
3. Only for default template, and easily updatable by user (like you do it)

Currently : we can not do 1 (`answer-container` come from template) . But maybe it best in 2.

:)

TonisOrmisson:
Thanks - it has got much clearer now!
I agree about 2. I'll change that.

Shnoulle:
Hey, it's for discussion, not decision : @tpartner : all default based or only template ? Did you see a reason where a user can show it (some js workaround ;) )

tpartner:
@Shnoulle, no, I don't see a need for it in any JS workarounds I can think of. I suppose, if that were the case, you could insert the required element(s) with the workaround.

PR #699, opened 2017-05-17
Fixed issue #12384: Question help text not rendered

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12384


LouisGac:
thank you

PR #700, opened 2017-05-18
Fix qanda merge issues

Shnoulle:
* Fixed issue #12360: Big difference in HTML produced with previou…
* Fixed issue #12270: XSS Vulnerability (new way)
** I think i get whole filtered with the new qanda and not filtered before or in this fix
** Did we must move whole filter to views to be clean (i think we must)
** See https://github.com/LimeSurvey/LimeSurvey/pull/683#issuecomment-295442090 about made it in view and not qanda (or EM)

### DIFF
--- a/application/config/internal.php
+++ b/application/config/internal.php
@@ -239,7 +239,7 @@ $internalConfig = array(
                 'tags' => array('if', 'for', 'set', 'autoescape'),
                 'filters' => array('escape', 'raw', 't', 'merge', 'length', 'gT', 'keys'),
                 'methods' => array(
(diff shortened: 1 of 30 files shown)


LouisGac:
so globally you replaced html input by CHtml methods to avoid to use htmlspecialchars, right?

Shnoulle:
yes, and reset qanda because broken before

Shnoulle:
We must make a decision about existing htmlentities do_longfreetext for example. I think (unsure for @c-schmitz ) we must fix it in view too.

PR #701, opened 2017-05-19
fixed issue #12414: exporting printable - template assets not visible…

TonisOrmisson:
… in some extractors (eg windows default)

https://bugs.limesurvey.org/view.php?id=12414


LouisGac:
thank you for this fix

PR #703, opened 2017-05-24
Fixed issue #12368: Remote control: properties that are available are…

Shnoulle:
Fixed issue #12368: Remote control: properties that are available are…not shown

* starting, but really long ...., merge it and do the job when found time or wait completed ?
* Review some part (manual is outdated and give false information)
* Usage of MarkDown ? Must control if http://api.limesurvey.org is clean with. According to [phpDoc doc](https://www.phpdoc.org/docs/latest/references/phpdoc/basic-syntax.html) it's allowed
* Even if we make a clean job now, another dev can brake it tomorrow if phpDoc are not updated

### DIFF
--- a/application/controllers/admin/remotecontrol.php
+++ b/application/controllers/admin/remotecontrol.php
@@ -66,10 +66,9 @@ class remotecontrol extends Survey_Common_Action
                 }
                 LSjsonRPCServer::handle($oHandler);
             }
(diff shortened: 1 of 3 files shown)


Shnoulle:
Oups, my gitonfig is out of date .... i fix it for a new one

PR #706, opened 2017-05-29
Composer package definition

pprossi:
May be this could be added to be able to fetch the repository with composer?
It is a simple Package definition. No function or prozess.
- The name key must be updated to "LimeSurvey/LimeSurvey" to fit to repository path.

If a timeout occurs try:
export COMPOSER_PROCESS_TIMEOUT=600 && composer install

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-	"name" : "LimeSurvey/LimeSurvey",
+	"name" : "pprossi/LimeSurvey",
 	"description" : "The most popular FOSS online survey tool on the web",
 	"homepage" : "https://www.limesurvey.org/",
 	"license" : "GPL 2.0",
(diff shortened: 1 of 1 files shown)


TonisOrmisson:
a good idea to have that but:

package should be named as
```
"name" : "limesurvey/LimeSurvey",
```

you can use the 

```
    "repositories": [
      {"url": "https://github.com/pprossi/LimeSurvey.git", "type": "vcs"},
    ],
```

to define your fork as what you want to use

and of course the authors section ... Put Carsten there?

pprossi:
i know that i am able to use my fork with:
```
"repositories": [
        {
            "type": "git",
            "url": "https://github.com/pprossi/LimeSurvey.git"
        }
]
```

But the main idea is to stay up to date and don't manage an additional repsotitory than the main one.
Name key: as i said :)

Author: what ever you like ;)

lacrioque:
We will merge it, but correct it quickly.
I have already created a company account on composer.

PR #708, opened 2017-05-29
Fixed issue #12440: Propertie TemplateConfiguration.name undefined

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12434


TonisOrmisson:
I'm validating through models...
There is some more of this there

Shnoulle:
About the redirect : i think Throw a 404 is better : HTTP error code are done for this.

TonisOrmisson:
@Shnoulle You're probably right

Shnoulle:
Are you sure you fix home page ? No error => survey listing only

PR #709, opened 2017-05-30
Fix sid/gid/qid in Survey_Common_Action (parent of all admoin controller)

Shnoulle:
* Fixed issue #12434: [security] Reflected XSS (Thanks to mrbreaker)
* Fixed issue #12433: [security] Reflected XSS (Thanks to mrbreaker)
* fix it and test it before using it, in Survey_Common_Action

### DIFF
--- a/application/core/Survey_Common_Action.php
+++ b/application/core/Survey_Common_Action.php
@@ -64,18 +64,8 @@ class Survey_Common_Action extends CAction
         $params = $this->_addPseudoParams($params);
 
         if (!empty($params['iSurveyId'])) {
-            $oSurvey=Survey::model()->findByPk($params['iSurveyId']);
-            if(!$oSurvey) {
-                Yii::app()->setFlashMessage(gT("Invalid survey ID"),'error');
-                $this->getController()->redirect(array("admin/index"));
-            } elseif (!Permission::model()->hasSurveyPermission($params['iSurveyId'], 'survey', 'read')) {
-                Yii::app()->setFlashMessage(gT("No permission"), 'error');
-                $this->getController()->redirect(array("admin/index"));
-            } else {
-                LimeExpressionManager::SetSurveyId($params['iSurveyId']); // must be called early - it clears internal cache if a new survey is being used
-            }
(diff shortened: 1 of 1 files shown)


Shnoulle:
@TonisOrmisson : can you review this pull request please ? Fix same bug than https://github.com/LimeSurvey/LimeSurvey/pull/708 but in a different way.

I test without casting to int 
`f(!$oSurvey || $params['iSurveyId']!=$oSurvey->sid) {`
=> but pgsql broke, then test if casted params is OK
`(string)(int)$params['iSurveyId']!==(string)$params['iSurveyId']`

I think we lust control too with `sid/<XSS>/surveyid/1234` : maybe there are issue here. I make a fix in next commit.

TonisOrmisson:
Looks good to me. This is obviously a lot better solution in general. Tested a  bit and seems to work. 

I am a bit unsure if it fixes all of this though. Im not sure about the workflow but the `$iSurveyId` is sometimes also `$surveyid` and I guess also `$surveyId` and currently they do not seem to be all filtered by this? or?

one more thing 
#708 was made on develop

Using this version I could close #708 except for a mistake here
https://github.com/LimeSurvey/LimeSurvey/pull/708/files#diff-ecb8aa59f6d0a7c38ea58ec03d0f5e48L73

This is not an issue in master. Maybe I revert the validation things there and keep only 
https://github.com/LimeSurvey/LimeSurvey/pull/708/files#diff-ecb8aa59f6d0a7c38ea58ec03d0f5e48L73
and rename issue #708 about error rendering

Shnoulle:
security issue must be made on master ;).

Else : for sid/surveyid on questiongroup : i try /sid/<XSS>/surveyid/1223 and  /sid/1234/surveyid/<XSS> i don't find a way to have not filtered part. Maybe this must be controlled with code too ....

And for error rendering in develop : maybe it's the same issue i have for surveylisting : https://bugs.limesurvey.org/view.php?id=12440 ? If yes , i'm happy to merge it in develop.

Shnoulle:
About sid!=surveyid!=surveyId : maybe send a 500 error if we have it but with different values ?

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12440 
Yup. This is also what i got. I'll change #708 to fix only that

PR #725, opened 2017-06-22
first wave of 508 fixes to admin home page

billdacat:
Section 508 fixes to admin homepage, fixing h# tag order, adding span tags for screen readers, etc.

### DIFF
--- a/application/extensions/PanelBoxWidget/views/box.php
+++ b/application/extensions/PanelBoxWidget/views/box.php
@@ -5,13 +5,14 @@
         echo ' data-target="_blank"';
     } ?> >
         <div class="panel-heading">
-            <h3 class="panel-title"><?php eT($title); ?></h3>
+            <h1 class="panel-title"><?php eT($title); ?></h1>
         </div>
         <div class="panel-body">
             <div class="panel-body-ico">
                 <a href="<?php echo $url; ?>"<?php if ($external) {
                     echo ' target="_blank"';
                 } ?>>
+		<span class="sr-only"><?php eT($title); ?></span>
                 <span class="icon-<?php echo $ico; ?>" style="font-size: 4em">
                 </span>
                 </a>
--- a/application/views/admin/super/admin_notifications.php
+++ b/application/views/admin/super/admin_notifications.php
@@ -4,6 +4,7 @@
         <a aria-expanded='false' 
             href='#'>
             <span class='fa fa-bell text-muted'></span>
+	    <span class='sr-only'>Notifications</span>
         </a>
     </li>
 <?php elseif($showLoader): ?>
@@ -20,7 +21,7 @@
         </a>
         <ul class='dropdown-menu' role='menu'>
             <li>
-                <a><span class='fa fa-spinner fa-spin'></span></a>
+                <a><span class='fa fa-spinner fa-spin'></span><span class='sr-only'>Loading notifications</span></a>
             </li>
         </ul>
     </li>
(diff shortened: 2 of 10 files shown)


application/views/admin/survey/subview/accordion/_accordion_container.php
```diff
@@ -23,6 +23,7 @@
             <h4 class="panel-title">
                 <a class="btn btn-default btn-xs hide-button hidden-xs opened handleAccordion hidden-sm">
                     <span class="glyphicon glyphicon-chevron-left"></span>
+		    <span class="sr-only">Expand/Collapse</span>
```
Shnoulle: 👍 except translation

application/extensions/PanelBoxWidget/views/box.php
```diff
@@ -5,13 +5,14 @@
         echo ' data-target="_blank"';
     } ?> >
         <div class="panel-heading">
-            <h3 class="panel-title"><?php eT($title); ?></h3>
+            <h1 class="panel-title"><?php eT($title); ?></h1>
```
Shnoulle: Maybe an h2 . We don't have a global h1 for each page with the real title ?

application/views/admin/super/admin_notifications.php
```diff
@@ -4,6 +4,7 @@
         <a aria-expanded='false' 
             href='#'>
             <span class='fa fa-bell text-muted'></span>
+	    <span class='sr-only'>Notifications</span>
```
Shnoulle: translation ;)

application/views/admin/super/footer.php
```diff
@@ -52,7 +53,7 @@ class='btn btn-warning btn-sm'
         <div class="modal-content">
             <div class="modal-header">
                 <button type="button" class="close" data-dismiss="modal">&times;</button>
-                <h4 class="modal-title"><?php eT("Confirm"); ?></h4>
+                <h3 class="modal-title"><?php eT("Confirm"); ?></h3>
```
Shnoulle: I'm upset about h1/h2 etc ... Did modal need a hX ? Or to be a `<article>` with h1 :(

Shnoulle:
@billdacat Can you add the translation system please ?

And currently : header are totally broken
capture du 2017-06-23 12-14-54

Think we must have

1. Title of website 
2. Tools Box subtitle (didn't exist currently)
3. Survey Listing (if exist)

No header for modal or box

billdacat:
I'll work on that.  For the modals, yeah, it needs to be a div which I will change -- noticed that when I went to mark up another page, and the h# tags were all out of whack again.

The h# tags in general on the admin side at least appear to be chosen at random and solely for CSS purposes -- will be changing those to divs as appropriate, and leaving the h# CSS alone.

LouisGac:
It's fine for me if you move it to h# tags, but for your information, they are not chosen randomly. They come from Bootstrap documentation:

http://getbootstrap.com/components/#panels-heading

billdacat:
Right, you can use h# tags, but they have to be in order for 508 compliance.  There are pages were the only h tag is an h3, or we jump from h2 -> h5 .. that's not correct.

Bootstrap panels also use divs too, so there's really no reason to not just switch over to them, especially for modal content.

LouisGac:
also: thank you for your great work ;-)

Shnoulle:
It's not only related to 508 compliance , and again a proof than boostrap doc are not the 'way to do'.

Else, here : maybe for a quick correction and evolution : we can move to `<div class='hX'>` in place of hX ? After we can add h1 and h2, and maybe h3/

Shnoulle:
I made an error : title of the web site MUST NOT be the title of the page.

Maybe : 

- nav
- The title of the content (here : welcome page ?) (h1)
- some h2 for sub-part ?
- some h3 for sub-sub-part

PR #726, opened 2017-06-23
wave of admin508 fixes, h# tags, panel boxes, sr-only, etc

billdacat:
New CSS classes written to support div usage instead of h# tags and keep same formatting.


Shnoulle:
Can you merge before to resolve conflict please :)

PR #731, opened 2017-06-29
Front end 508 fixes

billdacat:
Couple small front-end fixes, moving h4 tag to class, adding title to iframe.


Shnoulle:
Did you trye upload question type with a screen reader or you just make 508 testing ? Because i think we must really update whole system to be screen reader compatible :)

PR #732, opened 2017-06-30
508ls

billdacat:
Lots of 508 fixes, mainly cleaning up misused h# tags


application/views/admin/participants/displayParticipants_view.php
```diff
@@ -1,5 +1,5 @@
 <div class="col-xs-12">
-    <h3 class="pagetitle row">
+    <div class="pagetitle h3 row">
```
tammoterhark: Just for my information: why is it pagetitle and modal-title (with hyphen)? Is that by design of for retro-compatibility?

Shnoulle: I say fortuity ? dice roll ?

PR #735, opened 2017-07-03
DEV: Inspect models: remove unused local params & private methods

TonisOrmisson:


application/models/UpdateForm.php
```diff
@@ -241,7 +241,7 @@ public function getChangedFiles($destinationBuild)
      * @return object
      * //TODO $downloadid not used locally
      */
-    public function downloadUpdateFile($downloadid, $tobuild)
+    public function downloadUpdateFile($downloadid=null, $tobuild)
```
Shnoulle: `$downloadid=null, $tobuild` ?

TonisOrmisson: $downloadid not used locally. My logic was if I set to null it can be removed more easily. But since its the 1st param, its not actually.

Shnoulle: I think @LouisGac is the best reviewer for updater

c-schmitz:
@TonisOrmisson Can you please update it?

PR #743, opened 2017-07-04
FIX 12479 (partial): Add DB property descriptions in models 

TonisOrmisson:
Let's see how this model looks. 
Some gaps there I was unsure how to describe.

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -20,7 +20,7 @@ use \ls\pluginmanager\PluginEvent;
 /**
  * Class Survey
  *
- * @property integer $sid primary key
+ * @property integer $sid Survey ID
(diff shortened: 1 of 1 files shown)


TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12479

TonisOrmisson:
Maybe some feedback if in general this style is OK, then I can take on other models as well

LouisGac:
great job ;-) 
thank you very much

Shnoulle:
And WE (all) must fix it, review it when we update a file :)

Shnoulle:
PS : $savetimings Whether survey timings are saved (Y/N) is really better than Y : .... N : blah 👍

PR #748, opened 2017-07-05
DEV: remove deprecated getBaseLanguageFromSurveyID($surveyid) and replace all instances

TonisOrmisson:


application/controllers/admin/dataentry.php
```diff
@@ -508,7 +508,8 @@ public function editdata($subaction, $id, $surveyid, $language='')
         $survey = Survey::model()->findByPk($surveyid);
         $id = sanitize_int($id);
         $aViewUrls = array();
-        $sDataEntryLanguage = Survey::model()->findByPk($surveyid)->language;
+        $survey = Survey::model()->findByPk($surveyid);
```
Shnoulle: Why not `Survey::model()->findByPk($surveyid)->language;` ?

PR #752, opened 2017-07-06
DEV: kill SurveyObj: remove unused things

TonisOrmisson:
I really do not undertstand why there is SurveyObj and why Survey is not used instead. It is really confusing and in many places also the phpDoc is mixed up. 

I'll try to kill it. 
Here's the first part removing unused stuff

### DIFF
--- a/application/helpers/admin/export/SurveyObj.php
+++ b/application/helpers/admin/export/SurveyObj.php
@@ -6,7 +6,6 @@ class SurveyObj
     */
     public $id;
 
-
     /**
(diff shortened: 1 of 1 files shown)


LouisGac:
yep the export code is quiet messy.
are you absolutely sure this code is not used?

TonisOrmisson:
I checked all usages for the deleted parts and did not find usages. 
There is nothing absolute of course about it :)
I could re-check it once more if you ask

PR #754, opened 2017-07-07
DEV: Model DB property docs Vol3

TonisOrmisson:
We have a s***y vacation weather ! :(

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -18,22 +18,22 @@ if (!defined('BASEPATH'))
 /**
  * Class Question
  *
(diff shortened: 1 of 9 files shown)


application/models/Question.php
```diff
@@ -18,22 +18,22 @@
 /**
  * Class Question
  *
- * @property integer $qid
- * @property integer $parent_qid
+ * @property integer $qid Question ID. NB! Primary key is qid & language columns combined!
```
Shnoulle: I like the NB ;)

PR #760, opened 2017-07-07
DEV: clean frontend_helper

TonisOrmisson:


application/helpers/frontend_helper.php
```diff
@@ -2259,16 +2232,3 @@ function getSideBodyClass($sideMenustate = false)
     return "";$class;
 }
 
-/**
- * For later use, don't remove.
```
Shnoulle: Maybe ?

TonisOrmisson: Thanks, missed that. Reverted that!

PR #761, opened 2017-07-09
DEV: clean out survey->surveyinfo VOL1

TonisOrmisson:
The idea is to replace surveyinfo with oSurvey. Eventually hopefully killing surveyinfo. Clean out simply replaceable usages at first.


Shnoulle:
Great idea \o/

LouisGac:
Yes, at last, from template, we should access to basic objects: oSurvey, oGroup, oQuestion, oQuotas, etc.
but a lot of things to do before we can reach it. 

For surveyinfo: you can dump it from any twig file. There is still a lot of unnecessary stuff in it (like mail related stuff), and some datas are not directly related to oSurvey (like navigation datas: progress, etc )

If you need to find the output for the old aSurveyInfo before rendering, grep: "twigRenderer". Basically, it's inside frontend and surveyruntime helper. 

Thank you again for your contributions, they are greatly appreciated.

LouisGac:
Just read the PR in detail,  and I realized your changes concern backend indeed. 
Be careful not to collide with Markus work: 
https://github.com/LimeSurvey/LimeSurvey/tree/develop_adminpanel

for now, he's working only on menus, so it should be fine. But this week he should start to work on subquestion/answers etc.

TonisOrmisson:
Thanks for the tip. I'll try to have an eye on what Markus does.

PR #769, opened 2017-07-11
fix views: survey create

TonisOrmisson:
It seems there are some views still broken after all the changes I submitted. 
I'll try to clean up now a bit for a while :)

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -149,7 +149,8 @@ class SurveyAdmin extends Survey_Common_Action
         $aData                = $this->_generalTabNewSurvey();
         $aData                = array_merge($aData, $this->_getGeneralTemplateData(0));
         $aData['esrow']       = $esrow;
-        
+        $aData['oSurvey']       = $survey;
+
         //Prepare the edition panes
 
         $aData['edittextdata']              = array_merge($aData, $this->_getTextEditData($survey));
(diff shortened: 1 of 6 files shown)


LouisGac:
Tonis, I just pulled your work so I could push mine, and I get a notice while launching a survey:

> Trying to get property of non-object
> application/controllers/survey/index.php(265)
> if ($surveyExists == 1 && $survey->hasTokensTable){

Please, remember to set debug mode to 2 when you code.

Thank you again for your great work.

LouisGac:
(I just pushed a fix)

TonisOrmisson:
Thanks!
Yes I have the debug=>2 while coding. 
It's just I have trouble testing all possible views after the changes. I am quite sure I have broken something else too :(

I really miss some automated testing for the main views. 

I have done some basic testing with Codeception. Codeception seems to be the preferred testing framework for Yii2 so looking to future maybe it would be an option to implement that?
http://codeception.com/docs/modules/Yii1

It would be really helpful to run at least some a AcceptanceTests on a bunch of views automatically to see that nothing has been badly broken with changes.

And you are welcome to have my contribution. I have some time to spare right now and I am happy to contribute :)

LouisGac:
LimeSurvey is the result of years of developments, from PHP4 to today, from spaghetti code to Yii. So, of course, no TDD for LS... and touching one thing can affect many other ones. That's why I always try to keep the old logic (I never rewrite from scratch, but rather do cycles of refactorisation), and why I don't remove a method before I'm 100% sure it's not used anymore anywhere. 

But sure, some testing would be extremely useful. Olle started to quickly add some unitests in Master Branch. In Develop, it should be done using the Yii1 PHP Unit implementation:
http://www.yiiframework.com/doc/guide/1.1/en/test.unit

Feel free to do it if you want, you have our trust.

PR #773, opened 2017-07-13
DEV: implement Survey booleans3

TonisOrmisson:


LouisGac:
did you tested it?

TonisOrmisson:
Not really no. Mostly the same type of changes using the $oSurvey->isActive instead of $oSurvey->active == 'Y' / 'N'

LouisGac:
Knowing that we don't have unit testing, please, always test manually your changes. 
We can't blind push commits: just like in the 2000's 😕

PR #774, opened 2017-07-13
Indentconditional

billdacat:
add ls-conditional class so conditional questions can be targeted via css, for indenting, bordering, etc etc


Shnoulle:
I don't understand ?

I explain :
1. here you add ls-conditional in javascript. But you already have ls-irrelevant. Then the new class are not used.
2. This class don't show relation with another question (reading what did you put on irc). Relevance and “tied to the previous question” are nothing related. Sometime it's the previous question, but sometimes it's 5 question before. It's your usage , not ls related.
3. Really easy to do it in plugin

My point of view : there are no reason to add «this question have relevance somewhere and can be hidden» except to show it with sr-only.

Currently, you can easily add 'have-condition' with beforeQuestionRender simple example : 

```
$question = question::model()->find("qid=:qid and langage=:lang",array(
  ":qid"=>$this->getEvent()->get("qid"),
  ":lang"=>App()->getLanguage()
));
if($question->relevance!="1" or $question->relevance==="" or isnull($question->relevance)) {
  $this->getEvent()->set("class",$this->getEvent()->get("class")." have-condition");
}
```

See https://framagit.org/Shnoulle/addQuestionCodeClass/blob/master/addQuestionCodeClass.php#L102 for usage of beforeQuestionRender. You can do this on a 2.6lts

PR #775, opened 2017-07-14
FIX: Survey form:admin email

TonisOrmisson:


TonisOrmisson:
not fixed still :(

TonisOrmisson:
OK that should do it.
Moved the create/update difference from view to the model where it should be

c-schmitz:
Is this fix still needed?

TonisOrmisson:
I would say yes. 
I you create new survey, the owner & email fields are blank right now.
Im not sure about the conflicts right now, cant see them here.

c-schmitz:
Could it  be that you have 'tabs' activated in your source editor?
All tabbing should be done by using spaces.

TonisOrmisson:
no, spaces. I just merged, but now its broken

TonisOrmisson:
OK now should be good again. 
The logic & decisions are done in model rather than the view. 
If, its the new model, we use current users data

PR #780, opened 2017-07-14
FiX: some survey texts can not be cleared out

TonisOrmisson:
I wanted to clear out survey end-url and it was not possible. Also the basic text blocks should be able to be cleared out.


lacrioque:
The check for empty are absolutely necessary, because otherwise an empty post would always change the survey setting to empty. 
So we either need to find a better way to edit surveys, or this needs to stay in place.
Have you tested this?

Shnoulle:
Need !is_null ? `Yii::app()->request->getPost('description_'.$langname);` return null, BUT we fixCKeditor after …

TonisOrmisson:
The model validation should be done by Model->rules()

Currently - one can create a new survey with an empty description (and welcometext etc.), but one can not clear out the existing text if the text already exists. Which is not consistent logic. 

If the idea is that Survey must always have a welcometext then it should be set by the model rules. 

Also, the Empty post validation should be solved by using the Survey->save() method rather than any custom things. Also - in this PS oit still checks for the title not to be empty so an "empty post" will not pass here. And if I have an empty welcometext, then this is what i want :)

TonisOrmisson:
In a sense this was a quick fix for a problem that just faced me while I wanted to clear some texts. But in longer run the survey->save() and model rules is what should be worked on :)

TonisOrmisson:
Just retested it, works

lacrioque:
OK.
Yes I'm with you it should be done in Survey->save().
We just lack the time to transfer the logic to the surveymodel, and since we do only bugfixing for 3.0

Have you tested peripheral posting? Try posting any setting from somewhere and have a look on the values.

Better to use the yii->request->getPost() with null as standardvalue and check against null as shnoulle supposed.

TonisOrmisson:
No didn't test peripheral posting. 

Why only bugfixes for 3.0? Where the develop should go then? :)

And to be frank, this specific PR is a bug fix. There is different validation on creating and updating a model.

I'll work a bit on trying to use the yii way

LouisGac:
Don't worry Tony, you can continue to submit your great PR

PR #782, opened 2017-07-19
Fixed issue #12504: Tidy not available - Blocking quick translate

adamzammit:
Limesurvey depends on the PHP extension 'tidy' for the quick translate functionality. I've added this as a requirement in the installer.


LouisGac:
it has been fixed by Markus Friday. So, it's no more a requirement (we could make it a requirement for LS3).

PR #783, opened 2017-07-19
Fix issue 12452 Conditional routing doesn't work on scale and array questions

opinioni:
See https://bugs.limesurvey.org/view.php?id=12456 for more details


LouisGac:
it has been fixed by Markus Friday. So, it's no more a requirement (we could make it a requirement for LS3).

opinioni:
Hi all,
seems like this issue is still open, see:
https://github.com/LimeSurvey/LimeSurvey/blob/master/scripts/expressions/em_javascript.js#L789

didn't test the last version but maybe you fix the issue in other way?

lacrioque:
This is a working fix

PR #784, opened 2017-07-20
Fix multi array filtering

evently-nl:
Currently when trying to filter responses in the statistics view, the dual array questions are not used. This commit makes it work again: however when debugging is turned on, it now creates an error (echo htmlspecialchars($_POST[$myfield2]);, this expects a string but now gets an array (which makes sense). It was not really clear to me what the function of that echo was, so let me know if it should now loop the array, or whatever you want, and I'll update the pr!

off topic: whenever I create a pull request it always adds a lot of empty merge commits, how do I fix that? Thanks :)


TonisOrmisson:
off topic:

If you do a new branch for a fix: 
Update the local repo eg: `git fetch --all `
Checkout the upstream eg: `git checkout upstream/master`
Name the new branch `git checkout -b my-fix-branch`

This way you'll have the new branch as clean from the latest upstream master without the merge commits.

Shnoulle:
name accept # ? HTML validation ?

LouisGac:
the statistic code is pure hell. We'll rewrite it from scratch for 3.5 or 4.0. Until that time, only do bug fixing on it. Any modification can imply a lot of cascading bugs.

PR #786, opened 2017-07-24
Allow 'No answer' option to be translated in mobile version

toniginard:
The translation for 'NA' is available in desktop version, but not in mobile version (small screens)


Shnoulle:
Not like this : see https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/survey/questions/arrays/5point/rows/cells/answer_td_input.php#L18 .

The id must be answerSGQ-NA or something near (i think it's must be `answerSGQ-` to be EM compatible. 

Need another var : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/survey/questions/arrays/array/no_dropdown/rows/cells/answer_td.php#L33

PR #790, opened 2017-08-11
Update surveysettings.js

BertHankes:
Dev: 2 more standard templates added for template preview minimal and material


Shnoulle:
My pov : need to be a javascript variable, because `standardtemplaterootdir` can be updated by admin user. Then need to put it in LS var for example.

LouisGac:
the file has been removed

PR #791, opened 2017-08-12
Implemented Composer PSR-4 autoloading

machaven:
Hi,

I saw in the code comments that the autoloader class should be replaced by composer at some point. So here it is :)

When modifying the autoload configuration in composer.json, you will need to run the command: `composer dump-autoload`. This will recreate/update the autoload files in the vendor folder.


LouisGac:
hey, thank you very much.
Have you tested it?

machaven:
@LouisGac Yes, i've tested this on development, mainly around the plugin manager.

LouisGac:
thank you for the great job

PR #792, opened 2017-08-14
Better namespace naming convention

machaven:
Related to the previous composer autolaoding PR. Fix namespaces to conform to PSR naming conventions (StudlyCaps).


LouisGac:
Thank you very much for that great improvement.

Shnoulle:
After this update : there are no way to have a 2.6X + 3.0 compatible survey. 

Then : user can not update plugins and update limesurvey after OR update limesurvey + update plugins after.

Then must be a solution to deactivate plugins WITHOUT broke it (for example some plugins can delete their database when update) if they are not out of date …

PR #793, opened 2017-08-14
New feature #12563 remote control: process files on add_response()

pardo-bsso:
Dev: When calling add_response(), if the response data has fields for file
attachments, move those files to the final upload folder if they are not there.
This assumes that the files are already present on the temporary upload folder.


olleharstedt:
Looks good to me. Anyone else wants to comment? @Shnoulle @LouisGac

PR #796, opened 2017-08-17
New Feature: Set quotas by equation

TonisOrmisson:
Add the option to set a quota by Equation question


TonisOrmisson:
related to:
https://bugs.limesurvey.org/view.php?id=12402

Shnoulle:
At first look : seem great \o/ . No update on logic but on model and admin view :).

I check it :)

maziminke:
Great, finally we can use EM for quotas!

PR #798, opened 2017-08-30
fixed #12639 Quota end-url does not allow surrounding EM

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=12639

### DIFF
--- a/application/models/QuotaLanguageSetting.php
+++ b/application/models/QuotaLanguageSetting.php
@@ -83,7 +83,6 @@ class QuotaLanguageSetting extends LSActiveRecord
     {
         return array(
             array('quotals_message','required'),
-            array('quotals_url','url'),
             array('quotals_name','LSYii_Validators'),// No access in quota editor, set to quota.name
             array('quotals_message','LSYii_Validators'),
             array('quotals_url','LSYii_Validators','isUrl'=>true),


Shnoulle:
In commit history : https://github.com/TonisOrmisson/LimeSurvey/commit/bd86ccfd2884e7acb439fe2e425fdbf824a1e126#diff-f9dd264368810c6303206b6b095991cfR77

Else Surey admin can add XSS even if they don't have the right to do, and we fix some other part.

TonisOrmisson:
this is still there
https://github.com/LimeSurvey/LimeSurvey/pull/798/files#diff-f9dd264368810c6303206b6b095991cfL88

When I was changing the form to use Yii logic I also added Yii default url validation rule. But I did not realize that an url can be **surrounded** by EM also, so generic url validation is not an oprion.

Shnoulle:
Ah yes , we have 2 rules ??? `array('quotals_url','LSYii_Validators','isUrl'=>true),` and `array('quotals_url','urlValidator'),` urlValidator didn't break Expression ? It's needed with isUrl=>true ?

Shnoulle:
I think we must remove urlValidtor too , see http://www.yiiframework.com/doc/api/1.1/CUrlValidator#defaultScheme-detail

TonisOrmisson:
no
urlValidator refers to custom:
https://github.com/LimeSurvey/LimeSurvey/pull/798/files#diff-f9dd264368810c6303206b6b095991cfR94

This cheks url to be mandatory if redirection is enabled. 
the CUrlValidator should defined as  `array('quotals_url','url'),` Which Is what I deleted

PR #799, opened 2017-08-30
Fixed issue: malformed pagination query in SQL Server

odalecne:
Due to a weak implementation of the `CMssqlCommandBuilder::applyLimit()` method in Yii SQL Server driver the pagination query at the "Central participant database" page fails because "The column 'participant_id' was specified multiple times" (error 8156).

I could provide also a patch for the driver but since the Yii framework has reached its EOL (currently they accept only security fixes) one may easily avoid this error by qualifying the column names to be selected without duplicates that may arise (for example) from a * selection on joined tables.


c-schmitz:
Great fix - thank you!

PR #800, opened 2017-09-01
Update RegisterController.php

BertHankes:
DEV: added afterRegister event (see feature request)


Shnoulle:
Why not, but i think if we have more public function to display or get message or to get “default from core” and update it: it can be better :). 

For example, here : i use [beforeRegister](https://framagit.org/SondagePro-LimeSurvey-plugin/registerQuick/blob/master/registerQuick.php#L111) event to do my own job. 

Maybe set message in beforeRegister do the trick :) https://github.com/BertHankes/LimeSurvey/blob/d389ba03c5286995ae2967b472f5fbf0d5511661/application/controllers/RegisterController.php#L109

PR #801, opened 2017-09-04
Add better permission handling for expiration

realitix:
Currently, when an admin expires a survey, if he doesn't have the
permission, he sees a blank page. With this commit, he is now
redirected to the survey with an explanation message.

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -1450,7 +1450,8 @@ class SurveyAdmin extends Survey_Common_Action
         $iSurveyID = (int) $iSurveyID;
         if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveysettings', 'update'))
         {
-            die();
+            Yii::app()->setFlashMessage(gT("You do not have permission to access this page."),'error');
+            $this->getController()->redirect(array('admin/survey','sa'=>'view','surveyid'=>$iSurveyID));
         }
         Yii::app()->session['flashmessage'] = gT("The survey was successfully expired by setting an expiration date in the survey settings.");
         Survey::model()->expire($iSurveyID);


Shnoulle:
Did expires still accessible ? In fact : is user didn't have Permission : this button must be deactivated (and the check must be done too).

realitix:
Hello @Shnoulle,
Yes expire still accessible.

When you click on `Stop this survey`, you get this image
capture du 2017-09-04 10-48-23

When you set the permission `survey activation` , you can deactivate the survey but you can't expire it (is it the desired behavior by the way?).

Shnoulle:
@realitix OK, then the button must be disable if user don't have Permission. Didn't know if we must use (and fix) `$canactivate` or directly [test Permission in view](https://github.com/LimeSurvey/LimeSurvey/blob/c525eb99772ecf2530cd997e1b1a7f5ec5e41787/application/views/admin/survey/surveybar_view.php#L84)

realitix:
@Shnoulle Just to be sure: Is it normal that deactivation needs `survey activation` permission and expire needs `survey setting` permission ? From my point of view, it should be the same permission.

Shnoulle:
🤣 from my point of view : settings + content + surveylocale + activation + quota can be one Permission. Sorry mùade a little error in my comment finally

I think : is user have settings : he can update Date in settings, then no need deactivate (or activate) button. Then

1. If user don't have `activation` update : disable deactivation menu button
2. If user don't have `settings` update : disable or remove [_Expire survey_ button](https://github.com/LimeSurvey/LimeSurvey/blob/c525eb99772ecf2530cd997e1b1a7f5ec5e41787/application/views/admin/survey/deactivateSurvey_view.php#L48) in the confirmation form 

Maybe disable because else we need to remove other parts …

PR #802, opened 2017-09-06
Set expire permission to "activate"

realitix:
Hello,
In the survey view, when you click on "Stop this survey", you see this view:
capture du 2017-09-04 10-48-23

Like you can see, in the UI, you can choose between `Expire survey` and `Deactivate survey`. To deactivate the survey you need the `surveyactivation` permission but to expire the survey you need the `surveysettings` permission.

This patch fixes this inconsistency by setting the `surveyactivation` to expire. The `surveysettings` permission gives too much permission if you just want to allow the admin to expire the survey. Moreover, the UI is clear: it should be the same permission for both buttons.

Sincerely,
Jean-Sébastien

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -1448,7 +1448,7 @@ class SurveyAdmin extends Survey_Common_Action
     function expire($iSurveyID)
     {
         $iSurveyID = (int) $iSurveyID;
-        if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveysettings', 'update'))
+        if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveyactivation', 'update'))
         {
             Yii::app()->setFlashMessage(gT("You do not have permission to access this page."),'error');
             $this->getController()->redirect(array('admin/survey','sa'=>'view','surveyid'=>$iSurveyID));


Shnoulle:
@LouisGac i think we must have a discussion about Survey Permission ;)

realitix:
Indeed, I submitted this pull request to open the discussion.

c-schmitz:
I think this patch should not be accepted. Survey activation times are part of the survey settings currently.
Changing the permission only here would only make it confusing even more.

However it would be a different thing if start/endtime of the survey would not be part of the survey settings anymore, but part/setting of the activation process.

realitix:
Thanks @c-schmitz for your answer. Maybe someone else could confirm your answer to abandon this pull request.

c-schmitz:
@realitix Would you be willing to work on such a patch?

Shnoulle:
My POV is already here : https://github.com/LimeSurvey/LimeSurvey/pull/801#issuecomment-326918930

> I think : is user have settings : he can update Date in settings, then no need deactivate (or activate) button. Then
> 
> 1. If user don't have `activation` update : disable deactivation menu button
> 2. If user don't have `settings` update : disable or remove [_Expire survey_ button](https://github.com/LimeSurvey/LimeSurvey/blob/c525eb99772ecf2530cd997e1b1a7f5ec5e41787/application/views/admin/survey/deactivateSurvey_view.php#L48) in the confirmation form 
> 
> Maybe disable because else we need to remove other parts …

realitix:
@c-schmitz Sadly, I don't have enough spare time for that.
If my client (I work at Linagora) wants to pay for the new feature, I will let you know.

Thanks

PR #803, opened 2017-09-06
issue #12652: AuditLog plugin reset langage to default

Shnoulle:
- [12652: AuditLog plugin reset langage to default when register](https://bugs.limesurvey.org/view.php?id=12652)
- fix LSWebUser : no reason to set language here.
- really unsure original issue about language in plugin manager is fixed but i don't find a way to reproduce
- but pluginmanager extend admincontroller : then one can set lang
- admin part get adminlang
- other controller must check the language, but the default is set
- login form language are not updated after error.

### DIFF
--- a/application/controllers/AdminController.php
+++ b/application/controllers/AdminController.php
@@ -101,24 +101,21 @@ class AdminController extends LSYii_Controller
     {
         // From personal settings
         if (Yii::app()->request->getPost('action') == 'savepersonalsettings') {
-            if (Yii::app()->request->getPost('lang')=='auto')
-            {
+            if (Yii::app()->request->getPost('lang')=='auto') {
                 $sLanguage= getBrowserLanguage();
-            }
-            else
-            {
+            } else {
                 $sLanguage=sanitize_languagecode(Yii::app()->request->getPost('lang'));
             }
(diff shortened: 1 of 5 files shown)


Shnoulle:
@olleharstedt do you find any issue here for plugin usage ? I think a plugin for admin must use AdminController and plugin for user must use DirectRequest, then we set a language by default. I don't find a way to have issue (and done test with login part …) but who really know ?

olleharstedt:
Not sure what I'm looking at. Why would it be a problem for plugins, exactly?

Shnoulle:
Just because of https://github.com/LimeSurvey/LimeSurvey/pull/803/files#diff-707cce6f9455b8b2bf2497ceced9a0c7L13 … but tested and tested , didn't find any bug. 

But touch to a core controller can break other things.

olleharstedt:
Which commit added those lines?

Shnoulle:
https://bugs.limesurvey.org/view.php?id=12652#c44397 :)

olleharstedt:
OK, I will trust you. :)

Shnoulle:
Must fix this one : https://bugs.limesurvey.org/view.php?id=12655 at same time (to be sure) , seems to be a new one :/

PR #804, opened 2017-09-15
FIX: Typecheck the XML variable from the GroupImport method

Kbroeren:
Removed the validation from the groupimport just as the questionimport method becaus the validation always match the false state. So the groupimport can never be used.

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -30,7 +30,7 @@ function XMLImportGroup($sFullFilePath, $iNewSID)
     $xml                   = simplexml_load_string($sXMLdata,'SimpleXMLElement',LIBXML_NONET);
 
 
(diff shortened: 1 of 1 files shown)


tammoterhark:
Kevin: contactformulier op je site werkt niet, daarom maar zo:

https://www.limesurvey.org/forum/development/111625-dutch-user-group-of-limesurvey

Shnoulle:
Not best with $xml === false ? If lsg it's not an xml ? (and same must be applied to lsq if it's not the case)

Kbroeren:
@Shnoulle thats another option.. that also works. We can do the same for the question import. Then we have also type checking. I will update my PR

Kbroeren:
Thanks @Shnoulle. I have changed the message. Everything OK now ?

PR #805, opened 2017-09-16
Issue #12678: Save ranking question by model

Shnoulle:
- fix it for RemoteControl too, if RC use model …
- if RC didn't use model : he must use it ;)

### DIFF
--- a/application/models/QuestionAttribute.php
+++ b/application/models/QuestionAttribute.php
@@ -75,6 +75,27 @@ class QuestionAttribute extends LSActiveRecord
         );
     }
 
+    /**
+     * {@inheritdoc}
+     * Reset silently readonly attributes
+     */
+    public function beforeSave() {
+        if (!$this->isNewRecord) {
+            $attributeDefinition = \QuestionHelper::getAttributesDefinitions();
+            if(isset($attributeDefinition[$this->attribute]) && $attributeDefinition[$this->attribute]['readonly']) {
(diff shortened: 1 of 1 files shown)


Shnoulle:
@TonisOrmisson at https://github.com/LimeSurvey/LimeSurvey/pull/805/files#diff-1b4fbf906de3cb701eafa9ff1bec25a6R90 : i use `$oQuestion = Question::model()->find("qid=:qid",array('qid'=>$iQuestionID));` but because there are a relation : we can not get it directly ? If you know the good way, i'm happy to learn :)

Shnoulle:
There are issue : need to disable delete expect if question is deleted …

Shnoulle:
Must review whole

TonisOrmisson:
@Shnoulle 
Ideally yes it should be `$this->question`. But the relation is not defined ok
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/QuestionAttribute.php#L62
Relation should logically be `question` instead of `qid` and I think it does actually not work well due to the missing language attribute in relation (the key is qid-language)

And `$this->question->survey->isActive` :)

TonisOrmisson:
I think QuestionAttribute also could have a shortcut to `$this->survey` by defining the `getSurvey()`

TonisOrmisson:
https://github.com/LimeSurvey/LimeSurvey/pull/806

PR #806, opened 2017-09-17
dev: add Question & Survey Relations to QuestionAttribute

TonisOrmisson:


Shnoulle:
Question : it's not best to use fixed Relation ? 
Else : original issue was closed : need beforeDelete too (and beforeDelete not done when deleteAll) , then it' improvment of code not related to issue :). Maybe more for develop ?

PR #809, opened 2017-09-19
New feature: Add CC and BCC support to BeforeTokenEmail event

c4xp:
Hi,
We needed a way to add CC and BCC to emails being send out.
Those parameters should be editable by an external plugin.
Thank you.


Shnoulle:
Usage of phpmailer seems best (in my point of view) http://phpmailer.github.io/PHPMailer/classes/PHPMailer.PHPMailer.PHPMailer.html#method_addCC

Shnoulle:
stackoverflow : 2014 topic , see https://github.com/PHPMailer/PHPMailer/commit/1c43605340fe5ff26450b0ae93d4ad2058039ee6 and i test it woith mail if i remind : working with current version on LimeSurvey, https://github.com/PHPMailer/PHPMailer/issues/79#issuecomment-26056636

Shnoulle:
I'm for make a cleaner sendMail function and not update add a var in event again :). And maybe addCC more than cc

LouisGac:
Sorry, I agree with Denis. I understand you did it for now via a plugin, but for develop, feel free to update directly the core

PR #811, opened 2017-09-20
New feature : #12691: Allow initial superadmin to login via LDAP

Shnoulle:
- Just a settings more tested to disallow user 1
- User 1 is the only one needed even for other superadmin


Shnoulle:
@Aestu i think you use LDAP : can you review please :). I don't see any issue because it's disable by default :)

Shnoulle:
@tammoterhark : Done : https://github.com/LimeSurvey/LimeSurvey/pull/811/commits/5564716d4c10229b5bb7f87d5a47fb88f2e2030f

PS: next time : give me the sentence directly ;)

PR #812, opened 2017-09-20
Dev: allow to use Yii log system for tracevar even with debug=>0

Shnoulle:
- with debug=>0 : log is not done (http://www.yiiframework.com/doc/api/1.1/YiiBase#trace-detail)
- with log : we allow user to have their way to log
- Just a shortcut, but quickest way to trace ;)
- No change with default config, then easy merge.

### DIFF
--- a/application/helpers/globals.php
+++ b/application/helpers/globals.php
@@ -15,7 +15,6 @@
         return Yii::app();
     }
 
-
     /**
      * If debug = 2 in application/config.php this will produce output in the console / firebug
      * similar to var_dump. It will also include the filename and line that called this method.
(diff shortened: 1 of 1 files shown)


alexandcote:
This change introduced a bug. If we check [the logging documentation](https://www.yiiframework.com/doc/guide/1.1/en/topics.logging) of YII: 

>Messages can be logged by calling either Yii::log or Yii::trace. The difference between these two methods is that the latter logs a message only when the application is in debug mode.

So with the `Yii::log`, we always display trace. We should rollback theses change and tag a new 2.X version 😃

PR #814, opened 2017-09-25
Fixed issue #12704: Participants receive wrong tokens after testing

Shnoulle:
Dev: tempVars seem used for "forced" var
Dev: and template_replace with replacement replace the value for "forced" var

In fact : before _GetVarAttribute function return (in order) : value from survey (session) , if not set temporary value. Now : _GetVarAttribute return in order : temporary value(tempVars), value from survey (session) (knowVars).

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -4943,10 +4943,6 @@
                     {
                         $knownVar = $LEM->knownVars[$LEM->qcode2sgqa[$k]];
                     }
-                    else if (isset($LEM->tempVars[$k]))
-                    {
-                        $knownVar = $LEM->tempVars[$k];
-                    }
                     else
                     {
                         continue;
@@ -8810,20 +8806,16 @@ EOD;
         static public function isValidVariable($varName)
         {
             $LEM =& LimeExpressionManager::singleton();
-
-            if (isset($LEM->knownVars[$varName]))
-            {
+            if(isset($LEM->tempVars[$varName])) {
                 return true;
             }
-            else if (isset($LEM->qcode2sgqa[$varName]))
-                {
-                    return true;
-                }
-                else if (isset($LEM->tempVars[$varName]))
-                    {
-                        return true;
-                    }
-                    return false;
+            if(isset($LEM->knownVars[$varName])) {
+                return true;
+            }
+            if(isset($LEM->qcode2sgqa[$varName])) {
+                return true;
+            }
+            return false;
         }
 
         /**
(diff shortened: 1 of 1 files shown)


Shnoulle:
touching EM : need a review to don't break thousand of system ;)

application/helpers/expressions/em_manager_helper.php
```diff
@@ -4943,10 +4943,6 @@ static function StartSurvey($surveyid,$surveyMode='group',$aSurveyOptions=NULL,$
                     {
                         $knownVar = $LEM->knownVars[$LEM->qcode2sgqa[$k]];
                     }
-                    else if (isset($LEM->tempVars[$k]))
```
Shnoulle: Here it's for "prefill answer by survey". and $tempVars is not used for prefilling answers.

Shnoulle:
There are other place where tempVars is used and i don't understand why. https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/48a5d5c256cc126ad21a104eb9cb50fd938f6f0d/application/helpers/expressions/em_manager_helper.php#L9148 (SetVariableValue) used in ExpressionManager  https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/48a5d5c256cc126ad21a104eb9cb50fd938f6f0d/application/helpers/expressions/em_core_helper.php#L2214 assigment. BUT i didn't understand : assigment is done for all question type (knowVars) and not tempVars …

olleharstedt:
I'd prefer a unit test for this, personally.

olleharstedt:
Make a unit test _without_ the bug fix, and make sure the test _fails_. Then apply the fix and check so the test passes.

Shnoulle:
Maybe : but : i didn't really know the usage of tempVars : did they force or not the final value shown to user ? You can set unit test, but before : what is the desired behaviour …

Shnoulle:
> make sure the test fails

It's not the point, did we NEED to enforce tempVar or not … condition for fail :). Like always : there are no real comment/explanation on EM code. Then track "What usage" is difficult

olleharstedt:
You have to create a test that automatically runs the steps in "How to reproduce":

> 1. Create a new survey that uses tokens
> 2. Create a few test participant with different test tokens (e.g. 1234, abcd) and an e-mailaddres that you can access
> 3. Change the invitation e-mail and include the {TOKEN} and {TOKEN:TOKEN} field (maybe some others you want to test)
> 4. Test this survey while being logged in (use the same browser, different tab) with the token (1234) you have created before
> 5. Now send the invitation e-mails from the admin gui

olleharstedt:
You don't have to send any emails in the test, just checking the content in the email should be enough.

olleharstedt:
Step 4 might be hard...

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/814#issuecomment-332143868 : all done. I can confirm the mantis issue and THIS commit fix THIS issue. But maybe introduce another issue …

The point 4 is really easy : see [the bug report](https://bugs.limesurvey.org/view.php?id=12704).

I really don't want to introduce [new bug](https://bugs.limesurvey.org/view.php?id=12703)

Shnoulle:
PS : the test is more easy : just use [Replacefields](https://github.com/LimeSurvey/LimeSurvey/blob/fadb0cc4a2c2b5b82cb0b7af4eddffaf9863bc46/application/helpers/replacements_helper.php#L764) with one value in session + one value in $fieldsarray . The question is : did `$fieldsarray` must be forced or not …

PR #815, opened 2017-09-27
Fixed issue #12697: SPSS Export - responses truncated for incomplete answers

Shnoulle:
- force_refresh before exporting (like export by plugins)
- Get real size from ALL db for string
- Get the real decimal value and size from DB for numeric
- Fix issue with Array number/text input (show as 1 to 10 choice)

### DIFF
--- a/application/controllers/admin/export.php
+++ b/application/controllers/admin/export.php
@@ -370,16 +370,18 @@ class export extends Survey_Common_Action {
                 $iLength     = '255'; // Set the max text length of the Value
                 break;
             case 2:    //>=16
-                $iLength     = '16384'; // Set the max text length of the Value
-                break;
             default:
                 $iLength     = '16384'; // Set the max text length of the Value
         }
 
         $headerComment = '*$Rev: 121017 $' . " $filterstate $spssver.\n";
 
-        if ( isset($_POST['dldata']) ) $subaction = "dldata";
-        if ( isset($_POST['dlstructure']) ) $subaction = "dlstructure";
+        if ( Yii::app()->request->getPost('dldata') ) {
+            $subaction = "dldata";
+        }
+        if ( Yii::app()->request->getPost('dlstructure') ) {
+            $subaction = "dlstructure";
+        }
 
         if  ( ! isset($subaction) )
         {
@@ -441,8 +443,7 @@ class export extends Survey_Common_Action {
         Yii::app()->loadHelper("admin/exportresults");
         viewHelper::disableHtmlLogging();
 
-        if ( $subaction == 'dldata' )
-        {
+        if ( $subaction == 'dldata' ) {
             header("Content-Disposition: attachment; filename=survey_" . $iSurveyID . "_SPSS_data_file.dat");
             header("Content-type: text/comma-separated-values; charset=UTF-8");
             header("Cache-Control: must-revalidate, post-check=0, pre-check=0");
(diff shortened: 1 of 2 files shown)


Shnoulle:
Next step : move SPSS export to export plugins … really think it's more cleaner … for 3.0 because it was disable at start, then user can think it was removed.

maziminke:
SPSS export is a core feature. Besides easier code maintenance, what is the benefit of making it a plugin? I do not see a reason why someone may want to deactivate such a feature.
We could also create a single function/file/controller for it for easier maintaining the code.
If creating a plugin for it to separate the code it should always be activated by default.

Shnoulle:
@maziminke : XLS export is too a core feature BUT in a plugin … And really using Export plugin is really more cleaner than this spaghetti code with no respect of existing data … Same for R : it's in core, but in a cleaner plugin. 

I made [Export to Triple S (Survey Interchange Standard)](https://extensions.sondages.pro/export-statistics-and-database/triple-s-export/) and really more cleaner and adaptable than SPSS export.

Move it to a plugin make clean code, easy improvement, easy test, easy specific adaptation for user …

> I do not see a reason why someone may want to deactivate such a feature.

I really don't have a lot of user who use SPSS/PSPP …

tammoterhark:
All of my users user SPSS. Depends on the type of customers, apparently.

I also think that since "export data" (better wording than "export to application") in different export formats could be made in a module/plugin. First plugin for "export data" in general, plus sub-plugin per language/application: SPSS, Stata, Excel, R, etc.

tammoterhark:
I would love it if we could think of features of LimeSurvey as plugins in general. Makes it much easier to tailor an installation to customer's needs.

TonisOrmisson:
just my view. Most of the customers I see around work with SPSS. From my experiance SPSS is pretty much industrial standard in (commercial) Market research. So I also would support having this ON by default. As plugin? Why not. I Also like the modular approach.

tammoterhark:
Metadiscussion: maybe the word "plugin" is often used too narrow. I think of it as a module within LS (be it in the core distribution or not), that can be switched on/off and may have some settings. 

I often use Drupal as an example for this: it consists of modules:
* core modules that are neccessary to make it function at all
* additional modules in the core distribution (may be switched on/off at will)
* additional custom modules from third parties.

Shnoulle:
PS : i have some fix currently to do for SPSS. For information : current Export Plugin have all system needed to do exactly what you want with a clean way. Current SPSS code is dirty/spaghetti and must be reworked totally … Stata, R or Triple-S are pretty industrial too and they are plugin.

Current CSV or XLSX export are made in plugin way.

Shnoulle:
And more : with new QuestionAttribute : we can allow user to set some params for export plugin now. For example : export with 2 decimal value, export as integer etc …

Shnoulle:
OK, think i get all. Please SPSS specialist : time to revioew and test. For integer must return F8.0. or F8. only ?

tammoterhark:
As far as I know, F8 is good.

tammoterhark:
https://www.ibm.com/support/knowledgecenter/en/SSLVMB_20.0.0/com.ibm.spss.statistics.help/syn_universals_formats_numeric1.htm

tammoterhark:
So F8.0 maybe better.

Shnoulle:
It's here : https://github.com/LimeSurvey/LimeSurvey/pull/815/files#diff-491c165b0a7451a216f1c553f3bf98c9R2016 . Else need to update controller just a minute

Shnoulle:
@c-schmitz there are some SQL specific part. Can you test with MSSQL DB please ?

Shnoulle:
About [STATA export](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/plugins/ExportSTATAxml/STATAxmlWriter.php): i see only `stripTagsFull` usage (and this one can be moved elsewhere, maybe use one in the plugin ?

Shnoulle:
Some example file

- [survey_85678_newSPSS_syntax_file.sps](https://framadrop.org/r/Tim-qv0Ya2#oVSQoTN7F44uxGfJd9SAkNSlJ+BtPcQuOEUkJ1LroUY=)
- [survey_85678_oldSPSS_syntax_file.sps](https://framadrop.org/r/yOJWqDUoeL#lLnJ0Qj9sHMsz+s3EfbKFEXJtK11+O3ScxZSm1NIm/M=)
- [survey_85678_SPSS_data_file.dat](https://framadrop.org/r/za5DlDwjHg#aM7v+BigJGjPeDYsL4DiMOLLeq4qIGcgMGGiZmmftLw=)
- [survey_196567_newSPSS_syntax_file.sps](https://framadrop.org/r/JcZ5L-8U36#ullaieBigW5QWzIdmTok9ZNjZRtQc8Z6WZRpp3irpQI=)
- [survey_196567_oldSPSS_syntax_file.sps](https://framadrop.org/r/o1mFNdjk6D#AQM7sLZxzGUuULoGqnWPnXrEsDYHA44OtyqvZ/ffPdg=)
- [survey_683461_all_newSPSS_syntax_file.sps](https://framadrop.org/r/N0k4oNfOuz#B4BNxU8AmKiHvVH4KTeMB6yUikpHWMVHR98Q4VX10rE=)
- [survey_683461_all_oldSPSS_syntax_file.sps](https://framadrop.org/r/2NZHiH7gEX#D9OBs2YzUU3qR1izr4XrIt121PHrYeNnQrRHKoDkrc8=)
- [survey_683461_complete_newSPSS_syntax_file.sps](https://framadrop.org/r/3j0cZj80IE#VMUbptP379dG6sNq1iKrIzVv0SCoRkanKUOc0uTlpO4=)
- [survey_683461_complete_oldSPSS_syntax_file.sps](https://framadrop.org/r/6YZ_SeZHW1#P2max6xGLYPaf5/khT2Z8sAgHQFdRilXZZ+6rWfQOxQ=)
- [survey_683461_SPSS_data_file.dat](https://framadrop.org/r/AQryWo5plT#6V3x17aWtFKy5GC2FS4l2yW/9Ik88BadkjfLuV4L2AQ=)

PR #816, opened 2017-09-27
New feature #09235: Copy, export and import survey plugin settings

Shnoulle:
- update the XML lss file
- export using plugin name, not id
- import with plugin name find id
- must check with broken plugin
- must review Plugin::model()->rules() a little


Shnoulle:
Thanks :). No i do it for TSV export/import ;)

PR #817, opened 2017-10-02
enable video in spite of active xss filtering #112560

jackewitz:
New feature # : 12560
Link : https://bugs.limesurvey.org/view.php?id=12560
Desc :

Dear LS-Developer,

xss filtering is mandatory for us, but videos (self uploaded - YouTube is a no-go) in questions and help texts is the most requested feature at our organization.

LimeSurvey uses HtmlPurifier for xss filtering via yii-framework and the wrapper class CHhtmlPurifier.php. Unfortunately the wrapper class uses the old way to configure HtmlPurifier via an array. To enable video tag (HTML5) we must use the config-object of HtmlPurifier. The trick is:

1. change function getPurifier of CHtmlPurifier.php (line 113 - framework/web/widget) from proteced to public - so we can get to the configuration of the internal _purifier.
2. add some code to LSYii_Validators.php (130ff - application/core) - see file

My approach was to change classes from the yii-framework only minimal and add the maximum changes to the core code of LimeSurvey.

Hope, you can think about und maybe integrate it in LimeSurvey.

Best wishes .. Iver

### DIFF
--- a/application/core/LSYii_Validators.php
+++ b/application/core/LSYii_Validators.php
@@ -54,14 +54,14 @@ class LSYii_Validators extends CValidator {
             $object->$attribute=$this->xssFilter($object->$attribute);
             if($this->isUrl)
             {
-                $object->$attribute=str_replace('javascript:','',html_entity_decode($object->$attribute, ENT_QUOTES, "UTF-8")); 
+                $object->$attribute=str_replace('javascript:','',html_entity_decode($object->$attribute, ENT_QUOTES, "UTF-8"));
             }
         }
         // Note that URL checking only checks basic URL properties. As a URL can contain EM expression there needs to be a lot of freedom.
         if($this->isUrl)
         {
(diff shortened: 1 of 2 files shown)


Shnoulle:
Please : don't touch to framework but extend/add class via application/config/internal

LouisGac:
sorry, I can't merge this PR because you changed files from the framework. 
We update it regularly, so your changes would be overwritten.
If you want to extend Yii, read this:
https://stackoverflow.com/questions/23280492/override-yii-1-1-14-framework-classes-without-replacing-them

CClientScript Exemple to see how we do it in LimeSurvey case:
Original: https://github.com/LimeSurvey/LimeSurvey/blob/master/framework/web/CClientScript.php
LS Version: https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/core/LSYii_ClientScript.php

Shnoulle:
Better explanation by Louis :)

PR #818, opened 2017-10-03
Dev: Default survey language must be selectable, not fixed

TonisOrmisson:
Survey base language must be selectable not hard-fixed unchangeable by default. 
I'm not sure if I need to elaborate on that why?

also ... moved the googleanalyticsapikey defaults to Survey->init where the rest of the defaults are

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -126,8 +126,7 @@ class SurveyAdmin extends Survey_Common_Action
     */
     public function newsurvey()
     {
-        if (!Permission::model()->hasGlobalPermission('surveys','create'))
-        {
+        if (!Permission::model()->hasGlobalPermission('surveys','create')) {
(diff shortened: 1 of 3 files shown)


Shnoulle:
You mean we can not select the language when create a survey ? It's not a new feature to be able to update oSurvey->language after creation ?

TonisOrmisson:
@Shnoulle yes currently survey language is fixed to default language if you create a new survey. I am not sure about changing later. This does not bother me. Its just about creating and selecting a suitable language. One can not expect that a UI language must match all survey's base languages. EG I live in a multi-language environment and surveys with different base languages. While the preferred UI language is English (no matter what software).

Shnoulle:
Hi, then yes, yes, yes !!!! NEED it ;), **necessary** !!

Else : allopw update survey->language is something other, can be great but a new feature not necessary.

PR #823, opened 2017-10-05
Dev: TODOS replace Question->insertRecords()

TonisOrmisson:
Replaced the Question->insertRecords() with the Yii way.

Updated the Question model rules of missing attributes to be allow to push the data directly to ->attributes

### DIFF
--- a/application/controllers/admin/database.php
+++ b/application/controllers/admin/database.php
@@ -1386,24 +1386,20 @@ class database extends Survey_Common_Action
                         $oOldQuestion = Question::model()->findByPk($oldQID);
                     }
                     if (returnGlobal('copysubquestions') == 1) {
-                        $aSQIDMappings = array();
-                        $r1 = $oOldQuestion->subquestions;
-                        $aSubQuestions = $r1->readAll();
-
-                        foreach ($aSubQuestions as $qr1) {
-                            $qr1['parent_qid'] = $this->iQuestionID;
+                        $aSQIDMappings = [];
+                        foreach ($oOldQuestion->subquestions as $qr1) {
+                            $qr1->parent_qid = $this->iQuestionID;
                             $oldqid= '';
-                            if (isset($aSQIDMappings[$qr1['qid']])) {
-                                $qr1['qid'] = $aSQIDMappings[$qr1['qid']];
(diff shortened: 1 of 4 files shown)


Shnoulle:
👍 Working on TSV import : using tracevar to show error is really awfull. This disable using getErrors but more http://www.yiiframework.com/doc/api/1.1/CHtml#errorSummary-detail :)

Another solutiuin with less update (i must test it) : insertRecords can return id or object. If there are error return object, then view can use it

TonisOrmisson:
@Shnoulle thanks. i added the tracevar because it was here 
https://github.com/LimeSurvey/LimeSurvey/pull/823/files#diff-57d8ed613df5c805f7b2c706aca6ca52R363
and I did not know whether it was neccessary. 

I know about errorSummary() but its not this is a different sory. I just wanted to replace ....

Should I remove the tracevar at all then?

TonisOrmisson:
Ah ok, its only if save faile

Shnoulle:
About tracevar : yes it was here, before with current insertRecords : there are no clean solution to show the error to the user in a cleanest way (using errorSummary or do our own system). 

Then : since we don't have a system to show the error : we need something using Yii::log … and tracevar is the quickest

TonisOrmisson:
this type of logging then somewhere in LSActiveRecord->afterSave() to to cover all models similarly? Anyway I removed from here

Shnoulle:
No, i don't think : i think controller must use if(object->save) and add the error in a clean way for view :).
And currently : all error can be loggued with 
```
                'fileError' => array(
                    'class' => 'CFileLogRoute',
                    'levels' => 'info, warning, error',
                    'except' => 'exception.CHttpException.404',
                ),
```
(for example)
Maybe : send object to view. view can do 

```
if($object->hasErrors()) { 
  echo "There are error";
  echo CHtml::errorSummary($object);
}
```

TonisOrmisson:
Yes totally agree. 
In this specific case though the yii logging is not working because there is no log written :)
If it's a one-model action then errorSummary() is great, but if you are doing bulk actions then errorSummary() is no good, a custom solution is needed. 

Anyway, its not a subject for this PR, a question of general loging logic and I do not want to go into that now :)

PR #826, opened 2017-10-05
Dev: TODOS: Question: fix getOrderedSubQuestions() queries

TonisOrmisson:
cleared 2 TODOs

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -807,19 +807,17 @@ class Question extends LSActiveRecord
 
     /**
     * get subquestions fort the current question object in the right order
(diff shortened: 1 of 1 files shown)


Shnoulle:
PS  not related to this commit. But i like to have a solution here to

1. Get the records in fixed order (always : using SQL cache if available)
2. Add a method just before return available by external tool to update the order (and core do his random in it).
3. Use the new system for random and alphabetical. 
4. See how plugin can set different order.

Maybe the 1st point is not really great … but this allow to get RANDOM order of subquestions on same way for same participant in future.

PR #827, opened 2017-10-05
Dev: TODOS: SurveyLanguagesettings- remove unused getAllData()

TonisOrmisson:
also 
https://github.com/TonisOrmisson/LimeSurvey/commit/40956970073c71ca0dcf13de0a495ed448d84710#diff-7cc4f9528d95baae77c523d3532c1a19L391
Survey - Relation does not take input parameters!

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -388,7 +388,6 @@ class Survey extends LSActiveRecord
      * @param int $loginID
      * @return CActiveRecord
      *
-     * TODO: replace this by a correct relation
      */
     public function permission($loginID)
     {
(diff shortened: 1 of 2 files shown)


application/models/Survey.php
```diff
@@ -1484,6 +1483,7 @@ public function search()
      * @todo Should really be saved as three fields in the database!
      */
     public static function transcribeCaptchaOptions() {
+        // TODO POST handling should be done in controller!
```
Shnoulle: 👍

application/models/Survey.php
```diff
@@ -1524,7 +1524,7 @@ public static function transcribeCaptchaOptions() {
      * @todo Should really be saved as three fields in the database!
      */
     public static function saveTranscribeCaptchaOptions(Survey $oSurvey) {
-
+        // TODO POST handling should be done in controller!
```
Shnoulle: 👍

PR #830, opened 2017-10-05
DEV: SQL -> AR; add UserGroup->hasUser()

TonisOrmisson:
Some fixes regarding moving to AR instead of plain SQL queries. 
Ended up adding the UserGroup->hasUser() method also

### DIFF
--- a/application/core/Survey_Common_Action.php
+++ b/application/core/Survey_Common_Action.php
@@ -1282,40 +1282,23 @@ class Survey_Common_Action extends CAction
     public function _userGroupBar(array $aData)
     {
         $ugid = (isset($aData['ugid'])) ? $aData['ugid'] : 0 ;
-        if (!empty($aData['display']['menu_bars']['user_group']))
-        {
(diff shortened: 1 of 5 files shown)


Shnoulle:
👍 just one thing : i see `array_map('htmlspecialchars', $grpresult);` better because i think model must not filtering, but just validate if we filter the group name in views. I see it after merge is done :) except if you take a look ;)

TonisOrmisson:
I think the encoding should be done in views via CHtml::encode(), not in controller. By doing it in controller like that we loose the object oriented way.

TonisOrmisson:
In this case I do not see the places where the encoding would be needed. Or am I missing something ?

Shnoulle:
> I think the encoding should be done in views via CHtml::encode(), not in controller.

+1 encoding it's a view task :). 

Else : 20 caracter only … in `admin/usergroups/sa/view/ugid/1` and `/admin/usergroups/sa/view` (but here : we have an XSS in description with an old 3.0.0 beta. I check on master).

PR #832, opened 2017-10-06
removed API restriction to prevent survey ownership transfer

gekkedev:
Just thought this would be useful as we are working with a superuser anyway and there is no separate function for this.

Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -299,7 +299,6 @@ class remotecontrol_handle
     * Properties available are restricted
     * * Always
     *     * sid
-    *     * owner_id
     *     * active
     *     * language
     *     * additional_languages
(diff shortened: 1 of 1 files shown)


Shnoulle:
NOOOOOOOOOOO ! Use Persmission::model . Allowed only if you are able to do.

Shnoulle:
> we are working with a superuser anyway

False : any user can be used for API

gekkedev:
Just checked, and the unset-block is already encapsed by `if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveysettings', 'update'))`. Am I right that this is restricted enough?
I'm just suggesting that we remove this and a few other restrictions on the API to create new possibilities. For example the survey language.

Shnoulle:
Must look at GUI, but i think $aSurveyData['owner_id'] must test too if user are allowed to see the new owner_id (or must be).

But you have right on a part : we allow it to update via admin part (survey settings) => must be allowed in API on the same way.

gekkedev:
I am currently developing a small api connector for testing purposes, and with the default admin i was able to retrieve the user id on a newly created survey. Just changing it was a problem, kinda readonly.

Shnoulle:
Actually : you're right : model must control if update owner is OK for user connected :). Then 👍

gekkedev:
do I need to add anything?

PR #835, opened 2017-10-06
New feature: set_quota_properties via RemoteControl API

TonisOrmisson:
tested, works

Why? 
Eg: I have a external sample managing system using LS as data-collector interface for CATI. And quotas need to be "released" & reassigned sometimes from the external system.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2039,14 +2039,15 @@ class remotecontrol_handle
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Shnoulle:
Why not `set_quota_properties($sSessionKey, $iQuotaId, array|\struct $aQuotaData)` the can update active + limit + action + autoload_url ?

TonisOrmisson:
@Shnoulle thanks. Of course :)

I will also add a way to change the actual URL's for languagesettings

Shnoulle:
Great :).

About language_settings : like we do in survey ? I never set language settings by API, never test :)

PR #836, opened 2017-10-18
WIP: enabled optional LDAP user login via RC-API (incomplete, needs a fix)

gekkedev:


Shnoulle:
Hi,

Great idea BUY , not limited to LDAP please

1st solution : 

```
  /*
  * @access public
  * @param string $username
  * @param string $password
  * @param string $plugin
  * @return string|array
  */
  public function get_session_key($username, $password, $plugin = null)
  {
  }
```

But i nmy point of view : `_doLogin` must test ALL activated plugin, maybe with a plugin settings _Allow RC login_ (because webserver can allow RC login using user/password)

gekkedev:
I implemented your changes, the $plugin variant looks way better. **Still, I am afraid I still cannot succeed to debug & test this.**

c-schmitz:
If nobody is willing to work on this I will have to close this PR.

gekkedev:
I didn't figure out yet what is missing, but for my use-case I might develop something external to create users.

Shnoulle:
@gekkedev : can you open a feature request on bugs.limesurvey.org and assign it to me please.

Shnoulle:
@gekkedev : PS : in 2.73 current without update i use a new function for login. Then current workaround without update LS code : `get_session_key_ldap` in [RemoteControlHandler](https://framagit.org/SondagePro-LimeSurvey-plugin/extendRemoteControl/blob/master/RemoteControlHandler.php) .

But it's a great idea to have a common global way for this :).

gekkedev:
If the login by the LDAP plugin would work, we could use LDAP to autocreate the user, and then just use the LDAP username to search for the user and redeem his ID, etc.

Shnoulle:
Yes, but i don't want 2 different system for User creation and Plugin authentification : one must work with RC and GUI :)

gekkedev:
Yes, and something is different with ldapauth. It tells that username and password would be wrong, but they aren't. I remember that this also can come up when the LDAPsearch-user has a wrong password, so there might be the necessity to load up a config...sadly my limesurvey/yii understanding is too low; If i could find an example of how the ldap plugin is used when logging in, then it might be easy to port that into the api.

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/939

PR #838, opened 2017-10-21
WIP : Fixed issue #09567: no salt used for password hashing.

Shnoulle:
Dev: password_hash didn't need salt
Dev: PHP 5 >= 5.5.0, PHP 7 function
Dev: have an issue in admin/user/sa/index, must fix it before

But need to [fix User management](https://bugs.limesurvey.org/view.php?id=12814) and remove deprecated code for pgsql


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/840

PR #839, opened 2017-10-29
Merge pull request #1 from LimeSurvey/develop

ovillani:
Dev: fixed some optics. Added gsid for update from 2.72

Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:


c-schmitz:
The GSID patch is unnecessary. 
If your LimeSurvey version complains about a missing GSID field after update it is because the old model is still cached - just clearing out tmp/runtime/cache will fix this.
Unfortunately I wil not be able to accept the rest of the changes because it is in the same PR.
If you remove the unnecessary update I will have another look.

c-schmitz:
Closed due to missing fedback.

PR #849, opened 2017-11-03
DEV: Fix test class declarations invalid params

TonisOrmisson:


olleharstedt:
This can probably be closed, too.

TonisOrmisson:
why?
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/qanda_helper.php#L105

PR #850, opened 2017-11-03
DEV: tests: fix wait after submit

TonisOrmisson:
instead of directly looking for the element we click and wait until we see the element ...


olleharstedt:
Still relevant?

TonisOrmisson:
does not seem to be

PR #851, opened 2017-11-03
DEV: tests - add AdminViewsTest

TonisOrmisson:
the main idea is to have an identification tag on each page and loop all pages during tests to see if any of them throw errors (or all is fine). So if you make any changes one could easily loop through the pages to see if some errors get thrown up. 

its the start of that anyway.. 

A similar solution then later could be used also to identify pages by permission etc. 

Thanks @olleharstedt in getting set up


TonisOrmisson:
had a bit of trouble placing the class in a suite. 
This is an Acceptance /Functional test. Not an Unit test.  But the suits as they are do not have that logic implemented.  I would say it would be logical to separate the Functional and Unit tests :)

olleharstedt:
Hey, this looks great! I will have a closer look on Monday. Seems like Carsten already enabled travis CI, too! :balloon: 

No, there are no great separation of tests, only the folders for helpers, controllers, questions, ... We can discuss it in the meeting, maybe come up with something good. Louis will probably push a lot to do it the "Yii way". ;) Which makes sense for models and controllers, but a lot of LS logic is in the helpers. Well, we'll see.

TonisOrmisson:
I also like to be thinking the Yii way. It makes just so much more sense to me :)
I have been playing around with using codeception for the LS tests. Using a framework supported tests framework also kind of makes sense to me :)

Anyway I am very glad we go into testing stuff and let's discuss that ...

tests/TestBaseClassWeb.php
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Class TestBaseClassWeb
+ * this is the base class for functional tests that need browser simulation
+ * @package ls\tests
+ */
+class TestBaseClassWeb extends TestBaseClass
+{
+    /**
+     * @var int web server port
+     * TODO this should be in configuration somewhere
```
olleharstedt: Again problem with local phpunit.xml vs phpunit.xml in repository. For domain, I used command line environment. Could do the same for port of phantomjs. Too many command line arguments in the end?

TonisOrmisson:
uh I wish the  composer installed stuff would not be tracked :|

tests/TestBaseClassWeb.php
TonisOrmisson: I also kind of do not like the command line arguments. 
what about having a phpunit.xml.example and a gitignored phpunit.xml with modifiable config. This could also have the test DB stuff?

application/views/admin/token/tokensummary.php
```diff
@@ -4,6 +4,10 @@
  * @var Survey $oSurvey
  * @var array $queries
  */
+
+// DO NOT REMOVE This is for automated testing to validate we see that page
+echo viewHelper::getViewTestTag('surveyParticipantsIndex');
```
TonisOrmisson: This one is not visible to phantomjs for some reason. Some ajax thing?

olleharstedt: Don't know. Check if the web driver can dump the HTML for you.

olleharstedt:
I would prefer to have all tests inside _one_ framework, either codeception or phpunit, but not both. That's why I chose to use the Facebook web driver instead of codeception.

tests/data/views/adminSurveyViews.php
```diff
@@ -0,0 +1,45 @@
...
+    'surveyPlugins' =>['route'=>'survey/sa/rendersidemenulink/subaction/plugins/surveyid/{SID}'],
+
+
+    // going deeper
+    'surveyLogicFile' =>['route'=>'expressions/sa/survey_logic_file/sid/{SID}'],
+    // FIXME these FAIL !!
+    //'surveyResponsesIndex' =>['route'=>'responses/sa/index/surveyid/{SID}'],
+    //'surveyResponsesBrowse' =>['route'=>'responses/sa/browse/surveyid/{SID}'],
+
+    'dataEntryView' =>['route'=>'dataentry/sa/view/surveyid/{SID}'],
```
TonisOrmisson: The dataEntryView has a "division by zero" warning if I look it through browser. The test web-driver passes that and it seems that the test browser has the display_errors off or the warnings disabled. Anyway- it should not be Tests should be showing all of them. Im not sure if this is my environment config problem or something that could be configured for the testing web-driver.

application/views/admin/token/tokensummary.php
olleharstedt: Or you might have to wait until it's visible?

tests/TestBaseClass.php
```diff
@@ -11,12 +11,41 @@ class TestBaseClass extends TestCase
      */
     protected static $testHelper = null;
 
-    public function __construct()
```
TonisOrmisson: The __construct seems to mess up PHPUnits some funtionality. eg the DataProvider.

TonisOrmisson: https://stackoverflow.com/questions/10175414/phpunit-dataprovider-simply-doesnt-work

olleharstedt: Are you calling parent construct?

tests/TestBaseClassWeb.php
olleharstedt: Yes, but then it would be harder to push changes into the xml. The best would be to use _two_ phpunit.xml files, but could not find any support for that. :(

tests/data/views/adminBaseViews.php
```diff
@@ -0,0 +1,19 @@
...
+ * This contains a list of admin views that we can loop for bulk testing.
+ * Each element conntains an ID as key and the admin route that we can test
+ * // TODO currtnly these only contain pages that do not depend on opening a survey!
+ */
+
+return [
+    ['login' ,['route'=>'authentication/sa/login']],
+    ['index' ,['route'=>'index']],
+    ['createSurvey' ,['route'=>'survey/sa/newsurvey']],
+    ['listSurveys' ,['route'=>'urvey/sa/listsurvey']],
```
olleharstedt: Typo here. Didn't you test your own tests? oO

TonisOrmisson:
@olleharstedt 
> I would prefer to have all tests inside one framework, either codeception or phpunit, but not both. That's why I chose to use the Facebook web driver instead of codeception.

I'm not sure what you mean. Codeception runs on phpunit;
http://codeception.com/docs/05-UnitTests
Its a framework to make the usage of phpunit more efficient. 

What I like about the codeception idea is that it is the "official" supported way to go for yii2. 
https://github.com/yiisoft/yii2-app-advanced/blob/master/docs/guide/start-testing.md

I actually got the LS Unit tests mostly working on codeception. Since the back-end is PHPUnit anyway, its no ptoblem. But the acceptance testing is a bit different. And seems slower on my dev for some reason. 

But of course the decision is yours to make :) and I am not a testing specialist in any way

olleharstedt:
Any other reason you want to switch to Codeception, except it being standard in Yii 2?

TonisOrmisson:
no not really

olleharstedt:
Do you have a short how-to how you installed it? Composer require is enough?

olleharstedt:
Tip: The test helper can activate a survey: https://github.com/LimeSurvey/LimeSurvey/blob/develop/tests/TestHelper.php#L102

(For the admin views that require an activated survey.)

TonisOrmisson:
@olleharstedt 
I tested it over the weekend here
https://github.com/TonisOrmisson/LimeSurvey/tree/codeception
I put all the codeception stuff under tests/codeception

the composer question I can not answer. Since the lock is on higher versions than I can be, I do a number of updates to get the dependencies work for me. 

The codeception also used phantomjs webdriver. 
I got it running suite-by suite
`third_party/codeception/codeception/codecept run unit -vv`
`third_party/codeception/codeception/codecept run acceptance -vv`
 (had problems running all)

ah, and used the 'config'->urlManager->urlFormat = 'get'

TonisOrmisson:
> Tip: The test helper can activate a survey: https://github.com/LimeSurvey/LimeSurvey/blob/develop/tests/TestHelper.php#L102
(For the admin views that require an activated survey.)


Thanks! I was just thinking whether there is this. 
Although I would expect it would be in Survey Model?

Anyway I was thinking of having separate sataProviders & tests for different access type. Or status type. So a separate activatedSurveyViews or smth. 

But I will leave that for now, have some other duties also :)

PR #852, opened 2017-11-05
Render the view for any active plugin having substring AuthLDAP in the name, not only for the core AuthLDAP

digitmaster:
Fixed issue # : https://bugs.limesurvey.org/view.php?id=12869
New feature # :
Changed feature # :
Dev: 
Dev:


digitmaster:
https://bugs.limesurvey.org/view.php?id=12869

digitmaster:
Hi LS, can we discuss about this one? Whitout it, the plugin development for LDAP will be not possible  unless there is away to override this view by the developed plugin itself.
thanks.

gekkedev:
To improve it, I suggest to not define a new function (matchInstalledAndActive) but instead just storing the loop's result in a variable. It is probably not worth it as the only gets used once. What about this:
```
foreach ($haystack as $item){
    if ((strpos($item[name], $needle) !== false) && (($item[active]==1))) {
        echo "<div class=\"form-group\">";
        echo "<label  class='col-md-4 control-label'>";
        eT("Central database");
        echo "</label>";
        echo "<div class='col-md-8'>";
        echo CHtml::dropDownList('user_type',
            'DB',
            array(
                'DB' => gT("Internal database authentication",'unescaped'),
                'LDAP' => gT("LDAP authentication",'unescaped')
            ),
            array(
                'class' => ""
            )
        );
        echo "</div>";
        echo "</div>";
        break;
    }
    echo "<input type='hidden' id='user_type' name='user_type' value='DB'/>";
}
```

Shnoulle:
Think best solution is to add an event in [Permission::getGlobalBasePermissions](https://github.com/LimeSurvey/LimeSurvey/blob/b565d035ea4babd4e95c61f5651346142a69f845/application/models/Permission.php#L167) to [add new permission](https://github.com/LimeSurvey/LimeSurvey/blob/b565d035ea4babd4e95c61f5651346142a69f845/application/models/Permission.php#L265) from plugin.

Add this to AuthLDAP and AuthWebserver, review view to use a foreach (we have label etc …) to add a checkbox or not : we can use false VS null : false : checkbox default to false, null : no checkbox (and no rights).

But all of this must be in develop

digitmaster:
@gekkedev : thanks fo rthat! lovely, that was my first version, but indeed we might need to reuse the function here:
@line 33, /application/views/admin/user/setuserpermissions.php  
:-)

digitmaster:
@Shnoulle :  many thanks!  I agree, but at th emoment this one blocks the custom AuthLDAP plugin development. Right now I am too busy to jump  on **https://github.com/LimeSurvey/LimeSurvey/blob/b565d035ea4babd4e95c61f5651346142a69f845/application/models/Permission.php#L167** ...

Shnoulle:
Not need for 3.0 : https://github.com/LimeSurvey/LimeSurvey/commit/646e8bc3f13904b25f2692a135efdf9b2e013ed9#diff-f4567dff1849a50ed671994d03c036bdR151

New event globalBasePermissions

digitmaster:
Thanks Shnoulle!  You are definitely the best. The sad part is that you decided to not accept the patch for the 2.x.   
https://github.com/digitmaster/limesurveyplugins
Why?
regards

Shnoulle:
Because 3.0 is out start of 2018 … and need surely a lot f time to fix

But i think i can add same system in 2.73 if i need to start a 2.73lts. But i really dislike 2.5X version …

Shnoulle:
PS: in 2018 january : you can fork LS 2.73 without a lot of update from LS team (except security fix i hope):), then really more easy to fork it.

digitmaster:
good point but many of us (paying for confortupdate) run 2.7x. And who uses LS in Corporate Organizations will not migrate soon to 3.0. LDAP integration is important for the big players.

Shnoulle:
@LouisGac : ComfortUpdate replace only needed files ? Then there are not a lot of possibility that model was updated ?

LouisGac:
Yes CU only replace modified files.

Shnoulle:
Fixed in 3.0 
https://github.com/LimeSurvey/LimeSurvey/commit/646e8bc3f13904b25f2692a135efdf9b2e013ed9#diff-f4567dff1849a50ed671994d03c036bdR151

PR #854, opened 2017-11-07
DEV: TESTS- AdminViews2

TonisOrmisson:
I replace the #853 by this


TonisOrmisson:
@olleharstedt 
there was a problem with the test surveys not getting deleted. Went back to static tearDownAfterClass()
and all params to static

TonisOrmisson:
@olleharstedt 
Ok, now login only per class... and the setUpbeforeClass() for all set-ups
significantly faster for me :)

TonisOrmisson:
@olleharstedt 
I set up travis also for my LS

https://travis-ci.org/TonisOrmisson/LimeSurvey/builds

after also going on pre-built PHP 7.0 it runs well under 2 mins now

PR #859, opened 2017-11-10
Fixed issue #12869: Auth plugin must have a Permission settings acces…

Shnoulle:
New feature : getGlobalBasePermissions event to create new global Permission
Dev: and use it for AuthLDAP and AuthDB
Dev: Just need add the event to show it in GUI


Shnoulle:
I just wait travis and gitmate and merge

PR #860, opened 2017-11-15
DEV: remove xdebug from running tests (travis)

TonisOrmisson:
should be a bit quicker
https://docs.travis-ci.com/user/speeding-up-the-build/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ php:
     - 7.1
     - 5.6
 before_script:
-
+    # disable xdebug to speed up tests
+    - phpenv config-rm xdebug.ini
     # Install LimeSurvey.
     - phpunit --version
     - touch enabletests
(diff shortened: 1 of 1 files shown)


olleharstedt:
Do we want this? It want debugging tests harder?

TonisOrmisson:
How would you use xdebug in Travis envireonment?

the travis docs:

> PHP VM images on travis-ci.org provide several PHP versions which include XDebug. The XDebug extension is useful if you wish to generate code coverage reports in your Travis builds, but it has been shown to have a negative effect upon performance.
>
>You may wish to consider disabling the PHP XDebug extension for your builds if:
>
> you are not generating code coverage reports in your Travis tests; or
you are testing on PHP 7.0 or above and are able to use the PHP Debugger (phpdbg) which may be faster.

olleharstedt:
D'uh, didn't realize it was only for travis. OK, great!

TonisOrmisson:
just stumbled on it yesterday :)

PR #863, opened 2017-11-15
DEV: tests - GeneralSettings views (travis)

TonisOrmisson:
more tests

### DIFF
--- a/application/views/admin/homepagesettings/create.php
+++ b/application/views/admin/homepagesettings/create.php
@@ -1,6 +1,10 @@
 <?php
-/* @var $this BoxesController */
+/* @var $this AdminController */
 /* @var $model Boxes */
+
+// DO NOT REMOVE This is for automated testing to validate we see that page
+echo viewHelper::getViewTestTag('createNewBox');
+?>
 ?>
 <div class="col-lg-12 list-surveys">
 
--- a/application/views/admin/homepagesettings/index.php
+++ b/application/views/admin/homepagesettings/index.php
@@ -1,6 +1,9 @@
 <?php
-/* @var $this homepagesettings  */
+/* @var $this AdminController  */
 /* @var $dataProvider CActiveDataProvider */
+// DO NOT REMOVE This is for automated testing to validate we see that page
+echo viewHelper::getViewTestTag('homepageSettings');
+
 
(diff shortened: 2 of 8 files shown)


olleharstedt:
I wonder, what are you really testing with these tests? What kind of error would break them? Syntax errors?

olleharstedt:
Because if it's only checking for syntax error or rendering views, it would be faster to not use the web driver.

TonisOrmisson:
It will test anything that would throw an exception resulting in a page not being rendered. Syntax also. I would also run it in debug=2 to have a strict mode on syntax. But also a number of other stuff that would prevent a page from rendering. You name it. Any kind of exceptions etc. or even eg if a wrong view is linked to a wrong controller somehow etc. etc. 

The idea is that each page opens as expected for the end user with no exceptions thrown. 
It's the series of most basic acceptance tests. basically. just cklick through pages and see that the correct page opens and renders without errors. 
http://codeception.com/docs/03-AcceptanceTests.

To go further from here this type of logic could be also used to test access/permissions. Eg test if a page _will_not_ open if you try to open pages that need specific permissions. 

Some of the things for sure could be tested without web driver, but these are different type of tests. Eg Unit tests in general should be totally independent of environment etc.

olleharstedt:
But you don't actually click on anything, right? It's just direct links?

olleharstedt:
Yes, I reacted because the acceptance tests are not really deep but still takes a lot of time. Just have to live with that, I guess. :)

TonisOrmisson:
sure :)

But having the codebase we have ... I really prefer having any test possible to see if anything broke :)

olleharstedt:
Yes, just don't get a false feeling of security. Manual testing is still needed.

TonisOrmisson:
Also ... regarding the speed. I use the phpunit --filter option a lot locally to run just the things I work with. Travis runs on the background anyway :)

olleharstedt:
Yes, --filter or --group is absolutely necessary for local testing.

olleharstedt:
Btw, if you want to continue making tests, feel free to make some for question types, either one survey with all question types or many surveys with one question type each. Then you can submit with different values and check the database for correct response.

PR #891, opened 2017-12-05
Dev: queXML export honours group presentation settings

adamzammit:
Dev: queXML export honours survey group presentation settings


adamzammit:
Thanks Carsten,

This is already in master now as it was in develop before the merge so I'll close this.

Adam

PR #895, opened 2017-12-10
New feature : else part in Expression Manager function optionnable

Shnoulle:
Dev: just set default value to ""
Dev: use "" because if function is more used for show something (string) to user


maziminke:
Denis, can you explain this new feature in a sentence? Does it simply mean that I do not have to add details for the ELSE part anymore?

Shnoulle:
:) You're right …

A lot of time, i use EM if like that

`{if(QCODE=="Y","A sentence for you","")} {if(QCODE_SQ1=="Y","Another sentence for you","")}`

It's just a quick way to have `{if(QCODE=="Y","A sentence for you")} {if(QCODE_SQ1=="Y","Another sentence for you")}`

Some user on forum make error. And seems clean to have

If “condition is true” do “this”

No need the else part.

maziminke:
Thanks for the clarification, indeed a nice and very useful adjustment!

Shnoulle:
@maziminke : the n why did you close it ???? :(

c-schmitz:
@Shnoulle Please make sure to update the manual. Thank you.

maziminke:
@Shnoulle I closed it accidentally, sorry.

I also want to remind you that the manual has to be updated. If a new feature is not documented, no one will use it and that would be a pity.

PR #911, opened 2018-01-03
New feature: Import survey file via command line

TonisOrmisson:
First of all I removed some instances of accessing the session from models. The console application (as well as non-functional tests) have no access to session. 

The usage of importing surveys via command line can be useful for testing purposes. Eg. I have software that uses the LS as a third-party resource. To test my application I also need to run LS in the test environment. And to have any usage I also would need to import some surveys. Here comes the command-line import handy. 

### usage:
**Change to LS root in console**
```
$ cd /path/to/ls
```
**import file as specified user**
```
$ php application/commands/console.php import --file=tests/data/surveys/limesurvey_survey_88881.lss --userId=1
```

**If no user is set the first available superadmin user will be used.** 

**Importing using full path:**
```
$ php application/commands/console.php import --file=/home/me/limesurvey_survey_88881.lss
```

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -1176,6 +1176,7 @@ class SurveyAdmin extends Survey_Common_Action
      */
     public function copy()
     {
+        $user = User::model()->findByPk(Yii::app()->session['loginID']);
         $action = Yii::app()->request->getParam('action');
         $iSurveyID = sanitize_int(Yii::app()->request->getParam('sid'));
         $aData = [];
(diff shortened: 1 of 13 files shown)


TonisOrmisson:
Any thoughts on that one anybody?

application/controllers/admin/surveyadmin.php
```diff
@@ -1176,6 +1176,7 @@ public function editlocalsettings($iSurveyID)
      */
     public function copy()
     {
+        $user = User::model()->findByPk(Yii::app()->session['loginID']);
```
LouisGac: why not "Yii::app()->user" ?

LouisGac:
seems good to me: I could merge it just after we release today.
Have you tested it? Are you sure that normal web-import is still working fine?

c-schmitz:
Hello Tonis,

we discussed this patch here and we felt that we don't want to offer high-level functions on console level. For high level functions the RemoteControl API is the correct API to use, work, and it already exists.
Duplicating this command on console creates new point of failure, not to mention the exceptions for CConsoleProgram needed in the core. So I am sorry to say that we will not accept this PR. 
I hope you are not disappointed too much - to avoid future disappointments in the future feel free to discuss new features with us in advance, so we can find the best way of implementation. Idea: Maybe there is a way to encapsulate a JSONRPC client in a CCconsole program and so call the RC API by CConsole?

TonisOrmisson:
Reasonable actually. 
I don't know why I had the impression that RPC only allowed to copy, but not import.
We have a saying here that if the head is not working then the feet must :)

PR #912, opened 2018-01-04
DEV: change renderinf of .ls-questionhelp icon similar to ls-question-message

TonisOrmisson:
The .ls-questionhelp is very similar to  .ls-question-message in essence. They look the same basically. And I like to use them both in the same place. But. one of the icons is currently rendered via html, another via CSS. which makes it difficult to use them together. 

I swithced both to html.

### DIFF
--- a/themes/survey/bootswatch/css/theme.css
+++ b/themes/survey/bootswatch/css/theme.css
@@ -169,18 +169,8 @@ padding-top:1em;padding-bottom:1em;margin-bottom:2em;
 }
 /* little icon near the question-help */
 .ls-questionhelp{position: relative;}
-.dir-ltr .ls-questionhelp{margin-left:1.2em;}
 .dir-rtl .ls-questionhelp{margin-right:1.2em;}
 
-.ls-questionhelp:before{content:"\f059"}
-.ls-questionhelp:before {
-    display: block;
-    position:absolute;
-    font-family: FontAwesome;
-    font-size: inherit;
-    line-height: inherit;
-    height:100%;
-}
 
 /* If you want to center the icon */
 /*
--- a/themes/survey/fruity/css/theme.css
+++ b/themes/survey/fruity/css/theme.css
@@ -179,18 +179,7 @@ element.style {
 }
 /* little icon near the question-help */
 .ls-questionhelp{position: relative;}
-.dir-ltr .ls-questionhelp{margin-left:1.2em;}
 .dir-rtl .ls-questionhelp{margin-right:1.2em;}
-
-.ls-questionhelp:before{content:"\f059"}
-.ls-questionhelp:before {
-    display: block;
-    position:absolute;
-    font-family: FontAwesome;
-    font-size: inherit;
-    line-height: inherit;
-    height:100%;
-}
 
 /* If you want to center the icon */
 /*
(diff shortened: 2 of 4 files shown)


Shnoulle:
With html : it make more difficult to update view easily. Maybe it's best to do it only with css … 

Maybe best are 
`<div class="sr-only">Help for your answers : </div>{{ aQuestion.help.text | raw }}`

But in fact : question-help can be used for anything than something other than help … then with css you can easily use class attribute of question to change the icon to shown.

Unsure here … no best decision …, 2 choice maybe : 

1. put fa-icon for help and message 
2. Use css only solution for help or message

TonisOrmisson:
OK, valid points most probably. 
My only pint was that in general they look the same and seems they should also render similarly. 
If there is some good reason why they are done differently then I can just do a workaround on my custom templates. 

It just seemed not logical at first glance. 
I'll not change anything for now. Maybe someone has a good reason for that

Shnoulle:
At first : i take 1.7X template and try to have compatibility with old template (see https://github.com/LimeSurvey/LimeSurvey/blob/352c511eb5771c35cfa20fdf0d11f9fb56e6cc39/application/models/TemplateConfiguration.php#L281 ). But this compatibility system was already broken today :).

I don't say your commit is a bad idea :) , maybe best solution was to move the ls-question-message to css only system ?

I'm really unsure of the best solution, i just think core must not add HTML not needed :)

LouisGac:
I agree with Denis: the PR is good, but it should rather use CSS for both.
Thx for your work

TonisOrmisson:
I came across a problem related to this PR, so-ill post it here for discussion. 

problem: can't use Question help text area for holding any javaScript without having any visible help text content. 

we very often use the JS workarounds in questions and in order to keep the scripts separated from question text (not to end up in SPSS or other places) we place it in the help text field. very often we there is only the JS sitting there, no visible help text. 

The current vanilla theme will not render the JS in the help field if there is no visible help text 

https://github.com/LimeSurvey/LimeSurvey/blob/9c9ac0f888aec4604241b0c387c865f5a568177d/themes/survey/vanilla/views/subviews/survey/question_subviews/survey_question_help.twig#L23-L25

While having the icon as html I can do like:

https://github.com/TonisOrmisson/ls3-theme-km-base/blob/9b3b7f9bd5da9327a2280ab1b7c608062bb8f3c8/views/subviews/survey/question_subviews/survey_question_help.twig#L24-L28

which will render the js in the background, but not the help icon if no text. 

now. I can do all this in a custom template but I am thinking whether the similar use case of having only JS in help text field is a common practice and maybe should be allowed also in the default themes?

Im not suggesting it should be done like this PR I did :) maybe there is some other way to solve this (if this needs to be solved for the default themes)

Shnoulle:
It's the reason of https://www.limesurvey.org/index.php?option=com_sobipro&pid=104&sid=115:addscripttoquestion&Itemid=729 . JS must be same for all language (in a lot pof situation).

TonisOrmisson:
Good plugin. Maybe I'll try that. surely more clean & clear. but plugin is a plugin somethin extra you need to work with. 

and still - the behaviour of the software has changed. if I put something in the help text - visible or not - it should be rendered. If the JS usage is allowed and I put some JS in any of the fields, the reasonable expectation is that it should be rendered. I think. 

The question is whether its a bug or not. I would say its a bug. or?

Shnoulle:
I think it must be rendered even, if it's a space.

PR #913, opened 2018-01-04
DEV: fix question-container.row align offset

TonisOrmisson:
Fix the question row container going 15px offset left & right, see image:
by default the bootstrap .row has:

```
.row {
    margin-right: -15px;
    margin-left: -15px;
}
```
but for the questions it just goes out of align

image

### DIFF
--- a/themes/survey/vanilla/css/theme.css
+++ b/themes/survey/vanilla/css/theme.css
@@ -124,6 +124,10 @@ body {
   margin-bottom: 2em ;
   padding-bottom: 2em ;
 }
+.question-container.row {
+    margin-right: 0;
+    margin-left: 0;
+}
 
 .question-title-container {
   padding-top: 1em;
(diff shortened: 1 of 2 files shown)


Shnoulle:
It's not because row need col-XXX inside and don't have it in question ? If yes : must remove row class …

tammoterhark:
I agree with Denis, changing the standard behaviour of Bootstrap classes may give new problems later on. Better to look if HTML can be adjusted. 
Each div that has .row must have underlying col-xx-xx classes that fin in there. If these col-xx-xx classes are not present, the .row class is superfluous.

tammoterhark:
https://getbootstrap.com/docs/3.3/css/#grid

TonisOrmisson:
I agree. I actually tried removing the .row but I was not able to find a nice solution for that. 
I'm not that good into front-end. Maybe someone can find a nice solution. 
opened it here: 
https://bugs.limesurvey.org/view.php?id=13123

PR #914, opened 2018-01-05
Bugfix:Non webserver authentication.

pgarcpri:
checkPassword throwing ERROR_PASSWORD_INVALID when credentials are ok.

### DIFF
--- a/application/core/UserIdentity.php
+++ b/application/core/UserIdentity.php
@@ -33,14 +33,14 @@ class UserIdentity extends CUserIdentity
                 $this->errorCode = self::ERROR_USERNAME_INVALID;
                 return !$this->errorCode;
             }
-            
+
             if ($sOneTimePassword != '' && Yii::app()->getConfig("use_one_time_passwords") && md5($sOneTimePassword) == $user->one_time_pw) {
                 $user->one_time_pw = '';
                 $user->save();
                 $this->id = $user->uid;
                 $this->user = $user;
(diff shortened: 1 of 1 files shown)


vectorson:
More importantly, the else statement on line 45 will authenticate the user if the password is **incorrect**. Major security flaw present since 3.0.0-RC2 and this commit: https://github.com/LimeSurvey/LimeSurvey/commit/f5aa619faed7b2ed2a929228b0c93816334ed5ac#diff-81b80b77bee5b64a621cb3eee123121bR49

c-schmitz:
Good catch. Thank you very much - expect a new release today.

Shnoulle:
😮 how anybody see this before ?

TonisOrmisson:
The password comparison might be distinctively if true then authorized

Shnoulle:
Not updated LS3 : can not log in without the good password : authDB login don't have this issue . 

The issue here : why core can log in ? It's the plugin to log and control , not core …

vectorson:
Not to familiar with LS code base but it seems at least the RemoteControl API uses the method: https://github.com/LimeSurvey/LimeSurvey/blob/13b4e089b25b9a0d014e1483d79fd41b17410935/application/helpers/remotecontrol/remotecontrol_handle.php#L2979

Shnoulle:
Right with RC api , thanks @vectorson 

The fix must be applied … but maybe all this part must be removed …

And API must use AuthDB code …

PR #915, opened 2018-01-06
Fixed issue #13130: Translation method don't pass on count

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=13130


c-schmitz:
Already fixed. You can always commit small fixes like that directly.

TonisOrmisson:
I don't think I can commit to the LS repo directly

PR #917, opened 2018-01-06
DEV implement yii2-like model::findOne() instead of model::model()->findByPk()

TonisOrmisson:
Also replaced all 
`Survey::model()->findByPk($iSurveyId)` to Yii2-like `Survey::findOne($iSurveyId)` in the Models folder

The overall idea is that Yii1 end of maintenance was Dec 31, 2016. So going to yii2 eventually probably is inevitable

### DIFF
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -28,7 +28,8 @@ require_once(dirname(dirname(__FILE__)).'/helpers/globals.php');
 * @property CWebUser $user The user session information.
 * @property LSETwigViewRenderer $twigRenderer Twig rendering plugin
 * @property PluginManager $pluginManager The LimeSurvey Plugin manager
-* @property TbApi $boostrap The bootstrap renderer
+* @property TbApi $bootstrap The bootstrap renderer
+* @property CHttpSession $session The HTTP session
 * 
 */
 class LSYii_Application extends CWebApplication
--- a/application/core/LSYii_Controller.php
+++ b/application/core/LSYii_Controller.php
@@ -102,10 +102,14 @@ abstract class LSYii_Controller extends CController
         }
 
         if (ini_get("max_execution_time") < 1200) {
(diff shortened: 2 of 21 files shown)


TonisOrmisson:
http://www.yiiframework.com/doc-2.0/yii-db-baseactiverecord.html#findOne()-detail

Shnoulle:
It's to move to Yii2 after more easily ?

TonisOrmisson:
yes

olleharstedt:
> After that we will only respond to and fix security issues if any will be found until at least the end of 2018.

http://www.yiiframework.com/news/90/update-on-yii-1-1-support-and-end-of-life/

olleharstedt:
> Yii 1.1.19 is compatible with PHP 7.1 that, at the time of this writing, has an announced security support until December 1, 2019.

http://www.yiiframework.com/news/137/yii-1-1-19-is-released/

Shnoulle:
Because you call parent::findByPk : you losse the static call of findByPkCache ? No ?

We use a lot Survey::model()->findByPk and do one SQL request each time if we don't have a static.

c-schmitz:
Yii 1.1 is currently under maintenance mode. Continued support and bug fixes for this version has ended on December 31, 2016. Security fixes will be provided until December 31, 2019.

TonisOrmisson:
@Shnoulle 
I'm not sure it changes much of that since it's just a shortcut to the current usage. 
And regarding caching - I tend not to think too much on that and hope on the DB engine caching.

Shnoulle:
>  I'm not sure it changes much of that since it's just a shortcut to the current usage.

No because you use parent and not self :)

>  And regarding caching - I tend not to think too much on that and hope on the DB engine caching.

It's not caching, it static here : with previous version we put Survey::model()->fincByPk in memory 

Maybe you're right about DB caching

c-schmitz:
I don't think it makes much sense. Even when we update the code for Yii2 it is a simple seach & replace.

PR #919, opened 2018-01-07
DEV: Scrutinizer: fix spaces only

TonisOrmisson:
Scrutinizer fix spaces


TonisOrmisson:
Getting to know scrutinizer. Pushing some simple fixes.

PR #923, opened 2018-01-07
DEV: implement SurveyActivator

TonisOrmisson:
the activateSurvey() method is long and does a number of things. 
I split it up to pieces in a separate class.


application/controllers/admin/surveyadmin.php
```diff
@@ -782,7 +782,8 @@ public function activate($iSurveyID)
                 Survey::model()->resetCache(); // Make sure the saved values will be picked up
             }
 
-            $aResult = activateSurvey($iSurveyID);
+            $aResult = $surveyActivator->activate();
```
olleharstedt: Is `$aResult` ever checked?

LouisGac:
I would have not done it this way, but a new model is still better than a helper ;-) 
Is it deeply tested?

application/controllers/admin/surveyadmin.php
TonisOrmisson: I'm not sure about the controller. I was only aiming to replace the helper function with a class doing the same thing.

TonisOrmisson:
Not deeply. 
only MySql also.

Any suggestions on how you'd done it?

My aim was to create a separate class since I have a plan of testing a version with response data stored all in a  JSON field rather than table. So I can extend & customize the activator more simply.

TonisOrmisson:
depends on #925

TonisOrmisson:
Ok removed the dependency to #925 and created automated tests

olleharstedt:
I don't understand why we now, in the master branch, have both SurveyActivator *and* the old activateSurvey function?

The SurveyActivator is a good step towards putting business logic in service classes. I've made a small workshop about it in the office, you can read the PDF slides if if interested:
[cleancode_workshop.pdf](https://github.com/LimeSurvey/LimeSurvey/files/4319083/cleancode_workshop.pdf)

Shnoulle:
activateSurvey seems deleted/removed in this commit 

https://github.com/LimeSurvey/LimeSurvey/pull/923/files#diff-da74319d41b9ca5e8e86f9a11891ff6bL255

olleharstedt:
Yeah. So why is it still in the code-base? oO

Shnoulle:
And added 8 years ago ....

Shnoulle:
Removed in develop ? Merge issue develop to master ?????

olleharstedt:
Maybe. This event was never added to SurveyActivator, it seems: "afterSurveyActivate".

olleharstedt:
Fixed.

PR #925, opened 2018-01-09
DEV: Restructure tests folder

TonisOrmisson:
Discussed this also shortly with @olleharstedt . 
I restructured the tests folder to have the initial suite distinction depending on the test type (eg unit, acceptance, functional)

I reviewed the tests and divided them between the suites according to my best knowledge. Any change suggestions are welcome. 

Added the readme's and used some of the descritions from here:
https://stackoverflow.com/questions/4904096/whats-the-difference-between-unit-functional-acceptance-and-integration-test


olleharstedt:
You want to merge this to branch `develop`? Why not `master`?

olleharstedt:
Do we need a `data` folder if we already have `resources`?

olleharstedt:
Do we need folders for both acceptance tests and integration tests? Not really clear to me what the difference would be from LimeSurvey perspective.

Otherwise, it looks good! Better than before, for sure. (Y)

olleharstedt:
If you want to test a model (say, TemplateConfig), where would you put that test?

TonisOrmisson:
@olleharstedt I will push it to master. 

removed the data folder. I was also thinking about it but I hesitated about it. 
removed the integration tests suite. I'm also not sure. If we consider Admin & taking survey as separate modules there might it could be something between them but I'm also not sure we have something that could fit there. We can always add it later

olleharstedt:
We could separate functional testing in backend and frontend?

PR #927, opened 2018-01-09
DEV: Restructure tests folder

TonisOrmisson:
#925 
moved to master


TonisOrmisson:
I also changed the tests namespace from ls to LimeSurvey. It seems this one is used elsewhere

olleharstedt:
OK, so my understanding of the categories is like this:

1. Acceptance = anything that uses Selenium/headless browser, and/or writes to files/database
2. Functional = anything that writes to files/database but does NOT use Selenium
3. Unit = anything that does NOT write to files/database NOR uses Selenium

Correct?

TonisOrmisson:
I understand it pretty much the same.

olleharstedt:
OK, then I'll merge.

PR #932, opened 2018-01-10
removed survey language switching prevention in the API (only additional languages)

gekkedev:
It is not necessary to block language switching as a survey owner could do this by the webinterface, too.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -355,7 +355,7 @@ class remotecontrol_handle
      * * Always
      *     * sid
      *     * active
-     *     * additional_languages
+     *     * language
      * * If survey is active
      *     * anonymized
      *     * datestamp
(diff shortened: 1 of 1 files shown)


TonisOrmisson:
Main language can not be changed after the survey has been created

gekkedev:
True, thought that was the same as additional_languages, those are changeable though, but I adjusted the pull request now. To conclude, the survey attribute ```language``` should not be changeable, ```additional_languages``` should be.
Not sure if the API should support it, but we could split the languages, since all of them are written into one string, separated by spaces.

c-schmitz:
When you add/or delete a language from a survey the integrity of the survey must be checked and fixed (question groups/questions/answers etc.) need to be created/deleted for all language if they are missing/obsolete-
This is all missing currently. Please complete the patch accordingly.

gekkedev:
Sounds like we could as well write separate methods for language deletion and creation.

gekkedev:
Too busy, sorry. Feel free to continue or close it.

PR #938, opened 2018-01-15
removed restriction regarding survey enabling/disabling via rpc-api

gekkedev:
This should not cause any integrity failure when disabling an active survey, but could someone more experienced with Limesurvey's structure take a look at it?
Enabling a survey could cause trouble indeed since the survey's own table does not get created this way, but when used only for disabling it should be fine, right? For that case we could change the restriction to only allow disabling, but not enabling.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -384,6 +384,7 @@ class remotecontrol_handle
             if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveysettings', 'update')) {
                 // Remove fields that may not be modified
                 unset($aSurveyData['sid']);
+                if (@$aSurveyData['active'] == 'Y') unset($aSurveyData['active']);
                 unset($aSurveyData['language']);
                 unset($aSurveyData['additional_languages']);
                 // Remove invalid fields
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -354,7 +354,6 @@ class remotecontrol_handle
      * Properties available are restricted
      * * Always
      *     * sid
-     *     * active
      *     * language
      *     * additional_languages
      * * If survey is active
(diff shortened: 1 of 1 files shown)


gekkedev:
I've added a second commit for the mentioned alternative solution, so the preferred commit can be cherry-picked/merged.

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -385,8 +384,7 @@ public function set_survey_properties($sSessionKey, $iSurveyID, $aSurveyData)
             if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveysettings', 'update')) {
                 // Remove fields that may not be modified
                 unset($aSurveyData['sid']);
-                //unset($aSurveyData['owner_id']);
-                unset($aSurveyData['active']);
+                if (@$aSurveyData['active'] == 'Y') unset($aSurveyData['active']);
```
LouisGac: IMHO, the stfu operator should be avoided as much as possible.
is there any reason why you use it?

gekkedev: My clumsy way to keep it shorter than ```if (isset($aSurveyData['active']) && $aSurveyData['active'] == 'Y')```

gekkedev: And eventually these conditions might need to be recursive if done properly.

c-schmitz:
There are a number of reasons why a survey needs to be activated/deactivated. - mostly because the response table structure changes when you change certain survey settings/question/question groups/answer. So this patch will not be accepted.

gekkedev:
@c-schmitz could you point me into the right direction what else has to be done when disabling a survey? I think setting the ```active``` to ```N``` property should not harm the database's integrity. It's not a nice workaround to modify other attributes like e.g. setting the expiration date to a past day.
To test my idea, I've disabled a survey via phpmyadmin and ran a database integrity check, which does not find anything to complain about and amongst other things says:
> All answers meet consistency standards.
> All surveys meet consistency standards.
> All survey language settings meet consistency standards.
> All questions meet consistency standards.
It might sound stupid as there could still be unexpected behavior, but you'll probably know better.

Shnoulle:
See : http://api.limesurvey.org/classes/remotecontrol_handle.html#method_activate_survey

When deactivate : rename token + survey table

PR #940, opened 2018-01-18
Validation settings for "array by column" and others

Shnoulle:
[New feature #12240: Validation settings for "array by column" question type](https://bugs.limesurvey.org/view.php?id=12240)
New feature : Add `em_validation_q` to list of choice 

The only issue are name of var $rowdivid for column . I can update and name it $notrowdividButHaveSamePurpose ;)


Shnoulle:
For single choice ( ! and L) tested with `this=="A1" || Q00_other.NAOK == "A1"` .
Current test here : https://develop.sondages.pro/index.php/546711?newtest=Y&lang=en

`this` is set only to Q00 

I think there are another issue : because in question text i have `Q00_other.NAOK` and it's not updated

Shnoulle:
Found the issue :) . Fix it

PR #941, opened 2018-01-19
Fixed issue # : 13190 "Remove text or uncheck checkbox automatically" not work

lokiahk:
Fixed issue # : 13190 "Remove text or uncheck checkbox automatically" not work


lokiahk:
@Shnoulle Changed as your comments :)

PR #943, opened 2018-01-19
New feature : em_validation_q for 5 point, Language, Gender, Yes/No, …

Shnoulle:
Dev: All question have it except Array (dual scale) : EM subqs is broken
Dev: T/H/U can be moved the same way than 5,G,I, * …
Dev: subqs is used for single choice "array filter" … but it's not subquestions here …

Tested all question with : another question AND `this` condition. But need to add all question in EM->q2subqInfo …


c-schmitz:
Please target this PR to develop branch instead.

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/980 i just wait for scrutinizer

PR #945, opened 2018-01-21
Add support of CC and BCC into SendEmailMessage Helper

alexandcote:
Support CC and BCC into SendEmailMessage Helper

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -1943,7 +1943,7 @@ function javascriptEscape($str, $strip_tags = false, $htmldecode = false)
 * @param mixed $attachments
 * @return bool If successful returns true
 */
-function SendEmailMessage($body, $subject, $to, $from, $sitename, $ishtml = false, $bouncemail = null, $attachments = null, $customheaders = "")
+function SendEmailMessage($body, $subject, $to, $from, $sitename, $ishtml = false, $bouncemail = null, $attachments = null, $customheaders = "", $cc = array(), $bcc = array())
 {
     global $maildebug, $maildebugbody;
     require_once(APPPATH.'/third_party/html2text/src/Html2Text.php');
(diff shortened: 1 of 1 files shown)


c-schmitz:
Same here: please undo the .gitignore changes.

.gitignore
```diff
@@ -4,6 +4,9 @@
 # ignore CAssetManager for debugging
 /framework/web/CAssetManager.php
 
+# vendor directory
```
alexandcote: I add this because if you use composer to manage your plugins, you will have a `vendor` folder to ignore.

.gitignore
```diff
@@ -79,12 +82,6 @@ Thumbs.db
 .buildpath
 *.un~
 launch.json
-# ignore plugins directory by default except LS plugin
```
alexandcote: This was a duplicated entry in the `.gitignore` file. You will find the same code [here](https://github.com/LimeSurvey/LimeSurvey/pull/945/files#diff-a084b794bc0759e7a6b77810e01874f2L29)

.gitignore
TonisOrmisson: vendor dir is third-party
https://github.com/LimeSurvey/LimeSurvey/blob/a1cd6d20c7a5bd22b90cb8c4c3680fa432e18fc5/composer.json#L34

alexandcote: I delete this entry in the `.gitignore` file.

composer.json
```diff
@@ -3,14 +3,15 @@
 	"description" : "The most popular FOSS online survey tool on the web",
 	"homepage" : "https://www.limesurvey.org/",
 	"license" : "GPL 2.0",
+	"type": "limesurvey",
```
TonisOrmisson: https://getcomposer.org/doc/04-schema.md#type

Shnoulle:
Think we must create a new SendMail Class and add an event in this one. Global SendMail event for ALL event.

The hard point are : what information this event need (more than to/from/subject/body … 

1. Mail was sent to 'admin'/'particpant'/'token'
2. This mail was 'type' email (password for admin, invite for token etc …

See [eventBeforeTokenEmailExtended](https://framagit.org/SondagePro-LimeSurvey-plugin/eventBeforeTokenEmailExtended) this one already add CC and BCC to Token email, but not for admin and other email.

composer.json
c-schmitz: ?

alexandcote: It was a mistake... sorry about that

c-schmitz:
This needs to go to the develop branch.

PR #946, opened 2018-01-21
DEV: fix failing tests (missing upload/themes/survey folder)

TonisOrmisson:
removing the index.html files from upload/themes & upload/themes/survey caused the global settings to not open with no upload/themes/survey folder


TonisOrmisson:
back in green zone

PR #950, opened 2018-01-22
DEV: properly fix 13161

TonisOrmisson:
we get only one record. Thanks @c-schmitz


Shnoulle:
Or : 
`SavedControl::model()->deleteAll($criteria);`

Don't need search, because we don't use it ;)

c-schmitz:
Note: Careful with deleteAll() in general as it does not  properly run the  onBeforeDelete, onAfterDelete events of the model.
In this case it would be ok, though. But in general it is better to use delete()

Shnoulle:
Ah right :).

But if we do a find + a delete : there are 2 action on DB, and one is uneeded …

PR #963, opened 2018-01-24
New feature: allow setting RPC mode during console install

TonisOrmisson:
Optionally one can turn on the RPC mode during command line installation. For me - to have an automated test-instance I can then further manipulate further via my intregration apps/ app tests. 

Alternatively a separate console command for that, maybe even more reasonable?


TonisOrmisson:
notes: The Global Settings needs to be in a separate model. I will propose a solution for that (have a similar thing elsewhere already)

LouisGac:
"The Global Settings needs to be in a separate model": be careful. 
One object => One model.
So global settings could fall on different models. 

But at least, one model for the table setting would be a good start.

Shnoulle:
I'm for «Update any global settings» via command :). 

And maybe on a separate core plugin :  [direct (command)](https://manual.limesurvey.org/Direct_(command)) is done for such need.

TonisOrmisson:
Im also for updating any global setting via command. But even more so the globalSettings should be a separate model. Im not sure I understand you right :) but what I mean is that the collection of values inside the GlobalSetting (singular) compose a Model GlobalSettings (plural) where each attribute describes one record in the GlobalSetting table. This way the Settings collection will be nicely described, documented, is available to validate (model rules) etc.

LouisGac:
a model for GlobalSettings table with methods instead of the mess of database.php: sure.
but, not all the datas from "global settings"  form in admin UI are stored in this table.

Shnoulle:
Yes @LouisGac have right. We have some config.php GlobalSetting only. Maybe some other :).

My sentence was not about Model or not Model ;)

TonisOrmisson:
What i meant was something like that ...
https://github.com/LimeSurvey/LimeSurvey/compare/develop...TonisOrmisson:global-settings?expand=1

This way you can define for each setting:
 - validation rules (type, possible value etc)
 - descriptions, docs, labels
 - types

I understand that some settings are elsewhere. im not sure whether there is a clear reason or whether some of it could be integrated into the table. I was just aiming to make the setting table values make sense :) so that you have a clear reference point eg if i want to manipulate settings from some other place than the UI (eg the console) - and not duplicate the labels, logic, validation etc etc.

Shnoulle:
About "some settings are not in table", last example : i reallly think "super-super-admin" must not be in DB, but only in a php file :).

We can move it to another model if needed 'FixedGlobalSettings'

TonisOrmisson:
The note initially was not about moving settings from one media to another. It was just about defining a proper model for the values stored in table :)

c-schmitz:
This patch will not be accepted. If you want to activate RPC installation, you can do so in your config.file by overriding the default param from default-config.php..

TonisOrmisson:
ok, but the config.php is created after command-line installation:

https://github.com/LimeSurvey/LimeSurvey/blob/367ece0bc6ddff6974ad9acbc3d370e97dc62dba/.travis.yml#L29-L30

and having the `'RPCInterface'` in config.php later will not override?
or am I doing something wrong here?

c-schmitz:
Usually you would place the config.php first and then do the installation.
If you an installation without config.php the config-sample-mysql.php is used.
Setting the default on config.php will not set a value in the database btw, it will only be the default setting.
So, if someone enters global settings then RPC will be activated (and will also be function-wise before that). After that it can be still overridden by saving the global settings.

TonisOrmisson:
ok, good to know, thanks! I'll test that

PR #964, opened 2018-01-24
DEV: replace survey 'Y' with booleanAttributes

TonisOrmisson:

PR #965, opened 2018-01-24
Revert "DEV: replace survey 'Y' with booleanAttributes"

TonisOrmisson:
Reverts LimeSurvey/LimeSurvey#964


Shnoulle:
Think you can just fix it quickly or there are another issue ?

PR #970, opened 2018-01-29
DEV: remove the model->insertRecords() methods 1

TonisOrmisson:
does not make sense
```
      foreach ($data as $k => $v) {
                    $model->$k = $v;
        }
```
there is the `$model->attributes = $array`
I'll add the `@deprecated` with a date and refactor all current usages i find

### DIFF
--- a/application/controllers/admin/database.php
+++ b/application/controllers/admin/database.php
@@ -1321,15 +1321,12 @@ class database extends Survey_Common_Action
                         $r1 = Answer::model()->getAnswers((int) returnGlobal('oldqid'));
                         $aAnswerOptions = $r1->readAll();
                         foreach ($aAnswerOptions as $qr1) {
-                            Answer::model()->insertRecords(array(
-                                'qid' => $this->iQuestionID,
-                                'code' => $qr1['code'],
-                                'answer' => $qr1['answer'],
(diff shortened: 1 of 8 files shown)


Shnoulle:
Yep, and after we can show clean error when import bad lss :) :+1:

TonisOrmisson:
Talking off clean errors. 
I have seen a number of error throwing / logging versions. Can anybody say what is the suggested way of handling errors? Do we have one?

Shnoulle:
When importing : i think we must never throw error , but always send warning or error to result.
If result have error : delete the survey, with warning : just show the error. But we have 2 import system : one need array (return json) other need html :). 

Then don't know if we return $model->getErrors or html Error :)

TonisOrmisson:
regarding the errors: I was just eliminating the insertRecords() and kept all the error management as it was originally. If would say the error management change should be looked separately. or it there is very clear suggestion on how to change it? Otherwise i would commit

Shnoulle:
You totally right :) to really fix lss import we need

1. Export a clean simple lss
2. Update it manually to broke import
3. See what happen with THIS hack : did survey is OK ? Can be activated ? Did we just need to inform the user of something not imported etc …

PR #976, opened 2018-01-30
Feature: Survey default configs

kw-pr:
Let you set some default configs for new surveys, so you don't have to change them every time you create a new survey. Also contains a default for "mandatory" setting in new questions, so you can make new questions always as mandatory if you like. 
The config file in this push still has the old defaults values.

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -94,6 +94,12 @@ $config['showpopups']         = 1; // Show popup messages if mandatory or condit
 
 $config['maxemails']          = 50; // The maximum number of emails to send in one go (this is to prevent your mail server or script from timeouting when sending mass mail)
 
+// Survey default settings
+$config['survey_allowprev'] = 'N';
+$config['survey_showxquestions'] = 'Y';
+$config['survey_tokenanswerspersistence'] = 'N';
+$config['survey_sendconfirmation'] = 'Y';
+
 // Experimental parameters, only change if you know what you're doing
 //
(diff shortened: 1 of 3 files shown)


TonisOrmisson:
The idea to have a set of default survey setting is something I also have had for some time. In general I support the idea very much. I would myself have wanted to have this in DB & global settings somewhere so that it would be changeable via UI. 

The problem with this specific PR is that the config-defaults is not user-changable. config-defaults is tracked by git and will be overwritten with updates.

kw-pr:
@TonisOrmisson Yes, having this in the DB would be nice but I did not want to make my first pull request too big. The `getGlobalSetting()` I used can also read the same setting from the DB. So it is possible to move the option there later. For now you can overwrite the default in the config section of `config.php`. That should not be overwritten by any update.

Shnoulle:
getGlobalSettings must be replace by Yii:app()->getConfig … BUT default value must come from model, not frm controller. Else you have difference between default value from RC and GUI.

And more : why not all other value ? And if we add a new settings ? Need a global solution …

Shnoulle:
Oups, sorry already in model

c-schmitz:
I am sorry but this implementation is merely a hack. There will be a proper implementation done by us which is based on survey groups in the very near future.

maziminke:
I am looking forward to having some survey settings predefined at global settings or survey group setting. For systems with a large user base this will help a lot.

PR #979, opened 2018-02-01
Fixed issue #13283: SurveyGroups Title should be mandatory (used in dropdowns)

TonisOrmisson:


TonisOrmisson:
Can anybody explain why we need both name & title for groups. Wouln't just one of them do the job?
if we need both there should be hints in the form explaining the difference, otherwise its confusing.

LouisGac:
Title: shown to user (can contain spaces and special chars, be duplicate, etc)
Name: for internal references ( no spaces, no special char, uniq, etc )

TonisOrmisson:
This is what I thought, but I can not see any reasonable real life scenario for the actual usage of the name. I do not find any in the current code also that would be really necessary. 

Internally I would use the id instead. For me personally it seems like un-necessary complexity currently. 
:)

Shnoulle:
See all old issue's with TSV import and multiple language, and think «Why we don't add code/name for group before»

But i prefer code than name ;)

LouisGac:
using the auto-incremented ID for foreign in a context where you can do some import/export leads to so many problems...  

But you're right: it's not used at all for now in the code. Soon or later we'll need to be able to export survey groups, with their survey themes/survey settings configurations, with their "Survey Template", etc

PR #981, opened 2018-02-01
minor RPC-API method changes

gekkedev:
(one more Attribute; indentation and comment adjustments)


Shnoulle:
I din't understand the list users with only one user ?
If need information on a specific user : maybe create your own function ? 

It's easy : https://framagit.org/SondagePro-LimeSurvey-plugin/extendRemoteControl

gekkedev:
And I don't know if I understood your question correctly, but if you're asking why single user results are in an array too, that is because the search result can be empty in which case there is still something being returned, therefore validation is necessary.

PR #984, opened 2018-02-01
New feature: Let user set default settings for statistics export in c…

kw-pr:
…onfig.php

I think this would be useful. If wanted I can also add the other options to make the pull request more complete.

### DIFF
--- a/application/views/admin/export/statistics_subviews/_dataselection.php
+++ b/application/views/admin/export/statistics_subviews/_dataselection.php
@@ -28,7 +28,7 @@
             </div>
 
             <div class='form-group col-sm-6'>
-                <?php $sNoncompleted = (int) Yii::app()->request->getPost('noncompleted');?>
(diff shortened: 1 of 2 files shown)


LouisGac:
Yes, very good idea! Thank you. 
But please, submit it to dev branch. 
(it will be merge with master)

kw-pr:
@LouisGac Thanks for the answer!
I changed the commit base to develop.

c-schmitz:
Where can these new settings be set? I am missing some kind of configuration & documentation here.

kw-pr:
The settings as called statistics_noncompleted, statistics_showtextinline, statistics_usegraph. They can have the value 0 or 1. They can be set in application/config/config.php in the 'config' section.

I could add documentation in config-defaults.php or where would be a good place for that?

maziminke:
Please also update/extend the manual if this gets added to the core!

PR #985, opened 2018-02-02
Fixed issue #13001: Expression Manager evaluates answers to free text…

Shnoulle:
… questions enclosed by "{" and "}"
Dev: replace with encoded value : then EM don't update it
Dev: but HTML show it good (in source, shown encoded)

### DIFF
--- a/application/helpers/qanda_helper.php
+++ b/application/helpers/qanda_helper.php
@@ -3264,9 +3264,9 @@ function do_shortfreetext($ia)
         $answer = doRender('/survey/questions/answer/shortfreetext/location_mapservice/item_100', $itemDatas, true);
     } else {
         //no question attribute set, use common input text field
-        $value = $_SESSION['survey_'.Yii::app()->getConfig('surveyID')][$ia[1]];
+        $dispVal = $_SESSION['survey_'.Yii::app()->getConfig('surveyID')][$ia[1]];
         if ($aQuestionAttributes['numbers_only']) {
-            $value = str_replace('.', $sSeparator, $dispVal);
+            $dispVal = str_replace('.', $sSeparator, $dispVal);
         }
         $itemDatas = array(
             'value' => $value,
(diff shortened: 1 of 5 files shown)


Shnoulle:
@LouisGac : my first idea was to use `views/survey/questions/global/input/textItem` and `views/survey/questions/global/input/textarea` but:

- Using Chtml : this encode 2 times, then final have `&amp;#123;5+5&amp;#125;` for 5+5
- Using twig : we can have a lot of htmlOptions, with a lot of different attribute

But i don't like to use `echo str_replace(array("{","}"),array("&#123;","&#125;"),$value);` in all views (or twig version).

Maybe there are a better place for function ? A twig extended function ? 

But : with this version, we can more easily update to a twig extended function after.

Shnoulle:
PS: if OK : i do for whole text input in answers

LouisGac:
shnoulle, maybe I misunderstood, or maybe Dev branch is not sync with master, but I don't think this PR is necessary.

Please, have a look to this bug: 
https://bugs.limesurvey.org/view.php?id=13156

and this commit:
https://github.com/LimeSurvey/LimeSurvey/commit/cd2341a82471975d376abf0c00db2612fd16eca2

Here I gave some explanations to Markus:
https://bugs.limesurvey.org/view.php?id=13156

Could you tell me if the same kind of operations could solve #13001 ?

Shnoulle:
Totally unrelated : here it's USER entered value : if user enter {5+5} in a text input, EM do the job and replace it by 10 

I'm sure at 99% you can reproduce the issue in 2.6lts, 2.72, 3.1.1master and 3.1.1develop

See the issue : https://bugs.limesurvey.org/view.php?id=13001

LouisGac:
ok :laughing: I felt that I was misunderstanding your PR 
so is it like a new feature indeed?

Shnoulle:
No, it's an issue … public user (participant) enter {10+10} in a text input , we MUST show {10+10} (we save {10+10} in session i think (not verified)).

Let me check the solution proposed by the reporter [here](https://bugs.limesurvey.org/view.php?id=13001#c45272) but if i remind we **need** it for registering too, then same solution for all views seem better.

I have an issue with default value currently : i must test it work OK with default value (if admin put {10+10} in default value : it must be 20 for default value)

PR #986, opened 2018-02-02
Fix issue: Disable sorting in left box of ranking question.

kw-pr:
DEV: Some users where confused by this. They sorted the left box and did not understand why the required question is not accepted.
This disables sorting in the left box. Sorting into the right box still works. Also you can still remove items from the right box back into the left.


Shnoulle:
Oh yes, seems a good idea :) more clear

PR #990, opened 2018-02-03
DEV: rename insertRecords() to singular insertRecord() for dynamic models

TonisOrmisson:
It seems to me that the insertRecord method has its right place for the dynamic models. 

I changed the name to singular because
 - it describes more accurately what it does (inserts ONE record)
 - and its easier to deprecate /remove the insertRecords() for other models this way

### DIFF
--- a/application/controllers/admin/database.php
+++ b/application/controllers/admin/database.php
@@ -145,14 +145,16 @@ class database extends Survey_Common_Action
     public function _updateDefaultValues($qid, $sqid, $scale_id, $specialtype, $language, $defaultvalue)
     {
         if ($defaultvalue == '') {
-// Remove the default value if it is empty
+            // Remove the default value if it is empty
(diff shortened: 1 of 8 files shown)


Shnoulle:
My issue with insertRecord(s) is : we don't have a clean way to have the model->getErrors , else : why not ;)

TonisOrmisson:
this one is only about renaming. You'd prefer to kill the method at all for dynamic models?

Shnoulle:
No, no  :) , and in fact it's more for other function like this one:

Idea:
```
    public function insertRecord($data)
    {
        $record = new self;
        foreach ($data as $k => $v) {
            $search = array('`', "'");
            $k = str_replace($search, '', $k);
            $v = str_replace($search, '', $v);
            $record->$k = $v;
        }
        $record->save();
        return $record;
    }
```

Usage

```
  $oModel = Model::model()->insertRecords();
  if(!$oModel) {
      return $oModel->getErrors(); // Or fill an array, or show the errors and quit or another action
  }
  $iModel = $oModel->getPrimaryKey(); // or $iModel = $oModel->id;
  // Do action
```

PR #994, opened 2018-02-05
DEV: Answer/question  l10n model rules & phpdoc

TonisOrmisson:


application/models/QuestionL10n.php
```diff
@@ -52,7 +53,8 @@ public static function model($class = __CLASS__)
     public function relations()
     {
         return array(
-            'question' => array(self::BELONGS_TO, 'Question', 'qid'),
+            // FIXME this conflicts with the attribute "question"
+            //'question' => array(self::BELONGS_TO, 'Question', 'qid'),
```
TonisOrmisson: There is still a conflict here. 
$model->question refers to the question text, the oQuestion relation is not reached. 
It would be logical in my view to rename the question attribute to something like "name", "text" etc not using the model name.

Same goes for the answer :)

Shnoulle: text seems better than name :)

PR #995, opened 2018-02-07
New Feature: Additional mail header for bounce mails

kw-pr:
The email bounces we get do not contain the original header. Because of this the survey related headers (X-surveyid, X-tokenid) are missing and the bounces can not be processed. We arranged with t-systems that they include the full header if we send an additional email header.
This pull requests allows to add a custom header email header from globale settings, which fixes the issue for us. Might be useful for others too.

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -666,6 +666,7 @@ $config['bounceaccounthost'] = '';
 $config['bounceaccounttype'] = 'off';
 $config['bounceencryption'] = 'off';
 $config['bounceaccountuser'] = '';
+$config['bouncemailheader'] = '';
 
 // Question selector
 $config['defaultquestionselectormode'] = 'default';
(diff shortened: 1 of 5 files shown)


LouisGac:
thank you very much
could you please rather submit this PR to the dev branch?

kw-pr:
@LouisGac Sure, no problem.

Shnoulle:
I'm totally against this pull request. allow any header in event is better. (and any count of header)

kw-pr:
@Shnoulle You mean something like this for the `beforeTokenEmail` Event:

```
$event = new PluginEvent('beforeTokenEmail');
...
$event->set('customheaders', $customheaders);
$customheaders = $event->get('customheaders');
```

Shnoulle:
https://gitlab.com/SondagesPro/eventBeforeTokenEmailExtended/blob/master/eventBeforeTokenEmailExtended.php#L73
and
https://gitlab.com/SondagesPro/eventBeforeTokenEmailExtended/blob/master/eventBeforeTokenEmailExtended.php#L119

kw-pr:
@Shnoulle Yes, that is much better!

PR #996, opened 2018-02-07
New Feature: Add CustomHeaders with "beforeTokenEmail" Event

kw-pr:
As suggested by @Shnoulle in PR #995

### DIFF
--- a/application/controllers/RegisterController.php
+++ b/application/controllers/RegisterController.php
@@ -330,6 +330,7 @@ class RegisterController extends LSYii_Controller
         $sTo = $event->get('to');
         $sFrom = $event->get('from');
         $sBounce = $event->get('bounce');
+        $customheaders = (array) $event->get('customHeaders', array());
 
         $aRelevantAttachments = array();
         if (isset($aSurveyInfo['attachments'])) {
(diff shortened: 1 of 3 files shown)


Shnoulle:
Yes but here : it's only for email send by GUI and for invitation/reminder. Not for confirmation or register …

It's the reason i think we must rewrite this part before adding new system.

With your solution, i still using my plugin because it's done for the 4th event.

kw-pr:
@Shnoulle That's right. I added the two missing event cases. Also I changed the parameter name from `CustomHeaders` to `customHeaders`. That is more consistent with the other event return values.

What about [token_helper.php](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/admin/token_helper.php). Is the beforeTokenEmail event missing here?

Shnoulle:
And we need again to add each event part 5 times … ;) . Really more work, but really more clean.

kw-pr:
Why 5 times? In token_helper.php is only one SendEmailMessage for all emails. Should be enough to add the beforeTokenEmail event once. There is already a `type` parameter (for invite, remind, register, confirm).

Maybe there should be another parameter for the event like `isRemote` to detect the difference.

Shnoulle:
Fixed in https://github.com/LimeSurvey/LimeSurvey/pull/1246 (do anything with phpMailer ;) ) on ALL email sent.

Shnoulle:
Fixed in develop https://github.com/LimeSurvey/LimeSurvey/pull/1246

PR #998, opened 2018-02-08
Fix AuthLDAP: make multiple search bases work again.

ecraven:
Currently AuthLDAP does *not* work with multiple search bases.

This commit fixes the typo on account creation and searches through all
base DNs for login.

We use multiple base DNs, I'm looking into upgrading to a current version, but somewhere in between the one we use and current head this broke.


Shnoulle:
Maybe fix var names
```
$aUserSearchBases = explode(";",$this->get('usersearchbase'));
…
foreach($aUserSearchBases as $userSearchBase) {
```

c-schmitz:
Thank you!

PR #999, opened 2018-02-08
Deactivate, deactivedebugging.js

xvilo:
Not sure why this is. This is horrible and not even the correct way to disable debugging. Other scripts would not work for example the Facebook tracking pixel. It tries to bind on an undefined function for example. 

Dev: @xvilo

### DIFF
--- a/assets/scripts/deactivatedebug.js
+++ b/assets/scripts/deactivatedebug.js
@@ -1,6 +1,0 @@
-var dummyConsole = {
-    log : function(){},
-    error : function(){}
-};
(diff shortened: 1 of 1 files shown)


Shnoulle:
I like this too ;)

lacrioque:
This is obsolete, since LS3 will use a custom logging system.

xvilo:
@lacrioque why would this be obsolete? It's a redicoulse practice to just override native browser functions. This breaks a lot of other stuff.

PR #1000, opened 2018-02-08
Reflect all available console functions in deactivatedebug.js

DonMul:
Some scripts make use of other console functions than ` log` and `error`. These scripts kinda depend on those other functions to exist (like `console.warn` in the Facebook tracking pixel script) and therefor these other scripts will break if those functions don't exist.

Since the idea is to completely disable the `console` when debugging is disabled, the entire console should me mocked.

### DIFF
--- a/assets/scripts/deactivatedebug.js
+++ b/assets/scripts/deactivatedebug.js
@@ -1,6 +1,8 @@
-var dummyConsole = {
-    log : function(){},
-    error : function(){}
-};
+var dummyConsole = {};
+var realConsole = console || window.console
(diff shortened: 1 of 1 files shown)


xvilo:
++

Shnoulle:
@lacrioque think it must be merged ? Since adding some script in template can be done …

xvilo:
Can we merge this @lacrioque ?

xvilo:
Obsolete since we have https://github.com/LimeSurvey/LimeSurvey/pull/999 now.

PR #1001, opened 2018-02-08
Fix : Equation question reference not directly updated 

Shnoulle:
- Fixed issue [#13317](https://bugs.limesurvey.org/view.php?id=13317): Equation question reference not directly updated 
- Move em_javascript to own package (allow more easily inclusion and dependance)
    - Unsure for insideClass var name, if someone have a better idea …
    - but clearly coreClass must wrap answer part (like other item)


Shnoulle:
@olleharstedt since tests is out of git : i can update it. But test need update [here](https://github.com/LimeSurvey/LimeSurvey/blob/master/tests/helpers/ExpressionCoreAux.php#L153)
`./assets/packages/expressions/em_javascript.js` can do the trick i think.

olleharstedt:
Would you mind writing a test for this use-case?

olleharstedt:
Hm, someone should fix that Travis error first. Yes, it's not related to this commit.

Shnoulle:
@olleharstedt You need 3 question One to update the equation + equation + one to show equation value. Maybe add an id in the lss ?

`<div id="result">{Equation.NAOK}</div>`

I give a try when i found times …

TonisOrmisson:
@olleharstedt its not only "that" travis error. 
Merging master into develop created a bunch of errors. I was starting to address them in the beginning of the week but they just seemed endless :) anyway here's a run with all tests with a fork of current develop branch 

https://travis-ci.org/TonisOrmisson/LimeSurvey/jobs/339593111

olleharstedt:
Potentially, it looks worse than it is. Some errors will spit out error messages on several or every test, so often, one fix can be enough.

TonisOrmisson:
probably looks worse yes. still I fixed i guess 2-3 in one feature branch and I got still problems and gave up eventually :(

Anyway it would be excellent if anybody would look into the test failures. I think they are really valuable and helpful and I personally don't want to do much while the base tests are failing ....

olleharstedt:
I agree, but I'm getting increasingly tired of fixing other people's bugs. The one who broke it first should fix it. It should be visible in the Travis build history.

TonisOrmisson:
I also agree, but this started with master merged into develop and this is kind of a special case. 
It started here - the merge:
https://travis-ci.org/LimeSurvey/LimeSurvey/builds/337480044

I think fixing the test failures after merge should be a team effort anyway.

TonisOrmisson:
Also. It seems we need a bit more isolation between the tests. It should not be that one failing test will make most of the other tests fail also. Eg currently the installation test failing causes (i think) a lot of othr tests fail etc. They should be more isolated

TonisOrmisson:
but to conclude on a positive note. I am still very happy that we have the tests finally :)

olleharstedt:
Hm, yeah, I guess merge problems are a slightly different beast.

Isolation between tests is a reoccurring issue. It's fairly good right now, but yes, the installation test can definitely fail better.

If many tests fail, it's usually because of a bug in the code which affects all of those tests (syntax error being the most obvious example).

TonisOrmisson:
turns out it was "that" one fix afterall
here you go #1003

PR #1002, opened 2018-02-08
Remove MySQL specific quotes from a survey criteria

gwn:
Breaks Postgres if we don't.

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -1767,7 +1767,7 @@ return $s->hasTokensTable; });
         );
         $criteria->order = Yii::app()->db->quoteColumnName('groups.group_order').','
             .Yii::app()->db->quoteColumnName('t.question_order');
-        $criteria->addCondition('`groups`.`gid` =`t`.`gid`', 'AND');
+        $criteria->addCondition('groups.gid = t.gid', 'AND');
         return $criteria;
     }
 


Shnoulle:
Maybe must replace by Yii [quoteColumnName](http://www.yiiframework.com/doc/api/1.1/CDbSchema#quoteColumnName-detail) ?

TonisOrmisson:
@Shnoulle No. Yii will do it anyway after. The way @gwn did is how it should be.

TonisOrmisson:
@Shnoulle No. Yii will do it anyway after. The way @gwn did is how it should be.

Shnoulle:
@TonisOrmisson : Thanks, one question why Yii did it here but not on other part ? In Quota count management for example :).

TonisOrmisson:
@Shnoulle anything going through the `CACtiveRecord->query()` will get the quoting done by yii. 
Im not sure about the Quota count. Maybe there is raw SQL composed there?

Shnoulle:
No raw SQL : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Quota.php#L174 

Currently in model, before in common or frontend helper. Started the patch without the quoteColumnName : bug report : broke on mssql, fix the patch

TonisOrmisson:
Seems to be unnceccessary. Yii will do it anyway after

TonisOrmisson:
ok I was rechecking that to be sure and seems that bit everything gets quoted afterall. in this case also yii will not quote it

PR #1007, opened 2018-02-16
fix double token error

3limin4t0r:
Fixed issue #13370: When submitting an invalid token an error is displayed twice.


3limin4t0r:
The same check is already present in the same function a few lines lower.

https://github.com/LimeSurvey/LimeSurvey/blob/92dafe0fa1567aff38d2d334445d84dbe4542566/application/helpers/SurveyRuntimeHelper.php#L1650-L1658

PR #1012, opened 2018-03-09
Better view of readonly advanced setting

Shnoulle:
- better view of readonly advanced setting
- replace html by CHtml , seems cleaner
- The previous one work with input text (only current readonly situation) but we have a lot of other.


Shnoulle:
Travis issue seems not related

> here was 1 failure:
> 1) ls\tests\UpdateDbHelperTest::testCompareUpgradeAndFreshInstall
> (Upgraded from db version 258) Comparing field name "name" for column "showdatasecuritynotice" in table "lime_surveys":  upgraded value: "showdatasecuritynotice";  fresh install value: null
> Failed asserting that null matches expected 'showdatasecuritynotice'.

PR #1014, opened 2018-03-13
Dev: Add a button to show help inline

Shnoulle:
Help in collapse

capture du 2018-03-13 20-28-28


lacrioque:
Maybe loose the button and only show the icon?
I think the button makes it a bit to big

tammoterhark:
Hi Markus, I fully agree, a button around an icon makes the icon itself less easy to distinguish. Better to just display the icon (a little bit bigger then now). I would always show info in blue (blue mostly has a neutral meaning).

tammoterhark:
infoicon

Shnoulle:
@tammoterhark just right after label seem better too. About blue : no colour ;) info color or primary color. I update tomorrow.

tammoterhark:
Reason: asking info is not a real action. Buttons only for real actions. Although I do not know if that is consistent in LS.

Action: changing something?

Shnoulle:
Usage of text-info + inline. With default admin theme : text-info and text-primary are … green (near text-success)

+ Add a tooltip on hover, user have 

1. A quick view on hover
2. A long view (and more) on click
3. This allow us to add link to manual inside the help

application/views/admin/survey/Question/advanced_settings_view.php
```diff
@@ -158,8 +165,11 @@
 /* Launch all needed script (here after load) needed for widget */
 foreach (Yii::app()->clientScript->scripts as $index=>$script)
 {
+    // Add specific view script
+    $script[] = "$('.show-help').tooltip({ html:true, title : function() { return $($(this).attr('href')).html(); }, trigger: 'hover' });";
```
Shnoulle: Maybe here, we must disallow html, user can not really click on the tooltip

c-schmitz:
@lacrioque Should we merge this? Function looks fine to me.

Shnoulle:
And then ? Merging or not ?

Shnoulle:
@lacrioque and @c-schmitz : i can merge it after testing merging master … but need a confirmation.

It work good, no need extra js file and use already included boostrap part …

PR #1015, opened 2018-03-15
Fixed issue #13475: cant export numeric column to SPSS with pgsql

Shnoulle:
Dev: pgsql need casting
Dev: decimal seems OK
Dev: still issue with array number save as text …


Shnoulle:
@c-schmitz : can you just test with mssql please

Shnoulle:
Tested on pgsql and mysql.

PR #1016, opened 2018-03-15
Dev: pdf stat dowload (no popup) on available browser

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/commit/775fe9d94d97a0cfd4c13f3cbb01180cfe702967


Shnoulle:
Working on last FireFox , not tested on Chrome currently, but have another project where it work like that.

Project was started without dowload attribute : work but not after chrome update for my client. Fixed now (if i remind they test on edge too : https://caniuse.com/#feat=download)

Shnoulle:
@lacrioque can you quickly test ? I can merge it , it's working and it's really better

PR #1019, opened 2018-03-19
DEV: Token length to fit UUID 36 chars

TonisOrmisson:
it would be logical for the token field to fit the UUID (32 hex digits + 4 dashes=36 chars)
Currently tokens are 35 chars so one char is missing :)

Should this include also converting all existing survey tables to match this?


Shnoulle:
> Should this include also converting all existing survey tables to match this?

Since we don't have "DB version" in survey : surely yes … else potential bug in some week/month/year/decade (decade ? really ?)

TonisOrmisson:
Thanks @Shnoulle !

I have never done any db-version upgrades. Could anybody please check if I got this right?

application/helpers/update/updatedb_helper.php
```diff
@@ -2264,6 +2276,70 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
...
+            if (!in_array('token', $oTableSchema->columnNames)) {
+                continue;
+            }
+            // No token field in this table
+            switch (Yii::app()->db->driverName) {
+                case 'sqlsrv':
+                case 'dblib':
+                case 'mssql': dropSecondaryKeyMSSQL('token', $sTableName);
+                    alterColumn($sTableName, 'token', "string(36) COLLATE SQL_Latin1_General_CP1_CS_AS");
+                    $oDB->createCommand()->createIndex("{{idx_{$sTableName}_".rand(1, 40000).'}}', $sTableName, 'token');
```
TonisOrmisson: I am not sure about the index in mssql. I copied the logic from 181, but I don not understand whether the index also needs to be overwritten or not. Can this line be removed or will it be lost during the change and must be recreated?

Shnoulle: If index is already done in 181, don't need to add it after :)

TonisOrmisson: thanks!

LouisGac:
I like a lot this idea.
But why a PR to master? Why not to Dev?

TonisOrmisson:
@LouisGac Because I need it more urgently :) 
What's the merge plan for develop into master?
If you say to put it into dev then I will send it there of course. But I would like to have it ASAP personally since I am working on integrations that use UUID.

application/helpers/update/updatedb_helper.php
```diff
@@ -2264,6 +2276,68 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
...
+    $oDB = Yii::app()->db;
+    if (Yii::app()->db->driverName != 'pgsql') {
+        $aTables = dbGetTablesLike("tokens%");
+        if (!empty($aTables)) {
+            foreach ($aTables as $sTableName) {
+                switch (Yii::app()->db->driverName) {
+                    case 'sqlsrv':
+                    case 'dblib':
+                    case 'mssql': dropSecondaryKeyMSSQL('token', $sTableName);
+                        alterColumn($sTableName, 'token', "string(36) COLLATE SQL_Latin1_General_CP1_CS_AS");
```
Shnoulle: Are you sure it was the current collation for mssql

application/helpers/update/updatedb_helper.php
```diff
@@ -2264,6 +2276,68 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
     Yii::app()->setConfig('Updating', false);
     return true;
 }
+
+
+/**
+ * @param string $sMySQLCollation
```
Shnoulle: Can you look at code if here : it's currentColation, with mssql+pgsql or mysql.

Applys same collation than https://github.com/LimeSurvey/LimeSurvey/pull/1019/files#diff-4a95ed4f80f8713aaa18250f998ddfafL138

c-schmitz:
Patch goes to develop now.
Please note that the token fields have different collations than the other fields since they are case-sensitive.

application/helpers/update/updatedb_helper.php
TonisOrmisson: I am not sure about MMql collation. the SQL_Latin1_General_CP1_CS_AS seems to me the only reference in the updatedb_helper file. Can anybody help me with that one?

TonisOrmisson: and here ...
https://github.com/LimeSurvey/LimeSurvey/blob/2b720139cc7ad303481d89d65323329facf614da/application/models/Token.php#L129

PR #1021, opened 2018-03-20
New feature: Anonymize survey (GDPR)

TonisOrmisson:
I am in the middle of reviewing the new EU GDPR legislation and one principle that you can find there is that: 

> Personal data should be processed only if the purpose of the
processing could not reasonably be fulfilled by other means. In order to ensure that the
personal data are not kept longer than necessary, time limits should be established by the
controller for erasure or for a periodic review.

(from paragtaph 39 of the GDPR)

And It seems to me that in many cases a survey owner needs the personal data pieces during the active phase of data collection and data processing. But when the survey is finished, delivered to customer and some time has passed, often the personal data on the data collection interface (eg Ls) is not actually needed any more. So it might be a good idea to have a mechanism to force the anonymization process on a survey that originally was not anonymous. So while we might not want to delete the survey, we might want to clear out some data to make it fully anonymous. Additionally to make sure we don't have any "garbage" old tables that also might contain some data we actually should not keep.

Anyway. I did an initial test on how we could clear out some survey & token fields that could be used for identification. It's not a finalized PR but an initial version to get some feedback and discussion.


Shnoulle:
I like the idea, but a point : Personal data from token can be elsewhere (in attribute) : why not delete token table directly (with old token table)? 

Think best is : delete token table and delete data in token column, no ?

Maybe have another option : allow to delete specific column (value in column) : if survey ask name at start of survey for example. But think it can be done after the 1st commit.

Else : please review HTML part : the onclick is awfull and i think you must create views.

TonisOrmisson:
- I also was thinking about deleting the whole token table. It would be simpler also. 
- Deleting the token column: My first reaction would be that I would not do that, but then again, why not. I think the idea is to have a way to "archive" the project in an anonymous way and logically all data processing stuff should be done by that time. 

- Further I was also thinking about defining specific questions as "personal" so that triggering  anonymizing would clear out data in them

- HTML, I used copy from token-table/delete. It needs work, yes :). I would also not do the execution via GET

thanks for the comments

Shnoulle:
>  I used copy from token-table/delete. It needs work, yes :). I would also not do the execution via GET

🙀 it's done via GET to delete the token table today ? If right : it's a security issue … must be fixed 🙁

TonisOrmisson:
https://github.com/LimeSurvey/LimeSurvey/blob/135f66740a38d3ddecc942c278f93e2026da6df2/application/controllers/admin/tokens.php#L2240-L2257

c-schmitz:
I don't see this as a core feature as only a very small group of people will ever need this (besides the drama it can cause in unexperienced hands). This should rather be an extension.

PR #1022, opened 2018-03-21
beforeQuestionRender Class + attributes

Shnoulle:
- this one seems really needed … adding a class by plugin is needed
- but can not remove included function in old twig file


olleharstedt:
ping @LouisGac

olleharstedt:
Div id works even when no plugin is running the event?

Shnoulle:
Yes, event happen then this line : https://github.com/LimeSurvey/LimeSurvey/pull/1022/files#diff-b69c0b8c301d9dce52f005d2664f3dfbR1690
Force `id=>'questionQid'`

Same behaviour then 2.74

Shnoulle:
It's all done for me.

Shnoulle:
thanks :)

PR #1024, opened 2018-03-25
[WIP] Fixed issue #13529 : Survey and group text are not updated via EM js

Shnoulle:
Dev: add a new EM function to get current value (qid dependency)
Dev: WIP , maybe have another better solution


Shnoulle:
This one ; https://github.com/LimeSurvey/LimeSurvey/pull/1025 is better

PR #1025, opened 2018-03-26
Usage of Expression in theme [All done]

Shnoulle:
Fixed issue #13533 : Unable to use Expression Manager inside template 
Fixed issue #13529 : Survey/Group text with exppression are not updated via javascript

This leave choice for Theme developper, currently seems group descriptin must be added but can remove Survey title : less used.

Theme developer can use Expression like this `{{ processString(Expresssion) }}` expression can come from existing object 
exemples 

- `processString(aSurveyInfo.surveyls_title)` 
- `processString('{strip_tags(TOKEN:ATTRIBUTE_1)}')`

Purpose is to replace existing system and workaround already existing in template.

- [x] Move function to EM
- [x] Test group by group
- [x] Test question by question
- [x] Test all in one
- [x] Test other page (save, clearall)

### DIFF
--- a/application/core/LS_Twig_Extension.php
+++ b/application/core/LS_Twig_Extension.php
@@ -396,4 +396,16 @@ class LS_Twig_Extension extends Twig_Extension
         }
     }
 
+    /**
+     * Process any string with current page
+     * @param string to be processed
+     * @param boolean $static return static string (or not)
+     * @param integer $numRecursionLevels recursion (max) level to do
+     * @param array $aReplacement replacement out of EM
+     * @return string
+     */
+    public static function processString($string,$static=false,$numRecursionLevels=3,$aReplacement = array())
+    {
+        return LimeExpressionManager::ProcessStepString($string, $aReplacement,$numRecursionLevels, $static);
+    }
 }
--- a/application/core/LS_Twig_Extension.php
+++ b/application/core/LS_Twig_Extension.php
@@ -406,6 +406,13 @@ class LS_Twig_Extension extends Twig_Extension
      */
     public static function processString($string,$static=false,$numRecursionLevels=3,$aReplacement = array())
     {
+        if(!is_string($string)) {
+            /* Add some errors in template editor , see #13532 too */
+            if(Yii::app()->getController()->getId() == 'admin' && Yii::app()->getController()->getAction()->getId() == 'themes') {
+                Yii::app()->setFlashMessage(gT("Usage of processString without a string in your template"),'error');
+            }
+            return;
+        }
         return LimeExpressionManager::ProcessStepString($string, $aReplacement,$numRecursionLevels, $static);
     }
 }
--- a/application/config/internal.php
+++ b/application/config/internal.php
@@ -227,6 +227,7 @@ $internalConfig = array(
                 'unregisterScriptForAjax' => 'LS_Twig_Extension::unregisterScriptForAjax',
                 'listCoreScripts'         => 'LS_Twig_Extension::listCoreScripts',
                 'listScriptFiles'         => 'LS_Twig_Extension::listScriptFiles',
+                'processString'           => 'LS_Twig_Extension::processString',
                 'getAllQuestionClasses'   => 'LS_Twig_Extension::getAllQuestionClasses',
                 'intval'                  => 'intval',
                 'empty'                   => 'empty',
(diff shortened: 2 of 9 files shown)


Shnoulle:
@LouisGac : currently to allow usage of Expression Manager inside template : `LimeExpressionManager::FinishProcessingPage` need to be moved AFTER twig page construction : else javascript needed part are not added.

Then i move all this part in `LSETwigViewRenderer` .

Some issue/improvment (maybe)

1. Move this part to EM directly
2. Find a better solution for inputVars : i think even if template didn't have `{{ aSurvey.EM.ScriptsAndHiddenInputs }}` we must add it. Must come from core, not from template (best is to remove usage of EM, but hardest)
3. Maybe move all this part to EM function ? With LEMskipReprocessing too.

Some other idea : put the `processString` option in vanilla template : then, by default : less EM opération.

LouisGac:
1. Agree
2. Agree
3. Not sure.

Shnoulle:
About 2 : find `<form id="limesurvey">` with regexp :/  ? Or when start survey form : add directely the replemant par bfro str_replace (thing this one have much performance).

About 3: if 1 is part of EM , why not SkipReprocessing ? With enableSkipReprocessing + disableSkipReprocessing function ?

Currently : i'm unsure of the usage of this function ;)

Shnoulle:
PS : best is to add needed javascript each time ProcessString happen (sing registerScript) and just regsiter needed var at start of EM, but more for 5.0 or 6.0 ;)

application/core/LS_Twig_Extension.php
```diff
@@ -396,4 +396,16 @@ public static function listScriptFiles()
...
 
+    /**
+     * Process any string with current page
+     * @param string to be processed
+     * @param boolean $static return static string (or not)
+     * @param integer $numRecursionLevels recursion (max) level to do
+     * @param array $aReplacement replacement out of EM
+     * @return string
+     */
+    public static function processString($string,$static=false,$numRecursionLevels=3,$aReplacement = array())
```
Shnoulle: Thing static is more used than recursion, and aReplacemnt than recursion . Then theme dev have less optioon to write

Shnoulle:
@tpartner it's just for usage of this new function in theme. Replace {Expression] for 2.XX in pstpl

tpartner:
@Shnoulle, I find that not all EM placeholders are available.

I have tested in a /multiplechoice/answer.twig.

These do not work (there may be more)
- QUESTION_CODE
- GID
- QID
- GROUPNAME	
- SGQ	

These do work (there may be more)
- LANG	
- SID	
- SURVEYFORMAT	
- SURVEYNAME
- TEMPLATEURL

Do you want me to open a bug?

Shnoulle:
@tpartner : it's not the same issue. 

Can you confirm you can't use it in Question text too ? 

I think : 

1. if it can be used in Question text : it (can be) an issue here.
2. If it can not be an issue in Question text : it's totally another issue.

But about QID/SQG and GID : i's afraid of the way to fix it … maybe add it when load twig file ?

tpartner:
It's a mixture.

GIG, QID, SGQ work in the question text but not in the twig file.

QUESTION_CODE and GROUPNAME don't work in either (and I have enabled "Show question number and/or code" in survey settings
capture-3
)

Shnoulle:
It's what i thing 

1. QID/SQG and GID (all replacement with twig file dependance) : hard to fix, but have idea BUT : can be easily replaced by TWIG variable (you can do it `processString("{QID}",1,3,["QID"=>question.qid])` something near this. Twig var can be used with processString
2. QUESTION_CODE + GROUPNAME : did we fic it in core , after same than 1
3. If we have it at Question/title/etc … can easily fix it BUT : did we need it : TWIG variables already exist.

tpartner:
1. Yes, in some cases we can cheat by processing the twig vars, but doesn't that defeat the purpose of "Usage of Expression in theme"?

2. Don' know

3. I think if we provide some EM vars, we should provide all of them (that are available in the group/question text or old pstpl files.

Here's a better illustration:

capture-4

Shnoulle:
Just about 1 : currently using `{1+2}` in template didn't work like before. Then template workaround need update. 

I don't say i don't want to fix it, i just say : it make more work to fix it :). And unsure it's needed :) ( @tammoterhark have an idea about this maybe ?)

Shnoulle:
TEMPLATEURL is really strange :)

tpartner:
Yes, I understand that it's more work and I don't want to beat you up over it. :) 

But I think we have to consider that there will be "developers" who are even less able than me who will expect to have all of the possible EM vars (or none). I don't think it's easy to decide now what would be "required" and what is not.

Yes, the TEMPLATEURL does have an extra trailing slash (that seems to be another bug).

Shnoulle:
> But I think we have to consider that there will be "developers" who are even less able than me who will expect to have all of the possible EM vars (or none). I don't think it's easy to decide now what would be "required" and what is not.

Yes , totally and except me you and Tammo : i don't know a lot of template dev/hacker :)

tpartner:
I'll open a bug about the EM vars that are not available in the group/question text...

maziminke:
We often use {date("Y")} for a copyright information at the footer of custom templates so it would be great if we could make that work again!

tpartner:
@maziminke, this works as expected:

`{{ processString('{date("Y")}') }}`

Shnoulle:
> I'll open a bug about the EM vars that are not available in the group/question text...

Did you do it ? I answer_html, before twig : a create a global array for replacement in public controller. Maybe we must set it at the good time …

This global array are used every time, and can be extended

tpartner:
> Did you do it ?
Just did - https://bugs.limesurvey.org/view.php?id=13574 

It looks like everything now works except those listed in the bug.

capture-3

PR #1028, opened 2018-03-29
Fixed issue #13534: Language selection does not work in survey-list page

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=13534


olleharstedt:
Thanks!

PR #1031, opened 2018-03-30
Fixed issue #13519: No import button in Template listing 

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=13519


LouisGac:
great job

PR #1032, opened 2018-03-30
Theme +twig fix for index, usage for other part can be great too

Shnoulle:
Fixed issue #13542: Question index (head) show diable even if work
Fixed issue #13544: Unable to use flatEllipsizeText function in twig file
Dev: replace flatEllipsizeText by 2 LS_Twig_Extension
Dev: little css fix for index button : how to reset template config ?
Dev: processString in template are the theme roles, not core.


Shnoulle:
Moving to develop

PR #1033, opened 2018-03-31
Fixed issue #13368: Favicon transparent background

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=13368


TonisOrmisson:
this resolves replacing the favicon file for admin.
additionally it seems the favicon is missing for templates due to the templateUrl / oTemplate missing in twig. But that's another issue

PR #1036, opened 2018-04-03
Fixed issue #13565: Remove punctuation points from php

KRav007:
Fixed issue #13565: Remove punctuation points from php

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -6574,7 +6574,7 @@
                         {
                             $qmandViolation = true; // TODO - what about 'other'?
                         }
-                        $sMandatoryText = $LEM->gT('Please complete all parts').'.';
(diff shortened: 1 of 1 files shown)


olleharstedt:
You should probably change the translations and include the period.

KRav007:
What means "the period". If I put the point to the translation-string, the mo-files don't find the string anymore?! Right?! How to deal with the x mo files. Is there a job that move strings around - to the mo files ?

Shnoulle:
+1 for @olleharstedt , yes translation must be updated :).

olleharstedt:
@KRav007 Just move the point (full stop) into the string, and the translation tools will automatically mark the translation as "needs to be updated".

KRav007:
ok - cool - done

LouisGac:
Is this going to work with RTL languages?

KRav007:
From my logic - it should. RTL would not beginn with punctions on right side…

PR #1037, opened 2018-04-06
fixed issue #13556: breadcrumb link (from left menu) goes to basic view url

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=13556


TonisOrmisson:
@lacrioque Did not understand your question. 

I put the classes back and just removed the link.

Shnoulle:
>  But would you want to remove the link?

See bug report : 

1. Link is invalid
2. No need link (we are on this page)

PR #1038, opened 2018-04-09
New feature "template_copy" and "set_template_permission" on remote control

gabrieljenik:
New features on remote control.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -3066,4 +3066,100 @@ class remotecontrol_handle
         }
         return false;
     }
+
+    /**
+     * Create a copy of a template
+     *
+     * @param int $sSessionKey
+     * @param  string $copydir Template to copy
+     * @param  string $newname New template name
+     * @return array with status
+     */
+    public function template_copy($sSessionKey, $copydir, $newname)
+    {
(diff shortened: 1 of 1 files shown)


TonisOrmisson:
https://manual.limesurvey.org/LimeSurvey_roadmap#Limesurvey_v2.50_.28Feb_2016.29_-_Status:_Archived

gabrieljenik:
Sorry. What do you mean?

TonisOrmisson:
The version you want to push to is archived. 
All new features should be going into V3 develop branch as far as I have understood.

gabrieljenik:
I understand. But this could be considered as an update more than as a new feature.

Shnoulle:
ANd more : @c-schmitz ask new feature only in develop version ( next 4.0 …)

maziminke:
Why should we not add this to the next 3.x release? It can hardly break anything and that version will be in use the next months.

maziminke:
@gabrieljenik Did this feature made it into Limesurvey?

I consider it a very useful feature for the API.

gabrieljenik:
Nope... I think it was left behind :)

olleharstedt:
How about adding it to 4.x?

maziminke:
LS4 -> Of course!

But I'd love to see that also at LS3 as well since most users are using that version.

PR #1040, opened 2018-04-12
Rollback #812

alexandcote:
As mention in the PR https://github.com/LimeSurvey/LimeSurvey/pull/812#issuecomment-380826565, this introduce a regression and always display trace even if `debug = 0`.

I think we should tag a new 2.X version for this.

Thanks again guys.

### DIFF
--- a/application/helpers/globals.php
+++ b/application/helpers/globals.php
@@ -29,5 +29,5 @@
         if(isset($trace['file'],$trace['line']) && strpos($trace['file'],YII_PATH)!==0) {
             $msg = $trace['file'].' ('.$trace['line']."):\n" . $msg;
         }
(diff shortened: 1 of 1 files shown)


Shnoulle:
But ? I didn't add the needed
 `'enable' => defined('YII_DEBUG') && YII_DEBUG`

alexandcote:
@Shnoulle not sure to understand but yes all needed was [there](https://github.com/LimeSurvey/LimeSurvey/blob/master/index.php#L158) I think.

alexandcote:
Your suggestion work 👍

alexandcote:
I will let you do the change because I don't have many time is it good for you ?

PR #1042, opened 2018-04-13
Fixed issue #13604: beforeSurveyDelete didn't happen when delete a survey

Shnoulle:
- I know it's only for plugins ;)
- But i think we can add it in [AuditLog](https://github.com/LimeSurvey/LimeSurvey/tree/master/plugins/AuditLog)


Shnoulle:
😂
capture d ecran_2018-04-13_18-28-35

PR #977, opened 2018-01-30
New feature: allow selecting MySql engine (MyIsam vs InnoDB)

TonisOrmisson:
Something we discussed in one meeting.
I have tested the installation on local dev mariadb and multiple instances on Google Cloud SQL. 

I would appreciate any testing, & reviews enhancements. 
This probably needs a Warning note to describe what it means to select InnoDb

PR #1043, opened 2018-04-17
New Feature: allow choosing innoDb vs MyIsam engine for MYSQL during installation

TonisOrmisson:
Formerly opened form master #977, now for develop

Some notes: 
in order for innoDb to work one needs:
 - innodb_large_prefix = 1
 - innodb_file_format =  Barracuda;
 - innodb_file_format_max =  Barracuda;

otherwise you will get the "Error 1071: Specified key was too long; max key length is 767 bytes" for the text indexes, (Error 1071: Specified key was too long; max key length is 767 bytes)

I enabled this for travis 
https://github.com/LimeSurvey/LimeSurvey/compare/develop...TonisOrmisson:develop-innodb?expand=1#diff-354f30a63fb0907d4ad57269548329e3R45

and checking of this will be done in installer validation here
https://github.com/LimeSurvey/LimeSurvey/compare/develop...TonisOrmisson:develop-innodb?expand=1#diff-12a8fc159f48cb97e97609f6c43aab0aR258

Did some major changes in the installer logic. 
Why? because I needed to check the Innodb settings status in the model validation: so the dbConnection needed to move from the controller to the Model. And so all of the connection related things were moved to model. Controller got much more cleaner and is now not doing things that logically should not be there. 

Also cleaned up the view along that since there was code that should not be in the view. 

Tests: 
I enabled separate set of tests for InnoDB and MyIsam. So the whole test set runs both ways now. 
Also added PHP 7.2 to the matrix. 

... I think we should also use the pgsql for testing on travis. As far as I understand this is also included for travis so I would use an ENV DBTYPE to run test matrix also on available non-mysql options. 

I have tested this manually also during the work, but since there is a lot of changes here I would ask for you to review the code as well as also help to test manually. **Especially for MSSQL** that I do not have and is not tested against!


olleharstedt:
Phew, big PR. We should probably discuss it during the meeting today, if you have time.

olleharstedt:
Maybe we can just remove Psalm in the Travis script? `composer remove psalm` or or sim.

olleharstedt:
Psalm is a static analyzer. It's safe to remove in Travis environment. In the future we will use it to validate third party extensions (you can also integrate it into your IDE/editor of choice to get error message on save).

TonisOrmisson:
@olleharstedt `composer remove psalm` will not resolve since we are doing `composer install` and reading from `composer.lock`.

what we need to do is restrict the version not to use the dev-master, but a stable release eg `"vimeo/psalm": "~1.1"`
https://github.com/vimeo/psalm/releases

what we get now is a pre-relase of v2.0.0-dev which what we probably do not want.

TonisOrmisson:
https://github.com/LimeSurvey/LimeSurvey/pull/1046

olleharstedt:
The pre-release of Psalm contains some fixes I explicitly needed for it to work with our Frankenstein Yii setup.

olleharstedt:
Can't we just brutally remove it somehow in the Travis script?

TonisOrmisson:
I tried to run also an `composer install --no-dev` here 
https://travis-ci.org/TonisOrmisson/LimeSurvey/jobs/369265295
but this also fails

use your own fork of psalm with php5 ?
ask them to put php5 support back? :)
remove it totally ? :)

c-schmitz:
Remind me again why you want to use InnoDB?

mitchcapper:
Several platforms have dropped MyIsam support, this includes Amazon Aurora and Microsoft Azure Mysql.  There are replication issues with myisam as well along with backup limitations.   The feature set with innoDB also has largely created little reason not to support it in current versions, it also has more recovery options for failure.

TonisOrmisson:
If the core team would be prepared to review this commit I would go into sorting out the conflicts and re-testing this.

virgilwashere:
Yes I'm confused at their reluctance.

Thanks for your efforts on the PR.

TonisOrmisson:
@c-schmitz I remover the Traits, I agree its better if they are out of this. 
Regarding the rest - since I have no immediate need for this - and since time has passed since I initially wrote this - I will not go into re-doing this. You are welcome to use this any way you like. 

Adding this feature is a big change indeed and to be honest I gave this controller a fresh look (since there was anyway a big change) and probably changed all bits that I was editing to be better than before (more logical, standards, etc...).

c-schmitz:
Looks better, thank you.
I can see that during installation you set the engine for the session by using "SET default_storage_engine" during table creation.
How do you set the engine to be used for responses/token tables?

TonisOrmisson:
Excellent question, thanks!

As far as I understand - in the current set-up any default setting would be used. 

To be sure the proper engine is used for survey tables, it should be set in SurveyActivator. To have that the preference should be stored in Global Settings. Probably? Not sure about what happens if one would try to use an engine that is not supported.

There might be different approached here and I'm not sure what would be best. 

Should the the engine be an (advanced) option somewhere if the database supports both options (eg. if one runs into a column-count issue with innodb)? Or should we stick to the engine one chose during install?

My first guess would be to store the engine preference in global settings, but not allow to change it via UI.

olleharstedt:
Write default engine to config?

TonisOrmisson:
LimeSurvey Admin should not tweak the dbEngine unless he/she basically is also sysadmin in that sense. If you want to change the dbengine after installation, you should really understand the consequences.

TonisOrmisson:
And a separate idea for the future. Feel free to comment on that - I do not plan to implement it here. 

The need for MyIsam in mysql is due to the need for mode columns in longer surveys right?

But it is only the actual survey tables, nothing else. I feel like the default engine for system tables as well as token table should be InnoDb by default. I see no need to keep any of that in MyIsam (except simplicity - all tables same engine).

Shnoulle:
We must compare InnoDB vs MYisam about max number of question : https://dev.mysql.com/doc/refman/8.0/en/column-count-limit.html

olleharstedt:
Closed by mistake.

TonisOrmisson:
@olleharstedt @c-schmitz the preferred mysqlEngine is now being written to config.php and the SurveyActivator uses that value while creating the survey tables. 

Did some testing to assure it works as it should. Also worked nicely if changed in config after installation. 

Anybody see some other missing parts? (I also removed the WIP tag since I can not see any at this point)

c-schmitz:
Awesome! Thank you for your work. From my point of view this is ready to be merged.

Shnoulle:
If i don't make error [installer/sql-old/create-mysql.sql](https://github.com/LimeSurvey/LimeSurvey/pull/1043/files#diff-3eec665f83d854d3be81681fb081c77c) and other sql-old/ files can be removed now ?

Else : good work :)

TonisOrmisson:
Ok now.
The timings and responses tables are created by the SurveyActivator using the same session. So the timigs table was OK. But tokens table was created with a separate request. This is now fixed also. Also added the engine option to activate_helper and AuditLog table creations. I tested the activation parts wihle changing the engines with responses/timings/tokens table. Feel free to merge.

plugins/AuditLog/AuditLog.php
```diff
@@ -557,6 +557,11 @@ public function beforeActivate()
         {
             if (!$this->api->tableExists($this, 'log'))
             {
+                $options = '';
+                if (Yii::app()->db->driverName == 'mysqli' || Yii::app()->db->driverName == 'mysql') {
```
Shnoulle: I think more of something in API ;) : api->createTable function (but this api is … so … so … light)

dominikvitt:
I just finished with testing.
There is occasionally a following error when trying to import .lsa file:

>  Call to a member function getColumnNames() on null  
> _application/helpers/admin/import_helper.php(2205)_

There is also this error when trying to activate a survey with many columns:

> CDbCommand failed to execute the SQL statement: SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large (> 8126)

It is obvious those errors are related to innoDB limitation of having too many columns. 

In such case, import/activation should be aborted and warning should be shown to user.
Anybody have some idea about this?

Shnoulle:
> It is obvious those errors are related to innoDB limitation of having too many columns.

InnoDB don't sedn exception here ? https://github.com/LimeSurvey/LimeSurvey/blob/99d396dde8cd9a1c0beec81634a3488d16ddb085/application/helpers/admin/activate_helper.php#L457-L466

Shnoulle:
Just checkout : 

>  CDbCommand failed to execute the SQL statement: SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'firstname' used in key specification without a key length. The SQL statement executed was: ALTER TABLE `lime_participants` CHANGE `firstname` `firstname` text NOT NULL

TonisOrmisson:
Thanks for testing!
Can you tell whether the InnoDb acts differently compared to MyIsam when the column limit is reached? I would think that such error would be regardless of engine?

Maybe you can add some test files to the `tests/data/surveys` folder so that this could be tested automatically. Eg a survey with too much columns for MyIsam. And a survey with too much columns for InnoDb. 

Did you try with debugging = 0, was it the same?

dominikvitt:
When activating, it shows nice warning message when debug=0.

dominikvitt:
Column limit does not apply to the BLOB and TEXT type.
In current development version, all response columns are created as TEXT (because data encryption feature require it).
So, I'll try to merge your changes to my local develop version and see if problem is corrected.

TonisOrmisson:
@dominikvitt 
I guessing it is the same behaviour that we have currently (in master) and with MyIsam. If that's the case, maybe its not specifically related to this feature? If you turn off debugging there is a number of places you might end up with an exception thrown.

TonisOrmisson:
@Shnoulle can you check what is the status of the engine variables (see top of page):

    innodb_large_prefix;
    innodb_file_format;
    innodb_file_format_max;

your issue sounds like related to innodb_large_prefix. 
This should be checked during installation actually. Or maybe you changed some after installation?

Shnoulle:
I just checkout my current develop with your's.

Currently : even wity core develop : i have issue with my master DB to upgrade to develop.
@dominikvitt : do you know when we check DB updater ?

dominikvitt:
I would test now upgrading from master to develop.

Shnoulle:
> I would test now upgrading from master to develop.

Ok , i check too (maybe must do a `checkintegrity` , i test and if `checkintegrity` fix my issue i report to do one before upgrade (in fact i think we must do one each time we upgrade : new Feature …)

dominikvitt:
@Shnoulle Database problem is fixed now, you can try again.

Shnoulle:
@dominikvitt , sorry : https://bugs.limesurvey.org/view.php?id=14729 :)

dominikvitt:
Current develop version is already using TEXT db fields for all response columns, so we can't solve it like that. Let it stay for now as it is.

@TonisOrmisson Can you add check to see if table is actually created while importing .lsa file?
If table is not created, then you can show the same message as it is for survey activation.

TonisOrmisson:
created a new survey with this branch, activated and created some responses, exported lsa and imported as new survey. All relevant tables were created. responses, timings, tokens.

maziminke:
> In current development version, all response columns are created as TEXT (because data encryption feature require it).
Does that mean that e.g. for the date question type or the numeric input a TEXT field is used as well? 
 
How does that work with features like filtering statistics by date or calculating e.g. mean or average values at statistics and other features? Don't they rely on a proper DB field type?

Shnoulle:
> In current development version, all response columns are created as TEXT (because data encryption feature require it).

Really ? And this was merged in develop … for alpha version … the DB column type is the most complicated system to move to QuestionObject … don't do it at start of devlopment seem realy an issue in my opinion …

dominikvitt:
> How does that work with features like filtering statistics by date or calculating e.g. mean or average values at statistics and other features? Don't they rely on a proper DB field type?

All features you mentioned don't rely on specific DB field type and data would be read the same regardless of DB field type. So, there shouldn't be any different behaviour when using TEXT DB field type.

dominikvitt:
> created a new survey with this branch, activated and created some responses, exported lsa and imported as new survey. All relevant tables were created. responses, timings, tokens.

Please try to create a survey containing large number of response columns (e.g. over 100 or 200) and redo all steps to test it.
Then you should be able to reproduce that error.

TonisOrmisson:
> Maybe you can add some test files to the tests/data/surveys folder so that this could be tested automatically. Eg a survey with too much columns for MyIsam. And a survey with too much columns for InnoDb.

Shnoulle:
> Maybe you can add some test files to the tests/data/surveys folder so that this could be tested automatically. Eg a survey with too much columns for MyIsam. And a survey with too much columns for InnoDb.

A survey with 10 000 column must broke for all DB ;)

PR #1044, opened 2018-04-17
Fixed issue #13606: Unable to reset manually asset

Shnoulle:
Dev: get LS version in hash function
Dev: this replace Yii::getVersion() in parent function
Dev: then filemtime is still used asset by asset

### DIFF
--- a/application/core/LSYii_AssetManager.php
+++ b/application/core/LSYii_AssetManager.php
@@ -1,10 +1,9 @@
 <?php
-
 /**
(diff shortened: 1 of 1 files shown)


Shnoulle:
After we can fix https://bugs.limesurvey.org/view.php?id=13497 

Do ALL asset each time break css/js loading for ALL admin part + ALL public survey currently used …

Shnoulle:
@lacrioque why revert ?

PR #1047, opened 2018-04-20
Fixed issue #13599:  Table user_in_groups not cleaned

TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=13599

1) new deletions of groups will delete also user-group assignments

2) old undeleted assignments will be cleaned via the integrity checking tool (which looks awful btw)

### DIFF
--- a/application/controllers/admin/usergroups.php
+++ b/application/controllers/admin/usergroups.php
@@ -121,10 +121,9 @@ class Usergroups extends Survey_Common_Action
         if (Permission::model()->hasGlobalPermission('usergroups', 'delete')) {
             $ugid = Yii::app()->request->getPost("ugid");
             if (!empty($ugid) && ($ugid > -1)) {
-                $result = UserGroup::model()->requestEditGroup($ugid, Yii::app()->session["loginID"]);
-                if ($result->count() > 0) {
-                    $delquery_result = UserGroup::model()->deleteGroup($ugid, Yii::app()->session["loginID"]);
-                    if ($delquery_result) {
+                $userGroup = UserGroup::model()->requestEditGroup($ugid, Yii::app()->session["loginID"]);
+                if (!empty($userGroup)) {
+                    if ($userGroup->delete()) {
                         Yii::app()->user->setFlash("success", gT("Successfully deleted user group."));
                     } else {
                         Yii::app()->user->setFlash("notice", gT("Could not delete user group."));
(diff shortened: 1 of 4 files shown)


application/controllers/admin/usergroups.php
```diff
@@ -121,10 +121,9 @@ public function delete()
         if (Permission::model()->hasGlobalPermission('usergroups', 'delete')) {
             $ugid = Yii::app()->request->getPost("ugid");
             if (!empty($ugid) && ($ugid > -1)) {
-                $result = UserGroup::model()->requestEditGroup($ugid, Yii::app()->session["loginID"]);
-                if ($result->count() > 0) {
-                    $delquery_result = UserGroup::model()->deleteGroup($ugid, Yii::app()->session["loginID"]);
```
TonisOrmisson: The permission checking done in `UserGroup::model()->deleteGroup()` is done anyway in `UserGroup::model()->requestEditGroup()` and even that should not be in model. Added the `@deperecated` to `UserGroup::model()->deleteGroup()` (was also only isage I found)

TonisOrmisson: @Shnoulle I just wanted a second pair of eyes to have a look and github suggested you :)

and yes agree totally about 1 & 2
3 sometimes maybe (if 2 not possible).

Shnoulle: Since we don't have same system in Group than in Survey : yes, this don't break system.

application/models/UserGroup.php
```diff
@@ -231,6 +231,7 @@ public function requestViewGroup($ugId, $userId)
      * @param integer $ugId
      * @param integer $ownerId
      * @return bool
+     * @deprecated since 2018-04-21 use $this->delete and do the permissions check in controller!!
```
Shnoulle: Where place the 'listGroup' where dev can have : list of group with Permission usage ?

application/models/UserGroup.php
```diff
@@ -246,12 +247,21 @@ public function deleteGroup($ugId, $ownerId)
...
         if ($group->getErrors()) {
-                    return false;
+            return false;
         } else {
-                    return true;
+            return true;
         }
     }
 
+    public function delete()
```
Shnoulle: :+1: (just need inheritdoc ;) )

PR #1048, opened 2018-04-22
Fix multiple issues with theme assets

Shnoulle:
Fixed issue #13606 : Unable to reset manually single asset 
Fixed issue #13497 : Editor screen can be loaded without css after save
Fixed issue #13596: template assets is not resetted when update config
Dev: move injectglobalsettings to LSYii_Application (getGlobalSetting can be deprecated since a lot of time)


Shnoulle:
I think we can deprecate the 2 last function in globalsettings_helper and remove file in 4 or 5

application/controllers/admin/themeoptions.php
```diff
@@ -80,11 +80,12 @@ public function update($id)
     {
         if (Permission::model()->hasGlobalPermission('templates', 'update')) {
             $model = $this->loadModel($id);
-
             if (isset($_POST['TemplateConfiguration'])) {
                 $model->attributes = $_POST['TemplateConfiguration'];
                 if ($model->save()) {
                     Yii::app()->user->setFlash('success', gT('Theme options saved.'));
+                    /* If one TemplateConfiguration_files_* updated, we need to update asset */
+                    Template::model()->findByPk($model->template_name)->resetAsset();
```
LouisGac: No, it's not needed. Options don't change the content of the CSS files.

application/controllers/admin/themes.php
```diff
@@ -657,7 +657,8 @@ public function templatesavechanges()
 
                         // If the file is an asset file, we refresh asset number
                         if (in_array($relativePathEditfile, $cssfiles) || in_array($relativePathEditfile, $jsfiles)){
-                            SettingGlobal::increaseCustomAssetsversionnumber();
+                            //SettingGlobal::increaseCustomAssetsversionnumber();
+                            Template::model()->findByPk($sTemplateName)->resetAsset();
```
LouisGac: For now, assets are global, not Theme related. So the function to reset assets belongs to GlobalSetting Model, not Template Model. Also, it doesn't depend to any specific item (any instantiated object), so it should be static.

application/core/LSYii_Application.php
```diff
@@ -96,6 +96,17 @@ public function __construct($aApplicationConfig = null)
...
+        try {
+            $settingsTableExist = Yii::app()->db->schema->getTable('{{settings_global}}');
+            if (is_object($settingsTableExist)) {
+                $dbConfig = CHtml::listData(SettingGlobal::model()->findAll(), 'stg_name', 'stg_value');
+                $this->config = array_merge($this->config, $dbConfig);
+            }
+        } catch (Exception $exception) {
+            // Allow exception (install for example)
+            App()->log("Table settings_global not found"); // Log it as LEVEL_INFO ?
+        }
```
LouisGac: Is this code tested with the installer?

application/core/LSYii_AssetManager.php
```diff
@@ -14,27 +13,13 @@
...
-     */
-    protected function generatePath($file, $hashByName = false)
+    /* inheritdoc */
+    protected function hash($path)
     {
         $assetsVersionNumber       = Yii::app()->getConfig('assetsversionnumber');
         $versionNumber             = Yii::app()->getConfig('versionnumber');
         $dbVersion                 = Yii::app()->getConfig('dbversionnumber');
-        $iCustomassetversionnumber = (function_exists('getGlobalSetting') ) ? getGlobalSetting('customassetversionnumber'):1; // When called from installer, function getGlobalSetting() is not available
+        $iCustomassetversionnumber = Yii::app()->getConfig('customassetversionnumber',1);/* This always get by order : from db, from config, from config-default */
```
LouisGac: same: is it tested with the installer?

application/helpers/globalsettings_helper.php
```diff
@@ -15,20 +15,6 @@
...
-{
-    $settings = SettingGlobal::model()->findAll();
-
-    //if ($dbvaluearray!==false)
-    if (count($settings) > 0) {
-        foreach ($settings as $setting) {
-            Yii::app()->setConfig($setting->getAttribute('stg_name'), $setting->getAttribute('stg_value'));
-        }
-    }
-}
```
LouisGac: why removing this function in this commit? 
Is it related to theme assets ?

application/models/Template.php
```diff
@@ -509,6 +509,29 @@ public static function resetInstance()
...
+            $counter++;
+            // This is extremely unlikely.
+            if ($counter > 10) {
+                throw new CHttpException(500, 'Failed to reset template asset with new file.');
+            }
+            $fileToTouch= App()->securityManager->generateRandomString(42);
+        }
+        touch($templatePathWithDirSeparator.$fileToTouch);
+        unlink($templatePathWithDirSeparator.$fileToTouch);
+    }
```
LouisGac: No: we don't want to use any more the "touch"  logic. 
On 2.50+ we face the fact that the file system behave differently in many situation: SELinux, Linux, windows, specific config, etc 
So on LS3 the touch logic has been replaced by an Asset Version Number, one in the version file for the ComfortUpdate, the other one in DB for the user.

application/core/LSYii_Application.php
LouisGac: also: is it related to the commit? To the asset system?

Shnoulle:
Didn't use file system for asset is a mistake … and touch broke only for windows and directory.

We need a solution for : file update AND theme assets generation

Shnoulle:
Usage of DB only are a regression of touch system on all linux system …

Shnoulle:
User updatable templates are in upload directory : then are updatable. If user can't write in upload directory : no need to update template (and function didn't happen).

Else we can too create a assetnumber DB with path as pk and a number. We still need asset with filemtime and adding current path assetnumber ( file or directory with complete path).

Just need update DB

Shnoulle:
@LouisGac if i update commit like this : 

1. Add a assetversion db table with key: path (text) and version as int
2. Use it in LSYii_AssetManager::generatePath for directory only (file not needed ca return parent)
3. Increment (or set to 1 if not exist) in `Template::model()->findByPk($assetname)->resetAsset()`

It's OK ?

Shnoulle:
@LouisGac : i start adding delete asset instruction. But not sure of the place to call …

From Template in $this->renameTo() function : OK done

But when a Template is deleted, there are no Template::model()->findByPk()->delete() send ?

Shnoulle:
@LouisGac and @c-schmitz : all is fixed and tested. This fix a lot of issue with asset and template editor

- In some situation : can break public survey (asset generation for all asset, not only template asset)
- asset are not resetted when  sending an new file (or delete it)
- Still use default Yii system with mtime test on directory, this allow updating single file asset easily, or update directory asset via ftp without need to go to admin gui

Shnoulle:
@LouisGac and @c-schmitz : maybe i can do a plugin for this … see https://github.com/LimeSurvey/LimeSurvey/commit/3b152c658c15613de5e64da2d83e1d74e996041b for a way to replace core class by plugin class.

But it's really an issue here, not a feature …

Shnoulle:
For remind : this commit was done after https://github.com/LimeSurvey/LimeSurvey/commit/929dceeb60e9821fba235cd77f4bc826176b68e6 .

This one was reverted because of usage of App()->getConfig('configKey') , then clearing all usage of this one (`getGlobalSetting` must be deprecated since more than 2 years).

PR #1049, opened 2018-04-22
DEV: simplify optin controller

TonisOrmisson:
Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/application/controllers/OptinController.php
+++ b/application/controllers/OptinController.php
@@ -26,8 +26,8 @@
 class OptinController extends LSYii_Controller
 {
 
-        public $layout = 'bare';
-        public $defaultAction = 'tokens';
+    public $layout = 'bare';
+    public $defaultAction = 'tokens';
 
     function actiontokens($surveyid, $token, $langcode = '')
     {
@@ -42,7 +42,7 @@ class OptinController extends LSYii_Controller
         if (!$iSurveyID) {
             $this->redirect(array('/'));
         }
-        $iSurveyID = (int) $iSurveyID;
+        $iSurveyID = $oSurvey->primaryKey;
 
         //Check that there is a SID
         // Get passed language from form, so that we dont loose this!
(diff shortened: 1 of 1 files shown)


application/controllers/OptinController.php
```diff
@@ -42,7 +42,7 @@ function actiontokens($surveyid, $token, $langcode = '')
         if (!$iSurveyID) {
             $this->redirect(array('/'));
         }
-        $iSurveyID = (int) $iSurveyID;
+        $iSurveyID = $oSurvey->primaryKey;
```
Shnoulle: Maybe can remove this line : https://github.com/TonisOrmisson/LimeSurvey/blob/b8af9b7256deb22e8ed8d8b3b7f67b86652182cc/application/controllers/OptinController.php#L37 (directly use  `$oSurvey = Survey::model()->findByPk($surveyid);` )

And send a 404 error if $oSurvey is null

application/controllers/OptinController.php
```diff
@@ -54,13 +54,11 @@ function actiontokens($surveyid, $token, $langcode = '')
 
         Yii::app()->setLanguage($sBaseLanguage);
 
-        $aSurveyInfo = getSurveyInfo($iSurveyID, $sBaseLanguage);
-
-        if ($aSurveyInfo == false || !tableExists("{{tokens_{$iSurveyID}}}")) {
+        if (empty($oSurvey) || !$oSurvey->hasTokensTable) {
```
Shnoulle: empty test is not needed , done at https://github.com/TonisOrmisson/LimeSurvey/blob/b8af9b7256deb22e8ed8d8b3b7f67b86652182cc/application/controllers/OptinController.php#L42

application/controllers/OptinController.php
```diff
@@ -54,13 +54,11 @@ function actiontokens($surveyid, $token, $langcode = '')
...
             throw new CHttpException(404, "This survey does not seem to exist. It may have been deleted or the link you were given is outdated or incorrect.");
         } else {
-            LimeExpressionManager::singleton()->loadTokenInformation($iSurveyID, $token, false);
-            $oToken = Token::model($iSurveyID)->findByAttributes(array('token' => $token));
+            LimeExpressionManager::singleton()->loadTokenInformation($iSurveyID, $sToken, false);
+            $oToken = Token::model($iSurveyID)->findByAttributes(array('token' => $sToken));
 
             if (!isset($oToken)) {
                 $sMessage = gT('You are not a participant in this survey.');
```
Shnoulle: I like to have a HTML error here, unsure of code, maybe 403 ?

TonisOrmisson: Maybe, but I was not looking at that here at all :)
Just there was a sanitized $sToken there that was not used and an unsanitized $token used instead

Shnoulle:
Yep, but more : best to use 40X error when we have (again) a real clean error system

PR #1061, opened 2018-05-03
Fix "There are X questions in this survey." always displays

beat:
Even if the setting in Presentations "Display 'There are X questions in this suvery" is set to No (parameter value 'N'), the number of questions is still displayed in the Welcome message.

I'm not sure if this is the right place to fix the issue, but there is a mismatch in parameter name and a mismatch in the value tested. This fixes it, but feel free to reject this fix and propose a better one. Thanks !


dominikvitt:
It looks like fix for this issue is already applied, so I will close this PR.

PR #1063, opened 2018-05-09
Dev: move default QuestionAttribute to model and use it

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/commit/3b152c658c15613de5e64da2d83e1d74e996041b (and some other)

Question Attribute settings need some default always set, then set it always.


Shnoulle:
Bad merging … fixing it in some minutes

PR #1066, opened 2018-05-10
DEV: implement QuestionType

TonisOrmisson:
Originally I started to work with FieldMap a bit but seems this needs to be taken case of first. 

The main actual change for current codebase is moving the Question::typeList() to
QuestionType. If this gets merged we can move on to using the QuestionType for current use cases

And then later use QuestionType and Field to construct a FieldMap ...

I would appreciate a review. Not sure who to ask here :)


Shnoulle:
About fieldsName , i start to work on it too :). 
See https://gitlab.com/DRAAF_Bourgogne-Franche-Comte/surveyChaining/blob/master/helpers/surveyCodeHelper.php#L66

I think i can add it on your fork (and rework with your fork on base), maybe you already done.

Can you tell me your final fork where i can create a merge request ?

TonisOrmisson:
Im working currently here on doing the FieldMap on top of it:
https://github.com/TonisOrmisson/LimeSurvey/tree/fieldmap2

olleharstedt:
:+1:

application/models/Field.php
```diff
@@ -0,0 +1,68 @@
...
+    /** @var string $name Field column name */
+    public $name;
+
+    /**
+     * Field constructor.
+     * @param Question|null $question
+     */
+    public function __construct(Question $question = null)
+    {
+        $this->question = $question;
```
olleharstedt: Shouldn't it be the other way around, that question aggregates fields?

application/models/QuestionType.php
```diff
@@ -0,0 +1,538 @@
...
+ * @property string $fieldDataType numeric vs string types
+ * @property boolean $isText Whether the type is text (string longer than char)
+ * @property boolean $isChar Whether the type is char (one-character-string)
+ * @property boolean $isString Whether the type is string (text or char)
+ * @property boolean $isNumeric Whether the type numeric (integer, double)
+ * @property boolean $isInteger Whether the type integer
+ *
+ * {@inheritdoc}
+ */
+class QuestionType extends StaticModel
```
olleharstedt: Instead of class QuestionType, shouldn't we have one separate model for each question type?

PR #1073, opened 2018-05-11
 DEV: implement Question->delete() 

TonisOrmisson:
Question deletion did not take care of L10n-s

Implemented the Question->delete in model properly rather than in controller

### DIFF
--- a/application/controllers/admin/questions.php
+++ b/application/controllers/admin/questions.php
@@ -1325,6 +1325,7 @@ class questions extends Survey_Common_Action
     {
         $surveyid = sanitize_int($surveyid);
         $qid = (int) $qid;
+        $oQuestion = Question::model()->findByPk($qid);
         $rqid = $qid;
 
         if (Permission::model()->hasSurveyPermission($surveyid, 'surveycontent', 'delete')) {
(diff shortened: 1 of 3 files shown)


Shnoulle:
In other delete , we do : 

```` 
if(parent::delete()) {
   // Delete related
}
````

I think it's best : if current questuion are not deleted : must not delete sub question and language :)

Shnoulle:
Another question : when you use deleteAll on sub question : did self::delete() happen ?

Because sub question need deletion with related too :)

TonisOrmisson:
@Shnoulle 
regarding whether to delete the children after of before the parent relation:
consider you have Foreign keys with what eg Yii2 would create by default for mysql
https://www.yiiframework.com/doc/api/2.0/yii-db-migration#addForeignKey()-detail
If you would have a FK relationship between for example question and answers with default values, you would fail to delete the question before its FK relations. 

LS currently does not have FK relations, but looking in future it might be the case that it will

So unless you set the DB-engine to do deal with the "on delete" for FK-s, it would be only logical to do the deleting basically backwards. Relations first, then the parents. 

Anyway, this was the logic I used here. 
And in terms of failures. The deleting of children can probably fail similarly as the parent, no?

I can of course make it the way you suggest, but there was an actual logic to it :)

Any comments welcome!

TonisOrmisson:
Regrding subquestion relation deleting: 

this is why I was using this: 
https://github.com/LimeSurvey/LimeSurvey/pull/1073/files#diff-57d8ed613df5c805f7b2c706aca6ca52R388

all relations get deleted by using IN ([all sub-wuestion-ids])

TonisOrmisson:
I was thinking actually of doing the subquestion deleting recursively, but using all subquestion ids seemed more effective (less beautiful though)

Shnoulle:
> I was thinking actually of doing the subquestion deleting recursively, but using all subquestion ids seemed more effective (less beautiful though)

Yes really more effective :) :+1:

PR #1076, opened 2018-05-15
Deny access to hidden files and directories

RichieB2B:
Files like .editorconfig and .gitignore are accessible via HTTP. The best practice is to deny access to hidden files in the source tree.


Shnoulle:
Except it work only for apache :+1:

PR #1077, opened 2018-05-20
Update save way to add custom made EM functions (php and js)

ovillani:
Fixed issue # : 13513
New feature # : Update save way to add custom made EM functions 
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/application/extensions/emexpressions/sampleextension/sampleextension.js
+++ b/application/extensions/emexpressions/sampleextension/sampleextension.js
@@ -1,0 +1,17 @@
+// extending LEM wit js functions similar to the php file with same name
+// js functions are required for group-by-group or all-in-one presentations of surveys
+
+function sayHello(msg)
+{
+	return 'Hello ' + msg;
+}
+
+function sayBye(msg)
+{
+	return 'Good Bye ' + msg + ". " + notRegistered();
+}
+
+function notRegistered()
+{
+	return "Result is 42!";
+}
--- a/application/extensions/emexpressions/sampleextension/sampleextension.php
+++ b/application/extensions/emexpressions/sampleextension/sampleextension.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * Sample for a LEM extension to add new functions for LimeSurvey ExpressionManager
+ * Copyright (C) 2018 orvil (O. Villani)
+ * License: GNU/GPL License v2 or later, see LICENSE.php
+ * 
+ * Howto:
+ * 1. Create (if not already existing) a new folder below the path '/application/extensions/' in your LS installation named 'emexpressions'.
+ *    So path is now /application/extensions/emexpressions
+ * 
+ * 2. create in this (new) subfolder another folder with the name of your extension [in this case eg. 'sampleextension']
+ *    This is your extension-folder
+ * 
+ * 3. create or copy in your extension-folder a *.php file with exactly the same name as the folder has, but with '.php' at the end
+ *    This is your extension php file [in this case 'sampleextension.php']
+ * *
+ * 4. This extension php file has to contain a class with exactly the same name as the extension-folder / extension php file has
+ *    This is the extension class [ in this case 'sampleextension']
+ * 
+ * 5. The every extension class requires: 
+ * 		- a private variable named $RDP_ValidFunctions
+ * 		- a constructor where the additional EM functions are defined (the same way as they are in em_core_helper.php)
+ * 		- a function called 'newRDP_ValidFunctions()' where the property 'RDP_ValidFunctions' is returned
+ * 
+ * 6. In the extension php file but OUTSIDE(!) of the class definition implement those functions you need and 
+ *    that are defined for RDP_ValidFunctions
+ * 
+ * 7. Create or copy in this extension-folder a *.js file with exactly the same name as the folder has, but with '.js' at the end
+ *    This is the extension javaScript file [in this case 'sampleextension.js']
+ *    The extension javaScript file has to have equivalent functions to the additional php functions in extension php file
+ * 	  This is necesarry for surveys running in view-by-group and all-in-one mode
+ */ 
+
+ 
+// class has to have the same name as the file (except the .php at the end)
+class sampleextension
+{
+	private $RDP_ValidFunctions; // names and # params of valid functions
+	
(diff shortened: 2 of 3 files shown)


LouisGac:
I like it. But few comments: 
1. It should go to the develop branch.
2. the extension part should be written in twig for security reasons
3. if in php, maybe be it should use the plugin system?

Shnoulle:
EM function mean PHP+JS function.

Can not be twig …

I think use near same event than [newQuestionAttributes](https://github.com/LimeSurvey/LimeSurvey/blob/71031ae6a4fabe586521e9591418f117ccca8374/application/helpers/questionHelper.php#L1591) is better.

And em function is an array : 

````
'functionName'=>[
    $phpFunctionName,
    $jsFunctionName,
    $langDesciption,
    $emDescription,
    $emLink,
    $paramCount,
    $paramCount2
];
````

Maybe we can just add : 
- See if we can use autoloader with function name like this 'plugin.pluginName.'.functionName (maybe with a boolean) or something similar (for PHP)
- Allow to register a script file for new javascript function

**Comment by Shnoulle** [comment]
Another idea : add a new function in EM core: 

`pluginFunction($plugin,$functionName,$extraParams)`

And do the job in this function

**Comment by ovillani** [comment]
@Shnoulle do I understand it in the right way: your suggestion is to use the existing plugin directory /plugins/... to install the EM extension? Think it's no problem to change it to that - I'll give it a try. The advantage of this solution is of course the visibility in the admins plugin panes an the chance to switch the additional functions on and off.
But I'm not sure at the moment how to handle the suggested pluginFunction() - I'll try to find a way for that!
@LouisGac: ok, I'll do a request to the development branch if I solved the points above

**Comment by Shnoulle** [comment]
@ovillani not exactly :). 

I mean adding a new event where plugin dev can append to 'newEmFunction' like we do for newQuestionAttributes . 
- The function must return a php function, if we can call it via Yii Alias it's better
- The function must return a javascript function : then add a element for the js file to be registred.

Something like : 
'functionName'=>[
    $phpFunctionName,
    $jsFunctionName,
    $langDesciption,
    $emDescription,
    $emLink,
    $paramCount,
    $paramCount2,
    $alias=null, // alias added before call function (unusre we need one , to test : send 'myPluginname.myFunction' [here](https://github.com/ovillani/LimeSurvey/blob/a0c0f8b1198bf99e55f44fc6d4887dd0671be2e1/application/helpers/expressions/em_core_helper.php#L1883)  after set Yii::setPathOfAlias('myPluginname',__DIR__); in the plugin.
   $jsfile = null; // A js file to be registred (containing the js function)
];
About a new function inside EM it's more something similar than directRequest : 

Quick idea: 
```
pluginFunction($pluginName,$functionName,$extraParams=null) {
    $oEvent = new \LimeSurvey\PluginManager\PluginEvent('registerEmFunction');
    $oEvent->set('target',$pluginName);
    $oEvent->set('function',$functionName);
    $oEvent->set('params',$extraParams);
    App()->getPluginManager()->dispatchEvent($oEvent);
    $return = $oEvent->get('return');
    /* I have no idea on the way to treat this part */
}
```

LouisGac:
Twig == PHP

anything done in PHP can be done in twig.

Shnoulle:
Twig == PHP WITH restriction … if you want to add new function, must be added in [Twig allowed function](https://github.com/LimeSurvey/LimeSurvey/blob/51e827e030905d5ab78e64466fad2e658f4b8106/application/config/internal.php#L213)

I know plugin can update config, but unsure it's what you want …

LouisGac:
yes that why I said: do it with plugins if PHP
a new extension system for that is ok, if and only if it's done via twig, so it offers a higer security level than a plugin.

Shnoulle:
I don't understand what relation you have between twig file and https://manual.limesurvey.org/Expression_Manager#Access_to_Functions .

An expression manager function return a value, not an HTML string …

LouisGac:
sampleextension.php could be written in twig, so it would be secured, and people could install those kind of extension without any security issue. 

else, it should use the plugin system.

ovillani:
If it's possible using the existing plugin system would have the advantage that 
- no  additional directories are required
- an admin can see it being installed at the plugin panel

Anyhow I can't see the security issues in this case, 
1. because a plugin is always written in php too
2. not everybody is allowed to install a plugin, only those having ftp access to upload a plugin (meaning: if there are second thoughts on security, it's the admin itself ;) 
3. plugins like addScriptToQuestion are either distributed by the LS Homepage (==checked ?!) or downloaded from github etc (==you have to know what you are doing) or self developed (===you have to know what you are doing)

Shnoulle:
Current code have same issue, no ? 

Admin must upload a php file in `/extensions/emexpressions`

ovillani:
@shnoulle: yes upload is necessary, but not an issue. It's the homedirectory for the emexpressions i choose as a first idea - but I  the idea to place them in the plugins directory is better.

ovillani:
Just wondering for hot to put it in a plugin: can you give me an idea witch event is called every time 
- AFTER class ExpressionManager is initialized and
- BEFORE a survey/question is rendered

Shnoulle:
I think adding a new event in EM core constructor directly:

https://github.com/LimeSurvey/LimeSurvey/blob/23eaf7e094fd118ea1df78772ae1b935a688ac53/application/helpers/expressions/em_core_helper.php#L249

Why not ?

ovillani:
@Shnoulle: It's ok for me -  but: any suggestion where I can find a pattern in the code for learning how to implement a new event? Would speed things up - except someone else has time to implement that :)
P.S.: name for that event? newExpressionRegister? onExpressionRegister? ...

Shnoulle:
I make a pull request next week (give formation this week), but insure it work (put in WIP). And need to create a demo plugin :) . You can help me for improving/testing :)

ovillani:
@Shnoulle, that's great! & of course I'll help improving/testing. My pleasure ;)

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1081

Shnoulle:
Ok, solution in dev.

And for twig : current solution can be improved to allow twig file + package : https://github.com/LimeSurvey/LimeSurvey/pull/1081#issuecomment-394603961

PR #1078, opened 2018-05-22
New feature : Allow plugin to add or replace twig file

Shnoulle:
See usage https://gitlab.com/SondagesPro/registerQuick for example.

Need add for https://gitlab.com/SondagesPro/maintenanceMode

Have a plugin if not merged : https://gitlab.com/SondagesPro/twigExtendByPlugins (why all $_wtig is private in ETwigViewRenderer ? Someone see the reason ?

### DIFF
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -1884,14 +1884,20 @@ function checkCompletedQuota($surveyid, $return = false)
     $thissurvey['aQuotas']['sPluginBlocks']      = implode("\n", $blocks);
     $thissurvey['aQuotas']['sUrlDescription']    = $sUrlDescription;
     $thissurvey['aQuotas']['sUrl']               = $sUrl;
+    $thissurvey['active']                        = 'Y';
+    
 
-    $thissurvey['aQuotas']['hiddeninputs'] = '<input type="hidden" name="sid"      value="'.$thissurvey['sid'].'" />
+    $thissurvey['aQuotas']['hiddeninputs'] = '<input type="hidden" name="sid"      value="'.$surveyid.'" />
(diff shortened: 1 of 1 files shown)


Shnoulle:
@LouisGac : do you think plugin must be allowed to add [element on template editor](https://github.com/ecraven/LimeSurvey/blob/4d5ed420f106d40fe3e9c5693793dcd866c5dc22/themes/survey/vanilla/config.xml#L87). If yes : where di you think it's best to add this elements event ? In theme manager controller or one of  Template model ?

Shnoulle:
Still an issue about layout … layout didn't use twig path but only template layout views …

LouisGac:
theme editor is pretty messy. But just look where we get data from the XML file, it should be a perfect place to add plugin events.

Shnoulle:
And why not allowing Template OR plugin OR config to add path to twig loader ?

About layout : we don't use $this->_twig->getLoader()->paths … it's sad. Do you think we can, directly use twig here ? 

See [here](https://gitlab.com/SondagesPro/renderMessage/blob/master/renderMessageETwigViewRenderer.php#L224) , maybe replacing [this part](https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/7343ac4529393993a5a4a0f67ccc735b9709393a/application/core/LSETwigViewRenderer.php#L41) byè something similar ?

LouisGac:
it is just what I said: it should be a perfect place to add plugin events. 
...and we're using twig path ;-) 
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSETwigViewRenderer.php#L285

Shnoulle:
Yes, inside twig , but not for layout https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/7343ac4529393993a5a4a0f67ccc735b9709393a/application/core/LSETwigViewRenderer.php#L41

PR #1079, opened 2018-05-24
Add an empty footer to the vanilla theme.

ecraven:
This commit adds a file footer.twig, includes it in the relevant layout_*.twig, and includes a few lines of CSS to bottom-align the footer with css.

The CSS can be dropped, it could be added by the user, but the twig file cannot easily be added to extended themes (especially not to config.xml) :-/

Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/themes/survey/vanilla/views/subviews/footer/footer.twig
+++ b/themes/survey/vanilla/views/subviews/footer/footer.twig
@@ -1,0 +1,19 @@
+{#
+Add your footer in this file.
+
+You will need to add custom CSS to make this work, for example, the following
+should work for a bottom-aligned footer in all browsers that properly support
+flex:
+
+body {
+     padding-bottom: 0; /* no bottom padding, footer should be bottom aligned */
+     padding-top: 60px;/* now is redefine in JS to fit any title length */
+     flex-direction: column; /* article, then footer */
+     display: flex; /* flex, to extend the content to full size */
+     min-height: 100vh; /* full height, to bottom-align footer */
+}
+body > article {
+    flex: 1; /* the main article should use as much space as possible */
+}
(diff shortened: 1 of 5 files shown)


themes/survey/vanilla/views/subviews/footer/footer.twig
```diff
@@ -0,0 +1,3 @@
+<footer>
```
Shnoulle: I think it must be really empty : `!-- You can add your own footer here -->` or directly ``.

An [empty tag](https://stackoverflow.com/questions/560880/empty-html-tags) must be removed if we can

themes/survey/vanilla/css/theme.css
```diff
@@ -29,8 +29,14 @@
 */
 
 body {
-     padding-bottom: 10px;
+     padding-bottom: 0; /* no bottom padding, footer should be bottom aligned */
      padding-top: 60px;/* now is redefine in JS to fit any title length */
+     flex-direction: column; /* article, then footer */
+     display: flex; /* flex, to extend the content to full size */
```
Shnoulle: Using flex is a template choice, i didn't say it's a bad idea, but i don't think it must come from default lighter template.

ecraven:
I've modified the PR. No CSS is added, footer.twig is empty, except for a comment that shows what the CSS might look like.

ecraven:
Is there anything else I can do to help in merging this?

LouisGac:
ecraven: don't worry, we'll merge it. We've been very busy last weeks (moving to new office).

Shnoulle:
\o/ :)

PR #1080, opened 2018-05-28
FIX: endless redirections in theme setting menu

gekkedev:
If users view the theme settings they will face endless redirections to the same page. I've pointed the redirection back to the survey overview page.


LouisGac:
thanks

PR #1083, opened 2018-06-02
Updated feature: Added token_screenout to returned fields of API's get_summary() method

martinplattnr:
Fixed issue # : 
New feature # :
Changed feature # : 13737
Dev: Trivial change, as the new field token_screenout was fetched by the API method already using Token->summary(). It was just not returned - I guess because the screenout was added to Token->summary() after the implementation of API's get_summary() and no one updated it.

Report is at https://bugs.limesurvey.org/view.php?id=13737 

This is my first PR, I tried to follow your contribution guideline and hope it's ok.


LouisGac:
thank you

PR #1084, opened 2018-06-09
[READY] New feature : allow plugin to add or remove file for template editor

Shnoulle:
- this allow us to put twig files in core in template editor too : for example, currently question twig file
- todo : same for js and css file … (from package ? Using `->packages->css and ->packages->js` ?)
- I send a little demo plugin allowing hide firstname/lastname + show token form (quickRegister have send mail function)


Shnoulle:
@olleharstedt did you think we add a «admin theme twig editor» in 5.0 ? If yes : maybe need to review https://github.com/LimeSurvey/LimeSurvey/pull/1084/files#diff-9e735704a8412da2764aa0873fb8619dR529 to have oTemplate to current or false (don't load current one)

application/models/TemplateConfig.php
```diff
@@ -111,17 +111,22 @@ public function getTemplateURL()
...
+         * @param boolean $force file to be in template or mother template
+         * @return TemplateConfig|null|void
          */
-        public function getTemplateForFile($sFile, $oRTemplate)
+        public function getTemplateForFile($sFile, $oRTemplate,$force=true)
         {
             while (!file_exists($oRTemplate->path.$sFile) && !file_exists($oRTemplate->viewPath.$sFile) && !file_exists($oRTemplate->filesPath.$sFile)) {
                 $oMotherTemplate = $oRTemplate->oMotherTemplate;
                 if (!($oMotherTemplate instanceof TemplateConfiguration)) {
+                    if(!$force && Yii::app()->twigRenderer->getPathOfFile($sFile)) {
```
Shnoulle: Here, it's only valid for twig file, broke for other (js/css and «other») but currently use $force to false for twig file.

Some alternative solution : getTemplateForFile($sFile, $oRTemplate,$filtetype=false)

1. phpDoc : @param boolean|string search file type on other way, uninstall if false or not find
2. Find the way to search on package for css/js, and maybe picture
3. Find something for other file (unsure ? : when we have asset for survey ?)

Shnoulle:
Argl … since https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/d6913b8521f8ba5dabcb266d3d9e7188a265ad74/application/views/admin/themes/templatesummary_view.php#L111 we didn't have file type (css/js/view) make a test according to file type is harder …

Shnoulle:
The issue in Code factor is 2 new TODO

PR #1085, opened 2018-06-22
Joins the gettext text messages into one line because xgettext cannot…

imacat:
Hi.  I found that multi-line text messages do not exist in the translation PO files.  The reason is that the xgettext tool from GNU gettext can only process multi-line source text messages in the C language format, to concatenate the text lines with spaces as in C, but not with the string concatenation operator in PHP (.).  Hence several pieces of text in the tutorial do not appear in the PO file.

Search for "The most important settings of your survey can be reached from this sidebar:" in Pootle gets nothing:

https://translate.limesurvey.org/projects/limesurvey-3/zh-tw/traditional/?filters%5Bterm%5D=The+most+important+settings+of+your+survey+can+be+reached+from+this+sidebar%3A&filters%5Buser_login%5D=&filters%5Bstatus%5D=either&filter=Filter&sort%5Bby%5D=translation_date_added&sort%5Bhow%5D=asc

While it appears in application/core/LsDefaultDataSets.php line 464:

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LsDefaultDataSets.php#L464

Fixed issue # : 13805

https://bugs.limesurvey.org/view.php?id=13805

New feature # :
Changed feature # :
Dev: 
Dev:


lacrioque:
Thank you for your pull request!
We will merge it shortly. For the future can you check on our commit message format?
Here is a link: https://manual.limesurvey.org/Standard_for_Git_commit_messages
We try to check pull requests regularly, but sometimes it takes us a while.
Please feel free to do more pull requests!

PR #1086, opened 2018-06-25
[READY] New feature : allow multiple button and language changer in template

Shnoulle:
Dev: removing uneeded id

### DIFF
--- a/tests/questions/MultipleChoiceNextPreviousTest.php
+++ b/tests/questions/MultipleChoiceNextPreviousTest.php
@@ -58,20 +58,18 @@ class MultipleChoiceNextPreviousTest extends TestBaseClassWeb
             $lis[0]->click();
 
             // Click next.
-            $submit = self::$webDriver->findElement(WebDriverBy::id('ls-button-submit'));
-            $submit->click();
+            $submit = self::$webDriver->next();
 
             // Click previous..
-            $prev = self::$webDriver->findElement(WebDriverBy::id('ls-button-previous'));
+            $prev = self::$webDriver->findElement(WebDriverBy::cssSelector('[value="moveprev"]'));
             $prev->click();
             sleep(1);  // TODO: Does not work without this.
 
             // Click next.
-            $submit = self::$webDriver->findElement(WebDriverBy::id('ls-button-submit'));
-            $submit->click();
+            $submit = self::$webDriver->next();
 
             // Click previous..
-            $prev = self::$webDriver->findElement(WebDriverBy::id('ls-button-previous'));
+            $prev = self::$webDriver->findElement(WebDriverBy::cssSelector('[value="moveprev"]'));
             $prev->click();
 
             // Check value of checkbox.
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,4 +71,4 @@ before_script:
 
 script:
     # Run tests.
-    - DOMAIN=localhost phpunit
+    - DOMAIN=localhost phpunit --stop-on-failure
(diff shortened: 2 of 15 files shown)


Shnoulle:
@olleharstedt Can you review the test part ? I'm unbsure i found all related test. I `grep -r 'ls-button-submit'` and `grep -r 'ls-button-previous'`?

Strangely , no issue at first commit https://travis-ci.org/LimeSurvey/LimeSurvey/jobs/396333142 except the ajaxredirect …

tests/questions/MultipleChoiceNextPreviousTest.php
```diff
@@ -58,20 +58,18 @@ public function testNextPrevious()
...
             sleep(1);  // TODO: Does not work without this.
 
             // Click next.
-            $submit = self::$webDriver->findElement(WebDriverBy::id('ls-button-submit'));
-            $submit->click();
+            $submit = self::$webDriver->next();
 
             // Click previous..
-            $prev = self::$webDriver->findElement(WebDriverBy::id('ls-button-previous'));
+            $prev = self::$webDriver->findElement(WebDriverBy::cssSelector('[value="moveprev"]'));
```
Shnoulle: Unsure of way cssSelector work if there are multiple button, maybe need '[value="moveprev"]:first'

This was not an issue with core template since we use only one button.

olleharstedt:
Won't removing ID possibly break any current JS snippets/work-arounds?

Shnoulle:
If it's workaround : we don't support it. And LS 3 already broke a lot of workaround. And a lot of workaround broken in 2.5X to 2.7X …

Shnoulle:
Looking at https://manual.limesurvey.org/Workarounds:_Manipulating_a_survey_at_runtime_using_Javascript : all workaround use old id (movenext/movesubmit)

And we still have old css class name for input and button here : https://manual.limesurvey.org/Theme_editor#Input.2FButtons

olleharstedt:
Ok.

Shnoulle:
Best is to see with @tpartner :)

Shnoulle:
Oups, moved to WIP to add same class , but we already have specific class : https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/12100afaef9cc21e55a4ceb860b7fa96b3b18c02/application/models/TemplateConfig.php#L748 then READY again

tpartner:
Workarounds aside, this will render any advice given in the forums regarding manipulating the Next button in 3.x useless. Searching the forums for "ls-button-submit" yields many results. 

I thought 3.x was to have stable HTML.

Shnoulle:
I found 5 topics about usage of ls-button-submit.

In first alpha : button didn't have id (due to a forum issue where user want multiple movenext in template).

tpartner:
I don't know how many topics but I find at least 15 references. That's not the point. I thought the HTML was supposed to be stable. (unless to fix bugs, which this is not)

Having said that, I don't disagree with the idea of this feature. After 2.x, I do disagree with changing HTML in the "stable" versions.

Shnoulle:
OK, moving to develop, maybe i do it directly in develop (more simple to do it again)

tpartner:
Thanks.

Shnoulle:
Still a better usage to use `:submit:[move='next']` class selectro for next advice (sill valid for 4.0 :) )

maziminke:
I also agree to NOT change any basic HTML details. There are more JS workarounds out there then we can imagine.

Shnoulle:
Still need allowing easily adding more then one submit button in a page …

Currently : there are no easy way to do this. You don't want it ?  And don't tell me js workaround was not updated for 3.0 …

The real issue here : it's done before beta release, but removed after.

PR #1090, opened 2018-07-01
Fixed issue #13821: When importing a survey I receive a "Failed to In…

githubLewis:
…sert [3]" error message

This issue is caused by a change in the sql pdo driver, without this fix, new ids aren't returned from tables with identity columns in mssql. This affects any objects that use Identity columns, such as questions, assessments. This has the effect in the application to prevent questions from saving correctly either by creating new ones or via survey import.

This fix is copied from the yii 1.19-dev label, here: https://github.com/yiisoft/yii/commit/7b768c304289ede574768de68939224d1180e2b1

Consideration should be given to updating the whole framework.


Shnoulle:
> Consideration should be given to updating the whole framework.

Good idea, but this one is not in 1.1.19 release : https://github.com/yiisoft/yii/tree/1.1.19

Thanks you :)

PR #1092, opened 2018-07-07
Fixed issue #13845: Can't export and then import tab-separated text s…

Shnoulle:
…urvey file
Dev: real fix, number only in subquestion is allowed

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -2416,10 +2416,7 @@ function TSVImportSurvey($sFullFilePath)
                     }
                     // Insert sub question and keep the sqid for multi language survey
                     $question = new Question();
-                    //$question->attributes = $insertdata; : this broke call of $this when test rules when save : PLEASE : don't use
-                    foreach($insertdata as $attribute=>$value) {
-                        $question->setAttribute($attribute,$value);
-                    }
+                    $question->setAttributes($insertdata,false); //$question->attributes = $insertdata; : this broke call of $this when test rules when save : PLEASE : don't use
                     if (!$question->save()) {
                         $results['error'][] = gT("Error")." : ".gT("Could not insert subquestion").". ".gT("Text file row number ").$rownumber." (".$sqname.")";
                         break;
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -2390,8 +2390,8 @@ function TSVImportSurvey($sFullFilePath)
                         }
                     }
                 } else {
-                    $insertdata = array();
                     $scale_id = (isset($row['type/scale']) ? $row['type/scale'] : 0);
+                    $insertdata = array();
                     $insertdata['sid'] = $iNewSID;
                     $insertdata['gid'] = $gid;
                     $insertdata['parent_qid'] = $qid;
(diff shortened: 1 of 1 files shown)


Shnoulle:
This is the real fix , https://github.com/LimeSurvey/LimeSurvey/commit/00990367a1e06dd0a4773c9a48f1c820f9c456bc must be reverted.

olleharstedt:
Cool fix, will try on Monday. Why are you working on a Saturday? :O :O

application/helpers/admin/import_helper.php
```diff
@@ -2416,10 +2416,7 @@ function TSVImportSurvey($sFullFilePath)
                     }
                     // Insert sub question and keep the sqid for multi language survey
                     $question = new Question();
-                    //$question->attributes = $insertdata; : this broke call of $this when test rules when save : PLEASE : don't use
-                    foreach($insertdata as $attribute=>$value) {
-                        $question->setAttribute($attribute,$value);
-                    }
+                    $question->setAttributes($insertdata,false); //$question->attributes = $insertdata; : this broke call of $this when test rules when save : PLEASE : don't use
```
TonisOrmisson: nice!

Shnoulle: Maybe, but in general : must move from `$object->attributes = array()` to $object->setAttributes(array(),(true|false));`

I don't search on the file if we have another …

Here : it'sba bug fix, since other are not bug fix : i didn't touch […](https://github.com/LimeSurvey/LimeSurvey/commit/d67f292dfabd747ff5cfc97a3cd35217732084bb#commitcomment-29528163)

application/helpers/admin/import_helper.php
```diff
@@ -2416,7 +2416,7 @@ function TSVImportSurvey($sFullFilePath)
                     }
                     // Insert sub question and keep the sqid for multi language survey
                     $question = new Question();
-                    $question->attributes = $insertdata;
+                    $question->setAttributes($insertdata,false); //$question->attributes = $insertdata; : this broke call of $this when test rules when save : PLEASE : don't use
```
LouisGac: By-passing validation is not a real fix. I still prefer Dominik approach: at least by reading his code any we understand what the problem is and what the fix is doing. 
I'd say a "real" fix would be to change the subquestion's code generation for ranking question  so it respects the rule (what dominik started to do), or to change the rule.

Shnoulle: I don't bypass validation … I respect the rules : see the bug report PLEASE ! 

And yes : i'm angry here …

olleharstedt: Only two attributes are set to safe, no? https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Question.php#L114

olleharstedt: Sorry, I'm wrong. All attributes that have a validation rule are safe in Yii.

LouisGac: Ok, we reviewed everything a last time here. The problem is that exporting to tsv create subquestion being wrong in a way or another making the question rules to apply. Dominik will do a "real" fix today: he will call the lss generator (which works fine), and convert the result to TSV. So we'll have a single logic for both format.

Shnoulle: And there are the second issue : even valid TSV didn't import …

olleharstedt: There are multiple other issues with TSV export/import, for example number of question attributes being different (or maybe your fix fixed that too?).

Shnoulle:
About ranking : attacched export file from 2.73 + screenshot after import with this pull request
capture du 2018-07-09 19-18-43
[limesurvey_survey_rank.txt](https://github.com/LimeSurvey/LimeSurvey/files/2177277/limesurvey_survey_rank.txt)

I don't say TSV must not be rewritten , i say : The previous version was tested used again and again … the rewriting is not needed in 3.X, but must be done in 4.0

The only know issue was fixed here. If there are another : where is the mantis ?

PR #1095, opened 2018-07-14
[READY] Fixed issue #13877: Unable to update answers twig part in theme

Shnoulle:
Dev: using twig like we need to use it
Dev: must update for question template (but it can be done)

### DIFF
--- a/application/core/LSETwigViewRenderer.php
+++ b/application/core/LSETwigViewRenderer.php
@@ -125,8 +125,7 @@ class LSETwigViewRenderer extends ETwigViewRenderer
     {
         $this->_twig  = parent::getTwig(); // Twig object
         $loader       = $this->_twig->getLoader(); // Twig Template loader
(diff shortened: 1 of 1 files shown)


Shnoulle:
I just need to update a little for question template. Surely add an optionnable part on https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/ece310e5413ffdc0733b6b7df0b800ca7f9dfb34/application/core/LSETwigViewRenderer.php#L378

Something like `@param string $directory  specific directory for question` or `@param \QuestionTemplate` just need to see the best solution

Shnoulle:
OK, ready and tested. There are no clean solution (currently) to allow replacing QuestionTemplate twig files by template file : since directory (after basePath) are the same for core and template file.

Why not update this ? Directory not same for QuestionTemplate and other question.

PS : if not in LS core : i make it with plugin …

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1096

PR #1096, opened 2018-07-14
[READY] Fixed issue + feature : template twig files can replace any twig files (except Plugin forced)

Shnoulle:
Fixed issue #13877: Unable to update answers twig part in theme
New feature : Allow survey theme to update question template twig


Shnoulle:
Follow this https://github.com/LimeSurvey/LimeSurvey/pull/1095 (is child of this, no need the 2 pull request accepte, only one)

Shnoulle:
Previous commit from https://github.com/LimeSurvey/LimeSurvey/pull/1095 are in this one. 

capture du 2018-07-14 15-38-42

Don't like to add a `is_dir` but since it was only if Question template is set, not done a lot.

PR #1098, opened 2018-07-16
Dev : fix for usergroup view when admin

papa-mb:
Fixed issue # : 
New feature # :
Changed feature # :
Dev: fix for usergroup view when admin
Dev:

### DIFF
--- a/application/models/UserGroup.php
+++ b/application/models/UserGroup.php
@@ -405,15 +405,13 @@ class UserGroup extends LSActiveRecord
 
         $criteria->join .= 'LEFT JOIN {{users}} AS users ON ( users.uid = t.owner_id )';
 
-        if (!Permission::model()->hasGlobalPermission('superadmin', 'read')) {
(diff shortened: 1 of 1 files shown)


papa-mb:
Hi, 
It's just a fix for a bug in the view of groups when you are superadmin. There was a message that said that there's a call to a non object

lacrioque:
You removed a check if the user is not a superadmin, this is a rather critical call.
Can you maybe change it to check if the permission model exists instead of removing the call to it?

Shnoulle:
> You removed a check if the user is not a superadmin, this is a rather critical call.

In fact : this disable super-admin to see all survey … if super admin didn't create an user and not in his group : he don't see this user survey …

@papa-mb : best is to tell us exactly where you have this issue …

c-schmitz:
Closed due to lack of feedback and that the reason for this whole patch is uknown.

PR #1101, opened 2018-07-25
Fixed issue #13910: Unable to use saveall button (an other button)

Shnoulle:
Dev: for default action : see https://github.com/LimeSurvey/LimeSurvey/blob/ed9fc2b6455b0c09c6672dcd1a1d130139838e6b/themes/survey/vanilla/views/subviews/header/start_form.twig#L71
Dev: then same behaviour with and without js

### DIFF
--- a/themes/survey/vanilla/scripts/ajaxify.js
+++ b/themes/survey/vanilla/scripts/ajaxify.js
@@ -111,7 +111,7 @@ var AjaxSubmitObject = function () {
         // Restrict to [type=submit]:not([data-confirmedby])
         // - :submit is the default if button don't have type (reset button on slider for example),
         // - confirmedby have their own javascript system
-        $(document).on('click', '#ls-button-submit, #ls-button-previous', function (e) {
+        $(document).on('click', '#limesurvey [type=submit]:not([data-confirmedby])', function (e) {
             $('#limesurvey').append('<input id="onsubmitbuttoninput" name=\'' + $(this).attr('name') + '\' value=\'' + $(this).attr('value') + '\' type=\'hidden\' />');
             if (isIE10 || /Edge\/\d+\.\d+/.test(navigator.userAgent)) {
                 e.preventDefault();
@@ -123,6 +123,7 @@ var AjaxSubmitObject = function () {
         // If the user try to submit the form
         // Always bind to document to not need to bind again
         $(document).on('submit', '#limesurvey', function (e) {
+            console.log(activeSubmit);
             // Prevent multiposting
             //Check if there is an active submit
             //If there is -> return immediately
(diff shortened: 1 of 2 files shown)


Shnoulle:
Issue here https://bugs.limesurvey.org/view.php?id=13860
Answer is «advanced customization» … really ?

themes/survey/vanilla/scripts/ajaxify.js
```diff
@@ -111,7 +111,7 @@ var AjaxSubmitObject = function () {
         // Restrict to [type=submit]:not([data-confirmedby])
         // - :submit is the default if button don't have type (reset button on slider for example),
         // - confirmedby have their own javascript system
-        $(document).on('click', '#ls-button-submit, #ls-button-previous', function (e) {
+        $(document).on('click', '#limesurvey [type=submit]:not([data-confirmedby]:not(disabled))', function (e) {
```
lacrioque: Why you are so keen on having this selector? I still strongly oppose this.
It is not specific enough and the effects of pagetriggers should always only be done by page triggers.
One way around it is to add a specific selector class just for this kind of purpose? Would you be ok with that?
It is a simple change, just use i.g. .selector__trigger-page-submit to select the triggering buttons.

Shnoulle: A data , not a class. data-action='submit' for example. But i really think the purpose (action='submit') of a button must launch submit action … this action was done without js, then same action must be done with js …

Currently : something working on 2.73 + answer_html + 2.6lts didn't work in master. 

And the current system for disable multiple submit didn't work : https://bugs.limesurvey.org/view.php?id=13911

Shnoulle:
Javascript must add feature, not remove feature …

If an action can be done without javascript : it mus can be done with javascript. But an action doable only with javascript can be OK (doable without js is better, but not mandatory).

themes/survey/vanilla/scripts/ajaxify.js
Shnoulle: Yes, and … deactivate ajaxmode : click on Index button : it work. Activate ajaxmode, click on index button : it don't work …

LouisGac:
I didn't follow all your debate about the selector, but remember this: it's possible to create a theme without ajax and just using the EM js. I did it before 3.0 release. I will redo it and add it to premium theme.

Shnoulle:
Yes @LouisGac : i know : i can redo a template from scratch, like i do getSleleton and fix a lot of part in it. But still : i think LimeSurvey core must offer the best user experience by default …

LouisGac:
sure.

Shnoulle:
OK, then we leave broken system in LS core …

PR #1104, opened 2018-07-30
Add a new dropdown 'token type' for numeric tokens.

pefeigl:
This commit adds a new selection to let you choose between the normal tokens and numeric-only tokens.

maim

maim2

New feature # :
Dev: 
Dev:

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -691,6 +691,7 @@ class tokens extends Survey_Common_Action
 
             $aData['thissurvey'] = getSurveyInfo($iSurveyId);
             $aData['surveyid'] = $iSurveyId;
+            $aData['tokenType'] = !empty(Token::model($iSurveyId)->survey->tokentype) ? Token::model($iSurveyId)->survey->tokentype : 'default';
             $aData['iTokenLength'] = !empty(Token::model($iSurveyId)->survey->tokenlength) ? Token::model($iSurveyId)->survey->tokenlength : 15;
 
             $aData['sidemenu']['state'] = false;
@@ -915,6 +916,7 @@ class tokens extends Survey_Common_Action
 
             $amount = (int) Yii::app()->request->getPost('amount');
             $iTokenLength = (int) Yii::app()->request->getPost('tokenlen');
+            $tokenType = (int) Yii::app()->request->getPost('tokentype');
 
             // Fill an array with all existing tokens
             $existingtokens = array();
@@ -941,7 +943,7 @@ class tokens extends Survey_Common_Action
 
                 $attempts = 0;
                 do {
-                    $token->token = Token::generateRandomToken($iTokenLength);
+                    $token->token = Token::generateRandomToken($iTokenLength, $tokenType);
                     $attempts++;
                 } while (isset($existingtokens[$token->token]) && $attempts < 50);
 
@@ -975,11 +977,13 @@ class tokens extends Survey_Common_Action
 
         } else {
             $iTokenLength = !empty(Token::model($iSurveyId)->survey->tokenlength) ? Token::model($iSurveyId)->survey->tokenlength : 15;
+            $tokenType = !empty(Token::model($iSurveyId)->survey->tokentype) ? Token::model($iSurveyId)->survey->tokentype : 'default';
 
             $thissurvey = getSurveyInfo($iSurveyId);
             $aData['thissurvey'] = $thissurvey;
             $aData['surveyid'] = $iSurveyId;
             $aData['tokenlength'] = $iTokenLength;
+            $aData['tokentype'] = $tokenType;
             $aData['dateformatdetails'] = getDateFormatData(Yii::app()->session['dateformat'], App()->language);
             $aData['aAttributeFields'] = getParticipantAttributes($iSurveyId);
             $this->_renderWrappedTemplate('token', array('dummytokenform'), $aData);
@@ -2356,6 +2360,7 @@ class tokens extends Survey_Common_Action
         }
 
         $aData['iTokenLength'] = !empty(Token::model($iSurveyId)->survey->tokenlength) ? Token::model($iSurveyId)->survey->tokenlength : 15;
+        $aData['tokenType'] = !empty(Token::model($iSurveyId)->survey->tokentype) ? Token::model($iSurveyId)->survey->tokentype : 'default';
 
         $thissurvey = getSurveyInfo($iSurveyId);
         $aAdditionalAttributeFields = $thissurvey['attributedescriptions'];
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -41,6 +41,7 @@ use \LimeSurvey\PluginManager\PluginEvent;
  * @property string $allowregister Allow public registration (Y/N)
  * @property string $allowsave Is participant allowed save and resume later (Y/N)
  * @property integer $autonumber_start
+ * @property string $tokentype Token type: 'default' or 'numeric'
  * @property integer $tokenlength Token length: MIN:5 MAX:36
  * @property string $autoredirect Automatically load URL when survey complete: (Y/N)
  * @property string $allowprev Allow backwards navigation (Y/N)
@@ -472,6 +473,7 @@ class Survey extends LSActiveRecord
             array('format', 'in', 'range'=>array('G', 'S', 'A'), 'allowEmpty'=>true),
             array('googleanalyticsstyle', 'numerical', 'integerOnly'=>true, 'min'=>'0', 'max'=>'2', 'allowEmpty'=>true),
             array('autonumber_start', 'numerical', 'integerOnly'=>true, 'allowEmpty'=>true),
+            array('tokentype', 'in', 'range'=>array('default','numeric'), 'allowEmpty'=>true),
             array('tokenlength', 'default', 'value'=>15),
             array('tokenlength', 'numerical', 'integerOnly'=>true, 'allowEmpty'=>false, 'min'=>'5', 'max'=>'36'),
             array('bouncetime', 'numerical', 'integerOnly'=>true, 'allowEmpty'=>true),
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -193,10 +193,11 @@ abstract class Token extends Dynamic
     public function generateToken()
     {
         $iTokenLength = $this->survey->tokenlength;
-        $this->token = $this::generateRandomToken($iTokenLength);
+        $tokenType = $this->survey->tokentype;
+        $this->token = $this::generateRandomToken($iTokenLength, $tokenType);
         $counter = 0;
         while (!$this->validate(array('token'))) {
-            $this->token = $this::generateRandomToken($iTokenLength);
+            $this->token = $this::generateRandomToken($iTokenLength, $tokenType);
             $counter++;
             // This is extremely unlikely.
             if ($counter > 10) {
@@ -211,9 +212,13 @@ abstract class Token extends Dynamic
      * @param integer $iTokenLength
      * @return string
      */
-    public static function generateRandomToken($iTokenLength)
+    public static function generateRandomToken($iTokenLength, $tokenType)
     {
-        return str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
+        if($tokenType == 'numeric') {
+            return randomChars($iTokenLength, '123456789');
+        } else {
+            return str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
+        }
     }
 
     /**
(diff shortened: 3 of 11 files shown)


Shnoulle:
Such feature mus be in plugin

pefeigl:
> Such feature must be in plugin

Can this be done in a plugin?

Shnoulle:
Except with some hack, i don't think currently.

lacrioque:
Hey, thanks a lot.
We like this change and we will merge it, but not just now, we will wait for after the next release.
We get asked for numerical tokens often, so thank you very much for the effort made!

pefeigl:
Great to hear that! Please just ping me when you are considering merging this, then I'll rebase it and make sure it still works. One caveat: The random generator for the numbers is just `randomChars`, which might be unsafer than the `default` one. If necessary, I can look into changing that.

dominikvitt:
This is a great feature.
Can you close this one and create a new pull request in develop branch?
All new features go there.
As soon as you are finished, we'll merge it.

TonisOrmisson:
I would like to have a "startFrom' value there. Typically I like having similar tokens with the same length so I would like to start tokens eg from 10000,

Shnoulle:
@TonisOrmisson it's exacly why i ask it to be in plugin … Create an event + a core plugin to set it as number allow creating another plugin to have 

- start from plugin
- fixed part plugin (for example test{number})
- only capitals plugin
- … … lot of possibility.

pefeigl:
I'm willing to port the patch to the current dev branch, but could you
please first discuss whether it will be accepted or not? [not as a
plugin, but as a part of core, as it is still impossible to write as a
plugin at this point, I think].

Shnoulle:
> it is still impossible to write as a plugin at this point

I confirm, personnaly i ask for **a new event in core** to allow adding this.

application/models/Survey.php
```diff
@@ -472,6 +473,7 @@ public function rules()
             array('format', 'in', 'range'=>array('G', 'S', 'A'), 'allowEmpty'=>true),
             array('googleanalyticsstyle', 'numerical', 'integerOnly'=>true, 'min'=>'0', 'max'=>'2', 'allowEmpty'=>true),
             array('autonumber_start', 'numerical', 'integerOnly'=>true, 'allowEmpty'=>true),
+            array('tokentype', 'in', 'range'=>array('default','numeric'), 'allowEmpty'=>true),
```
TonisOrmisson: I would suggest using constants for the types. Much easier to refer to later and refactor if needed

application/models/Token.php
```diff
@@ -211,9 +212,13 @@ public function generateToken()
      * @param integer $iTokenLength
      * @return string
      */
-    public static function generateRandomToken($iTokenLength)
+    public static function generateRandomToken($iTokenLength, $tokenType)
     {
-        return str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
+        if($tokenType == 'numeric') {
+            return randomChars($iTokenLength, '123456789');
+        } else {
```
TonisOrmisson: no `else` is needed here if true above will return

application/views/admin/survey/subview/accordion/_tokens_panel.php
```diff
@@ -36,6 +36,20 @@ function alertPrivacy()
 <div id='tokens-panel' class="container-fluid">
     <div class="row">
         <div class="col-sm-12 col-md-6">
+            <!-- Set token type -->
+            <div class="form-group">
+                <label class=" control-label" for='tokentype'><?php  eT("Select token type:"); ?></label>
+                <div class="">
+                    <select name='tokentype' id='tokentype' class="form-control">
```
TonisOrmisson: please use CHtml::dropDownList here. That's why we have a framework for

maziminke:
I like this feature. While using a plugin is nice to separate code, I consider this a core feature which should go into LS 4 directly.

Shnoulle:
@maziminke : The reason why i think it's best by event is allowing forcing to anythging else … One user can want number only, one another uppercase only etc …

dominikvitt:
@pefeigl: Please port this PR to the current develop branch.
It should be developer as core feature and all suggestions mentioned above implemented (if possible).

Shnoulle:
@dominikvitt : then need another Pull request for [Omit ambiguous characters from token](https://bugs.limesurvey.org/view.php?id=15026)

Again and agin : best solution : add an event for plugin to do anything with token …

olleharstedt:
@Shnoulle Why would this be put in a plugin instead of core?

application/controllers/admin/tokens.php
```diff
@@ -691,6 +691,7 @@ public function addnew($iSurveyId)
 
             $aData['thissurvey'] = getSurveyInfo($iSurveyId);
             $aData['surveyid'] = $iSurveyId;
+            $aData['tokenType'] = !empty(Token::model($iSurveyId)->survey->tokentype) ? Token::model($iSurveyId)->survey->tokentype : 'default';
```
olleharstedt: When ported to LS4, you can use the `??` operator here.

tammoterhark:
Just to make things more complicated.

From a user point of view: please put this in core and not in a plugin.

tokens

Example from LastPass.

I would already be very glad with the (default) omission of ambiguous charaters, for people that have to fill in by hand.

Shnoulle:
@olleharstedt : because see https://github.com/LimeSurvey/LimeSurvey/pull/1104#issuecomment-488280401
> I would like to have a "startFrom' value there. Typically I like having similar tokens with the same length so I would like to start tokens eg from 10000,

Or [Omit ambiguous characters from token](https://bugs.limesurvey.org/view.php?id=15026) (most valuable in my opinion than only number)

I ever have user with `specificcode{number}` only token code. And why not only alpah (not numeric), only uppercase (great to have too) 

This pull request do ONE system when we can need a lot more … Adding a event + a plugin for numeric only care all of options …

Shnoulle:
@tammoterhark is a solution for all too, but : more and more system to care by core dev in LimeSurvey … need a test here for combination.

olleharstedt:
We can do both. Add new core features _and_ event to allow more customization.

olleharstedt:
Also note that Travis fails because the newly installed database column is different from the upgraded one:

> (Upgraded from db version 258) Comparing field name "allowNull" for column "tokentype" in table "lime_surveys":  upgraded value: false;  fresh install value: true
Failed asserting that true matches expected false.

Shnoulle:
> We can do both. Add new core features and event to allow more customization.

? Why do both , in my opinion, needed

1. Plugin event and one or more core plugins
2. @tammoterhark solution to do whole in core

Jan-E:
> We can do both. Add new core features _and_ event to allow more customization.

How about a change like this for adding a new event `customToken`:
```
diff --git a/application/models/Token.php b/application/models/Token.php
index e4b90ce..3005821 100644
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -237,7 +237,19 @@ abstract class Token extends Dynamic
      */
     public static function generateRandomToken($iTokenLength)
     {
-        return str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
+        /**
+         * We fire the customToken event
+         */
+        $event = new PluginEvent('customToken');
+        // $surveyId = $this->dynamicId; <- fails
+        $surveyId = isset($_SESSION['LEMsid']) ? intval($_SESSION['LEMsid']) : NULL;;
+        $event->set('surveyId', $surveyId);
+        $event->set('iTokenLength', $iTokenLength);
+        $event->set('generatedToken', '');
+        App()->pluginManager->dispatchEvent($event);
+        $token = $event->get('generatedToken');
+        if ($token == '') $token = str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
+        return $token;
     }
 
     /**
```
The idea is that a plugin that subscribes to the event `customToken` can set a \$token in the event.
If the token is not generated by a plugin the existing function will be used.

I tested this with a plugin that used this functiion to generate the token:
```
/**
 * The custom generate function
 */

public function generateCustomToken()
{
    $event = $this->getEvent();
    $iSurveyID=$event->get('surveyId');
    if (!$this->get('enabled', 'Survey', $iSurveyID, false)) {
        //echo "<pre>pkugin not active for iSurveyID {$iSurveyID} ".$this->get('enabled', 'Survey', $iSurveyID, false)."</pre>";
        return;
    }
    $iTokenLength = $event->get('iTokenLength');
    $event->set('generatedToken', randomChars($iTokenLength, '123456789'));
    //echo "<pre>iSurveyID = {$iSurveyID} generatedToken ".$event->get('generatedToken')." ".$this->get('enabled', 'Survey', $iSurveyID, false)."</pre>\n";
}
```

Jan-E:
I now have these functions:
* Numeric tokens  
* Without ambiguous characters  
* CAPITALS ONLY  

See the [functions in the source code](https://github.com/toolsforresearch/TFRcustomToken/blob/master/TFRcustomToken.php#L38) of our [TFRcustomToken plugin](https://github.com/toolsforresearch/TFRcustomToken)

Shnoulle:
@Jan-E :+1: for `$event = new PluginEvent('afterGgenerateCustomToken');` 

With 

````
$token = str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
new PluginEvent('afterGgenerateCustomToken');
$event->set('surveyId', $this->getSurveyId());
$event->set('oToken', $this);
$event->set('token', $token);
App()->pluginManager->dispatchEvent($event);
$token = $event->get('token');
return $token;
````

Then : 
1. sure to have current sid, 
2. can check some value indide current token (Object), 
3. can get the current token and update or totally replace.
4. Without plugin : nothings are updated

Jan-E:
@Shnoulle 
I do not mind reshuffling the order of the commands. And the name of the pluginevent is something for the LimeSurvey devs to decide. The real trouble is in getting the sid. I tried various methods. The first 3 methods are used somewhere in Tokens.php:
* $event->set('surveyId', $this->getSurveyId());
* $event->set('surveyId', $this->dynamicId);
* $event->set('surveyId', $this->getDynamicId());
* $event->set('surveyId', intval($_SESSION['LEMsid']));

However, the first 3 fail, apparently throw an Exception and generate an error 500.
Just try your proposed method and add only this to the `function generateRandomToken`, nothing else:

`$iSurveyId = $this->getSurveyId();`

Only the method using the \$_SESSION succeeds when you try to Generate tokens for a survey. Result of the others: error 500. Just a educated guess: is the error caused by the fact that the `generateRandomToken` function is not called by a scenario like `Token::model(12345)->generateRandomToken`

BTW. You forgot to pass the \$iTokenLength to the plugin:

`$event->set('iTokenLength', $iTokenLength);`

Jan-E:
Let us continue the discussions in the new PR without any code changes yet:
https://github.com/LimeSurvey/LimeSurvey/pull/1307
This way we separate the discussions about adding a new dropdown in core and adding a new event. better not get too off-topic here.

Shnoulle:
Ah right … no surveyId because it's a static function, 

Think it's best to have $this totally when generate token, then 2 solution:

1. Add the event in https://github.com/LimeSurvey/LimeSurvey/blob/e5d117578b8db82df1ba56bf3e26a72077a27759/application/models/Token.php#L223
2. Update https://github.com/LimeSurvey/LimeSurvey/blob/e5d117578b8db82df1ba56bf3e26a72077a27759/application/models/Token.php#L244 to `public static function generateRandomToken($iTokenLength,$iSurveyId=null,$aTokenAttributes=null)`

Looking at usage in develop, not a lot of place for usage
````
$ grep -Ri "generateRandomToken" *
application/models/Token.php:        $this->token = $this::generateRandomToken($iTokenLength);
application/models/Token.php:            $this->token = $this::generateRandomToken($iTokenLength);
application/models/Token.php:    public static function generateRandomToken($iTokenLength)
application/models/Token.php:                $newtoken = $this::generateRandomToken($iTokenLength);
application/controllers/admin/tokens.php:                    $token->token = Token::generateRandomToken($aData['tokenlength']);
````

In my opinion : `$_SESSION` must be totally avoided…

Jan-E:
To all participants in and subscribers to this PR: please review [Add a new event to generate custom tokens](https://github.com/LimeSurvey/LimeSurvey/pull/1307) at https://github.com/LimeSurvey/LimeSurvey/pull/1307

Jan-E:
@Shnoulle approved an already outdated version of https://github.com/LimeSurvey/LimeSurvey/pull/1307
Does anyone here have additional comments?

Shnoulle:
@Jan-E , i merge it in develop (and take responsability of fixing if needed).

Jan-E:
@Shnoulle Do you mean adding a dropdown 'token type' for numeric tokens?

There are several people here that want to see it in core, not in a plugin. I know you are a fan of plugins, but having this in core as well miht be eqally inportant.

Jan-E:
New PR for afterGenerateToken plus customToken plugin is here now: https://github.com/LimeSurvey/LimeSurvey/pull/1308

If this PR is merged every LimeSurvey admin can activate the plugin and choose for the 3 token types in the plugin:
* Numeric tokens  
* Without ambiguous characters  
* CAPITALS ONLY

Please review the PR.

PR #1107, opened 2018-08-03
Fixed Postgres Database Creation

hovanessb:
PostgreSQL does not allow integers to be stored in Boolean type columns. The file was updated to create the database without errors. Specifically two active columns for `surveymenue_entries` and `surveymenu`, `showincollapse` the `active` column are now `integer`. This was tested on PostgreSQL 10 & 11


installer/create-database.php
```diff
@@ -351,7 +351,7 @@ function createDatabase($oDB){
         $oDB->createCommand()->createTable('{{quota_languagesettings}}', array(
             'quotals_id' => "pk",
             'quotals_quota_id' => "integer NOT NULL default '0'",
-            'quotals_language' => "string(45) NOT NULL default 'en'",
+            'quotals_language' => "sstring(45) NOT NULL default 'en'",
```
LouisGac: "ss"tring

Shnoulle:
Maybe it's better to fix the way we set showincollapse ? And set it in [rules](https://github.com/LimeSurvey/LimeSurvey/blob/108cf3522890408673977783c9bce8564fdfe189/application/models/Surveymenu.php#L37) : https://www.yiiframework.com/wiki/56/reference-model-rules-validation#validation-rules-reference

We can event fix it before save and before load …

hovanessb:
Prior to setting `showincollapse`  on saving, the database installation doesn't create this table in a fresh install and dies with this error on Postgres. I think the safest would be to set them as integers. Any other logic within this application that I saw tests this field as true or false by explicitly checking `showincollapse==0`  or `showincollapse=1` so I'm not exactly sure why they were set as booleans in the first place?

installer/create-database.php
hovanessb: Thanks, this was fixed.

hovanessb:
So I would like to be able to fix the other Databases but I can't tell how Travis is testing the other builds? Or what is it comparing?

```bash
1) ls\tests\UpdateDbHelperTest::testCompareUpgradeAndFreshInstall
(Upgraded from db version 258) Comparing field name "dbType" for column "active" in table "lime_surveymenu_entries":  upgraded value: "tinyint(1)";  fresh install value: "int(11)"
Failed asserting that two strings are equal.
```

olleharstedt:
If you install limesurvey 2.73 and then upgrade to your commit, the fields will still have type boolean.

hovanessb:
Ok now everything seems ready

installer/sql-old/create-mysql.sql
```diff
@@ -631,7 +631,7 @@ CREATE TABLE `prefix_surveymenu` (
   `title` varchar(255)  NOT NULL DEFAULT '',
   `position` varchar(255)  NOT NULL DEFAULT 'side',
   `description` text ,
-  `active` int(1) NOT NULL DEFAULT '0',
+  `active` int(11) NOT NULL DEFAULT '0',
```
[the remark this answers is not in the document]
olleharstedt: There's a manual page on how to add a new database version. I'll try to find it tomorrow.

olleharstedt:
Is the same bug fixed by this commit? https://github.com/LimeSurvey/LimeSurvey/commit/0537858d80bc92b12eba643eb785bf87b17ebfaa

olleharstedt:
Follow the instructions here to create a new database version: https://manual.limesurvey.org/Database_versioning

hovanessb:
@olleharstedt Yes it seems so... well that was fun :1st_place_medal:

olleharstedt:
Sorry, bad communication within the team. Thanks, though.

PR #1108, opened 2018-08-08
Add a new setting "forgotpasswordurl".

pefeigl:
If it is set, use it on the login page, to link to an external page. We need this because all our users are LDAP users, where LS cannot change the password directly.

New setting:
settings

Link on login page:
login

### DIFF
--- a/application/controllers/admin/globalsettings.php
+++ b/application/controllers/admin/globalsettings.php
@@ -317,6 +317,7 @@ class GlobalSettings extends Survey_Common_Action
         setGlobalSetting('RPCInterface', Yii::app()->getRequest()->getPost('RPCInterface'));
         setGlobalSetting('rpc_publish_api', (bool) Yii::app()->getRequest()->getPost('rpc_publish_api'));
         setGlobalSetting('characterset', Yii::app()->getRequest()->getPost('characterset'));
+        setGlobalSetting('forgotpasswordurl', Yii::app()->getRequest()->getPost('forgotpasswordurl'));
         setGlobalSetting('sideMenuBehaviour', Yii::app()->getRequest()->getPost('sideMenuBehaviour', 'adaptive'));
         $savetime = intval((float) Yii::app()->getRequest()->getPost('timeadjust') * 60).' minutes'; //makes sure it is a number, at least 0
         if ((substr($savetime, 0, 1) != '-') && (substr($savetime, 0, 1) != '+')) {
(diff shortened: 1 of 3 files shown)


Shnoulle:
This must be in the plugin : not in core. 

Update event if needed, but please not in core … core do core and offer event for plugin. If plugin need something : see if need event update …

pefeigl:
> @pefeigl How do your users authenticate? By using the LDAP authentication plugin?

Yes, exclusively. Thus right now there is no way to send them to the
correct page for resetting their password, thence the pull request ;)
If you are willing to merge it, I'll rework it to work against current
master.

Greetings, Peter

c-schmitz:
As @Shnoulle already pointed this settings needs to be a part of the LDAP authentication plugin, not the core.
Also any new features need to go to the develop branch.

Shnoulle:
I didn't say it can be done currently with current event to update (in PHP) to show a different url BUT

1. You can add a css file or a js to hide/update it (just register a css file or a script)
2. You can easily https://manual.limesurvey.org/BeforeControllerAction to do a redirect if controller is admin, action is authentification and request->getParam('sa') is forogtpassword

About a plugin : see how add an event here : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/authentication.php#L221

ecraven:
OK. I definitely disagree with your view here, this is something that belongs with the LDAP code, as LDAP authentication is rather useless without it. I am willing to port the code to work with current master, but I will not implement it as a separate plugin.
If that is unacceptable, I'll just close this, and keep it as a separate private branch. No hard feelings, but I think this is core functionality that is missing in core...

Shnoulle:
I say : you can add an event or update current event to add 'forgotlink' … i don't say core don't have to do nothing …

dominikvitt:
@pefeigl: Please port this PR to the current develop branch.
It should be developer as core feature.
We hope to have this great feature merged into develop before final LS4 release.

pitbulk:
I made that using the js suggestion.

In my plugin I defined a newLoginForm method that I subscribed and has the following content:
```
$alternativePw = "http://desiredpw.com"; // Value I read from the setting plugin
$authtype_base = 'Authdb';
$container = $this->getEvent()->getContent($authtype_base);
$container->addContent('
            <script type="text/javascript">
                $( document ).ready(function() {
                    $(\'div.login-submit\').find(\'a\').first().attr(\'href\', "'.$alternativePw.'");
                });

            </script>');
```
It would be great to add id or class to the <a> element to make easiler its recognition.

olleharstedt:
New feature? Need a Mantis ticket!

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/pull/1422 , approch is better (IMO)

And IMO : https://github.com/LimeSurvey/LimeSurvey/pull/1422#issuecomment-631987602 seems best solution too (IMO)

cdorin93:
So, shall we stick to #1422 ?

olleharstedt:
Use #1422 instead.

PR #1113, opened 2018-08-22
Fixed issue #13950: SQL Error when saving a response or getting a session token via API

githubLewis:
Fixed issue #13950: SQL Error when saving a response or getting a session token via API

New feature # :
Changed feature # :
Dev: Without this fix Responses aren't saved & API authentication fails - only applies to SQL Server.
Dev: This commit has also pulled in some of my defensive coding changes - these changes are required for me as they throw errors in my development environment.


installer/create-database.php
```diff
@@ -396,7 +396,7 @@ function createDatabase($oDB){
         $oDB->createCommand()->createTable('{{sessions}}', array(
             'id' => "string(32) NOT NULL",
             'expire' => "integer NULL",
-            'data' => "binary",
+            'data' => "string(255) NULL",
```
c-schmitz: That can't be right. Session data is huge and cannot fit into such short field type.

Shnoulle:
@githubLewis : can you try https://github.com/Shnoulle/LimeSurvey-Shnoulle/commit/b7868f2964488fd792601bbb4bf00aca0a130a7e .

Totally WIP and currently untested ;)

installer/create-database.php
Shnoulle: Since it totally broke on MSSQL : why not just an text (or ntext) ?

dominikvitt:
Can I close this PR?
It looks patch is already applied:
https://github.com/LimeSurvey/LimeSurvey/commit/1a28a4639865647c8756d9b07c740be4e9df35f1

Shnoulle:
Yes, https://github.com/LimeSurvey/LimeSurvey/commit/1a28a4639865647c8756d9b07c740be4e9df35f1 fix THIS issue.

PR #1115, opened 2018-09-07
Best practise to talk about LibreOffice rather than OpenOffice

jinformatique:
Changed # : replaced OpenOffice occurence by LibreOffice

### DIFF
--- a/application/views/admin/participants/importCSV_view.php
+++ b/application/views/admin/participants/importCSV_view.php
@@ -101,7 +101,7 @@ echo viewHelper::getViewTestTag('importParticipants');
                         <div class='panel-body'>
 
                             <p>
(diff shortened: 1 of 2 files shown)


Shnoulle:
I think i prefer something like 

> major spreadsheet tool

🙂

jinformatique:
Sure, but people who only heard of OpenOffice will believe OpenOffice is the major spreadsheet and  won't know about LibreOffice until someone tell them. This could be a written way to let them know about LibreOffice.

c-schmitz:
I would suggest 'default for most Spreadsheet tools', that way we don't give a preference to anyone.

Shnoulle:
@jinformatique : please, review your pull request using 
> default for most Spreadsheet tools

:)

c-schmitz:
Since OP did not took care of it I fixed it myself.

PR #1118, opened 2018-09-12
issue #14050: Unable to import response from deactivated survey

Shnoulle:
Fixed issue #14050: Unable to import response from deactivated survey after update datestamp
Dev: set a default value to the 2 value for datestamp
Dev: didn't see why it work in 2.6 and not in 2.73 …

If someone found a better sentence (in less word …). Must test in msql and pgsql (but NOW() is a common sql word. 

If didn't use NOW : can not set to "0000-00-00 …" since i think it's only for mysql


Shnoulle:
Fixed in https://github.com/LimeSurvey/LimeSurvey/commit/d21f31c02ad945ff362ae79541ad6b45ebfab69c

In my opinion : default value can be set by rules when 'import'

PR #1119, opened 2018-09-14
Fixed issue #14061: Duplication check when adding participant to survey not working

singerla:
Fixed issue #14056: duplication check when adding participant to survey not working
Dev: $oTokens cant be accessed within callback function for 'array_reduce'. The $oToken's participant_id has to be compared to the oParticiant's.


Shnoulle:
Thank you

PR #1121, opened 2018-09-14
Easiest solution for translation

Shnoulle:
- Usage of get_class($this) allow to find source more easily
- This allow to use gT function in view without send whole translation or plugin in views or twig
- I think it can allow to update some string with translation more easily in twig public file too : see https://bugs.limesurvey.org/view.php?id=13711 , current solution is to do it with fixed string by language, here : can do it with fixed en_us string + translation.
- **This don't break current plugin translation**

### DIFF
--- a/application/core/LSCGettextMessageSource.php
+++ b/application/core/LSCGettextMessageSource.php
@@ -29,7 +29,7 @@ class LSCGettextMessageSource extends CGettextMessageSource
             $messageFile .= self::PO_FILE_EXT;
         }
 
-        if (false && $this->cachingDuration > 0 && $this->cacheID !== false && ($cache = Yii::app()->getComponent($this->cacheID)) !== null) {
+        if ($this->cachingDuration > 0 && $this->cacheID !== false && ($cache = Yii::app()->getComponent($this->cacheID)) !== null) {
             $key = self::CACHE_KEY_PREFIX.$messageFile.".".$category;
             if (($data = $cache->get($key)) !== false) {
                 return unserialize($data);
--- a/application/core/LSCGettextMessageSource.php
+++ b/application/core/LSCGettextMessageSource.php
@@ -22,26 +22,25 @@ class LSCGettextMessageSource extends CGettextMessageSource
         // Default catalog to langauge (e.g. de)
         // TODO: Where is catalog set (except default value)?
         $this->catalog = $language;
-
         $messageFile = $this->basePath.DIRECTORY_SEPARATOR.$language.DIRECTORY_SEPARATOR.$this->catalog;
         if ($this->useMoFile) {
-                    $messageFile .= self::MO_FILE_EXT;
+            $messageFile .= self::MO_FILE_EXT;
         } else {
-                    $messageFile .= self::PO_FILE_EXT;
+            $messageFile .= self::PO_FILE_EXT;
         }
 
-        if ($this->cachingDuration > 0 && $this->cacheID !== false && ($cache = Yii::app()->getComponent($this->cacheID)) !== null) {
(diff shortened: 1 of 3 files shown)


olleharstedt:
Do you have a usage example?

application/core/LSCGettextMessageSource.php
```diff
@@ -22,26 +22,25 @@ public function loadMessages($category, $language)
...
         if ($this->useMoFile) {
-                    $messageFile .= self::MO_FILE_EXT;
+            $messageFile .= self::MO_FILE_EXT;
         } else {
-                    $messageFile .= self::PO_FILE_EXT;
+            $messageFile .= self::PO_FILE_EXT;
         }
 
-        if ($this->cachingDuration > 0 && $this->cacheID !== false && ($cache = Yii::app()->getComponent($this->cacheID)) !== null) {
+        if (false && $this->cachingDuration > 0 && $this->cacheID !== false && ($cache = Yii::app()->getComponent($this->cacheID)) !== null) {
```
LouisGac: if false? :stuck_out_tongue_winking_eye:

Shnoulle: Arg …

Shnoulle:
@olleharstedt : no existing plugin, since i can use it like that currently :) 

But, i can use it here : https://gitlab.com/SondagesPro/managament/responseListAndManage/blob/master/views/content/subContent/modalSurvey.php#L13 or here : https://gitlab.com/SondagesPro/managament/responseListAndManage/blob/master/views/content/subContent/modalUser.php#L12

for example.

application/core/LSCGettextMessageSource.php
Shnoulle: thanks

Shnoulle:
@olleharstedt : another think i use in my plugin : specific cacheId for each plugin : i'm unsure it work really best, but : not same file => not same cacheId , no ?

Shnoulle:
Some example of usage : 
1. Create a "textForMyCompany" plugin : only transaltion inside with (for example) : "Legal notice".
2. Add {{ gT('Legal notice') }} in template footer (surely with a link but it's not the point here)
3. Add fr.po with "Note légale" for translation
4. Then french user see "Note légale", other see "Legal notice"

No need complex part except usage of poedit for plugin :)

Shnoulle:
Example of usage : https://github.com/olleharstedt/MassAction/blob/master/views/index.twig#L4 😉

olleharstedt:
>     2. Add {{ gT('Legal notice') }} in template footer (surely with a link but it's not the point here)
> 
>     3. Add fr.po with "Note légale" for translation
> 
>     4. Then french user see "Note légale", other see "Legal notice"

No plugin name in `{{ gT('Legal notice') }}`?

Shnoulle:
You're right, `{{ gT('Legal notice','html,null,'myCompanyString') }}`

Shnoulle:
gT($string,$type,null,$plugin) can be seen as a shortcut (or long cut) to $plugin->gT($string).

olleharstedt:
I suggest:

1. A new method `pT` that reads from registered plugin locale file.
2. A new event RegisterLocaleEvent, which lets plugins register new locale files.

To be discussed in team meeting.

maziminke:
Ohh, sorry! Must have been the wrong button.

olleharstedt:
Other ideas for a more complete translation system:

* Translation is own extension.
* All non-core translation files are compiled to one on first page load, for faster access (see e.g. Twig).

Shnoulle:
OK, let's close this one before have a more complete (and easy please) solution :)

Shnoulle:
@maziminke : you can already use $this->gT in plugin, adding your language file in locale/de/de.po

maziminke:
Nice!

PR #1126, opened 2018-09-21
Dev: trim survey & quota redirect urls from leading & trailing spaces

TonisOrmisson:
Spent a few hours today struggling with a problem initiated by having a leading space in the beginning of an URL. While having the leading space, the URL converted all `&` chars in URL to `&amp;` in database.

it of course looked like an `&` in the form, but ended up breaking the actual parameters since the redirection URLs contained `&amp;`-s instead. Anyway, there is no reason for the URLs to have leading or trailing spaces.


Shnoulle:
:+1:

PR #1127, opened 2018-09-25
New feature: Allow filtering "Out of quota" separately in responses table

TonisOrmisson:
The "No" option will display only the responses whose token is "open" and the questionnaire can be re-opened to fill. The "out of quota" will display responses that are closed due to quotas.

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -1952,7 +1952,7 @@ return $s->hasTokensTable; });
         $dynamic = SurveyDynamic::model($this->sid);
         if ($dynamic->bHaveToken) {
             $criteria = new CDbCriteria;
-            $criteria->addCondition('tokens.completed = \'Q\'');
+            $criteria->addCondition('tokens.completed = ' . Yii::app()->getDb()->quoteValue('Q'));
             $dynamic->with('tokens')->count($criteria);
             return intval($dynamic->with('tokens')->count($criteria)) > 0;
         }
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -87,6 +87,7 @@ use \LimeSurvey\PluginManager\PluginEvent;
  * @property QuestionGroup[] $groups
  * @property Quota[] $quotas
  * @property Question[] $quotableQuestions
+ * @property bool $hasQuota
  *
  * @property array $fullAnswers
  * @property array $partialAnswers
@@ -1936,6 +1937,26 @@ return $s->hasTokensTable; });
         }
 
         return $dataSecurityNoticeLabel;
+
+    }
+
+    /**
+     * @return bool
+     */
+    public function getHasQuota()
+    {
+        if (!empty($this->quotas)) {
+            return true;
+        }
+        // check if any tokens have Quota completes (quota models deleted)
+        $dynamic = SurveyDynamic::model($this->sid);
+        if ($dynamic->bHaveToken) {
+            $criteria = new CDbCriteria;
+            $criteria->addCondition('tokens.completed = \'Q\'');
+            $dynamic->with('tokens')->count($criteria);
+            return intval($dynamic->with('tokens')->count($criteria)) > 0;
+        }
+        return false;
 
     }
 }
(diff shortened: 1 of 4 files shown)


Shnoulle:
I like to have it only if survey as quota :), else : :+1:

TonisOrmisson:
How about you set some quota during the collection and then end up removing them?
A token can be complete="Q" even when survey does not have quotas (any more).
So having this option only when any tokens are actually out of quota (or when survey has quota)
OK?

application/models/Survey.php
```diff
@@ -1938,4 +1939,24 @@ public static function replacePolicyLink($dataSecurityNoticeLabel, $surveyId) {
...
+    public function getHasQuota()
+    {
+        if (!empty($this->quotas)) {
+            return true;
+        }
+        // check if any tokens have Quota completes (quota models deleted)
+        $dynamic = SurveyDynamic::model($this->sid);
+        if ($dynamic->bHaveToken) {
+            $criteria = new CDbCriteria;
+            $criteria->addCondition('tokens.completed = \'Q\'');
```
Shnoulle: `$criteria->addCondition('tokens.completed = '.Yii::app()->getBd()->quoteValue('Q'));` https://www.yiiframework.com/doc/api/1.1/CDbConnection#quoteValue-detail

Best for DB compat

application/models/SurveyDynamic.php
```diff
@@ -518,12 +524,9 @@ public function timeline($sType, $dStart, $dEnd)
...
      * @return bool
      */
-    private function getbHaveToken()
+    public function getbHaveToken()
     {
-        if (!isset($this->bHaveToken)) {
-            $this->bHaveToken = tableExists('tokens_'.self::$sid) && Permission::model()->hasSurveyPermission(self::$sid, 'tokens', 'read'); // Boolean : show (or not) the token;
-        }
-        return $this->bHaveToken;
+        return tableExists('tokens_'.self::$sid) && Permission::model()->hasSurveyPermission(self::$sid, 'tokens', 'read'); // Boolean : show (or not) the token;
```
TonisOrmisson: A general comment. I do not think its a good idea to make permission checks in models. This should be covered in other places (eg controller).

Shnoulle:
Last ;) i think currently completed return Quota + submitted … maybe have 

- quota : `== 'Q'`
- completed `!= 'N' `
- submitted : `!= 'Q' and !='N' ` since it can be Y or date string

TonisOrmisson:
@Shnoulle completed Y/N is not using token entity at all. Its using submitdate.

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/SurveyDynamic.php#L624...L630

Shnoulle:
Yes, i edited after, sorry … i don't know exactly what is set for submitdate when token is set to Q :)

TonisOrmisson:
nothing, it will be what it was before, since it will not be submitted, `null`

application/models/SurveyDynamic.php
```diff
@@ -628,6 +631,9 @@ public function search()
         if ($this->completed_filter == "N") {
             $criteria->addCondition('t.submitdate IS NULL');
         }
+        if ($this->completed_filter == "Q") {
+            $criteria->addCondition('tokens.completed = ' . Yii::app()->getDb()->quoteValue('Q'));
```
dominikvitt: This breaks with error message if token table doesn't exists:
500: Internal Server Error
CDbCommand failed to execute the SQL statement: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'tokens.completed' in 'where clause'

Check if token table exists should be implemented, something like: if ($this->bHaveToken)

dominikvitt:
Currently there are only two response status icons: Yes and No. 
It would be nice to have different response status icon (or icon + text "Out of quota" ) for "Out of quota" status.

dominikvitt:
@TonisOrmisson: Please port this PR to the current develop branch.

maziminke:
@TonisOrmisson Can you share a screenshot of how the new feature loos like?

application/models/SurveyDynamic.php
Shnoulle: General comment : some model function NEED Permission : you think we can add a array var 

```
public var $permission = [
'create' => false,
'read' => false,
'update' => false,
'delete' => false,
'import' => false,
'export' => false,
]
```

At minima for buttons it's needed (in my opinion).

Maybe you can open a Development mantis about this ? If we are OK , i can start on it in develop version.

PR #1129, opened 2018-09-25
Dev: updated feature #14093: Move simple plugin default to Survey set…

Shnoulle:
Just update default menu


lacrioque:
I'm against putting it under Plugin settings. It is not really a "Setting".
Let's talk about this in the open on the meeting later today.

Shnoulle:
It's the default way to add settings by survey for plugin …

And currently : its at the same place but under a submenu … this submenu are not used anywhere …

Shnoulle:
OK, i redo the pull request : plugin in Survey menu at end, like now, but without the submenu.

This can be removed when we have a better solution to add "Survey settings" by plugin easily (my choice now : use Tool menu ;) )

PR #1130, opened 2018-09-25
Fix JSON editor options to load the right content

yent:
Fix JSON editor options to load the right content into the embedded JSON editor, issue was that content, whatever its type, was passed to CHtml::encode which expects a string due to lack of encode htmlOption for CHtml::textarea. Also added needed json_encode so that textarea value is a string in the end.

Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:


c-schmitz:
@LouisGac @dominikvitt Please review and merge if okay.

PR #1134, opened 2018-10-04
Fixed issue #14127: Slider with only right part broken

Shnoulle:
Dev: each row can have different width


Shnoulle:
capture du 2018-10-04 10-43-28

lacrioque:
This seems to be good.
Great fix!

Shnoulle:
Yes, and after thinking : i think its' best to do it like this : if you want "empty" right part : you can just add a [space] inside it : we don't trim

`'|subquestion|right' !== ' |subquestion|right'`

PR #1135, opened 2018-10-04
Fixed issue #14126: google map without API broke

Shnoulle:
Dev: remove gmap if no API
Dev: set to OSM if currently on gmap
Dev: disable location_nodefaultfromip if ipInfoDbAPIKey is empty
Dev: best solution is really to move all of this in question_template …

This is the quickest solution BUT we can not currently create a clean question template : https://bugs.limesurvey.org/view.php?id=14128

### DIFF
--- a/application/helpers/qanda_helper.php
+++ b/application/helpers/qanda_helper.php
@@ -3032,6 +3032,11 @@ function do_shortfreetext($ia)
     $answer = "";
     $sQuestionHelpText = '';
 
+    /* Fix map service according to google map API key */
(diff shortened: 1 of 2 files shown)


Shnoulle:
This commit can use this another commit : https://github.com/LimeSurvey/LimeSurvey/pull/1159

Surely better : hide it, but dodn't delete existing attribute :)

Shnoulle:
And see : https://bugs.limesurvey.org/view.php?id=14190 , have a new attribute : disable or forcedefault  can be used too.

Shnoulle:
develop

PR #1136, opened 2018-10-04
Dev: remove binary for third_party and framework

Shnoulle:
Dev: since we add own fix to third_party and framework : more easy
Dev: see https://www.limesurvey.org/forum/development/116591-binary-third_party

Related commit : https://github.com/LimeSurvey/LimeSurvey/commit/b8ff705b4583882bbea90b6fe5421d603731ba71


Shnoulle:
Each third_party need own .gitattribute file since some can be OK, some can be crlf etc … need more testing time than i have currently

PR #1139, opened 2018-10-06
Fixed issue: {SID} placeholder does not work in email templates

vectorson:
Fixed issue: {SID} placeholder does not work in email templates
Dev: It's already available in the placeholder list in the GUI (see image below) but it just gets replaced with an empty string.

screen shot 2018-10-06 at 23 31 23


Shnoulle:
Not related to this commit, but there are an issue somewhere else.
See https://github.com/LimeSurvey/LimeSurvey/blob/6bb4566a41c46c55d8da8ee595a005d3d5cd5f38/application/helpers/expressions/em_manager_helper.php#L3655

SID must be always set to current survey id …

PR #1140, opened 2018-10-08
Dev: clean up PluginsController to leave only public action

Shnoulle:


application/controllers/PluginsController.php
```diff
@@ -7,143 +7,6 @@
...
-    {
-        return $this->properties[$property];
-    }
-
-    public function __set($property, $value)
-    {
-        $this->properties[$property] = $value;
-    }
-
-    public function _init()
```
Shnoulle: Personnaly : i use direct for own page : but i register BS myself …

application/controllers/PluginsController.php
```diff
@@ -7,143 +7,6 @@
...
-        Yii::app()->bootstrap->init(); // Make sure bootstrap css is rendered in time
-    }
-
-    /**
-     * Activates plugin with $id
-     *
-     * @param int $id
-     * @return void
-     */
-    public function actionActivate($id)
```
Shnoulle: Ore leave it and redirect to admin controller ? I never use it in plugins (and it's not in API)

application/controllers/PluginsController.php
```diff
@@ -7,143 +7,6 @@
...
-    }
-
-    /**
-     * Show configuration for plugin with $id
-     *
-     * @param int $id
-     * @return void
-     */
-    public function actionConfigure($id)
-    {
```
Shnoulle: Or leave it and redirect to controller : but it's not in API (i never use it in plugin)

application/controllers/PluginsController.php
```diff
@@ -7,143 +7,6 @@
...
-    }
-
-    /**
-     * Deactivates plugin with $id
-     *
-     * @param int $id
-     * @return void
-     */
-    public function actionDeactivate($id)
-    {
```
Shnoulle: See other comment about function ;)

application/controllers/PluginsController.php
olleharstedt: I say remove it. It has no place here.

application/controllers/PluginsController.php
olleharstedt: Same.

application/controllers/PluginsController.php
olleharstedt: Same.

olleharstedt:
Can merge this tomorrow.

Shnoulle:
@TonisOrmisson : you can chek it ? @olleharstedt : if you put an OK : i can merge it :)

PR #1149, opened 2018-10-19
New feature #14156: Need a way to easily disable table to block for p…

Shnoulle:
…articular questions
Dev: just add forced-table in Class name : done for user
Dev: really don't found a cleaner solution …

Someone have a better and cleaner solution. Still easy for user ?


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1152

PR #1151, opened 2018-10-21
Fixed issue: Unable to batch-edit email address and reminder count of survey participants.

vectorson:
Fixed issue: Unable to batch-edit email address and reminder count of survey participants.
Dev: Faulty logic during validation resulted in email address always being set to "lskeep". 
Dev: Changes to reminder count was being ignored completely.


vectorson:
The email validation will unfortunately fail silently if an invalid email is submitted as I couldn't find an easy way to print errors inside the modal popup. There seems to be a bug with printing errors in modal popups, for example the flash message:

`                            Yii::app()->setFlashMessage(sprintf(gT('%s cannot be left empty'), $desc['description']), 'error');
                            $this->getController()->refresh();`

that is currently being executed after validating mandatory attributes is not being displayed in neither single-edit nor batch-edit.

maziminke:
A very useful feature/add-on!

vectorson:
Just made an additional commit to allow setting an empty email address.

vectorson:
The issues have been solved in other commits but setting an empty email address is still not possible so this pull request might still be somewhat relevant.

PR #1153, opened 2018-10-22
Update Question.php

st4nny:
fix error query in group filter in listquestions view:
error from db pgsql: "operator does not exist: integer ~~ unknown"

Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -1002,7 +1002,7 @@ class Question extends LSActiveRecord
         $criteria2->compare($qid_reference, $this->title, true, 'OR');
 
         if ($this->gid != '') {
-            $criteria->compare('groups.gid', $this->gid, true, 'AND');
+            $criteria->compare('groups.gid', $this->gid, false, 'AND');
         }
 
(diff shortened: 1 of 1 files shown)


application/models/Question.php
```diff
@@ -1002,7 +1002,7 @@ public function search()
         $criteria2->compare($qid_reference, $this->title, true, 'OR');
 
         if ($this->gid != '') {
-            $criteria->compare('groups.gid', $this->gid, true, 'AND');
+            $criteria->compare('groups.gid', $this->gid, false, 'AND');
```
Shnoulle: See https://www.yiiframework.com/doc/api/1.1/CDbCriteria#compare-detail this change the partial match then add 1 and 10 and 11 and … if `$this->gid` is 1, `false` fixed it

Question : why it's an issue in pgsql ?

dominikvitt:
I can see this issue is already fixed, so I would close this PR.
https://github.com/LimeSurvey/LimeSurvey/commit/64abf721296a93f9e48003e0b9bdad31b9de5900

PR #1155, opened 2018-10-23
Develop merge master 181023

Shnoulle:


application/helpers/expressions/em_manager_helper.php
```diff
@@ -3944,6 +3944,13 @@ public function setVariableAndTokenMappingsForExpressionManager($surveyid,$force
                             $rowdivid = $sgqa;
                         }
 
+                        break;
+                    case Question::QT_H_ARRAY_FLEXIBLE_COLUMN:
```
Shnoulle: I'm unsure here, since it don't exist in develop, but exist in master …

PR #1158, opened 2018-10-24
[READY] Upload image in survey for Template option

Shnoulle:
1. List image in survey (if sid is not empty)
2. Upload image in survey


Shnoulle:
Alt solution : list image in dropdown separatly : Maybe 3 categories : one global + one template + one survey.

But then : need 2 upload in Template options : one global + one template

And … 3 upload in Survey template options : one global + one template + one survey (and need check Permission before showing the upload form …)

Shnoulle:
PS : with this solution : logo.png from template was replaced by logo.png from survey (if file exist and user validate Template Option)

lacrioque:
At first glance I'd say good work, but we will have to check the impact on LimeSurvey Pro seperately.

Shnoulle:
I separate image in survey, maybe adding a `sprintf("%s (in survey)",$filename)` but need to add it in options.twig …

Shnoulle:
For LimeSurvice : there are no change if user didn't update anything in Template option.
Change is done ONLY if user select a image in Survey Template option.

When user delete Survey : image uploaded here is deleted (like all survey files), but he can only select image in template (selected) or Survey

Shnoulle:
See related (current) issue : https://bugs.limesurvey.org/view.php?id=14180

We must choose before :)

Shnoulle:
capture du 2018-10-26 00-29-43

Shnoulle:
@lacrioque can you check LimeSurvey Pro/SAAS impact with last commit ?

I think there are not : i check : you can upload a file `logo.png` in fruity for example : saved in global file. You can select it in dropdown after.

It's not the case today. 

It don't break other template with own options.twig (but if you find a better solution for https://github.com/LimeSurvey/LimeSurvey/pull/1158/commits/c2933177698e737ae01facf9eacf99c639a68fac#diff-deb99980251a9aef800d3fccb44b9be6R210 …)

Shnoulle:
@lacrioque : any news ?

Shnoulle:
Fix https://bugs.limesurvey.org/view.php?id=13827 too

PR #1159, opened 2018-10-27
[WIP] New feature #14185: Need a clean way to hide custom question theme attributes

Shnoulle:
Dev: still a little issue, explain in pull request


Shnoulle:
There still an issue, but don't find where : using `<visible>0</visible>` from a core settings reset it to empty …

Shnoulle:
Arg … `QuestionTemplate` system only use current question template : we must allow forcing another template …

Shnoulle:
OK, tested whole GUI : really adding visible + really allow question template to update core attribute

Shnoulle:
No good solution since we can set $sTemplate to not current one in QuestionTemplate::model …

Shnoulle:
Already done in develop

PR #1162, opened 2018-10-30
Add event to allow survey search tweaking

yent:
Add 'beforeSurveySearch' event dispatch after standard criteria/count/other search options have been build.

The handler gets the data provider and can modify/swap it. This allows further plugin-based search refinements like allowing a certain subset of users (that cannot be superadmin) to see all surveys or having virtualhost owned surveys.

One may also want to catch the 'beforeHasPermission' event to restrict access using the same rules ...

Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -1615,7 +1615,12 @@ class Survey extends LSActiveRecord
                 'pageSize'=>$pageSize,
             ),
         ));
-
+        
+        $oEvent = new \LimeSurvey\PluginManager\PluginEvent('beforeSurveySearch');
+        $oEvent->set('oCActiveDataProvider', $dataProvider);
+        App()->getPluginManager()->dispatchEvent($oEvent);
+        $dataProvider = $oEvent->get('oCActiveDataProvider');
+        
         $dataProvider->setTotalItemCount($this->count($criteria));
 
         return $dataProvider;


Shnoulle:
Why only for Survey ? Maybe better in https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/behaviors/PluginEventBehavior.php

yent:
Absolutely right, I may need a bit more time to understand the principle of it tough :)
As far as I understand I have to 1) add a 'beforeSearch' method which takes the dataprovider as an argument to the PluginEventBehavior class (with related entry in events method) and 2) call it from Survey::search instead of the original code I wrote, is that it ?

Shnoulle:
Maybe more difficult than i think … since don't seems Yii have already a search event … system like [beforeSave](https://www.yiiframework.com/doc/api/1.1/CActiveRecord#onBeforeSave-detail) for example

Shnoulle:
PS : i'm not the one who merge this :) (in master).

yent:
Well, it seems that in the end [CActiveDataProvider calls findAll upon model](https://github.com/LimeSurvey/LimeSurvey/blob/master/framework/web/CActiveDataProvider.php#L199) so using [onBeforeFind](https://www.yiiframework.com/doc/api/1.1/CActiveRecord#onBeforeFind-detail) should work ... I'm going to try that :)

Shnoulle:
unsure it work with findAll, but this allow to set CDbCriteria too :)

yent:
Well ... Did some testing, can't get it to work :( Search criteria seems to be lost along the way ... From the call to [findAll](https://github.com/LimeSurvey/LimeSurvey/blob/master/framework/web/CActiveDataProvider.php#L199) onwards the criteria is passed as an argument, before any call to [beforeFind](https://github.com/yiisoft/yii/blob/master/framework/db/ar/CActiveRecord.php#L1351) happens so I don't see a way to access it ...

Shnoulle:
I think beforeFind work only with find … not with findAll , it's a common issue with Yii event

cdorin93:
Hmm, what is the status of this PR? Does anyone know more :) ?

c-schmitz:
I don't see that this makes any sense as the concept does not be too well thought through  and does not work - so I am closing this.

PR #1164, opened 2018-11-07
Develop pattern numeric

Shnoulle:
New feature 14224: Add pattern for some question type (numeric)

- need radix too (or remove radix for develop ?)
- need some attribute to have the best pattern we can ?


Shnoulle:
@dominikvitt why closing this ?

Shnoulle:
For information : it's already in skelvanilla : https://gitlab.com/SondagesPro/SurveyThemes/skelvanilla/tree/master/views/survey/questions/answer

Shnoulle:
Sh… there are issue with index and move previous …
And surely too  with something like this
- Numeric question Q1 type with condition `is_empty(Q0.NAOK)` in same page
- Enter `aaa` in Q1
- Enter `aaa` in Q0
- Q1 is hidden but validation br'oken 

Maybe this need to be done in js only … (and with js only : better to use `input:number`)

Shnoulle:
No : validation is an issue, then must be able to disable validation in js, must do it in JS only …

PR #1165, opened 2018-11-07
Fixed issue #14227: Some token with existing reponse can send PHP War…

Shnoulle:
…ning about seed
Dev: set a seed if empty

### DIFF
--- a/application/libraries/MersenneTwister.php
+++ b/application/libraries/MersenneTwister.php
@@ -15,6 +15,12 @@ function setSeed($surveyid)
     if (isset($_SESSION['survey_'.$surveyid]['srid'])) {
         $oResponse = \Response::model($surveyid)->findByPk($_SESSION['survey_'.$surveyid]['srid']);
         $seed = $oResponse->seed;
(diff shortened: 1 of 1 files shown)


Shnoulle:
We can fix invalid seed too (badly VV import for example) or throw an error (or just log it …)

olleharstedt:
Looks good?

Shnoulle:
It work, the only potential issue : here https://github.com/LimeSurvey/LimeSurvey/blob/ad436aa43e8e5833825bde8ad39980738e0de249/application/libraries/MersenneTwister.php#L17 (current ls too) : if column don't exist.

I'm unsure we need https://github.com/LimeSurvey/LimeSurvey/blob/ad436aa43e8e5833825bde8ad39980738e0de249/application/libraries/MersenneTwister.php#L31

Shnoulle:
PS : why ? : i assign to you, because you have already some bug on seed :). And you know best than me .

The question : what to do if seed is not set, and what to do if seed is invalid ( ' 1234 ` (see space : unsure it work) or `abcd`) ?

PR #1166, opened 2018-11-08
[READY] Confirm action in grid with POST and urlParam get

Shnoulle:
Fixed issue #14222: When deleting a single response : all page is reloaded and current filter lost
Fixed issue #14221: [security] No XSS control when delete a token
Dev: usage of CButtonColumn
Dev: return real error, show it in alert
Dev: LS need a system to show result after ajax …
Dev: i don't find it


Shnoulle:
Still token to fix

PR #1167, opened 2018-11-08
Fix #14229

amicci:
Fixed issue # : 14229
Dev: Popover is triggered only in visible elements that have data-limesurvey-submit.


Shnoulle:
Not about this fix : but i think this script must be moved to template … it's really a template related part …

Here: really hard to show when template dev want …

PR #1170, opened 2018-11-14
Fixed issue # 14253: Copying survey does not copy survey group

TonisOrmisson:
Fixed issue # 14253: Copying survey does not copy survey group

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -799,6 +799,7 @@ function importSurveyFile($sFullFilePath, $bTranslateLinksFields, $sNewSurveyNam
 */
 function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = null, $iDesiredSurveyId = null, $bTranslateInsertansTags = true, $bConvertInvalidQuestionCodes = true)
 {
+    $isCopying = ($sNewSurveyName != null);
     Yii::app()->loadHelper('database');
     $results = [];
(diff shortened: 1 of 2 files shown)


TonisOrmisson:
https://bugs.limesurvey.org/view.php?id=14253

Shnoulle:
Question : even when export/import : why this need to be resetted ? Maybe a checkbox set to yes by default «reset Survey group» ?

And why 1 ? Default survey group is not null ?

(i never use Survey group currently ;) )

TonisOrmisson:
I understand the need for resetting while importing. 

Currently the group is stores as id in the xml. So if you export-import between different instances (which is usually the case for import - else copy) then the id-s of groups will most likely not match the imported. 

Using a group name instead of id in the importing process would make more sense here. This way we could match the group by name and use the correct match (if found) or default (if not). Having an ID referring to an external component is pretty much meaningless.

TonisOrmisson:
Why 1?

1 is the default value for gsid
https://github.com/LimeSurvey/LimeSurvey/blob/master/installer/create-database.php#L515

Shnoulle:
> Using a group name instead of id in the importing process would make more sense here. This way we could match the group by name and use the correct match (if found) or default (if not). Having an ID referring to an external component is pretty much meaningless.

Yep, it's the reason why we export name of plugin :).
Thanks for the information

c-schmitz:
Thank you!

PR #1171, opened 2018-11-15
Fixed issue #14256: .gitignore does not properly ignore /tmp directory

mfavetti:
https://bugs.limesurvey.org/view.php?id=14256

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,16 +46,16 @@
 
 # tmp directory : whole except the index.html for directory create
 /tmp/*
-!/tmp/*/*index.html
-/tmp/assets/**/
-/tmp/runtime/gii-1.1.18/*.*
-/tmp/runtime/cache/*.bin
-/tmp/runtime/plugin.log
-/tmp/runtime/twig_cache/*
-/tmp/runtime/CSS/*.ser
-/tmp/runtime/HTML/*.ser
-/tmp/runtime/URI/*.ser
-
+/tmp/assets/*
+/tmp/runtime/*
+/tmp/upload/*
+!/tmp/index.html
+!/tmp/assets/
+!/tmp/assets/index.html
+!/tmp/runtime/
+!/tmp/runtime/index.html
+!/tmp/upload/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -46,16 +46,16 @@
...
-/tmp/runtime/twig_cache/*
-/tmp/runtime/CSS/*.ser
-/tmp/runtime/HTML/*.ser
-/tmp/runtime/URI/*.ser
-
+/tmp/assets/*
+/tmp/runtime/*
+/tmp/upload/*
+!/tmp/index.html
+!/tmp/assets/
```
Shnoulle: I don't think it's OK here ? Just the line after, no ?

caseylucas: This has worked well for us:
```
!/tmp/
/tmp/*
!/tmp/index.html
!/tmp/assets/
/tmp/assets/*
!/tmp/assets/index.html
!/tmp/runtime/
/tmp/runtime/*
!/tmp/runtime/index.html
!/tmp/upload/
/tmp/upload/*
!/tmp/upload/index.html
```
See https://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder#5534865

Shnoulle: But : git ignore empty directory (directory didn't exist), then i think

```
/tmp/runtime/*
!/tmp/runtime/index.html
```
only must work , no ?

caseylucas: Yeah, I think you are correct.  I verified that I got the same (desired) results with: 
```
/tmp/*
!/tmp/index.html
/tmp/assets/*
!/tmp/assets/index.html
/tmp/runtime/*
!/tmp/runtime/index.html
/tmp/upload/*
!/tmp/upload/index.html
```

Shnoulle: Yes, but stackoverflow link seems clear here 

> If you exclude application/, then everything under it will always be excluded (even if some later negative exclusion pattern (“unignore”) might match something under application/).

This point are unsure and maybe can bve related to git version, then : leave it :)

PR #1172, opened 2018-11-15
Fixed issue #14217: Unpleasant rebound effect when editing questions

Shnoulle:
Dev: still one rebound, but not a second one 500ms after …


Shnoulle:
But more final time out before page is ready to click on an option after open its 500ms+500ms …

PR #1175, opened 2018-11-20
New feature: Access participant information from Twig

TonisOrmisson:
New feature: access oToken from twig

eg want to display a name or other token-identifiable content in questions
or get the token id to create some third-party content elements related to token

### DIFF
--- a/application/controllers/survey/index.php
+++ b/application/controllers/survey/index.php
@@ -118,14 +118,15 @@ class index extends CAction
             $token = $clienttoken;
         }
 
-        if (empty($token) && isset($_SESSION['survey_'.$surveyid]['token'])) {
-            $token = $_SESSION['survey_'.$surveyid]['token'];
-        }
-
         //SEE IF SURVEY USES TOKENS
         if ($oSurvey->hasTokensTable) {
             $tokensexist = 1;
         }
+
+        if (empty($token) && isset($_SESSION['survey_'.$surveyid]['token']) && !$oSurvey->isAnonymized && $tokensexist === 1) {
+            $token = $_SESSION['survey_'.$surveyid]['token'];
+        }
+
 
         if ($tokensexist == 1 && isset($token) && $token != "" && tableExists("{{tokens_".$surveyid."}}") && !$previewmode) {
             // check also if it is allowed to change survey after completion
--- a/application/controllers/survey/index.php
+++ b/application/controllers/survey/index.php
@@ -118,14 +118,16 @@ class index extends CAction
             $token = $clienttoken;
         }
 
+        if (empty($token) && isset($_SESSION['survey_'.$surveyid]['token'])) {
+            $token = $_SESSION['survey_'.$surveyid]['token'];
+        }
+
         //SEE IF SURVEY USES TOKENS
         if ($oSurvey->hasTokensTable) {
             $tokensexist = 1;
         }
 
-
         if ($tokensexist == 1 && isset($token) && $token != "" && tableExists("{{tokens_".$surveyid."}}") && !$previewmode) {
-
             // check also if it is allowed to change survey after completion
             if ($thissurvey['alloweditaftercompletion'] == 'Y') {
                 $oToken = $tokenInstance = Token::model($surveyid)->editable()->findByAttributes(array('token' => $token));
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -92,6 +92,7 @@ class SurveyRuntimeHelper
 
         // Start rendering
         $this->makeLanguageChanger(); //  language changer can be used on any entry screen, so it must be set first
+        /** @var Token $oToken */
         extract($args);
 
         ///////////////////////////////////////////////////////////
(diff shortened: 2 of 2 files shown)


Shnoulle:
Related ? https://bugs.limesurvey.org/view.php?id=14228

Question : sending object ? Allow twig to update and save ?

TonisOrmisson:
@Shnoulle not related, I was not aware of that.
I just happened to be in need of accessing the token in theme. But generally I agree with the #14228
No not modify. Just access the object. Object is just structured & documented thing to work with (as opposed to an ambiguous aSurveyData)

Shnoulle:
About not to modifiy : i'm totally unsure if twig can use Object function or not ?
Then maybe : 

`'aSurvey'=> Survey::model()->findByPk($this->iSurveyid)->getAttributes(),`
and
`$params['aToken'] = $oToken->getAttributes(),`

For oSurvey : can easily test : oSurvey->getAllLanguages(); …

TonisOrmisson:
@Shnoulle Objects seem to work fine (which is great actually)

with this change I can also display like `{{ oToken.survey.defaultlanguage.surveyls_title }}`

Shnoulle:
OK, looking at twig documentation : [attribute](https://twig.symfony.com/doc/2.x/functions/attribute.html) can be used. We don't add it in default [twig config](https://github.com/LimeSurvey/LimeSurvey/blob/4aca4e4a938d38def389038f4276d2fe2ebeb5ef/application/config/internal.php#L233), then i think it's OK to send whole object :).

We don't have to give access to something than can update DB in another way we need (except if user want it and is allowed to edit `config.php` file). Twig is done to allow upload extension without giving all access to extension developer :)

application/controllers/survey/index.php
```diff
@@ -118,14 +118,16 @@ public function action()
             $token = $clienttoken;
         }
 
+        if (empty($token) && isset($_SESSION['survey_'.$surveyid]['token'])) {
+            $token = $_SESSION['survey_'.$surveyid]['token'];
+        }
+
```
c-schmitz: What about anonymized surveys? In that case the Token information should not be available.
Also the patch should go to develop, instead.

Shnoulle: `{{ processString('{strip_tags(TOKEN:LASTNAME)}') }}` take care of anonymous survey. Maybe oToken are not the best candidate : 1. Can be EM accessible, 2. anonymous are issue

TonisOrmisson: anynomized surveys wont have the `$_SESSION['survey_'.$surveyid]['token']` set, or? Should not be applicable to them at all. 

Regarding the develop: im sorry, I have lost the understanding what is allowed for master and what not. I have seen a bunch of features added to master since the new-features-for-develop-only policy landed. I was not sure if it was relevant any more.

Shnoulle: `$_SESSION['survey_'.$surveyid]['token']` is always set to control if user is allowed to access to this survey :).

It's not related to anonymous here :)

c-schmitz: Even if recently someone commited a new feature it was reverted. As said, please submit for develop instead. Please make sure the token information is not available for anonymized surveys.
Thank you!

TonisOrmisson: not sure about the reverting. I have just seen regularly new features in new releases while develop branch has not been merged into master
the tail of the log:

> Changes from 3.15.3 (build 181108) to 3.15.4 (build 181109)   November 09, 2018
-New feature : list image in Survey for Brand logo file in Theme option (Denis Chenu)
>
> Changes from 3.15.1 (build 181017) to 3.15.2 (build 181107)  November 07, 2018
>
> -New feature: allow to change config file location (LouisGac)
-New feature: allow to change config file location, part 2 (LouisGac)
-New feature: configuration option to use a single upload directory for all surveys (LouisGac)
>
> Changes from 3.14.11 (build 180926) to 3.15.0 (build 181008) October 8, 2018
-New feature #14119: minute step for datetimepicker (Denis Chenu)
-New feature: Added Italian language to multilingual sample survey (Luciano Fanfoni) (Carsten Schmitz)
>
> Changes from 3.14.9 (build 180917) to 3.14.10 (build 180924) September 24, 2018
-New Feature: added Twig function getAllAnswers() (LouisGac)

Its just a bit confusing. Some new features clearly have gone straight to master and I was assuming it was ok. Can anybody explain why eg 
> Changes from 3.14.9 (build 180917) to 3.14.10 (build 180924) September 24, 2018
-New Feature: added Twig function getAllAnswers() (LouisGac)

... went straight into master & release, which in essence seems to be pretty similar to this commit?

Shnoulle: About 

>  -New feature : list image in Survey for Brand logo file in Theme option (Denis Chenu)

It's more because fixing related issue (user without template update rights can not upload logo in Survey theme option) . Fixing issue need this new feature :)

c-schmitz: We should not name it oToken - oParticipant would be better, since the Token is just an information bit of the participant.

TonisOrmisson: I did not name anything there. I'm just passing the existing things. And if I look at what It is, Its called Token_{SID} representing the class Token or TokenDynamic. 

But the class name is confusing, I agree.

LouisGac:
great PR, but why in master? Could it be pushed to develop please?

maziminke:
Was this feature actually added?
If so, do we have some additional information available at the manual on how to use it and the available options/details?

PR #1176, opened 2018-11-22
[WIP] Fixed issue #14270: value, valueNAOK etc … broken with GET params

Shnoulle:
- move prefillFromCommandLine from buildsurveysession to initFirstStep
- must test prefilling in some condition
- must test when setVariableAndTokenMappingsForExpressionManager without surveyOptions


Shnoulle:
Bad solution until develop or other rewriting

PR #1177, opened 2018-11-22
Fixed issue #14271: Some words used by LS can be used as question codes

Shnoulle:
- disable Bad word on same way than unicity and pattern : don't broke existing survey with this code
- Subquestion already fixed : time and other
- Don't see for answer ?


Shnoulle:
Report another issue with Multiple text with comment, else all seems OK

PR #1178, opened 2018-11-22

Shnoulle:
- this ONLY a quick fix …
- To work seems $this->RDP_errs must be array of errors (string and token)
- but currently receive array of array of errors … sometimes
- with debug = 0 , null == 0, and $globalErrs[] = something not set
- mimic debug=0 behaviour with debug=2

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -1297,9 +1297,17 @@ class ExpressionManager
         $globalErrs = array();
         $bHaveError = false;
         while ($errIndex < $errCount) {
-            if ($errs[$errIndex++][1][1] == 0) {
+            /**
+             * Shnoulle on 2018-11-22
+             * This is really a quick fix … i really don't understand the logic here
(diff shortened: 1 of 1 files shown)


Shnoulle:
@dominikvitt it's really a quick fix , since you have https://bugs.limesurvey.org/view.php?id=13630 , please find a better solution …

Seems OK when
````
    0 => array
    (
        0 => 'Variable non définie'
        1 => array
        (
            0 => 'ASSESSMENT_CURRENT_TOTAL'
            1 => 0
            2 => 'WORD'
        )
    )
````

But not when 
````
    0 => array
    (
        0 => array
        (
            0 => 'Variable non définie'
            1 => array
            (
                0 => 'ASSESSMENT_CURRENT_TOTAL'
                1 => 0
                2 => 'WORD'
            )
        )
    )
````

I think it's an bad array merging/push somewhere ?

Shnoulle:
Maybe this one is better : https://github.com/LimeSurvey/LimeSurvey/pull/1179

PR #1179, opened 2018-11-22

Shnoulle:
- This seems to be a better fix
- doing $errors = array_merge($errors, $this->RDP_errs); $this->RDP_errs = $errors; seems … really strange
- But maybe there are some good reason …


Shnoulle:
All this function need rewrite : show only the 1st error, not all. Surround 1st error, not all etc …

PR #1180, opened 2018-11-23

Shnoulle:
Fixed issue : Only partial error shown in logic file
Dev: Previous system is totally broken , rewrite it
Dev: adding $this->RDP_errs inside $this->RDP_errs : remove this …


Shnoulle:
Before the fix
withoutfix-current

With the fix
withfix

PR #1185, opened 2018-11-27
Fixed issue #9762: Token column is blank when selecting "Not complete…

mfavetti:
Fixed issue # : [14289](https://bugs.limesurvey.org/view.php?id=14289)

### DIFF
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -1964,12 +1964,14 @@ function tokensExport($iSurveyID)
         if ($bIsNotAnonymous) {
             $oRecordSet->leftJoin("{{survey_$iSurveyID}} ls", 'lt.token=ls.token');
             $oRecordSet->andWhere("ls.id IS NULL");
+            $oRecordSet->select("lt.*, ls.id");
         }
     }
     if ($iTokenStatus == 3 && $bIsNotAnonymous) {
         $oRecordSet->leftJoin("{{survey_$iSurveyID}} ls", 'lt.token=ls.token');
         $oRecordSet->andWhere("lt.completed='N'");
         $oRecordSet->andWhere("ls.id IS NULL");    
+        $oRecordSet->select("lt.*, ls.id");
     }
     if ($iTokenStatus == 4 && $bIsNotAnonymous) {
         $oRecordSet->selectDistinct('lt.tid, lt.firstname, lt.lastname, lt.email, lt.emailstatus, lt.token, lt.language, lt.sent, lt.remindersent, lt.remindercount, lt.completed, lt.usesleft, lt.validfrom, lt.validuntil, MAX(ls.startdate) as started');


Shnoulle:
I see : https://github.com/LimeSurvey/LimeSurvey/blob/bcb56b5fd4155d51ea38daf0b47303dbd9a31794/application/helpers/export_helper.php#L1977

for `$iTokenStatus == 4` , we don't need distinct or we need it ?

mfavetti:
OK, I looked more in depth at the current options and criteria. Here is what we have now:
0) All
1) Completed (completed != "N" e.g. the completion date)
2) Not completed (completed == "N", if non-anonymized then no entry in response table)
3) Not started (completed == "N", and no entry in response table)
4) Started but not completed (completed == "N", and incomplete entry in response table)

Options 3+4 are only shown for/work for non-anonymized surveys. Things get more interesting when considering multiple responses. I think there are the following cases:

Let uses left start at some number n >= 1

A)Token marked completed (0 uses left, n completed responses, x incomplete responses) [if token based persistence, x = 0, otherwise, x >= 0]

B)Token marked incomplete, no responses (n uses left, 0 responses)

C)Token marked incomplete, no complete responses, some incomplete responses (n uses left, x incomplete responses) [if token based persistence, x will be 1, otherwise x may be >= 1]

D)Token marked incomplete, complete and incomplete responses (n-m uses left, m completed responses, x incomplete responses) [if token based response persistence, x will be 0 or 1, otherwise, x may be >= 0)

Case 1 above (completed), that should be only A? A+D? (Currently A)
Case 2 above (not completed), that should be B+C? B+C+D? (Currently B+C+D if anonymous, **otherwise its B only**)
Case 3 above (not started), that should be only case B here? (Currently B)
Case 4 above (started, but not completed), that should be C only? C+D? (Currently C+D)

Can you guys weigh in on what you think it should be? We could also amend the dropdown options to better reflect the possibilities.

@Shnoulle, to answer your question about whether or not we need SELECT DISTINCT or SELECT, I think you can get rid of DISTINCT. The query already groups by the token and aggregates startdate as the max startdate of all matched responses, so there wouldn't be duplicates anyways.

Shnoulle:
About distinct : yes, it's what i'm think :).

I think 0,1 and 2 can use only token ? I'm unsure what happen with Use left > 1 or Allow multiple response (no token answer persistance + alow multiple or edit), but i don't think we must search response here.

3: Yes : B sems OK. For anonymous survey : we can't search 3, 3 didn't exist
4 : … hum … unsure : C+D or C … sorry …

The issue here : it's 0,1,2,3,4 are done for "use left == 1 at start + token answer persistance + not anonymous. But each survey type need different value/choice …

I think : 
1. we must fix current blank value" BUT with existing behaviour. (i don't test your fix, but seems OK for me)
2. Update system to allow more export options **according to current survey settings** (because , yes : we can change answer persistance after survey start …)

c-schmitz:
Thank you very much!

PR #1189, opened 2018-12-06
Fixed issue #14313: Top part of drop-down box on participants screen is hidden

Shnoulle:
Fixed issue : tokenListActions can be out of screen

Wrap grid table in a duv , with an min-height fi the issue


Shnoulle:
@dominikvitt see the issue : https://github.com/LimeSurvey/LimeSurvey/commit/b9ca9fed3115927704488549dc77b094c1e54784#commitcomment-31584591 with current solution.

Here : we just add a wrapper to be sure to show the dropup :)

PR #1197, opened 2018-12-22
 Fixed issue #14312: Import participant CSV is not possible with surveys update right

Shnoulle:
Dev: fix it in Permission::model, then fixed for RC and plugin too


Shnoulle:
But more : read tokens need only read surveys, not update surveys

PR #1200, opened 2019-01-03
Master 14392 fix theme zip

Shnoulle:
Fixed issue #14392: Don't include some directory or file in theme zip export
Dev: Use same function then import when export for filtering
Dev: remove hidden directory when import and export


Shnoulle:
Broken … need more work

PR #1201, opened 2019-01-04
Fixed issue #14398: Unable to set a different standardthemerootdir than default one

Shnoulle:
Dev: get core template list from standardthemerootdir
Dev: Set DB settings by copnfig.php settings in default TemplateManifest


Shnoulle:
Still an issue after TRUNCATE the theme table : think we must try to install default theme if it's not installed

PR #1202, opened 2019-01-07
Fixed issue #14400: Unable to use self ot that fixed string in Expression

Shnoulle:
Fixed issue #14400: Unable to use self or that fixed string in Expression
Dev: add self,that and this in reserved var for Question rules 
Fixed issue : don't replace title starting with self or that in expression

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -171,13 +171,10 @@ class Question extends LSActiveRecord
             /* ExpressionManager reserved word (partial) */
             $aRules[] = array('title', 'in', 'not' => true,
                 'range' => array(
-                    'LANG',
-                    'SID',
-                    'QID',
-                    'GID',
-                    'SAVEDID',
-                    'SGQ',
-                    'TOKEN',
+                    'LANG','SID', // Global var
+                    'SAVEDID','TOKEN', // current survey related var
+                    'QID','GID','SGQ', // current question related var
+                    'self','that','this', // EM reserved variables
                 ),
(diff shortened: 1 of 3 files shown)


Shnoulle:
Sill have issue with question title as `selfToto`

Shnoulle:
CodeFactor is unhappy :/ But in general : CodeFactor didn't like EM …

application/models/Question.php
```diff
@@ -171,13 +171,10 @@ public function rules()
...
-                    'SID',
-                    'QID',
-                    'GID',
-                    'SAVEDID',
-                    'SGQ',
-                    'TOKEN',
+                    'LANG','SID', // Global var
+                    'SAVEDID','TOKEN', // current survey related var
+                    'QID','GID','SGQ', // current question related var
+                    'self','that','this', // EM reserved variables
```
maziminke: Does that mean that "self" as an answer code will not be accepted anymore?

Shnoulle: As a question title , not answer or Subquestion title

Shnoulle:
@maziminke : i think it's OK now BUT : i like more testing with this fix. Since current 3.15 is old … think i wait next release to merge (and then have a week to test)

application/models/Question.php
maziminke: Ahh, that makes sense. Being able to use it as answer text would be very helpful so if possible let's keep that.

Shnoulle: Yes, no issue in answer text : EM have issue with it currently, but it's an issue in EM.

But clearly : we must not allow it in question title/code

PR #1203, opened 2019-01-08
New feature: Support HTML Video and Audio Elements if XSS filter is activated

Oz-men:
New feature: Add Support for HTML Video and Audio Elements

Dev: This implements video and audio elements into HTML Purifier.

Dev: With this change, video and audio tags will no longer be removed when XSS filter is activated.

### DIFF
--- a/application/core/LSYii_Validators.php
+++ b/application/core/LSYii_Validators.php
@@ -127,10 +127,9 @@ class LSYii_Validators extends CValidator
                 )
         );
 
-	$config = $filter->getPurifier()->config;
+	    $config = $filter->getPurifier()->config;
         $config->set('HTML.DefinitionID', 'html5-definitions');
         $def = $config->maybeGetRawHTMLDefinition();
-        $max = $config->get('HTML.MaxImgLength');
         if ($def) {
             $def->addElement('video', 'Inline', 'Flow', 'Common', array(
                 'controls' => 'Bool',
--- a/application/core/LSYii_Validators.php
+++ b/application/core/LSYii_Validators.php
@@ -108,7 +108,7 @@ class LSYii_Validators extends CValidator
      */
     public function xssFilter($value)
     {
-        $filter = new CHtmlPurifierHelper();
+        $filter = new LSYii_HtmlPurifier();
         $filter->options = array(
             'AutoFormat.RemoveEmpty'=>false,
             'Core.NormalizeNewlines'=>false,
--- a/application/core/LSYii_Validators.php
+++ b/application/core/LSYii_Validators.php
@@ -108,7 +108,7 @@ class LSYii_Validators extends CValidator
      */
     public function xssFilter($value)
     {
-        $filter = new CHtmlPurifier();
+        $filter = new CHtmlPurifierHelper();
         $filter->options = array(
             'AutoFormat.RemoveEmpty'=>false,
             'Core.NormalizeNewlines'=>false,
@@ -127,7 +127,7 @@ class LSYii_Validators extends CValidator
                 )
         );
 
-	    $config = $filter->getPurifier()->config;
+	$config = $filter->getPurifier()->config;
         $config->set('HTML.DefinitionID', 'html5-definitions');
         $def = $config->maybeGetRawHTMLDefinition();
         if ($def) {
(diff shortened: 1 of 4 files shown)


Shnoulle:
No need to update http://htmlpurifier.org/download ?

Else can you extend CHtmlPurifier ? To don't update framework ? (more easy to fix when update (but Yii1 didn't update a lot …)

framework/web/widgets/CHtmlPurifier.php
```diff
@@ -110,7 +110,7 @@ public function getOptions()
 	 * Get the HTML Purifier instance or create a new one if it doesn't exist.
 	 * @return HTMLPurifier
 	 */
```
c-schmitz: I agree that the framwork should not be changed. Can you properly extend CHTMLPurifier instead, please?

application/core/LSYii_Validators.php
```diff
@@ -126,6 +126,27 @@ public function xssFilter($value)
                 'news' => true,
                 )
         );
+
+	$config = $filter->getPurifier()->config;
+        $config->set('HTML.DefinitionID', 'html5-definitions');
+        $def = $config->maybeGetRawHTMLDefinition();
+        $max = $config->get('HTML.MaxImgLength');
```
c-schmitz: What is this line for?

application/core/LSYii_Validators.php
```diff
@@ -126,6 +126,27 @@ public function xssFilter($value)
...
+        if ($def) {
+            $def->addElement('video', 'Inline', 'Flow', 'Common', array(
+                'controls' => 'Bool',
```
c-schmitz: Does this probably open up vulnerabilities like https://html5sec.org/#55 and https://html5sec.org/#56 ?

application/core/LSYii_Validators.php
Oz-men: Line accidentally committed. Now removed.

application/core/LSYii_Validators.php
Oz-men: Tested those and other potential vulnerabilities. They get filtered.

framework/web/widgets/CHtmlPurifier.php
Oz-men: I extended CHtmlPurifier by creating a helper class. However, I don't know how LimeSurvey usually deals with these problems, so feel free to tell me another way if this doesn't comply.

c-schmitz: @Shnoulle Any idea?

Shnoulle: I don't know, in general we extend Yii class in `core` if i don't make error. 

But i have same issue for a project in develop : in helpers it's more easy because you can use autoloader. I don't have fixed idea of the good place.

For example : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSETwigViewRenderer.php or https://github.com/LimeSurvey/LimeSurvey/blob/ba08ebf863baa4dd7b770093c7e1904b7325b92a/application/core/LSCGettextMessageSource.php#L10

`class LSYiiClassName extends YiiClassName` maybe

dominikvitt:
@Oz-men Can you implement this feature without changing framework files?

Oz-men:
> @Oz-men Can you implement this feature without changing framework files?

I did implement it by creating a helper class which extends the framework class.
I now removed that helper class and moved it to core and edited the instantiated class in LSYii_Validators.php to reflect the change.

dominikvitt:
Sorry if I didn't make clear in previous message.
It looks like file  **framework/web/widgets/CHtmlPurifier.php** would still be modified.
You have change from
**protected function getPurifier()**
to  
**public function getPurifier()**

Oz-men:
I'm sorry, I forgot to commit that in my previous commits a few months ago.
Everything should be good now.

dominikvitt:
@Oz-men: Please port this PR to the current develop branch.

olleharstedt:
What's the status for this PR?

cdorin93:
Split into #1589 and #1591 . To be tested soon.
Many thanks @Oz-men for your initial contributions!

PR #1206, opened 2019-01-10
Fixed issue #14415: (partial) unable to use Expression Manager var in submit.twig

Shnoulle:
Dev: no SAVEDID, but core EM var (question code etc …)

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -1184,12 +1184,11 @@ class SurveyRuntimeHelper
                 $this->aSurveyInfo['aLEM']['debugvalidation']['message'] .= "<table><tr><td align='left'><b>Group/Question Validation Results:</b>".$this->aMoveResult['message']."</td></tr></table>\n";
             }
 
-            // kill survey session before redirecting
-            if ($this->aSurveyInfo['printanswers'] != 'Y') {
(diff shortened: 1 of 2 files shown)


application/helpers/SurveyRuntimeHelper.php
```diff
@@ -1208,7 +1207,13 @@ private function moveSubmitIfNeeded()
             if(!$surveyActive) {
                 $this->aSurveyInfo['include_content'] = 'submit_preview';
             }
-            Yii::app()->twigRenderer->renderTemplateFromFile("layout_global.twig", array('oSurvey'=> $oSurvey, 'aSurveyInfo'=>$this->aSurveyInfo), false);
+            $sHtml = Yii::app()->twigRenderer->renderTemplateFromFile("layout_global.twig", array('oSurvey'=> $oSurvey, 'aSurveyInfo'=>$this->aSurveyInfo), true);
+            $oTemplate = Template::model()->getInstance();
```
Shnoulle: I test with default template only, think it take the current one (related to survey)

Shnoulle:
@LouisGac : did you see any potential issue here ? About partial : it's due to SAVEDID not in Expression Manager, but i think it can be fixed in develop only.

Shnoulle:
Thanks :)

PR #1207, opened 2019-01-16
New feature: Warn when activating a survey that is expired by config

TonisOrmisson:
We have had some weird issues with survey settings somehow not being saved. And survey being activated with the expiry date set in past. 

It is only logical to have a warning when one tries to activate a survey that has the configuration that will end up in activating a survey which will not be ready for participating.

### DIFF
--- a/application/views/admin/survey/activateSurvey_view.php
+++ b/application/views/admin/survey/activateSurvey_view.php
@@ -187,7 +187,7 @@
         <p class='col-sm-7 col-sm-offset-2'>
             <?php eT("Please note that once responses have collected with this survey and you want to add or remove groups/questions or change one of the settings above, you will need to deactivate this survey, which will move all data that has already been entered into a separate archived table."); ?><br /><br />
         </p>
-        <?php if($oSurvey->getIsExpiredByConfig()):?>
+        <?php if($oSurvey->getIsDateExpired()):?>
         <div class="row">
             <div class='col-sm-7 col-sm-offset-2'>
 
--- a/application/views/admin/survey/activateSurvey_view.php
+++ b/application/views/admin/survey/activateSurvey_view.php
@@ -1,3 +1,7 @@
+<?php
+/** @var Survey $oSurvey */
+
+?>
 <?php if ((isset($failedcheck) && $failedcheck) || (isset($failedgroupcheck) && $failedgroupcheck)): ?>
 <div class='side-body <?php echo getSideBodyClass(false); ?>'>
     <div class="row welcome survey-action">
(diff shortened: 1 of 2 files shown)


TonisOrmisson:
Looks like that:

screenshot from 2019-01-16 08-41-59

maziminke:
Great feature!

application/views/admin/survey/activateSurvey_view.php
```diff
@@ -183,6 +187,15 @@ function alertPrivacy()
...
         </p>
+        <?php if($oSurvey->getIsExpiredByConfig()):?>
+        <div class="row">
+            <div class='col-sm-7 col-sm-offset-2'>
+
+                <div class="alert alert-danger"><?= Yii::t('app', 'NB! The survey you are activating is currently expired! The survey will not be available for participants.')?><div>
+            </div>
+
+        </div>
+        <?php endif;?>
```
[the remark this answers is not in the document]
TonisOrmisson: I also did it like that at first, but. It does not match the context well and it would give a wrong impression. By having a (past) datetime set for a survey does not actually mean it is expired. If the survey is not activated yet, then it is not expired by the definitions set here: 

https://github.com/LimeSurvey/LimeSurvey/blob/4dc6d4ff14cff15361eb81b1a07c9684905e453e/application/models/Survey.php#L1133..L1159

According to the `survey->state` the state is `inactive`. So having an `isExpired = true` while `->status = 'inactive'` is confusing.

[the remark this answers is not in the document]
Shnoulle: Ah, you mean isExpired must mean : `active== 'Y' && expires < "now"` ?

I really don't like this ByConfig … Maybe something like isDateExpired ?

Else : you put the comment where i just put idea about isStarted(ByConfig|ByDate) ;)

application/views/admin/survey/activateSurvey_view.php
```diff
@@ -183,6 +187,15 @@ function alertPrivacy()
         <p class='col-sm-7 col-sm-offset-2'>
             <?php eT("Please note that once responses have collected with this survey and you want to add or remove groups/questions or change one of the settings above, you will need to deactivate this survey, which will move all data that has already been entered into a separate archived table."); ?><br /><br />
         </p>
+        <?php if($oSurvey->getIsDateExpired()):?>
+        <div class="row">
+            <div class='col-sm-7 col-sm-offset-2'>
+
+                <div class="alert alert-danger"><?= Yii::t('app', 'NB! The survey you are activating is currently expired! The survey will not be available for participants.')?><div>
```
c-schmitz: What's 'NB' ?  Also should be a note not a warning - it is not like you cannot remove the expiration date after activation.
Also the hint should be helpful, not keep the user from activating a survey. So rather like:

"Note: This survey has a past expiration date configured and is currently not available to participants. Please remember to update/remove the expiration date in the survey settings after activation."

TonisOrmisson: Being someone who has activated a number of surveys while the expiry date was set and passed, I would say that treating this as a warning is adequate. 

When one is activating a survey - what is the expected behaviour? The expected behaviour is that after activating the survey will be active and ready to accept responses. If you have an expiry date set in past you will not reach the expected outcome. Thus - a warning, not a note. 

But of course it's juts my opinion. If you really feel it should be less visible, I will change it of course, but I personally will clearly regard this as an warning not a mere note. 

ps.
NB is Nota bene :)
https://en.wikipedia.org/wiki/Nota_bene

Shnoulle: @TonisOrmisson you set it as alert-danger, can be moved to alert-warning ;)

TonisOrmisson: have you seen alert-warning? :)

TonisOrmisson: @Shnoulle I can live with alert-warning :) thanks

c-schmitz:
Thank you!

application/views/admin/survey/activateSurvey_view.php
Shnoulle: Not an issue ;) personally : danger is OK for me, warning more for «active but not started : don't send the url now, but in some day :)». Just point of view, not major.

PR #1208, opened 2019-01-18
Fixed issue #14389: never same order in Theme list

Shnoulle:
Dev: a bad default order is always better then no fixed order


Shnoulle:
Ordering in dropdown too

PR #1211, opened 2019-02-04
Fix #14488: Error with multiple choice type

realitix:
Fixed issue #14488 :

### DIFF
--- a/framework/db/schema/mysql/CMysqlTableSchema.php
+++ b/framework/db/schema/mysql/CMysqlTableSchema.php
@@ -22,4 +22,11 @@ class CMysqlTableSchema extends CDbTableSchema
 	 * Defaults to null, meaning no schema (or the current database).
 	 */
 	public $schemaName;
+
+	/**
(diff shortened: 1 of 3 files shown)


Shnoulle:
1. the issue is only due to `other` then i think it's this rules that mùust be fixed
2. framework must be clean then we can update it more easily

Shnoulle:
Rules is here : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Question.php#L135

Then can surely be fixed here : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSYii_CompareInsensitiveValidator.php

Shnoulle:
But more looking at exception the issue is in database … `$oSubQuestion->title = rand(11111, 99999);`

Shnoulle:
Fixed in https://github.com/LimeSurvey/LimeSurvey/commit/5df2bf4cd28b9d240e11cd6fcdb1c32f2ad75746 thanks for report :)

PR #1213, opened 2019-02-06
Fix#14495: error with P type and comment suffix

realitix:
Fixed issue #14495

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -134,6 +134,10 @@ class Question extends LSActiveRecord
             if ($oParentQuestion->other == "Y") {
                 $aRules[] = array('title', 'LSYii_CompareInsensitiveValidator', 'compareValue'=>'other', 'operator'=>'!=', 'message'=> sprintf(gT("'%s' can not be used if the 'Other' option for this question is activated."), "other"), 'except' => 'archiveimport');
             }
+            // #14495: comment suffix can't be used with P Question (collapse with table name in database)
+            if ($oParentQuestion->type == "P") {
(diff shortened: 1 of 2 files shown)


Shnoulle:
See mantis comment : must fix it in model ( import, api, plugin etc … must use same rules)

realitix:
I updated the pull request with your indication. It works!

Shnoulle:
@realitix OK, seems excellent, thank you.
Just a final question : did you try to fix https://github.com/LimeSurvey/LimeSurvey/blob/51f685f31172bee68802a0885600568377d89845/application/helpers/admin/activate_helper.php#L22 in this pull request ?

- if yes : i wait for the fix
- if no : i squash and merge and see what happen with a previous survey :)

application/models/Question.php
```diff
@@ -134,6 +134,10 @@ public function rules()
             if ($oParentQuestion->other == "Y") {
                 $aRules[] = array('title', 'LSYii_CompareInsensitiveValidator', 'compareValue'=>'other', 'operator'=>'!=', 'message'=> sprintf(gT("'%s' can not be used if the 'Other' option for this question is activated."), "other"), 'except' => 'archiveimport');
             }
+            // #14495: comment suffix can't be used with P Question (collapse with table name in database)
+            if ($oParentQuestion->type == "P") {
+                $aRules[] = array('title', 'match', 'pattern'=>'/.+comment$/', 'not'=>true, 'message'=> gT("'comment' suffix can not be used with question of type P."));
```
Shnoulle: Unsure of what to do with archiveimport ? I think we can't activate since some years : then we can't have LSA.

realitix:
Helle @Shnoulle.
I didn't try to fix the `fixNumbering` function because with this pull request, the problem is fixed earlier.
So you can merge it.

Thanks

Shnoulle:
OK, great :)

PR #1214, opened 2019-02-07
#14499: Add first and last name to the "To" of confirmation email

realitix:
Fixed issue #14499


application/helpers/frontend_helper.php
```diff
@@ -474,6 +474,8 @@ function submittokens($quotaexit = false)
                 //Only send confirmation email if there is a valid email address
             $sToAddress = validateEmailAddresses($token->email);
             if ($sToAddress) {
+                // #14499: Add first and last name to the "To" of confirmation email
```
Shnoulle: Must validate if it's not empty i think

PR #1217, opened 2019-02-12
Fixed issue: JsonEditor to not encode the value if it is already a valid json

TonisOrmisson:
If the value is already a valid JSON, do not encode it
related to 
https://github.com/LimeSurvey/LimeSurvey/pull/634

### DIFF
--- a/application/extensions/yii-jsoneditor/JsonEditor.php
+++ b/application/extensions/yii-jsoneditor/JsonEditor.php
@@ -47,7 +47,13 @@
 		{
 			$htmlOptions = $this->htmlOptions;
             list($name, $id) = $this->resolveNameID();
-			echo CHtml::tag('div', $htmlOptions, CHtml::textArea($name, json_encode($this->value), array(
+			$value = $this->value;
+			// not a json, encoding
+			if (!isJson($this->value)) {
+				$value = json_encode($this->value);
+			}
+
+			echo CHtml::tag('div', $htmlOptions, CHtml::textArea($name, $value, array(
 				'id' => $id,
                 'encode' => false,
 			)));


pitbulk:
I had the same issue and the PR fixed it.

This is an easy fix that should be applied.

Shnoulle:
Don't see why it can broke something.

pitbulk:
@Shnoulle the patch avoid to encode a value that already was json encoded.
It has nothing to do with the jsoneditor library.

Basically if we store a json value, if we save again, with the current implementation it double json encode it.

Adding this patch will prevent to json encode a value that already is a json

Shnoulle:
@pitbulk yes, but why it work in 2.06 ? And not now ? `$this->value` was decoded before ?

@TonisOrmisson : you don't have merge right ? I never use json in plugin settings …

pitbulk:
I tested it on Limesurvey 3.1 and had the issue. I will try now on Limesurvey 4.2

pitbulk:
As you can see in those images, each time I save the form, the value is JSON encoded again and again.

image

image

image

The field was defined as:
``
'idp_multi_support' => [
    'type' => 'json',
    'editorOptions'=> ['mode'=>'text'],
    'label' => 'Multiple IdP support',
    'value' => '<textarea name="idp_multi_support"></textarea>'
]
``

I tested and the bug exists on both Limesurvey3 and Limesurvey4.

TonisOrmisson:
@Shnoulle I have merge right, but to be honest I do not understand the overall policy of what is OK to merge myself and what not. So I generally do not merge stuff myself here :)

Shnoulle:
😄 😄 😄 😿 😿

pitbulk:
any progress?

TonisOrmisson:
@pitbulk I requested a code review from the core team @c-schmitz

PR #1218, opened 2019-02-13
New feature: Add placeholder text for text and numeric responses

adamzammit:
Adds a "Placeholder answer" text box to the "Display" section during question edits

editing

Adds the placeholder attribute to short/long/huge text and numeric responses

placeholder-example

### DIFF
--- a/application/core/QuestionTypes/HugeFreeText/RenderHugeFreeText.php
+++ b/application/core/QuestionTypes/HugeFreeText/RenderHugeFreeText.php
@@ -39,6 +39,7 @@ class RenderHugeFreeText extends QuestionBaseRenderer
         $maxlength = "";
         $withColumn = false;
         $inputsize = null;
+        $placeholder = "";
 
         $drows = $this->setDefaultIfEmpty(
             $this->getQuestionAttribute('display_rows'), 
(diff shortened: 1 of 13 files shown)


maziminke:
This looks pretty similar to the feature "Default answer" which is available for text question types and some others. What's the difference?

TonisOrmisson:
I like it. Why not. 
A placeholder is entirely different than default answer. A placeholder is not the value of the input so the value is empty if not set.

maziminke:
Ahh, so it doesn't really fill the field. Makes sense, nice feature!

Shnoulle:
I like it but o prefer to have it in plugin or theme option …

adamzammit:
If it isn't acceptable as a core feature I'll work on a plugin - let me know.

tammoterhark:
In my opinion this could be a core feature as it may benefit all users and it is not obtrusive.

Shnoulle:
>  Else you would loose a lot of work when switching themes (if you have set placeholders). 
Good point :) 

till think Survey theme can add some question attribute

adamzammit:
I have refactored this and tested it on the develop branch. Please let me know if you are happy for me to merge.

adamzammit:
@c-schmitz are you happy for this feature to be merged into develop?

Shnoulle:
Questuon : i see we send this attribute encoded : it's not the task to view to encode this ? Then the twig file must encode it .

We flatten option here https://github.com/LimeSurvey/LimeSurvey/blob/5bee16c650597a2558055bf5efe274b5ae5bae77/application/views/survey/questions/answer/list_dropdown/rows/option.twig#L13

adamzammit:
@Shnoulle - sorry I don't follow "we send this attribute encoded" - can you please give me some more detail and I'll look into this.

Adam

Shnoulle:
`htmlspecialchars` : we send attribute encoded, i think it's the view that must encode and not the core.

adamzammit:
thanks @Shnoulle I'll fix this.

dominikvitt:
@adamzammit - it looks this feature works okay in current develop version, it shows in question overview, question editor and preview. Please check if everything works as you wanted.

adamzammit:
thanks @dominikvitt it looks good in the dev version. @Shnoulle  - I've made those changes now.

PR #1219, opened 2019-02-14
Master fix num auto theme

Shnoulle:
Fixed issue : usage of bNumRealValue from config
 New feature : allow update of fix num auto in core themes

### DIFF
--- a/assets/scripts/expressions/em_javascript.js
+++ b/assets/scripts/expressions/em_javascript.js
@@ -509,11 +509,6 @@ function LEMval(alias)
     var varName = alias;
     var suffix = 'code';    // the default
     var value = "";
-    if(typeof bNumRealValue == 'undefined'){
-        bNumRealValue=false;
-    } // Allow to update {QCODE} even with text
-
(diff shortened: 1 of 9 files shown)


Shnoulle:
@olleharstedt You ask, i do it …

Shnoulle:
Need revert : https://github.com/LimeSurvey/LimeSurvey/commit/2e13dfb5b3155939851778e7f91c2a72231018de to be complete (else need only one option)

olleharstedt:
OK, will try.

olleharstedt:
Tried to revert and merge this, but for me it does not replicate the behaviour that was the point of Dominik's commit: To make "asd" get deleted.

Shnoulle:
I NEVER say asd was deleted …

Let me check if current «fix» `123.123.123` to an empty value and do same in next commit

olleharstedt:
> I NEVER say asd was deleted …

I think that was our problem. :laughing: 

Thanks for fixing it, it's all good now.

PR #1224, opened 2019-02-18
Fixed issue #14544: Response Table - "completed" uses a smaller font

cdorin93:
Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:


Shnoulle:
No :).

Because with debug =2 : need too in ystrap.css.

But crrently : i think it's better to add it in adminstyle.css and didn't touch to yisstrap.

PR #1227, opened 2019-02-19
New feature: Add reference to group relevance (if any) in question view

TonisOrmisson:
Question view should display all logic that determines whether this question is shown or not. If question has a parent relevance from group applied, it really should be shown.

### DIFF
--- a/application/views/admin/survey/Question/question_view.php
+++ b/application/views/admin/survey/Question/question_view.php
@@ -1,3 +1,6 @@
+<?php
+/** @var Question $oQuestion */
+?>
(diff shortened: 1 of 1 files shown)


Shnoulle:
Why not, maybe add `trim($oQuestion->groups->grelevance)!='1')` too ?

TonisOrmisson:
Minimal impact? Copied from group view.

Shnoulle:
> Minimal impact? Copied from group view.

Right :+1:

LouisGac:
makes sense, thx

PR #1228, opened 2019-02-19
Fix#14495: Enforce filter to avoid 'comment' keyword

realitix:
Fixed issue #14495 : 
I already fixed this issue but we still can use the 'comment' keyword.
With this fix, 'comment' becomes forbidden.


Shnoulle:
Please : `Fixed issue #… ` ;)

PR #1231, opened 2019-02-23
Fixed issue #14563: Save question before attributes loaded remove all attributes

Shnoulle:
Dev: Add an hidden input with false value when editing question
Dev: set it to true in question_advanced_settings


Shnoulle:
For diff : better with _Hide whitespace changes_ ;)

LouisGac:
another great PR from Denis Chenu :+1:

Shnoulle:
Thanks : dis you know/think database.php was used by reote control or another controller than default GUI ?

PR #1232, opened 2019-02-23
New feature #14571: Expression manager string (and html) updatable in template

Shnoulle:
Dev: just move php to twig
Dev: remove the PHP file is cleaner (and broke if one renderPartial still use php)

Need some other : all in `question_help` for starting

Unsure for auto set aSurveyInfo, but think it's really best.


Shnoulle:
For ExpressionManager : i think it can be great to have questionStatus … but don't really know where is it.

LouisGac:
great, thx

Shnoulle:
The most questionnable part was this one : https://github.com/LimeSurvey/LimeSurvey/pull/1232/files#diff-9e735704a8412da2764aa0873fb8619dR533

Maybe need a static variable somewhere ?

PR #1234, opened 2019-02-28
Dev: Updated feature #14584: More with in global plugin settings

Shnoulle:
Dev: same the global settings (except for tab)
Not really a new feature … more an usuability fix …
capture d ecran du 2019-02-28 18-02-28
capture d ecran du 2019-02-28 18-11-46


Shnoulle:
Don't think travis issue are related : https://travis-ci.org/LimeSurvey/LimeSurvey/jobs/499920895#L566

PR #1237, opened 2019-03-05
Fixed issue #14598: Bad order shown in List question

Shnoulle:
Dev: add order by group_order in question_order directly


Shnoulle:
Alternative solution : maybe better ? Unsure for separator here `/` but anything can be better ?
capture d ecran du 2019-03-05 09-50-14

PR #1240, opened 2019-03-06
Fix #14609: Prevent import of survey with SID 0

realitix:
Fixed issue #14609

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -863,11 +863,10 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         }
         $iOldSID = $results['oldsid'] = $insertdata['sid'];
 
-        if ($iDesiredSurveyId != null) {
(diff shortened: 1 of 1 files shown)


Shnoulle:
I think the return value must be tested : and if false (null, 0, empty string): get an new one .

In fact : i think it's happen like this ?

````
$iDesiredSurveyId = empty($iDesiredSurveyId) ? $insertdata['sid'] : $iDesiredSurveyId;
unset($insertdata['sid']);
$insertdata['wishSID'] = null;
if (!empty($iDesiredSurveyId)) {
$insertdata['wishSID'] = GetNewSurveyID($iDesiredSurveyId);
} 
````

Maybe update here : https://github.com/LimeSurvey/LimeSurvey/blob/157b44c446a754ad45ec3eef6ceb4bbb7df4729b/application/helpers/admin/import_helper.php#L875
`

realitix:
I updated with your advice.
I didn't need to unset `sid` because if `withSid` exists, it overwrites `sid`.
And i don't need to check if `$iDesiredSurveyId` is equal to 0 because this check is done in the `GetNewSurveyID` function.

PR #1243, opened 2019-03-13
Fix the user's creation error when first connecting with the AuthCAS plugin

pjenicot:
When attempting to add a user to the database who is try to authentificate with the AuthCAS plugin :

```
Internal Server Error
CDbCommand failed to execute the SQL statement: SQLSTATE [HY000]: General error: 1364 Field 'entity_id' does not have a default value
An internal error occurred while the Web server was processing your request. Please contact the webmaster to report this problem.
```

Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/installer/create-database.php
+++ b/installer/create-database.php
@@ -256,7 +256,7 @@ function createDatabase($oDB){
         $oDB->createCommand()->createTable('{{permissions}}', array(
             'id' =>  "pk",
             'entity' =>  "string(50) NOT NULL",
-            'entity_id' =>  "integer NOT NULL",
+            'entity_id' =>  "integer NOT NULL default 0",
             'uid' =>  "integer NOT NULL",
             'permission' =>  "string(100) NOT NULL",
             'create_p' =>  "integer NOT NULL default 0",


Shnoulle:
Why not fix AuthCas plugin ?

pjenicot:
Because the AuthCas plugin uses the authentication mechanism of Limesurvey and doesn't alter the table schema during its installation.

It seems that the bug has already been raised [[07982: Internal Server Error When adding users](https://bugs.limesurvey.org/view.php?id=7982)]

Shnoulle:
OK, but : 
1. i think it's better to fix it in model : set the defaults when create.
2. And here : you didn't fix when update (but we don't need to update DB for such issue).
3. see https://github.com/Shnoulle/LimeSurvey-ShibbolethAuth/blob/5d9e8edcbf2b6fab7405ff75d2805c3d9878b8ea/ShibbolethAuth.php#L116 to set the entity_id in AuthCas without need to update limesurvey …

pjenicot:
You're right :)

I'm going to ask for an evolution for the AuthCas plugin. [ [AuthCAS.php#L320](https://github.com/univlorraine/limesurvey-cas/blob/eede337aa6135a2441cc3a96e0eaa4fae42e9d32/AuthCAS/AuthCAS.php#L320) ]

I closed this request.

Shnoulle:
PS : i think have it in model/Permission default is a god idea too :)

PR #1246, opened 2019-03-21
New feature #14622: allow to update any email text

Shnoulle:
New feature #09668: attachment in beforeTokenEmail
Fixed issue #12243: beforeTokenEmail didn't happen for remote control
Fixed issue #10529: Unable to send email via some smtp server
New feature : allow to add cc, bcc to any email.

Questions : 

- [ ] remove SendEmailMessage ? Yes or no
- [ ] review the emailType string ? [See current](https://github.com/LimeSurvey/LimeSurvey/blob/45f65b0c2aa9daa294281b01a55a8846a3c000c2/application/core/LimeMailer.php#L43-L47)
- [ ]  Did event()->get('to') must be 3.X plugin compatible
- [ ] Did we add the url to all survey or only if in aUrlsPlaceholders ?
- [ ] Someone have an idna email adress for testing ? Seems PHPMailer have his own system, then our need to be ?
- [ ] Old update are not needed (get('to') event->set('to') for example) : maybe disable the automatic update ? Plugin must fill the "update" array to be updated. See [$event->get('updateDisable')](https://github.com/LimeSurvey/LimeSurvey/blob/45f65b0c2aa9daa294281b01a55a8846a3c000c2/application/core/LimeMailer.php#L505) and [mailSenderToFrom plugin](https://github.com/LimeSurvey/LimeSurvey/blob/45f65b0c2aa9daa294281b01a55a8846a3c000c2/plugins/mailSenderToFrom/mailSenderToFrom.php#L42-L46)


Shnoulle:
Maybe some decision about : 

1. Did we remove SendEmailMessage , i'm for removing it in 4.0 not in 5.0
2. Maybe review the emailType string … i'm copy paste action or try to find a clear string but …
3. Did event()->get('to') must be 3.X plugin compatible
4. Did we add the url to all survey or only if in aUrlsPlaceholders ?
5. Someone have an idna email adress for testing ? Seems PHPMailer have his own system, then our need to be ?
6. I dislike to update and update again now : since the class can be used directly : i think we must totally remove this part https://github.com/LimeSurvey/LimeSurvey/pull/1246/files#diff-7d3d8e9dcd030b4be92f267328c3c15eR496 . Or maybe better don't update by default (Plugin must put update['part'] to true to be updated by the core system.

application/helpers/admin/token_helper.php
```diff
@@ -14,171 +14,72 @@
...
-*
+* Seems used only in remote_control : maybe move it to ?
 * @param integer $iSurveyID
 * @param CActiveRecord[]  $aResultTokens
 * @param string $sType type of notification invite|register|remind|confirm
 * @return array of results
 */
 function emailTokens($iSurveyID, $aResultTokens, $sType)
 {
-    Yii::app()->loadHelper('common');
```
olleharstedt: Hi ho! This function is so different in dev compared to master it's not possible to merge Dominik's ONE-LINE change without major conflicts. Can you have a look?

olleharstedt: This is the fix:

`$fieldsarray["{SID}"] = $iSurveyID`

but fieldsarray is removed.

Shnoulle:
Yes , i take it … let me create a RC test script …

Shnoulle:
No need any update in my manual test
Capture d’écran du 2019-10-17 16-11-45

PR #1247, opened 2019-03-22
Fixed issue #14659: uploaded files filename pattern.

Shnoulle:
Dev: srid_indexForSrid-questionCode_indexInQuestion-OriginalFileName


Shnoulle:
Not really a new feature, right ?

Before
Capture d’écran du 2019-03-22 14-27-11

After
Capture d’écran du 2019-03-22 14-23-38

dominikvitt:
actually, it is a new feature and it should go to develop

PR #1249, opened 2019-03-27
Revert "Dev: Inbetween"

Shnoulle:
This reverts commit 788b9ab73f8b2047d1a21fee0b5c8b939d0f6ba4.

You can not broke API like this : 
1. In my opinion : redirect don't need send email
2. Plugin can send email itself (then 2 mails sent after this update)
3. This broke the API

### DIFF
--- a/application/controllers/RegisterController.php
+++ b/application/controllers/RegisterController.php
@@ -131,12 +131,14 @@ class RegisterController extends LSYii_Controller
         if (empty($this->aRegisterErrors) && $iTokenId && $this->sMessage === null) {
             $directLogin = $event->get('directLogin', false);
             if ($directLogin == true) {
+                if($event->get('sendRegistrationEmail', false)) {
+                    self::sendRegistrationEmail($iSurveyId, $iTokenId);
+                }
                 $oToken = Token::model($iSurveyId)->findByPk($iTokenId);
                 $redirectUrl = Yii::app()->getController()->createUrl('/survey/', array('sid' => $iSurveyId,'token' => $oToken->token, 'lang'=>$sLanguage));
                 Yii::app()->getController()->redirect($redirectUrl);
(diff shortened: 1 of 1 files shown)


Shnoulle:
I merge it tomorrow

lacrioque:
I think it is better to send the email by the base system, and to give the possibility to send it also to plugins used in registration. So I'd not approve this revert.

Shnoulle:
It's a API broker : and plugin can send email , with your Dev : theis broke https://gitlab.com/SondagesPro/SurveyAccess/registerQuick/blob/master/registerQuick.php#L155

It's not because **You** think it's better to send email than it must be sent … there are reason wher he must NOT be sent.

Shnoulle:
@c-schmitz : can we really do plugin API broker in 3.X without any discussion in a `Dev` commit? It's the way of LimeSurvey work ? 
Here : this commit break an old plugin, and there are no way to fix it easily (i must reconstruct whole system with beforeControllerAction).

Shnoulle:
> give the possibility to send it also to plugins used in registration
You don't give a new possibility … your force sending email : no way to disbale this …

maziminke:
Please do not break any API related features. The API is used more and more and breaking things can have a huge impact for lots of other LimeSurvey users and customers. 
In addition to that it would be really nice to keep an eye in plugin compatibility. The plugin system was developed for features not to be broken at each new LimeSurvey core version.

Shnoulle:
I send a fix where all are happy, still new feature : add an event()->get("sendEmail") to send email or not.

Shnoulle:
````
                if($event->get('sendRegistrationEmail', false)) {
                    self::sendRegistrationEmail($iSurveyId, $iTokenId);
                }
````
Then : not an API broker (not send mail for already exitsing pluginà but a new feature.

PR #1252, opened 2019-03-30
Fixed issue #14649: CDbException when a "default answer" option is not compatible with a question type

Shnoulle:
Dev: Add an option to set or not the invalid answer string
Dev: since getValidityString is call only for POSTED value, seem not really needed
Dev: but the param unsure it don't happen if user can not update survey

I like to shown the error on page for admin user, currently it don't work …

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -6942,9 +6942,9 @@
                             $LEM->updatedValues[$sgqa] = $updatedValues[$sgqa] = array('type'=>$qInfo['type'],'value'=>$_SESSION[$LEM->sessid][$sgqa]);
                         }
                         /* Add it in view for user with Permission surveycontent update right */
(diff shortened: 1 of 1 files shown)


Shnoulle:
Maybe : 

1. add a (public|private|protected) function `setInvalidAnswerString($setInvalidAnswerString = true)`
2. Use it here : and use Yii::log : create a new log ? `limesurvey.expressionmanager.invalidanswer.surveyid` ?
3. Find a way to separate in the log : error from default (it's an admin error) and error from user (maybe try to hack system)

Maybe the best way is to create a invalidAnswer class ?

Shnoulle:
Ok, added error for user with updatecontent rights
Capture d’écran du 2019-04-02 09-50-53

And log error for server admin (someone seems trying to broke DB)

Shnoulle:
Updated 
Capture d’écran du 2019-04-02 14-29-18

PR #1253, opened 2019-04-01
14578 create function fix

rene-s:
Fixed issue # : 14578 (partially I guess)
New feature # :
Changed feature # :
Dev: 
Dev: 

This fixes a "created_function() is deprecated" messages when logging in using PHP 7.2.14.


Shnoulle:
My opinion on this bug report : the only way to fix it move to Yii2.

If there are no way to have it in real situation : no need to fix. 
Here :set debugsql to 1 : you have the issue : i check and merge

PR #1256, opened 2019-04-05
[READY] New feature setVariableExpressionEnd event to update knowVars

Shnoulle:
Dev: and sample expressionQuestionHelp plugin
See : https://bugs.limesurvey.org/view.php?id=14206

- [x] Test if it work for help (mantis : https://bugs.limesurvey.org/view.php?id=8309)
- [x] Add QCODE.question plugin for multiple question : mantis : https://bugs.limesurvey.org/view.php?id=8858
- <strike>Allow  QCODE.count etc … for some var (related https://bugs.limesurvey.org/view.php?id=11589 )</strike>
- [x] Allow static DB column : https://bugs.limesurvey.org/view.php?id=14445 (SEED)


Shnoulle:
I wait for plugin priority for QCODE.question  (because QCODE.help is needed too with multiple question …)

maziminke:
Denis, what is a possible use case of this feature? 

And how would it be used? Do you need to install a new plugin first?

Shnoulle:
@maziminke there are a demo plugin … to add QCODE.help here …

Use case : 3 link on https://github.com/LimeSurvey/LimeSurvey/pull/1256#issue-267761092

maziminke:
Thanks for your feedback, Denis!

Shnoulle:
@maziminke i add the SEED etc … plugin to the sample plugin. Think this one is great in core/plugins not in demo.

maziminke:
Thanks for adding SEED support!

Shnoulle:
Have an issue with QCODE.help on same page : ok on page after … js not updated.

maziminke:
@Shnoulle I would also prefer to add this to the core since EM placeholders like {SAVEDID} or the new {SEED} belong to the core.
Could you also add a few sentences to the manual later? I can then take care of the German translation.

Shnoulle:
@maziminke : i already explain why i think make feature in plugin is better

1. More easy to manage Core : less code
2. More easy to create other plugin : sample code
3. You speak for SAVEDID or SEED, but : why not for last action ? Or submitdate Or … **different people, different usage**

Create plugin or function is , in my opinon, to have less line in code to fix , and more improvment to do easily;

Fixing a plugin is easy, fixing core is complex …

Shnoulle:
@maziminke : remind THIS pull request add 4 new feature (existing in mantis) and potentially a lot more new feature …

Shnoulle:
@maziminke : 3rd point, personnaly i don't care about SEED, but best need usage of [SUBMITDATE](https://github.com/LimeSurvey/LimeSurvey/pull/1256/files#diff-7648a98f52a8709a5a38c24f24f0c7bfR39) and DATESTAMP (last action) (i add it in next commit in plugin)

Shnoulle:
Ok, fixes QCODE.help to allow usage of expression on same page

Delete QCODE.count (current statitics of this question) : better in function `getCount(Q00,"A1")` for example (need a improvment in em_core : allow only PHP function)

Shnoulle:
Ready to be checked. I just want to add a checkbox in Survey plugin settings for expressionQuestionHelp and expressionQuestionForAll activate/deactivate.

And maybe rename expressionQuestionForAll ?

Shnoulle:
Travis fix/control to do

- [ ] ls\tests\ArrayNumberCheckboxTest::testArrayCheckboxExpression
- [ ] ls\tests\DateTimeValidationTest::testBasic
- [ ] ls\tests\ScreenOutTest::testNoAnswer
- [ ] ls\tests\SelfExpressionTest::testBasic

Shnoulle:
@olleharstedt : Thank‘s a lot . Don't find new broken test currently. Can you double check ?

(and now : why set as empty array return a null value …)

olleharstedt:
A test that displays the new feature would be nice.

olleharstedt:
Or a manual page. Or both. ;)

Shnoulle:
@olleharstedt since it's Plugin feature : a test for activating plugin, import survey file  ?

Else : there are 3 plugin sample , and i update manual only after is merged.

olleharstedt:
Yes, the test will have to activate the plugin and then perform the test(s), whatever it needs to do.

olleharstedt:
The test can hack the database to set active = 1.

Shnoulle:
OK, i put it in my todo list when this is merged. I still have the min/max and the js compare to do (but for master)

Shnoulle:
2 things to do

- [ ] Fix question.help in same page
- [ ] Add plugin test for the event + 3 plugins

Shnoulle:
question.help in same page not broken : usage of {self.NAOK} get current question for .question or .help

PR #1261, opened 2019-04-09
Dev: Revert "Release 3.17.1+190408"

Shnoulle:
Dev: This reverts commit 9d1f0ebf8152e3ab5646d679db062f246ffeab51.
Dev: This commit is an API broker
Dev: see issue https://bugs.limesurvey.org/view.php?id=14759

### DIFF
--- a/application/config/version.php
+++ b/application/config/version.php
@@ -12,9 +12,9 @@
  */
 
 
-$config['versionnumber'] = '3.17.2';
+$config['versionnumber'] = '3.17.0';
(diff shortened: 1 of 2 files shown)


Shnoulle:
@LouisGac  can you explain how you have error with image ? If imageSrc return false : image function return an empty string : no issue …
After your fix : same things BUT image return a string …

Shnoulle:
Oups … bas revert …

PR #1263, opened 2019-04-15
Fixed issue #14268. LDAP import. duplicate filter does not work

plyusninva:
LDAP import. Checkbox "filterduplicatetoken" and "filterblankemail" send values != '1'. 
there is no need to check the value of the variable, it is enough that it exists in the POST request.
without these changes is not working filter "Filter Duplicates", this fix solve problem in https://bugs.limesurvey.org/view.php?id=14268

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -1704,8 +1704,8 @@ class tokens extends Survey_Common_Action
         if (!Yii::app()->request->getPost('submit')) {
             $this->_renderWrappedTemplate('token', array('ldapform'), $aData);
         } else {
-            $filterduplicatetoken = Yii::app()->request->getPost('filterduplicatetoken') == '1';
-            $filterblankemail = Yii::app()->request->getPost('filterblankemail') == '1';
+            $filterduplicatetoken = Yii::app()->request->getPost('filterduplicatetoken') == true;
+            $filterblankemail = Yii::app()->request->getPost('filterblankemail') == true;
 
             $ldap_queries = Yii::app()->getConfig('ldap_queries');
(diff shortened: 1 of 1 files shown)


plyusninva:
there is no value attribute. the default value is "On", not "1"
https://github.com/LimeSurvey/LimeSurvey/blob/d2adb9703d9f2847413239509525e9c682fbc47d/application/views/admin/token/ldapform.php#L67

similarly:
https://github.com/LimeSurvey/LimeSurvey/blob/d2adb9703d9f2847413239509525e9c682fbc47d/application/views/admin/token/ldapform.php#L59

Shnoulle:
+1 , but then : why not `$filterduplicatetoken = Yii::app()->request->getPost('filterduplicatetoken')` ?

getPost send null if not exist :)

plyusninva:
You're right, but the variable should be written boolean :)

plyusninva:
also taken into account the download via CSV, because it passes the value "1"
https://github.com/LimeSurvey/LimeSurvey/blob/d2adb9703d9f2847413239509525e9c682fbc47d/application/views/admin/token/csvupload.php#L102

Shnoulle:
boolval(Yii::app()->request->getPost('filterblankemail')) === boolval( Yii::app()->request->getPost('filterblankemail')== 'on') === boolval( Yii::app()->request->getPost('filterblankemail')== '1') 

and boolval(null) == boolval("") == boolval("0") == boolval(0)

On if you need to check with on/off : `filter_var(Yii::app()->request->getPost('filterblankemail'),FILTER_VALIDATE_BOOLEAN)`

plyusninva:
You can open your PR, let them choose the best solution.

PR #1264, opened 2019-04-15
Proposed changes to enable password strength rules

alexmarketaccess:
As posted in https://bugs.limesurvey.org/view.php?id=14636#c51463

Note that my proposed changes use the gettranslation feature, which of course should exist if merged into the master repo. 

A complement to this contribution would be to use the config params when LS automatically generates passwords for new user accounts. I didn't require this because I don't send out user passwords via email. That is, in my config I have:

	'display_user_password_in_email' => false,
	'display_user_password_in_html' => false,

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -1,0 +1,77 @@
+<?php if (!defined('BASEPATH')) exit('No direct script access allowed');
+/*
+| -------------------------------------------------------------------
+| DATABASE CONNECTIVITY SETTINGS
+| -------------------------------------------------------------------
+| This file will contain the settings needed to access your database.
+|
+| For complete instructions please consult the 'Database Connection'
+| page of the User Guide.
+|
+| -------------------------------------------------------------------
+| EXPLANATION OF VARIABLES
+| -------------------------------------------------------------------
+|
+|    'connectionString' Hostname, database, port and database type for 
+|     the connection. Driver example: mysql. Currently supported:
+|                 mysql, pgsql, mssql, sqlite, oci
+|    'username' The username used to connect to the database
+|    'password' The password used to connect to the database
+|    'tablePrefix' You can add an optional prefix, which will be added
+|                 to the table name when using the Active Record class
+|
+*/
+return array(
+	'components' => array(
+		'db' => array(
+			'connectionString' => 'mysql:host=localhost;port=3306;dbname=;',
+			'emulatePrepare' => true,
+			'username' => '',
+			'password' => '',
+			'charset' => 'utf8mb4',
+			'tablePrefix' => '',
+		),
+		
+		// Uncomment the following line if you need table-based sessions
+		// 'session' => array (
+			// 'class' => 'application.core.web.DbHttpSession',
+			// 'connectionID' => 'db',
+			// 'sessionTableName' => '{{sessions}}',
+		// ),
+		
+		'urlManager' => array(
+			'urlFormat' => 'path',
+			'rules' => array(
+				// You can add your own rules here
+			),
+			'showScriptName' => true,
+		),
+	),
+	// For security issue : it's better to set runtimePath out of web access
+	// Directory must be readable and writable by the webuser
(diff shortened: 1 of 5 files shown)


TonisOrmisson:
I would not limit the password max length. What would be the benefit of limiting an user not to use longer than 22 chars?

alexmarketaccess:
Valid point for general LS deployment -- I just had a requirement for this side project to put a max length. Maybe if it was set to zero in the config, it could mean unlimited max length.

Shnoulle:
In my opinion, best way to have the parameters is to set in config array

See here for example
https://github.com/LimeSurvey/LimeSurvey/blob/f471785dae1a84bb9c43903d34ac3b457d3d902e/application/config/config-defaults.php#L189-L194

Then
````
$config[passwordValidator'] => array(
    'min' => 4,
    'max' => null,
    'upper' => null,
    'numeric' => null,
    'symbol' => null,
),
````

Else : linked to [Authdb](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/plugins/Authdb/Authdb.php) : maybe it's best is [Authdb](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/plugins/Authdb/Authdb.php)

In fact : only [Authdb](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/plugins/Authdb/Authdb.php) plugin use the password (then this password must be managed by [Authdb](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/plugins/Authdb/Authdb.php))

And finally : only for 4.0 i think

File: application/views/admin/user/modifyuser.php [hunk 1]
```diff
@@ -55,6 +55,9 @@
                     <div class="">
                         <?php echo $form->passwordField($oUser, 'password',array('value'=>'', 'placeholder'=>html_entity_decode(str_repeat("&#9679;",10),ENT_COMPAT,'utf-8'))); ?>
                     </div>
+                    <div class="">
+                        <span class='text-info'><?php eT($passwordHelpText); ?></span>
```
Shnoulle: eT not needed, just echo here

application/models/User.php
```diff
@@ -285,6 +287,81 @@ public function checkPassword($sPassword)
...
+          $lower = $settings['lower'];
+          if($lowercase < $lower) $error = "Password must include at least ".$lower." lowercase letter".($lower != 1 ? 's' : '').".";
+        }
+        if(isset($settings['numeric'])){
+          $numeric = $settings['numeric'];
+          if($number < $numeric) $error = "Password must include at least ".$numeric." number".($numeric != 1 ? 's' : '').".";
+        }
+        if(isset($settings['symbol'])){
+          $symbol = $settings['symbol'];
+          if($specialChars < $symbol) $error = "Password must include at least ".$symbol." special character".($symbol != 1 ? 's' : '').".";
```
Shnoulle: Must use sprintf and gT

application/config/config.php
```diff
@@ -0,0 +1,77 @@
...
+	// on your webspace.
+	// LimeSurvey developers: Set this to 2 to additionally display STRICT PHP error messages and get full access to standard templates
+		'debug'=>1,
+		'debugsql'=>0, // Set this to 1 to enanble sql logging, only active when debug = 2
+	// Update default LimeSurvey config here
+	
+	),
+
+    'params' => array(
+        'passwordValidator' => array(
```
lacrioque: This needs to be set in config-default.php. I think we can start to add this into develop and have a rather low password security setting by default

alexmarketaccess: Seems a reasonable idea to me. I'm not in a position to do coding on this at this time, but if anyone wants to move forward with it, that is great!

dominikvitt:
No problem, I'll implement this feature into develop branch.

dominikvitt:
It is implemented in develop branch:
https://github.com/LimeSurvey/LimeSurvey/commit/b8d7499e05977abffe8811b88588c56f8c74b46c

PR #1266, opened 2019-04-18
Dev: create $qid and $other var for checkValidityAnswer

Shnoulle:
Dev: this don't change anything, but seems more understandable ?
Dev: alternaive : $qid = null;
Dev: if(!empty($qinfo['qid']) { $qid = $qinfo['qid'];}
Dev: if(empty($qid) && !empty($qinfo['info']['qid']) { …


Shnoulle:
The way is set is not important, but think set $qid and $other at start is more understandable , no ?

PR #1267, opened 2019-04-19
Develop order for plugins

Shnoulle:
New feature #12127: Order of plugin 

Needed for example for https://gitlab.com/SondagesPro/ExportAndStats/getStatInSurvey/issues/2#note_162517567 : getStatInSurvey must happen after addScriptToQuestion.
(I have another issue where need to install a plugin after another one but don't find it currently

After in can add a `setOrder` or maybe `setOrderAfter($pluginName)` ? If a new plugin appear ?

### DIFF
--- a/application/config/version.php
+++ b/application/config/version.php
@@ -12,7 +12,7 @@
  */
 
 $config['versionnumber'] ='4.0.0-alpha';
-$config['dbversionnumber'] = 410;
+$config['dbversionnumber'] = 411;
 $config['buildnumber'] = '';
 $config['updatable'] = true;
 $config['assetsversionnumber'] = '30084';
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -2699,7 +2699,7 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
 
         if($iOldDBVersion < 411) {
             $oTransaction = $oDB->beginTransaction();
-            $oDB->createCommand()->addColumn('{{plugins}}','loadorder',"int default 0");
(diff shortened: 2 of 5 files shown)


Shnoulle:
```
<?xml version="1.0" encoding="UTF-8"?>
<config>
    <metadata>
        <name>answersAsReadonly</name>
        <type>plugin</type>
        <creationDate>2016-11-17</creationDate>
        <lastUpdate>2019-04-19</lastUpdate>
        <author>Denis Chenu</author>
        <authorUrl>https://extensions.sondages.pro/addscripttoquestion</authorUrl>
        <supportUrl>https://support.sondages.pro/</supportUrl>
        <version>2.4.2</version>
        <license>GNU Affero General Public License v3.0</license>
        <description><![CDATA[Just to have order]]></description>
    </metadata>
    <order>10</order>
    <compatibility>
        <version>2.73</version>
        <version>3.0</version>
        <version>4.0</version>
    </compatibility>
</config>

```

olleharstedt:
Order = priority? Or load order?

Shnoulle:
loadorder : just "ORDER by loadorder'

You think it's better naming loadorder ?

I check with -100 or 100 : OK.

A system with `dependance` can be better, but more hard to do.

pluginOne need and must be loaded after pluginTwo (need)
pluginOne must be loaded after pluginThree (but only if exist (use))
pluginOne must be loaded before pluginFour

Shnoulle:
Make another one with `priority`

olleharstedt:
`priority` sounds good. :+1: 

For dependence I'd use `require`, as composer does, but only check if required plugin is installed and active. A combination of `require` and `priority` should be able to solve all situations, no?

Shnoulle:
You already have require in develop ?

olleharstedt:
Don't think so, have to check.

PR #1268, opened 2019-04-24
New feature #07494: Adding List-Unsubscribe header for token email

Shnoulle:
Fixed issue : encrypt/decryot issue for token email
Fixed issue : Api don't use publicurl from config


Shnoulle:
There are some issue fixed with decrypt/uncrypt … then i redo it after this issue was fixed …

Shnoulle:
I redo the plugin with mast develop version

PR #1272, opened 2019-05-07
Fixed issue #14858: Upload status is not visible enough

Shnoulle:
API broker or not ?


Shnoulle:
Capture d’écran du 2019-05-07 17-50-50

PR #1273, opened 2019-05-09
Cache result of getQuestionAttributesForEM (WIP)

olleharstedt:
This fix will speed up front-end rendering with 14% on my machine. Benchmarked with:

    $ ab -lk -c 5 -n 50 "http://localhost/limesurvey/index.php?r=survey/index&sid=437521&newtest=Y&lang=pt-BR"

The speed-up will probably be bigger with bigger surveys (more questions -> bigger speed-up).

The crucial question is when to CLEAR the cache. I suggest:

* When newtest=Y is set
* When survey, question or group is saved

NB: Cache is never active when debug = 2.

More discussion here: https://bugs.limesurvey.org/view.php?id=14859


olleharstedt:
@Shnoulle Feel free to comment. :)

application/helpers/expressions/em_manager_helper.php
```diff
@@ -8653,6 +8653,12 @@ public static function UpgradeQuestionAttributes($changeDB=false,$iSurveyID=NULL
         */
         private function getQuestionAttributesForEM($surveyid=0,$qid=0, $lang='')
         {
+            $cacheKey = $surveyid . $qid . $lang;
+            $value = Yii::app()->cache->get($cacheKey);
```
olleharstedt: Cache key should be prefixed with `em_cache` or something, so that we can clear all EM cache objects in one go.

Shnoulle:
When survey, question or group is saved only in my opinion. 
+ preview mode active (group or question) + checksurveylogicfile (force cahce reset by admin user)

Else `$cacheKey = $surveyid . $qid . $lang;` mus be set to `$cacheKey = $surveyid ."_". $qid ."_". $lang;`

Else surveyid 1234 question 12 have same cache id than survey 123241 question 2 (very rare issue, but easy fix)

application/helpers/expressions/em_manager_helper.php
Shnoulle: Why not create a emcache in internal ?
````
		'emcache'=>array(
			'class'=>!empty($userConfig['config']['disableemcache']) ? 'system.caching.CDummyCache' : 'CFileCache',
		),
````

**Comment by Shnoulle** [comment]
Did this cache optimisation must not goes in develop ?

File: application/helpers/expressions/em_manager_helper.php (hunk 1 above)
Comment (olleharstedt) [reply application/helpers/expressions/em_manager_helper.php]: Wouldn't that lead to duplicated configuration? I'd assume cache and emcache would always be the same cache, no?

**Comment by olleharstedt** [comment]
> Did this cache optimisation must not goes in develop ?

Yeah. This is just for discussion and testing.

**Comment by Shnoulle** [comment]
Updated my comments :)

**Comment by olleharstedt** [comment]
> When survey, question or group is saved only in my opinion (in afterSave ?) + preview mode active (group or question) + checksurveylogicfile (**force cache reset by admin user**)

Preview mode always has newtest=Y, no?

File: application/helpers/expressions/em_manager_helper.php (hunk 1 above)
Comment (Shnoulle) [reply application/helpers/expressions/em_manager_helper.php]: current cache is used for DB caching only (and some Yii automatic cache (Language))

Yes, duplicate configuration but : it can be very interesting to have different cache system : one for big data, one for little data, one updated a lot, one not updated a lot etc …

For example : i think createFieldmap can be best in FileCache : because it's very huge (sometimes 2Mo or more …) and loaded one time by user.

**Comment by Shnoulle** [comment]
> Preview mode always has newtest=Y, no?
Not currently (it's true in 2.73, false in 3.X (i must report)) : you see it when you start a survey, and goes to preview question after (see «already submlitted»).

Else : newtest=Y is used by public user, preview question/group ONLY by admin user. 
I think public user don't need to reset cache if we do a good job.

File: application/helpers/expressions/em_manager_helper.php (hunk 1 above)
Comment (Shnoulle) [reply application/helpers/expressions/em_manager_helper.php]: Else : Cache key should be prefixed  by EM + SURVEYID : we need to reset cache by survey, not globally (for all survey).

**Comment by olleharstedt** [comment]
> newtest=Y is used by public user

Really? newtest=Y should never be included in email invitation.

**Comment by Shnoulle** [comment]
> Really? newtest=Y should never be included in email invitation.

Start a survey, click on «Exit on clear survey» , click on reload. And it's not because is not shown is not used …

**Comment by olleharstedt** [comment]
> Start a survey, click on «Exit on clear survey» , click on reload. And it's not because is not shown is not used …

OK, good info, thanks.

**Comment by Shnoulle** [comment]
A good workaround for Call center : endurl with newtest=Y . A lot in forum …

**Comment by Shnoulle** [comment]
More clear about newtest=Y : it's needed to **Reset current value of actual survey** but we use too for **reset current informations for actual survey**

Here : we must separate : current value VS current informations

**Comment by olleharstedt** [comment]
OK, so what about this:

* EM cache is only active when survey is activated
* EM cache is reset when survey is activated
* EM cache is reset when survey/question/group is saved

**Comment by Shnoulle** [comment]
+1
(and reset when survey_logic_files if we made an error ;) )

**Comment by Shnoulle** [comment]
Need : answer/condition updated too
What for token attribute number ?

Maybe quota ? (i don't think)

**Comment by olleharstedt** [comment]
> Need : answer/condition updated too

Well, yeah, everything that can change a relevance expression or any expression need to flush the EM cache when saved. Good point.

**Comment by olleharstedt** [comment]
@c-schmitz This cache cuts loading time of our problematic survey from 800k time units (ms?) to 365k.

**Comment by olleharstedt** [comment]
Note to self: Same should be applied to getGroupInfoForEM.

**Comment by Shnoulle** [comment]
> This cache cuts loading time of our problematic survey from 800k time units (ms?) to 365k.

Cut by 2 : you can do best ;).

Important things : what system cache is used and what session system is used (session in `tmpfs` ,for example)

**Comment by olleharstedt** [comment]
This was the file cache. Maybe better can be done with memcache or something else. Again, mostly visible for big surveys.

**Comment by Shnoulle** [comment]
With fil cache : you move from 800 to 350 ! 😃 then it can be better with memcache \o/ hourra !!!

**Comment by olleharstedt** [comment]
^^ Yeah, it's good (for big surveys - sorry I can't give you my test survey, it's not ours).

File: application/helpers/expressions/em_manager_helper.php [hunk 2]
```diff
@@ -21,6 +21,8 @@
     * @author Denis Chenu <http://sondages.pro>
     */
     Yii::import('application.helpers.expressions.em_core_helper', true);
+    Yii::import('application.helpers.expressions.em_cache_exception', true);
+    Yii::import('application.helpers.expressions.em_cache_helper', true);
```
Shnoulle: I like to see a good way for autoloader to work …

This don't work https://github.com/LimeSurvey/LimeSurvey/blob/2528b1178cc26c384ee519f56cf95cfb7ca4bf5a/application/helpers/questionHelper.php#L14 ?
viewHelper can not be autoloaded because it don't have the namespace.

Shnoulle: Arg … helpers inside subdirectory … maybe `namespace LimeSurvey\Helpers\expressions;`

application/helpers/expressions/em_cache_helper.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+/**
+ * Discussion here: https://bugs.limesurvey.org/view.php?id=14859
+ *
+ * @since 2019-05-23
+ * @author Olle Haerstedt
+ */
+class EmCacheHelper
```
Shnoulle: Currently : usage os only to set surveyinfo ? But have idea to extend it after ?

application/helpers/expressions/em_manager_helper.php
olleharstedt: Separate issue, won't deal with in this PR.

application/helpers/expressions/em_cache_helper.php
olleharstedt: The helper class? Sure, lots of other functionality. See later commits.

application/helpers/expressions/em_cache_helper.php
```diff
@@ -0,0 +1,135 @@
...
+     * @param string $key
+     * @param mixed $value
+     * @return void
+     */
+    public static function set($key, $value)
+    {
+        if (!self::useCache()) {
+            return;
+        }
+
```
Shnoulle: Here, it's not inside $surveyCache ?

application/helpers/expressions/em_cache_helper.php
```diff
@@ -0,0 +1,135 @@
...
+        \Yii::app()->cache->set($key, $value);
```
Shnoulle: `\Yii::app()->emcache->set($key, $value);` if you want to allow different cache for em and for db

Shnoulle: Ok temporary solution for checking :)

application/helpers/expressions/em_cache_helper.php
olleharstedt: Inside?

olleharstedt:
Shit, I just got this error when running my test:

```
..Segmentation fault (core dumped)
```

:D

application/helpers/expressions/em_cache_helper.php
Shnoulle: I mean you don't use `self::$surveyinfo` in set action.

olleharstedt: Hm, let me push. :) You read to fast! XD

application/helpers/expressions/em_manager_helper.php
Shnoulle: I can quikcly check : starting with good name space seems better in my opinion.

olleharstedt: A PR should be as small as possible. Not sure I agree with making changes to namespaces at the same time.

Shnoulle: Maybe when moving to develop then ?

application/config/internal.php
```diff
@@ -186,8 +186,9 @@
             'class' => defined('YII_DEBUG') && YII_DEBUG ? 'system.caching.CDummyCache' : 'CFileCache',
         ),
         // For more info about the emcache, see application/helpers/expressions/em_cache_helper.php.
+        // Disabled by default. Enable by adding emcache in config.php after installation.
```
Shnoulle: Arg … i'm not ready now … and previous default seems OK, no ?

olleharstedt: No, there are risks for bugs, so I don't want to enable it by default until we feel more sure. Also, don't waste time reviewing this until it's more finished. ;)

application/helpers/expressions/em_cache_helper.php
```diff
@@ -144,13 +144,23 @@ protected static function getSurveyCache()
...
-     * True if all conditions are set to use the emcache.
+     * True if all conditions are met to use the emcache.
      *
      * @return boolean
      */
     public static function useCache()
     {
-        // If forced, always use.
+        // Never use from CLI.
+        if (php_sapi_name() === 'cli') {
```
Shnoulle: Since we rae in Yii, maybe `Yii::app() instanceof CConsoleApplication` ?

olleharstedt: Not sure that works with PHPUnit.

olleharstedt: Wait, I needed to remove it to make it testable. See updated code.

olleharstedt:
Using big survey, no welcome page, all-in-one, testing with ab

    ab -lk -c 5 -n 50 "http://localhost/LimeSurvey/index.php/428719?newtest=Y&lang=en"

With emcache: 30 sec
Without emcache 90 sec

Pretty good! Not sure all caching can be kept, didn't try to submit the survey. :d  Also, Travis broken.

Shnoulle:
The all in one page is THE survey to caching !

You test on local ? (for internet speed don't broke test)

olleharstedt:
Yes, on local machine. Apache, PHP module (not fast setup, but shouldn't matter for the test).

Shnoulle:
Thsi one : check static and move it to cache seems a great idea : https://github.com/LimeSurvey/LimeSurvey/pull/1273/files/78fe17af5ea1c087fa171ca631e69174c0c02d1f..9b8a3dd64e0b2f23faf93a960dedc125ea3368ac
:)

olleharstedt:
I just realized that a cached value will skip all plugin events.

Shnoulle:
Oh yes :)

olleharstedt:
Copied into dev branch.

PR #1274, opened 2019-05-09
Dev: debug 2 as default, max_execution_time at 240

Shnoulle:
Dev: A way to debug installer

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -707,11 +707,11 @@ $config['hide_update_key'] = false;
 
 // Dev part
 // 1 : looking for errors, 2 : PHP STRICT error messages
-$config['debug'] = 0;
+$config['debug'] = 2;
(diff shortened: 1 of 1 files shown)


olleharstedt:
Hm, I don't think I support this. Better to just set debug = 2 on your local machine while debugging.

olleharstedt:
OR: Let PHP check for CLI arguments if you install from command-line.

Shnoulle:
It's only for pre-alpha and alpah : reset to 0 by default when beta :)

olleharstedt:
https://www.php.net/manual/en/function.getenv.php

Shnoulle:
It's not really for us, more for tester :).

olleharstedt:
Yes, but test must happen on debug = 0. Some bugs are ONLY visible when debug = 0, and those are the ones we care about most.

Shnoulle:
OK, good point :)

PR #1275, opened 2019-05-09
Dev: Fixed : bad checking of sodium when install

Shnoulle:
Dev: if for debug to get all install error …
With debug forced to 2 : PHP notice about `$sodiumPresent` in installer.

PS : still show warning in PHP7.2


Shnoulle:
@dominikvitt : I don't understand why it return false in my PHP7.2 ? https://www.php.net/manual/en/sodium.installation.php

>  As of PHP 7.2.0 this extension is bundled with PHP. For older PHP versions this extension is available via PECL.

PR #1276, opened 2019-05-11
Fixed issue #14875: No error is shown at debug=0 if DB is broken

Shnoulle:
Dev: really deprecate getGlobalSetting because getGlobalSetting('debug') can update version
Dev: and all is set when needed (see setSetting::model->setSettings …)

### DIFF
--- a/application/controllers/SurveysController.php
+++ b/application/controllers/SurveysController.php
@@ -61,6 +61,10 @@
          */
         public function actionError()
         {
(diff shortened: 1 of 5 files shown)


Shnoulle:
Sorry for push branch in LimeSurvey …

Shnoulle:
I redo it withouth updating injectglobalsetting (do it in develop after merging)

PR #1277, opened 2019-05-12
Dev: Fixed issue : 504 with big DB upgrade

Shnoulle:
Dirty …dirty but work


Shnoulle:
Other test : https://github.com/Shnoulle/LimeSurvey-Shnoulle/commit/e78e672d7cf2dc52bd43b6e3bf25956813dbbe69

Shnoulle:
OK, SQL optimization : https://github.com/LimeSurvey/LimeSurvey/pull/1278 is OK.

Still need a ajax mode i think , but when found times.

PR #1278, opened 2019-05-12
[WIP] Fixed issue #14874: Big DB can send a 504 when upgrade …

Shnoulle:
Dev: SQL usage only , no while
Dev: INSERT INTO with DISTINCT in select

### DIFF
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -2422,7 +2422,8 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
                 'relevance' =>  "text",
                 'modulename' =>  "string(255) NULL"
             ));
-            $oDB->createCommand("INSERT INTO {{questions}} (qid, parent_qid, sid, gid, type, title, preg, other, mandatory, question_order, scale_id, same_default, relevance, modulename) SELECT qid, parent_qid, sid, gid, type, title, preg, other, mandatory, question_order, scale_id, same_default, relevance, modulename from {{questions_old}} GROUP BY qid")->execute();
+            /* Since qid is unique : can be used, but mysql < 5.7.5 throw error , fix NULL to emty string when needed */
+            $oDB->createCommand("INSERT INTO {{questions}} (qid, parent_qid, sid, gid, type, title, preg, other, mandatory, question_order, scale_id, same_default, relevance, modulename) SELECT qid, parent_qid, sid, gid, type, title, COALESCE(preg,''), other, COALESCE(mandatory,''), question_order, scale_id, same_default, COALESCE(relevance,''), COALESCE(modulename,'') from {{questions_old}} GROUP BY qid, parent_qid, sid, gid, type, title, COALESCE(preg,''), other, COALESCE(mandatory,''), question_order, scale_id, same_default, COALESCE(relevance,''), COALESCE(modulename,'') ")->execute();
             $oDB->createCommand()->createIndex('{{idx1_questions}}', '{{questions}}', 'sid', false);
             $oDB->createCommand()->createIndex('{{idx2_questions}}', '{{questions}}', 'gid', false);
             $oDB->createCommand()->createIndex('{{idx3_questions}}', '{{questions}}', 'type', false);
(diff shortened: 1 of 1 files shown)


Shnoulle:
Not tested in pgsql (but i do it when found time)
Not tested in mssql 

`INSERT INTO {{answers}} (qid, code, sortorder, scale_id) SELECT qid, code, sortorder, scale_id FROM {{answers_old}} GROUP BY qid, code, scale_id"` is SQL OK in whole i think ?

Shnoulle:
Remind : it's NOT a php time_limit issue : SQL update and php script still working after 504 (checked via mytop)

Then no set_time_lit can fix this issue.

And : in my opinion: we still need

1. A step by step system
2. Separate 400 upgrade in more function : see : https://github.com/Shnoulle/LimeSurvey-Shnoulle/commit/e78e672d7cf2dc52bd43b6e3bf25956813dbbe69#diff-fb32bdf0b19fecbe2d24aa8be88536e4R2420

application/helpers/update/updatedb_helper.php
```diff
@@ -2634,47 +2620,25 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
...
-                $iPrevScaleId = $row['scale_id'];
-                $iPrevSqid = $row['sqid'];
-                $iPrevSpecialType = $row['specialtype'];
-            }
-            switchMSSQLIdentityInsert('defaultvalues', false);
-            try{ setTransactionBookmark(); $oDB->createCommand()->dropIndex('label_idx_10','{{defaultvaluestemp}}');} catch(Exception $e) { rollBackToTransactionBookmark(); };
-            $oDB->createCommand()->dropTable('defaultvaluestemp');
-            $oDB->createCommand()->createIndex('{{idx1_defaultvalue}}', '{{defaultvalues}}', ['qid', 'scale_id', 'sqid', 'specialtype'], true);
+            ]);
+            $oDB->createCommand("INSERT INTO {{defaultvalues}} (qid, scale_id, sqid, specialtype) SELECT qid, scale_id, sqid, specialtype FROM {{defaultvalues_old}} GROUP BY qid, scale_id, sqid, specialtype")->execute();
```
olleharstedt: Does this work on MSSQL too?

Shnoulle: Didn't know

Shnoulle:
@olleharstedt : didn't check and didn't check with pgsql

@dominikvitt : seems only for mysql < 5.7.5 . https://stackoverflow.com/questions/34913281/select-list-is-not-in-group-by-clause-and-contains-nonaggregated-column i check it.  I think we can add all column too.

Shnoulle:
Sh… `GROUP BY qid, parent_qid, sid, gid, type, title, preg, other, mandatory, question_order, scale_id, same_default, relevance, modulename` : duplicate entry due to modulename null VS empty.

I check it :)

dominikvitt:
When doing insert into question table, qid should be removed because we want new qid to be generated.
Then, new qid would be needed in question_l10ns table, connected to questions table.
New question table would contain less rows then the old one.

Shnoulle:
You CAN‘T remove qid : you broke ALL survey using INSERTANS

In current : https://github.com/LimeSurvey/LimeSurvey/blob/447e78a6a032b6f0ba0101f755c11f983533dabb/application/helpers/update/updatedb_helper.php#L2411
You just set qid as primary index (move for qid,language to qid) : you don't remove qid …

dominikvitt:
If survey has 3 languages, one question in old questions table would contain 3 rows.
After you convert it to new format, table questions would have 1 row and table question_l10ns would have 3 rows. So the problem is that you can't easily convert from 3 rows to 1 row.

Shnoulle:
Current behaviour … copying whole base, do a foreach to delete extta qid … but keeoing the same qid.

Look at the database (and code) : current unique index is [qid,language], you remove extra language. But all other must be same …

Shnoulle:
Maybe you can't easily convert : but you MUST convert … since it must be fixed …

Shnoulle:
Another idea is to fix this difference in https://github.com/LimeSurvey/LimeSurvey/blob/e8d629684f78fdcaabe064bf70abdc3885940d22/application/controllers/admin/checkintegrity.php#L382 and do it before launch update.

Shnoulle:
Arg …
Need fixing (but : how to fix it ?, current do 1 but why not 0 ?)
Capture d’écran du 2019-05-13 13-52-26

olleharstedt:
@dominikvitt First insert language rows (3 rows are 3 rows), then insert question rows (3 rows are 1 row)?

dominikvitt:
@olleharstedt  Yes, it should work

Shnoulle:
Current : 

1. copy questions to question_i10n
2. Delete columns frm question_i10n except qid+question+text
3. Do a foreach in questions to delete multiple lines with same qid : 1st are not deleted , other are deleted
4. Delete language,question and text columns

I want to remove the 3 with a `GROUP BY` , using only qid is OK but …

1. Clean database can have some null values against empty values
2. Dirty database (imported lss here) can have different values for same qid (it muts not, but it happen).

dominikvitt:
@Shnoulle 
Your approach to do everything with insert queries is very good and you should try to make it work.
Also, if DB is dirty, it is impossible to fix.

Shnoulle:
For dirty DB i think to do a preupgrade part :

Search the lines with difference (other than is null VS is empty)
get the value from the 1st line to update the other line.

But after i'm afraid for mssql … i have a mssql server, but can't really playing with DB … (for pgsql : i test and fix when it's ready for mysql).

Shnoulle:
The worst part seems only answers BUT i didn't have a lot of survey with multiple language …

Shnoulle:
Still not working, but https://github.com/Shnoulle/Test-LS-checkDbForUpgrade :D

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1282

Shnoulle:
- [ ] Check on MySQL > 5.7 (MariaDB 10.3.12)
- [ ] Check on MySQL < 5.7
- [ ] Check on Postgre
- [ ] Check on MSsql

PR #1279, opened 2019-05-13
Fixed issue #14875: No error is shown at debug=0 if DB is broken

Shnoulle:


Shnoulle:
Capture d’écran du 2019-05-13 14-53-49

PR #1281, opened 2019-05-20
Just to check

Shnoulle:


dominikvitt:
@Shnoulle I just merged it, please check

Shnoulle:
My coomit don't seem broken … ?

Shnoulle:
Surely need to rebuild admin js and css

dominikvitt:
@Shnoulle You maybe don't know, but in develop branch we don't use gulp anymore to rebuild admin assets.
Instead you should use something like this: 
/usr/local/bin/yarn run build

It is defined in assets/packages/adminbasics/package.json

Shnoulle:
Ok, thanks for information :)

PR #1282, opened 2019-05-21
[READY] Fixed issue #14874: Big DB can send a 504 when upgrade

Shnoulle:
Dev: Usage of INNER JOIN to get Survey->language

Method is here : https://github.com/Shnoulle/Test-LS-checkDbForUpgrade Can not check another currently since i have MySQL < 5.7 only pn a php5.6 system …

- [x] Check on MySQL > 5.7 (MariaDB 10.3.12)
- [ ] Check on MySQL < 5.7
     - [ ] Checked with plugin
- [ ] Check on Postgre 9
     - [ ] Checked with plugin
- [ ] Check on Postgre 10
- [ ] Check on MSsql 
     - [ ] Checked with plugin

### DIFF
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -2558,7 +2558,8 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
             $oDB->createCommand("INSERT INTO {{labels}}
                 (id, lid, code, sortorder, assessment_value)
                 SELECT id, lid, code, sortorder, assessment_value
-                FROM {{labels_update400}} GROUP BY id
+                FROM {{labels_update400}}
+                GROUP BY id, lid, code, sortorder, assessment_value
                 ")->execute();
             switchMSSQLIdentityInsert('labels', false); // Untested
             $oDB->createCommand()->dropTable('{{labels_update400}}');
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -2709,7 +2709,7 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
             $oDB->createCommand()->createIndex('{{idx1_defaultvalue}}', '{{defaultvalues}}', ['qid', 'scale_id', 'sqid', 'specialtype'], false);
             $oDB->createCommand("INSERT INTO {{defaultvalue_l10ns}} (dvid, language, defaultvalue)
                 SELECT {{defaultvalues}}.dvid, {{defaultvalues_update407}}.language, {{defaultvalues_update407}}.defaultvalue
-                FROM {{defaultvaluest}}
+                FROM {{defaultvalues}}
                 INNER JOIN {{defaultvalues_update407}}
                     ON {{defaultvalues}}.qid = {{defaultvalues_update407}}.qid AND {{defaultvalues}}.sqid = {{defaultvalues_update407}}.sqid AND {{defaultvalues}}.scale_id = {{defaultvalues_update407}}.scale_id
                 ")->execute();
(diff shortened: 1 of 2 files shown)


Shnoulle:
Method is here : https://github.com/Shnoulle/Test-LS-checkDbForUpgrade

Shnoulle:
Sh … still get a time out … even if all DB action is done before time out …

Shnoulle:
OK 504 is due to 413 update …

Shnoulle:
Checked on litlle pg file (with multilingual survey, but only one token table) : really quickest

Shnoulle:
All checked done for the DB action https://github.com/Shnoulle/Test-LS-checkDbForUpgrade

Shnoulle:
Arg … thanks for testing …
See the issue : https://github.com/Shnoulle/Test-LS-checkDbForUpgrade/commit/f7bc11c2bf050edec54aabde0d6368cba23fc151

I check again on different DB before apply here.

Shnoulle:
Ok CLI result on mysql

````
Update mysql:host=localhost;port=3306;dbname=ls_develop2; with prefix :lime_ from 358 to 414
Database has been successfully upgraded to version 414 
This process used 2774 ms for its computations
It spent 229 ms in system calls

Update mysql:host=localhost;port=3306;dbname=ls_develop2; with prefix :lime_ from 359 to 414
Database has been successfully upgraded to version 414 
This process used 1672 ms for its computations
It spent 96 ms in system calls
````

Shnoulle:
@dominikvitt Ok, fixed this part. Thanks for testing :)

Else : i check pg 10 and mariadb 10 with develop, but for pg9, mssql and mysql < 5.7 : i can only test with the 3.X plugin. 

PS : currently : no 504 with this version, still have 504 with current.

Shnoulle:
\o/ checked on mysql < 5.17 ?

@c-schmitz : can you check on a mssql DB please ?

dominikvitt:
It works on Microsoft SQL Server 2017 (RTM-CU15) (KB4498951) - 14.0.3162.1 (X64).

Shnoulle:
@dominikvitt for default value : did you check if question in this language exist ? I have it on MSSQL test table. 
Else : did you make a checkintegrety before upgrade ?

Argl … totally right about labels …

Shnoulle:
Can you send ma your SQL db with default value : with my DB : 165 for the 2 part.

Shnoulle:
For label : `SELECT lid,code,sortorder,assessment_value FROM lime_labels GROUP BY lid,code ` do the trick but not mysql compat …

Shnoulle:
@dominikvitt  : lables fixed … sorry.

About defaultvalues : i have same on mssql and i add a checker in https://github.com/Shnoulle/Test-LS-checkDbForUpgrade

My checker show this default values are related to deleted survey
Capture d’écran du 2019-05-25 17-10-16

Can you check with the plugin please (and it's more quick : no way to broke DB : all is done on _test _old _l10ns non existing table on master)

dominikvitt:
I just checked it again: subqid 89 exists.
There are 3 subquestions for multiplechoice question qid 20: 88, 89 and 90, so your query should take all subquestions into account.
Only thing that may matter here is that I don't have default value for main language, which is 'en'.
It looks that problem is defaultvalues_update407 table.

Shnoulle:
Ok,

I get defaultvalues table by Survey->language https://github.com/LimeSurvey/LimeSurvey/pull/1282/files#diff-fb32bdf0b19fecbe2d24aa8be88536e4R2707 except imported survey : i think it was always set.
But right : with multiple choice : can not be set for language …

OK : i fix this. And since it can happen for Imported survey for answer or question too : i review all part where i make a join on Survey->language …

Shnoulle:
Fixed : last update not tested with other sql : but 
` SELECT qid, sqid, scale_id, specialtype GROUP BY qid, sqid, scale_id, specialtype`

No reason of broken SQL

INER JOIN in l10ns : i finally know the usage of specialtype …

PR #1258, opened 2019-04-06
[WIP] Hidden attribute by Expression + hide with css or HTML

Shnoulle:
New feature #[13835](https://bugs.limesurvey.org/view.php?id=13835) : hidden attribute : by Expression Manager
New feature #[14748](https://bugs.limesurvey.org/view.php?id=14748): Really hidden equation question type (not in HTML)
New feature #[11665](https://bugs.limesurvey.org/view.php?id=11665): More ways to hide a question

PR #1283, opened 2019-05-22
14260 mark hidden questions for LS4

mbischof:
Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:


maziminke:
@lacrioque Can you please check if this looks fine and then merge?
Thanks!

Shnoulle:
Can you explain the exact feature here ?

Is this related to such feature : https://github.com/LimeSurvey/LimeSurvey/pull/1258 ?

PS : i don't update this feature when i can test Hidden attribute by Expression + hide with css or HTML …

Shnoulle:
OK, checked : i‘m totally against this commit : only using color/opacity is reall a bad solutiuon.

See my proposal
Capture d’écran du 2019-05-22 12-20-42

Shnoulle:
And nned screenshot.

Ps : if you show it elsewhere : i think we must show EM error before hidden or not …

maziminke:
This is the related bugtracker ticket: https://bugs.limesurvey.org/view.php?id=14260

As outlined before, we simply add a CSS class which allows users to mark hidden questions at their admin theme. Nothing more, nothing less.
We do NOT add any colors or the like so I don't agree to your objections.

Shnoulle:
Then : i don't really understand the point …

Usage of text or data- or informatin on titke is best.

And since css3 is great `[data-attribite-hidden]{color: red}` do the trick for admin template

Shnoulle:
If it's for adding CSS/data- in left part : need

1. Question type class/data (i need it …)
2. Hidden attribute
3. Random group attribute (i need it …)
4. Have or not condition/expression
5. …

lacrioque:
I will close this and reapply it without so many unecessary whitespace changes.
Also data-attributes are applied to the question blocks.

Shnoulle:
Yep data-attributes is great :)

I must find a way for random-group for a client now …

Shnoulle:
@lacrioque : one think more needed maybe : EM error inside question ? 
More PHP time consuming maybe … but things it's more needed than other one.

PR #1285, opened 2019-05-28
[READY] Master checkintegrity

Shnoulle:
Fixed issue #14938: Check data integrity : die with renaming a non existing table

- Leave  Yii exception : clean
- Check really whole table name
- Same for token

### DIFF
--- a/application/controllers/admin/checkintegrity.php
+++ b/application/controllers/admin/checkintegrity.php
@@ -160,7 +160,7 @@ class CheckIntegrity extends Survey_Common_Action
     {
         foreach ($tokenTables as $aTokenTable) {
             Yii::app()->db->createCommand()->dropTable($aTokenTable);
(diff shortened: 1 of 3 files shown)


Shnoulle:
OK , all done and checked.

Lot can be improved using direcltly LEFT JOIN WHERE IS NULL …

Shnoulle:
Surely merge issue in develop, but i merge it …

Else 

- [x] Checked on mssql because it's the original issue

Shnoulle:
I still have an issue : i have single surveyLanguag setting again and again … Seems OK in code, i look again :)

Shnoulle:
WTF : import 2 surveys, delete manually the 2 surveys in lime_surveys, check integrity : show only one survey …
Chek the SQL , do same in SQL : show the 2 surveys …

Shnoulle:
Got it : defaultScope `'index'=>'surveyls_language'` disable multiple id in findAll since key of array is `surveyls_language`

PR #1287, opened 2019-06-03
Dev: feature : allow plugin to use another plugin translation

Shnoulle:
Just update name of name of component , can be improved by `gT($sToTranslate, $sEscapeMode = 'html', $sLanguage = null,$pluginClass = null)`

### DIFF
--- a/application/libraries/PluginManager/PluginBase.php
+++ b/application/libraries/PluginManager/PluginBase.php
@@ -97,7 +97,7 @@ abstract class PluginBase implements iPlugin
 
         // Set plugin specific locale file to locale/<lang>/<lang>.mo
         \Yii::app()->setComponent(
-            'pluginMessages'.$this->id,
+            get_class($this).'Messages',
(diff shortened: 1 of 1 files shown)


olleharstedt:
OK to merge, no?

Shnoulle:
Yes, it's just to know if you have any reason to use plugin->id :)

olleharstedt:
No, looks fine. It's not yet possible to install more than one instance of a plugin (and never will be?).

Shnoulle:
It's not possible to have 2 plugins with same ClassName :). I think PHP broke at 100% here … except when we allow plugin to have namespace (but then get_class return `\namespace\ClassName` : maybe i can check this for LS version 12 ?) 

Maybe it's possible to have 2 plugins with same name (since it's in a variable).

It's the reason why i choose ClassName (and never use $this->name now …)

olleharstedt:
Good point.

PR #1291, opened 2019-06-08
How could I share the print answers to my friend?

suoba:
Some questions are privacy but very fun or useful !  
I am anonymous to do this.
But I also want to share the answers and the report with my friend ! Hope they can join it!
My friend also want to share this to his friend, and other guy maybe repeat this!
How should I do ?

PS:
1. Export PDF  are not good for this ,and it never work well ,always something wrong.
2. I'm not good at coding , only can do step by step.
3. I already install LimeSurvey on my website ,it's Good! thanks to you guys.


Shnoulle:
It's not a clean pull request …
Please : open o forum post or contact a limesurvey partner.

PR #1294, opened 2019-06-18
fix bug 14986 wrong decimals because of math with floats

Joost1982:
Fixed issue # : 14986
New feature # :
Changed feature # :
Dev: Joost1982
Dev: 

More info: https://bugs.limesurvey.org/view.php?id=14986
and: http://www.sqlfiddle.com/#!9/117c55/15


Shnoulle:
1. INSTR didn't exist in pgsql (didn't check with mssql)
2. SUBSTRING_INDEX didn't exist in pgsql (didn't check with mssql)

PR #1296, opened 2019-06-19
fix 14986

Joost1982:
Fixed issue # : 14986 (https://bugs.limesurvey.org/view.php?id=14986)

Replication of the problem (for mysql db):
Download this structure:
https://www.kbanijmegen.nl/Limefiles/limesurvey_survey_763116.lss
run survey and add the number 3.2 in one of the fields, save and export data to SPSS. Or download this survey archive which already has respondent data in it:
https://www.kbanijmegen.nl/Limefiles/limesurvey_survey_763116.lsa

In the SPSS-syntax of the survey archive there will be " V6 F19.17" which is not correct and the syntax will fail.

I've made a fiddle with in which you can see what the problem is (for mysql). It has to do with doing math with floating numbers: http://www.sqlfiddle.com/#!9/117c55/15
Floating-Point Arithmetic: https://floating-point-gui.de/ 

- The first column shows example values as they are stored in the databases (i.e. the values entered by the respondents).
- The second column shows what happens when you are trying to substract floating numbers with this commands: ABS(197558X162X1469SQ001_SQ001) - floor(ABS(197558X162X1469SQ001_SQ001). A lot of extra decimals and seemingly random numbers appear at the end. Those extra numbers create problems later on: using CAST() as CHAR, REVERSE() and MAX() to define $maxDecimal as in line 2250 of https://github.com/LimeSurvey/LimeSurvey/blob/6f2d32c2a47665c41ff81bfe1eb1af43cd03ffb1/application/helpers/export_helper.php#L2250 will make it '81000000000000002.0' and the $decimalMaxLen (line 2256) becomes 17 because strlen(intval(81000000000000002.0)) = 17.
- The third column shows a possible solution by casting the value to a decimal first. This is however problematic because we do not know a priori how many decimals the respondents entered.
- The fourth column shows my final solution (for mysql) by using SUBSTRING_INDEX(). This splits the number in front and after the dot ('.').

(I think this survey(question) also gives an error when using pgsql (a different error; here the $maxDecimal becomes zero), see this fiddle http://www.sqlfiddle.com/#!17/c371d/6  but I'm not entirely sure because I only have a system with mysql; the exact same query in mysql gives the error as above http://www.sqlfiddle.com/#!9/117c55/23)

I've made some changes to export_helper.php to fix this issue, but this needs testing!

### DIFF
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -2246,9 +2246,31 @@ function numericSize($sColumn,$decimal=false)
         if(Yii::app()->db->driverName == 'pgsql') {
             $castedColumnString = "CAST($sColumn as FLOAT)";
         }
-        $maxDecimal = Yii::app()->db
-        ->createCommand("SELECT MAX(REVERSE(CAST(ABS($castedColumnString) - FLOOR(ABS($castedColumnString)) as CHAR))) FROM {{survey_".$iSurveyId."}}")
-        ->queryScalar();
+	/* pgsql */
+        if(Yii::app()->db->driverName == 'pgsql') {
+            $maxDecimal = Yii::app()->db
+            ->createCommand("SELECT MAX(CAST(nullif(split_part($castedColumnString, '.', 2),'') as integer))
+			    FROM {{survey_".$iSurveyId."}}")	
+            ->queryScalar();
+	/* mssql */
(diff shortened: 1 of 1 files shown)


lacrioque:
Thank you very much for your PR, sometimes we need a few days to check everything!
Merged now.

PR #1297, opened 2019-06-19
New feature #16781: autodetection of path in spss syntax

Joost1982:
I've added automatic path detection for the data file in the spss syntax so you do not have to edit the FILE= line. Everytime you save the syntax and the data file in the same folder, the path is automatically detected. Python is needed, but that is included in all newer versions of SPSS, see this table on https://www.spss-tutorials.com/python-for-spss-installing-and-testing/ .

I've made the automatic path detection optional. Next to the "Prior to 16" and '16 or up" I have added an extra option "version with Python Plugin / Essentials" for this new "feature". I edited the instructions accordingly.

We are using this in our company, maybe others find it useful too...

### DIFF
--- a/application/controllers/admin/export.php
+++ b/application/controllers/admin/export.php
@@ -455,7 +455,7 @@ class export extends Survey_Common_Action
             header("Content-type: text/comma-separated-values; charset=UTF-8");
             header("Cache-Control: must-revalidate, no-store, no-cache");
 
-            if ($spssver == 2) {
+            if ($spssver == 2 || $spssver == 3) {
                 echo "\xEF\xBB\xBF";
             }
             $sNoAnswerValue = Yii::app()->getRequest()->getPost('noanswervalue');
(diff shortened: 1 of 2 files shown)


cdorin93:
Hello, and thanks for your contribution!
Is it possible to rebase and solve the conflicts before having it tested?

Joost1982:
> 
> 
> Hello, and thanks for your contribution!
> Is it possible to rebase and solve the conflicts before having it tested?

Done! :)

cdorin93:
Looks good!
Do you think there is any other name we can give to the version to keep it shorter? Maybe Only "SPSS with Python Essentials"? Just thinking how it would be easier for users to understand the feature without reading the textbox below.

Joost1982:
I've updated the description.

> can the functionality be made part of SPSS vers 16 or new option? What would be the cons?

I would not do that. When someone has SPSS without Python support the syntax will throw an error. Python support is not installed by default. (the user is prompted to install "IBM SPSS Statistics - Essentials  for Python" during installation)

Joost1982:
> I've updated the description.

Now one of the tests fails. How is that possible? I've only edited the button description?

c-schmitz:
@Joost1982 Can you update it one more last time, please?

Joost1982:
> @Joost1982 Can you update it one more last time, please?

Done! :)

c-schmitz:
Thank you!

PR #1300, opened 2019-06-25
Fixed issue #15004 show correct id attribute for additional attributes

evently-nl:
Fixed issue #15004 Show correct id attribute for additional attributes on the registration form

Dev: should be merged in version Version 3.17.* as well


Shnoulle:
@evently-nl : do you understand why extra attribute have input-sm ?

evently-nl:
> @evently-nl : do you understand why extra attribute have input-sm ?

No idea, probably to make them seem less prominent than the first and last name, although the size difference is really small. It's on the e-mail field as well which is weird as that one is alway mandatory..

Shnoulle:
Yes … i remove whole input-sm in registerQuick …

PR #1305, opened 2019-07-15
Dev: Fixed #15048: Broken HTML in questionexplorer

Shnoulle:
Dev: build for production


lacrioque:
I don't remember exactly why, but i had a reason to not use li there.
I think it would be better to change the <ul> toa <div>

Shnoulle:
No issue with li. And since there are ul : it MUST be li

lacrioque:
It is bootstraps listgroup, meaning that the ulö can easily be made to a div.
I have checked, why I used div instead of li, the rendering needs a block item in the list.
I will change the ul to div.

Shnoulle:
Since it's a list : ul>li is better,

But … make like you want … like all time …

PR #1307, opened 2019-07-20
New Feature: Add an event to generate custom tokens

Jan-E:
New issue to separate the discussions in https://github.com/LimeSurvey/LimeSurvey/pull/1104

### DIFF
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -295,7 +295,7 @@ abstract class Token extends Dynamic
         foreach ($tkresult as $tkrow) {
             $bIsValidToken = false;
             while ($bIsValidToken == false && $invalidtokencount < 50) {
-                $newtoken = $this::generateToken($iTokenLength);
+                $newtoken = $this->_generateRandomToken($iTokenLength);
                 if (!isset($existingtokens[$newtoken])) {
                     $existingtokens[$newtoken] = true;
                     $bIsValidToken = true;
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -214,19 +214,26 @@ abstract class Token extends Dynamic
      * Generates a token for this object.
      * @throws CHttpException
      */
-    public function generateToken()
+    public function generateToken($tokenlength = NULL)
     {
-        $iTokenLength = $this->survey->tokenlength;
-        $this->token = $this::generateRandomToken($iTokenLength);
+        $iTokenLength = $tokenlength ? $tokenlength : $this->survey->tokenlength;
+        $this->token = $this->_generateRandomToken($iTokenLength);
         $counter = 0;
         while (!$this->validate(array('token'))) {
-            $this->token = $this::generateRandomToken($iTokenLength);
+            $this->token = $this->_generateRandomToken($iTokenLength);
             $counter++;
             // This is extremely unlikely.
-            if ($counter > 10) {
-                throw new CHttpException(500, 'Failed to create unique token in 10 attempts.');
+            if ($counter > 50) {
+                throw new CHttpException(500, 'Failed to create unique token in 50 attempts.');
             }
         }
+        $event = new PluginEvent('afterGenerateToken');
+        $event->set('surveyId', $this->getSurveyId());
+        $event->set('iTokenLength', $iTokenLength);
+        $event->set('oToken', $this);
+        $event->set('token', $token);
+        App()->pluginManager->dispatchEvent($event);
+        $token = $event->get('token');
     }
 
     /**
@@ -234,8 +241,10 @@ abstract class Token extends Dynamic
      *
      * @param integer $iTokenLength
      * @return string
+     *
+     * Be sure to call only from generateToken() to facilitate the pluginevent 'afterGenerateToken'
      */
-    public static function generateRandomToken($iTokenLength)
+    private function _generateRandomToken($iTokenLength)
     {
         return str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
     }
@@ -287,7 +296,7 @@ abstract class Token extends Dynamic
         foreach ($tkresult as $tkrow) {
             $bIsValidToken = false;
             while ($bIsValidToken == false && $invalidtokencount < 50) {
-                $newtoken = $this::generateRandomToken($iTokenLength);
+                $newtoken = $this::generateToken($iTokenLength);
                 if (!isset($existingtokens[$newtoken])) {
                     $existingtokens[$newtoken] = true;
                     $bIsValidToken = true;
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -218,11 +218,9 @@ abstract class Token extends Dynamic
     {
         $iTokenLength = $tokenlength ? $tokenlength : $this->survey->tokenlength;
         $this->token = $this->_generateRandomToken($iTokenLength);
-        $this->token = $this->_afterGenerateToken($iTokenLength);
         $counter = 0;
         while (!$this->validate(array('token'))) {
             $this->token = $this->_generateRandomToken($iTokenLength);
-            $this->token = $this->_afterGenerateToken($iTokenLength);
             $counter++;
             // This is extremely unlikely.
             if ($counter > 50) {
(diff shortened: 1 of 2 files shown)


Jan-E:
In reply to @Shnoulle in https://github.com/LimeSurvey/LimeSurvey/pull/1104#issuecomment-513457586

I do not mind reshuffling the order of the commands. And the name of the pluginevent is something for the LimeSurvey devs to decide. The real trouble is in getting the sid. I tried various methods. The first 3 methods are used somewhere in Tokens.php:
* $event->set('surveyId', $this->getSurveyId());
* $event->set('surveyId', $this->dynamicId);
* $event->set('surveyId', $this->getDynamicId());
* $event->set('surveyId', intval($_SESSION['LEMsid']));

However, the first 3 fail, apparently throw an Exception and generate an error 500.
Just try your proposed method and add only this to the `function generateRandomToken`, nothing else:

`$iSurveyId = $this->getSurveyId();`

Only the method using the \$_SESSION succeeds when you try to Generate tokens for a survey. Result of the others: error 500. Just a educated guess: is the error caused by the fact that the `generateRandomToken` function is not called by a scenario like `Token::model(12345)->generateRandomToken`

BTW. You forgot to pass the \$iTokenLength to the plugin:

`$event->set('iTokenLength', $iTokenLength);`

Shnoulle:
I look for `getSurveyId` , `$this->getSurveyId() == $this->dynamicId == $this->getDynamicId()\

Using $_SESSION is a bad solution since it was unsure …

Shnoulle:
PS : muts move to `develop` … sorry

Shnoulle:
My opinion : 
1. https://github.com/LimeSurvey/LimeSurvey/blob/e5d117578b8db82df1ba56bf3e26a72077a27759/application/models/Token.php#L244 to `private static function generateRandomToken`
2. https://github.com/LimeSurvey/LimeSurvey/blob/e5d117578b8db82df1ba56bf3e26a72077a27759/application/models/Token.php#L223 to `public function generateToken($tokenLegth = null)`
3. https://github.com/LimeSurvey/LimeSurvey/blob/e5d117578b8db82df1ba56bf3e26a72077a27759/application/controllers/admin/tokens.php#L990 to `$token->generateToken($aData['tokenlength']);`
4. Add the Event inside https://github.com/LimeSurvey/LimeSurvey/blob/e5d117578b8db82df1ba56bf3e26a72077a27759/application/models/Token.php#L223 function (maybe with counter ?)

Jan-E:
I guess this must be it:

1. In tokens.php: change the call from addDummies() to generateRandomToken() into a call to generateToken()
2. Because there already is a loop in generateToken() I removed the loop in addDummies(), but increased the attempts in generateToken to 50 (which was the failure threshold in addDummies)
3. Change the call in generateTokens() to generateRandomToken() into a call to generateToken()
4. Change the `public static function generateRandomToken` into a `private function _generateRandomToken` and call that function only from within `generateToken()`
5. Change `function generateToken()` into `function generateToken($tokenlength = NULL)`
6. If a tokenlength argument is passed to generateToken() that one is used, otherwise `\$this->survey->tokenlength`
7. Add a call to a new plugin event `afterGenerateToken`
8. Validate the generated custom token: use it in the attempts loop. Use a private function _afterGenerateToken to invoke the plugin event in stead of adding the whole code twice in generateToken()

@Shnoulle Pleaase review the code changes.

Jan-E:
> PS : must move to `develop` … sorry

@Shnoulle I will close this PR and create a new one for the develop branch, once this discussion is completed.

application/models/Token.php
```diff
@@ -214,28 +214,49 @@ public function findByToken($token)
...
     /**
      * Creates a random token string without special characters
      *
      * @param integer $iTokenLength
      * @return string
+     *
+     * Be sure to call only from generateToken() to facilitate the pluginevent 'afterGenerateToken'
      */
-    public static function generateRandomToken($iTokenLength)
+    private function _generateRandomToken($iTokenLength)
```
Shnoulle: Can be static

Jan-E: In PHP 7 non-static seems to be faster. See https://www.codedwell.com/post/59/static-vs-non-static-methods-in-php with a timing example for PHP 7.2.6. Besides that it is a bit odd to make a private function static, because it should never be called as $classname::aStaticMethod() from outside the class.

Jan-E: On second thoughts: since _afterGenerateToken needs `$this` we might even merge _generateRandomToken and _afterGenerateToken. I will test the third commit locally.

Jan-E: Commit 3 proved to be OK (tested in the train from Amsterdam to Rotterdam). In the end we are almost back at the initial idea, except that $_SESSION is avoided and the calling sequence has been streamlined.

Jan-E: Would you like the PR to develop as 1 commit or with these 3 commits, so that the history is preserved?

Shnoulle:
Always one commit please :)

Shnoulle:
Approved for 4.0 :) If you can provide a sample plugin for only numeric it's really cool :)

Jan-E:
@Shnoulle there is a serious problem with the current patch. Up until now I tested with Generate tokens. That went fine. But 'Create dummy tokens' throws an error:
> tokens and its behaviors do not have a method or closure named "_generateRandomToken".

If I revert all the changes in Token.php and keep only those in application/controllers/admin/tokens.php the error is another, but as severe:
> Property "tokens.survey" is not defined.

I do not really grasp (yet) why it fails. add_participants() in the RPC uses comparable code with a call to generateToken():
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/remotecontrol/remotecontrol_handle.php#L1738

Comparable code is in application/controllers/RegisterController.php as well:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/RegisterController.php#L413
I must assume the call to generateToken() there does not fail.

Then why does a call to generateToken() in tokens.php at this point fail:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/tokens.php#L977 ???

Jan-E:
Found it: https://github.com/LimeSurvey/LimeSurvey/pull/1307/commits/3f35b036e14d939455078e47853e345f583492d4

Jan-E:
Plugin is here: https://github.com/toolsforresearch/TFRcustomToken

Jan-E:
BTW While testing Generatetokens() in LS 4.0.0-RC1 I ran into
> 500: Internal Server Error
> Call to a member function decrypt() on array

One should not call 'member' functions on a non-object like an array. The error occurs in findAllAsArray()
https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/models/LSActiveRecord.php#L103
The line
`        return $records->decrypt();`
should be changed back into
`        return $records;`
git blames this commit https://github.com/LimeSurvey/LimeSurvey/commit/bc1a1a290a9f8bc9dca43d82a191eceb8a538769 by by @dominikvitt

application/models/Token.php
olleharstedt: About commit, we can squash when we merge anyway, so there will always be one commit no matter how many commits there are in the PR.

olleharstedt:
This should be for dev, not master, no?

Shnoulle:
@olleharstedt : yes it's the project : https://github.com/LimeSurvey/LimeSurvey/pull/1307#issuecomment-513494109

@Jan-E : about plugin : i mean plugin inside LimeSurvey core. Unsure of the place : plugins/ seems best in my opinion. see [mailSenderToFrom](https://github.com/LimeSurvey/LimeSurvey/tree/develop/plugins/mailSenderToFrom)  for example : thius fix a mantis LimeSUrvey issue (and the [2 new plugins](https://github.com/LimeSurvey/LimeSurvey/pull/1256/files#diff-a084b794bc0759e7a6b77810e01874f2R75) in this pull request : [New feature setVariableExpressionEnd ](https://github.com/LimeSurvey/LimeSurvey/pull/1256))

Jan-E:
New PR to develop here: https://github.com/LimeSurvey/LimeSurvey/pull/1308

Jan-E:
@olleharstedt 
> This should be for dev, not master, no?

We need it in master too, for our use case in https://bugs.limesurvey.org/view.php?id=15026
Will it be backported from dev to Limesurvey 3.17.x? I tested https://github.com/LimeSurvey/LimeSurvey/pull/1308 both in 4.0.0-RC1 as well as in 3.17.x.

PR #1308, opened 2019-07-22
New feature: add afterGenerateToken event with customToken plugin

Jan-E:
@Shnoulle @olleharstedt 
This PR contains the same changes as in https://github.com/LimeSurvey/LimeSurvey/pull/1307, but now to the develop branch. Let us continue the conversation here. I will close https://github.com/LimeSurvey/LimeSurvey/pull/1307

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,7 +72,7 @@
 !/plugins/Demo/
 !/plugins/AuditLog/
 !/plugins/mailSenderToFrom/
-
+!/plugins/customToken/
 
 #ignore old styles
 /styles/
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -985,15 +985,7 @@ class tokens extends Survey_Common_Action
                 $token->lastname = str_replace('{TOKEN_COUNTER}', $newDummyToken, $token->lastname);
                 $token->email = str_replace('{TOKEN_COUNTER}', $newDummyToken, $token->email);
 
-                $attempts = 0;
-                do {
-                    $token->token = Token::generateRandomToken($aData['tokenlength']);
-                    $attempts++;
-                } while (isset($existingtokens[$token->token]) && $attempts < 50);
-
-                if ($attempts == 50) {
-                    throw new Exception('Something is wrong with your random generator.');
-                }
+                $token->generateToken($aData['tokenlength']);
 
                 $existingtokens[$token->token] = true;
                 $token->encryptSave();
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -220,17 +220,17 @@ abstract class Token extends Dynamic
      * Generates a token for this object.
      * @throws CHttpException
      */
-    public function generateToken()
+    public function generateToken($tokenlength = NULL)
     {
-        $iTokenLength = $this->survey->tokenlength;
-        $this->token = $this::generateRandomToken($iTokenLength);
+        $iTokenLength = $tokenlength ? $tokenlength : $this->survey->tokenlength;
+        $this->token = $this->_generateRandomToken($iTokenLength);
         $counter = 0;
         while (!$this->validate(array('token'))) {
-            $this->token = $this::generateRandomToken($iTokenLength);
+            $this->token = $this->_generateRandomToken($iTokenLength);
             $counter++;
             // This is extremely unlikely.
-            if ($counter > 10) {
-                throw new CHttpException(500, 'Failed to create unique token in 10 attempts.');
+            if ($counter > 50) {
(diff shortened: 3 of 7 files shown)


Jan-E:
In reply to https://github.com/LimeSurvey/LimeSurvey/pull/1307#issuecomment-513716509
> @Jan-E : about plugin : i mean plugin inside LimeSurvey core. Unsure of the place : plugins/ seems best in my opinion. see [mailSenderToFrom](https://github.com/LimeSurvey/LimeSurvey/tree/develop/plugins/mailSenderToFrom) for example : thius fix a mantis LimeSUrvey issue (and the [2 new plugins](https://github.com/LimeSurvey/LimeSurvey/pull/1256/files#diff-a084b794bc0759e7a6b77810e01874f2R75) in this pull request : [New feature setVariableExpressionEnd ](https://github.com/LimeSurvey/LimeSurvey/pull/1256))

OK, now I see what you mean. I will see what I can do.

Jan-E:
OK, I added a customToken plugin and tested it in LS 4.0.0-RC1 and LS 3.17.0+190402.
What next?

Jan-E:
Is anything blocking this to be merged? We need at least the first commit (the new afterGenerateToken event) for our use case in https://bugs.limesurvey.org/view.php?id=15026:
15026: Omit ambiguous characters from token

Jan-E:
I have merged the customToken plugin part of the PR into 1 commit.
Please review and merge.

Shnoulle:
Personnaly , i think it's a really good pull request :) Allowing to generate token by an external API can be needed too :)

Jan-E:
> Allowing to generate token by an external API can be needed too :)

Tokens generated by `add_participants($sSessionKey, $iSurveyID, $aParticipantData, $bCreateToken = true)` in the RPC will already follow the customToken settings.

The same applies to `getTokenId($iSurveyId)` in the RegisterController class
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/RegisterController.php#L413

Both do essentially:
1. Initiate the Token class: `$token = Token::create($iSurveyID)`
2. Add a generated token to that class: `$token->generateToken()`

addDummies() will follow the same procedure, once this PR is merged.

Jan-E:
@Shnoulle What is the standard procedure for new features like this? Merge it to develop will only add it to 4.0.0-RCx, I guess. When and how will it be backported to 3.17.x?

Shnoulle:
It was never BackPorted …

About new feature : i have same issue than you : like to have some new (little) feature in 3.X , but i didn't decide … see https://github.com/LimeSurvey/LimeSurvey/pull/1271 for example … light featire, can not broke anything but …

Jan-E:
I asked the same to @olleharstedt : https://github.com/LimeSurvey/LimeSurvey/pull/1307#issuecomment-514147618

If we want it in 3.17.x and it will not be backported, what is the way to go? Create another PR for the master branch?

Shnoulle:
@Jan-E : only @c-schmitz can answer.

Jan-E:
@c-schmitz I can imagine that adding a new plugin to LS 3.x is a step too far. Will a PR with only the new afterGenerateToken event to master be approved?

LouisGac:
If everybody agrees on it, I bet Carsten will aprove it.

Jan-E:
It will be the first commit in this PR: https://github.com/LimeSurvey/LimeSurvey/pull/1308/commits/dbc8d47cdaeecf2608954e8abd8f5b099d04ba51

Shnoulle:
> If everybody agrees on it, I bet Carsten will aprove it.

I ask for some really minor feature‘s in master some week ago, and the answer are «All new feature for 4.0».
Then , if this one is OK : i NEED a clear situation about new feature !

Jan-E:
Changes in that commit summarized:
1. In tokens.php: change the call from addDummies() to generateRandomToken() into a call to generateToken()
2. Because there already is a loop in generateToken() I removed the loop in addDummies(), but increased the attempts in generateToken to 50 (which was the failure threshold in addDummies)
3. Change `function generateToken()` into `function generateToken($tokenlength = NULL)`
4. If a tokenlength argument is passed to generateToken() that one is used, otherwise `$this->survey->tokenlength`
5. Change the `public static function generateRandomToken` into a `private function _generateRandomToken`
6. Change the calls to `$this::generateRandomToken()` into `$this->_generateRandomToken()`
7. Add a call to a new plugin event `afterGenerateToken` in `_generateRandomToken()`

Shnoulle:
[Issue reopen as feedback.](https://bugs.limesurvey.org/view.php?id=15026)

Jan-E:
Thanks for merging,

PR #1315, opened 2019-08-29
fix JS error in firstStartTour onShow

m-unkel:
Fixes an invalid JavaScript code within the firstStartTour tutorial.

Issue:
Browser console output: "console.ls.logT is not a function"

Fix:
Remove (debug) call to invalid function "console.ls.logT" in firstStartTour onShow code.


lacrioque:
+1 for merging
Thank you for spotting that typo!

PR #1316, opened 2019-09-03
fix: Modify the exported stata file encoding to utf-8

wuzhihui1123:
The exported stata file will have strange content When the survey response contains non-English content


Shnoulle:
I don't know STATA : UTF8 is accepted ? In all STATA version ?

For example some [triple-s compatible tool accept only ANSI](https://gitlab.com/SondagesPro/ExportAndStats/exportTripleS/blob/master/exportTripleS.php#L100).

cdorin93:
Hmm, what is the status of this PR - is there any ticket for this one? :thinking:

PR #1317, opened 2019-09-04
Fixed issue #15228: This survey throw a Error when try to print

Shnoulle:
- Use parents relation to be sure to have fixed gid
- https://bugs.limesurvey.org/view.php?id=15228


Shnoulle:
@c-schmitz : since you have https://bugs.limesurvey.org/view.php?id=15229 this one is related

LouisGac:
Seems ok to merge.

PR #1318, opened 2019-09-06
Fixed issue #15237: Upload 3gpp, amr, aac audio files using File Upload Question

Shnoulle:
Dev: allowing to set mime database from config
Dev: allowing to set magic file (extension by magic type)
Dev: review file to check : ico can have image/vnd.microsoft.icon mime type (not in system.utils.fileExtensions)

Currently, checking with `'magic_database' => '/usr/share/misc/magic.mgc',` in Fedora : have ` finfo_file(): Null byte in regex ` … unsure it was Fedora issue or not …

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -557,6 +557,21 @@ $config['iFileUploadTotalSpaceMB'] = 0;
 
 $config['uniq_upload_dir'] = false; // Use a single KCFinder upload directory for all surveys
 
+/**
+ * Allow to use a different MIME database for finfo_open
+ * @see https://www.php.net/manual/en/function.finfo-open.php
+ * Example : '/usr/share/misc/magic.mgc' for redhat based linux
+ */
+$config['magic_database'] = null;
+
+/**
+ * Allow to use a different magic file array 
+ * @see https://www.yiiframework.com/doc/api/1.1/CFileHelper#getExtensionByMimeType-detail
+ * This file must return a PHP array of extension by mimeTypes
+ * Example : https://github.com/LimeSurvey/LimeSurvey/blob/master/framework/utils/fileExtensions.php
+ */
+$config['magic_file'] = null;
+
 
 // defines if the CKeditor toolbar should be opened by default
 $config['ckeditexpandtoolbar'] = true;
@@ -724,6 +739,5 @@ $config['pluginCoreList'] = [
 ];
 
 $config['pluginWhitelist'] = [];
-
 return $config;
 //settings deleted
--- a/application/controllers/UploaderController.php
+++ b/application/controllers/UploaderController.php
@@ -177,13 +177,13 @@ class UploaderController extends SurveyController
                 Yii::app()->end();
             }
             /* extension checked : check mimeType */
-            $extByMimeType = LSFileHelper::getExtensionByMimeType($_FILES['uploadfile']['tmp_name'], null,false);
+            $extByMimeType = LSFileHelper::getExtensionByMimeType($_FILES['uploadfile']['tmp_name'], null);
             $disableCheck = false;
             if(is_null($extByMimeType)) {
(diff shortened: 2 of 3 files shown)


LouisGac:
great.

Shnoulle:
Can someone check with the files in report on a debian server ?
I want a better file than .ico (Fedora mime type send image/vnd.microsoft.icon not in Yii fileinfo) and html send a PHP notice with /usr/share/misc/magic.mgc

Shnoulle:
Remind : `$magicFile` from getExtensionByMimeType != `$magicFile` from getMimeType

1. getMimeType : setenv('MAGIC=…') in config.php can fix the issue
2. getExtensionByMimeType : we can use CFileHelper since it was in param.

Then : i can fix it with
1. Adding App->config to CFileHelper::getExtensionByMimeType call
2. Document setenv('MAGIC=…') in manual.

But about 2 : setenv can be disable (?)

olleharstedt:
Is this a bug fix or a new feature or both?

Shnoulle:
Both :) 

New feature needed to fix an issue …

olleharstedt:
Question: Can this bug be fixed _without_ a new feature?

Shnoulle:
No, we need at minima `$config['magic_file'] = null;` 

For the aac file : `CFileHelper::getMimeType` return `audio/x-hx-aac-adts` (php7.3) but not included in Yii fileInfo file

olleharstedt:
> but not included in Yii fileInfo file

Yes, that's why I think it's an issue for the Yii team. :) To add the mime type in the list.

Shnoulle:
> Yes, that's why I think it's an issue for the Yii team. :) To add the mime type in the list.

Maybe in Yii2, but not in Yii1 since it's a App issue (App can use another one) :)

olleharstedt:
Why not in Yii 1? It's still maintained, no?

Shnoulle:
https://www.yiiframework.com/news/90/update-on-yii-1-1-support-and-end-of-life

> We have now decided to not accept any features or enhancements for Yii 1.1 anymore.

Know file extension seems more an enhancements since it can be done in App :) (IMO)

olleharstedt:
In that case it's without risk to change the file ourselves, no? Or better: Subclass it and add what we want.

Shnoulle:
> Subclass it and add what we want.

But since using a updatable file is more easy to fix quickly for user . Nesxt time maybe there are mvg or mxd file … with this fix : admin user can fix it himself without wait a new release :)

olleharstedt:
So add new feature in LS4 instead?

Shnoulle:
Its a `Bug reports` and seems more a bug. Fixing this bug need a feature :)

LouisGac:
Thank you Denis, you did a great job.

Shnoulle:
Let's wait @c-schmitz decision about new feature or not ;)

c-schmitz:
Well, mime type checking is an additional measure but not really secure and can easily be tricked. I will allow this though, because we have wandered down this path and need to properly finish it.

Shnoulle:
> Well, mime type checking is an additional measure but not really secure and can easily be tricked.

You know a php way to check mime type in a **really** secure way ?

LouisGac:
you can fake a meme type by editing the first bytes of a script. I think that's what Carsten means.

Shnoulle:
@LouisGac : yes i know, but in this case : i think the OS must disable execution etc … no ? And we take the security measure **we can**, we can not have security measure we don't know ;)

LouisGac:
sure, that's why we must check mime type and file extension. To be sure a .php is not validated as a picture and uploaded.  Your PR does it in a perfect way, you even pointed to the Yii function to do it and offer a way to update it since it's not updated anymore... so... again, great work :+1: 

https://github.com/LimeSurvey/LimeSurvey/blob/master/framework/utils/fileExtensions.php

PR #1322, opened 2019-09-26
Fixed issue #15334: in _deleteParticipant

WalAffe:
**admin/participants/sa/displayParticipants** delete modal has different values for second option if you click on the delete Button directly, or select User and use the delete Button in the Dropdown Menu. According to **application/controllers/admin/participantsaction.php** there are 3 cases: po, ptt, ptta. 

If you click on the delete button directly in the user row the second value is pt.


c-schmitz:
Thank you!

PR #1326, opened 2019-10-04
Feature/bug Unfinished code

KRav007:
Fixed issue # : 15370
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/application/controllers/admin/UserManagement.php
+++ b/application/controllers/admin/UserManagement.php
@@ -140,33 +140,7 @@ class UserManagement extends Survey_Common_Action
         }
 
         if (!isset($aUser['uid']) || $aUser['uid'] == null) {
-            $sendMail = (bool) Yii::app()->request->getPost('preset_password', false);
-            $aUser = $this->_createNewUser($aUser);
-            $sReturnMessage = '';
-            $success = true;
-
-            if ($sendMail && (isset($newUser['sendMail']) && $newUser['sendMail'] == true)) {
-                if ($this->_sendAdminMail('registration', $aUser)) {
-                    $sReturnMessage = gT("Success");
-                    $sReturnMessage .= CHtml::tag("p", array(), sprintf(gT("Username : %s - Email : %s."), $aUser['users_name'], $aUser['email']));
-                    $sReturnMessage .= CHtml::tag("p", array(), gT("An email with a generated password was sent to the user."));
-                } else {
-                    // has to be sent again or no other way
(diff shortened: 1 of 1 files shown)


application/controllers/admin/UserManagement.php
```diff
@@ -959,7 +933,12 @@ public function _sendAdminMail($type = 'registration', $aUser, $newPassword = nu
         $mailer->Body = $body;
         $mailer->isHtml(true);
         $mailer->emailType = $emailType;
-        return $mailer->sendMessage();
+
```
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/blob/409556bf2fc939f6b159c26f9d5f5a40cdde1328/application/core/LimeMailer.php#L540 already return a boolean, thencan return it ?

application/controllers/admin/UserManagement.php
```diff
@@ -959,7 +933,12 @@ public function _sendAdminMail($type = 'registration', $aUser, $newPassword = nu
         $mailer->Body = $body;
         $mailer->isHtml(true);
         $mailer->emailType = $emailType;
```
Shnoulle: It's set `registration` but `registration` it's not for survey register ?

KRav007: seems so, it's only for user admin part Add & reset password - could be refactored to a globale mailer class, which has all cases in switch , reset , register for survey, add admin user .......

What do you mean? This class is a nightmare - the CSV-Stuff shoaled exclude too?!

application/controllers/admin/UserManagement.php
KRav007: We should return the $oMailer to handle error case in calling method?

application/controllers/admin/UserManagement.php
Shnoulle: [This class is a nightmare](https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/core/LimeMailer.php) ? It can be improved, 

Else

https://github.com/LimeSurvey/LimeSurvey/blob/409556bf2fc939f6b159c26f9d5f5a40cdde1328/application/core/LimeMailer.php#L44-L49

This use for
1. Default template to be used (if exist in survey)
2. Event name (then event can know easily what email it send).

cdorin93:
Hmm, what is the status of this one? :thinking:

eddylackmann:
@cdorin93  this is not needed anymore , we worked and changed many things in the user management .

PR #1327, opened 2019-10-08
[WIP] Dev: fix travis for ranking

Shnoulle:
Dev: issue https://bugs.limesurvey.org/view.php?id=15385
Dev: is worst in develop than in master
Dev: manual checking, unsure it‘s OK ;)

### DIFF
--- a/application/views/survey/questions/answer/ranking/answer.twig
+++ b/application/views/survey/questions/answer/ranking/answer.twig
@@ -61,8 +61,8 @@
 {% set script %}
 try{
  var ranking{{qId}} = new RankingQuestion({
-        max_answers      : "{{ processString(max_answers,1) }}",
-        min_answers      : "{{ processString(min_answers,1) }}",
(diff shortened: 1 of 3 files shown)


Shnoulle:
Let's see if travis is happy

olleharstedt:
You can also hack `.travis.yml` to use `--filter Ranking` to only run the ranking test. Slightly faster feedback.

Shnoulle:
Yes, but don't want to setup extra part on my allday system …
Then i do it in a VM

olleharstedt:
I mean hacking the main .travis.yml.

Shnoulle:
@olleharstedt : there are 2 different fix … maybe revert and check only with https://github.com/LimeSurvey/LimeSurvey/pull/1327/commits/2313cfc645be3644894a5dcda7d32b234e513aa3 ?

Shnoulle:
Else : we don't have the same test in master and develop …

olleharstedt:
Yes, I don't want to merge this, since it changes to test.

PR #1330, opened 2019-10-09
added new action events to PluginCommand

mbischof:
Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev: 

`php application/commands/console.php plugin events`
could replace `application/libraries/PluginManager/EventDocumentor.php`

### DIFF
--- a/application/commands/PluginCommand.php
+++ b/application/commands/PluginCommand.php
@@ -1,48 +1,90 @@
 <?php
-    /*
-    * LimeSurvey (tm)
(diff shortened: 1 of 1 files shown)


Shnoulle:
What purpose ? 

Scanning a whole directory for php file , without "manage installed" event ? Seems strange , no ?

mbischof:
as I mentioned before, this could be a replacement for `application/libraries/PluginManager/EventDocumentor.php`, which outputs a list of all used plugin events in Limesurvey by a yii console command instead of a simple php script, imho the right place to do such things

PR #1332, opened 2019-10-16
14668 logo quexml export

mbischof:
Fixed issue # : 
New feature # :
Changed feature # :
Dev: 
Dev:


application/libraries/admin/quexmlpdf.php
```diff
@@ -3806,6 +3806,16 @@ protected function newPage($init = false)
             $this->SetHeaderMargin(0);
             $this->SetFooterMargin(0);
 
+            $oTemplate = Template::model()->getInstance();
```
olleharstedt: Could you please provide instructions on how to (manually) test this code? Something like

1. Go to x
2. Click on y
3. See z
4. ...

Thank you.

application/libraries/admin/quexmlpdf.php
```diff
@@ -3806,6 +3806,16 @@ protected function newPage($init = false)
...
+            $sLogoFileName = $oTemplate->filesPath . Yii::app()->getConfig('pdflogofile');
+            if (file_exists($sLogoFileName)) {
+                $result = LSYii_ImageValidator::validateImage($sLogoFileName);
+                if (isset($result['check']) && $result['check'] == true) {
+                    $sLogo = "@" . file_get_contents($sLogoFileName);
```
olleharstedt: You probably want to check if `file_get_contents` returned anything valid.

Shnoulle: Done in `$result = LSYii_ImageValidator::validateImage($sLogoFileName);` :)

application/libraries/admin/quexmlpdf.php
Shnoulle: Think it's hard to test pdf file ? Check if logo is shown in pdf file ?

Shnoulle:
Question : did logo must be same for queXML than from pdf ?

In my opinion : logo for pdf and queXML pdf can be different. Maybe adding a new global settings ?

application/libraries/admin/quexmlpdf.php
mbischof: In your upload/themes/survey/{...}/files - folder must be the image defined in your application/config/config-defaults.php `pdflogofile` setting, then pdf queXML export (responses/sa/browse, second icon in list view) should display the image in the pdfs header

Shnoulle: Yes : but @olleharstedt talk for a way to add a [test](https://github.com/LimeSurvey/LimeSurvey/tree/develop/tests) for this. I don't think there are automatic way to add a test to check PDF content.

Shnoulle: Arg … sorry … **to (manually) test ** sorry both … please : delete my comments ;)

application/libraries/admin/quexmlpdf.php
olleharstedt: No? It can return `false`, which should be checked.

maziminke:
@c-schmitz , can we please merge this one? Its details were discussed long time ago (see https://bugs.limesurvey.org/view.php?id=14668).
Thanks!

olleharstedt:
I'd suggest not to merge any new features without:

1) Manual QA
2) Accepting tests

Sorry for the bureaucracy, but we're trying to come up with a more stable workflow for new features, both internally and externally. Also quite busy making LS4 bug free, so I didn't have time yet to test this.

PR #1335, opened 2019-10-22
[READY] Working on "5" lt "20"

Shnoulle:
Fixed issue : more number VS string comparaison difference in js and php
Dev: add test for "5" le "18"

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -325,7 +325,7 @@ class ExpressionManager
         // Not sure if needed to test if [2] is set. : TODO review
         if ($bBothNumeric) {
             $aForceStringArray = array('DQ_STRING', 'DS_STRING', 'STRING'); // Question can return NUMBER or WORD : DQ and DS is string entered by user, STRING is a result of a String function
-            if ((isset($arg1[2]) && in_array($arg1[2], $aForceStringArray) || (isset($arg2[2]) && in_array($arg2[2], $aForceStringArray)))) {
+            if ((isset($arg1[2]) && in_array($arg1[2], $aForceStringArray) && (isset($arg2[2]) && in_array($arg2[2], $aForceStringArray)))) {
                 $bBothNumeric = false;
                 $bBothString = true;
                 $bMismatchType = false;
--- a/tests/data/surveys/limesurvey_survey_checkCompareTestSurvey.lss
+++ b/tests/data/surveys/limesurvey_survey_checkCompareTestSurvey.lss
@@ -17,34 +17,7 @@
   </fields>
   <rows>
    <row>
-    <qid><![CDATA[12400]]></qid>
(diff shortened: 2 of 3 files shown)


Shnoulle:
It's just the test to do here : with Q0 as "5" test done is 
https://github.com/LimeSurvey/LimeSurvey/pull/1335/files#diff-5a3f04a9357d7c5320bbb989015c7f62R102-R108

application/helpers/expressions/em_core_helper.php
```diff
@@ -325,7 +325,7 @@ public function RDP_EvaluateBinary(array $token)
         // Not sure if needed to test if [2] is set. : TODO review
         if ($bBothNumeric) {
             $aForceStringArray = array('DQ_STRING', 'DS_STRING', 'STRING'); // Question can return NUMBER or WORD : DQ and DS is string entered by user, STRING is a result of a String function
-            if ((isset($arg1[2]) && in_array($arg1[2], $aForceStringArray) || (isset($arg2[2]) && in_array($arg2[2], $aForceStringArray)))) {
+            if ((isset($arg1[2]) && in_array($arg1[2], $aForceStringArray) && (isset($arg2[2]) && in_array($arg2[2], $aForceStringArray)))) {
```
olleharstedt: For me, there's absolutely no way to know if this change breaks anything without a complete test suite. :|

olleharstedt:
I'll checkout this PR and have a look.

olleharstedt:
Shouldn't `+` be banned on strings too? Use strconcat or something.

application/helpers/expressions/em_core_helper.php
Shnoulle: The issue is (in 3.19) : JS and PHP does 2 different think.

See the 1st commit : https://github.com/LimeSurvey/LimeSurvey/pull/1335/commits/84e7ebe993b6f879795085a651689387152e0217
I test ONLY if js value == php value

Shnoulle:
About `+` maybe warning always … use `sum` or `join`

olleharstedt:
`+` on numbers is OK because it's _easy_ to have same behvaiour on JS and PHP. IMHO.

Shnoulle:
Yes, but 2 + "" + 2 => number or string ? ("" come from a no answer value)

It's not a JS vs PHP issue, it's desired behaviour issue

Shnoulle:
And 2 + QCODE_SQ01.NAOK + "2"  with QCODE_SQ01 filtered ? number or string

olleharstedt:
2 + "" is `+` used on string, so should be warning.

olleharstedt:
So `+` only on questions that are forced to numerical. They still have empty string as no answer?

olleharstedt:
Otherwise, to use `+` you must first `intval(NAOK)`.

Shnoulle:
@ollehar : now there are warnings : did i merge and test this one again ?

PR #1343, opened 2019-11-19
Add missing translation in tutorial question 29

m-unkel:
Fixed issue # : Tutorial question 29 is missing a gT(...) call for the text inside the alert box.
Dev:  The text "Please add at least two answer options to proceed." must be added to the .po translations. at translate.limesurvey.org

Add translation option for string "Please add at least two answer options to proceed.".
Wrapped string inside a gT(...)-call


c-schmitz:
Thank you!

PR #1344, opened 2019-11-20
New feature #15513: Missing function floatval() in EM

Shnoulle:
Dev: check 150.03_page-section php vs js : OK

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -184,6 +184,7 @@ class ExpressionManager
 'date' => array('exprmgr_date', 'date', gT('Format a local date/time'), 'string date(format [, timestamp=time()])', 'http://php.net/date', 1, 2),
 'exp' => array('exp', 'Decimal.asNum.exp', gT('Calculates the exponent of e'), 'number exp(number)', 'http://php.net/exp', 1),
 'fixnum' => array('exprmgr_fixnum', 'LEMfixnum', gT('Display numbers with comma as decimal separator, if needed'), 'string fixnum(number)', '', 1),
+'floatval' => array('floatval', 'LEMfloatval', gT('Get float value of a variable'), 'number floatval(number)', 'http://php.net/floatval', 1),
 'floor' => array('floor', 'Decimal.asNum.floor', gT('Round fractions down'), 'number floor(number)', 'http://php.net/floor', 1),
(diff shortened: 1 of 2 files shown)


olleharstedt:
Acceptance test needed. Could be done as unit-test, not functional, right? See my tiny test added to EM warning.

Shnoulle:
I already started survey test but … [15598: intval : broken interpretation of string start by integer in JS](https://bugs.limesurvey.org/view.php?id=15598)

Shnoulle:
My opinon : fix intval in 4.0 : add a test for intval + floatval in 4.X

olleharstedt:
My question is why LEMintval is *worse* than the JS standard `parseInt`?

Shnoulle:
Don't know …
Seems it was implemented like this : https://github.com/SondagesPro/LimeSurvey-SondagesPro/commit/237c43a5ca362d1d1591f01ae2ea202e7454b789

olleharstedt:
```
function LEMis_string(a)
{
    return isNaN(a);
}
```

???

olleharstedt:
Looks like Carsten moved the file without using `git mv`.

Shnoulle:
Apr 2012 : 8 years ago …

Previous : https://github.com/SondagesPro/LimeSurvey-SondagesPro/commit/41fe91a45ce37ce5a0a304fa012b8afe07acf401

olleharstedt:
What about replacing LEMintval with parseInt in JS?

olleharstedt:
Missing acceptance test, but maybe I can write that later.

Shnoulle:
@olleharstedt I'm out of office this week. I can write acceptance test for floatval only (i already start but intval broken).

olleharstedt:
Maybe we can use nodejs to run the expression both in PHP and JS? I did this for an earlier test.

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/blob/master/tests/surveys/MinMaxCompareTest.php already check JS vs PHP

It does

1. JS vs PHP compare
2. Fixed string compare (current way)

olleharstedt:
Yes, but it could be done without using the browser, I mean, in a unit test.

But good. :)

Shnoulle:
Maybe : but using the browser : you are sure it use expressionmanager js : no ?

olleharstedt:
I'm not sure about anything. :|

Shnoulle:
Right … «You have more chance it use the same js than a survey participant» …

PR #1347, opened 2019-11-25
Adds PostgreSQL compatibility

mwehr:
Fixed issue # :  Compare with "1" should be "true" for PostgreSQL

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -863,7 +863,7 @@ class Survey extends LSActiveRecord
         $criteria->addCondition('level=:level');
 
         if ($collapsed === true) {
-            $criteria->addCondition('showincollapse=1');
+            $criteria->addCondition('showincollapse=true');
         }
 
         $criteria->params = [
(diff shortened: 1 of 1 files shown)


c-schmitz:
This does not make sense. 'showincollapse' is an integer column, not a boolean.

mwehr:
Hmm, this is my exception with =1:
CDbCommand konnte das SQL-Statement nicht ausführen: SQLSTATE[42883]: Undefined function: 7 FEHLER: Operator existiert nicht: boolean = integer
LINE 1: ...t_id IS NULL AND (position='collapsed' OR showincollapse=1 )
^
HINT: Kein Operator stimmt mit dem angegebenen Namen und den Argumenttypen überein. Sie müssen möglicherweise ausdrückliche Typumwandlungen hinzufügen.. The SQL statement executed was: SELECT * FROM "lime_surveymenu" "t" WHERE survey_id IS NULL AND parent_id IS NULL AND (position=:position OR showincollapse=1 )

mwehr:
Hmm ok I may have found the problem.
Our Limesurvey didn't go through the "boolean" -> "integer" migration
https://github.com/LimeSurvey/LimeSurvey/commit/0537858d80bc92b12eba643eb785bf87b17ebfaa

PR #1352, opened 2019-12-17
Fixed issue #15662: Non utf8 caracter save as ????? in DB

Shnoulle:
Dev: same fix than 4.X …


Shnoulle:
review for a better commit

PR #1354, opened 2019-12-20
[WIP] Dev: Issue with MSSQL and 0 or 0.1 or 0.001

Shnoulle:
Fixed issue #15684: When reloading decimal value with 0 with MSSQL : 0 disappear
Fixed issue #15685: Issue when exporting decimal value in MSSQL 
Dev: check if value start by `. ` before removing lastb 0 and last dot
Dev: add 0 is start by .
Dev: Maybe can check with regexp, but use $string[0] seems the speediest solution ?


olleharstedt:
I'd like to test this a bit more before merge.

Shnoulle:
@olleharstedt ask for a mysql test : reload a LSA with token and answer persistence with 0 and 0.1 entered seems.

PR #1358, opened 2020-01-03
Dev: script and xss 

Shnoulle:
Fixed issue #15690: User with XSS enable can add/update scripts
New feature #15096: XSS for super-admin too
New feature : Allow configure question script allowed

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -229,9 +229,9 @@ $config['filterxsshtml'] = true;
 
 // disablescriptwithxss
 // Allow update of script in question
-// 1 : Default : follow XSS rules
-// 0 : allowed for all
-$config['disablescriptwithxss'] = 1;
+// true : Default : follow XSS rules
+// false : allowed for all
+$config['disablescriptwithxss'] = true;
 
 // superadminfilterxsshtml
 // Enables filtering of suspicious html tags in survey, group, questions
@@ -240,10 +240,11 @@ $config['disablescriptwithxss'] = 1;
 // This settings can not be updated by GUI interface
 $config['superadminfilterxsshtml'] = false;
 
-// superadmindisablescript
+// superadminenablescript
 // Totally disable question script update
 // Remove the script part in question edition
-$config['superadmindisablescript'] = false;
+// Used only if superadminfilterxsshtml and disablescriptwithxss are true
+$config['superadminenablescript'] = false;
 
 // usercontrolSameGroupPolicy
 // If this option is set to true, then limesurvey operators will only 'see'
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -233,17 +233,17 @@ $config['filterxsshtml'] = true;
 // false : allowed for all
 $config['disablescriptwithxss'] = true;
 
-// superadminfilterxsshtml
+// forcedfilterxss
 // Enables filtering of suspicious html tags in survey, group, questions
 // and answer texts in the administration interface
 // Even super-administrator have this restriction.
 // This settings can not be updated by GUI interface
-$config['superadminfilterxsshtml'] = false;
+$config['forcedfilterxss'] = false;
 
 // superadminenablescript
 // Totally disable question script update
 // Remove the script part in question edition
-// Used only if superadminfilterxsshtml and disablescriptwithxss are true
+// Used only if forcedfilterxss and disablescriptwithxss are true
 $config['superadminenablescript'] = false;
 
 // usercontrolSameGroupPolicy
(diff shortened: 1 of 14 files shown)


Shnoulle:
Maybe need a last setting : 'noquestionscript' where even OLD scripts are removed. No GUI

olleharstedt:
Too many indentation fixes in this PR...

Shnoulle:
The identitentation fix was here : https://github.com/LimeSurvey/LimeSurvey/pull/1358/commits/af5a74094f260e1b6fdac21a9d6c7cd10e815fbe

You can not fix a file if it's totally broken …

The file just have 2 new functions : https://github.com/LimeSurvey/LimeSurvey/pull/1358/files#diff-707cce6f9455b8b2bf2497ceced9a0c7R119-L116

olleharstedt:
New features should have acceptance tests. Is there a good reason to exclude them here?

Shnoulle:
Really hard to do a acceptance test with a `update only via config.php file`

Shnoulle:
And if i remind : acceptance test only for LimeSurvey internal dev …

olleharstedt:
Acceptance tests should be done for all new features, both internal and external devs.

olleharstedt:
Can you create a mantis ticket for this feature, please? 

>  New feature : Allow configure question script allowed

olleharstedt:
The emcache test writes to config.php: https://github.com/LimeSurvey/LimeSurvey/blob/develop/tests/functional/frontend/EmCacheSurveyTest.php#L38

Maybe there's a better way?

Shnoulle:
OK for the config , but i my opinion : need a XSS checker for simple user BEFORE superadmin

Shnoulle:
> Can you create a mantis ticket for this feature, please?
> New feature : Allow configure question script allowed

The new feature didn't exist currently because any admin can add scrip (it's the issue). Then …

Shnoulle:
https://bugs.limesurvey.org/view.php?id=15693

application/config/config-defaults.php
```diff
@@ -227,6 +227,25 @@
...
+// and answer texts in the administration interface
+// Even super-administrator have this restriction.
+// This settings can not be updated by GUI interface
+$config['forcedfilterxss'] = false;
+
+// superadminenablescript
+// Totally disable question script update
+// Remove the script part in question edition
+// Used only if forcedfilterxss and disablescriptwithxss are true
+$config['superadminenablescript'] = false;
```
lacrioque: I think this should be a global option.
The script field may be disabled by a personal setting, it is missing a global setting, yet.

lacrioque:
I do like this PR very much.
It takes into account missing functionalities that need to be implemented for the script field to be useful also for potentially more paranoid organisations.

application/config/config-defaults.php
Shnoulle: This settings can disable superadmin to have script … by security : disallow ALL user to add scripts.

If you enable `forcedfilterxss` : superadmin can not add script inside question->text (or anywhere) => Must not be in GUI.
But if you enable at same time `disablescriptwithxss` : superadmin can not add (or update) script. We must leave allowing this ONLY in php file : Some server admin can want to disable script for superadmin too (see some bug report about XSS in mantis : XSS by superadmin).

Shnoulle:
> It takes into account missing functionalities that need to be implemented for the script field to be useful also for potentially more paranoid organisations.

See https://github.com/LimeSurvey/LimeSurvey/pull/1358#issuecomment-570585206

> Maybe need a last setting : 'noquestionscript' where even OLD scripts are removed. No GUI

But it's another feature (and can be done separatly)

application/config/config-defaults.php
olleharstedt: This line:

>  // Totally disable question script update

does not make sense with regard to the option. Maybe change to:

>  // If false, will totally disable question script update

and of course also include a description if set to true.

application/models/QuestionL10n.php
```diff
@@ -27,6 +27,15 @@
 class QuestionL10n extends LSActiveRecord
 {
 
+    /**
+     * @inheritdoc
+     * Add a handler to before save for script
+     */
+    public function init()
```
olleharstedt: Is this relevant for the XSS fix?

olleharstedt:
How come you didn't edit this line? https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/core/LSYii_Validators.php#L52

Shnoulle:
> How come you didn't edit this line? https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/core/LSYii_Validators.php#L52

https://github.com/LimeSurvey/LimeSurvey/pull/1358/files#diff-5fddc6ba03308f8df045a9051b434d06R48 ? 

Checked by [LSWebUser->xssFiltered() function](https://github.com/LimeSurvey/LimeSurvey/pull/1358/files#diff-707cce6f9455b8b2bf2497ceced9a0c7R143)

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1364

PR #1355, opened 2020-01-02
Dev: Optimisation on SQL request in Expression Manager

Shnoulle:
Move for ->find ->save to updateByPk 

With find+save : all value are updated : update whole line. With updateByPk : update ONLY the needed value.

With find+save : you update value from 1st page at 2nd page, with updateByPk : only value on page + lastpage + datestamp are updated : less SQL is **always** better. See the SQL log

PR #1359, opened 2020-01-07
Dev: 2 different action for save : plugin event happen 2 times

Shnoulle:
Dev: this is just a test for https://github.com/LimeSurvey/LimeSurvey/commit/27957d68e9ae3226515524c1774493d86706233a#commitcomment-36629663

**But** before https://github.com/LimeSurvey/LimeSurvey/commit/27957d68e9ae3226515524c1774493d86706233a#commitcomment-36629663 beforeResponseSave didn't happen. Then : i think best solution is to use updateByPK : really optimum SQL. 
And add a new event in 4.X in a updateByPk for Response .


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1355#event-2930577306 merged

PR #1360, opened 2020-01-08
Fixed issue #15684: When reloading decimal value with 0 with MSSQL : 0 disappear

Shnoulle:
Fixed issue #15685: Issue when exporting decimal value in MSSQL
Dev: check if value start by . before removing last 0 and last dot
Dev: add 0 is start by .
Dev: Maybe can check with regexp, but use $string[0] seems the speediest solution ?
Dev: doing it for Survey doing , export, response browse and data edit
Dev: don't update ExpressionManager

### DIFF
--- a/application/controllers/admin/dataentry.php
+++ b/application/controllers/admin/dataentry.php
@@ -598,6 +598,9 @@ class dataentry extends Survey_Common_Action
                             $aDataentryoutput .= $fname['subquestion'].'&nbsp;';
                             /* Fix DB DECIMAL type */
                             $value = $idrow[$fname['fieldname']];
+                            if($value[0] === ".") {
+                                $value = "0".$value;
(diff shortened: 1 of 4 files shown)


Shnoulle:
In  previous commit : error in ExpressionManager must be 

````
            if($arg1[0][0]  === ".") {
                $arg1[0]  = "0".$arg1;
            }
            if (strpos($arg1[0] , ".") !== false) {
                $arg1[0]  = rtrim(rtrim($arg1, "0"), ".");
            }
````

But since : i'm unsure we need it in Expression manager : i didn't update EM

olleharstedt:
OK, will test.

Shnoulle:
I just add the “fix” where i'm sure it's needed :)

Thanks a lot

PR #1361, opened 2020-01-09
15632 remotecontrol cpd import participants update

mbischof:
Fixed issue # : https://bugs.limesurvey.org/view.php?id=15632
New feature # : 
Changed feature # : cpd_importParticipants allows updating records
Dev: 
Dev:

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -3112,107 +3112,105 @@ class remotecontrol_handle
      * @param int $sSessionKey
      * @param array $aParticipants
      * [[0] => ["email"=>"dummy-02222@limesurvey.com","firstname"=>"max","lastname"=>"mustermann"]]
+     * @param bool $update
      * @return array with status
      */
-    public function cpd_importParticipants($sSessionKey, $aParticipants)
+    public function cpd_importParticipants($sSessionKey, $participants, $update = false)
     {
-
         if (!$this->_checkSessionKey($sSessionKey)) {
-            return array('status' => 'Invalid session key');
+            return 'Invalid session key';
         }
 
-        $aAttributeData = array();
         $aDefaultFields = array('participant_id', 'firstname', 'lastname', 'email', 'language', 'blacklisted');
-        $bIsValidEmail = true;
-        $sMandatory = 0;
-        $sAttribCount = 0;
-        $aResponse = array();
-        $aResponse['ImportCount'] = 0;
+        $aResponse = array(
+            'ImportCount' => 0,
+            'UpdateCount' => 0
+        );
 
-        // get all attributes for mapping
-        $oFindCriteria = new CDbCriteria();
-        $oFindCriteria->offset = -1;
-        $oFindCriteria->limit = -1;
-        $aAttributeRecords = ParticipantAttributeName::model()->with('participant_attribute_names_lang')->findAll($oFindCriteria);
+        $aAttributeRecords = ParticipantAttributeName::model()
+            ->with('participant_attribute_names_lang')
+            ->findAll();
 
-        foreach ($aParticipants as $sKey => $aParticipantData) {
+        foreach ($participants as $participant) {
 
-            $aData = array(
-                'firstname' => $aParticipantData['firstname'],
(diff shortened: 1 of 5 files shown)


application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -3115,107 +3115,105 @@ protected function _checkSessionKey($sSessionKey)
...
      * [[0] => ["email"=>"dummy-02222@limesurvey.com","firstname"=>"max","lastname"=>"mustermann"]]
+     * @param bool $update
      * @return array with status
      */
-    public function cpd_importParticipants($sSessionKey, $aParticipants)
+    public function cpd_importParticipants($sSessionKey, $participants, $update = false)
     {
-
         if (!$this->_checkSessionKey($sSessionKey)) {
-            return array('status' => 'Invalid session key');
```
Shnoulle: No, all error return an arry with status. Don't update this part

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -3115,107 +3115,105 @@ protected function _checkSessionKey($sSessionKey)
...
+                    $model->participant_id = Participant::gen_uuid();
                 }
+            }
 
-                // Write to database if record not exists
-                if (empty($arRecordExists)) {
-                    // save participant to database
-                    Participant::model()->insertParticipantCSV($aData);
+            $scenario = $model->getScenario(); // insert or update
+            if ($scenario == 'update' && $update === false) {
```
Shnoulle: Strange to use getScenario :)

Why not , maybe more simple to understand 

````
if (!$model) {
    // Create
} elseif (!$update) {
    continue;
}
````

**Comment by olleharstedt** [comment]
Looks like lots of tests fail for some reason. I will run them locally and check.

**Comment by olleharstedt** [comment]
The test fails because there are too many failed login attempts.

**Comment by olleharstedt** [comment]
Nope, wrong. The test fails because the default user is overwritten from "admin" to "remotecontroladmin".

**Comment by mbischof** [comment]
this password? https://github.com/LimeSurvey/LimeSurvey/blob/master/tests/helpers/RemoteControlTest.php#L33

**Comment by Shnoulle** [comment]
Scrutinizer is more happy now :). 

Else : can you review the website and the email ?

File: tests/helpers/remotecontrol/BaseTest.php [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+    /* @var remotecontrol_handle*/
+    public $handler;
+
+    const REMOTECONTROL_USERNAME = 'admin';
+    const REMOTECONTROL_PASSWORD = 'password';
+    const REMOTECONTROL_EMAIL    = 'admin@test.de';
+
+    public static function setUpBeforeClass()
+    {
+        Yii::import('application.helpers.common_helper', true);
```
olleharstedt: These lines are duplicated - they are already here: https://github.com/LimeSurvey/LimeSurvey/blob/master/tests/TestHelper.php#L23

olleharstedt:
I'm guessing you want this merged into new branch 3.x-LTS, not master (which is now LS4)?

mbischof:
thats correct

maziminke:
We also need this fix to be available at Limesurvey 4 later.

olleharstedt:
> We also need this fix to be available at Limesurvey 4 later.

No problem.

olleharstedt:
> thats correct

:+1: 

Please note that I won't merge until the issues are resolved (code duplication, in this case).

Shnoulle:
https://tools.ietf.org/html/rfc2606 or limesurvey.org for example . limesurvey.example can be good too :)

olleharstedt:
Can you adjust the comments by Denis, please?

application/helpers/remotecontrol/remotecontrol_handle.php
Shnoulle: Still think i prefer the elseif method. What it's your opinion @olleharstedt ?

olleharstedt: No strong opinion. So will merge.

olleharstedt:
FYI, this PR has been cherry-picked into the master branch.

PR #1363, opened 2020-01-09
Fix #14108 - Registration of LDAP users with an identifier in lowercase

pjenicot:
No correct error message when the connection between the AuthLDAP plugin and LimeSurvey API fails due to the case sensitivity on the connection identifier

Fixed issue # : 14108
New feature # :
Changed feature # :
Dev: 
Dev:

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -426,7 +426,7 @@ class AuthLDAP extends LimeSurvey\PluginManager\AuthPluginBase
         /* unsubscribe from beforeHasPermission, else updating event */
         $this->unsubscribe('beforeHasPermission');
         // Here we do the actual authentication
-        $username = $this->getUsername();
+        $username = strtolower($this->getUsername());
         $password = $this->getPassword();
 
         $ldapmode = $this->get('ldapmode');


olleharstedt:
Is this enough to resolve the bug? Can you explain how to test in detail, in a step-by-step manner, please?

halliballi123:
In my opinion this won't solve the problem. I think the search for the user object should be case insensitve. this fix won't work if the username in limesurvey database has uppercase letters, or will it?

pjenicot:
Searching in LDAP is insensitive by default.
You are obliged to execute an SQL query on the database to change the identifiers to lowercase.

halliballi123:
I know that. But if you expect that to work, there are more things to consider. 

First in the update of LimeSurvey there should be a SQL update statement that changes the usernames to lowercase (not everybody who has that problem now will find this discussion here). 

Second LimeSurvey backend should force lowercase usernames. Otherwise there is always the problem of creating mixed case or uppercase usernames in LimeSurvey backend.

Or the lookup of the user object has to be case insensitve, as I suggested before.

Shnoulle:
> First in the update of LimeSurvey there should be a SQL update statement that changes the usernames to lowercase (not everybody who has that problem now will find this discussion here).

No, because case insensitive is only needed for LDAP

My opinion for a solution : update API https://github.com/LimeSurvey/LimeSurvey/blob/d45979c431ee14adffc66874d0a1d620f2331648/application/libraries/PluginManager/LimesurveyApi.php#L319 
To add a option 

````
    /**
     * Get the user object for a given username
     *
     * @param string $username
     * @param boolean $casesensitive search (default true)
     * @return User|null Returns the user, or null when not found
     */

````

halliballi123:
> 
> 
> > First in the update of LimeSurvey there should be a SQL update statement that changes the usernames to lowercase (not everybody who has that problem now will find this discussion here).
> 
> No, because case insensitive is only needed for LDAP
> 
> My opinion for a solution : update API
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/d45979c431ee14adffc66874d0a1d620f2331648/application/libraries/PluginManager/LimesurveyApi.php#L319
> 
> To add a option
> 
> ```
>     /**
>      * Get the user object for a given username
>      *
>      * @param string $username
>      * @param boolean $casesensitive search (default true)
>      * @return User|null Returns the user, or null when not found
>      */
> ```

Isn't that what I wrote before, change the behavior of the user object lookup?

LDAP user verification is already case insensitve. The problem is, that after successful authentication via LDAP the userobject could not be found if the entered Username was not exactly like it was written in LimeSurvey.

Therefore I think you are right, best way would be to change the getUserByName method and use it at the position after successful LDAP authentication.

Shnoulle:
Remind : system must be mysql + pgsql + mssql compatible :).

Then must use : https://www.yiiframework.com/doc/api/1.1/CDbExpression :) witth LOWER and set usernema to lowercase (or upper ;) )

cdorin93:
Hmm, what is the status of this one :thinking: ?

c-schmitz:
The PR itself was not enough and there does not seem to be any activity. Please submit a proper pull request implementing it as suggested in https://github.com/LimeSurvey/LimeSurvey/pull/1363#issuecomment-573561664. Alternatively submit a bug report or feature request.

halliballi123:
I think this discussion is a result of my bug report
[https://bugs.limesurvey.org/view.php?id=14108](https://bugs.limesurvey.org/view.php?id=14108)

PR #1364, opened 2020-01-09
Fixed issue #15690: User with XSS enable can add/update scripts

Shnoulle:
Dev: Add 2 function : one for XSS and one for script in LSWebUser
Dev: The 1st commit is the needed action, 2nd commit : just the view

### DIFF
--- a/application/controllers/admin/questionedit.php
+++ b/application/controllers/admin/questionedit.php
@@ -454,8 +454,9 @@ class questionedit extends Survey_Common_Action
             "update" => Permission::model()->hasSurveyPermission($oQuestion->sid, 'survey', 'update'),
             "editorpreset" => App()->session['htmleditormode'],
             "script" =>
-            SettingsUser::getUserSetting('showScriptEdit', App()->user->id)
-            && Permission::model()->hasSurveyPermission($oQuestion->sid, 'survey', 'update'),
+            Permission::model()->hasSurveyPermission($oQuestion->sid, 'survey', 'update')
+            && App()->user->isScriptUpdateAllowed()
+            && SettingsUser::getUserSetting('showScriptEdit', App()->user->id),
         ];
 
         $this->renderJSON($aPermissions);
--- a/application/core/LSWebUser.php
+++ b/application/core/LSWebUser.php
@@ -50,7 +50,7 @@ class LSWebUser extends CWebUser
     }
 
     /**
-     * Replace default system to return only one flash … 
+     * Replace default system to return only one flash …
      */
     public function getFlashes($delete = true)
     {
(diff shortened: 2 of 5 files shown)


Shnoulle:
I got it in rules \o/

olleharstedt:
Well, now I can't test it, because I can't get the damn script text field to show.

1. New user
2. New survey
3. My account -> show script field: yes
4. Edit question
5. Still no script field

What am I missing?

Shnoulle:
> What am I missing?

XSS disable for simple user ?
And maybe disable this option when XSS is enable for user ?

PS : checking with superadmin : i try to hide scripts : still shown …

Shnoulle:
PS : in you fork : script still show to user as editable even if they don't have the right.

If you want same : 1st commit do it. 2n,d commit only disable the script edit part.

olleharstedt:
image

image

olleharstedt:
Oh, you hide the field now...

Shnoulle:
I hide the edit field if user can not edit.

But let me check

Shnoulle:
It's OK with simple user, XSS disable : script are shown and editable

olleharstedt:
Well, I can't test the validator unless the field is visible. :) I'm hacking it to test.

olleharstedt:
OK, works, but I'd prefer if a message was shown to the user. "Not allowed to edit script" instead of just hiding it, because it's set shown = yes in the profile.

Shnoulle:
…  
It's must be in User settings …

And … really … asking more and more and more form external … But [LS team can do anything](https://github.com/LimeSurvey/LimeSurvey/commit/1eee6cf2833261dcd020bac974b154f34750fa27) ?

Shnoulle:
Your fork : https://github.com/LimeSurvey/LimeSurvey/tree/bug/15690-user-with-xss-filter-enabled-can-add-script

Use see script, can think it can be updated, try to update and … nothings happen …

Shnoulle:
You prefer i revert https://github.com/LimeSurvey/LimeSurvey/pull/1364/commits/1924ea89b950d49f64b69c42de745c43e8a7e282 ?

Then user see script but when save : script is resetted (like XSS system)

olleharstedt:
Urgh. I'll merge this and then revert 1924ea8 manually. Let's discuss later about marking script field read-only, we can do it in the office.

Shnoulle:
@olleharstedt : wait for "text information" my 2 other pull request

1. Global GUI : scriptAllowUpdate follow XSS : yes/no
2. PHP : xss for superadmin

Shnoulle:
And i do a 

3. Add information about all of this in UserSettings and Question edit/preview when all it done

olleharstedt:
Issue here: https://bugs.limesurvey.org/view.php?id=15702

PR #1366, opened 2020-01-13
New feature #15693: Allow configure question script allowed

Shnoulle:
Dev: add config disablescriptwithxss
Dev: Add information on Personal setting : don't disable setting,
Dev: but help show it as unused
Dev: col-md-6 : need row to be clean on md


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1372

PR #1368, opened 2020-01-14
Fixed issue #15696: Multiple email addresses in token are not working

Shnoulle:
Fixed issue : Allow to update bad Opt out or set Opt out manually


Shnoulle:
I don't use email : i report another issue …

Shnoulle:
Why i didn't use email input : [15712: When doing error in token create : unable to submit, need to reload page](https://bugs.limesurvey.org/view.php?id=15712)

else : email multiple allow , and ; but not |. Model control allow  , and ; and | . Then it's an API broker.

olleharstedt:
Thanks, will test and merge.

olleharstedt:
Test failed.

olleharstedt:
Merged but need more info.

PR #1369, opened 2020-01-16
Dev : fix potential issue when error trying to submit

Shnoulle:
Dev: previous way use $model->save, but new one ->updateByPk : no errors returned
Dev: master use $model->save : no need to merge

**3LTS only fix : do not merge in master**

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -5582,9 +5582,9 @@
                     if ($oResponse->submitdate == null || Survey::model()->findByPk($this->sid)->alloweditaftercompletion == 'Y') {
                         $iCountUpdated = Response::model($this->sid)->updateByPk($oResponse->id,$aResponseAttributes);
                         if (!$iCountUpdated) {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Very rare issue …

1. Need a submitfailed (this is hard to get now)
2. we use updateByPk to have less difference with previous system : update only needed part
3. But we try to get $response->errors : response can be null OR don't have errors

Then : this fix is clearly a fix without potential issue, unsure on the way to fix such issues ?

olleharstedt:
NB: Since 3.x is in another branch now, it will never be merged into master.

olleharstedt:
In case we have to fix something in both lts and master, a cherry-pick would be appropriate.

Shnoulle:
> In case we have to fix something in both lts and master, a cherry-pick would be appropriate.

cherry-pick master to 3LTS or 3LTS to master ?

I mean : if there are an issue to fix in 3LTS and master : we fix it in 3LTS 1st or second ?

Shnoulle:
Sorry :; lost a part on the fix : https://github.com/LimeSurvey/LimeSurvey/commit/a4d2c73bdd7909cc817ce6f8a9d5b128f5958116

Such fix is OK directly or not ?

PR #1371, opened 2020-01-17
Fixed issue #15733: Survey administrator can receive invalid mail about error

Shnoulle:
Dev: PDO system ?
Dev: updateByPk return updated answer, if answer are not updated : return 0

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -5582,8 +5582,14 @@
                     if ($oResponse->submitdate == null || Survey::model()->findByPk($this->sid)->alloweditaftercompletion == 'Y') {
                         $iCountUpdated = Response::model($this->sid)->updateByPk($oResponse->id,$aResponseAttributes);
                         if (!$iCountUpdated) {
-                            $message = submitfailed('',$this->gT('Error on response update')); // NO response->errors when use ->updateByPk
-                            LimeExpressionManager::addFrontendFlashMessage('error', $message, $this->sid);
+                            // Fix #15733 comparing values of previous response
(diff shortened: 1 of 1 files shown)


application/helpers/expressions/em_manager_helper.php
```diff
@@ -5582,8 +5582,13 @@ private function _UpdateValuesInDatabase($finished=false)
                     if ($oResponse->submitdate == null || Survey::model()->findByPk($this->sid)->alloweditaftercompletion == 'Y') {
                         $iCountUpdated = Response::model($this->sid)->updateByPk($oResponse->id,$aResponseAttributes);
                         if (!$iCountUpdated) {
-                            $message = submitfailed('',$this->gT('Error on response update')); // NO response->errors when use ->updateByPk
-                            LimeExpressionManager::addFrontendFlashMessage('error', $message, $this->sid);
+                            // Fix #15733 comparing values of previous response
+                            $aReponseCurrentAttributes = $oResponse->getAttributes();
+                            $aIntersect = array_intersect_key($aReponseCurrentAttributes,$aResponseAttributes);
```
olleharstedt: Won't this check only key difference, not value difference?

olleharstedt:
Will test

Shnoulle:
@olleharstedt : seems more (maybe) a workaround ?

I have when reload difference with page : `1 !== '1'` but `& == '1'`

Maybe : try to update ONLY if needed ? I mean : check if there are difference before try to update ?
What is the speediest : light SQL + PHP check or PHP check + light SQL ?

olleharstedt:
Denis, *please* write more detailed instructions on how to reproduce. I can't follow your instructions in the bug ticket. https://bugs.limesurvey.org/view.php?id=15733

PR #1372, opened 2020-01-17
New feature #15693: Allow simple user to update script with XSS enable

Shnoulle:
Dev: add settings default and GUI
Dev: use it


Shnoulle:
Thank you @lacrioque : now the last one : [15096: XSS for super-admin too](https://bugs.limesurvey.org/view.php?id=15096) (with the scriptXSS allowed or not)

PR #1375, opened 2020-01-24
Fixed issue #15769: Password can not be changed from the user settings page

aharenaz:
bug report #15769

The password it's never updated on user model so it's never saved on database making it impossible to change an users password.

When $newPassword is non empty, it never enters the else part so it doesn't updates the model.

Steps to reproduce this bug:
1- Make a new clean installation (i.e. using limesurvey4.0.1+200120.tar.gz, PHP 7, mariadb).
2- Login as admin, using "password" as password
3- Go to user account, and try to change the pasword to another one
-> The feedback the page gives is that changes were successfully applied
-> Password isn't really updated

### DIFF
--- a/application/controllers/admin/useraction.php
+++ b/application/controllers/admin/useraction.php
@@ -643,28 +643,31 @@ class UserAction extends Survey_Common_Action
                     $error = $oUserModel->checkPasswordStrength($newPassword);
                     if ($error) {
                         Yii::app()->setFlashMessage(gT($error), 'error');
-                        $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                        return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
                     }
-                } elseif (!$oUserModel->checkPassword($oldPassword)) {
+                }
+                if (!$oUserModel->checkPassword($oldPassword)) {
                     // Always check password
                     Yii::app()->setFlashMessage(gT("Your new password was not saved because the old password was wrong."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
-                } elseif (trim($oldPassword) === trim($newPassword)) {
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                }
+                if (trim($oldPassword) === trim($newPassword)) {
                     //First test if old and new password are identical => no need to save it (or ?)
                     Yii::app()->setFlashMessage(gT("Your new password was not saved because it matches the old password."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
-                } elseif (trim($newPassword) !== trim($repeatPassword)) {
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                }
+                if (trim($newPassword) !== trim($repeatPassword)) {
                     //Then test the new password and the repeat password for identity
                     Yii::app()->setFlashMessage(gT("Your new password was not saved because the passwords did not match."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
                 //Now check if the old password matches the old password saved
-                } elseif (empty(trim($newPassword))) {
-                    Yii::app()->setFlashMessage(gT("The password can't be empty."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
-                } else {
-                    // We can update
-                    $oUserModel->setPassword($newPassword);
                 }
+                if (empty(trim($newPassword))) {
+                    Yii::app()->setFlashMessage(gT("The password can't be empty."), 'error');
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                }
+                // We can update
+                $oUserModel->setPassword($newPassword);
             }
 
             $uresult = $oUserModel->save();


application/controllers/admin/useraction.php
```diff
@@ -643,28 +643,31 @@ public function personalsettings()
...
                 //Now check if the old password matches the old password saved
-                } elseif (empty(trim($newPassword))) {
+                }
+                if (empty(trim($newPassword))) {
                     Yii::app()->setFlashMessage(gT("The password can't be empty."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
-                } else {
-                    // We can update
-                    $oUserModel->setPassword($newPassword);
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
```
Shnoulle: redirect is a redirect, no return please

application/controllers/admin/useraction.php
```diff
@@ -643,28 +643,31 @@ public function personalsettings()
...
                     Yii::app()->setFlashMessage(gT("Your new password was not saved because the passwords did not match."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
                 //Now check if the old password matches the old password saved
-                } elseif (empty(trim($newPassword))) {
+                }
+                if (empty(trim($newPassword))) {
                     Yii::app()->setFlashMessage(gT("The password can't be empty."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
-                } else {
```
Shnoulle: In your method : seem this else must be remove

Shnoulle:
A mantis issue please ? With a how to reproduce ?

application/controllers/admin/useraction.php
aharenaz: It's already removed in my code ... your quote is from the original one

application/controllers/admin/useraction.php
aharenaz: Here is needed so action returns immediately showing the appropiate flash message (https://www.yiiframework.com/doc/api/2.0/yii-web-controller#redirect()-detail).
Without a return, it would show "Your personal settings were successfully saved." leading to confusion.
Otherwise it wold require more changes on then follow up code to check if errors happened on password change for showing the appropiate flash message.
Any suggestion at this respect?

application/controllers/admin/useraction.php
Shnoulle: Its' just looking at code, sicne i didn't have a clean way to reproduce …

Shnoulle: It's the one in this pull request …

application/controllers/admin/useraction.php
Shnoulle: I mean : 

Since `$this->getController()->redirect(array("admin/user/sa/personalsettings"));` is a redirect : you CAN'T return it …

It's an end : no return 

See https://www.yiiframework.com/doc/api/1.1/CHttpRequest#redirect-detail

return void …

application/controllers/admin/useraction.php
Shnoulle: Right :)

Shnoulle:
My opinion : 

If password was not updated : add a error / alert, but don't redirect before saving other data.

Then when user save (with a bad password)
1. Add a alert/error with [error setFlashMessage](https://github.com/LimeSurvey/LimeSurvey/blob/9500fe6b2f4ab6955e221a51a72a18341636cf8a/application/core/LSYii_Application.php#L250)
2. tag it must come back to user edit.
3. Save other data, and add [success setFlashMessage](https://github.com/LimeSurvey/LimeSurvey/blob/9500fe6b2f4ab6955e221a51a72a18341636cf8a/application/core/LSYii_Application.php#L250)
4. If must redirect : [check it here](https://github.com/LimeSurvey/LimeSurvey/blob/9500fe6b2f4ab6955e221a51a72a18341636cf8a/application/controllers/admin/useraction.php#L701) even if it's saveandclose 

Because (for example) if user want to upadte is name at same time than his password : you must save is name …

Maybe : when save with success : put a different string https://github.com/LimeSurvey/LimeSurvey/blob/9500fe6b2f4ab6955e221a51a72a18341636cf8a/application/controllers/admin/useraction.php#L695

> “Password excepted, your personal settings were successfully saved.”

Shnoulle:
PS : https://github.com/LimeSurvey/LimeSurvey/blob/9500fe6b2f4ab6955e221a51a72a18341636cf8a/application/core/LSYii_Application.php#L250

Allow multiple message : then can have error + warning + success + success + … …

Just add message :)

olleharstedt:
As Denis said, please add a bug report in bugs.limesurvey.org with precise step-by-step instructions on how to reproduce.

olleharstedt:
Oh, yeah, just copy your "Steps to reproduce this bug:" to the ticket.

application/controllers/admin/useraction.php
aharenaz: OK, thanks, you are right!
I'll change that.

aharenaz:
> Oh, yeah, just copy your "Steps to reproduce this bug:" to the ticket.

application/controllers/admin/useraction.php
```diff
@@ -643,28 +643,31 @@ public function personalsettings()
                     $error = $oUserModel->checkPasswordStrength($newPassword);
                     if ($error) {
                         Yii::app()->setFlashMessage(gT($error), 'error');
-                        $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                        return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
```
olleharstedt: Remove `return`, please.

application/controllers/admin/useraction.php
```diff
@@ -643,28 +643,31 @@ public function personalsettings()
...
                     }
-                } elseif (!$oUserModel->checkPassword($oldPassword)) {
+                }
+                if (!$oUserModel->checkPassword($oldPassword)) {
                     // Always check password
                     Yii::app()->setFlashMessage(gT("Your new password was not saved because the old password was wrong."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
-                } elseif (trim($oldPassword) === trim($newPassword)) {
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
```
olleharstedt: Remove `return`.

application/controllers/admin/useraction.php
```diff
@@ -643,28 +643,31 @@ public function personalsettings()
...
+                }
+                if (trim($oldPassword) === trim($newPassword)) {
                     //First test if old and new password are identical => no need to save it (or ?)
                     Yii::app()->setFlashMessage(gT("Your new password was not saved because it matches the old password."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
-                } elseif (trim($newPassword) !== trim($repeatPassword)) {
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
```
olleharstedt: Same.

application/controllers/admin/useraction.php
```diff
@@ -643,28 +643,31 @@ public function personalsettings()
...
+                }
+                if (trim($newPassword) !== trim($repeatPassword)) {
                     //Then test the new password and the repeat password for identity
                     Yii::app()->setFlashMessage(gT("Your new password was not saved because the passwords did not match."), 'error');
-                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
+                    return $this->getController()->redirect(array("admin/user/sa/personalsettings"));
```
olleharstedt: Same.

cdorin93:
The issue is already fixed in current master. Thank you all for your contributions! I will now close the PR.

PR #1376, opened 2020-01-25
[WIP] 15224 get question attributes event

Shnoulle:
Fixed issue #15224: newQuestionAttributes deprecated : need a replacer
Dev: create a function
Dev: use it when needed

### DIFF
--- a/application/helpers/questionHelper.php
+++ b/application/helpers/questionHelper.php
@@ -1644,28 +1644,10 @@ class questionHelper
             $aQuestionAttributes[$aCoreAttribute['name']] = $aCoreAttribute;
         }
 
-        /**
-         * New event to allow plugin to add own question attribute (settings)
-         * Using $event->append('questionAttributes', $questionAttributes);
-         * $questionAttributes=[
-         *  attributeName=>[
-         *      'types' : Aply to this question type
-         *      'category' : Where to put it
-         *      'sortorder' : Qort order in this category
-         *      'inputtype' : type of input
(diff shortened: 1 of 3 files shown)


application/helpers/questionHelper.php
```diff
@@ -1644,28 +1644,10 @@ public static function getQuestionThemeAttributeValues($type, $sQuestionThemeNam
             $aQuestionAttributes[$aCoreAttribute['name']] = $aCoreAttribute;
         }
 
-        /**
-         * New event to allow plugin to add own question attribute (settings)
```
Shnoulle: Since old plugin using this event
1. Can broke with debug
2. Happen only in some condition (didn't work … with all question type)
3. value are not saved

I choose to remove

application/models/Question.php
```diff
@@ -295,8 +295,10 @@ public function getAdvancedSettingsWithValues($iQuestionID, $sQuestionType, $iSu
         // TODO: move getQuestionAttributesSettings() to QuestionAttribute model to avoid code duplication
         $aAttributeNames = QuestionAttribute::getQuestionAttributesSettings($sQuestionType);
 
-        // If the question has a custom template, we first check if it provides custom attributes
+        // for questioneditor and public question view. Used QuestionBaseDataSet and Survey_Common_Action
+        //$aAttributeNames = array_merge($aAttributeNames, QuestionAttribute::getQuestionAttributesPlugins($sQuestionType));
```
Shnoulle: In my test : this work without this one … need more testing, maybe related to question type

olleharstedt:
IMO: The question attribute code is needlessly complex and needs to be remade, with a proper design.

PR #1370, opened 2020-01-16
[WIP] Fixed issue #15224: newQuestionAttributes deprecated : need a replacer

Shnoulle:
Dev: replace by getQuestionAttributes, because restricted by type
Dev: create QuestionAttribute::getQuestionAttributesPlugins to get the definition
Dev: add new definition in QuestionAttribute::model()->getQuestionAttributes
Dev: add new definition in Question::model()->getAdvancedSettingsWithValues

PR #1377, opened 2020-01-27
Fix DB update issues when updating from old Limesurvey versions

mbischof:
Fixed issue # : 15759
New feature # :
Changed feature # :
Dev:

### DIFF
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -1291,7 +1291,9 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
         // LS 2.5 table start at 250
         if ($iOldDBVersion < 250) {
             $oTransaction = $oDB->beginTransaction();
-            createBoxes250();
+            if (!tableExists('boxes')) {
+                createBoxes250();
+            }
             $oDB->createCommand()->update('{{settings_global}}', array('stg_value'=>250), "stg_name='DBVersion'");
             $oTransaction->commit();
         }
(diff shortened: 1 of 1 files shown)


maziminke:
There is also a bug report at https://bugs.limesurvey.org/view.php?id=15759.

Shnoulle:
Question : maybe the existing table must be renamed (or deleted) and created again ?

Shnoulle:
Or better : just remove this createBoxes …

Created again : https://github.com/LimeSurvey/LimeSurvey/blob/50ab52cf35fb5e5f146782592211ae4aacdd783b/application/helpers/update/updatedb_helper.php#L1946-L1961

at dbversion 338

application/helpers/update/updatedb_helper.php
```diff
@@ -1339,6 +1339,7 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
         if ($iOldDBVersion < 256) {
             $oTransaction = $oDB->beginTransaction();
             upgradeTokenTables256();
+            $oDB->createCommand()->update('{{participants}}', array('email' => ''), 'email IS NULL');
```
Shnoulle: Think is not related to this issue …


I do same … staring from a fork to do another fork : https://github.com/LimeSurvey/LimeSurvey/pull/1370#issuecomment-578443506 :)

olleharstedt: If this is not related to the issue, please remove from this PR and add it to another one.

maziminke:
Both commits deal with DB update issues when updating from old versions. We will simply rename the pull request to "Fix DB update issues when updating from old Limesurvey versions".

maziminke:
Please review and merge, thanks.

olleharstedt:
I already requested a change.

maziminke:
@olleharstedt Seriously, this is simply fixing issues during updating from old versions. There is no complexity. Why that additional work of separating it into different pull requests? 
There is no benefit and dealing with such really minor fixes that way will lead to less fixes by the community or Limesurvey partners. It is not worth the effort.

olleharstedt:
Ah, I was mostly following Denis' advice that it was unrelated. A commit should be a unit of work that does only one thing. This makes the git history more clear. But I will merge this. The commit is so small anyway.

olleharstedt:
Thank you for the contribution!

maziminke:
Thanks, Olle!

PR #1379, opened 2020-01-28
Fixed issue #15794: Question is LS Expression (replacementfields) show question from next group

Shnoulle:
Dev: usage of compare Yii function

Please : have a look at Yii code : https://www.yiiframework.com/doc/api/1.1/CDbCriteria#compare-detail

### DIFF
--- a/application/controllers/admin/limereplacementfields.php
+++ b/application/controllers/admin/limereplacementfields.php
@@ -446,16 +446,16 @@ class limereplacementfields extends Survey_Common_Action
         if ($gid != null && $qid == null) {
             $oGroup = QuestionGroup::model()->findByPk($gid);
             $oCriteria->with = ['group'];
-            $oCriteria->compare('group_order', $oGroup->group_order, false, '<=');
+            $oCriteria->compare('group_order', '<='.$oGroup->group_order,);
         }
 
         if ($qid != null) {
             $oCriteria->with = ['group'];
-            $oCriteria->compare('group_order', $oCurrentQuestion->group->group_order, false, '<=');
+            $oCriteria->compare('group_order', '<='.$oCurrentQuestion->group->group_order);
             if ($oCurrentQuestion->parent_qid != 0) {
-                $oCriteria->compare('question_order', $oCurrentQuestion->parent->question_order, false, '<');
+                $oCriteria->compare('question_order', '<'.$oCurrentQuestion->parent->question_order);
             } else {
-                $oCriteria->compare('question_order', $oCurrentQuestion->question_order, false, '<');
+                $oCriteria->compare('question_order', '<'.$oCurrentQuestion->question_order );
             }
         }
 


Shnoulle:
Checked save question in russion with this fork : seems OK
Peek 28-01-2020 18-30

olleharstedt:
Have to test on postgres also. I can do it on Friday (workshop on Thursday).

application/controllers/admin/limereplacementfields.php
```diff
@@ -446,16 +446,16 @@ private function collectQuestionReplacements($surveyid, $gid = null, $qid = null
         if ($gid != null && $qid == null) {
             $oGroup = QuestionGroup::model()->findByPk($gid);
             $oCriteria->with = ['group'];
-            $oCriteria->compare('group_order', $oGroup->group_order, false, '<=');
+            $oCriteria->compare('group_order', '<='.$oGroup->group_order,);
```
olleharstedt: Cast to int?

Shnoulle: Don't think : 
See Yii code

$value is set here https://github.com/yiisoft/yii/blob/d01658e47d799e5778608982a45bfcf2d4062a17/framework/db/schema/CDbCriteria.php#L427

And use PDO param for value : https://github.com/yiisoft/yii/blob/d01658e47d799e5778608982a45bfcf2d4062a17/framework/db/schema/CDbCriteria.php#L447

The original issue seems top be about column name , not value

(I check with my postgre version an no issue. )

Shnoulle: Yes confirm:

$this->addCondition($column.$op.self::PARAM_PREFIX.self::$paramCount,$operator);

Become 
`$this->addCondition('group_order < :ycp0,'AND');`

And before the previous fix : 
`$this->addCondition('group.group_order < :ycp0,'AND');`

olleharstedt: `addCondition` is not same as `compare`?

olleharstedt: Oh, they actually split the value. Good lord.

olleharstedt: Better  to use array instead of string, then. More clear.

Shnoulle: > Better to use array instead of string, then. More clear.

You want to rewrite YII ?

In place of compare we can use addCondition directly :) .

Else we can use [Yii::app()->db->quoteColumnName](https://github.com/LimeSurvey/LimeSurvey/blob/eba81f2ab24d2b2484d7b2b1d730df191b573aed/application/models/Quota.php#L181) in addCondition or compare …

olleharstedt: > You want to rewrite YII ?

No, $value can be both array and string.

Shnoulle:
Yes, but with array : you can not compare with `>` or `<=` etc …

> the column value to be compared with. If the value is a string, the aforementioned intelligent comparison will be conducted. I**f the value is an array, the comparison is done by exact match of any of the value in the array**. If the string or the array is empty, the existing search condition will not be modified.

More like a addInCondition ?

Maybe addCondition is the cleanest and most understandable way ?

PS : there are some other place where we have compare …

Shnoulle:
@olleharstedt : i can use addCondition in place of compare …

olleharstedt:
> Yes, but with array : you can not compare with > or <= etc …

Ah, shit, I thought the first element in the array could be the operator. Yii sucks. :(

Shnoulle:
RTFM :D :dancing_women:

olleharstedt:
^^

PR #1380, opened 2020-01-31
Update AuthLDAP.php "Error: Undefined Variable userentry"

orionmcg:
Avoid undefined variable error by initializing $userentry before for loop.


olleharstedt:
Good PR, thanks you!

PR #1383, opened 2020-02-05
Fix for empty token value when calling Token->generateToken()

bkidwell:
On a Windows installation I had a mysterious problem where calls the remote API function `add_participants` were resulting in new participants with empty tokens.

I found out that the problem is:
* `Yii::app()->securityManager->generateRandomString()` always returns **false** if there is an error.
* In my Windows installation, out of the box, the random string function is failing, presumably because `mcrypt` and `openssl` extensions were disabled. I enabled `openssl` and my problem went away.
* `Token->_generateRandomToken()` doesn't check the return value.

I believe that `Token->_generateRandomToken()` should fail hard when it gets a failure value from the Yii framework random string function.

This patch enacts that policy. **Please edit the error message to suit your project's style, if the message is not to your liking.**

### DIFF
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -245,7 +245,7 @@ abstract class Token extends Dynamic
     {
         $token = Yii::app()->securityManager->generateRandomString($iTokenLength);
         if ($token === false) {
-            throw new CHttpException(500, 'Failed to generate random string for token. Please check your configuration and ensure that the openssl or mcrypt extension is enabled.');
+            throw new CHttpException(500, gT('Failed to generate random string for token. Please check your configuration and ensure that the openssl or mcrypt extension is enabled.'));
         }
         $token = str_replace(array('~', '_'), array('a', 'z'), $token);
         $event = new PluginEvent('afterGenerateToken');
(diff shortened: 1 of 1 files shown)


application/models/Token.php
```diff
@@ -243,7 +243,11 @@ public function generateToken($tokenlength = NULL)
      */
     private function _generateRandomToken($iTokenLength)
     {
-        $token = str_replace(array('~', '_'), array('a', 'z'), Yii::app()->securityManager->generateRandomString($iTokenLength));
+        $token = Yii::app()->securityManager->generateRandomString($iTokenLength);
+        if ($token === false) {
+            throw new CHttpException(500, 'Failed to generate random string for token. Please check your configuration and ensure that the openssl or mcrypt extension is enabled.');
```
olleharstedt: Thank you for your contribution. Can you please put the string inside the `gT` function so that our translation bot can find it? Thank you.

bkidwell: Done. Sorry about that. I wasn't sure if it was wanted, because an adjacent Exception error message in the calling function wasn't formatted this way.

maziminke:
Thanks for tracking this down and providing a fix!

PR #1384, opened 2020-02-07
[WIP][DONTMERGE] Dev: 400 update step by step

Shnoulle:
For discussion.

Alternate way : 390 + 391 + 392  ?

Current **major** issue : if it broke during 400 upgrade and show a solution (and user have DB access) : there are no way to launch again the upgrade process

### DIFF
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -2431,191 +2431,238 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
             $oTransaction->commit();
         }        
 
+        /* 400 upgrade … */
         if ($iOldDBVersion < 400) {
             // This update moves localization-dependant strings from question group/question/answer tables to related localization tables
-            $oTransaction = $oDB->beginTransaction();
-            
-            // Question table 
-            /* l10ns question table */
-            if(Yii::app()->db->schema->getTable('{{question_l10ns}}')){
-                $oDB->createCommand()->dropTable('{{question_l10ns}}');
-            }
-            $oDB->createCommand()->createTable('{{question_l10ns}}', array(
-                'id' =>  "pk",
-                'qid' =>  "integer NOT NULL",
-                'question' =>  "text NOT NULL",
-                'help' =>  "text",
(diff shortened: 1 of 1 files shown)


Shnoulle:
@maziminke maybe you have broken DB ? Then we can test if fix the error returned directly inside DB and relauch upgrade process can work.

olleharstedt:
IMO, first step is to move all `dropTable` to the end of the update.

Shnoulle:
> IMO, first step is to move all dropTable to the end of the update.

More work for user :  right ?

olleharstedt:
> More work for user : right ?

If it fails? Yes, the restore procedure would be manual.

olleharstedt:
But less testing for us (me).

Shnoulle:
1st step : have some broken DB (with mariadb or pg)

olleharstedt:
> 1st step : have some broken DB (with mariadb or pg)

Easy, just manipulate the collation before upgrading to LS4.

Shnoulle:
Maybe but must have too different errors. I see some pk in Label (really don't understand how this can happen ....)

explorerdk:
I have the same issue, I had to rollback to backup copy of Site / DB, if helpful i will supply a copy of Database for debugging....   the same error posted on Mantis Bug report is what it does to me also.  Tried changing collation, but that did not help with the 1st error.  

My Survey DB is not to big either.,.  

/Explorerdk

olleharstedt:
@explorerdk You should write something on Mantis so that we know more people are affected. Thank you.

maziminke:
Currently I can't share a broken DB due to privacy issues, sorry.

explorerdk:
> Currently I can't share a broken DB due to privacy issues, sorry.

@maziminke  I have mine Anonymized, so I was proposing to share my small Survey db  :)

Shnoulle:
@olleharstedt 

> IMO, first step is to move all dropTable to the end of the update.

What i can do is : 

1. remove the step by step
2. move the drip table at end
3. Keep the checking of XX_400 existence before working on XX and use the _400 one by rename ? (for example question_400 exist ; delete question + questioni10n, rename question_400 to question and try again)

PR #1385, opened 2020-02-07
Fixed issue #15816: Order groups when loading them in UpdateGroupList.

Jelle-S:
Fixed issue #15816 : Order groups when loading them in UpdateGroupList.


Jelle-S:
I'm not sure what the test is failing on, but it doesn't seem related?

PR #1387, opened 2020-02-13
Fixed issue #15859: Token can be truncated in SPSS export

TonisOrmisson:
Token length is wrong. Or can be wrong in many cases. Token column in survey is set to 'string(36)'. Tokens can be imported or created via API in a longer format than set in the survey settings. So We are ending up truncated tokens in SPSS file. The survey->tokenlength should be also used while creating the token field in DB (and API/import length check) OR the SPSS should generate a field allowing max length that can come from DB (32).

### DIFF
--- a/application/helpers/admin/activate_helper.php
+++ b/application/helpers/admin/activate_helper.php
@@ -403,7 +403,7 @@ function activateSurvey($iSurveyID, $simulate = false)
                 }
                 break;
             case "token":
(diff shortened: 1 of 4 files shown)


c-schmitz:
Thank you!

TonisOrmisson:
Will this end up in v3 LTS also?

maziminke:
Please also merge this into 3.x LTS!

PR #1390, opened 2020-02-19
Dev: broken SQL for PG or MS in customToken plugin

Shnoulle:
Dev: and always better to use PDO with Yii
Dev: if elseif is really awfull. Usage of switch
Dev: Add translation system (check if it's OK , $this->gT must work)

### DIFF
--- a/application/core/plugins/customToken/config.xml
+++ b/application/core/plugins/customToken/config.xml
@@ -7,7 +7,7 @@
         <last_update>2019-07-22</last_update>
         <author>Tools for Research</author>
         <authorUrl>https://www.toolsforresearch.com</authorUrl>
-        <version>1.0.0</version>
+        <version>1.0.1</version>
         <license>MIT</license>
         <description><![CDATA[Numeric, non-ambiguous or CAPITAL tokens]]></description>
     </metadata>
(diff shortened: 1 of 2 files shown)


Shnoulle:
I k,now it's bad to fix without issue. But i can not report issue because i don't have a pgsql currently.

But this code are awfull
https://github.com/LimeSurvey/LimeSurvey/blob/592cafefee9f682feb50476e1876dd29dee8a07e/application/core/plugins/customToken/customToken.php#L119
We **must not** show broken SQL code in plugins.

Included plugin are used by external developer .. for inspiration

Else : usage of if /elseif/elseif/esleif .... Arg .... 

The last update gT for translation where we can.

Shnoulle:
About https://github.com/LimeSurvey/LimeSurvey/blob/592cafefee9f682feb50476e1876dd29dee8a07e/application/core/plugins/customToken/customToken.php#L119

This delete other plugin with customToken key usage .... too .

olleharstedt:
Steps how to reproduce the bug?

Shnoulle:
Get a PGSQl
Activate plugin
Deactivate plugin

Else : i don't create a plugin using customToken setting too ..but reading code is OK.

It is a clear issue in code ..

Shnoulle:
About "Steps how to reproduce the bug?" then we never rewrite for better understandable code or optimization ?

Shnoulle:
With customToken bug and mysql

1. Create a plugin `myPlugin` using customToken setting (global for example)
2. Update this plugin setting
3. Activate customToken plugin
4. deativate customToken plugin
5. Look at `myPlugin` : customToken resetted to null.

olleharstedt:
:+1:

PR #1391, opened 2020-02-21
Added plugin events for Twig Renderer

tgoellner:
Creates two new plugin events for rendereing Twig templates:

* `TwigRendererConvertTwigToHtml` is called just before a template is rendered and enables a plugin to update the `$aDatas` array as well as the `$sString` before those are rendered by the Twig Renderer
* `TwigRendererRenderHtmlPage` is called before the whole rendered template is output to the buffer and provides the whole `$sHtml `content and the `$oTemplate` variable

### DIFF
--- a/application/core/LSETwigViewRenderer.php
+++ b/application/core/LSETwigViewRenderer.php
@@ -124,7 +124,7 @@ window.addEventListener('message', function(event) {
     event.source.postMessage({msg: 'EVENT COLLECTED', event: event.data}, '*');
     var getUrl = window.location;
     var baseUrl = getUrl.protocol + '//' + getUrl.host + '/' + getUrl.pathname.split('/')[1];
-    
+
     if (baseUrl.match(event.origin) && event.data.run == 'trigger::pjax:scriptcomplete' ) {
         console.log('runScriptcomplete');
         jQuery(document).trigger('pjax:scriptcomplete');
@@ -135,7 +135,7 @@ window.addEventListener('message', function(event) {
         jQuery('#questionPreview--content').text('');
         jQuery('#questionPreview--content').html(event.data.content);
     }
-}, false);  
+}, false);
 ";
             App()->getClientScript()->registerScript(
                 'postMessageScripts',
@@ -155,7 +155,7 @@ window.addEventListener('message', function(event) {
             }
 
             $sHtml     = $this->convertTwigToHtml($line, $aDatas, $oTemplate);
-            
+
             $sEmHiddenInputs = LimeExpressionManager::FinishProcessPublicPage(true);
             if ($sEmHiddenInputs) {
                 $sHtml = str_replace(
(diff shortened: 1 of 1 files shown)


Shnoulle:
We don't have the twig file name ? 

Can you explain the exact purpose here ?

tgoellner:
The purpose is that a plugin is now able to alter the $aData variable so within the twig templates we can access those added/altered variables. Also it might be necessary for some plugins (at least I have some use cases) to alter the complete HTML output before it gets rendered.

Shnoulle:
It's not purpose here, it's too global ...

>  to alter the $aData variable so within the twig templates

We can already replace twig file AND create tiwg function if i don't make error.

>  alter the complete HTML output before it gets rendered

https://gitlab.com/SondagesPro/coreAndTools/maintenanceMode

maziminke:
In general I like the idea of having such additional options to adjust the output.

Shnoulle:
Yes, so am i. But here : i don't think it's the best situation.

Personnaly : i lke to have a way to "addAnyTwig" or "addAnyHTML"  to any twig file.

Bu here : current twig file are not in event (and i don't see how adding it). Maybe there are some place to add it.

My idea was more something like this

```
$event = new PluginEvent('TwigRenderer');
$event->set('template',$twigfile);
$event->set('twigfile',$twigfile);
```
And plugin must be allowed to add twig file (or html content) before and after the htmùl produced by twig. And mybe replace the twig file.

About the data : i really think `$aSurveyInfo` muts be extendable by plugins. But all of this array used are really a mess .... afraid of API update.

Shnoulle:
Some sample usage

1. Add a dialog box but not with JS : https://gitlab.com/SondagesPro/QuestionSettingsType/questionExtraSurvey/-/blob/master/questionExtraSurvey.php#L573-576
2. Replace some specific twig file by plugin : https://gitlab.com/SondagesPro/QuestionSettingsType/arrayFilterHtml/-/blob/master/arrayFilterHtml.php (here : best is to allow plugin to replace answer_row.twig` after checked if need to be replaced)
3. etc ...

olleharstedt:
> The purpose is that a plugin is now able to alter the $aData variable so within the twig templates we can access those added/altered variables. Also it might be necessary for some plugins (at least I have some use cases) to alter the complete HTML output before it gets rendered.

Can you be a bit more specific of the use-cases, please?

c-schmitz:
The plugin event is too unspecific and reasoning was not properly given.

Shnoulle:
> Some sample usage
> 
>     1. Add a dialog box but not with JS : https://gitlab.com/SondagesPro/QuestionSettingsType/questionExtraSurvey/-/blob/master/questionExtraSurvey.php#L573-576
> 
>     2. Replace some specific twig file by plugin : https://gitlab.com/SondagesPro/QuestionSettingsType/arrayFilterHtml/-/blob/master/arrayFilterHtml.php (here : best is to allow plugin to replace answer_row.twig` after checked if need to be replaced)
> 
>     3. Add an element about RGPD/CNIL in all survey, directly in HTML
> 
>     4. Add some shortcut links for a11Y : https://gitlab.com/SondagesPro/SurveyThemes/skelvanilla/-/blob/master/views/subviews/navigation/shortcuts_links.twig (in all theme)
> 
>     5. ....

You really need more ?

It's really interesting to allow **update** (without replacing) existing twig file

olleharstedt:
Maybe open a feature request on Mantis...?

Shnoulle:
Got it : https://bugs.limesurvey.org/view.php?id=13889

PR #1392, opened 2020-02-21
Fixed issue #15873: setVariableExpressionEnd have lack of system to udate question part

Shnoulle:
- Add questionSeq2relevance in set / get
- plugin have all they can have but .... not really stable since EM must be rewritten
- **but** EM is same since years now ....

PR #1393, opened 2020-02-21
Master 15873 set variable expression end api update

Shnoulle:
- Add function in API
- Remove the get(knowVars) from event (needed)
- use API in plugins

We can not be really EM safe, since EM is not safe by itself ....


Shnoulle:
My opinion : before a clean rewriting of Expression : create API function is silly ...

Shnoulle:
Fixed by https://github.com/LimeSurvey/LimeSurvey/pull/1392

Before open Expression to API :! must clean up this Expression ....

PR #1394, opened 2020-02-27
Fixed issue #15920: PHP notice when submit a survey with refurl

Shnoulle:
Dev: move $this->LEMsessid to setVarFromArgs function
Dev: send surveyid from the param in setVarFromArgs
Dev: cherry-pick


Shnoulle:
Not major issue except for user with debug > 1

PR #1395, opened 2020-02-27
Fixed issue #15924: Can not send invitations to imported participants where emailstatus IS NULL

acran:
Fixed issue [#15924](https://bugs.limesurvey.org/view.php?id=15924): Can not send invitations to imported participants where emailstatus IS NULL

Dev: although the WHERE condition "emailstatus <> 'OptOut'" *intuitively* should match everything that is not equal to "OptOut" it actually does *not* match NULL values in MySQL since it is not NULL-safe.

Dev: This is related to [#15772](https://bugs.limesurvey.org/view.php?id=15772) and [#15823](https://bugs.limesurvey.org/view.php?id=15823) since emailstatus should not be NULL in the first place.

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -2615,7 +2615,7 @@ class tokens extends Survey_Common_Action
         if ($request->getPost('bypassbademails') == '1') {
             return "emailstatus = 'OK'";
         } else {
-            return "emailstatus <> 'OptOut'";
+            return "emailstatus <> 'OptOut' OR emailstatus IS NULL";
         }
     }
 


application/controllers/admin/tokens.php
```diff
@@ -2615,7 +2615,7 @@ protected function getSQLemailstatuscondition()
         if ($request->getPost('bypassbademails') == '1') {
             return "emailstatus = 'OK'";
         } else {
-            return "emailstatus <> 'OptOut'";
+            return "emailstatus <> 'OptOut' OR emailstatus IS NULL";
```
acran: although the WHERE condition "emailstatus <> 'OptOut'" intuitively should match everything that is not equal to "OptOut" it actually does not match NULL values in MySQL since it is not NULL-safe.

Shnoulle:
My opinion : 

replace OK by 'OK' or null .... null => not set, default, default MUST be OK

acran:
@Shnoulle i.e. change the `if` branch to
```php
        if ($request->getPost('bypassbademails') == '1') {
            return "emailstatus = 'OK' OR emailstatus IS NULL";
        }
```
?

Makes sense to me, but I'm not familiar enough with the remaining code/logic around this. Should I update my PR with the above?

olleharstedt:
Denis is sick this week, but maybe he can review it the next. @Shnoulle for reminder.

Shnoulle:
Sorry didn't answered ....

Not sure about the ideal solution .... but globally my opinion on emailstatus are OK or NULL or '' mean : it's OK to send it ...

But we use 'ok' since year now .... then maybe it can broke some setup ?

I think https://github.com/LimeSurvey/LimeSurvey/pull/1395#issuecomment-593096033 is OK, but what it's your opinion on this solution ? 

It's not a dev opinion, more a user opinion .... @cdorin93 maybe must give his opinion ....

olleharstedt:
Sounds good. Dorin is on vacation next week, but let's ping him when he's back.

acran:
@olleharstedt @Shnoulle @cdorin93 I'm not quite sure about the current situation of this PR: is it waiting for an update by me respective to https://github.com/LimeSurvey/LimeSurvey/pull/1395#issuecomment-593096033 or should I wait for feedback first?

c-schmitz:
I think you could just update your tokens to OK where the email status is empty. I am against allowing for NULL as being equal to OK.

Shnoulle:
@c-schmitz i think in 3 and since 2.50 (maybe before) there are no way to set NULL to emailstatus (except directly in database).

acran:
@c-schmitz:

> I think you could just update your tokens to OK where the email status is empty.
> You can use the participant batch-edit to filter for tokens that are null and set them to ok.

In the UI there is no way to differentiate between NULL values and empty strings since both are rendered as empty strings. This makes this issue so confusing since saving the token "without making any changes" the value from NULL to empty string in the background and thus its behavior.

> I am against allowing for NULL as being equal to OK

Which is also not what this PR does. Instead it ensures that NULL is _not equal to 'OptOut'_ see https://github.com/LimeSurvey/LimeSurvey/pull/1395#pullrequestreview-365560362

> as it could have unintended side-effects with people using this in some way. The underlying bug causing the empty emailstatus has long been fixed.

As said before: in the frontend NULL can not be distinguished from an empty string in the database. The underlying bug may had been fixed at some point but in the master branch this PR is based off of it was still present. See also the reated open bugs: [15772](https://bugs.limesurvey.org/view.php?id=15772), [15823](https://bugs.limesurvey.org/view.php?id=15823)

olleharstedt:
@acran Can you please join us at our next team meeting so we can discuss this issue?

acran:
@olleharstedt of course. Where can I find the information on when the next meeting takes place and how to join?

olleharstedt:
> @olleharstedt of course. Where can I find the information on when the next meeting takes place and how to join?

chat.freenode.net, #limesurvey 

Some more info here: https://manual.limesurvey.org/How_to_join_the_LimeSurvey_project_team

PR #1396, opened 2020-02-27
Fixed issue #15844: api get_question_properties not responding the su…

m-rm:
It seems that a few instances of a variable where missed during refactoring.

Fixed issue # : 15844 https://bugs.limesurvey.org/view.php?id=15844


olleharstedt:
Thanks!

PR #1397, opened 2020-03-10
Fixed issue #15973: Invalid submitfailed received

Shnoulle:
Dev: PDOStatement::rowCount are unstable and unsure.
Dev: then don't use it

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -5586,23 +5586,16 @@
                     }
                     //If the responses already have been submitted once they are marked as completed already, so they shouldn't be changed.
                     if ($oResponse->submitdate == null || Survey::model()->findByPk($this->sid)->alloweditaftercompletion == 'Y') {
-                        $iCountUpdated = Response::model($this->sid)->updateByPk($oResponse->id,$aResponseAttributes);
-                        if (!$iCountUpdated) {
-                            // Fix #15733 comparing values of previous response
-                            $aReponseCurrentAttributes = $oResponse->getAttributes();
-                            $aIntersect = array_intersect_key($aReponseCurrentAttributes,$aResponseAttributes);
-                            if (!empty(array_diff_assoc($aResponseAttributes,$aIntersect))) {
-                                $message = submitfailed('',$this->gT('Error on response update'));
-                                LimeExpressionManager::addFrontendFlashMessage('error', $message, $this->sid);
-                            }
-                        } else {
-                            // Action in case its saved with success : to be move in Response::aferSave ?
-                            // Save Timings if needed
-                            if ($this->surveyOptions['savetimings']) {
-                                Yii::import("application.libraries.Save");
-                                $cSave = new Save();
-                                $cSave->set_answer_time();
-                            }
+                        Response::model($this->sid)->updateByPk($oResponse->id,$aResponseAttributes);
+                        /* See https://www.php.net/manual/en/pdostatement.rowcount.php :
+                         * this behaviour is not guaranteed for all databases and should not be relied on for portable applications.
+                         */
+                        // Action in case its saved with success : to be move in Response::aferSave ?
+                        // Save Timings if needed
+                        if ($this->surveyOptions['savetimings']) {
+                            Yii::import("application.libraries.Save");
+                            $cSave = new Save();
+                            $cSave->set_answer_time();
                         }
                     } else {
                         LimeExpressionManager::addFrontendFlashMessage('error', $this->gT('This response was already submitted.'), $this->sid);


Shnoulle:
I do the fix for master when this one was merged.

Since [PDOStatement::rowCount](https://www.php.net/manual/en/pdostatement.rowcount.php) is not guaranted and can not be used for portable applications. 

And : SQL error throw error directly.

olleharstedt:
>this behaviour is not guaranteed for all databases and should not be relied on for portable applications.

99% of LS installations use MySQL? So OK to rely on?

Shnoulle:
> 99% of LS installations use MySQL? So OK to rely on?

Some month ago : we never have issue, seems mariadb return 0 if there are no real update since only some month ? And mariadb is used a lot. The original issue is for mariadb and mariadb replace mysql a lot ....
Seems PGSQL need a specific settings for this

Shnoulle:
Maybe an update in default system ?

`PDO::MYSQL_ATTR_FOUND_ROWS` https://www.php.net/manual/en/ref.pdo-mysql.php

I have the issue in a debian server and a fedora server ...

olleharstedt:
Ah, right, I remember.

Shnoulle:
> 99% of LS installations use MySQL? So OK to rely on?

I think it's false : most mariadb, after pgsql, after mysql and finally MSSQL.

But i think big system, big company use most PGSQL or MSSQL.

otheus:
UPDATE: "99% of LS installations use MySQL? So OK to rely on?" @Shnoulle and @olleharstedt  The reason PDF Update-rowCount should not be relied upon *to indicate success or failure* is because if the update changes 0 rows, MySQL-PDO will report a rowCount 0. Instead, the exception handler should be used to catch problems doing the UPDATE call. See  https://stackoverflow.com/questions/25167835. I also confirmed this on my system. 

I spent hours tracking down why the system was failing survey responses in the installation using the version prior to this commit; and it appears to be caused by two updates executed in a row, possibly do a double-click on the submit page or something. (There also seems to be a related event to a switch from one php-session to another for the same survey.)

olleharstedt:
Double click? Can you reproduce that?

Shnoulle:
I don't understand the question.

olleharstedt:
> possibly do a double-click on the submit page or something

Shnoulle:
> > possibly do a double-click on the submit page or something

I try manually : really hard to do currently.  Modern browser work good (even without JS)

otheus:
> double click

I don't really know. 

We see these surveys that every now and then (once every few days) call `submitfailed()` for no apparent reason, and the analysis with debugging SQL led me to this hypothesis. 

It might also (instead?) have to do with session expiration. The user had a session, started a survey with response id 117, but mid-way through, the session gets changed and response id 118 is used. 

```
2021-07-25T18:43:08.290955+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[1170]:[system.db.CDbCommand] Executing SQL: UPDATE `sessions` SET `id`=:id WHERE id=:oldID. Bound with :oldID='k3ibdjl95gqu4v394q81sif22f', :id='nff802fsvhc6aequ8sbrekafcp'
...
2021-07-25T18:43:29.604102+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[1148]:[system.db.CDbCommand] Executing SQL: UPDATE `sessions` SET `id`=:id WHERE id=:oldID. Bound with :oldID='nff802fsvhc6aequ8sbrekafcp', :id='dp0a8l6em5t3du2s4ourcsn8pg'
```

The timestamp of the debug statement which would have generated the `submitfailed` call corresponds precisely to the second query above (pid, timestamp). 

In the end, survey response id 118 was completely filled out, while 117 has only a few fields populated.

otheus:
It happened again yesterday. Similar features. Apparently the same survey participant:
```
??:48:40.321326+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[14641]:[system.db.CDbCommand] Querying SQL: SELECT `data`, FROM `sessions`, WHERE expire>:expire AND id=:id. Bound with :expire=1627318119, :id='kio2kgl8tllp7huk0kc3q638t2'

??:48:40.334778+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[14641]:[system.db.CDbCommand] Executing SQL: UPDATE `sessions` SET `id`=:id WHERE id=:oldID. Bound with :oldID='kio2kgl8tllp7huk0kc3q638t2', :id='p1puehoa6c4anid366hk1pid1v'

??:48:56.788220+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[14601]:[system.db.CDbCommand] Executing SQL: UPDATE `sessions` SET `id`=:id WHERE id=:oldID. Bound with :oldID='p1puehoa6c4anid366hk1pid1v', :id='080rkpcrnn0ulceeb3694e1em8'
??:48:56.814861+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[14601]:[system.db.CDbCommand] Executing SQL: UPDATE `survey_652142` SET `lastpage`=:yp0, `datestamp`=:yp1, `652142X35812X407463`=:yp2, `652142X35812X407463_filecount`=:yp3 WHERE `survey_652142`.`id`=117. Bound with :yp0=1, :yp1='<REDACTED>:48:56', :yp2=NULL, :yp3=NULL
??:48:56.815104+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[14601]:[system.db.CDbCommand] Executing SQL: UPDATE `saved_control` SET `saved_thisstep`=:yp0 WHERE `saved_control`.`scid`=1761. Bound with :yp0='1'
??:48:56.816067+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[14601]:[system.db.CDbCommand] Executing SQL: UPDATE `survey_652142` SET `lastpage`=:yp0, `datestamp`=:yp1, `652142X35812X407463`=:yp2, `652142X35812X407463_filecount`=:yp3 WHERE `survey_652142`.`id`=117. Bound with :yp0=1, :yp1='<REDACTED>:48:56', :yp2=NULL, :yp3=NULL
??:48:56.816297+02:00 limesurvey [local5.debug] limesurveyDB-allgemein[14601]:[system.db.CDbCommand] Executing SQL: UPDATE `saved_control` SET `saved_thisstep`=:yp0 WHERE `saved_control`.`scid`=1761. Bound with :yp0='1'

```

Shnoulle:
I have some issue with reload system 
> Notice: Undefined variable: sLoadPass in .\application\helpers\frontend_helper.php on line 61 Notice: Undefined variable: sLoadPass in .\application\helpers\frontend_helper.php on line 62 Notice: Undefined variable: sLoadPass in .\application\helpers\frontend_helper.php on line 63

Maybe related ? I have the fix, but not the way to reproduce …

Shnoulle:
Seems you have 2 `saved_thisstep` then, maybe related to the way it saved currently.

Doing some test : seems it('s broken on last version (with token enable survey)

otheus:
I will continue to collect data. So far, every double `saved_control` is indicative of the UPDATE-rowCount == 0 condition / event (Which would have triggered submitfailed). However, (1) single `saved_control` queries are not associated with this event, and (2) there was at least one event that had no `saved_control` query at all.

PR #1398, opened 2020-03-13
Fixed issue #15984: Order of subquestion are not respected in print answers

Shnoulle:
Dev: fix order in relations
Dev: use App()->getDb()->quoteColumnName for security


Shnoulle:
I check on mssql about `'order'=> App()->getDb()->quoteColumnName('subquestions.question_order').' ASC'`

PR #1400, opened 2020-03-24
 Fixed issue #16024: Language changer dropdown autoupdated : keybord navigation issue

Shnoulle:
Fixed issue #16024: Language changer dropdown autoupdated : keybord navigation issue
Dev: checked on chrome and firefox
Dev: disable change if last key was key up or down on select.

### DIFF
--- a/assets/packages/limesurvey/survey.js
+++ b/assets/packages/limesurvey/survey.js
@@ -213,7 +213,16 @@ function activateLanguageChanger(){
         }
     });
     /* Language changer dropdown */
+    /* Don't activate change when using key up / key down */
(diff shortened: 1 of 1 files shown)


Shnoulle:
Maybe need to remove lastkey data in some condition ? Let me check

Shnoulle:
Good idea :)

I must check 
- Key up/down
- Update with mouse ....

Unsure it work

Shnoulle:
I use limesurvey : package name

Else : i made a system for Key up/down + Update with mouse : OK on Firefox.

See https://github.com/LimeSurvey/LimeSurvey/pull/1400/files#diff-4fa8092528a97c23ef397647f75e653fR223

I no other advice : i merge it tomorrow :)

PR #1405, opened 2020-04-17
Fixed issue #16142: Deleting a participant after he/she started the survey keeps the survey available

Shnoulle:
Dev: unset invalid token ($oToken is null)


Shnoulle:
This is a quick fix currently, i think we need a better system to check valid / invalid token .
Unsure if in SurveyRuntimeHelper or somewhere else ?

Maybe a new getCurrentToken function, validate and /or throw error.

olleharstedt:
Should same fix be cherry-picked into master?

Shnoulle:
Yes, surely, i do it after.

olleharstedt:
OK, we just need someone to test this. @cdorin93 Busy, busy?

PR #1409, opened 2020-04-23
Fixed issue #16175: statCountIf returns nothing for first participant 

Shnoulle:
Dev: $result = $result ? $result : false; can never return 0
Dev: js accept null ? No ? Why adding this ? No reason …
Dev: add a test to allow return 0


application/helpers/expressions/em_core_helper.php
```diff
@@ -2120,7 +2120,7 @@ private function RDP_RunFunction($funcNameToken, $params)
                                         if (is_numeric($params[0])) {
                                             $result = $funcName(floatval($params[0]));
                                         } else {
-                                            $result = NAN;
+                                            $result = false;
```
olleharstedt, line 2123: Phew. Almost impossible to figure out the effect of this change. :sweat:

Shnoulle: Yes, BUT : return value must be same if function have 1 or 2 or any parameters … 

See https://github.com/LimeSurvey/LimeSurvey/blob/857b89f0df595790363ea3d429ab32d71d2ffaaf/application/helpers/expressions/em_core_helper.php#L2110-L2125

`atan` return false, but not`atan2` … 

I can reset to NAN but : there are an issue

olleharstedt: Maybe we can look for a unit-test design next week.

Shnoulle: If you're OK : i revert just this part (atant2 : NAN to false)

Just keep https://github.com/LimeSurvey/LimeSurvey/pull/1409/files#diff-7f2b63ae2aaab1ffe6caffbaae39751cL2144

olleharstedt: I'm OK with anything, but we need a unit-test for RDP_RunFunction. Let me glance the code.

Shnoulle: I add it for `0` return , but clearly : the line `$result = $result ? $result : false;` was ONLY HERE if function have 2 parameters.

And NAN return only for atan2, not for all other math function

olleharstedt: Want me to set it up? The method needs to be made public, and maybe some state needs to be set before running the function.

Shnoulle: If only NAN (for atan2) function is an issue, i can revert.

Shnoulle:
I add comments on the NAN part (NAN in PHP) since review are only on this one.

olleharstedt:
OK, but now the only change is in the acceptance test?

Shnoulle:
No

https://github.com/LimeSurvey/LimeSurvey/pull/1409/files#diff-7f2b63ae2aaab1ffe6caffbaae39751cL2144

set `0, `''`, `[]`, `null` to false.

Shnoulle:
I'm sure for 100% we need this commit, i merge.

olleharstedt:
Without a unit-test?

olleharstedt:
RDP_RunFunction is a fragile function, like all of EM. That's why I'd like a unit test of it before doing changes.

Shnoulle:
> OK, but now the only change is in the acceptance test?
https://github.com/LimeSurvey/LimeSurvey/pull/1409#issuecomment-619987716

> Without a unit-test?
https://github.com/LimeSurvey/LimeSurvey/pull/1409#issuecomment-623365927

> RDP_RunFunction is a fragile function, like all of EM. That's why I'd like a unit test of it before doing changes. But maybe your acceptance test already covers it...?

Plese : just check code and update done … 

Before the fix : 

number of argsPassed : 
- 1 : return result (false, null, empty array , 0 , empty string)
- 2 : return false for  (false, null, empty array , 0 , empty string) , else result
- 3 : return result (false, null, empty array , 0 , empty string)
- 4, 5 or 6 (or more) :  return result (false, null, empty array , 0 , empty string)

Just explain me why 2 arguments are not same logic then 1 or 3 …

olleharstedt:
> Plese : just check code and update done …

Reading the change it's _impossible_ to tell the consequence without understanding the whole context. That's why unit tests are needed.

> number of argsPassed :

Good explanation. You just wrote the unit test in pseudo-code. ;)

Shnoulle:
> Good explanation. You just wrote the unit test in pseudo-code. ;)

Unsure, but maybe can use count : count(Q00,Q01) return 0 if no answer set, count(Q00,Q01,Q03) return false if no answer set.

Do you want a check like this ?

olleharstedt:
Let's create a bug report for this instead.

https://bugs.limesurvey.org/view.php?id=16223

PR #1410, opened 2020-04-26
In usergroups make name length max in PHP to correspond to the error …

jmantysalo:
Fixed issue #16188: Off-by-one limit of user group name length

### DIFF
--- a/application/controllers/admin/usergroups.php
+++ b/application/controllers/admin/usergroups.php
@@ -135,7 +135,7 @@ class Usergroups extends Survey_Common_Action
                 $db_group_description = $_POST['group_description'];
 
                 if (isset($db_group_name) && strlen($db_group_name) > 0) {
-                    if (strlen($db_group_name) > 21) {
+                    if (strlen($db_group_name) > 20) {
                         list($aViewUrls, $aData) = $this->index(false, array("type" => "warning", "message" => gT("Failed to add group! Group name length more than 20 characters.")));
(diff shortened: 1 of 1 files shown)


Shnoulle:
Better : 

1. Try to save or validate
2. If not : show validation error
3. Model->rules are OK https://github.com/LimeSurvey/LimeSurvey/blob/388c6debd3c5c4225f45741d641513a18ff9af68/application/models/UserGroup.php#L59
4. We can use [CHtml to show clean error](https://www.yiiframework.com/doc/api/1.1/CHtml#errorSummary-detail)

Shnoulle:
Better : 

1. Try to save or validate
2. If not : show validation error
3. Model->rules are OK https://github.com/LimeSurvey/LimeSurvey/blob/388c6debd3c5c4225f45741d641513a18ff9af68/application/models/UserGroup.php#L59
4. We can use [CHtml to show clean error](https://www.yiiframework.com/doc/api/1.1/CHtml#errorSummary-detail) 

Please : mantis issue ?

jmantysalo:
Mantis issue is 16188: Off-by-one limit of user group name length. Otherwise I currently does not know enough to make better correction for the code.

Shnoulle:
whole file seems deprecated.

https://github.com/LimeSurvey/LimeSurvey/commit/df59832ff0af208397568038261b7171be38a796#diff-8db47338464471400005e3f52e24bca5

Maybe only a 3.5 issue ?

cdorin93:
Mantis issue closed: #16188 . I will close this ticket.

PR #1413, opened 2020-04-29
Fixed issue #16203: Multiple choice questions with "other" optionacti…

Shnoulle:
Fixed issue #16203: Multiple choice questions with "other" optionacti: Setting an answer as "exclusive" and selecting it gives an error message
Fixed issue #16226: Invalid count on multi with exclude : js vs php diff
Dev: irrelevance mean NULL
Dev: exlusive mean irrelevance since years now
Dev: If need a new feature need an option
Dev: 'N' is invalid …


Shnoulle:
Set irrelevant to '' are a big change , you can 't broke a lot of analysis system where null is know .
It's like this since years and years … 

This revert https://github.com/LimeSurvey/LimeSurvey/commit/30a85bb9791d4650653276618a50e3fb1424f911

Shnoulle:
For dev meeting

Shnoulle:
2 bug fixed, revert whole.

And i better understand why it work without other : i think $_SESSION[$this->sessid][$qInfo['rowdivid']] take the last one. Then if last one is exclude : it's OK; But if last one is not the exclude one (for example other) : this break.

I didn't check whole process, but it can be an explanation.

PR #1414, opened 2020-04-30
Fixed issue #16208: escape in Multiple choice with comments

Shnoulle:
Dev: input need to be encoded
Dev: encode must be done by view
Dev: question theme can want json … or another way

### DIFF
--- a/application/core/QuestionTypes/MultipleChoiceWithComments/RenderMultipleChoiceWithComments.php
+++ b/application/core/QuestionTypes/MultipleChoiceWithComments/RenderMultipleChoiceWithComments.php
@@ -120,7 +120,7 @@ class RenderMultipleChoiceWithComments extends QuestionBaseRenderer
                 'inputCommentId'       => 'answer'.$myfname2,
                 'commentLabelText'     => gT('Make a comment on your choice here:'),
                 'inputCommentName'     => $myfname2,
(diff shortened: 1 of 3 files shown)


olleharstedt:
One in our team should test this and then merge.

Shnoulle:
> One in our team should test this and then merge.

We must decide (before) if text by user must be encoded by view (twig) or core.

I'm unsure of the need, but maybe some theme can want not encoded input.

olleharstedt:
> We must decide (before) if text by user must be encoded by view (twig) or core.

Do we? Why not decide after? Fixing database corruption is more important, IMO.

Shnoulle:
There are no DB corruption here … it's ONLY a view issue

Shnoulle:
> Why not decide after?

Because: this broke API … for QuestionTheme …

olleharstedt:
> There are no DB corruption here … it's ONLY a view issue

The bug report says that the database content is corrupted. It's not correct?

>  When the users type a single quote in the comments field, every characters after the quote are lost and not registered in the database !!

Shnoulle:
> The bug report says that the database content is corrupted. It's not correct?

No, i can confirm : saved in DB. But if reload, and resubmit : updated :)

Shnoulle:
(14:59:56) Shnoulle: We sent it encoded in 3.X : we must send it encoded in 4.X …

PR #1415, opened 2020-05-01
New PluginEvent "beforeEmailBody"

tgoellner:
To alter the email that is sent out when a user saves a survey I created a new PluginEvent named `beforeEmailBody` which holds the following parameters:

```
    emailType:  [type, e.g. 'savesurveydetails']
    surveyId:   Survey ID
    subject:    The default subject of the email
    body:       The default body message of the email
    email:      Email address the mail is going to be sent to
```

and which might hold other parameters – for my use case ('savesurveydetails') for example:

```
    surveyName: Name of Survey
    surveyLink: Link to the survey
    savename: Name of the user
    savepass: Password the user chose
```

After the plugin event has been dispatched all properties of the event that can be obtained by $event->get() are replaced in the email body – so in

```
    The survey %surveyName% has the link %surveyLink%
```

the properties of `$event->get('surveyName')` and `$event->get('surveyLink')` will be replaced by the values of the event.


Shnoulle:
This was already done in https://manual.limesurvey.org/BeforeSurveyEmail

For ALL email , not only save … and type are https://github.com/LimeSurvey/LimeSurvey/blob/d1b34d38e1fd920d376043ba2db069152af70aba/application/core/LimeMailer.php#L49

If there are an issue : please : report.

Shnoulle:
Please : confirm i can close.

Shnoulle:
I really don't understand purpose of this …

olleharstedt:
Closed for now, since the use-case is already supported by current event system. Feel free to reopen if an unsupported use-case wants to be discussed.

PR #1422, opened 2020-05-19
New feature #16780: Add a new setting alternativeforgotpwurl to the LDAP plugin

pitbulk:
Related to See #1108

This PR implements a way for Ldap Auth developers to be able to set an alternative Forgot Password URL.


Shnoulle:
Why not add 'forgotpassword' part in plugin content directly ?

I mean 

1. Add the link for AuthDB in https://github.com/LimeSurvey/LimeSurvey/blob/58f3d7bb21d3423c05b088b2f4696fd10e7d17d8/application/core/plugins/Authdb/Authdb.php#L123-L125
2. (in a second step) move all dedicated part to AuthDB plugin : https://github.com/LimeSurvey/LimeSurvey/blob/58f3d7bb21d3423c05b088b2f4696fd10e7d17d8/application/controllers/admin/authentication.php#L221

Since forgot password must be only related to plugin .

Or add a 'extraContent' AuthPluginn part where AuthDB can add forgotpassword part and otherplugin : what they want.

olleharstedt:
Manual should be updated here: https://manual.limesurvey.org/Authentication_plugins#LDAP

Mark new features as available since 4.3.

A unit test or functional test would also be nice. Do we have any automatic tests for core plugins, @Shnoulle ?

Shnoulle:
> Do we have any automatic tests for core plugins, 

Not for LDAP since we don't setup LDAP in travis …

olleharstedt:
> Not for LDAP since we don't setup LDAP in travis …

OK, so maybe a new unit-test will be better, to inspect the generated HTML programmatically?

Shnoulle:
Oh right :)

But then : maybe with a new event param ? forgotpasswordlink ?

If it's null : show current, if it's false : hide it in the forum.

current forgotpassword link muts be move to AuthDB

olleharstedt:
Dunno, I don't have the brain space to think about new features like that.

olleharstedt:
Just, keep it simple...

Shnoulle:
Question : did we accept a jquery action as an official way to fix this. 

@pitbulk : do you think you can update the feature to have a new plugin param ? But without moving dedicated part to AuthDB plugin . Just for the link ? Or the action maybe ( forgotpassword take the param of current plugin).

pitbulk:
Do you want me to move the logic implemented for the LDAP plugin to the AuthDB plugin?

Shnoulle:
Not exactly : add a new event param `forgotpasswordlink` in newLoginForm event

Set it to current one by default (or not ? unsure here) for API compatibility

Something like this

````
            $newLoginForm = new PluginEvent('newLoginForm');
            $newLoginForm->set('forgotpasswordlink',App()->createUrl("admin/authentication/sa/forgotpassword"));
            App()->getPluginManager()->dispatchEvent($newLoginForm);
            $aData['summary'] = self::getSummary('logout');
            $aData['pluginContent'] = $newLoginForm->getAllContent();
            $aData['forgotpasswordlink'] = $newLoginForm->get('forgotpasswordlink');
````

here : https://github.com/LimeSurvey/LimeSurvey/blob/3b1b238b5148087a1b7d43d0e9d4e197f28a0571/application/controllers/admin/authentication.php#L144

pitbulk:
Ok, I will work on something next week ;)

cdorin93:
Hey there! What is the status of this PR ? :)

c-schmitz:
@pitbulk Unless you can modify the PR the way that was requested  by @Shnoulle shnoulle we will have to close it because it appears to be pretty dead right now.

glimz:
Dear @pitbulk we have asked you several times to update the PR without any responses from you. Therefore I will be closing the PR.

PR #1424, opened 2020-05-20
Fixed issue #16247: Launch console can create new security.php file

Shnoulle:
Dev: Throw exception if file already exist
Dev: Read real default directory in ConsoleApplication

Dev:

### DIFF
--- a/application/core/ConsoleApplication.php
+++ b/application/core/ConsoleApplication.php
@@ -44,15 +44,27 @@ class ConsoleApplication extends CConsoleApplication
 
         // Set webroot alias.
         Yii::setPathOfAlias('webroot', realpath(Yii::getPathOfAlias('application').'/../'));
-
         /* Because we have app now : we have to call again the config : can be done before : no real usage of url in console, but usage of getPathOfAlias */
         $coreConfig = require(__DIR__.'/../config/config-defaults.php');
         $consoleConfig = require(__DIR__.'/../config/console.php'); // Only for console : replace some config-defaults
         $emailConfig = require(__DIR__.'/../config/email.php');
         $versionConfig = require(__DIR__.'/../config/version.php');
         $updaterVersionConfig = require(__DIR__.'/../config/updater_version.php');
-        $lsConfig = array_merge($coreConfig, $consoleConfig, $emailConfig, $versionConfig, $updaterVersionConfig);
 
+        $lsConfig = array_merge(
+            $coreConfig,
+            $consoleConfig,
+            $emailConfig,
+            $versionConfig,
+            $updaterVersionConfig
+        );
+
+        if (file_exists(__DIR__.'/../config/security.php')) {
+            $securityConfig = require(  __DIR__.'/../config/security.php');
+            if (is_array($securityConfig)) {
+                $lsConfig = array_merge($lsConfig, $securityConfig);
+            }
+        }
         /* Custom config file */
         $configdir = $coreConfig['configdir'];
         if (file_exists( $configdir .  '/security.php')) {
--- a/application/core/LSSodium.php
+++ b/application/core/LSSodium.php
@@ -125,10 +125,10 @@ class LSSodium {
 	 * Write encryption key to version.php config file
 	 */
 	protected function generateEncryptionKeys(){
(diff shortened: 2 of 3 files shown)


Shnoulle:
I just want to test on the broken server. But it's realla a major issue to update an existing security.php file
Then

1. Check if security.php exist before replace, if exist : throw exception (tested on broken server)
2. Read really the security.php (not tested currently)

Shnoulle:
https://bugs.limesurvey.org/view.php?id=16247

Shnoulle:
Since update securitry.php can broke ALL data : it's a major issue

application/core/LSSodium.php
```diff
@@ -124,7 +124,11 @@ public function decrypt($sEncryptedString, $bReturnFalseIfError=false){
 	 * 
 	 * Write encryption key to version.php config file
 	 */
-	protected function generateEncryptionKeys(){			 	
+	protected function generateEncryptionKeys(){
+	if(is_file(APPPATH.'config/security.php')) {
```
olleharstedt, line 128: Wrong indentation?

olleharstedt: Also, curly brace is not PSR-2 compliant.

Shnoulle: Since ALL file have wrong identation and are PSR2 broke : i didn'tb update identitention …

Shnoulle: Capture d’écran du 2020-05-20 14-23-26

Shnoulle: Related : https://github.com/LimeSurvey/LimeSurvey/pull/1417#discussion_r423551410

Shnoulle: Done in https://github.com/LimeSurvey/LimeSurvey/pull/1424/commits/b65235b4a2ee0e2c65fd74bb746e3048990fde08 but ONLY on new lines

application/core/ConsoleApplication.php
```diff
@@ -44,15 +44,27 @@ public function __construct($aApplicationConfig = null)
...
 
+        $lsConfig = array_merge(
+            $coreConfig,
+            $consoleConfig,
+            $emailConfig,
+            $versionConfig,
+            $updaterVersionConfig
+        );
+
+        if (file_exists(__DIR__.'/../config/security.php')) {
```
olleharstedt, line 62: Should probably use `$configdir` here instead of hard-coded location?

olleharstedt: Sorry, Denis...

olleharstedt: Oh, the location is hard-coded everywhere above. Strange.

Shnoulle: $configdir is broken in some server …

I think config-default `$config['rootdir'] = getcwd();` muts be fixed for console.

A lot of user don't know how to create a multi line cron … on server.

Shnoulle:
@olleharstedt since this one is fixed : i start to work on other console issue …

1. Fix this rootdir/configdir … why not using `__FILE__` ?

application/core/ConsoleApplication.php
Shnoulle: > Should probably use `$configdir` here instead of hard-coded location?

This is done after , before this fix : default security are never read … in specific server …

olleharstedt: Hm OK. :thinking: :cow:

Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/1424#issuecomment-631497967

Shnoulle:
Why we can't use in config-default : 
`$config['rootdir'] =basename(basename(basename(realpath(__FILE__)));`
?

olleharstedt:
@Shnoulle Console is very low priority for us. But if it helps you, go ahead. :)

Shnoulle:
I copy/paste a lmot opf code each time i create a new console plugin 

https://gitlab.com/SondagesPro/mailing/sendMailCron/-/blob/master/sendMailCron.php#L435-473

PR #1430, opened 2020-06-05
Fixed issue #16364: validator assigned to non existent field

mfrankiewicz:
Fixed issue #16364


olleharstedt:
What kind of model is DefaultValue...? Never seen before. You know where it's used? Thanks for the fix, though.

Shnoulle:
https://manual.limesurvey.org/Default_answers

Shnoulle:
It's a child of moving i10n value to own table.

* In 3.LTS : lime_defaultvalue have language and defaultvalue
* In 4.X : both are move to lime_defaultvalue_l10ns

Shnoulle:
the rules are in i10ns , no need other update

mfrankiewicz:
> What kind of model is DefaultValue...? Never seen before. You know where it's used? Thanks for the fix, though.

I know only that it's created during saving default value for question and next it's linked with DefaultValueL10n model which stores value (validated with CDefaultValueValidator).

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/database.php#L182

olleharstedt:
Alright, good info, good fix, will merge.

PR #1432, opened 2020-06-10
Fixed #16210: Timer bug in qanda_helper.php list_dropdown

BertHankes:
Fix a bug: https://bugs.limesurvey.org/view.php?id=16125
Timer bug in qanda_helper.php list_dropdown
Attention: This bug is in Lime3 and Lime4!

Fixed issue #16210:


Shnoulle:
4 didn't use all qanda_helper (it must be cleaned …)

thansk for pull request :)

PR #1433, opened 2020-06-11
Dev: WIP : use cleaner rules for code uniqueness

Shnoulle:
Fixed issue #16377: When editing a question and saving deletion of responses items only one stays


cdorin93:
Hmm, what is the status of this one? Ping: @olleharstedt

Shnoulle:
It's related to the issue with bad compare for unicity.

It's broken in some situation 

Let close (don't have time)

PR #1435, opened 2020-06-18
Fixed issue #15037: Data protection statement texts not saved if text for checkbox is too long

gabrieljenik:
Fixed issue #15037: Data protection statement texts not save if text for checkbox is too long.

Added maxlength = 192 in Data policy settings view for data sec label (editDataSecurityLocalSettings_view.php).
Also added a control in the controller (database.php): if there is an error while saving the language settings, the error is shown and the survey is not saved.

### DIFF
--- a/application/controllers/admin/database.php
+++ b/application/controllers/admin/database.php
@@ -814,6 +814,8 @@ class database extends Survey_Common_Action
         $languagelist = $oSurvey->additionalLanguages;
         $languagelist[] = $oSurvey->language;
 
+        $saveSurvey = true;
+
         Yii::app()->loadHelper('database');
 
         if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveylocale', 'update')) {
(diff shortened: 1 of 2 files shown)


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1435/files#diff-05b8c46923ca0091db7ee406d881eb5bR34 192 character only … 😧

gabrieljenik:
Yes, that comes from the DB

Shnoulle:
All DB ?

gabrieljenik:
https://github.com/LimeSurvey/LimeSurvey/blob/071204150cb4a5affbc1532c9ca1936238b88467/application/helpers/update/updatedb_helper.php#L2220

Shnoulle:
Beurk … ;)

Shnoulle:
@gabrieljenik : why not adding it in rules ?

Then : more clean when import or use API.

Without your fix : it's done only for HTML GUI part, must be fixed too (not throw SQL error) when import and

gabrieljenik:
@Shnoulle The rule was already there.
` array('surveyls_policy_notice_label', 'length', 'min' => 0, 'max'=>192),`

Part of the issue was that when the text was too loong the saving failure was not reported back with an error message.

Shnoulle:
Thank you :)

application/controllers/admin/database.php
```diff
@@ -814,6 +814,8 @@ private function actionUpdateSurveyLocaleSettings($iSurveyID)
         $languagelist = $oSurvey->additionalLanguages;
         $languagelist[] = $oSurvey->language;
 
+        $saveSurvey = true;
```
olleharstedt, line 817: Please do not introduce needless state into the logic. This is spaghetti code. Instead, throw an exception if survey can't be saved.

application/controllers/admin/database.php
```diff
@@ -881,129 +883,137 @@ private function actionUpdateSurveyLocaleSettings($iSurveyID)
                     if (count($data) > 0) {
                         $oSurveyLanguageSetting = SurveyLanguageSetting::model()->findByPk(array('surveyls_survey_id'=>$iSurveyID, 'surveyls_language'=>$langname));
                         $oSurveyLanguageSetting->setAttributes($data);
-                        $oSurveyLanguageSetting->save(); // save the change to database
+                        // Save the change to database
+                        if (!$oSurveyLanguageSetting->save()) { 
+                            $saveSurvey = false;
+                            Yii::app()->setFlashMessage(Chtml::errorSummary($oSurveyLanguageSetting,Chtml::tag("p",array('class'=>'strong'),gT("Survey could not be updated, please fix the following error:"))), "error");
```
olleharstedt, line 889: Throw exception here, and add flash message in the `catch` block instead.

olleharstedt: Or just throw an exception and let Yii error handler show it.

olleharstedt: OK, I'm getting some understanding of your fix. database.php is _not_ an easy file to work with, with so big methods. I pushed a fix where I split it into two methods instead, and added a try-catch block around those methods.

gabrieljenik: Nice! Will take into account the advice. Agree: Sometime code reorgs are needed.

olleharstedt: Yeah, not an easy call as consultant. In-house I can do "bolder" changes. But you can always ask us.

gabrieljenik: Yes, when it come to fixes,I try to keep the scope tight

olleharstedt: That's exactly right, as small fix as possible (with any given interpretation of "possible" ^^ ), to make cherry-picking between branches easier.

PR #1436, opened 2020-06-18
Fixed issue #16316: Condition for list radio questions with question design "image_select" broken for questions on the same page

gabrieljenik:
Fixed issue #16316: Condition for list radio questions with question design "image_select" broken for questions on the same page

The image_select’s answer_row.twig for listradio was using {{ sCheckconditionFunction }} (which we believe it was carried over from multiple choice design) instead of the fixed “checkconditions” function

### DIFF
--- a/themes/question/image_select/survey/questions/answer/listradio/rows/answer_row.twig
+++ b/themes/question/image_select/survey/questions/answer/listradio/rows/answer_row.twig
@@ -13,7 +13,6 @@
  * @var code
  * @var answer
  * @var checkedState
- * @var sCheckconditionFunction
  * @var myfname
  * @var sValue
  **** Additional attributes:
@@ -43,9 +42,9 @@ question_template_attribute.horizontal_scroll   : {{question_template_attribute.
             id="answer{{ name}}{{ code }}"
             value="{{code}}"
             {{ checkedState }}
-            onclick='cancelBubbleThis(event); {{ sCheckconditionFunction }}'
+            onclick='cancelBubbleThis(event); checkconditions(this.value, this.name, this.type)'
         />
-        <label for="answer{{ name }}{{ code }}" class="imageselect-label" onclick='cancelBubbleThis(event); {{ sCheckconditionFunction }}'>
+        <label for="answer{{ name }}{{ code }}" class="imageselect-label" onclick='cancelBubbleThis(event);'>
 
          <img class="unforce-height" src="{{ answer }}" 
          style="{%if question_template_attribute.fix_width > 1%} 
--- a/themes/question/image_select/survey/questions/answer/listradio/rows/answer_row.twig
+++ b/themes/question/image_select/survey/questions/answer/listradio/rows/answer_row.twig
@@ -33,7 +33,7 @@ question_template_attribute.horizontal_scroll   : {{question_template_attribute.
 </p>#}
 
 <!-- answer_row -->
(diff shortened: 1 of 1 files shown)


Shnoulle:
No …

It's must trigger the originial event and nothing else (if needed)

See : https://github.com/LimeSurvey/LimeSurvey/blob/master/assets/packages/expressions/em_javascript.js#L210

Shnoulle:
My opinion : you can remove `cancelBubbleThis(event)`

gabrieljenik:
Remove the `cancelBubbleThis(event)` and trigger a `change` event. Right?

gabrieljenik:
That doesn't work. Those events are triggered for `radio-item` inputs, which is not the case. That's way we need an explicit call to `checkconditions`

Shnoulle:
Because there are another issue : **[it's not a checkbox-item](https://github.com/LimeSurvey/LimeSurvey/blob/d4d01fd0a7a50a432989daf4a5e129132f968345/themes/question/image_select/survey/questions/answer/listradio/rows/answer_row.twig#L36)** 

The issue is here, it's a button item or a radio-item

If really needed

https://gitlab.com/SondagesPro/QuestionTheme/EmailInput/-/blob/master/survey/questions/answer/shortfreetext/config.xml#L38

https://gitlab.com/SondagesPro/QuestionTheme/EmailInput/-/blob/master/survey/questions/answer/shortfreetext/assets/expression-email.js

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/pull/1439 

In 2.6, each time there are a onclick system : really more hard to upate/improve. Move to cleaner and external solution is allways best :)

I can't set you as reviewer … :(

gabrieljenik:
I know what you mean. The class is wrongly set on the parent tag. Will try that and let you know. Thanks.

Shnoulle:
> I know what you mean. The class is wrongly set on the parent tag. Will try that and let you know. 

Please : review : https://github.com/LimeSurvey/LimeSurvey/pull/1439

;)

gabrieljenik:
OK, I see, you have added fixes for "other" and "no answer" cases. 
If I got you right, we should merge and also put the inline script JS on external file inside the question theme, right?

Shnoulle:
For other script : copy/pasted from radio
For noanswer : i just move the script (radio have the onclik …)

themes/question/image_select/survey/questions/answer/listradio/rows/answer_row.twig
```diff
@@ -43,13 +42,12 @@ question_template_attribute.horizontal_scroll   : {{question_template_attribute.
...
             {{ checkedState }}
-            onclick='cancelBubbleThis(event); {{ sCheckconditionFunction }}'
         />
-        <label for="answer{{ name }}{{ code }}" class="imageselect-label" onclick='cancelBubbleThis(event); {{ sCheckconditionFunction }}'>
+        <label for="answer{{ name }}{{ code }}" class="imageselect-label">
 
-         <img class="unforce-height" src="{{ answer }}" 
-         style="{%if question_template_attribute.fix_width > 1%} 
-            width: {{question_template_attribute.fix_width}}px 
+         <img class="unforce-height" src="{{ answer }}"
```
olleharstedt, line 48: Just a note, if possible, avoid white-space changes, or do white-space changes in a separate PR. It will make reviewing the code easier. Thank you. :)

PR #1437, opened 2020-06-18
Fixed Issue #16103: Switching language without inline policy notice being ticked creates empty responses.

gabrieljenik:
Fixed Issue #16103: Switching language without inline policy notice being ticked creates empty responses.

Error Diagnose:
- After language was changed, LEM was set to dirty. 
- After showing the datapolicy error message, the process was killed, so the LEM was not regenerated.
- Still, the regular diplay process saved the dirty LEM on session. 
- Later when showing the first group, the InitDirtyStep was not called as it seems the LEM was dirty, but not TOO dirty. 
- So, a semi dirty LEM didn't have the surveyOptions['active'] attribute set to TRUE, which caused the saving process not to update the DB


Shnoulle:
Seems clear : need to initMove before checking security :+1:

PR #1440, opened 2020-06-22
Fixed issue #16375: Theme with valid config.xml (according to manual)…

Shnoulle:
Fixed issue #16375: Theme with valid config.xml (according to manual) can not be extendable
Dev: create if not set
Dev: allow last_update or lastUpdate only for 4.X
Dev: then 3.X still an LTS

### DIFF
--- a/application/models/TemplateManifest.php
+++ b/application/models/TemplateManifest.php
@@ -849,15 +849,27 @@ class TemplateManifest extends TemplateConfiguration
      */
     public static function changeDateInDOM($oNewManifest, $sDate = '')
     {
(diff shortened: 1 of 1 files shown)


olleharstedt:
How about someone tests this before merge?

Shnoulle:
I don't see a way to add a test

Shnoulle:
And 11 days between report, Marcel do a test because he have the issue too.

olleharstedt:
No, but Sushma, our QA, should test it before merge. :)

Shnoulle:
OK, for next time then :)

PR #1442, opened 2020-06-22
Fixed issue #15783: Print answers : all is empty

gabrieljenik:
Fixed issue #15783: Print answers : all is empty

SurveyDynamic::getPrintAnswersArray and SurveyDynamic::getQuestionArray() were not taking into account the i10n tables for Group, Question and Answer.

Added a $sLanguage optional parameter to getQuestionArray, because language is not an attribute of the question now. 

Set as optional as to keep backwards compatibiity.
If the parameter is not set, the survey’s main language is used.

There was a comment in line 973 of SurveyDynamic.php ( https://github.com/LimeSurvey/LimeSurvey/blob/778ad15d3206e95a815db171543521e6fe7a4793/application/models/SurveyDynamic.php#L973 ) referring to an issue that may got solved by this fix.

### DIFF
--- a/application/models/SurveyDynamic.php
+++ b/application/models/SurveyDynamic.php
@@ -780,7 +780,7 @@ class SurveyDynamic extends LSActiveRecord
      * @param boolean $getComment
      * @return array | boolean
      */
-    public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $subquestion = false, $getComment = false)
+    public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $subquestion = false, $getComment = false, $sLanguage = null)
     {
 
         $attributes = QuestionAttribute::model()->getQuestionAttributes($oQuestion->qid);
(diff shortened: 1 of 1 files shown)


application/models/SurveyDynamic.php
```diff
@@ -792,22 +792,27 @@ public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $su
             return false;
         }
 
+        // Use survey language is no language is specified
+        if (empty($sLanguage)) $sLanguage = $oQuestion->survey->language;
```
olleharstedt, line 796: Not PSR-12 compliant. Please always use `{}` around if-statements.

application/models/SurveyDynamic.php
```diff
@@ -780,7 +780,7 @@ protected function filterColumns(CDbCriteria $criteria)
      * @param boolean $getComment
      * @return array | boolean
      */
-    public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $subquestion = false, $getComment = false)
+    public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $subquestion = false, $getComment = false, $sLanguage = null)
```
olleharstedt, line 783: When adding a new argument, don't forget to update the docblock too! :partying_face:

application/models/SurveyDynamic.php
```diff
@@ -907,19 +917,19 @@ public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $su
...
-                        ? $oQuestion->parents->answers
+                    : ($oQuestion->parent != null
+                        ? $oQuestion->parent->answers
                         : []
                     )
             );
 
             foreach($aAnswers as $key=>$value){
-                $aAnswerText[$value['code']] = $value['answer'];
+                $aAnswerText[$value['code']] = Answer::model()->getAnswerFromCode($value->qid, $value->code, $sLanguage, $value->scale_id);
```
olleharstedt, line 932: Is this database query needed here, if we already did `$oQuestion->answers` above?

PR #1443, opened 2020-06-22
Fixed issue #15596: Problem saving question with question type list (options) design image_select

gabrieljenik:
Fixed issue #15596: Problem saving question with question type list (options) design image_select
Added explicit min attribute as to have front_end validation not to fail

-----

Image_select question theme has an attribute “fix_width”, of type “integer”, with a default of “-1” and no explicit minimum set in the config.xml file. 
As there was no minimum set, the default minimum used was 1. That made front end validation to fail. 

That was not noticeable.
- When fix_width field was displayed on screen, 
   - No big red message appeared, but a warning message below the attribute.
   - The saving wheel kept spinning. 
- When fix_width field was not on screen (collapsed) 
   - a Jquery error appeared (“field not focusable”)

It is to highlight the default attribute in the config.xml it’s not needed, because the attribute should either be set to a positive value or left empty. Removing the <default> tag from the config.xml also seems to solve the issue, although not sure if not having a default value for the attribute is valid.

I personally would remove the default attribute, as I don't like "-1" as an attribute value. But not sure if it the best way to go as to not break anything.

What do you think?

### DIFF
--- a/themes/question/image_select/survey/questions/answer/listradio/config.xml
+++ b/themes/question/image_select/survey/questions/answer/listradio/config.xml
@@ -80,6 +80,7 @@
             <category>Display</category>
             <sortorder>90</sortorder>
             <inputtype>integer</inputtype>
+            <min>-1</min>
             <default>-1</default>
             <help>Fix width of the images to this value. Leave empty to not change them.</help>
             <caption>Fix width</caption>
--- a/themes/question/image_select/survey/questions/answer/listradio/config.xml
+++ b/themes/question/image_select/survey/questions/answer/listradio/config.xml
@@ -80,8 +80,6 @@
             <category>Display</category>
             <sortorder>90</sortorder>
             <inputtype>integer</inputtype>
-            <min>-1</min>
-            <default>-1</default>
             <help>Fix width of the images to this value. Leave empty to not change them.</help>
             <caption>Fix width</caption>
         </attribute>


Shnoulle:
> I personally would remove the default attribute, as I don't like "-1" as an attribute value. But not sure if it the best way to go as to not break anything.
 
+1, yes an empty string is clear enough … 

> What do you think?

In 2.6 and 3.X : default value are not saved in database. Then if 4.X are good enough on this : -1 are not saved.

Just check : my opinion : 
1. Move default to empty string (mean not set or default) (no default must set "")
2. Since -1 are never saved because it was invalid : i think you can remove the min = -1 and set min to 1 (or 0 ?)

PS : a good system must be show the error, even if need to `collapse` show the part with error. But again : some dev didn't  know standard …

gabrieljenik:
> +1, yes an empty string is clear enough …
If we make an empty string `>default></default>`, advanced_settings_view render crashes.

> Just check : my opinion :
> Move default to empty string (mean not set or default) (no default must set "")
advanced_settings_view render crashes.
> Since -1 are never saved because it was invalid : i think you can remove the min = -1 and set min to 1 (or 0 ?)
If we leave -1, the min is needed as to avoid the frontend validation failure (as it is the initial solution pushed). If we remove default, I agree.

Shnoulle:
> If we make an empty string >default></default>, advanced_settings_view render crashes.

[Not set must be empty string (default)](https://github.com/LimeSurvey/LimeSurvey/blob/e55a12f35b1f62032d0b79a79ede7d727b708681/application/models/QuestionAttribute.php#L404-L421) . It was in 3.X. 

And here: it's another issue :)

gabrieljenik:
ok, so how do we follow this?
Scenario A
1 - Create a ticket for crash when default doesn't allow empty string.
2 - After that is done, we make the fix_width attribute to default to empty

Or?:
Scenario B
1 - We make the fix_width attribute to default to empty, removing the whole `<default>` line.
2 - Create a ticket for crash when default doesn't allow empty string.
3 - We can think of adding the `<deafult></default>` line back (if desired)

Shnoulle:
B for me (if it work)

Else : min can be 0 ? fixed_with to 0 ???? Why not min to 1?

gabrieljenik:
OK, will go for B then.

gabrieljenik:
https://bugs.limesurvey.org/view.php?id=16413

https://bugs.limesurvey.org/view.php?id=16414

themes/question/image_select/survey/questions/answer/listradio/config.xml
```diff
@@ -80,7 +80,6 @@
             <category>Display</category>
             <sortorder>90</sortorder>
             <inputtype>integer</inputtype>
-            <default>-1</default>
```
olleharstedt, line 83: Not a fan of this change, since for a code reader it will look like default tag is forgotten. Will it work with `<default></default>`?

Shnoulle: No : it never work with `<default></default>` , issue was reported

olleharstedt: _Should_ it work?

Shnoulle: I think : it should work, not necesssary if we use `QuestionAttribute::getDefaultSettings();` but someone remove the usage of this function done to give the default in 4 … previously

olleharstedt: Sigh.

PR #1444, opened 2020-06-23
Fixed issue #16421 - surveyGetXMLData() not producing valid XML

BertHankes:
Fixed issue #16421 - surveyGetXMLData() not producing valid XML


c-schmitz:
This was fixed in a877f78df366fd5ff1546f41389e3a775eb0a83c

PR #1445, opened 2020-06-23
Fixed issue #16415: Can't upload files with Google Chrome

gabrieljenik:
Fixed issue #16415: Can't upload files with Google Chrome

From my understanding, this issue happens because of some false positive load events of the iframe.
I mean some load events are triggered for the iframe (example for the about:blank doc) while they should be ignored.

There are already some filters, so I am adding a new one.

This wasn't enough. Somehow it worked when we tested. 

This is a WIP.
We are rolling back prior bug change (the kne from couple of weeks ago) and reviewing that solution also.

### DIFF
--- a/assets/scripts/ajaxupload.js
+++ b/assets/scripts/ajaxupload.js
@@ -509,7 +509,7 @@
             // "This page contains both secure and nonsecure items"
             // Anyway, it doesn't do any harm.
             iframe.setAttribute('id', id);
-            iframe.setAttribute('srcdoc', this._settings.action);
+            // iframe.setAttribute('srcdoc', this._settings.action);
 
             iframe.style.display = 'none';
             document.body.appendChild(iframe);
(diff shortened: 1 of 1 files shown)


assets/scripts/ajaxupload.js
```diff
@@ -509,7 +509,7 @@
             // "This page contains both secure and nonsecure items"
             // Anyway, it doesn't do any harm.
             iframe.setAttribute('id', id);
-            iframe.setAttribute('srcdoc', this._settings.action);
+            // iframe.setAttribute('srcdoc', this._settings.action);
```
olleharstedt, line 512: You can remove this line, no? If you only comment it, the reader won't know why it's commented (without checking git blame).

Shnoulle: Maybe comments temporary ? The issue was really strange.

But clearlky : all upload question type must be rewritten to use real ajax

gabrieljenik: I am following up with patrick on why that line was added. There are some stackoverflow posts which recommend that. Maybe the chrome change that produced that was rolled back after?

We need to be able to reproduce the old bug (the one from couple of weeks ago).

I believe, we need to retest the initial bug, or said in other words identify the key browsers (and specific possible engine variations) where to test this.

gabrieljenik: > You can remove this line, no?

Yes, was a work in progress. I asked Sushma to help me test it

olleharstedt: :+1:

olleharstedt:
Thanks, Gabriel and Sushma!

PR #1447, opened 2020-06-25
New Feature #16429: Added config variable to implicitly declare MyISAM engine when creating tables in MySQL

jenseeckhout:
Dev:
Created a new MySQL config variable to enforce specification of the MyISAM database engine when creating a new table.
This is necessary to facilitate MySQL replication in the event that you have large survey tables (i.e. lots of columns). Reason for this is that MySQL replication ignores the default engine variable. Hence, if your replication is set to use InnoDB by default, replication will crash when trying to create the table (see https://dev.mysql.com/doc/refman/8.0/en/column-count-limit.html). By implicitly adding the ENGINE=MyISAM to the CREATE TABLE statement, this will be picked up by replication correctly.
See Mantis: https://bugs.limesurvey.org/view.php?id=16429

### DIFF
--- a/application/core/db/MysqlSchema.php
+++ b/application/core/db/MysqlSchema.php
@@ -18,6 +18,10 @@ class MysqlSchema extends CMysqlSchema
             $options = 'DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci';
         }
 
+        if (Yii::app()->getConfig('forceMyISAM') === true) {
+            $options .= ' ENGINE=MyISAM';
+        }
+
         // Below copied from parent.
         $cols = array();
         foreach ($columns as $name => $type) {
--- a/application/core/db/MysqlSchema.php
+++ b/application/core/db/MysqlSchema.php
@@ -15,11 +15,7 @@ class MysqlSchema extends CMysqlSchema
     public function createTable($table, $columns, $options = null)
     {
         if (empty($options)) {
-            $options = 'DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci';
-        }
-
-        if (Yii::app()->getConfig('forceMyISAM') === true) {
-            $options .= ' ENGINE=MyISAM';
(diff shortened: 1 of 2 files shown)


Shnoulle:
A related issue **must** be reported in mantis

jenseeckhout:
> A related issue **must** be reported in mantis

Updated PR description with link to the issue.

application/core/db/MysqlSchema.php
```diff
@@ -18,6 +18,10 @@ public function createTable($table, $columns, $options = null)
             $options = 'DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci';
         }
 
+        if (Yii::app()->getConfig('forceMyISAM') === true) {
```
Shnoulle, line 21: Why not
````
If (App()->db->driverName == 'mysql') {
     $options .= ' ENGINE='.App()->getConfig('mysqlEngine');
}
````

Or better : 
````
        if (empty($options)) {
            $options = 'ENGINE='.App()->getConfig('mysqlEngine').' DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci';
        }
````
in : https://github.com/LimeSurvey/LimeSurvey/blob/d905cdce4ad13e401a2ceb4ac50b13144b799431/application/core/db/MysqlSchema.php#L18

Then : we have the same behaviour in 3.X (and before) and 4.X

jenseeckhout: Good point. This allows for better backwards compatibility. Thanks, I updated the PR.

Shnoulle:
:+1: exactly the same (by default) than 3.X.

jenseeckhout:
Is anything else required before this can be merged?

Shnoulle:
@olleharstedt ?

olleharstedt:
Remind me on Monday... No, wait, the week after. Germany will change its taxes 1 July, I have to deal with that first.

PR #1448, opened 2020-06-26
Fixed issue #15985: print answers: list with comments, comments not printed

gabrieljenik:
This  was implemented on getFullResponseTable(), LSv2.6. 
But seems to got lost in the transitio to LSv3 when moving to SurveyDynamic::getPrintAnswersArray()

Implementing comments for question type 'O'
Rename getComment parameter.

### DIFF
--- a/application/models/SurveyDynamic.php
+++ b/application/models/SurveyDynamic.php
@@ -971,6 +971,12 @@ class SurveyDynamic extends LSActiveRecord
             }
         }
 
+        // If trying to retrieve main question ($getComment = false), retrieve comment in a new attribute
+        // Check if $getComment = false to avoid endless recursivity
+        if ($oQuestion->type=='O'&&!$getComment) {
+            $aQuestionAttributes['comment'] = $this->getQuestionArray($oQuestion, $oResponses, $bHonorConditions, true, true);
+        }
+
         return $aQuestionAttributes;
     }
 
--- a/application/models/SurveyDynamic.php
+++ b/application/models/SurveyDynamic.php
@@ -849,7 +849,7 @@ class SurveyDynamic extends LSActiveRecord
         }
 
 
-        if ($getCommentOnly === true) {
+        if ($getCommentOnly) {
             $fieldname .= 'comment';
         }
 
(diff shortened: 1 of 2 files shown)


application/models/SurveyDynamic.php
```diff
@@ -964,13 +964,19 @@ public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $su
...
+
         if ($oQuestion->type=='N' || ($oQuestion->parent_qid != 0 && $oQuestion->parents['type'] === "K")) {
             if (strpos($aQuestionAttributes['answervalue'], ".") !== false) { // Remove last 0 and last . ALWAYS (see \SurveyObj\getShortAnswer)
                 $aQuestionAttributes['answervalue'] = rtrim(rtrim($aQuestionAttributes['answervalue'], "0"), ".");
             }
         }
 
+        // If trying to retrieve main question ($getCommentOnly = false), retrieve comment in a new attribute
+        // Check if $getCommentOnly = false to avoid endless recursivity
+        if ($oQuestion->type=='O'&&!$getCommentOnly) {
```
olleharstedt, line 976: PSR-12 recommends (or implies?) space between `==`. It should be:

```php
if ($oQuestion->type == 'O' && !$getCommentOnly) {
```

Please adjust.

application/models/SurveyDynamic.php
```diff
@@ -849,7 +849,7 @@ public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $su
         }
 
 
-        if ($getComment === true) {
+        if ($getCommentOnly === true) {
```
olleharstedt, line 852: `=== true` is not needed, should be

```php
if ($getCommentOnly) {
```

application/models/SurveyDynamic.php
```diff
@@ -770,10 +770,10 @@ protected function filterColumns(CDbCriteria $criteria)
      * @param SurveyDynamic $oResponses
      * @param boolean $bHonorConditions
      * @param boolean $subquestion
-     * @param boolean $getComment
+     * @param boolean $getCommentOnly If should only returns the "comments" or "other" response.
```
olleharstedt, line 773: +1 for proper explanation of the argument :)

application/models/SurveyDynamic.php
gabrieljenik: Done

application/models/SurveyDynamic.php
gabrieljenik: Fixed

PR #1449, opened 2020-06-26
Fixed issue #15940: Export as printable : PHP Notice wuth debug

gabrieljenik:
When using global template on a survey, the template name saved on DB is "inherit".
That was taken literal sometimes and caused errors, as there is no template called "inherit"

"inherit" template is now converted to the proper template name, set at global level.

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -836,6 +836,33 @@ class Survey extends LSActiveRecord
     }
 
     /**
+     * Returns the name of the template to be used for the survey.
+     * It resolves inheritance from group and from default settings.
+     * 
+     * @return string
+     * 
+     * @todo:  Cache this on a private attribute?
+     */
+    public function getTemplateEffectiveName()
+    {
+        // Fetch template name from model
+        // This was already filtered on afterFind, so if the one at load time is not valid, will be replaced by default one
+        // If it is "inherit", means it will inherit from group, so we will replace it.
+        $sTemplateName = $this->template;
+
+        // if it is "inherit", get template name form group
+        if ($sTemplateName == 'inherit') {            
+            if (!empty($this->oOptions->template)) {
+                $sTemplateName = $this->oOptions->template;
+            } else {
+                throw new CException("Unable to get a template name from group for survey {$this->sid}");
+            }
+        }  
+        
+        return $sTemplateName;
+    }
+
+    /**
      * Get surveymenu configuration
      * This will be made bigger in future releases, but right now it only collects the default menu-entries
      */
--- a/application/models/Template.php
+++ b/application/models/Template.php
@@ -285,7 +285,9 @@ class Template extends LSActiveRecord
     public static function getTemplatePath($sTemplateName = "")
     {
         // Make sure template name is valid
-        $sTemplateName = self::templateNameFilter($sTemplateName);
+        if (!self::checkIfTemplateExists($sTemplateName)) {
+            throw new \CException("Invalid {$sTemplateName} template directory");
+        }
 
         static $aTemplatePath = array();
         if (isset($aTemplatePath[$sTemplateName])) {
@@ -293,9 +295,8 @@ class Template extends LSActiveRecord
         }
 
         $oTemplate = self::model()->findByPk($sTemplateName);
-        if (empty($oTemplate))
-        {
-            throw new \Exception("Survey theme {$sTemplateName} not found.", 1);
+        if (empty($oTemplate)) {
+            throw new \CException("Survey theme {$sTemplateName} not found.", 1);
         }
 
         if (self::isStandardTemplate($sTemplateName)) {
@@ -375,7 +376,9 @@ class Template extends LSActiveRecord
     public static function getTemplateURL($sTemplateName = "")
     {
         // Make sure template name is valid
-        $sTemplateName = self::templateNameFilter($sTemplateName);
+        if (!self::checkIfTemplateExists($sTemplateName)) {
+            throw new \CException("Invalid {$sTemplateName} template directory");
+        }
 
         static $aTemplateUrl = array();
         if (isset($aTemplateUrl[$sTemplateName])) {
--- a/application/models/Template.php
+++ b/application/models/Template.php
@@ -284,6 +284,9 @@ class Template extends LSActiveRecord
      */
     public static function getTemplatePath($sTemplateName = "")
     {
+        // Make sure template name is valid
+        $sTemplateName = self::templateNameFilter($sTemplateName);
+
         static $aTemplatePath = array();
         if (isset($aTemplatePath[$sTemplateName])) {
             return $aTemplatePath[$sTemplateName];
@@ -367,6 +370,9 @@ class Template extends LSActiveRecord
      */
     public static function getTemplateURL($sTemplateName = "")
     {
+        // Make sure template name is valid
+        $sTemplateName = self::templateNameFilter($sTemplateName);
+
         static $aTemplateUrl = array();
         if (isset($aTemplateUrl[$sTemplateName])) {
             return $aTemplateUrl[$sTemplateName];
--- a/application/models/Template.php
+++ b/application/models/Template.php
@@ -293,6 +293,10 @@ class Template extends LSActiveRecord
         }
 
         $oTemplate = self::model()->findByPk($sTemplateName);
+        if (empty($oTemplate))
+        {
+            throw new \Exception("Survey theme {$sTemplateName} not found.", 1);
+        }
 
         if (self::isStandardTemplate($sTemplateName)) {
             return $aTemplatePath[$sTemplateName] = Yii::app()->getConfig("standardthemerootdir").DIRECTORY_SEPARATOR.$oTemplate->folder;
--- a/application/controllers/admin/export.php
+++ b/application/controllers/admin/export.php
@@ -1260,8 +1260,8 @@ class export extends Survey_Common_Action
     private function _exportPrintableHtmls($iSurveyID, $readFile = true)
     {
         $oSurvey = Survey::model()->findByPk($iSurveyID);
-        $assetsDir = substr(Template::getTemplateURL($oSurvey->template), 1);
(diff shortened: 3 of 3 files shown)


Shnoulle:
Question : what happen if you name a template `inherit` 🤔 another issue to report ?

Shnoulle:
In  my opinion : there are an issue before … (but this one is a fix) $oSurvey-template MUST never be set to an inexisting survey. `getTemplatePath` and `getTemplateURL` must not be called with an invalid template :)

gabrieljenik:
> Question : what happen if you name a template `inherit`  another issue to report ?

Yes, noted that on the bug

gabrieljenik:
> In my opinion : there are an issue before … (but this one is a fix) $oSurvey-template MUST never be set to an inexisting survey. `getTemplatePath` and `getTemplateURL` must not be called with an invalid template :)

Will fix that. We should throw an exception, right?

Shnoulle:
> > In my opinion : there are an issue before … (but this one is a fix) $oSurvey-template MUST never be set to an inexisting survey. `getTemplatePath` and `getTemplateURL` must not be called with an invalid template :)
> 
> Will fix that. We should throw an exception, right?

If you find the other issue, but don't take to more time on it :)

Yes : think best is to throw clean error if template is invalid. But it's **my** opinion.

I think too : if we get there and template is inavlid : maybe some css/js was not loaded ( theme package)

gabrieljenik:
> If you find the other issue, but don't take to more time on it :)

Here you mean?
https://github.com/LimeSurvey/LimeSurvey/pull/1449/files#diff-66fb738711f2db21da8c5b6029ad0794R295
If oTemaplte is empty, throw error.

Error message could be "Survey template not found"?

Shnoulle:
> > If you find the other issue, but don't take to more time on it :)
> 
> Here you mean?
> https://github.com/LimeSurvey/LimeSurvey/pull/1449/files#diff-66fb738711f2db21da8c5b6029ad0794R295
> If oTemaplte is empty, throw error.
> 
> Error message could be "Survey template not found"?

Not exactly : i mean : template must be checked before this function. Then : why the template is not OK for print version.

gabrieljenik:
> Not exactly : i mean : template must be checked before this function. Then : why the template is not OK for print version.

Oh, I agree. Template must be checked. Actually the template name comes out from the DB. So it is kind of checked. Also templateNameFilter() does lot of checkings. Still, I suggest the "Survey template not found" exception as a sanity check (I think they are called like that), just in case. Don't expect to be thrown, but better to ave an exception than a dump.

Where did you picture an extra validation?

Shnoulle:
> Actually the template name comes out from the DB.

in Survey->find ? Then it must be already fixed … else all is broken.

> Don't expect to be thrown, but better to ave an exception than a dump.

Throw an exception id really the best solution :)

> Where did you picture an extra validation?

I don't want extra validation : only Survey->find must validate it 

Currently : inherit is disable of fixing … 

https://github.com/LimeSurvey/LimeSurvey/blob/18f3377f2e0eb94ba4bc6ee29438bc8b4d5ba0df/application/models/Survey.php#L565-L567

And this in an issue for Print …

gabrieljenik:
> And this in an issue for Print …

Agree this is what is making things break.
Maybe we can have a method on Survey model, like "GetEffectiveTemplate()" or "GetEffectiveTemplateName()" ?

Shnoulle:
I don't know how it done with Survey, but maybe this explan why sometimes it broke at 1st call, but not at second call (when delete or import some specific survey)

gabrieljenik:
>Maybe we can have a method on Survey model, like "GetEffectiveTemplate()" or "GetEffectiveTemplateName()" ?

As to not loose track fo the conversation, @olleharstedt @Shnoulle Should I do this?

olleharstedt:
What happens if I create a theme with name "inherit"?

olleharstedt:
What's wrong with this fix, exactly?

gabrieljenik:
> What's wrong with this fix, exactly?

Nothing is wrong really. Fix should work good.
I was saying that:
- If the survey uses the global default theme, "inherit" is saved on the DB
- If that's the case, the survey model doesn't validate the theme when laoding a survey record.
- When using the global theme, as to be able to get the effective theme to be used, I was thinking about adding a method.
But I realize think it is out of the scope...

gabrieljenik:
> > If you find the other issue, but don't take to more time on it :)
> 
> Here you mean?
> https://github.com/LimeSurvey/LimeSurvey/pull/1449/files#diff-66fb738711f2db21da8c5b6029ad0794R295
> If oTemaplte is empty, throw error.
> 
> Error message could be "Survey template not found"?

Done

Shnoulle:
Still an isue in this fix : if template is set as inherit (from Survey group) : it's default template used.

I think this commit didn't fix the issue

Shnoulle:
Survey use `$oTemplate = Template::model()->getInstance('', $surveyid);`

printable version **MUST** use `$oTemplate = Template::model()->getInstance('', $surveyid);`

Shnoulle:
Your commit didn't take inherited template , only default template

https://github.com/LimeSurvey/LimeSurvey/pull/1455

application/models/Template.php
```diff
@@ -293,6 +293,10 @@ public static function getTemplatePath($sTemplateName = "")
         }
 
         $oTemplate = self::model()->findByPk($sTemplateName);
+        if (empty($oTemplate))
```
olleharstedt, line 296: https://www.php-fig.org/psr/psr-12/

Should be

```php
if (empty($oTemplate)) {
```

`{` on the same line. You should include PSR-12 checking in your editor (will show LOTS of errors, but make sure all new code follows it, please).

application/models/Template.php
```diff
@@ -293,6 +293,10 @@ public static function getTemplatePath($sTemplateName = "")
...
+        {
+            throw new \Exception("Survey theme {$sTemplateName} not found.", 1);
```
olleharstedt, line 298: Not required, but if you want to avoid `\`, you can add `use Exception;` at top of file (if there's a namespace, not sure there is right now).

olleharstedt: Something to discuss is which exception messages should be translated with gT().

Shnoulle:
@olleharstedt : did you read my comment ? If we want to use template we NEED to get inherited template. Else : it's another issue … 

https://github.com/LimeSurvey/LimeSurvey/pull/1455

application/models/Template.php
Shnoulle: Must use CException since we use Yii (and there are no namespace)

olleharstedt: OK, sure.

gabrieljenik:
Added method/attribute on survey model for getting the effective template name (handling survey group inheritance). Used that method/attribute while exporting printable surveys.

Added some existence checkings on getTemplateUrl() and getTemplatePath()

gabrieljenik:
> @olleharstedt : did you read my comment ? If we want to use template we NEED to get inherited template. Else : it's another issue …
> 
> #1455

Please review my latest commit. I think it goes in the same direction of your PR

application/models/Template.php
gabrieljenik: Fixed

application/models/Template.php
gabrieljenik: Fixed

application/models/Survey.php
```diff
@@ -835,6 +835,33 @@ public function getSurveyTemplateConfiguration()
...
+        // This was already filtered on afterFind, so if the one at load time is not valid, will be replaced by default one
+        // If it is "inherit", means it will inherit from group, so we will replace it.
+        $sTemplateName = $this->template;
+
+        // if it is "inherit", get template name form group
+        if ($sTemplateName == 'inherit') {            
+            if (!empty($this->oOptions->template)) {
+                $sTemplateName = $this->oOptions->template;
+            } else {
+                throw new CException("Unable to get a template name from group for survey {$this->sid}");
```
Shnoulle, line 858: I like the new function :+1: .

But here : i don't know how it work for parent : parent can not be inherit too ? Then get it for "global" ?

olleharstedt:
Looks good to me! Denis, you happy?

Shnoulle:
I don't really know what happen fro group , i don't know if there are a way where group can return inherit.
I don't work on inherit system : then don't really know if we can have a more global function.

But think it can be checked after.

sushmanadendlalimesurvey:
Tested by pulling the PR. Able to successfully download and view the survey in debug mode

olleharstedt:
Gah, now I merged with the typo D:

application/models/Survey.php
gabrieljenik: I actually tooke the code from digging into your example.
The code is from here: https://github.com/gabrieljenik/LimeSurvey/blob/65061857c0ae8514300746b481285e0004a7d576/application/models/TemplateConfiguration.php#L259

So, I think survey group inheritance is there. 
Now that you mention it, not sure how it would work with multiple survey groups inherted.
Is that possible? Does the TemplateConfiguration code supports it?

@Shnoulle What I don't like that much your intial approach (sorry :) ) is that there is a lot of processing (template instancing) just for getting the template name.

gabrieljenik:
> I don't really know what happen fro group , i don't know if there are a way where group can return inherit.
> I don't work on inherit system : then don't really know if we can have a more global function.
> 
> But think it can be checked after.

Agree: We can make the new method thicker if needed after.

application/models/Survey.php
Shnoulle: Thanks : never look at getInstanceFromSurveyId :)

Shnoulle:
> > I don't really know what happen fro group , i don't know if there are a way where group can return inherit.
> > I don't work on inherit system : then don't really know if we can have a more global function.
> > But think it can be checked after.
> 
> Agree: We can make the new method thicker if needed after.

And since : you create a clean function : really more easy to fix :)

gabrieljenik:
> And since : you create a clean function : really more easy to fix :)

Thanks for the comment :)

PR #1452, opened 2020-06-26
Fixed issue #16434: Special characters shown as HTML entities in sidebar

gabrieljenik:
Ok, so special characters should be shown as they are and not as html entities.
What about special characters which form tags? Should they be shown? Isn't that a security issue?

My call for fixing this was to use strip_tags.
So, sending special characters to look good. But tags are not sent.

Text is flatten. After that. special chars are sent to the sidebar. Tags, which could be formed by special chars re-coded, are not sent.

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -575,7 +575,8 @@ class SurveyAdmin extends Survey_Common_Action
                         $questionText = isset($question->questionl10ns[$baselang])
                             ? $question->questionl10ns[$baselang]->question
                             : '';
-                        $curQuestion['question_flat'] = viewHelper::flatEllipsizeText($questionText, true);
+                        $decodedQuestionText = htmlspecialchars_decode(viewHelper::flatEllipsizeText($questionText, true));
+                        $curQuestion['question_flat'] = strip_tags($decodedQuestionText);
                         $curGroup['questions'][] = $curQuestion;
                     }
 


Shnoulle:
Here : there are a double encoding, no ?

gabrieljenik:
> Here : there are a double encoding, no ?

What do you mean? sorry, couldn't get it. 
I count 1 encoding only, the one done here https://github.com/LimeSurvey/LimeSurvey/pull/1452/files#diff-ae6ef646cdcd6b65eca67ba2a3b15a27R578
What would be the 2nd one?

Shnoulle:
Here there are double encoding since you need to decode …

If there are only one : you don't have to decode … maybe in vue.js ? 

But more `viewHelper::flatEllipsizeText` already (must) remove tag , then why this &lt; is here.

One encode : `&lt;` show as '>' in HTML
then  here : double encode (and why this tag was not removed …)

Shnoulle:
See sample , checked with 3 : 1 > 2 show with &gt; , but real tag are not shown as tag.

Her e: it's real tag or false tag ?

Shnoulle:
Since `question.question_flat` is almready filtered and encoded :  
https://github.com/LimeSurvey/LimeSurvey/blob/18f3377f2e0eb94ba4bc6ee29438bc8b4d5ba0df/assets/packages/adminsidepanel/src/components/subcomponents/_questionsgroups.vue#L378

Must send it raw : 
`[{{question.title}}] &rsaquo; {{ question.question_flat | raw }} `

gabrieljenik:
`{{ question.question_flat | raw }}`
Nice, will add that to the PR

Shnoulle:
I didn't check : but think it's the best solution. 

If you can fix it for 3.X too : it's cool

gabrieljenik:
> `{{ question.question_flat | raw }}`
> Nice, will add that to the PR

Sorry, we were confused because of lanugage constructs.
`raw` on vue doesn't seem to be standard

Still, made me think, how should special characters travel from server to front end. I believe they should travel as text. 
Then Vue takes care of escaping.

That's how it is set now.
So wouldn't change the current PR.

If it is fine, I will wrap it there and then open a new ticket for v3

Shnoulle:
>   raw on vue doesn't seem to be standard

Yes sorry : https://vuejs.org/v2/guide/syntax.html#Raw-HTML 

All text are already filtered by `flatEllipsizeText` , the name of the variable are `question_flat` then no reason to escape it again …

Here : the filtered text contain `&gt;` and `&` see json
Capture d’écran du 2020-06-30 08-14-52

Shnoulle:
@gabrieljenik : can you check if you build https://github.com/LimeSurvey/LimeSurvey/pull/1456

My opinion : it's the way to do. Or if you want to send it decoded (and to leave encode to bue) remove strip_tags

thedirtypanda:
So whats the problem here?

gabrieljenik:
> @gabrieljenik : can you check if you build #1456

I will try to build it.

> My opinion : it's the way to do. Or if you want to send it decoded (and to leave encode to bue) remove strip_tags

What do you mean by "it is the way to do"?
From my point of view, the question still remains: Why do we need to send html-encoded text if it is supposed to be "flat" (as the attribbute name is "question_flat")?

I agree on removing strip_tags. It is not needed after all, as that flat text will probable be escaped.

application/controllers/admin/surveyadmin.php
```diff
@@ -575,7 +575,8 @@ public function getAjaxQuestionGroupArray($surveyid)
                         $questionText = isset($question->questionl10ns[$baselang])
                             ? $question->questionl10ns[$baselang]->question
                             : '';
-                        $curQuestion['question_flat'] = viewHelper::flatEllipsizeText($questionText, true);
+                        $decodedQuestionText = htmlspecialchars_decode(viewHelper::flatEllipsizeText($questionText, true));
+                        $curQuestion['question_flat'] = strip_tags($decodedQuestionText);
```
Shnoulle, line 579: Since vue.js already encoded, you can just decode 
and put a comment
` // vue.js encpode and no filter to get raw text, then decode it`

gabrieljenik:
At last, I would just follow any direction. I think we are spending too much time on this and maybe the subject doesn't worth it (in a good sense I mean it) :)

application/controllers/admin/surveyadmin.php
Shnoulle: No need to remove "false" tag put by user.

gabrieljenik: ok, so only missing would be the comment, right?

gabrieljenik:
> 1. No need to remove `&gt;p&lt;` is user put it.

Not sure I agree. User typed `<p>` and that is the flat text. `&gt;p&lt;`  is the encoded one which is saved.

As to wrap this, I would say to make this the last exchange. What do you think?
- If you agree with my argument, I will remove the strip_tags and put some comment. 
- If not, I will follow your previous directions: 
  + Send text html-encoded (as it was before) 
  + Make vue to show it raw.

What do you think?

Shnoulle:
> Why do we need to send html-encoded text if it is supposed to be "flat" 

Because `&gt;` is flat … an `&gt;p&lt;` too 

It's what i didn't understand first on your

sushmanadendlalimesurvey:
Tested the issue by checking out to the branch, Special characters are not diplayed in question code.
Working as expected

olleharstedt:
@Shnoulle Feel free to open any related bugs to this issue.

PR #1453, opened 2020-06-26
Fixed issue #16394: If text too long, data protection statement text not stored

gabrieljenik:
Added max length and type-counter to data sec label input.
TypeCounter.vue was missing. Cloned it and adapted it for showing OK on empty string


Shnoulle:
Did we really need type.counter ?

thedirtypanda:
This was the last feature which Markus implemented. Its good and you will get more information, about how many chars you can type in etc. In my point of view, it has something to do with user expirience. Our UX Designer will decide in the future, if we will stick to that or not.

Shnoulle:
> This was the last feature which Markus implemented. Its good and you will get more information, about how many chars you can type in etc. In my point of view, it has something to do with user expirience. Our UX Designer will decide in the future, if we will stick to that or not. 

OK, why not, but before adding information : all system must work :)

PR #1454, opened 2020-06-29
fix composer.json license format to match composer standard

TonisOrmisson:
the license tag does not match the [composer standard](https://getcomposer.org/doc/04-schema.md#license)

not having standard licence causes eg crash for packagist automation.


c-schmitz:
While I commend to adhere to standards it should be "GPL-2.0-or-later"'

TonisOrmisson:
@c-schmitz thanks, done!

PR #1459, opened 2020-06-30
Fixed issue #16430: Testing a question will not be possible if admin participated in the survey (cookies must be activated from survey settings)

gabrieljenik:
Avoidng validation if we are on preview mode.


sushmanadendlalimesurvey:
Tested the issue by checking out to branch, followed the steps mentioned above was able to view the question successfully after clicking on Preview question button from Question Editor and also in question overview. Working as expected

PR #1461, opened 2020-07-01
Fix table name regex in buildXMLFromQuery

miniBill:
When the table name gets quoted (on MySQL) the current regex doesn't match inside the query, which makes the following code malfunction and prevents survey copying


miniBill:
Should I create a mantis issue? It's literally a 4 characters fix...

olleharstedt:
This should already be OK by giving explicit table name "groups".

miniBill:
On our installation it didn't work with the current code.
MySQL: 5.6.36
PHP: 7.1.32

olleharstedt:
on which version?

Shnoulle:
Maybe time to report issue with step by step way to reproduce ?

miniBill:
My company decided to use LS 3 for the time being

PR #1465, opened 2020-07-01
Fixed issue #15903: Sending email to all group users (copy is sent in cc for every user)

gabrieljenik:
Copy email is not sent as CC but as adding and extra sender to the list of users.

### DIFF
--- a/application/models/UserGroup.php
+++ b/application/models/UserGroup.php
@@ -476,9 +476,12 @@ class UserGroup extends LSActiveRecord
         $oUserFrom = User::model()->findByPk(Yii::app()->session['loginID']);
         $fromName = empty($oUserFrom->full_name) ? $oUserFrom->users_name : $oUserFrom->full_name;
         $mailer->setFrom($oUserFrom->email, $fromName);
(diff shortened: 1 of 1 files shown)


cdorin93:
looks great, thanks!

application/models/UserGroup.php
```diff
@@ -476,9 +476,12 @@ public function sendUserEmails($ugid, $subject, $body, $copy){
         $oUserFrom = User::model()->findByPk(Yii::app()->session['loginID']);
         $fromName = empty($oUserFrom->full_name) ? $oUserFrom->users_name : $oUserFrom->full_name;
         $mailer->setFrom($oUserFrom->email, $fromName);
-        //todo: this will send user in cc amountNumberOfUsers (spaming the email of user ...)
+        
+        // Add the sender to the list of users in order to receive a copy
         if ($copy == 1) {
-            $mailer->addCC($oUserFrom->email, $fromName);
```
Shnoulle, line 481: Send me a copy : add me to CC : It's not the way to do ;)

sushmanadendlalimesurvey:
Tested by  Dorin. Working as expected

PR #1470, opened 2020-07-08
Real name verification code

xgdz:
I am now carrying out secondary development. May I ask where is the relevant code of the verification code of the questionnaire?


Shnoulle:
Pull request is not done to ask dev support.

PR #1471, opened 2020-07-08
Dev: TypeHints in QuestionEditorController

Shnoulle:
Dev: related #16470

### DIFF
--- a/application/controllers/QuestionEditorController.php
+++ b/application/controllers/QuestionEditorController.php
@@ -28,7 +28,7 @@ class QuestionEditorController extends LSBaseController
      * @param string $view
      * @return bool
      */
-    protected function beforeRender($view)
+    protected function beforeRender(string $view)
     {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Capture d’écran du 2020-07-08 15-31-18
Capture d’écran du 2020-07-08 15-52-53

olleharstedt:
This breaks travis.

Shnoulle:
Yep :) i check.

Still think it's a good idea, but i create another pull request for https://bugs.limesurvey.org/view.php?id=16469 

Currently : permssion access in LimeSurvey 4 is broken

olleharstedt:
> Currently : permssion access in LimeSurvey 4 is broken

It's always been broken.

Shnoulle:
OK : 
Capture d’écran du 2020-07-09 14-04-05

Since call didn't check … i close

PR #1472, opened 2020-07-08
Fixed issue #16471: Fix renderKeys not using baseURL

jatoko:
Fixed issue #[16471](https://bugs.limesurvey.org/view.php?id=16471):


cdorin93:
Hey @jatoko , have you had time to take a look at the comments from : https://bugs.limesurvey.org/view.php?id=16471 ?

c-schmitz:
After reviewing this I don't think that the fix is correct, but merely fixes a symptom, not the underlying issue.
Therefore, I am closing this PR.

PR #1473, opened 2020-07-08
Fixed issue #16204: Arrays (all types) not displayed correctly during survey-taking (mobile phone)

gabrieljenik:
Created a new DB migration for adding missing noTablesOnMobile.css to vanilla and bootswatch configs
Beware: DBversion was updated.


application/helpers/update/updatedb_helper.php
```diff
@@ -3084,6 +3084,31 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
...
+        /**
+         * Add missing noTablesOnMobile.css to vanilla and bootswatch configs
+         */
+        if ($iOldDBVersion < 427) {
+            $oTransaction = $oDB->beginTransaction();
+            // Update vanilla config
+            $oDB->createCommand()->update(
+                '{{template_configuration}}',
+                [
+                    'files_css' => '{"add":["css/ajaxify.css","css/theme.css","css/custom.css","css/noTablesOnMobile.css"]}',
```
Shnoulle, line 3097: If someone update himself a template_configuration in DB to add or remove a css : you broke it …

You must
1. Read current
2. decode current 
3. Leave inherit as inherit
4. Add `css/noTablesOnMobile.css` 
5. Save it

gabrieljenik: These changes only apply for vanilla and bootswatch. I think template_configuration can't be changed for those. right?

gabrieljenik:
> Must not update already updated part. User can add or remove his own file.

I think vanilla and bootswatch are not to be updated by user. right?

> And your current one seems outdated 

I will double check. Thanks!

application/helpers/update/updatedb_helper.php
gabrieljenik: Reviewed. No in sync with config.xml

gabrieljenik:
Reviewed. No in sync with config.xml

application/helpers/update/updatedb_helper.php
Shnoulle: It can be manually edited (master screenshot)
Capture d’écran du 2020-07-09 08-37-36

gabrieljenik: > It can be manually edited (master screenshot)
> Capture d’écran du 2020-07-09 08-37-36

Honestly, I couldn't find that screen in v4.3.3
Let me take another look.

gabrieljenik: I have checked with v4.3.3 and didn't found that screen.
Let me triple check and be back with you

Shnoulle: Oh right , but can be updated in 3.X.

Your right : it's a 3.X screenshit (think i took it on master sorry).

Still : dislike the fix but not the fix : the global way for updating a theme …

sushmanadendlals:
Tested the issue by pulling the PR. Working as expected.

PR #1474, opened 2020-07-08
Fixed issue #16135 - Map based question with Google Map - displaying as static in mobile browser (LS3)

gabrieljenik:
Question help was getting over the map for small screens. That made the map look static.


sushmanadendlals:
Tested the issue by pulling the PR. Working as expected

PR #1475, opened 2020-07-08
Fixed issue #16451: When admin previews surveys timer is not reset in between runs

gabrieljenik:
On preview mode, set the 'resetQuestionTimers' js flag from server side, so timer.js can pick it and unset the localstorage


sushmanadendlals:
Tested following issues by pulling the PR:
1. Timer expiration-Page expires immediately after the set time is over and displays a message saying page expired.
2. Timer is reset in between runs  when the survey is viewed in Preview mode and in Active mode the timer does not reset.
3. Tested Timer by Previewing and Activating the survey. 
Seems to work fine in all cases.

PR #1477, opened 2020-07-08
Fixed issue #16151: Map question not displayed if loaded after a condition was met (both have to be on the same page)

gabrieljenik:
MapQuest setup was not being triggered on JS.

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -7744,7 +7744,8 @@
                             $relParts[] = "  $('#question" . $arg['qid'] . "').trigger('relevance:on');\n";
                             if ($arg['type'] == 'S')
                             {
-                                $relParts[] = "  if($('#question" . $arg['qid'] . " div[id^=\"gmap_canvas\"]').length > 0)\n";
+                                $relParts[] = "  if($('#question" . $arg['qid'] . " div[id^=\"gmap_canvas\"]').length > 0 || \n";
+                                $relParts[] = "     $('#question" . $arg['qid'] . " div[id^=\"map\"]').length > 0)\n";
                                 $relParts[] = "  {\n";
                                 $relParts[] = "      resetMap(" . $arg['qid'] . ");\n";
                                 $relParts[] = "  }\n";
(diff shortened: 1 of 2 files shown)


application/helpers/expressions/em_manager_helper.php
```diff
@@ -7744,7 +7744,8 @@ public static function GetRelevanceAndTailoringJavaScript($bReturnArray=false)
                             $relParts[] = "  $('#question" . $arg['qid'] . "').trigger('relevance:on');\n";
                             if ($arg['type'] == 'S')
                             {
-                                $relParts[] = "  if($('#question" . $arg['qid'] . " div[id^=\"gmap_canvas\"]').length > 0)\n";
```
Shnoulle, line 7747: For master : can we move this part on $(document).on("relevance:on") function ?

Shnoulle: Maybe in 3.X too in fact …

sushmanadendlals: Tested the issue by pulling the PR. Working as expected.

sushmanadendlals:
Tested the issue by pulling the PR. Working as expected.

application/helpers/expressions/em_manager_helper.php
Shnoulle: Yes, it's not working or not working :) . It's Map question javascript must be included in Map question system. Not added in core (if we can)

gabrieljenik: I can do an extra function inside maps.js that will do the setup taking a `qid` as parameter.
So that block will be moved into a new function.
Would that work?

Shnoulle: Remind : it's only my opinion :) 

Think adding here : https://github.com/LimeSurvey/LimeSurvey/blob/9fd5e78aa6a67b1078a05746c29bceb9f8fab7c0/assets/scripts/map.js#L18

```
$("#question"+question_id).on("relevance:on", function() { 
 resetMap(question_id);
});
```

But make it only in 4.X , and we can remove the EM part

Shnoulle: No need extra function : map.js have already function no need more

gabrieljenik: Sorry. This has been tested and merged. Not sure about the idea. I don't dare to do that now.

PR #1479, opened 2020-07-09
Master ls base controller check params

Shnoulle:
Fixed issue #16469: Any admin user can see any question (without read right on survey)
Fixed issue #16467: Reflected XSS vulnerabilities
Dev: add a function to validate int parameters
Dev: throw 400/403 or 404 error


olleharstedt:
Not too happy about this, as I've expressed in our chat. Can lead to false sense of security if a controller uses another name than the ones listed in checkparams.

Shnoulle:
> Can lead to false sense of security if a controller uses another name than the ones listed in checkparams.

Yes, right

But : seems some developer don't care about security … https://github.com/LimeSurvey/LimeSurvey/blob/9fd5e78aa6a67b1078a05746c29bceb9f8fab7c0/application/controllers/QuestionEditorController.php#L19

thedirtypanda:
This is under refactoring. So please say that she doesnt care. This is not true Denis! First we are refactoring the controllers afterwards the code refactoring will follow ;)

Shnoulle:
Controller are done without any Permission control : it's not «under» refactoring after releasing 4 minor version …

thedirtypanda:
So Denis, u dont know our schedule. Dont say anything thats not true. U know nothing.

Shnoulle:
True :

- [x] QuestionEditorController are done without any Permission control 
- [x] 3 minor version since LimeSurvey 4 is out as «**stable**»

Sorry : my error 3 minor version not 4

olleharstedt:
Markus forgot to add permission checks on most of his controllers. We've been adding them step by step. The access rules are irrelevant at the moment. Calm down, everybody. :)

Shnoulle:
>  Can lead to false sense of security if a controller uses another name than the ones listed in checkparams.

Fixed: create a function : public function **must** use it for security. The mast part are to review getQuestionObject

Think must be `getQuestionObject($surveyId, $qid=null, $gid = null,  $sQuestionType = null)` surveyId must come **only** from public function.

Shnoulle:
We can add `Permission::model()->hasSurveyPermission($surveyid, 'survey', 'read')` in «helper» function, but need more specific control after.

thedirtypanda:
I dont have this functions inside my master version of LSBaseController. Is the branch correct?

Shnoulle:
What function ?

I add : getValidatedSurveyId function and use it … don't understand your question.

thedirtypanda:
Ahhh u added it. Okay cause it looked like, that the function already exists in our codebase. The new Github Ui is confusing.

Shnoulle:
?
Capture d’écran du 2020-07-14 12-22-01

(dislike the new UI too, but here : seems OK ?)

Shnoulle:
Any news on this one ?

olleharstedt:
Vacation here, back next week. :ice_cream: :sun_with_face: :swimmer:

thedirtypanda:
We will discuss this next week, when @olleharstedt  is back ;)

olleharstedt:
Merged anyway? Does this solve the problem of passing both survey id and question id (or group id)?

application/controllers/LSBaseController.php
```diff
@@ -55,6 +55,66 @@ protected function _init()
...
+                throw new CHttpException(404);
+            }
+            if ($oQuestion && $sid != $oQuestion->sid) {
+                // Try to hack : 400
+                throw new CHttpException(400);
+            }
+            if ($oGroup && $sid != $oGroup->sid) {
+                // Try to hack : 400
+                throw new CHttpException(400);
+            }
```
olleharstedt, line 106: Does not check permissions?

olleharstedt:
Not a proper fix of 16469, IMO, since there's no permission checks.

olleharstedt:
And above all, where's the link to the discussion?

application/controllers/LSBaseController.php
```diff
@@ -202,8 +188,6 @@ public function run($action)
                 }
             }
         }
-        $this->checkParams();
```
olleharstedt: I like this change, makes it more explicit.

application/controllers/LSBaseController.php
Shnoulle: No, you ask to remove Permission checker on the global system …

> Not too happy about this, as I've expressed in our chat. Can lead to false sense of security if a controller uses another name than the ones listed in checkparams.

The only way to disable usage of a different surveyid/sid/survyId … is to get it by another function …

Then : you need the real survey id , and check permission after … the function return the sid according to qid or gid.

olleharstedt: Yeah, I think it's OK since it's not global anymore, more explicit. Thank you, Denis. Enjoy your vacation. :)

Shnoulle: In my opinion: there still an issue with this https://github.com/LimeSurvey/LimeSurvey/pull/1479/files#diff-44c5279b74ca1be33751c347b23e1585R91

Else about control on Permission : currently, need only read, but when add a function to update : need 'update' etc … Persmission check done in each function (and must be only in public function)

Shnoulle:
@olleharstedt and @thedirtypanda : maybe (unsure) need to check why encoding is not donein vue before ?

The question : did encode or filter must be donein view (Vue) each time or not. My opinion : we must always encode/filter in Vue/View even if var is already checked by controller/model. It's an opinion.

PR #1480, opened 2020-07-10
Fixed issue #16481: Easy way to launch CDBException

Shnoulle:
Dev: use checkValidityAnswer to allow or not prefilling
Dev: set other to true : can be invalid, but don't broke DB


Shnoulle:
Looki_ng at code , `$this->setMoveResult();` seems happen for 1st page too . Maybe we have all qinfo at this time.

Shnoulle:
for master : we need a better solution : move after setMoveResult ?

PR #1482, opened 2020-07-10
Fixed issue #16454: Sometimes, question_order is assigned from survey_id

gabrieljenik:
When deleting a question, the order of questions in the group were reassigned by calling updateQuestionOrder, but the survey ID was passed as starting position.

updateQuestionOrder definition was modified to use 1 as default value. Now reordering will follow the same schema as when questions are added, both starting from 1.

we found "database::actionUpdateQuestion()" to be deprecated. 
https://github.com/gabrieljenik/LimeSurvey/blob/1b975c60d3c4128fc72a542b07d3d981bb2eae9c/application/controllers/admin/database.php#L579
Should it be removed?

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -428,7 +428,7 @@ class Question extends LSActiveRecord
         $this->removeFromLastVisited();
 
         if (parent::delete()) {
-            Question::model()->updateQuestionOrder($this->gid, $this->sid);
+            Question::model()->updateQuestionOrder($this->gid);
             return true;
         }
         return false;
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -250,10 +250,12 @@ class Question extends LSActiveRecord
 
     /**
      * Fix sort order for questions in a group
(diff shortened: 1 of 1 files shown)


gabrieljenik:
I think this instance is missing to review.
https://github.com/gabrieljenik/LimeSurvey/blob/1b975c60d3c4128fc72a542b07d3d981bb2eae9c/application/controllers/admin/database.php#L1518

I will pickup on Monday

gabrieljenik:
> I think this instance is missing to review.
> https://github.com/gabrieljenik/LimeSurvey/blob/1b975c60d3c4128fc72a542b07d3d981bb2eae9c/application/controllers/admin/database.php#L1518
> 
> I will pickup on Monday

Could this be deprecated too?

cdorin93:
@olleharstedt , do you have time to review quickly this one please?

olleharstedt:
Code looks good. Did someone test it?

PR #1485, opened 2020-07-16
ENH #16378: New Plugin Event: beforeTokenImport

gabrieljenik:
New plugin event implemented for CSV import.
Demo plugin implementation

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -58,7 +58,7 @@ class tokens extends Survey_Common_Action
 
         // CHECK TO SEE IF A Survey participants table EXISTS FOR THIS SURVEY
         if (!$survey->hasTokensTable) {
-//If no tokens table exists
+            //If no tokens table exists
             $this->_newtokentable($iSurveyId);
         } else {
             $aData['thissurvey'] = $thissurvey;
@@ -2148,7 +2148,42 @@ class tokens extends Survey_Common_Action
                             }
                         }
 
-                        if (!$bDuplicateFound && !$bInvalidEmail && !$bInvalidToken) {
+                        // Dispatch beforeTokenImport event
+                        $params = array(
+                            'csvcharset' => $sUploadCharset,
+                            'filterduplicatetoken' => $bFilterDuplicateToken,
+                            'filterblankemail' => $bFilterBlankEmail,
+                            'allowinvalidemail' => $bAllowInvalidEmail,
+                            'filterduplicatefields' => $aFilterDuplicateFields,
+                            'separator' => $sSeparator,
+                            'showwarningtoken' => Yii::app()->request->getPost('showwarningtoken'),
+                        );
+                        $event = new PluginEvent('beforeTokenImport');
+                        $event->set('importType', 'CSV');
+                        $event->set('surveyId', $iSurveyId);
+                        $event->set('params', $params);
+                        $event->set('recordCount', $iRecordCount);
+                        $event->set('token', $aWriteArray);
+                        App()->getPluginManager()->dispatchEvent($event);
+                        $bPluginReportedError = !is_null($event->get('importValid')) && !$event->get('importValid');
+
+                        if ($bPluginReportedError) {
+                            // If plugin says import is not valid, append the error
+                            $sErrorMessage = $event->get('errorMessage');
+                            if (empty($sErrorMessage)) $sErrorMessage = gT("%s records with other errors");
+                            
+                            $sTokenSpecificErrorMessage = $event->get('tokenSpecificErrorMessage');
+                            if (!empty($sTokenSpecificErrorMessage)) {
+                                $aPluginErrorMessageList[$sErrorMessage][] = $sTokenSpecificErrorMessage;
+                            } else {
+                                $aPluginErrorMessageList[$sErrorMessage][] = sprintf(gT("Line %s : %s %s (%s)"), $iRecordCount, $aWriteArray['firstname'], $aWriteArray['lastname'], $aWriteArray['email']);
+                            }
+                        } else {
+                            // If plugin says import is OK, replace token data from the event
+                            $aWriteArray = $event->get('token');
+                        }
+
+                        if (!$bDuplicateFound && !$bInvalidEmail && !$bInvalidToken && !$bPluginReportedError) {
                             // unset all empty value
                             foreach ($aWriteArray as $key => $value) {
                                 if ($aWriteArray[$key] == "") {
@@ -2190,6 +2225,7 @@ class tokens extends Survey_Common_Action
                 $aData['aInvalidTokenList'] = $aInvalidTokenList;
                 $aData['aInvalidFormatList'] = $aInvalidFormatList;
                 $aData['aInvalidEmailList'] = $aInvalidEmailList;
+                $aData['aPluginErrorMessageList'] = $aPluginErrorMessageList;
                 $aData['aModelErrorList'] = $aModelErrorList;
                 $aData['iInvalidEmailCount'] = $iInvalidEmailCount;
                 $aData['thissurvey'] = getSurveyInfo($iSurveyId);
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -2800,6 +2800,16 @@ class tokens extends Survey_Common_Action
 
     /**
      * Dispatches the beforeTokenImport event
+     * @param int $iSurveyId
+     * @param array $aToken
+     * @param string $sUploadCharset
+     * @param boolean $bFilterDuplicateToken
+     * @param boolean $bFilterBlankEmail
+     * @param boolean $bAllowInvalidEmail
+     * @param array $aFilterDuplicateFields
+     * @param string $sSeparator
+     * @param boolean $bShowWarningToken
+     * @param int $iRecordCount
      * @return array
      */
     protected function dispatchBeforeTokenImport($iSurveyId, $aToken, $sUploadCharset, $bFilterDuplicateToken, $bFilterBlankEmail, $bAllowInvalidEmail, $aFilterDuplicateFields, $sSeparator, $bShowWarningToken, $iRecordCount)
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -2148,18 +2148,17 @@ class tokens extends Survey_Common_Action
                             }
                         }
 
-                        $aEventResult = $this->dispatchBeforeTokenImport(
-                            $iSurveyId,
-                            $aWriteArray,
-                            $sUploadCharset,
-                            $bFilterDuplicateToken,
-                            $bFilterBlankEmail,
-                            $bAllowInvalidEmail,
-                            $aFilterDuplicateFields,
-                            $sSeparator,
-                            Yii::app()->request->getPost('showwarningtoken'),
-                            $iRecordCount
+                        // Dispatch beforeTokenImport event
+                        $aOptions = array(
+                            'csvcharset' => $sUploadCharset,
+                            'filterduplicatetoken' => $bFilterDuplicateToken,
+                            'filterblankemail' => $bFilterBlankEmail,
+                            'allowinvalidemail' => $bAllowInvalidEmail,
+                            'filterduplicatefields' => $aFilterDuplicateFields,
+                            'separator' => $sSeparator,
+                            'showwarningtoken' => Yii::app()->request->getPost('showwarningtoken'),
                         );
+                        $aEventResult = $this->dispatchBeforeTokenImport($iSurveyId, $aWriteArray, $iRecordCount, $aOptions);
                         $bPluginReportedError = $aEventResult['pluginReportedError'];
                         $bImportDone = $aEventResult['importDone'];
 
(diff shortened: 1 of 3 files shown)


Shnoulle:
@gabrieljenik did this pull request allow to create import and update system ?

For example : import token with token column, search Token with `token`, if exist : update the existing Token, if not : create a new Token. Something like VV import :)

application/controllers/admin/tokens.php
```diff
@@ -2148,7 +2148,42 @@ public function import($iSurveyId)
...
+                            'showwarningtoken' => Yii::app()->request->getPost('showwarningtoken'),
+                        );
+                        $event = new PluginEvent('beforeTokenImport');
+                        $event->set('importType', 'CSV');
+                        $event->set('surveyId', $iSurveyId);
+                        $event->set('params', $params);
+                        $event->set('recordCount', $iRecordCount);
+                        $event->set('token', $aWriteArray);
+                        App()->getPluginManager()->dispatchEvent($event);
+                        $bPluginReportedError = !is_null($event->get('importValid')) && !$event->get('importValid');
```
Shnoulle, line 2168: Why not add $event->set('importValid', true) bedore dispatching event. Then after event we can just test `if(!$event->get('importValid'))` ?

Shnoulle:
Maybe need 2 value to return
 - [importValid](https://github.com/LimeSurvey/LimeSurvey/pull/1485/files#diff-091a0a33007efec2513a2239ea06c9ebR2168) like currentky
- [importDone](https://github.com/LimeSurvey/LimeSurvey/pull/1485/files#diff-091a0a33007efec2513a2239ea06c9ebR2181) (or just import) : if importDone is true : just add as succees. Something like `sent` for email

gabrieljenik:
> @gabrieljenik did this pull request allow to create import and update system ?
> 
> For example : import token with token column, search Token with `token`, if exist : update the existing Token, if not : create a new Token. Something like VV import :)

After that I apply the new parameter "importDone", it should allow you to do whatever import process you want.

gabrieljenik:
> Add `$event->set('importValid',true)` seems easiest to understand :).
> 
> Add `$event->set('importDone',false)` (or another setting name ( `doImport => true` ?) can be a great improvement to allow plugin to do the import , sending success, but disable import by core.

Agree. I will ad it.

Shnoulle:
Maybe `doImport => true` is better for this param ? 

Forr exemple [`send` param](https://manual.limesurvey.org/BeforeEmail) in mail event is more like `doImport` , if we follow same we use `import => true` but personnaly : i think it's unclear.

Use what you want ;)

gabrieljenik:
What about now?

Shnoulle:
For : https://github.com/LimeSurvey/LimeSurvey/pull/1485/files#r455560055 
You can replace by ` $bPluginReportedError = !$event->get('importValid');` directly (since is set to true before)

And still need a minor PSR2 :)

gabrieljenik:
> For : https://github.com/LimeSurvey/LimeSurvey/pull/1485/files#r455560055
> You can replace by ` $bPluginReportedError = !$event->get('importValid');` directly (since is set to true before)

Fixed

> And still need a minor PSR2 :)

Oh, right. Sorry. They were shown as warnings :) Fixed them

application/controllers/admin/tokens.php
```diff
@@ -2148,7 +2148,47 @@ public function import($iSurveyId)
...
+                            
+                            $sTokenSpecificErrorMessage = $event->get('tokenSpecificErrorMessage');
+                            if (!empty($sTokenSpecificErrorMessage)) {
+                                $aPluginErrorMessageList[$sErrorMessage][] = $sTokenSpecificErrorMessage;
+                            } else {
+                                $aPluginErrorMessageList[$sErrorMessage][] = sprintf(gT("Line %s : %s %s (%s)"), $iRecordCount, $aWriteArray['firstname'], $aWriteArray['lastname'], $aWriteArray['email']);
+                            }
+                        } else {
+                            // If plugin says import is OK, replace token data from the event
+                            $aWriteArray = $event->get('token');
```
Shnoulle, line 2188: :+1:

application/controllers/admin/tokens.php
```diff
@@ -2148,7 +2148,47 @@ public function import($iSurveyId)
...
+                        $event->set('importType', 'CSV');
+                        $event->set('surveyId', $iSurveyId);
+                        $event->set('params', $params);
+                        $event->set('recordCount', $iRecordCount);
+                        $event->set('token', $aWriteArray);
+                        $event->set('importDone', false);
+                        $event->set('importValid', true);                        
+                        App()->getPluginManager()->dispatchEvent($event);
+                        $bPluginReportedError = !$event->get('importValid');
+                        $bImportDone = $event->get('importDone');
```
olleharstedt, line 2171: Can we split out this code section into a new function (in same file)? I don't like it when functions grow indefinitely. :|

olleharstedt (changes requested):

application/controllers/admin/tokens.php
gabrieljenik: Something like wrapping the whole plugin call in a function?

olleharstedt: If possible, yes, would be great.

application/views/admin/token/csvimportresult.php
```diff
@@ -101,6 +102,24 @@
                                 </li>
                             <?php } ?>
 
+                            <?php if (!empty($aPluginErrorMessageList)) { 
+                                        $iPluginErrorIndex=0;
+                                        foreach ($aPluginErrorMessageList as $sPluginErrorMessage => $aTokenSpecificErrorList) {
```
olleharstedt, line 107: Consider using alternative syntax here.

https://www.php.net/manual/en/control-structures.alternative-syntax.php

application/views/admin/token/csvimportresult.php
```diff
@@ -101,6 +102,24 @@
                                 </li>
                             <?php } ?>
 
+                            <?php if (!empty($aPluginErrorMessageList)) { 
+                                        $iPluginErrorIndex=0;
+                                        foreach ($aPluginErrorMessageList as $sPluginErrorMessage => $aTokenSpecificErrorList) { 
+                                            $iPluginErrorIndex++; ?>
```
olleharstedt, line 108: Kind of too much logic in a viewfile, but not sure how to fix. Maybe do a small widget? Or too much work?

olleharstedt (changes requested):

application/views/admin/token/csvimportresult.php
Shnoulle: At this line ? I didn't see how :)

olleharstedt: On all lines.

Well, it gets hard because of this: "$iPluginErrorIndex=0;". But lines like that don't belong in viewfiles. Hence, widget might be better?

cdorin93:
What is the status of this fix :thinking: ?

application/views/admin/token/csvimportresult.php
gabrieljenik: Widget for this would be an overkill I think.

> But lines like that don't belong in viewfiles. 
Don't agree too much. I understand alternative syntax is better for view files, but this is just used for numbering. This kind of logic doesn;t belong to a controller either. I would say is more a view logic.

I would just leave it like that or maybe just chnage it to the Alternative Syntax using multiple lines.
But honestyl, I don't believe it worth doing it.

application/views/admin/token/csvimportresult.php
gabrieljenik: The only difference in between previous blocks and this new one is we are just numbering lines... I wouldn;t do a widget as it is not reusable  If not, I could add some logic in the top of the file as to add the numbering and create a plain array. What do you think?

application/views/admin/token/csvimportresult.php
eddylackmann: I don t see any needs to use alternative synthax. is that needed @olleharstedt ?

olleharstedt: Me and Louis used alternative syntax as we thought it had higher readability. But it's not mandated in any style guide.

application/views/admin/token/csvimportresult.php
olleharstedt: Hm, such logic should preferably be in the controller. Bah. Yes, it's not an easy task to separate view from logic. Leave for now and concentrate on the event function above instead.

application/controllers/admin/tokens.php
gabrieljenik: Done!

gabrieljenik:
When approved, will apply it for LS4 as well

application/controllers/admin/tokens.php
```diff
@@ -2769,4 +2797,58 @@ protected function tokenIsSetInEmailCache($iSurveyId, $tid, $bIsInvitation)
             && isset($_SESSION[$cacheName][$sType][$tid])
             && $_SESSION[$cacheName][$sType][$tid] > 0;
     }
+
+    /**
+     * Dispatches the beforeTokenImport event
+     * @return array
+     */
+    protected function dispatchBeforeTokenImport($iSurveyId, $aToken, $sUploadCharset, $bFilterDuplicateToken, $bFilterBlankEmail, $bAllowInvalidEmail, $aFilterDuplicateFields, $sSeparator, $bShowWarningToken, $iRecordCount)
```
olleharstedt, line 2805: Missing PHPDocs.

olleharstedt (changes requested):

olleharstedt:
Thanks for the updates!

Other question: Should we start writing unit-tests for events?

application/controllers/admin/tokens.php
gabrieljenik: Updated it

gabrieljenik:
Yes, we could. We should think for some approach. On a separate item?

olleharstedt:
> Yes, we could. We should think for some approach. On a separate item?

I'll create a Mantis ticket.

application/controllers/admin/tokens.php
Shnoulle: I really dislike a bunch of params like this …

I really prefer something like 
````
public function dispatchBeforeTokenImport($iSurveyId, $aToken, $aOptions)
{
    $aOptions =array_merge(
        array(
            'csvcharset' => 'UTF-8',
            'filterduplicatetoken' => false,
            'filterblankemail' => true,
            'allowinvalidemail' => false,
            'filterduplicatefields' => array(),
            'separator' => ',',
            'showwarningtoken' => true,
        ),
        $aOptions
    );
    // Do the job
}

````
Really easiest to understand in my point of view.

**Review by Shnoulle — COMMENTED** [review]
Just commented, not request. Prefer a array for options than a lot of iline options.

There are issue in some of functions between 3 and 4 because we remove an unused param an ExpressionManager function. If we have a array of option: less code to update :)

**Comment by Shnoulle** [comment]
For ExpressionManager event : i write core plugins … i don't see how we can do except have a plugin using the event ?

File: application/controllers/admin/tokens.php (hunk 6 above)
Comment (gabrieljenik) [reply application/controllers/admin/tokens.php:2815]: I agree. it is more readable. Let me re-write it. Honestly didn't want to create a block whre I need to setup that array just for calling this method.

**Comment by gabrieljenik** [comment]
> Just commented, not request. Prefer a array for options than a lot of iline options.
> 
> There are issue in some of functions between 3 and 4 because we remove an unused param an ExpressionManager function. If we have a array of option: less code to update :)

Updated

File: application/controllers/admin/tokens.php (hunk 6 above)
Comment (Shnoulle) [reply application/controllers/admin/tokens.php]: Thanks you :) And better it was send in array in plugin :)

**Review by Shnoulle — APPROVED** [review]

**Comment by olleharstedt** [comment]
> Just commented, not request. Prefer a array for options than a lot of iline options.

Even better are data-value objects, or data-transfer objects, but not sure it's worth it in this case.

File: application/controllers/admin/tokens.php [hunk 7]
```diff
@@ -2769,4 +2796,70 @@ protected function tokenIsSetInEmailCache($iSurveyId, $tid, $bIsInvitation)
...
+     * @param boolean $bFilterDuplicateToken
+     * @param boolean $bFilterBlankEmail
+     * @param boolean $bAllowInvalidEmail
+     * @param array $aFilterDuplicateFields
+     * @param string $sSeparator
+     * @param boolean $bShowWarningToken
+     * @param int $iRecordCount
+     * @return array
+     */
+    protected function dispatchBeforeTokenImport($iSurveyId, $aToken, $iRecordCount, $aOptions)
```
olleharstedt, line 2814: I'm sorry, but now PHPDocs is not correct anymore. :laughing: Just replace options with `array $aOptions`.

olleharstedt (changes requested):

olleharstedt:
> > Just commented, not request. Prefer a array for options than a lot of iline options.
> > There are issue in some of functions between 3 and 4 because we remove an unused param an ExpressionManager function. If we have a array of option: less code to update :)
> 
> Updated

:+1:

Shnoulle:
> > Just commented, not request. Prefer a array for options than a lot of iline options.
> 
> Even better are data-value objects, or data-transfer objects, but not sure it's worth it in this case.

Yes, but only when we need a clean object :)

olleharstedt:
> Yes, but only when we need a clean object :)

That's always the case. :|

https://steemit.com/php/@crell/php-use-associative-arrays-basically-never

application/controllers/admin/tokens.php
gabrieljenik: Updated. :)

olleharstedt:
That import() function is in severe need of refactoring.

olleharstedt:
Merged. Thank you!

olleharstedt:
We need to cherry-pick it into master now.

gabrieljenik:
https://bugs.limesurvey.org/view.php?id=16785

PR #1489, opened 2020-07-20
Fixed Issue #15749 - Adding new users as superadmins doesn't check all boxes

gabrieljenik:
Original LS3 behavior implemented: Other checkboxes should checked automatically.

There was a missmatch when appluing selection queries.


sushmanadendlals:

PR #1491, opened 2020-07-20
Fixed issue #16149: See survey list - owner column is not correct

gabrieljenik:
Showing owner's full name


sushmanadendlals:
Tested the issue before pulling the PR, Issue exist. Tested the issue after pulling the PR, but the column is blank. I tried clearing application cache and closing the tab and coming back but it still remains the same.

sushmanadendlals:
Tested the issue after pulling the PR, working as expected.

PR #1492, opened 2020-07-20
Fixed issue #16498: Can't run survey with child theme because of enheritance problems.

gabrieljenik:
1- Removed comments after loading config.xml.
2 - Had empty nodes to be saved in DB as empty strings.


sushmanadendlals:
Tested also by enabling debug mode

PR #1493, opened 2020-07-20
Fixed issue #15978: Multiple numerical input (Equals sum value: Remaining and Total missing)

gabrieljenik:
Functionality was not implemented in RenderMultipleNumerical.php from LSv3


sushmanadendlals:
Tested the issue before pulling the PR, Issue exist. Tested the issue after pulling the PR, working as expected

PR #1498, opened 2020-07-28
Fixed issue #16520: For specific screen navigation, condition is not saved properly

gabrieljenik:
Resetting cqid when using token fields


sushmanadendlals:
Tested the issue before pulling the PR, Issue exist. Tested the issue after pulling the PR working as expected

PR #1499, opened 2020-07-28
Fixed issue #16273 - Validation regex including unicode characters fails

gabrieljenik:
Decoding html before running regex. This (a decode string) is similar to what the PHP side regex function gets.


sushmanadendlals:
Working as expected,Please refer Mantis for more details

PR #1502, opened 2020-07-28
Fixed issue #15951: No creation date when importing participants to CPDB

gabrieljenik:
Made creation date to be current date

### DIFF
--- a/application/models/Participant.php
+++ b/application/models/Participant.php
@@ -2124,7 +2124,8 @@ class Participant extends LSActiveRecord
             'language' => $data['language'],
             'blacklisted' => $data['blacklisted'],
             'created_by' => $data['owner_uid'],
-            'owner_uid' => $data['owner_uid']);
+            'owner_uid' => $data['owner_uid'],
+            'created' => date('Y-m-d H:i:s', time()));
         Yii::app()->db->createCommand()->insert('{{participants}}', $insertData);
     }
 


application/models/Participant.php
```diff
@@ -2124,7 +2124,8 @@ public function insertParticipantCSV($data)
             'language' => $data['language'],
             'blacklisted' => $data['blacklisted'],
             'created_by' => $data['owner_uid'],
-            'owner_uid' => $data['owner_uid']);
+            'owner_uid' => $data['owner_uid'],
+            'created' => date('Y-m-d H:i:s', time()));
```
Shnoulle, line 2128: Need time shift ? No ? (`timeadjust`)
https://github.com/LimeSurvey/LimeSurvey/blob/1daf9d4347b9a102a26aa8ba67e70a20fd98438a/application/helpers/common_helper.php#L2491

gabrieljenik: Will fix

sushmanadendlals:
Tested the issue before pulling the PR, Issue exist. Tested the issue after pulling the PR, below are my findings:
Screenshot 1: before the Import Created On Date is displayed
Screenshot 2: After the Import Created On Date is not displayed
Screenshot 3: In the Branch before the Import Created On Date is displayed
Screenshot 4: In the Branch after the Import Created On Date is displayed
Please refer the attachment for more details, Working as expected
15951_CPDB_CreatedOnDate

application/models/Participant.php
gabrieljenik: OK, so after some review I got the following.
- When a token is created, `created ` is assigned `NOW()` from DB.
- When a token is copied, `created ` is assigned `date('Y-m-d H:i:s', time()));`
https://github.com/LimeSurvey/LimeSurvey/blob/1d87c0e7c3fa972b32b16a2ef0ea052376d81a2e/application/models/Participant.php#L2030

So, seems dateShift is not used here at the Participants Model.
Maybe we should, but we could open a new ticket for that as to review it specificlly and handle it easier.
What do you think?

Shnoulle: > So, seems dateShift is not used here at the Participants Model.

Seems to be an issue :) 

>  Maybe we should, but we could open a new ticket for that as to review it specificlly and handle it easier.

Yes :+1: (but i'm on holiday on August …)

gabrieljenik: https://bugs.limesurvey.org/view.php?id=16568

Happy hollidays!

PR #1505, opened 2020-07-28
Fixed issue #15945: Cannot add condition based on Survey participant attributes

gabrieljenik:
Considering editSourceTab for setting up the condition fieldName

As the question select was initialized, the p_cquestions field had a value.
As so, it was always set, no matter if a token field was chosen.
So, when selecting a token field, a question field was wrongly set on the condition.


sushmanadendlals:

PR #1507, opened 2020-07-28
Fixed issue #16521: After updating an answer option, the related condition doesn't get updated

gabrieljenik:
Updated dependant conditions when an answer's code is updated.


sushmanadendlals:
Tested the issue before pulling the PR, Issue exist. Tested the issue after pulling the PR working as expected

PR #1512, opened 2020-07-28
Fixed issue #16250: Slider for numeric question display text area

gabrieljenik:
Hide the slider options, which were not available on LS3 anyway.
If we want to implement slider for LS3, we could make a new ticket.


sushmanadendlals:
Tested the issue before pulling the PR, Issue exist. Tested the issue after pulling the PR, below are my findings:
Screenshot 1: before Pulling the PR questiontype -Numerical input Slider tab was displayed
Screenshot 2:Preview of survey even when you select slider options nothing is been shown
Screenshot 3: After pulling the PR , the slider tab is been hidden and not shown
Please refer the attachment for more details
16250_Numericalinput_Slidertabhidden

PR #1513, opened 2020-07-28
Fixed issue #16529: After deleting a user who owned a survey, the survey list generates a dump

gabrieljenik:
Added getOwnerUserName in Survey model, and use that ownerUserName property in listSurveys.php.


sushmanadendlals:
Tested the issue in the latest version. Was not able to replicate the issue. Please refer the attachment for more details.
Scenario 1: Before Deleting the User
Scenario 2: After Deleting the User I don't see any name displayed in master
Master: I don't get any message to move surveys to other users.

I tried pulling PR and tested but still I see the same behavior as above.

sushmanadendlals:
16529_UserDelete_OwnerBlank

PR #1514, opened 2020-07-28
Fixed issue #16533: Double clicking on question overiew for read only user navigates to question editor

gabrieljenik:
Checking permissions before switching to edit mode


sushmanadendlals:

PR #1516, opened 2020-07-28
Fixed issue #16380: On clicking on delete user icon from user control screen lands in different page

gabrieljenik:
Route was relative instead of complete (including controller/action). When usng not friendly urls, redirection failed.


sushmanadendlals:
Tested the issue before pulling the PR, Issue exist. Tested the issue after pulling the PR, working as expected

PR #1521, opened 2020-07-29
Fixed issue #16531: Validation regex including unicode characters fails

gabrieljenik:
Decoding html before running regex. This (a decode string) is similar to what the PHP side regex function gets.

### DIFF
--- a/assets/packages/expressions/em_javascript.js
+++ b/assets/packages/expressions/em_javascript.js
@@ -850,7 +850,8 @@ function LEMregexMatch(sRegExp,within)
         var flags = sRegExp.replace(/.*\/([gimy]*)$/, '$1');
         var pattern = sRegExp.replace(new RegExp('^/(.*?)/'+flags+'$'), '$1').trim();
         var reg = new RegExp(pattern, flags); // Note that the /u flag crashes IE11
-        return reg.test(within);
+        var decodedValue = html_entity_decode(within);
+        return reg.test(decodedValue);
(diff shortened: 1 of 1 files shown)


Shnoulle:
It's do the same than https://github.com/LimeSurvey/LimeSurvey/blob/1daf9d4347b9a102a26aa8ba67e70a20fd98438a/application/helpers/expressions/em_core_helper.php#L3057 ?

I'm nbot a regexp specialist : if you can take 5 minutes to explain why we need `flags = sRegExp.replace(/.*\/([gimy]*)$/, '$1');` in js ?
Maybe put more comments on this 2 functions ;)

Shnoulle:
Maybe : add a expressionmanager regexmatch [functional test](https://github.com/LimeSurvey/LimeSurvey/pull/1521) ?

gabrieljenik:
> I'm nbot a regexp specialist : if you can take 5 minutes to explain why we need flags = sRegExp.replace(/.*\/([gimy]*)$/, '$1'); in js ?

Me neither.
I just found that the text to the regex may come html-encoded, which breaks the regular expression. That's why we added the decode.

About the flags, I believe that they are being extracted, if any, from the regular expression string.
Later they are inputted as flags when creating the RegExp object.
I think it is part of a "string regex" --> "object regex" conversion.

sushmanadendlals:
Working as expected,Please refer Mantis for more details

PR #1522, opened 2020-07-29
Fixed issue #16539: "Structure" should be shown when editing conditions (via the condition designer)

gabrieljenik:


sushmanadendlals:
Tested in all the scenario's, Working as expected

PR #1527, opened 2020-07-31
Fixed issue #16555: bootswatch Image uploaded via Survey theme option are uploaded in Global directory

Shnoulle:
Dev: Add hidden input surveyid
Dev: Check if empty : no need to send if empty


Shnoulle:
Oups it's for 3.X, didn't check for master currently :)

PR #1528, opened 2020-07-31
Fixed issue #16554: Upgrading from 4.0.1 to 4.3.7, the "Save" button is now missing in question creation

gabrieljenik:
Checking permission in a different way.


Shnoulle:
Where canEdit is set ? Already set in another commit ?

gabrieljenik:
Nope. Here:
https://github.com/LimeSurvey/LimeSurvey/pull/1528/files#diff-79264db2fb815d47cde6ee5541abf936R257

PR #1530, opened 2020-08-04
Fixed issue #16536: Batch editing reminder sent date leads to wrong data being added

gabrieljenik:
Reviewed fields initialization and jquery handling.
- Fixed some element IDs on the mass edit modal dialog.
- Changed update method of hidden input from attr('value',xxx) to val(xxx)
- Added initialization of YesNo and YesNoDate fields on "shown.bs.modal" and "actions-updated" event.

### DIFF
--- a/application/views/admin/token/massive_actions/_update.php
+++ b/application/views/admin/token/massive_actions/_update.php
@@ -419,15 +419,12 @@
 <?php App()->getClientScript()->registerScript("Tokens:MassActionUpdateView_Scripts", "
 
    var bindBSSwitch = function(formGroup){
-       console.log(\"bindBSSwitch run on:\",formGroup);
-    //Script to update the completed settings
-    formGroup.find('.YesNoSwitch').on('switchChange.bootstrapSwitch', function(e, state){
-        
-        formGroup.find('.selector_datechange').css('display', (state ? '' : 'none'));
-        //formGroup.find('.selector_submitField').val(state ? 'Y' : 'N');
-
-    });
-   };
+        console.log(\"bindBSSwitch run on:\",formGroup);
+        //Script to update the completed settings
+        formGroup.find('.YesNoSwitch').on('switchChange.bootstrapSwitch', function(e, state){        
+            formGroup.find('.selector_datechange').css('display', (state ? '' : 'none'));
+        });
+    };
 
    var bindDatepicker = function(myFormGroup){
     myFormGroup.find('.action_datepickerUpdateHiddenField').on('change dp.change', function(){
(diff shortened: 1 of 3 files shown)


sushmanadendlals:
Working as expected, Please refer Mantis for more details

assets/scripts/admin/tokens.js
```diff
@@ -38,19 +38,19 @@ $.fn.YesNoDate = function(options)
             {
                 // Show date
                 $elDateContainer.show();
-                $elHiddenInput.attr('value', moment().format($elDate.data('date-format')));
+                $elHiddenInput.val(moment().format($elDate.data('date-format')));
```
olleharstedt, line 41: What's the difference?

gabrieljenik: When using val() the value is set in the DOM. When using attr, sometimes, that's not updated on the DOM

application/views/admin/token/massive_actions/_update.php
```diff
@@ -424,7 +424,7 @@ class="YesNoDatePicker form-control action_datepickerUpdateHiddenField"
     formGroup.find('.YesNoSwitch').on('switchChange.bootstrapSwitch', function(e, state){
         
         formGroup.find('.selector_datechange').css('display', (state ? '' : 'none'));
-        formGroup.find('.selector_submitField').val(state ? 'Y' : 'N');
+        //formGroup.find('.selector_submitField').val(state ? 'Y' : 'N');
```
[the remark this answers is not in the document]
gabrieljenik: Fixed

application/views/admin/token/massive_actions/_update.php
```diff
@@ -419,15 +419,12 @@ class="YesNoDatePicker form-control action_datepickerUpdateHiddenField"
...
-       console.log(\"bindBSSwitch run on:\",formGroup);
-    //Script to update the completed settings
-    formGroup.find('.YesNoSwitch').on('switchChange.bootstrapSwitch', function(e, state){
-        
-        formGroup.find('.selector_datechange').css('display', (state ? '' : 'none'));
-        //formGroup.find('.selector_submitField').val(state ? 'Y' : 'N');
-
-    });
-   };
+        console.log(\"bindBSSwitch run on:\",formGroup);
```
olleharstedt, line 422: Do we need this debugging statement?

olleharstedt: I don't remember if we had a proper logger in JS? That's only on if debug = 2, with different logging levels.

gabrieljenik: Fixed

PR #1531, opened 2020-08-04
Fixed issue #16326: Statistics filters do not reflect percentages and totals

gabrieljenik:
In LS3, each value is retrieved by a query to the DB. In LS4, the responses are retrieved first, and then each value is calculated by looping through the responses.
Some filters that were added to the query in LS3 were not working on LS4 because the SQL was still appended to a query that's never used. This was solved by adding the filters to the response selection criteria.
Also, the value for "No answer" that was calculated by comparing to empty string in SQL query (LS3), in LS4 was implemented with an "equal" operator (==). That comparison also matched null values, so it was replaced with an "identical" operator (===).


sushmanadendlals:

PR #1534, opened 2020-08-04
Fixed issue #16538: Can add 2 survey participantes with the same token with the RPC api

gabrieljenik:
Adding validation to the save operation.

What I have seen is that encryptSave doesn't do validation by default.
I haven't updated that, but shouldn't we? Why it was set to false?

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1809,7 +1809,7 @@ class remotecontrol_handle
                 if ($bCreateToken) {
                     $token->generateToken();
                 }
-                if ($token->encryptSave()) {
+                if ($token->encryptSave(true)) {
                     $aParticipant = $token->getAttributes();
                 } else {
                     $aParticipant["errors"] = $token->errors;
--- a/tests/unit/helpers/RemoteControlTest.php
+++ b/tests/unit/helpers/RemoteControlTest.php
@@ -33,6 +33,79 @@ class RemoteControlTest extends TestBaseClass
     }
 
     /**
+     * Test the add_participants API call.
+     */
+    public function testAddParticipants()
+    {
+        \Yii::import('application.helpers.remotecontrol.remotecontrol_handle', true);
+        \Yii::import('application.helpers.viewHelper', true);
+        \Yii::import('application.libraries.BigData', true);
+        $dbo = \Yii::app()->getDb();
+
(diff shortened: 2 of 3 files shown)


Shnoulle:
> What I have seen is that encryptSave doesn't do validation by default.
> I haven't updated that, but shouldn't we? Why it was set to false?

:+1:

sushmanadendlals:
Needs to be tested after the merge, as per discussion with Dorin

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -1809,7 +1809,7 @@ public function add_participants($sSessionKey, $iSurveyID, $aParticipantData, $b
                 if ($bCreateToken) {
                     $token->generateToken();
                 }
-                if ($token->encryptSave()) {
+                if ($token->encryptSave(true)) {
```
olleharstedt, line 1812: We need a unit-test for this, I think. It's not clear when and how validation can be used with encryption.

gabrieljenik: I agree unit-test could be a good way approach.

olleharstedt: Can you make one, in this PR, please? Ping me if you need help or instructions. There already is a test for the remote API that you can look at.

gabrieljenik:
Uploaded unit test

application/helpers/remotecontrol/remotecontrol_handle.php
olleharstedt: What you added looks good, but I was thinking of a new test in this file: https://github.com/LimeSurvey/LimeSurvey/blob/master/tests/unit/helpers/RemoteControlTest.php

For add_participants RPC call. Sorry for being vague. But, more test code is always good! :)

gabrieljenik: Added

olleharstedt: Great, will try it!

Shnoulle:
> Also, we could enhance this is with some kind of attribute which tells if the token is encrypted or not as to not double encrpyt or not double decrypt.

oToken->token are never encrypted, no ? I think it must never been encrypted. 

For other attributes : it's not in https://github.com/LimeSurvey/LimeSurvey/blob/7dc7fcf7268ee64d8e7d4bafa6b2f8e017a51ca0/application/models/Survey.php#L662  ?

PR #1537, opened 2020-08-04
Fixed issue #16217: Survey will not show if "inherit on" is chosen.

gabrieljenik:
Removed "public" scope from Survey model, because it doesn't work with inheritance (DBCriteria can't manage inheritance recurrency). Added a "findAllPublic" method to the model to use in public listing.
Also fixed survey overview.

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -460,7 +460,6 @@ class Survey extends LSActiveRecord
                 ':now2' => dateShift(date("Y-m-d H:i:s"), "Y-m-d H:i:s", Yii::app()->getConfig("timeadjust"))
             )
             ),
-            'public' => array('condition' => "listpublic = 'Y'"),
             'registration' => array('condition' => "allowregister = 'Y' AND startdate > :now3 AND (expires < :now4 OR expires IS NULL)", 'params' => array(
                 ':now3' => dateShift(date("Y-m-d H:i:s"), "Y-m-d H:i:s", Yii::app()->getConfig("timeadjust")),
                 ':now4' => dateShift(date("Y-m-d H:i:s"), "Y-m-d H:i:s", Yii::app()->getConfig("timeadjust"))
@@ -2075,6 +2074,22 @@ return $s->hasTokensTable; });
     public function getOwnerUserName()
     {
         return isset($this->owner["users_name"]) ? $this->owner["users_name"] : "";
+    }
+
+    /**
+     * Find all public surveys
+     * @return Survey[]
+     */
+    public static function findAllPublic()
+    {
+        $surveys = self::model()->findAll();
+        $surveys = array_filter(
+            $surveys,
+            function($s) {
+                return $s->isListPublic;
+            }
+        );
+        return $surveys;
     }
 
 }
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -460,6 +460,7 @@ class Survey extends LSActiveRecord
                 ':now2' => dateShift(date("Y-m-d H:i:s"), "Y-m-d H:i:s", Yii::app()->getConfig("timeadjust"))
             )
             ),
+            'public' => array('condition' => "listpublic = 'Y'"),
             'registration' => array('condition' => "allowregister = 'Y' AND startdate > :now3 AND (expires < :now4 OR expires IS NULL)", 'params' => array(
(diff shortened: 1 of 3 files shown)


sushmanadendlals:

olleharstedt:
Please resolve conflict.

application/models/Survey.php
```diff
@@ -2069,4 +2068,20 @@ public function setToInherit()
         }
     }
 
+    /**
+     * Find all public surveys
+     * @return Survey[]
+     */
+    public static function findAllPublic()
+    {
+        $surveys = self::model()->findAll();
```
olleharstedt, line 2077: This won't scale. Filter in the query. :laughing:

olleharstedt: Also it doesn't seem to be called statically in the change above?

gabrieljenik: > This won't scale. Filter in the query.

It is not possible to flter in the query as `public` could be an inherited value. That's the main reason of the bug. I could try to narrow the query a bit more still (Public and Inherited). Though, a loop for reviewing one by one the inherited ones will be needed

olleharstedt: Can it be both inactive and public?

olleharstedt: Looks like it filters by active? `Survey::model()->active()->open()->with('languagesettings')->findAllPublic()`

You should probably change the entire line, current way makes no sense.

olleharstedt: Not sure what `active()` actually calls here.

gabrieljenik: I was trying to substitute the AR scope with a new method that goes through attributes.
Seems I missed something. Will review it.

olleharstedt: Thanks, ping me when you want a new review.

gabrieljenik:
@sushmanadendlals I think this may need a retesting after @olleharstedt approves it

application/models/Survey.php
```diff
@@ -2069,4 +2068,22 @@ public function setToInherit()
...
     }
 
+    /**
+     * Find all public surveys
+     * @return Survey[]
+     */
+    public function findAllPublic()
+    {
+        $oCriteria = new CDbCriteria();
+        $oCriteria->condition = "listpublic = 'Y' or listpublic = 'I'";
```
olleharstedt, line 2078: Also check if it's active?

olleharstedt (changes requested):

olleharstedt:
Hey Gabriel, you also need to pull upstream and merge into this branch, to solve the conflict with the survey model.

application/models/Survey.php
gabrieljenik: That's aleady included on the scopes (filters) before on the chain, which are taken into account by doing `$this->findAll()`.

gabrieljenik:
> also need to pull upstream and merge into this b

Done!

gabrieljenik:
Please, let's remove the "testedOk" label

olleharstedt:
> Please, let's remove the "testedOk" label

No, I need this to make sure Sushma has tested it successfully.

olleharstedt:
Or wait, you mean we should test again? Yes, good point.

application/models/Survey.php
olleharstedt: Are you sure...?

gabrieljenik: Yes. I mean, we are changing the `findAll()` by a `findAllPublic()`, which calls internally `findAll` with a custom criteria.

Goging deeper: `findAll` calls `query()`, which in turns calls `applyScopes`, which applies prior scopes to the latest criteria. https://www.yiiframework.com/doc/api/1.1/CActiveRecord#query-detail

You can see debug in picture attached.
image
From the watch: `$c->condition = "(active = 'Y') AND ((startdate <= :now1 OR startdate IS NULL) AND (expires >= :now2 OR expires IS NULL))"`

olleharstedt: Oh God, so much "spooky action at a distance". Thanks for having good insight!

sushmanadendlals:
Tested the scenario's below:
1. When List survey Publicly in Publication & access control settings is set 'On', Listed Publicly in overview is set to 'Yes'.
2. When List survey Publicly in Publication & access control settings is set 'Inherit on', Listed Publicly in overview is set to 'Yes'.
3. When List survey Publicly in Publication & access control settings is set 'Off', Listed Publicly in overview is set to 'No'.

sushmanadendlals:
Working as expected

PR #1542, opened 2020-08-10
Fixed issue #16483: Adding condition triggers bulk update for complete database

gabrieljenik:
Function updated to get the survey ID from the question if surveyId is null and a qid is specified.

UpgradeConditionsToRelevance receives two params: surveyId and qid. If the qid is specified, it should only update that question (as per the function doc). But, if surveyId is null, it loops trough all the surveys, which only makes sense if no qid is specified.


olleharstedt:
Good fix!

PR #1543, opened 2020-08-10
Fixed issue #16566: We cannot create multiple choice questions 

gabrieljenik:
Random characters sometimes could come on qid when creating new questions as those are used by vue to identify internally new rows.

Before issue #16469, QuestionEditorController tried to find the questions using 'findByPk', which automatically casts the value to int, removing characters. 

Now, 'find' is used because the survey ID is added to the search criterias.
qid is not casted and no objects are found.

When saving a new question, Question object is empty, and the resulting SQL fails in Postgres because of the data type.
Solved by casting 'qid' to int.

### DIFF
--- a/application/controllers/QuestionEditorController.php
+++ b/application/controllers/QuestionEditorController.php
@@ -268,7 +268,7 @@ class QuestionEditorController extends LSBaseController
         );
 
         // Store changes to the actual question data, by either storing it, or updating an old one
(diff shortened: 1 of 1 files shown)


sushmanadendlals:
WORKING AS EXPECTED, please refer mantis for more details

olleharstedt:
> Random characters sometimes could come on qid when creating new questions as those are used by vue to identify internally new rows.

Uh, how is this even possible? Strange behaviour.

PR #1544, opened 2020-08-10
Fixed issue #15864: When deleting a question at LS 4 the system switches to the "Settings" menu

gabrieljenik:
The 'delete' method now supports the 'redirectTo' parameter: to question list or group overview. Depending on that, the sidebar is set as well

### DIFF
--- a/application/controllers/admin/questions.php
+++ b/application/controllers/admin/questions.php
@@ -960,9 +960,12 @@ class questions extends Survey_Common_Action
      * @access public
      * @param int $surveyid
      * @param int $qid
+     * @param int $qid
+     * @param int $massAction
+     * @param string  $redirectTo Redirect to question list ('questionlist' or empty), or group overview ('groupoverview').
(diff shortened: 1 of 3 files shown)


sushmanadendlals:
Working as expected, Please refer Mantis for more details

application/controllers/admin/questions.php
```diff
@@ -988,6 +991,29 @@ public function delete($surveyid=null, $qid=null, $gid = 0, $massAction = false)
             $gid_search = null;
         }
 
+        if (empty($redirectTo)) {
+            $redirectTo = Yii::app()->getRequest()->getPost('redirectTo', 'questionlist');
+        }
+
+        if ($redirectTo == 'groupoverview') {
+            $redirect = Yii::app()->createUrl(
```
olleharstedt, line 999: Slightly unclear variable name, feel free to use `$redirectUrl` in the future.

PR #1545, opened 2020-08-10
Fixed answer #16441: Problem in field type = List (Dropdown),

gabrieljenik:
The language attributes for the answer options were not retrieved.
Merged the answer attributes with the corresponding answerl10ns attributes in getQuestionArray (SurveyDynamic.php).


sushmanadendlals:
Issue is not fixed Completely, Please refer mantis for more  details

sushmanadendlals:
As per the discussion the answer options appear on the PDF now, the above mentioned issue will be solved later as it exists in LS3 as well

PR #1548, opened 2020-08-11
Fixed issue #16335: Cannot see more label sets if the list is long

gabrieljenik:
Set a max-height of 290px for the list (10 items).


sushmanadendlals:

PR #1549, opened 2020-08-11
Fixed issue #16504: Lable Sets Assessment Value Not Copied to Answer options I have created a label set comprising 5 options with assessment value

gabrieljenik:
In LabelSets.vue, assessment_value was ignored while copying. Now it's copying that value when both the source and target have the assessment_value property.


sushmanadendlals:

olleharstedt:
Merged manually.

PR #1551, opened 2020-08-17
Fixed issue #16549: Error message is shown while trying to delete and reupload file using File upload

gabrieljenik:
Removed the iframe after receiving the result.
Also fixed the order of _getResponse and form.submit() calls, which doesn't seem to affect this case but was wrong anyway: _getResponse only adds the load event for the iframe, and it was being called after form.submit(). The event should be defined before submitting the form.


sushmanadendlals:
Issue not fixed. Please refer Mantis for more details

gabrieljenik:
The fix for this should be found with #1552

PR #1555, opened 2020-08-18
Fix https://bugs.limesurvey.org/view.php?id=16083

level420:
After uncommenting this line editing and deleting assessements works again

Fixed issue #16083:


olleharstedt:
Thank you! I will check why it was commented out in the first place.

sushmanadendlals:

PR #1557, opened 2020-08-18
Fixed issue #16591 to #16594: Lime Survey 4.3.10+200812 Stored Cross Site Scripting (Users)

gabrieljenik:
Fixed issue #16591: Lime Survey 4.3.10+200812 Stored Cross Site Scripting (Survey Menu)
Fixed issue #16592: Lime Survey 4.3.10+200812 Stored Cross Site Scripting (User Details)
Fixed issue #16593: Lime Survey 4.3.10+200812 Stored Cross Site Scripting (UserName)
Fixed issue #16594: Lime Survey 4.3.10+200812 Stored Cross Site Scripting (Permission Roles)

Flattening everywhere

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -109,6 +109,10 @@ class UserManagementController extends LSBaseController
         }
 
         $aUser = Yii::app()->request->getParam('User');
+        // Sanitize full name to prevent XSS attack
+        if (isset($aUser['full_name'])) {
+            $aUser['full_name'] = flattenText($aUser['full_name'], false, true);
+        }
         $passwordTest = Yii::app()->request->getParam('password_repeat', false);
         if (!empty($passwordTest)) {
             if ($passwordTest !== $aUser['password']) {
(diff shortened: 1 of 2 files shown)


Shnoulle:
I really dislike to filter/encode in controller for view.

Can we encode it or filter it in view/vue ?

gabrieljenik:
Hmm Not sure I agree. You are saying contents should be saved as inputted and then filtered while showing them?
As for this fix, which was needed quick, I just took the same approach when editting as it was before for creation.

I think that filtering on the view (when data is displayed I mean) it is a bigger thing with more impact that should be addressed on a different story.

eddylackmann (approved):

Shnoulle:
> Hmm Not sure I agree. You are saying contents should be saved as inputted and then filtered while showing them?

Saved as inputted : already done for superadmin, filtered or encoded when shown , yes. Only when shown and done by view, not by controller.

>  As for this fix, which was needed quick, I just took the same approach when editting as it was before for creation.

Yes, but maybe old approch was bad …

> I think that filtering on the view (when data is displayed I mean) it is a bigger thing with more impact that should be addressed on a different story.

Encoding OR filtering (if needed) **must** be done in view for string. Then view can choose to show it encoded or filtered.

For example , for survey title is `<strong>The title</strong><script>Some script</script>`

1. in public : show it not encoded and not filtered
2. not public : can choose to show it encoded (default for Yii grid for example) , flatten or just filtered. And can show it as h& : flatten and in detail as encoded :).

Shnoulle:
PS : i take the topic for next meeting ;) : where string must be filtered/encoded/flatten etc ;)

gabrieljenik:
> Saved as inputted : already done for superadmi

Keep in mind the fields we are dealing with are et inputs which were never expected to have HTML and which don't have any filtering at all. Neither for superadmin nor for regular users.

Shnoulle:
In my opinion : 

- `$aSurveymenu['description']` MUST allow HTML (or markdown)
- `$aSurveymenu['title']` can allow html.

PR #1558, opened 2020-08-19
Fixed issue #16313: Issues saving/modifying answers: "Answer codes must be unique by question"

gabrieljenik:
Replaced comparison operators as they could end up having different types


Shnoulle:
:+1:

vitormattos:
I applyed this patch and the problem was solveld

eddylackmann (approved):

PR #1564, opened 2020-08-24
Fixed issue #16604: Adding condition triggers bulk update for complete database

gabrieljenik:
If surveyId is not given but qid is, fetch surveyd from the question

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -8616,13 +8616,26 @@ report~numKids > 0~message~{name}, you said you are {age} and that you have {num
         public static function UpgradeQuestionAttributes($changeDB=false,$iSurveyID=NULL,$onlythisqid=NULL)
         {
             $LEM =& LimeExpressionManager::singleton();
-            if (is_null($iSurveyID))
+            $aSurveyIDs = array();
+
+            // If survey given, set target surveys.
+            if (!is_null($iSurveyID)) {
+                $aSurveyIDs=array($iSurveyID);
+            }
+
+            // If no surveys fetched but question given, fetch survey from question
+            if (empty($aSurveyIDs) && $onlythisqid) {
+                $oQuestion = Question::model()->find("qid = :qid", array(":qid"=> $onlythisqid);
+                if($oQuestion) {
+                    $aSurveyIDs = array($oQuestion->sid);
+                }
+            }            
+
+            // If no surveys fetched still, go for all surveys
+            if (empty($aSurveyIDs))
             {
                 $sQuery='SELECT sid FROM {{surveys}}';
                 $aSurveyIDs = Yii::app()->db->createCommand($sQuery)->queryColumn();
-            }
-            else{
-                $aSurveyIDs=array($iSurveyID);
             }
 
             $attibutemap = array(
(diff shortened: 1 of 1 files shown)


gabrieljenik:
Same fix could be applied here.
https://github.com/LimeSurvey/LimeSurvey/blob/bd038776d000dfd4fc629d5f3455a2548e3ec10b/application/helpers/expressions/em_manager_helper.php#L8618

Not sure it is worth it, as it is only called from here.
I would do it, what do you think? @olleharstedt

Shnoulle:
A fix on `$LEM =& LimeExpressionManager::singleton(); ` ???

Shnoulle:
Oh no : you mean : usage of `onlythisqid` variable :).

:+1:

gabrieljenik:
Yes. Not sure it is worth it. What do you think?

Shnoulle:
> Yes. Not sure it is worth it. What do you think?

```
            if (is_null($iSurveyID) && $onlythisqid) {
                $oQuestion = Question::model()->find("qid = :qid", array(":qid"=> $onlythisqid);
                if($oQuestion) {
                    $iSurveyID = array($oQuestion->sid);
                }
            }
```
just before https://github.com/LimeSurvey/LimeSurvey/blob/bd038776d000dfd4fc629d5f3455a2548e3ec10b/application/helpers/expressions/em_manager_helper.php#L8619

This fix without update. But in my opinion : must check if $onlythisqid and $iSurveyID are compatible, throw error if not.

gabrieljenik:
So far I think this can be closed. This will probably not be tested nor merged. Already merged in master. So, let me know if this should be closed.

olleharstedt:
Don't close yet, please. LS3 might live long still.

olleharstedt:
Alternatively we can cherry-pick the fix from master into 3.x-LTS.

gabrieljenik:
This is the patch done for 3.x LTS. Not cherry picked but done with the same spirit. Not sure you can cherry pick.
The biggest issue is testing this is hard

olleharstedt:
But you can test it doesn't break anything?

gabrieljenik:
That's the biggest part to test.

olleharstedt:
Travis is red?

gabrieljenik:
Seems so, but I believe it is not because of the code.

```
0.10s$ wget localhost
--2020-08-25 17:08:38--  http://localhost/
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:80... connected.
HTTP request sent, awaiting response... 500 Internal Server Error
2020-08-25 17:08:38 ERROR 500: Internal Server Error.
```

olleharstedt:
Or? :) 

PHP Parse error:  syntax error, unexpected ';' in application/helpers/expressions/em_manager_helper.php on line 8628

gabrieljenik:
Strange. This is so old and I am sure we have been through this and did testing!
Nevermind, will review it

gabrieljenik:
Pushed a fix in the code.
Let's see what travis has to say :)

gabrieljenik (commented):
> Or? :)
> 
> PHP Parse error: syntax error, unexpected ';' in application/helpers/expressions/em_manager_helper.php on line 8628

Fixed. Travis green

PR #1565, opened 2020-08-26
Fixed issue #16565: Transfer surveys window to be displayed to ask to transfer surveys to other users while deleting the survey owner

gabrieljenik:
Add transfer to screen, shown after confirmation modal

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -218,7 +218,7 @@ class UserManagementController extends LSBaseController
      * @return void|string
      * @throws CException
      */
-    public function actionDeleteConfirm()
+    public function actionDeleteUser()
     {
         if (!Permission::model()->hasGlobalPermission('users', 'delete')) {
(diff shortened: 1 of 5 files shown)


application/controllers/UserManagementController.php
```diff
@@ -228,16 +228,85 @@ public function actionDeleteConfirm()
...
+                                    'users' => $aUsers,
+                                ],
+                                true
+                            ),
+                        ]
+                    ]
+                );
+            }
+        } else {
+            // If $transferTo is not null, transfer the surveys
```
Shnoulle, line 269: If $transferTo is not null but false or empty ? It's a dropdown to select it ?

gabrieljenik: Fixed it. now checks for empty before.
Yes, it is a dropdown to select it. That's on the view.

eddylackmann (approved):

PR #1569, opened 2020-08-31
Ckeditor 4, Data Policy, Label Sets, Email Templates, Question Organizer

gabrieljenik:
DEV: Added CKEditor v4
DEV: Restored LS3 Global Settings for HTML Editor
DEV: Rebuild Data Policy Screen
DEV: Rebuild Labelsets Screen
DEV: New HTML Modal Editor: Different than the Regular HTML Editor on PopUp mode, which opens on a new screen.
DEV: Modal HTML Editor for LabelSets


gabrieljenik:
@olleharstedt For your reviewal

Shnoulle:
@gabrieljenik CKeditor4 can be tested here ?

olleharstedt:
Nice :balloon:

gabrieljenik:
> @gabrieljenik CKeditor4 can be tested here ?

Actually, I think I am missing a file to push.
Sorry, will be back with you soon

gabrieljenik:
Label sets screen needs some refinement.
As for example: 
- When adding new items on the label set, the popup editor is used. Later on the modal one. We should be using always the modal one.
- When adding new items, beahviour is fuzzy when loosing focus.

Will update soon

gabrieljenik:
When testing, please remember to reset the global configuration for the html editor, as that option has been restated also.

gabrieljenik:
I fon't like how labelset adding screen flow is, still is the same as in LS3. We may enhance it, although I am not sure you want to do it now.

olleharstedt:
> We may enhance it, although I am not sure you want to do it now.

Not now, getting LS4 production ready is top priority.

gabrieljenik:
Merged LS/develop into this.

# About porting question orgaizing
- Had to add jquery-ui to AdminTheme.php. Otherwise, it doesn't work when loaded by ajax (jquery-ui goes on HEAD).
- Added Save button using "showSaveButton" from Top Bar Vue component. In LS3, there is also a Save and Close button. Could be added (if it doesn't already work) when Top Bar is reverted.
- There are issues with beforeunload, already present in LS3. Question organizer sets the beforeunload event to show an alert when leaving the page without saving, but there are two issues:
1) Doesn't work when loaded by ajax because it uses window.onload (not needed)
2) Submit is not properly recognized, so the message is also shown when Save button is clicked.

Shnoulle:
> Had to add jquery-ui to AdminTheme.php. Otherwise, it doesn't work when loaded by ajax (jquery-ui goes on HEAD).

Shit, we still use jquery-ui …

I don't check other part but know there are a lot opf issue about «««ajax»»» badly integrated in LimeSurvey …

gabrieljenik:
> I don't check other part but know there are a lot opf issue about «««ajax»»» badly integrated in LimeSurvey …

Something we have been noticing is that pjax doesn't work well with scripts registered as packages. But if we register scripts as registerScript, it works well. Still, this is a good point. Is pjax going to be used on the new LS4? or the whole page will be refreshed all time (old fashioned) @olleharstedt  @Shnoulle

Shnoulle:
registerPackage can be extended … and it must be **before** adding ajax as stable … 

See https://bugs.limesurvey.org/view.php?id=16558 and https://bugs.limesurvey.org/view.php?id=16559

olleharstedt:
We didn't discuss pjax yet, needs to be checked what it does and if it actually brings some value to the customer.

olleharstedt:
This is the wrong URL in the test: 

> "FAILED viewing createSurvey on route survey/sa/newsurvey, full url http://localhost/index.php?r=admin/survey/sa/newsurvey"

Correct URL is

> http://localhost/index.php?r=surveyAdministration/newSurvey

It will be fixed automatically after you resolve the merge conflicts.

gabrieljenik:
@olleharstedt Why the route for surveyadministration uses a different survey i parameter? isn't that going to broke security matters which could be standarized?

gabrieljenik:
> @olleharstedt Why the route for surveyadministration uses a different survey i parameter? isn't that going to broke security matters which could be standarized?

Also, friendly urls I think could have got broken. Have to double check. Not sure if you already have it on your radar

gabrieljenik:
At last, sorry, I need some guidelines on how to move actions to the new surveyAdministration controller.

actionOrganize needs to be migrated.
I stumbled with changed on how to pick parameters and with views. I was able to solve that.

But later I got errors while rendering topbar, which seems to be still in Vue.
Could be good to review that.

gabrieljenik:
@Trischi80 Please, can you look at my comments below? I could use your input and I think some could be interesting for you. Thansk!

olleharstedt:
Thanks, Gabriel!

PR #1571, opened 2020-09-02
Fixed issue #16588: False warning about the screen resolution

gabrieljenik:
The window size used for validation was only retrieved on "document ready" event of the login page.
Added a handler for the window resize event to keep the 'width' field updated until form sumbission.

### DIFF
--- a/application/views/admin/authentication/login.php
+++ b/application/views/admin/authentication/login.php
@@ -154,4 +154,7 @@ $( document ).ready(function() {
     $('#user').focus();
     $("#width").val($(window).width());
 });
+$( window ).resize(function() {
+    $("#width").val($(window).width());;
+});
 </script>


Shnoulle:
Why not use RWD to show/hide an `alert` box ?

1. Remove the alert-box from notify
2. At https://github.com/LimeSurvey/LimeSurvey/blob/4698bac29da586f9f81a6b9985466edb7fdbc6ff/application/views/layouts/adminmenu.php#L10 add an alert element
3. Hide it in CSS for screen bigger than XXXX

gabrieljenik:
I think it is a great idea, still not the perfect timing to implement it.
As many other times, looks like more as a re-do than a bugfix.

Will let @olleharstedt decide how to handle it.

Shnoulle:
@olleharstedt I can check for the redo : it's more fun ;)

Shnoulle:
PS : 
1. 1920 screen-with : not OK
Capture d’écran du 2020-09-03 15-28-05

2. 800 screen-width : OK
Capture d’écran du 2020-09-03 15-31-05

USage of em for media is better then pixels ;)

olleharstedt:
No idea, still busy with invoices. :weary:

eddylackmann:
we can first provide this fix . I will put it in our todo list so we can redo it later :) !

eddylackmann (approved):

Shnoulle:
> we can first provide this fix . I will put it in our todo list so we can redo it later :) !

Yes sure, a link to the todo ?

PR #1574, opened 2020-09-03
Fixed issue #16633: Error message when changing the data policy message language unless I check the box

gabrieljenik:
Making a form of its own for the language changer of the first page. This form as only have the language will not trigger server side validation of the policy. This is how it is in LS4

### DIFF
--- a/themes/survey/vanilla/views/subviews/navigation/language_changer_first_page.twig
+++ b/themes/survey/vanilla/views/subviews/navigation/language_changer_first_page.twig
@@ -17,6 +17,24 @@
     @see: http://www.yiiframework.com/wiki/48/by-example-chtml/#hh5
 
 #}
-{{ include('./subviews/navigation/language_changer.twig') }}
+{% if aSurveyInfo.alanguageChanger.show == true %}
+    {% set htmlOptions = {
+        'id'          : 'firstpage-changelang',
+        'autocomplete': 'off',
+        'class'       : 'form',
+    } %}
+
+    <!-- Start of language change Form-->
+    {{ C.Html.form(
+        ( aSurveyInfo.surveyUrl ),
+    'post',
+    (htmlOptions)
+    ) }}
+
+
+    {{ include('./subviews/navigation/language_changer.twig') }}
+
+    </form>
+{% endif %}
 
 
--- a/themes/survey/vanilla/views/subviews/navigation/language_changer_first_page.twig
+++ b/themes/survey/vanilla/views/subviews/navigation/language_changer_first_page.twig
@@ -17,24 +17,6 @@
     @see: http://www.yiiframework.com/wiki/48/by-example-chtml/#hh5
 
 #}
-{% if aSurveyInfo.alanguageChanger.show == true %}
-    {% set htmlOptions = {
-        'id'          : 'firstpage-changelang',
-        'autocomplete': 'off',
-        'class'       : 'form',
-    } %}
-
-    <!-- Start of language change Form-->
-    {{ C.Html.form(
-        ( aSurveyInfo.surveyUrl ),
-    'post',
-    (htmlOptions)
-    ) }}
-
-
-    {{ include('./subviews/navigation/language_changer.twig') }}
-
-    </form>
-{% endif %}
+{{ include('./subviews/navigation/language_changer.twig') }}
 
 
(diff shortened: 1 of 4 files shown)


themes/survey/vanilla/views/subviews/navigation/language_changer_first_page.twig
```diff
@@ -17,6 +17,24 @@
...
+    {% set htmlOptions = {
+        'id'          : 'firstpage-changelang',
+        'autocomplete': 'off',
+        'class'       : 'form',
+    } %}
+
+    <!-- Start of language change Form-->
+    {{ C.Html.form(
+        ( aSurveyInfo.surveyUrl ),
+    'post',
```
Shnoulle, line 30: It can be a get here ? Think it's best if it's a get.

Shnoulle (commented):
To be checked (and fixed)

- [ ] token by URI or POST value
- [ ] prefilling url https://manual.limesurvey.org/URL_fields#Start_URL
- [ ] captcha

themes/survey/vanilla/views/subviews/navigation/language_changer_first_page.twig
gabrieljenik: Yes, I think so

gabrieljenik:
> To be checked (and fixed)
> 
> * [ ]  token by URI or POST value
> * [ ]  prefilling url https://manual.limesurvey.org/URL_fields#Start_URL
> * [ ]  captcha

This fails when using the language changer from the top menu. Will review again

Shnoulle:
> This fails when using the language changer from the top menu. Will review again

Dis it work currently ? in 3.X and 4.X ?

gabrieljenik:
> This fails when using the language changer from the top menu. Will review again

Now top menu doesn't show error.

gabrieljenik:
> To be checked (and fixed)
> 
> * [ ]  token by URI or POST value
> * [ ]  prefilling url https://manual.limesurvey.org/URL_fields#Start_URL
> * [ ]  captcha

Tested with all. Still sending via POST as the form has all the info, as for example the move parameter, which is used.

gabrieljenik:
> > This fails when using the language changer from the top menu. Will review again
> 
> Dis it work currently ? in 3.X and 4.X ?

This fix is for LS3. I tested in LS3.
On LS4 the error was not reproducible.

Shnoulle:
>  On LS4 the error was not reproducible.

Yes, but i mean : prefill value, token etc ;) .
And to be checked with JS/ without JS too :(

gabrieljenik:
> And to be checked with JS/ without JS too :(

Wihtout JS? Does LS works with no JS support?

Shnoulle:
Public part : yes a lot , i work for it …

gabrieljenik:
> Public part : yes a lot , i work for it …

Haven't tested with no JS support. How do you approach that testing? Any chrome settings could be usefull?

Let me ask, why? JS is a standard!
Isn't that an unnecessary overhead and a limit to development of the tool?

Shnoulle:
> Let me ask, why? JS is a standard!
> Isn't that an unnecessary overhead and a limit to development of the tool?

Not everywhere, and when you connecte with a really light connexion :; it's really great

https://softwareengineering.stackexchange.com/questions/26179/why-do-people-disable-javascript

But the best one is https://softwareengineering.stackexchange.com/a/25996
https://en.wikipedia.org/wiki/Progressive_enhancement

gabrieljenik:
> > Let me ask, why? JS is a standard!
> > Isn't that an unnecessary overhead and a limit to development of the tool?
> 
> Not everywhere, and when you connecte with a really light connexion :; it's really great
> 
> https://softwareengineering.stackexchange.com/questions/26179/why-do-people-disable-javascript
> 
> But the best one is https://softwareengineering.stackexchange.com/a/25996
> https://en.wikipedia.org/wiki/Progressive_enhancement

I see. I like the idea, but is that something to focus right now...?
I will test it..

Shnoulle:
> I see. I like the idea, but is that something to focus right now...?

It's like security … best is to start always by simple part, and improve after.

gabrieljenik:
When testing with no JS the upper menu doesn't ahow up.
Only the language changer from the body. Languages are changed good.

Bug is not reproducible in LS4.
Tested it with no JS as well.

Shnoulle:
Great !

PR #1575, opened 2020-09-03
Fixed issue #16562: Bar charts at statistics start with min value=1 but has to be min value=0

gabrieljenik:
Setting options to explicitly start on 0


eddylackmann (approved):

Shnoulle:
It's not about tis issue, but there are no way to set whole default in one step using [Chart.defaults](https://www.chartjs.org/docs/latest/configuration/) ?

PR #1577, opened 2020-09-03
Fixed issue #15615: Mandatory multiple choice questions are not assigned a red background if not being answered

gabrieljenik:
Added new code on tip event handler for updating question-container, evaluating status of its tips and then add/remove the error class

### DIFF
--- a/assets/packages/template-core/template-core.js
+++ b/assets/packages/template-core/template-core.js
@@ -81,8 +81,8 @@ var TemplateCoreClass = function () {
                         $parent.removeClass('hide-tip', 1);
                         $parent.addClass('tip-was-hidden', 1);
                     }
-                    $questionContainer = $(this).parents('div.question-container');
-                    $questionContainer.addClass('input-error'); /* No difference betwwen error after submit and error before submit : think (Shnoulle) it's better to have a difference */
+                    var questionContainer = $(this).parents('div.question-container');
+                    updateQuestionClassFromTips(questionContainer);
                 });
 
                 $(this).on('classChangeGood', function () {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Like the move to function, no new event here , no ?

gabrieljenik:
> Like the move to function, no new event here , no ?

Thought about new event. This seems easier. Do you see any upside on using events for that?

Shnoulle:
You said «Added new event handler for question-container,» , i don't see a new event here. It was just a question. 

Since it's template-core : i already remove this in theme … My opinion : we must not have the same color for 'submitted' and not submitted errors :)

gabrieljenik:
> You said «Added new event handler for question-container,» , i don't see a new event here. It was just a question.

Fixed description

> Since it's template-core : i already remove this in theme … My opinion : we must not have the same color for 'submitted' and not submitted errors :)

I don't understand sorry. You already have something on your theme which differentiates submitted and not submitted (dirty fields and non dirty fileds)?

Shnoulle:
> I don't understand sorry. You already have something on your theme which differentiates submitted and not submitted (dirty fields and non dirty fileds)?

I just don't update wrapper with JS solution. Project is to add it, but must find times.

cdorin93:
@gabrieljenik , could you please test why Travis is red? thanks!

gabrieljenik:
I will

gabrieljenik:
AutoCalcTest::testBasic() is erroring here:
`$this->assertEquals('Body Mass Index (BMI) is 25.308.', $bmi, $bmi);`

It shouldn't fail. Still, the comaprison is wrong! Is comparing the result with itself, not a calculated value.

About the other error RankingFilterHideShowTest::testRankingFilterShown, not sure why. 
We will go deeper into it.

gabrieljenik:
We found the issue I think. We will push the fix.

eddylackmann (approved):

PR #1584, opened 2020-09-08
Fixed issue #15994: Batch-Edit shows "lskeep" after modification of email status

gabrieljenik:
Code Review.
Created a checkbox for "emailstatus" edit only
Don't update "email" if email is empty even if checkbox is checked (no way to set [empty string] to email with Batch edit
Removed placeholder for emailstatus. It was confusing.


maziminke:
@cdorin93 Why wasn't this one added to today's release?

cdorin93:
I forgot about it - we will do another release tomorrow

PR #1586, opened 2020-09-10
DEV: Edit Question Test

gabrieljenik:


cdorin93:
@olleharstedt , have time to take a quick look at it?

olleharstedt:
Looks like one element can't be found?

> NoSuchElementException: Unable to locate element: #cke_question_en

olleharstedt:
If you run the test locally, skip `MOZ_HEADLESS=1` and you'll see the browser stepping through the test.

gabrieljenik:
This is failing as this test is being run against a revert-vue-js branch while it was done for LTS

olleharstedt:
> A hint: This is failing as this test is being run against a revert-vue-js branch while it was done for LTS

Is it? The branch says 3.x-LTS.

olleharstedt:
This test works perfect for me locally, in isolation.

gabrieljenik:
> > A hint: This is failing as this test is being run against a revert-vue-js branch while it was done for LTS
> 
> Is it? The branch says 3.x-LTS.

I swear I saw something about revert-vue-js
I will try increasing the wait time before the failing line.

Shnoulle:
When i have issue like this : the screenshot help me a lot …

Shnoulle:
Oups , sorry @olleharstedt already tested :)

olleharstedt:
> When i have issue like this : the screenshot help me a lot …

Yeah, but running the suite locally takes forever. T_T Can't reproduce. Grrr.

gabrieljenik:
> > A hint: This is failing as this test is being run against a revert-vue-js branch while it was done for LTS
> 
> Is it? The branch says 3.x-LTS.

I swear I saw something about revert-vue-js

olleharstedt:
Not sure how to continue to debug. Dump HTML at failure??

gabrieljenik:
I want to switch to use somethign like this: frameToBeAvailableAndSwitchToIt

olleharstedt:
> Facebook\WebDriver\Exception\NoSuchElementException: Unable to locate element: #cke_question_en

This means that the page is not loaded at all? Can you put a sleep(2) before this call? Or maybe try to find all h1 and h2 tags and dump them (will tell us which page we're on).

Shnoulle:
maybe we set config to something other tha inline HTML editor before ? Did you try to lauch all test locally @olleharstedt ?

olleharstedt:
> maybe we set config to something other tha inline HTML editor before ? Did you try to lauch all test locally @olleharstedt ?

I tried, but it's _really_ slow.

olleharstedt:
I tried to run this test with different HTML editor options, it worked for all alternatives.

gabrieljenik:
> > Facebook\WebDriver\Exception\NoSuchElementException: Unable to locate element: #cke_question_en
> 
> This means that the page is not loaded at all? Can you put a sleep(2) before this call? Or maybe try to find all h1 and h2 tags and dump them (will tell us which page we're on).

I have already tried adding a sleep(10) I think. Didn't work.
Changed the approach on how to get the div. Didn't work either.
I will try to add some echoing of html to check what is rendered.

olleharstedt:
Does the test work if you _only_ run this test? You can edit it in .travis.yml file.

gabrieljenik:
Ran good with a single test! Very strange!
`The command "DOMAIN=localhost ./third_party/bin/phpunit --filter EditQuestionTest" exited with 0`

gabrieljenik:
Testing changing the order in which the tests run

Shnoulle:
Even if update order fix the issue : we need to find what happen before ;)

gabrieljenik:
Now failed on another. I will find out. I see some collisions.

olleharstedt:
Also try running phpunit with `--testdox` to see the names of the tests being run.

olleharstedt:
Another try could be to use a separate survey to import instead of current one, so that not two tests use the one survey. Not sure it matters, but maybe.

Shnoulle:
> Now failed on another. I will find out. I see some collisions.

Yes, but this one is OK. If ChangeQuestionTemplateTest::testChangeQuestionTemplate happen after the same, it's the previous one that break …
The previous one update something but don't reset :)

Shnoulle:
> Also try running phpunit with `--testdox` to see the names of the tests being run.

Question : you see all test ? For example

```
ls\..\testOne : OK
ls\..\testTow : OK
…
```

If yes : surely a great idea :)

gabrieljenik:
> Another try could be to use a separate survey to import instead of current one, so that not two tests use the one survey. Not sure it matters, but maybe.

I think there is some kind of problem. The import at EditQuestionTest was some kind of failing. When moving it up in the order, worked good, but then the testChangeQuestionTemplate is failing. EditQuestionTest  has a TearDown method, so it is not supposed to be an issue. I will double check

> Also try running phpunit with `--testdox` to see the names of the tests being run.
Good!

gabrieljenik:
Either teardownAfterClass is not being run or self::$testSurvey is not being reset. Keep on investigating

Shnoulle:
The deletion seems partially OK … https://travis-ci.org/github/LimeSurvey/LimeSurvey/jobs/729673984 

Realy hard to fix here …

gabrieljenik:
Last test, changing the survey file, worked good.
Not sure how that made a difference. I will double check.

Also, smell, something is wrong with the cleanup.
I think the problem is we should be using static sometimes, and we are using self.

gabrieljenik:
https://bugs.limesurvey.org/view.php?id=16692

I will review what's wrong and fix it :)

gabrieljenik:
We have green Travis!

Shnoulle:
About lss file : i think it's best if they have a name in relation with the test done (or a functionality).

**Really more easy** for other dev to check and test when try to update something.

olleharstedt:
> About lss file : i think it's best if they have a name in relation with the test done (or a functionality).

Yeah, I agree with this, Denis.

olleharstedt:
> Also, smell, something is wrong with the cleanup.
> I think the problem is we should be using static sometimes, and we are using self.

Yeah, clean-up is hacky. One reason I want to move more to unit-tests with mocking. But integrity tests like these are needed too. Thanks for fixing!!

Shnoulle:
next time : please, clear name for lss …

gabrieljenik:
> next time : please, clear name for lss …

Yes, I iwas just keeping status quo. Does it worth to rename all current files and deduplicate if any is duplicate?

Shnoulle:
> Yes, I iwas just keeping status quo.

I just speak for new files, current status quo i set a clear name for new test … , it's best to review previous one, but take time.

PR #1589, opened 2020-09-11
New feature #12560: enable video in spite of active xss filtering

gabrieljenik:
Added new configuration to the HtmlPurifier.
Extended the standard HtmlPurifier so it exposes the config in a public method.

### DIFF
--- a/application/core/LSYii_HtmlPurifier.php
+++ b/application/core/LSYii_HtmlPurifier.php
@@ -1,0 +1,30 @@
+<?php if (!defined('BASEPATH')) {
+    exit('No direct script access allowed');
+}
+/*
+ * LimeSurvey
+ * Copyright (C) 2007-2011 The LimeSurvey Project Team / Carsten Schmitz
+ * All rights reserved.
+ * License: GNU/GPL License v2 or later, see LICENSE.php
+ * LimeSurvey is free software. This version may have been modified pursuant
+ * to the GNU General Public License, and as distributed it includes or
+ * is derivative of works licensed under the GNU General Public License or
+ * other free or open source software licenses.
+ * See COPYRIGHT.php for copyright notices and details.
+ *
+ */
+
+class LSYii_HtmlPurifier extends CHtmlPurifier
+{
+
+    /**
+	 * Get the config object for the HTML Purifier instance.
+	 * @return mixed the HTML Purifier instance config
+	 */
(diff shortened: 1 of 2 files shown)


application/core/LSYii_HtmlPurifier.php
```diff
@@ -0,0 +1,30 @@
...
+ * License: GNU/GPL License v2 or later, see LICENSE.php
+ * LimeSurvey is free software. This version may have been modified pursuant
+ * to the GNU General Public License, and as distributed it includes or
+ * is derivative of works licensed under the GNU General Public License or
+ * other free or open source software licenses.
+ * See COPYRIGHT.php for copyright notices and details.
+ *
+ */
+
+class LSYii_HtmlPurifier extends CHtmlPurifier
```
Shnoulle, line 17: :+1:

application/core/LSYii_Validators.php
```diff
@@ -128,6 +128,44 @@ public function xssFilter($value)
         );
         // To allow script BUT purify : HTML.Trusted=true (plugin idea for admin or without XSS filtering ?)
 
+        // Enable video
```
Shnoulle, line 131: Just a question : can not be in `LSYii_HtmlPurifier` ?

gabrieljenik (commented):
What about now?

Shnoulle:
I think only for https://github.com/LimeSurvey/LimeSurvey/pull/1589/files#diff-a832a00a927d6b1c5bb94e437eb7881fR54 but : it's just a question :)

gabrieljenik:
Sorry, what you mean? Don't understand.

Shnoulle:
Something like a enableVideo function

https://github.com/LimeSurvey/LimeSurvey/pull/1589/files#diff-a832a00a927d6b1c5bb94e437eb7881fR54-R90

```
public function enableVideo() {
    $config = $this->getConfig();
    // Do the job
}
```

gabrieljenik:
It is a possibility. Stil, I believe xss filtering should have almost always the same filters. That's why I didn't include that method.
So, this new class is more like a factory, which gives you filters based on scenarios. We could have it return based on individual settings, as for example if video is allowed or not, but I think that could be left for later, right?

Shnoulle:
Oh yes, sure !

I just ask a question, not make an advice ...

PR #1594, opened 2020-09-16
Fixed issue #16185: Typo "be have"

jmantysalo:
Fixed issue #<Mantis issue number>: 16185

### DIFF
--- a/application/models/UserGroup.php
+++ b/application/models/UserGroup.php
@@ -67,7 +67,7 @@ class UserGroup extends LSActiveRecord
                 'length',
                 'min' => 1,
                 'max' => 20,
-                'tooShort' => gt("Name should be at least have one sign"),
+                'tooShort' => gt("Name should have at least one sign"),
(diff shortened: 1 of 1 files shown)


cdorin93:
Hmm - what about: "Name should contain at least 1 character" ? :thinking: . Wouldn't that sound better? Indeed, the current form is not nice / user friendly

Shnoulle:
Name can not be empty ?

Shnoulle:
My opinion : never shown : https://github.com/LimeSurvey/LimeSurvey/blob/334af4225e4c3dcfead5adcaa7ba188cb2d35b83/application/models/UserGroup.php#L62

jmantysalo:
> My opinion : never shown :

Good point. So the error message is currently just a placeholder, if we later want to require slightly longer name.

Well, at least a good case to see if GitHub, PR and mirroring change to translations page works. If I can do this, maybe even something more usefull is doable too.

Shnoulle:
PS : i think even if you set to 3 : there are already some translation by Yii :)

PR #1595, opened 2020-09-16
Fixed issue #16650: Timing not recorded when exporting results

gabrieljenik:
Made the response to be based on the queried response and not start from scratch. Timings are there initially. If starting from scratch, we loose them.

### DIFF
--- a/application/helpers/admin/export/Writer.php
+++ b/application/helpers/admin/export/Writer.php
@@ -298,7 +298,7 @@ abstract class Writer implements IWriter
         foreach ($oSurvey->responses as $response) {
             // prepare the data for decryption
             $sTokenTableName='tokens_'.$oSurvey->id;
-            $aResponse = array();
+            $aResponse = $response;
             if (tableExists($sTokenTableName)) {
                 $oToken = Token::model($oSurvey->id);
                 $oToken->setAttributes($response, false); 


gabrieljenik:
@olleharstedt @Shnoulle Maybe for later, but I believe decription should happen on SurveyDao and not in writer. What do you think? We assure then that all data in the surveyObj is decrypted and exportable.

Shnoulle:
SurveyDao send whole data if i remind ?

>  I believe decription should happen on SurveyDao and not in writer. 

decription : you mean uncrypt ? If yes : :+1: see https://bugs.limesurvey.org/view.php?id=15669

https://bugs.limesurvey.org/view.php?id=15670 and https://bugs.limesurvey.org/view.php?id=15668 are related.

This make usage of encryption  in reponse unusuable.

Reponse ocject must decrypt i think.

gabrieljenik:
I was meaning only about exporting process.
Some of those tickets relate to other scenarios.

I think these is solved
- https://bugs.limesurvey.org/view.php?id=15669
- https://bugs.limesurvey.org/view.php?id=15670

About this https://bugs.limesurvey.org/view.php?id=15668
I believe this should depend on the profile. Info shouldn't be shown uncrypted to everyone.

Shnoulle:
I muts check for issue.

olleharstedt:
> @olleharstedt @Shnoulle Maybe for later, but I believe decription should happen on SurveyDao and not in writer. What do you think? We assure then that all data in the surveyObj is decrypted and exportable.

Don't have time to look into this right now, but I trust you and Denis. :)

gabrieljenik:
> > @olleharstedt @Shnoulle Maybe for later, but I believe decription should happen on SurveyDao and not in writer. What do you think? We assure then that all data in the surveyObj is decrypted and exportable.
> 
> Don't have time to look into this right now, but I trust you and Denis. :)

We can approach it when reviewign export on new ls4. Changing this now could lead to export plugins to stop working

Shnoulle:
> Changing this now could lead to export plugins to stop working

Dod you have check exiting export plugin ? I diodn't update mines since a lot of time … and don't really check in 4.X

olleharstedt:
Sounds good.

gabrieljenik:
> > Changing this now could lead to export plugins to stop working
> 
> Dod you have check exiting export plugin ? I diodn't update mines since a lot of time … and don't really check in 4.X

Me neither.

PR #1598, opened 2020-09-21
Revert Question Group Editor

gabrieljenik:
Reverted Vue-JS from View, Add, Edit

### DIFF
--- a/application/controllers/QuestionGroupsAdministrationController.php
+++ b/application/controllers/QuestionGroupsAdministrationController.php
@@ -47,8 +47,10 @@ class QuestionGroupsAdministrationController extends LSBaseController
      */
     protected function beforeRender($view)
     {
-        $this->aData['topBar']['type'] = 'group';
-        //$this->aData['topBar']['showSaveButton'] = true;
+        // Set topbar type if not already set
+        if(!isset($this->aData['topBar'])||!isset($this->aData['topBar']['type'])) {
+            $this->aData['topBar']['type'] = 'group';
+        }
         if (empty($this->aData['topBar']['showCloseButton'])) {
             $this->aData['topBar']['showCloseButton'] = false;
         }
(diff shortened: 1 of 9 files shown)


thedirtypanda:
I found some issues:

1. if you press at 'create group' then the view will be rendered. But there is a heading called 'Description'. It's not possible to add any description (at this point) to the group (there is no ckeditor). Please hide the description at this stage. After saving enable the visibility for the description again.

2. What happens to the topbar? Dorin told me that you need help.

gabrieljenik:
> 1. if you press at 'create group' then the view will be rendered. But there is a heading called 'Description'. It's not possible to add any description (at this point) to the group (there is no ckeditor). Please hide the description at this stage. After saving enable the visibility for the description again.

Will review.

> 2. What happens to the topbar? Dorin told me that you need help.

We just didn't spend too much time on it as we are not sure if it will be vue-js based or PHP based or how it will be the basic structure for setting up in the future. So looks weird now. Any updates on how topbar will be setup? Reverted or kept in php?

gabrieljenik:
Added a test and a review for languages

thedirtypanda:
I prefer to stick to vue JS for now. In the future it might be changing. The problem is, that we want to release fast and if we change the top bar now, it will take a while, cause the top bar will affect every page in the admin views.

If you want we can discuss this at discord, together with the other developers.

gabrieljenik:
> I prefer to stick to vue JS for now. In the future it might be changing. The problem is, that we want to release fast and if we change the top bar now, it will take a while, cause the top bar will affect every page in the admin views.
> 
> If you want we can discuss this at discord, together with the other developers.

No worries. If we will be keeping the vueJs toolbars then we will follow it.

gabrieljenik:
@thedirtypanda Please see latest commits.

cdorin93:
hey @thedirtypanda , have time to take a look today at this one?

PR #1599, opened 2020-09-22
Fixed issue #16647: Can't select seperator for CPDB export - export doesn't recognise text

gabrieljenik:
Used 'fputcsv' in CPDBExport() with default args.


Shnoulle (approved):
It work 

Before the fix
Capture d’écran du 2020-09-22 16-27-40

After the fix
Capture d’écran du 2020-09-22 16-29-06

Shnoulle:
PS : value are not encoded, it was filtered. I prefer to have it encoded , i report.

PR #1600, opened 2020-09-22
Fixed issue #16672: Can't select seperator for CPDB export - export doesn't recognise text

gabrieljenik:
Used 'fputcsv' in CPDBExport() with default args


Shnoulle:
If io don't make error : it's better for memory too ?

gabrieljenik:
> If io don't make error : it's better for memory too ?

Not sure... performance is not my strongest point!

eddylackmann (approved):

eddylackmann (approved):

Shnoulle:
We already have it for 3.X , right ? :smile_cat:

PR #1603, opened 2020-09-24
Fixed issue #16695: JSON RPC Remote Control methods are not sent as expected to server

gabrieljenik:
Restated the call method

### DIFF
--- a/application/libraries/jsonRPCClient.php
+++ b/application/libraries/jsonRPCClient.php
@@ -216,4 +216,9 @@ class jsonRPCClient
     {
         $this->enableCurl = false;
     }
+
+    public function call($method, $params)
+    {
+        return $this->__call($method, $params);
+    }
 }


Shnoulle:
We use the client ? Not only the server ?

gabrieljenik:
What do you mean?

Shnoulle:
I mean : we don't use jsonRPCClient class, we use jsonRPCServer. LimeSurey is a JSON RPC server , not a client.

And proof : the only place we call this Class : we die
https://github.com/LimeSurvey/LimeSurvey/blob/b4192e9bbb5019377334abc1e60b31a61daab83b/application/controllers/admin/remotecontrol.php#L102

Shnoulle:
Maybe it's best to remove the client since we don't uyse it in test too. We can add it in tests when we add a test.

```
[shnoulle@poledra 3LTS]$ cd application/
[shnoulle@poledra application]$ grep -r "jsonRPCClient" *
controllers/admin/remotecontrol.php:            Yii::app()->loadLibrary('jsonRPCClient');
controllers/admin/remotecontrol.php:            $client = new jsonRPCClient($serverUrl);
controllers/admin/remotecontrol.php:            // Set $client = new jsonRPCClient($serverUrl, true); to activate debug output
libraries/jsonRPCClient.php:class jsonRPCClient
[shnoulle@poledra application]$ grep -r "jsonRPCServer" *
controllers/admin/remotecontrol.php:                Yii::app()->loadLibrary('LSjsonRPCServer');
controllers/admin/remotecontrol.php:                LSjsonRPCServer::handle($oHandler);
libraries/jsonRPCServer.php:class jsonRPCServer
libraries/LSjsonRPCServer.php:    Yii::import('application.libraries.jsonRPCServer');
libraries/LSjsonRPCServer.php:    class LSjsonRPCServer extends jsonRPCServer
[shnoulle@poledra application]$ cd ..
[shnoulle@poledra 3LTS]$ cd tests/
[shnoulle@poledra tests]$ grep -r "jsonRPCClient" *
[shnoulle@poledra tests]$ 
```

Less code : less security issue. Remove unused code is always a great improvment

gabrieljenik:
The client is for remote connections, 3rd party php apps that want to connect.
Or also, intercmonunication in between LS servers.

gabrieljenik:
I agree maybe we could reorganize the code.

Shnoulle:
Then : it's not used by LimeSurvey core or test. Only if another tool include the file from LimeSurvey ?

My opinion : we don't use it ? We don't need it.

Shnoulle (changes requested):
We don't use it ? We don't need it.

Shnoulle:
If another LimeSurvey want to use jsonRPCClient : create a plugin and include your own jsonRPCClient.

Can be an helper : see https://gitlab.com/SondagesPro/coreAndTools/toolsDomDocument for example.

gabrieljenik:
I don't follow sorry. This jsonRPCClient has been in LS for years.

Shnoulle:
Yes, and ?

Shnoulle:
It's not because a bad code is here for years we need to keep it … 

@olleharstedt any advice ?

gabrieljenik:
It allows 3rd party apps to connect to LS. It is the API gateway.
Why do you want to remove ?

Shnoulle:
> It allows 3rd party apps to connect to LS. It is the API gateway.

No : the Server is the API gateway, not the client :). remove the file it work.

>  Why do you want to remove ?

We don't use it.

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/1604

gabrieljenik:
How are the 3rd party supposed to connect?
They need a client.

You are saying remove it from here. 
I say, repackage it on a differen repo maybe.

I think we agree, right?

Shnoulle:
>  They need a client.

Yes, and ? Thius client was writtedn in python, R, C## or PHP. We don't need client.

gabrieljenik:
> We don't need client.

We need to provide a suggested client for 3rd party apps to connect.
If you want someone to connect, we need to show them how to

Shnoulle:
LimeSurvey don't care about client : 

https://github.com/GruposInternet/jsonRpcClient-PHP 
https://github.com/soundasleep/jsonrpcclient
https://github.com/brushwood-field/jsonrpcclient
etc …

Shnoulle:
> We need to provide a suggested client for 3rd party apps to connect.

Manual (like for python) https://manual.limesurvey.org/RemoteControl_2_API#Example_and_helper

We have the link to https://github.com/weberhofer/jsonrpcphp

olleharstedt:
@gabrieljenik The client can be made in any programming language, with any library. But also with the one provided in LS core. Maybe we use it for tests...?

Shnoulle:
See @c-schmitz advice : https://github.com/LimeSurvey/LimeSurvey/pull/1604#issuecomment-698961404

Shnoulle:
PS : or include it totally : with own directory + README + licence  (via composer ;) ) : more easy to update (

gabrieljenik:
I think this discussion is not related to this bug.
There was code used and mantained on the repo. The code had a bug.
This is the PR for it.

Then we get to a new discussion if we should keep on including the client libraries on the repo.
That's another discussion, right?

gabrieljenik:
From my point of view, if the code is in the repo, the bug is in the repo.
The fix was really easy and the bug was introduced while maintaing that code.

After getting things straight, I believe it is OK about moving code out.
But as far as now, the code is in the repo.

There are users who rely on that code.
So if the code is broken, removing it will not fix their situation.
If it is easy engough to fix it, why not fixing it first and then removing it later.
That way users will be happier and I believe it will be more fair for them.

Shnoulle:
>  The fix was really easy and the bug was introduced while maintaing that code.

Please : a way to reproduce …

Else : my proposition is to include it but in a better way … don't put any file anywhere without source information …

Shnoulle:
`$sessionKey= $myJSONRPCClient->get_session_key( LS_USER, LS_PASSWORD );`

No need ->call here.

You update something for bad code from other dev ?

gabrieljenik:
> `$sessionKey= $myJSONRPCClient->get_session_key( LS_USER, LS_PASSWORD );`
> 
> No need ->call here.

I don't find the method `get_session_key` on the jsonRPCClient

> You update something for bad code from other dev ?
This is not fair

Shnoulle:
The issue are in test function, not in jsonRpcClient

We muts give good advice and advice working with public other jsonRpcClient version.

Shnoulle (commented):
The issues are in the test function, not in current jsonRPCClient.

We can update to the last one without make any change.

Shnoulle (changes requested):
The sample function muts be updated. Not the jsonRPCCLlient.

Give good advice not broken advice …

gabrieljenik:
This created the issue.
https://github.com/LimeSurvey/LimeSurvey/commit/ddde6a413a4635b41f298fb7e3b25e8adc7b1d2c#diff-cef1e7e070938ffe0879f6a5d42ee28cL171

Shnoulle:
> I don't find the method `get_session_key` on the jsonRPCClient

Not needed … please : read the manual and all sample put on forum https://gitlab.com/SondagesPro/coreAndTools/twigExtendByPlugins/-/snippets/1729193

gabrieljenik:
> Give good advice not broken advice …

I could have made a mistake. I don't think so, but I could.
I just feel it is not fair to say that I don't give good advice.

Shnoulle:
Sorry … but it's in the manual since years …

gabrieljenik:
OK, so there has been a bug for years on the client and on the test script. (`call` may have not been there)
After passing so many years, making things straight (removing the `call` method) may leave some people behind which have already build scripts based on the bug. 

After so many years I would say the bug is part of the accepted code.
So fixing the original bug caused that people to get their scripts not working.

Shnoulle:
1. This function is disable by a `die()`
2. The current Class you can found elsewhere don't all call …
3. The manual about remote control don't use call

> After so many years I would say the bug is part of the accepted code.

Please no … I already report REAL situation where something broke without update AI, in really minor version. Sometimes minor version broke API, but here : clearly : a function starting by a `die();`

We have already update function parameters inside model : you can not tell : leave this function, maybe someone use it in own script …
We already remove Model function too, some of my plugins are broken (and have multiple test of version) : since it's nt API usage : we must allow it.

gabrieljenik:
> This function is disable by a die()
The comment next to that function says: "Remove if you want to test this function"

I don't think this is a conversation to have here in the Github chat.
Is it so bad to restate a code that has been forever in that file and was only taken for a couple months?

c-schmitz:
To give some history:
The goal of the test code was that it runs the exact same way without change, no matter if it is instantiating the JSonRPCCLient or the XMLRPCClient.
So, the question is ,if we change the test() function, will it still work with the XMLRPCCLient. 
I think it won't (but I have not tested it) so I think that Gabriel's approach is the better one.

Besides that, please be respectful when commenting. Always assume that the other party only has the best intentions.

Shnoulle:
Sorry @gabrieljenik it's not what i mean … : i don't think you make bad code. My english is really unclear. 

Else : I confirm we need to have the most compatible code with the current way of coding. 

1. Have a function starting by a die just show we put something without make a clean system
2. We have now tests system
3. remote control must be tested
4. Then this function must be moved to tests.

Else about usage of `jsonRPCClient` : if we update for ourself : we must rename it and clearly show there are difference. Else someone can thinbk using jsonRPCClient Johannes Weberhofer he can use the same code : and it's false …

My opinion : 

- create a test to test XMLRPC call
- create another test for JsonRPC

**If we can not use a library : it's best.** but if we use a library : best is to don't update the lib.

Shnoulle (commented):
See last comment

gabrieljenik:
> Sorry @gabrieljenik it's not what i mean … : i don't think you make bad code. My english is really unclear.

All good I know all comments are with the best intentions.

> Else : I confirm we need to have the most compatible code with the current way of coding.
> 
> 1. Have a function starting by a die just show we put something without make a clean system
> 2. We have now tests system
> 3. remote control must be tested
> 4. Then this function must be moved to tests.

I agree some reorganization could be done, bot in terms of testing and library.
Maybe we can open a ticket for that.

> Else about usage of `jsonRPCClient` : if we update for ourself : we must rename it and clearly show there are difference. Else someone can thinbk using jsonRPCClient Johannes Weberhofer he can use the same code : and it's false …

This can be approached during the reorganization.
I think we can do that with out much hurt using namespaces.
\base\jsonRPCClient would be the original library.
jsonRPCClient could be the overloaded one, with some extra methods for backward compatibility and for compatibility with the XML client.

> My opinion :
> 
> * create a test to test XMLRPC call
> * create another test for JsonRPC

Agree

Have a nice weekend!

maziminke:
Guys, as a general approach to such fixes: Please keep the discussion limited to the fix itself. There clearly is an issue with the code and this PR fixes it, at least at for our project. It also seems to be backwards compatible, which is important for any API adjustments.
We should also keep the sample code, even if it is not used. it helps other users.

Of course, there can always be improvements, cleanup and the like. But for this we can create a new ticket at the bug tracker.
Maybe when thinking of LS5, we can complately clean up the API and warn users that things can be broken (though we should try hard to keep all the API functions working as before).

eddylackmann (approved):

PR #1604, opened 2020-09-25
Fixed issue #16695 : we don't use jsonRPCClient Class

Shnoulle:


Shnoulle:
LimeSurvey is a API server , not a client.

c-schmitz:
The jsonRPCClient is used by the test() function of the remotecontrol controller.
It is an easy way to check/test the RC api. So please don't remove it.

Shnoulle:
> The jsonRPCClient is used by the test() function of the remotecontrol controller.

No, there are a die() at start. I have another place to make my test (with more complex RC function) my self. 

If we need a test : best is to really create a test : https://github.com/LimeSurvey/LimeSurvey/tree/master/tests

c-schmitz:
So, if you want to remove it, then please create the tests first. Otherwise just leave it.

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/pull/1603 

I don't understand how we can fix something we don't really use …

maziminke:
> > The jsonRPCClient is used by the test() function of the remotecontrol controller.
> 
> No, there are a die() at start. I have another place to make my test (with more complex RC function) my self.
> 
> If we need a test : best is to really create a test : https://github.com/LimeSurvey/LimeSurvey/tree/master/tests

Well, **you** may have another way to test. But other may use the existing code for testing :-)

Shnoulle:
> Well, you may have another way to test. But other may use the existing code for testing :-)

External dev search od `die()` function before looking at manual ?
Even if search show the manual at 1st : https://duckduckgo.com/?q=limesurvey+api+php  ?
stackoverflow too (usage of ->function) https://stackoverflow.com/search?q=limesurvey+API+php 
etc …

Usage of https://github.com/weberhofer/jsonrpcphp 

This one : https://github.com/nuxwin/JsonRPC seem to use `execute`

c-schmitz:
As mentioned this will not be merged. It is neither necessary nor important and why remove something useful that does not bother anyone.

PR #1605, opened 2020-09-30
Fixed issue #16506: Fruity theme: question type Matrix (texts), you do not see the sums of the total columns and grand total.

gabrieljenik:
Updating CSS clasess for the inputs as to have the data to be more visible

### DIFF
--- a/application/views/survey/questions/answer/arrays/texts/columns/col_total.twig
+++ b/application/views/survey/questions/answer/arrays/texts/columns/col_total.twig
@@ -12,7 +12,7 @@
         &nbsp;
     {% else %}
         <label class="sr-only">{{ gT("Total") }}</label>
-        <input title="[[ROW_NAME]] total" size="{{ inputsize }}" value="" type="text" disabled="disabled" class="disabled form-control"  data-number='1' />
+        <input title="[[ROW_NAME]] total" size="{{ inputsize }}" value="" type="text" disabled="disabled" class="disabled form-control total"  data-number='1' />
     {% endif %}
 </td>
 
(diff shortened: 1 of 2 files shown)


Shnoulle:
There already a css for total ? text-info or something lime this didn't work ?

gabrieljenik:
image
Yes, css class was there

Shnoulle:
:( i didn't like usage of class like this … i prefer to use `ls-total-sum` or a boostrap class.

we remove `.odd` `.even` usage :(

gabrieljenik:
ok, let me try with ls-total-sum

gabrieljenik:
ls-total-sum is not defined in the css.
Adding it may imply bigger changes for updating all the variations and also all the derived themes.

On the other hand, the total class definition is there and seems to fit what was designed for.

Alternatives: 

### 1 - Rename the total clase?
.total -> ls-sum-total

### 2 - Add exception rule
The text is currently very light and is not visible because of this defintion.
image

So, an alternative solution would be to add an exception to that definition in the case they are in the total row / column.
In that case, they would use the inherited.
image

But that again will require to update all variations.

Shnoulle:
> But that again will require to update all variations.

Yes : keep total is the quickest way , but i dislike ;)

eddylackmann:
@Shnoulle  at least we have to review our css classes in the future und redefine them in a proper way but it still a low priority for the moment.

eddylackmann (approved):

Shnoulle:
Yes , already said that. `total` seems unclear to me . I already update some css class in 3.X , somes other still there like this one 

We must just remind of this when adding new class name …

PR #1606, opened 2020-10-01
New feature #16710: add onUpdateResponse plugin event

TonisOrmisson:
Add the plugin event for any actual data submission/update of the response.

Will do the same pull request for develop branch of v4.

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -5597,6 +5597,15 @@
                             $cSave = new Save();
                             $cSave->set_answer_time();
                         }
+
+                        $event = new PluginEvent('onUpdateResponse');
+                        $event->set('surveyId', $this->sid);
+                        $event->set('responseId', $oResponse->id);
+                        $event->set('step', $thisstep);
+                        $event->set('responseAttributes', $aResponseAttributes);
+                        $event->set('finished', $finished);
+
+
                     } else {
                         LimeExpressionManager::addFrontendFlashMessage('error', $this->gT('This response was already submitted.'), $this->sid);
                     }


Shnoulle:
With 4.X : i think [afterResponseSave](https://manual.limesurvey.org/Dynamic_model_events) event happen (ut must)

2 event at same time ? Really a good idea ?

https://github.com/LimeSurvey/LimeSurvey/blob/987fa4bb04bb56b64648d322c425574445759d86/application/helpers/expressions/em_manager_helper.php#L5656

Shnoulle:
I con,firm : we already have the event in 4.X, i didn't add it in 3.X because it was a time where «No feature oin 3.X» …

Then PLEASE ! don't merge …

Shnoulle:
Related commits : https://github.com/LimeSurvey/LimeSurvey/pull/1353 and https://github.com/LimeSurvey/LimeSurvey/commit/a452cf022ff9a9bc747b2d1584bb3e59dd9b5bbc

To have the beforeReponseSave and afterResponseSave event in 3.X : 

Replace here
https://github.com/LimeSurvey/LimeSurvey/blob/6c4108ef0adb90739facc616f1e1c5fb4f18a1bd/application/helpers/expressions/em_manager_helper.php#L5589

by 
```
$oResponse->setAttributes($aResponseAttributes);
$oResponse->save();
```

Shnoulle (changes requested):
Better to fix 3.X to have beforeResponseSave and afterResponseSave event

TonisOrmisson:
@Shnoulle I am unable to find afterResponseSave & beforeResponseSave events in v4 nowhere. Of course there is no point of adding this if it is already existing somewhere, but I was unable to find it earlier as well as now. 

Sure, don't merge until we find out where afterResponseSave has disapperared? 
https://github.com/LimeSurvey/LimeSurvey/search?q=afterResponseSave

Deleteted at some point?

Shnoulle:
Plugin 

```
class checkResponseSave extends PluginBase {

  static protected $description = 'Did it work ?';
  static protected $name = 'checkResponseSave';


  public function init()
  {
    $this->subscribe('beforeResponseSave');
    $this->subscribe('afterResponseSave');
  }

  public function beforeResponseSave()
  {
    $oEvent = $this->getEvent();
    $this->log("survey id is : ".$oEvent->get('surveyId'));
    tracevar("survey id is : ".$oEvent->get('surveyId'));
    $response = $oEvent->get('model');
    $this->log("response id is : ".$response->id);
    tracevar("response id is : ".$response->id);
  }

  public function afterResponseSave()
  {
    $oEvent = $this->getEvent();
    $this->log("survey id is : ".$oEvent->get('surveyId'));
    tracevar("survey id is : ".$oEvent->get('surveyId'));
    $response = $oEvent->get('model');
    $this->log("response id is : ".$response->id);
    tracevar("response id is : ".$response->id);
  }
}
````

All models have this event, but i didn't create whole afterXXXXSave page

Shnoulle:
See https://manual.limesurvey.org/Specific_model_events too …

TonisOrmisson:
again, checkResponseSave is in docs, but not in actual source. `afterModelSave` seems too generic, if you need to do something with onUpdateResponse then a good bet is that you also need to see some context eg, what page/question was saved.

Shnoulle:
I create the clean dynamic model event …

Please : try the plugin : it's clear enough …

Shnoulle:
See pmy sample plugin : 

   $this->subscribe('**beforeResponseSave**');
    $this->subscribe('**afterResponseSave**');

PR #1608, opened 2020-10-01
Fixed issue #16642: PHP warning when showing array questions

gabrieljenik:
For dual scale questions, the 'cellwidth' variable calculation involved the number of labels. But, the number of labels is only calculated when the question is shown as radios, and not when it's shown as dropdown.

However, fixing the 'cellwidth' calculation (diving by 1 when 'numrows' is not set) affected the rendering because the "answer_dropdown" view used the 'cellwidth' value for styling and it now changed from "INF" (infinity) to an actual number. So, the use of 'cellwidth' in dropdown view was removed in order to not alter how the question was rendered.


Shnoulle:
What is the code from qanda in 3.X ?

gabrieljenik:
Sorry, didn't get that... I am slow today :)

eddylackmann (approved):

PR #1609, opened 2020-10-01
Fixed issue #16644: Audit log not save all actions

gabrieljenik:
There was a mismatch between the default values for "Audit log for this survey" setting. The default value shown on screen was Yes (true), and for most events it was indeed treated as 'true' by default when logging actions. But for tokens and survey settings, it defaulted to 'false' when logging actions.

It was fixed by defaulting to 'true' in all cases, as shown in the screen and expected by the user.

### DIFF
--- a/application/core/plugins/AuditLog/AuditLog.php
+++ b/application/core/plugins/AuditLog/AuditLog.php
@@ -337,7 +337,7 @@
         {
             $event = $this->getEvent();
             $iSurveyID=$event->get('iSurveyID');
-            if (!$this->checkSetting('AuditLog_Log_TokenSave') || !$this->get('auditing', 'Survey', $iSurveyID, false)) {
+            if (!$this->checkSetting('AuditLog_Log_TokenSave') || !$this->get('auditing', 'Survey', $iSurveyID, true)) {
                 return;
             }
 
@@ -615,7 +615,7 @@
             $event = $this->getEvent();
             $oModifiedSurvey = $event->get('modifiedSurvey');
             $iSurveyID = $oModifiedSurvey->sid;
-            if (!$this->checkSetting('AuditLog_Log_SurveySettings') || !$this->get('auditing', 'Survey', $iSurveyID, false)) {
+            if (!$this->checkSetting('AuditLog_Log_SurveySettings') || !$this->get('auditing', 'Survey', $iSurveyID, true)) {
                 return;
             }
 


Shnoulle:
You mean (in boolena) send NULL as current set as "checked" ?

https://github.com/LimeSurvey/LimeSurvey/blob/39e7bb96119ccf645ad9980c5706be8356c083a4/application/core/plugins/AuditLog/AuditLog.php#L598

Maybe the issue is more here, no ? `null == false == 0 == ''`

gabrieljenik:
You mean, why are they appearing as checked when they shouldn't... right?
I will review.

Shnoulle:
I'm not against set autiting at true by default, but i think (for a dev) `'current' => $this->get('auditing', 'Survey', $event->get('survey')) ` must set as false :).

In my plugins : i alway use a clean default value (for boolean or other) but : think  `$this->get('auditing', 'Survey', $event->get('survey')) ==  $this->get('auditing', 'Survey', $event->get('survey'),false)` 

For this issue : you can just replace by `$this->get('auditing', 'Survey', $event->get('survey'), true)` more clear.

For `$this->get('auditing', 'Survey', $event->get('survey')) !=  $this->get('auditing', 'Survey', $event->get('survey'),false)`  it's another issue ;)

gabrieljenik:
I think the problems comes as following:
- When looking at audit plugin settings, all checkboxes are checked, as default = 1
```            
'AuditLog_Log_UserLogin' => array(
                'type' => 'checkbox',
                'label' => 'Log if a user has logged in successfully',
                'default' => '1',
            ),
```

- Still, if user never saves that screen, the records for the settings are not saved.

- Then, for example on `beforeSurveySettingsSave`, the on/off is checked as follows.
```
            if (!$this->checkSetting('AuditLog_Log_SurveySettings') || !$this->get('auditing', 'Survey', $iSurveyID, false)) {
                return;
            }
```
While `checkSetting` seems to use the default value set on the plugin settings, `get` default is `false`, triggering the return, and not logging.

Shnoulle:
Seems really complex for … no reason …

gabrieljenik:
So, you think this:
```
            if (!$this->checkSetting('AuditLog_Log_SurveySettings') || !$this->get('auditing', 'Survey', $iSurveyID, false)) {
                return;
            }
```

should be replaced by something like this?
```
            if (!$this->get('auditing', 'Survey', $iSurveyID, false)) {
                return;
            }
```

or maybe something like this?
```
            if (!$this->checkSetting('AuditLog_Log_SurveySettings')) {
                return;
            }
```

I like the latest one as it pick ups the default values from the settings.
Even better, I would move the checkSetting method to the base plugin (in the future).

What do you think?

Shnoulle:
No : your fix seems OK, **i missunderstand the issue**, i think there are the checkbox at survey level.

1. If user check on Survey settings, but uncheck ALL Global settings : nothing is loggued : OK
2. If user check all at global but choose No at Survey settings : Nothing is loggued
3. If user check all at global and choose Yes at Survey settings : All is loggued
4. If user don't look at global settings : same than 2 and 3 : :+1: 

The only feature that can be intersting are : 

An new checkbox at global level "Activate log at surey level by default" checked by default

And at Survey level we have "Yes", "No" and "Leave default".

```
$auditing = $this->get('auditing', 'Survey', $iSurveyID, ""); 
if($auditing === "") {
$auditing = $this->get('auditing', null, null, 1); 
}
```

(I use [private function](https://gitlab.com/SondagesPro/coreAndTools/autoSaveAndQuit/-/blob/master/autoSaveAndQuit.php#L562-572) for this) **But it's a new feature**

gabrieljenik:
Good. I will leave it like this :)

gabrieljenik:
>  i think there are the checkbox at survey level.

You are right. I forgot for a while

eddylackmann (approved):

PR #1611, opened 2020-10-01
Fixed issue #16704: Inheritance not working when importing themes created for LS3.x

gabrieljenik:
The "theme options" page has multiple forms, and the right one was not being sent.


maziminke:
@cdorin93 Would be wonderful if we can include this fix at the next ls3 release.

eddylackmann (approved):

PR #1612, opened 2020-10-01
Fixed issue #16692: Clear process after test script is failing

gabrieljenik:
Cache was not properly cleared in between tests neither after survey deletion

The issue is caused by 'findByPkCache':
1) Test A imports survey 123456. The process involves using findByPK, and that methods saves the survey in a cache.
2) Test A finishes and the survey is deleted.
3) Test B imports survey 123456 again. This is a whole new survey. However, findByPk finds it in the cache, and returns the cached object.
4) Test B now has a copy of survey 123456 that is different from the last imported survey (groups and questions IDs differ).

Solved by clearing the findByPkCache in TestBaseClass::importSurvey() and clearing cache on Survey::delete()


gabrieljenik:
Made EditQuestionTest.php and ChangeQuestionTemplateTest.php to use the same survey (limesurvey_survey_573386) as to double check how the fix works in action.

Shnoulle:
stactic not cache , but great job :+1: , test is done in one step :)

gabrieljenik:
`stactic not cache`. You mean should be static, right?
I agree. Also, it should have its own methods for managing. Also, could be a separate clase... but... one step at a time... stability :)

olleharstedt:
Can you make a PR for master too? Just cherry pick the commits.

gabrieljenik:
> great job  , test is done in one step :)

Thanks!

Shnoulle:
> `stactic not cache`. You mean should be static, right?
> I agree. Also, it should have its own methods for managing. Also, could be a separate clase... but... one step at a time... stability :)

I mean : it's name `findByPkCache` but it use static var, not cache. Better name was `findByPkStatic`

I mean cache must be done for all calls.

And it explain why we have this issue in test and not in real. In general : after deletion: coma back to a new HTML page : \Survey::model()->findByPkCache is resetted :)

gabrieljenik:
Travis failed, not sure why..

olleharstedt:
> Travis failed, not sure why..

> PHP Parse error:  syntax error, unexpected '$this' (T_VARIABLE), expecting '(' in application/models/Survey.php on line 303

olleharstedt:
> And it explain why we have this issue in test and not in real. In general : after deletion: coma back to a new HTML page : \Survey::model()->findByPkCache is resetted :)

Yes, we've had lots of issues in testing, since it's the "same" request, in some sense.

Shnoulle:
> > And it explain why we have this issue in test and not in real. In general : after deletion: coma back to a new HTML page : \Survey::model()->findByPkCache is resetted :)
> 
> Yes, we've had lots of issues in testing, since it's the "same" request, in some sense.

reset findByPk can be a solution :). Maybe need to reset some other part (for example QuestionAttributes ?)

application/models/Survey.php
```diff
@@ -297,6 +297,12 @@ public function delete($recursive=true)
             // Delete all uploaded files.
             rmdirr(Yii::app()->getConfig('uploaddir').'/surveys/'.$this->sid);
         }
+
+        // Remove from cache
+        if (array_key_exists($this->sid, $this->findByPkCache)) {
+            unset $this->findByPkCache[$this->sid];
```
olleharstedt, line 303: Unset need `()`?

olleharstedt (changes requested):

application/models/Survey.php
gabrieljenik: My bad! Parenthesis missing. Sorry. Seems I overviewed that.

Unset is needed as to remove the survey from the cache if it is deleted.
Although cached will usually be refreshed as page is refreshed after delete, this code will help avoiding future problems, for example when testing.

I think is theorically needed and practically usefull in few situtations. But still, needed I think

olleharstedt (approved):

PR #1615, opened 2020-10-07
Normalize htmleditor import

gabrieljenik:
Sometimes it was being loaded as `admin.htmleditor` and some others as `admin/htmleditor`
That caused some collisions and error `function redeclare` as the file was included twice. as the alias check was failing.


Shnoulle:
Seems great :) (not tested)

eddylackmann (approved):

PR #1620, opened 2020-10-13
Fixed issue #16734 :update question edit mandatory default Y and same_default default 0

454335025:
Fixed issue #16734:


olleharstedt:
As Denis mentioned in the bug ticket, the correct solution is to make "Save as default" work good in LS4.

PR #1622, opened 2020-10-13
Set default folder for getFileList(); to "global"

eliasmagn:
Set default folder for getFileList();
Could be fixed via url generation as well...
https://bugs.limesurvey.org/view.php?id=16130
Fixed issue #16130

### DIFF
--- a/application/controllers/admin/LimeSurveyFileManager.php
+++ b/application/controllers/admin/LimeSurveyFileManager.php
@@ -138,6 +138,10 @@ class LimeSurveyFileManager extends Survey_Common_Action
      */
     public function getFileList($folder, $iSurveyId = null)
     {
+        if ($folder == '') {
+            $folder = 'global';
+        }
+
         // Get real folder name from alias.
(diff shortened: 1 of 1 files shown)


Shnoulle:
I like it, but i thhin there are an issue about send an empty folder : when it's done ?

On the issue «everytime i click on "filemanager"» filemanager msut be set according to current survey in general (with some exception : if not : it's an issue :)

eliasmagn:
Just click filemanager in configuration tab. no folder is submitted. Was talking about url/uri generation. I dont know which folder you want to show when opening the filemanager so i choosed a default one. Isnt it better to prevent faults coming from gui, i think about fuzzing. But yes, would be nicer to say something like: if folder not found send exception and set to default folder.
I dont kow if i understand you correctly ;-)

Shnoulle:
Have my answer : https://bugs.limesurvey.org/view.php?id=16130#c60185

Thanks :) 

Just  
```
        if (empty($folder)) {
            $folder = 'global';
        }
```
maybe

eddylackmann (approved):

PR #1624, opened 2020-10-14
Fixed issue #16738: Check logic, error in question group shown as error in question

Shnoulle:
Dev: reset errors and warning between each questions and after group
Dev: add error lines in group name

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -1613,6 +1613,18 @@ class ExpressionManager
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cdorin93:
Pfiu, good catch :) . Do you have time to provide the same fix to master as well @Shnoulle ?

Shnoulle:
Yes, already assigned to me :  https://bugs.limesurvey.org/view.php?id=16731

But it's @jmantysalo for the catch :)

I don't understand why we don't have this issue before ?????

PR #1460, opened 2020-07-01
Fixed issue #16442: Unable to use string with "on" in theme option label

Shnoulle:
Dev: User theme options must not be update totally like this
Dev: New settings already use optionlabels

PR #1626, opened 2020-10-16
Fixed issue #16442: Unable to use string with "on" in theme option label

Shnoulle:
Dev: User theme options must not be update totally like this
Dev: New settings already use optionlabels
Dev: just in case …


Shnoulle:
Copy of https://github.com/LimeSurvey/LimeSurvey/pull/1460 just in case.

Shnoulle:
@eddylackmann just to more clear of the update done, same fix.

eddylackmann (approved):

PR #1631, opened 2020-10-22
Fixed issue #16110: resources not found (vanilla based template)

gabrieljenik:
Removed spectrum colorpicker. Doesn't seem to be used anywhere.
It is included in template-core package definition (commit cf1ee28) and the js and css files are also registered in the Fruity's options.twig (commit bce3f7c).
But the 'spectrum' JS function is never used.


Shnoulle:
:+1:  great !

Shnoulle (approved):

Shnoulle:
We don't need spectrum currently : https://caniuse.com/input-color input type:color is best :)

maziminke:
I know we have some color picker for certain theme options. Is that a different tool that is used for that? I always assumed it was based on these spectrum files but I may be wrong.

Shnoulle:
@maziminke see https://github.com/LimeSurvey/LimeSurvey/pull/1631#issuecomment-714261694 we REALLY don't need javascript for color picker

PR #1633, opened 2020-10-23
Bug/16733  when importing a theme bigger in size than the allowed php.ini settings, there is no proper description of the error

gabrieljenik:
Fix applied to:
- Theme import (Themes list and Theme editor)
- File upload in Theme editor (uploading files to a theme)
- Image upload (in global theme options)
- Image upload (in survey theme options

There are other places where this should be applied. I have some already fixed. Want to review these first.

These are fixed and pendign to be applied:
- CPDB (Import CSV)
- Labelset import
- Token import
- Question import (had to add 'action' and 'sid' to the URL)

These are pending to be fixed:
- UploaderController
- questiongroups
- surveyadmin - Import survey - Need to split the method (used for copy and import), or move the 'action' to the URL.
- frontend_helper, checkUploadedFileValidity ?
- dataentry - Import VV file (need to split the method or assume an upload action when the request is POST)


gabrieljenik:
Notes to be reviewed in the future:

There are many redirects to admin/themes/sa/upload that don't seem right. They should probably redirect to admin/themeoptions.

When uploading a file to a theme (in theme editor), if the POST max size is exceeded, the page is redirected to the 'vanilla' theme. That's because the redirect goes to the theme editor without a target theme (the target theme is specified in the POST, which is empty).

Shnoulle:
I don't like to have it in common_helper 

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSFileHelper.php is miore for exsiting file on server.

Think usage (ane extend) https://www.yiiframework.com/doc/api/1.1/CFileValidator make more work , but maybe someting between ?

gabrieljenik:
> I don't like to have it in common_helper
> https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSFileHelper.php is miore for exsiting file on server.

Agree.

> Think usage (ane extend) https://www.yiiframework.com/doc/api/1.1/CFileValidator make more work , but maybe someting between ?

We can continue enhancing on the future? 
From this patch, we will have everything using the helpers, so enhancing means working on the helpers only.

Shnoulle:
Create a uploadHelper ?

Not in common_helper usedfor all anf nothing (my opinion)

gabrieljenik:
Creatied Uploadhelper and moved helper functions to class methods.

Once this push is OK, I have some other fixes already done for other screens as well.

application/core/LSUploadHelper.php
```diff
@@ -0,0 +1,99 @@
...
+            $iMaximumSize = min((int) $customMaxSize, getMaximumFileUploadSize());
+        }
+
+        // When 'post_max_size' is exceeded $_POST and $_FILES are empty.
+        // There is no way to confirm if the superglobals are empty because 'post_max_size' was
+        // exceeded, or because nothing was posted.
+        if (empty($_POST) && empty($_FILES)) {
+            throw new \Exception(
+                sprintf(
+                    gT("No file was uploaded or the request exceeded %01.2f MB."),
```
Shnoulle, line 39: :+1:

Shnoulle (approved):
Seems cool for me :)

PR #1634, opened 2020-10-28
Dev/16784 revert vuejs top bars

gabrieljenik:


gabrieljenik:
Travis is failing. Let's pick that up after the tollbar matters are totally addressed. What do you think?

gabrieljenik:
Continued here https://github.com/LimeSurvey/LimeSurvey/pull/1648

PR #1635, opened 2020-10-29
Fixed issue #14558: No difference shown for auto set Permission in survey

Shnoulle:
Dev: html have great feature … best is to use it


Shnoulle:
Capture d’écran du 2020-10-29 11-11-21

PR #1638, opened 2020-11-02
Fixed issue #16817: Missing MIME type check on survey import

bataineh:
Fixed issue #16817: fix bypassing-file-upload-restrictions in import survey
Dev:

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -1186,7 +1186,7 @@ class SurveyAdmin extends Survey_Common_Action
                 } elseif (!in_array(strtolower($sExtension), array('lss', 'txt', 'tsv', 'lsa'))) {
                     $aData['sErrorMessage'] = sprintf(gT("Import failed. You specified an invalid file type '%s'."), CHtml::encode($sExtension));
                     $aData['bFailed'] = true;
-                }elseif(!in_array($mimeType,['text/tab-separated-values','text/plain'])){
+                }elseif(!in_array($mimeType,['text/tab-separated-values','text/plain','application/xml','application/zip','text/xml'])){
                     $aData['sErrorMessage'] = sprintf(gT("Import failed. You specified an invalid file type '%s'."), CHtml::encode($mime));
                     $aData['bFailed'] = true; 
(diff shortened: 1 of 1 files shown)


Shnoulle:
We have a file helper, maybe we can use it ? https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSFileHelper.php


The type must be checked after error and extension (in my opinion) 

Did we really need `'application/octet-stream'` ? 

Can your report an issue on our mantis please ? https://github.com/LimeSurvey/LimeSurvey/blob/master/CONTRIBUTING.md#getting-started

Shnoulle:
You can report as our mantis ? 
https://bugs.limesurvey.org/

bataineh:
https://bugs.limesurvey.org/view.php?id=16817

olleharstedt:
Not really a security issue if it's impossible to rename files.

Shnoulle:
> Not really a security issue if it's impossible to rename files.

Maybe but you put an exe file in tmp directory and don't delete it, right : random name still there. But moved to LimeSurvey tmp

c-schmitz (approved):

application/controllers/admin/surveyadmin.php
```diff
@@ -1183,6 +1186,9 @@ public function copy()
                 } elseif (!in_array(strtolower($sExtension), array('lss', 'txt', 'tsv', 'lsa'))) {
                     $aData['sErrorMessage'] = sprintf(gT("Import failed. You specified an invalid file type '%s'."), CHtml::encode($sExtension));
                     $aData['bFailed'] = true;
+                }elseif(!in_array($mimeType,['text/tab-separated-values','text/plain'])){
```
Shnoulle, line 1189: Arg … unsure  on : 

- lss : `application/xml` or `text/xml`
- lsa : `application/zip`

?

c-schmitz: good question - not sure, either.

Shnoulle: Must test ;) with different server … :/

olleharstedt: Merged without resolving this?

PR #1639, opened 2020-11-02
Fixed issue: wrong port and TLS description in bounce email/smtp settings

ViliusS:
Dev: As per imap_open() manual {hostname/ssl} could be actually both SSL or implicit TLS connection depending on PHP configuration/version, which almost always uses standard IMAPS/POP3S ports. And {hostname/tls} means that a client requires a server to allow STARTTLS which usually runs on the same port as plain text IMAP/POP3. Make sure we won't set wrong port here if an administrator didn't specify one.

Dev: The same goes for SMTP connection and PHPMailer library. If "tls" is specified the SMTPSecure option in the library initiates STARTTLS support not the implicit TLS connection, see https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L2024

Dev: While at it, inform administrators that not using SSL/TLS or StartTLS makes a connection very insecure.

Dev: Also, make values lowercase as most of their default options in the code are lowercase.

The fix should be compatible with LS3.x branch.

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -99,13 +99,13 @@ class tokens extends Survey_Common_Action
                 $hostname       = getGlobalSetting('bounceaccounthost');
                 $username       = getGlobalSetting('bounceaccountuser');
                 $pass           = LSActiveRecord::decryptSingle(getGlobalSetting('bounceaccountpass'));
-                $hostencryption = strtoupper(getGlobalSetting('bounceencryption'));
+                $hostencryption = strtolower(getGlobalSetting('bounceencryption'));
             } else {
                 $accounttype    = strtoupper($thissurvey['bounceaccounttype']);
                 $hostname       = $thissurvey['bounceaccounthost'];
                 $username       = $thissurvey['bounceaccountuser'];
                 $pass           = LSActiveRecord::decryptSingle($thissurvey['bounceaccountpass']);
-                $hostencryption = strtoupper($thissurvey['bounceaccountencryption']);
+                $hostencryption = strtolower($thissurvey['bounceaccountencryption']);
             }
 
             @list($hostname, $port) = explode(':', $hostname);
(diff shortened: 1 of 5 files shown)


application/controllers/admin/tokens.php
```diff
@@ -153,13 +149,13 @@ public function bounceprocessing($iSurveyId)
 
             switch ($hostencryption) // novalidate-cert to have personal CA , maybe option.
```
c-schmitz, line 150: Because you switched to lowercase settings you must make sure that you convert the setting to lowercase first before comparing here.

application/views/admin/token/bounce.php
```diff
@@ -98,9 +98,9 @@
                                         'name' => 'bounceaccountencryption',
                                         'value'=> $settings['bounceaccountencryption'] ,
```
c-schmitz, line 99: Same here.

application/views/admin/globalsettings/_email.php
```diff
@@ -64,9 +64,9 @@
                     'name' => 'emailsmtpssl',
                     'value'=> getGlobalSetting('emailsmtpssl') ,
```
c-schmitz, line 65: Maybe here, too?

c-schmitz (changes requested):

application/controllers/admin/tokens.php
ViliusS: $hostencryption is converted on Line 102 and Line 108 above. The same lines which were used to convert to uppercase. I didn't change the logic.

application/views/admin/globalsettings/_email.php
ViliusS: emailsmtpssl option was already lowercase, I didn't change the case for it, just the bounceencryption.

application/views/admin/token/bounce.php
ViliusS: You are right here. Ideally I would have used database migration script for this, but it looked very complicated to me. Fix adjusted here and in global _bounce.php too.

ViliusS:
Everything should be fine now.

ViliusS:
Ping. Does this need any other adjustments?

c-schmitz (approved):

c-schmitz:
Thank you!

ViliusS:
Thank you too. Could this be merged to LTS-3.x? It should be straightforward.

maziminke:
@cdorin93 Please merge this into LS3 as well. it is a very useful enhancement.

@ViliusS Thanks for your contribution!

PR #1640, opened 2020-11-03
Dev: disable account type buttons if global configuration is used

ViliusS:
Dev: It is still possible to click on the buttons, even though the labels are disabled, but at least this shows a clear indication to the user.

### DIFF
--- a/assets/scripts/admin/tokenbounce.js
+++ b/assets/scripts/admin/tokenbounce.js
@@ -9,14 +9,14 @@ var LS = LS || {
 function updateParameters()
 {
 if ($('#bounceprocessing input:radio:checked').val()!='L'){
-        $("#bounceaccounttype label").attr('disabled','disabled');
+        $("#bounceaccounttype label").addClass('disabled');
         $("#bounceaccounthost").attr('disabled','disabled');
         $("#bounceaccountuser").attr('disabled','disabled');
         $("#bounceaccountpass").attr('disabled','disabled');
         $('#bounceaccountencryption label').addClass('disabled');
     }
     else {
-        $("#bounceaccounttype label").removeAttr('disabled');
+        $("#bounceaccounttype label").removeClass('disabled');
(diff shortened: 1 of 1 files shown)


Shnoulle:
disbaled is **not** supported by label
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

css issue, not js. Don't broke HTML please :)

ViliusS:
How about now?

Shnoulle:
Leave the attr disabled to the input maybe ;)

It work with class ? Maybe add class to all other lable (PS : **i didn't check** !)

ViliusS:
The reason I've submitted the patch is that disabling the input didn't work. These buttons are constructed via <label> and CSS classes.

Now it works and is on par with "bounceaccountencryption" type below.

Shnoulle:
> The reason I've submitted the patch is that disabling the input didn't work.

? Strange issue here … someone broke HTML before you ;)

Shnoulle:
OK, got it : seems in 3 : label have class, but no update on radio :)

What i have in current 4.3.23  seems clear enough in global settings
Peek 03-11-2020 16-21

Here : it's for survey settings ?

Shnoulle:
Near same aspect in Survey settings : in my opinion : it's clear enough

@cdorin93 ?

Shnoulle:
In survey settings : seems button have already the disabled class … (but the text input didn't have good class)
Peek 03-11-2020 18-45

ViliusS:
You are checking the wrong field. The patch is against bounceaddresstype, not the bounceaddressencryption.

Shnoulle:
> The fix is for survey settings, not the global settings.

I don't know we can disable it … where can i get the page please ?

Shnoulle:
OK !!!

 Server type  … finally i got it …

Shnoulle:
Thank you and sorry for my misunderstanding ...

PR #1646, opened 2020-11-05
Fixed issue #16819: Delete a single user in user management and throw an exception.

gabrieljenik:


Shnoulle:
@gabrieljenik the input already exist , no ?

gabrieljenik:
It is added by JS.

Shnoulle:
We talk about remove this part https://github.com/LimeSurvey/LimeSurvey/blob/8263c8c55a521570bf086eb7c808f4c47fd2a215/application/extensions/LimeScript/LimeScript.php#L30  with @olleharstedt 

Then : if only added by this line, maybe need to fix it differently :)


`TbHtml::formTb` didn't add the CRSF token it self ?

olleharstedt:
If it's JSON encoded it will probably be surrounded by an extra pair of quotation marks.

DELETE should always be a POST, so it needs CSRF.

Shnoulle:
What i mean exactly : 

Use CHtml::form add the crsf token (name and value) : the `TbHtml::formTb` must add it too, no ? This is my original question

Else  : 

1. I always prefer to submit an existing form (or better : use an existing button) in javascript : always more easy to understand 
2. This input are totally broken : it don't use `Yii::app()->request->csrfTokenName` ;)

gabrieljenik:
My bad in diagnosis. Still the patch is OK.

The extra CSRF hidden input  field is added, as you say @Shnoulle, with the `TbHtml::formTb`.
It was added by it and then explicitily added later.

Sorry about the confusion

Shnoulle:
Thanks for your confirmation and your patch, you don't have to be sorry …

Trischi80:
May be this link could be interesting for discusion about CSRF:
https://www.yiiframework.com/wiki/274/how-to-validate-csrf-token-with-session

PR #1647, opened 2020-11-06
DEV #16785: Revert VueJs Topbars II

gabrieljenik:
Integrated the new top bar system into the new controllers.
Needed a way for sharing topBars among views and also a general framework where to place common topBar parts (ease maintenance).

Key parts:
LayoutHelper:renderTopbar() called from layouts.
Base top bar with Left and Right side configurable.
Totally overridable if needed.
Implemented it for email templates.

WIP
Next step survey toolbar, which is shared among many screens.


gabrieljenik:
Wrong branch name.

PR #1649, opened 2020-11-08
Fixed issue: Allow to login if LDAP user's DN contains double quotes

ViliusS:
Dev: Patch fixes an issue where LDAP users with double quotes or other special characters in their DN could not login if a group filter is set.

Should be applied to LS3.x branch too.


c-schmitz:
Thank you, again!

PR #1654, opened 2020-11-13
Fixed issue #16042  generated token length is too short with add participants()

gabrieljenik:
Updated code from generateTokens()
Code reorg

### DIFF
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -61,6 +61,9 @@ abstract class Token extends Dynamic
     /** @var int Maximum token length */
     const MAX_LENGTH = 36;
 
+    /** @var int Default token length */
+    const DEFAULT_LENGTH = 15;
+
     /**
      * Set defaults
      * @inheritdoc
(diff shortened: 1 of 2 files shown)


application/models/Token.php
```diff
@@ -61,6 +61,9 @@ abstract class Token extends Dynamic
     /** @var int Maximum token length */
     const MAX_LENGTH = 36;
 
+    /** @var int Default token length */
+    const DEFAULT_LENGTH = 15;
```
Shnoulle, line 65: We don't have a config for this ? I think yes, but no ....

application/models/Token.php
```diff
@@ -231,13 +234,34 @@ public function findByToken($token)
...
+    }
+
     /**
      * Generates a token for this object.
      * @throws CHttpException
      */
     public function generateToken($tokenlength = NULL)
     {
-        $iTokenLength = $tokenlength ? $tokenlength : $this->survey->tokenlength;
+        if ($tokenlength) {
```
Shnoulle, line 260: Why not `generateToken($iTokenLength = NULL)`

And 
```
if(empty($iTokenLength)) {
    $iTokenLength = $this->getSurveyTokenLength();
}
```

I don't understand why use 2 different parameter name here ....

gabrieljenik: You are right.

application/models/Token.php
gabrieljenik: Would be the best, but we don't have it and I just wanted to wrap the ticket :)
The less changes, the less tests, the less destabilization

Shnoulle: totally right :+1:

Shnoulle (approved):

eddylackmann (approved):

PR #1662, opened 2020-11-20
Fixed issue #16847: Sort users when adding to usergroup

jmantysalo:
Fixed issue #<Mantis issue number>: 16847


Shnoulle:
order by uid ? Whny not by name ?

eddylackmann (approved):

PR #1663, opened 2020-11-20
New feature #16832: Add user group functions to LimeSurveyAPI

flowdnb:
New feature #16832: Add user group functions to LimeSurveyAPI

Dev: Allow plugins to read and create user group, add/remove user from group

### DIFF
--- a/application/libraries/PluginManager/LimesurveyApi.php
+++ b/application/libraries/PluginManager/LimesurveyApi.php
@@ -473,4 +473,90 @@ class LimesurveyApi
             false
         );
     }
+
+    /**
+     * Returns an array of all user groups
+     *
+     * @return array
+     */
+    public function getUserGroups()
+    {   
+        return \UserGroup::model()->findAllAsArray();
+    }
+
+    /**
+     * Adds a new user group
+     *
+     * @param string $groupName Name of user group to be created
+     * @param string $groupDescription Description of user group to be created
+     * @return boolean True or false if user group was added or not
+     */
+    public function addUserGroup($groupName, $groupDescription)
+    {
+        $db_group_name = flattenText($groupName, false, true, 'UTF-8', true);
+        $db_group_description = $groupDescription;
+
+        if (isset($db_group_name) && strlen($db_group_name) > 0) {
+            if (strlen($db_group_name) > 21) {
+                return false;
+            } elseif (\UserGroup::model()->find("name=:groupName", array(':groupName'=>$db_group_name))) {
+                return false;
+            } else {
+                $newUserGroup = new \UserGroup();
+                $newUserGroup->owner_id = 1;
+                $newUserGroup->name = $groupName;
+                $newUserGroup->description = $groupDescription;
+                $result = $newUserGroup->save();
+                if ($result) {
+                    \UserInGroup::model()->insertRecords(array('ugid' => $newUserGroup->getPrimaryKey(), 'uid' => 1));
+                }
+                return $result;
+            }
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Adds or removes a user from a user group
+     *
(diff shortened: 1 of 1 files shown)


application/libraries/PluginManager/LimesurveyApi.php
```diff
@@ -473,4 +473,90 @@ public function renderTwig($file, array $data)
...
+
+    /**
+     * Adds or removes a user from a user group
+     *
+     * @param integer $ugid The user group ID
+     * @param integer $uid The user ID
+     * @param string $action The desired action 'add' or 'remove'
+     * @return boolean True or false if user was added/removed to/from group or not
+     */
+    public function userInGroupAction($ugid, $uid, $action)
```
olleharstedt, line 528: Please remove `$action` and split the function in two parts, one for remove, one for add. If the functions have code in common, split it into a third private function.

olleharstedt (changes requested):

Shnoulle:
Not about this function but a general remark : if we add function again on API we have a really great file finally  … maybe find another way to set "this part was in API" ?

application/libraries/PluginManager/LimesurveyApi.php
```diff
@@ -473,4 +473,90 @@ public function renderTwig($file, array $data)
...
+    {
+        $group = \UserGroup::model()->findByAttributes(array('ugid' => $ugid));
+
+        if (empty($group)) {
+            return false;
```
Shnoulle, line 533: Throw Exception ?

application/libraries/PluginManager/LimesurveyApi.php
```diff
@@ -473,4 +473,90 @@ public function renderTwig($file, array $data)
...
+        } else {
+            if ($uid > 0 && \User::model()->findByPk($uid)) {
```
Shnoulle, line 535: Same : throw Exception , no ?

application/libraries/PluginManager/LimesurveyApi.php
```diff
@@ -473,4 +473,90 @@ public function renderTwig($file, array $data)
...
+     * @param string $groupDescription Description of user group to be created
+     * @return boolean True or false if user group was added or not
+     */
+    public function addUserGroup($groupName, $groupDescription)
+    {
+        $db_group_name = flattenText($groupName, false, true, 'UTF-8', true);
+        $db_group_description = $groupDescription;
+
+        if (isset($db_group_name) && strlen($db_group_name) > 0) {
+            if (strlen($db_group_name) > 21) {
```
Shnoulle, line 500: No … must try to save or validate and return model->errors…

application/libraries/PluginManager/LimesurveyApi.php
```diff
@@ -473,4 +473,90 @@ public function renderTwig($file, array $data)
...
+                return false;
+            } elseif (\UserGroup::model()->find("name=:groupName", array(':groupName'=>$db_group_name))) {
```
Shnoulle, line 502: Same : try to validate : return model->errors

olleharstedt:
Merged, but please consider also returning model errors if save fails.

maziminke:
@olleharstedt can we also backport this to LS 3.x API? That version is the LTS version which most customers use so it makes sense making it available there as well.
It is just a new API function, that should not cause any harm...

olleharstedt:
I can try to cherry-pick this into LS3 today. Shouldn't be a problem, as you say.

olleharstedt:
Welp, forgot. Please remind me next week.

olleharstedt:
Cherry-picked to LS3.

Reasons we approved internally:

* Zero or very low risk
* LS4 has been unstable long time
* Request by long-time partner

PR #1664, opened 2020-11-24
Fixed issue #16308: Multi Numeric Slider custom Font Awesome icons

gabrieljenik:
There were several problems:
- numeric-slider.js set the CSS for the custom slider handle, but it used a wrong selector.
- numeric-slider.js set the content to the unicode value directly, without using backslash.
- In the case of Fruity, it was setting the slider handle's background color, which works for 'round' and 'square' but not for 'triangle' and 'custom'. Fixed by adding this rule to all variations: .fruity .slider-handle.custom,.fruity .slider-handle.triangle{background-color:transparent}


application/views/survey/questions/answer/multiplenumeric/rows/sliders/answer_row.twig
```diff
@@ -39,7 +39,7 @@
                 <div class='col-xs-12 col-sm-2 slider-left text-right'>{{ sliderleft }}</div>
             {% endif %}
 
-            <div class="slider-container ls-input-group col-xs-12 col-sm-{{sliderWidth}}" style="min-height:15px;">
+            <div class="slider-container ls-input-group col-xs-12 col-sm-{{sliderWidth}}" style="min-height:15px;" id="answer{{myfname}}slid-container">
```
Shnoulle, line 42: `min-height:15px` , maybe we can include this in a css file ? It don't work without ?

PR #1666, opened 2020-11-24
Fixed issue #16853: Inconsistent filter behaviour when create token

Shnoulle:
Dev: remove flattenText in controller
Dev: usage of filter_var on attribute for all user (admin or not)

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -436,7 +436,7 @@ class tokens extends Survey_Common_Action
             $aCoreTokenFields = array('firstname', 'lastname', 'emailstatus', 'token', 'language', 'sent', 'remindersent', 'remindercount', 'completed', 'usesleft');
             foreach ($aCoreTokenFields as $sCoreTokenField) {
                 if (trim(Yii::app()->request->getPost($sCoreTokenField, 'lskeep')) != 'lskeep') {
-                    $value = flattenText(Yii::app()->request->getPost($sCoreTokenField));
+                    $value = App()->request->getPost($sCoreTokenField);
                     if ($sCoreTokenField == 'language' && empty($value)) {
                         continue;
                     }
@@ -451,7 +451,7 @@ class tokens extends Survey_Common_Action
             $attrfieldnames = GetParticipantAttributes($iSurveyId);
             foreach ($attrfieldnames as $attr_name => $desc) {
                 if (trim(Yii::app()->request->getPost($attr_name, 'lskeep')) != 'lskeep') {
-                    $value = flattenText(Yii::app()->request->getPost($attr_name));
+                    $value = App()->request->getPost($attr_name);
                     if ($desc['mandatory'] == 'Y' && trim($value) == '') {
                         $aResults['global']['result'] = false;
                         $aResults['global']['message'] = sprintf(gT('%s cannot be left empty'), $desc['description']);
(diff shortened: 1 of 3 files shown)


application/models/Token.php
```diff
@@ -366,7 +383,11 @@ public function rules()
             array('emailstatus', 'default', 'value' => 'OK'),
         );
         foreach (decodeTokenAttributes($this->survey->attributedescriptions) as $key => $info) {
-                $aRules[] = array($key, 'LSYii_Validators', 'except'=>'FinalSubmit');
+            $aRules[] = array(
+                $key, 'filter',
+                'filter' => array(self::class, 'sanitizeAttribute'),
+                'except'=>'FinalSubmit'
```
Shnoulle, line 389: I muts add `'except'=>'FinalSubmit'` for firstname or lastname too ? It's a sanitizer, not a checker.

application/models/Token.php
```diff
@@ -263,6 +263,15 @@ public static function sanitizeToken($token)
...
     }
 
+    /**
+     * Sanitize string for any attribute
+     * @param string $attribute to sanitize
+     * @return string sanitized attribute
+     */
+    public static function sanitizeAttribute($attribute)
+    {
+        return filter_var($attribute, FILTER_SANITIZE_STRING);
```
Shnoulle, line 273: With a checker (is valid) `return filter_var($attribute, FILTER_SANITIZE_STRING) == $attribute` maybe bettre : don't update, just send a error and disable save.
But : unsure admin GUI was ready, same with import …

Shnoulle:
See https://bugs.limesurvey.org/view.php?id=16884 :)

application/controllers/admin/tokens.php
```diff
@@ -541,17 +541,17 @@ public function editToken($iSurveyId)
 
             echo $from . ',' . $until;
             $aData = array(
-                'firstname' => flattenText(Yii::app()->request->getPost('firstname')),
```
olleharstedt, line 544: Why is flattenText removed?

Shnoulle: Because it was ONLY for admin … via GUI : not for

- register
- remote control

for example.

Question : did you read the mantis issue ?

Shnoulle: With this one we flatten for all action (using model …) `return filter_var($attribute, FILTER_SANITIZE_STRING);` not **only** for admin … user via GUI

Shnoulle: See related : https://bugs.limesurvey.org/view.php?id=16884 …

Shnoulle: CPDB allow tag too

olleharstedt: OK :)

Shnoulle: Can resolve ?

application/models/Token.php
Shnoulle: Check : admin GUI not ready …

olleharstedt:
Did anyone test this?

Shnoulle:
Me …

olleharstedt:
The dev should not test his/her own code. :)

olleharstedt:
At least not only the dev. :d

Shnoulle:
;)

Here : see the original issue one the [«way to reproduce»](https://bugs.limesurvey.org/view.php?id=16853). But need 3 seperate test (minimum 3)

But this fix fix https://bugs.limesurvey.org/view.php?id=16884 too …

c-schmitz:
Denix, can you port this also to master and develop, please?

Shnoulle:
Done on both

olleharstedt:
Seems like this fix is incorrectly escaping `'` (apostrophe) in participant names.

olleharstedt:
Possible fix here: https://github.com/LimeSurvey/LimeSurvey/commit/b4953e4a4c15db37fca875c3b21c59a6eaf31156

Shnoulle:
Thank you

PR #1673, opened 2020-12-03
Fixed issue #16749: Make sure Vanilla and Bootswatch look like Fruity

gabrieljenik:
BootsWatch already works like Fruity.
Added options page entry to Vanilla's config to enable the core options page. Tested with basic settings and seems to work OK.
Also, themes that extended Vanilla before and after the change work too.


Shnoulle:
I need to move to core ;) maybe …

eddylackmann (approved):

PR #1674, opened 2020-12-03
Fixed issue #16670: Extended theme shows all options in one page

gabrieljenik:
When the theme config is copied, the engine node is removed (apparently it caused issues before).
Since the optionspage config is inside the engine node, it's removed from the extended theme config, and then the options.twig is used.
One option is to move the optionspage out of the engine node, so it's not removed.
Other option is to remove the engine node contents EXCEPT optionspage, instead of the whole engine node. This option seems easier and it was the one implemented.

### DIFF
--- a/application/models/TemplateManifest.php
+++ b/application/models/TemplateManifest.php
@@ -892,7 +892,7 @@ class TemplateManifest extends TemplateConfiguration
     }
 
     /**
-     * Delete files and engine node inside the DOM
+     * Delete engine node inside the DOM, except the optionspage configuration
      *
      * @param DOMDocument   $oNewManifest  The DOMDOcument of the manifest
      */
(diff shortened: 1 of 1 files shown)


Shnoulle:
>  One option is to move the optionspage out of the engine node, so it's not removed.

My opinon

I think options part are not part of engine

PS : i think too https://github.com/LimeSurvey/LimeSurvey/blob/c1e46c896af57055d3e0d87948f91b7437b522bb/themes/survey/vanilla/config.xml#L115 are not part of «engine»

optionspage in metadata ? Unsure ?

Shnoulle:
PS : more easiest solution : don't remove engine ;)

eddylackmann (approved):

PR #1679, opened 2020-12-03
Fixed issue #16854: Cannot attach PDF to email invitation (email template for surveys)

gabrieljenik:
Previously, at emailtemplates controller's 'update' method, the settings were saved directly using updateAll. After commit '2c8ccdd5', in order to apply the model's validation, the settings are saved using 'setAttributes($attributes)' and 'save()'.
The setAttributes method, by default, only assigns 'safe' attributes (attributes that have a validation rule). That excludes the 'attachments' attribute.
Passing 'false' as second parameter solves the issue.

### DIFF
--- a/application/models/SurveyLanguageSetting.php
+++ b/application/models/SurveyLanguageSetting.php
@@ -135,6 +135,8 @@ class SurveyLanguageSetting extends LSActiveRecord
 
             array('surveyls_dateformat', 'numerical', 'integerOnly'=>true, 'min'=>'1', 'max'=>'12', 'allowEmpty'=>true),
             array('surveyls_numberformat', 'numerical', 'integerOnly'=>true, 'min'=>'0', 'max'=>'1', 'allowEmpty'=>true),
+
+            array('attachments', 'attachmentsInfo'),
         );
     }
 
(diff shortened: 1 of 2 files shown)


Shnoulle:
Maybe we can set attachments to safe ? Or : check if it's a valid json before save ?
https://www.yiiframework.com/doc/api/1.1/CModel#setAttributes-detail

It just need a valid rules to be safe

gabrieljenik:
I like the approach. Let me review.

Shnoulle:
PS : comment on discord ;)

cdorin93:
Everything is fine with the PR. Will there be any additional change pushed @gabrieljenik ?

gabrieljenik:
> Everything is fine with the PR. Will there be any additional change pushed @gabrieljenik ?

I have uploaded now an alternative approach according to latest comments.
Should be reviewed again. Sorry

Shnoulle:
You filter current attachment in validation here ?

Gloups … it's a serialized not a json …

gabrieljenik:
> You filter current attachment in validation here ?
> 
> Gloups … it's a serialized not a json …

Don't understand sorry.

Shnoulle:
> > You filter current attachment in validation here ?
> > Gloups … it's a serialized not a json …
> 
> Don't understand sorry.

For filter VS validation : 
- filter update the attribute just before save
- validation check if value is correct and disable save if not.

In LimeSurvey : we use a lot filter :) . Sometimes filter is great, sometimes not :)

About serialized : i just don't know how value is saved in DB. I really think itr's json.

application/models/SurveyLanguageSetting.php
```diff
@@ -183,6 +185,30 @@ public function lsdefault($attribute)
...
     }
 
+    /**
+     * Defines the customs validation rule attachmentsInfo
+     *
+     * @param mixed $attribute
+     */
+    public function attachmentsInfo($attribute)
+    {
+        if (empty($this->$attribute)) return;
```
Shnoulle, line 195: PSR don't ask

```
if (test) { 
return;
}
```

Shnoulle (commented):
I apply the fix on a windows IIS system … seems it don't work too … Can not activate debug mode currently … i check more

Shnoulle:
OK , more issues for windows , but unrelated to this commit :+1:

Shnoulle (approved):

gabrieljenik:
> OK , more issues for windows , but unrelated to this commit

Thanks for reviewing!!! Appreciate it!

Shnoulle:
See the solution without saving path : https://github.com/LimeSurvey/LimeSurvey/pull/1800

This fix issue with publicurl and windows server

Shnoulle:
Review a second time : https://github.com/LimeSurvey/LimeSurvey/pull/1801



:)

PR #1682, opened 2020-12-03
Fixed issue #16799: User with read + create survey : invalid list shown

gabrieljenik:
LIsting survey which have at least read permissions

### DIFF
--- a/application/controllers/admin/surveyadmin.php
+++ b/application/controllers/admin/surveyadmin.php
@@ -1266,9 +1266,9 @@ class SurveyAdmin extends Survey_Common_Action
     public function copy()
     {
         if (!Permission::model()->hasGlobalPermission('surveys', 'create')) {
-            Yii::app()->user->setFlash('error', gT("Access denied"));
-            $this->getController()->redirect(Yii::app()->request->urlReferrer);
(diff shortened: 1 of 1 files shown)


application/controllers/admin/surveyadmin.php
```diff
@@ -1265,6 +1265,10 @@ public function editlocalsettings($iSurveyID)
      */
     public function copy()
     {
+        if (!Permission::model()->hasGlobalPermission('surveys', 'create')) {
```
Shnoulle, line 1268: 😢 403 error …

gabrieljenik: What do you mean?

Shnoulle: Send a 403 error : https://bugs.limesurvey.org/view.php?id=16470

Some reason why it's better are in the comments , some issue take really more time without a clean error : https://bugs.limesurvey.org/view.php?id=16470#c60490

gabrieljenik: Done

Shnoulle:
Thanks you !!! Realy thank you ! (because it's not mandatory ;)

Shnoulle (approved):
:+1: :heart:

PR #1683, opened 2020-12-10
Fixed issue #16895: Question radio button with Question Theme "Bootstrap Buttons" does not keep value in "Other" field

gabrieljenik:
There were two problems:
1) The 'other' answer value from the DB was never used by the theme to feed it into the screen ($answer_other was not used in answer_row_other.twig).
2) The show/hide code is part of the radio's 'change' event, which was only trigger on an actual change, not initially to setup the screen

An additional problem is that, in this theme, the 'other' input control is outside the scope of answer_row_other.twig, so the value cannot be directly assigned when rendering. Instead, a hidden input had to be added to hold the value until it can be transfered by javascript to the actual control.

At last, as the doBootstrapRadio is called multiple times (once per question in the screen), the event assignment happens multiple times as well.

### DIFF
--- a/themes/question/bootstrap_buttons/survey/questions/answer/listradio/assets/scripts/bootstrapbuttons.js
+++ b/themes/question/bootstrap_buttons/survey/questions/answer/listradio/assets/scripts/bootstrapbuttons.js
@@ -1,12 +1,25 @@
 function doBootstrapRadio(){
-    $(".button-item .bootstrap-radio").change(function(){
-        name = $(this).attr('name');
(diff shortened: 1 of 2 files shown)


Shnoulle:
I don't know … maybe it's better to call doBootstrapRadio with param ?
doBootstrapRadio(qid) ?

gabrieljenik:
I don't know. I Thought about it, but this way seems a bit more performant and involved less changes to the structure of the code.
I am comfortable with this approach. Could be a better one, but feel this is accpetable.

PR #1684, opened 2020-12-10
Fixed issue #16833: new config setting for date format and question code updates

gabrieljenik:
For date questions, the length of the placeholders is set accoding to the format of the question
Question Title Column width can be set from config-defaults
For questions with subquestion, the question code is shown withouth the subquestion part. [Bug]

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -424,6 +424,9 @@ $config['quexmlusequestiontitleasid'] = false;
 // QueXML: If set to true, the Data/Time answers will be formated with the survey's date format
 $config['quexmlkeepsurveydateformat'] = false;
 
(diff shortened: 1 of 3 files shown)


maziminke:
@cdorin93 I think this can be merged?

application/config/config-defaults.php
```diff
@@ -424,6 +424,9 @@
 // QueXML: If set to true, the Data/Time answers will be formated with the survey's date format
 $config['quexmlkeepsurveydateformat'] = false;
 
+// QueXML: Width of the question title column in MM
+$config['quexmlquestiontitlewidth'] = 14;
```
olleharstedt, line 428: I can't find where this setting is used?

olleharstedt (changes requested):

PR #1686, opened 2020-12-11
Fixed issue #16896: Javascript shown when editing subquestions

gabrieljenik:
The HTML breaks when the subquestion's text contains double quotes (").
Added "escape('html_attr')" to subquestion text and code (subquestionRow.twig).

It seems the same happens with other fields (like the "Array filter" attribute, just to mention one).
To be checked on #16904

### DIFF
--- a/application/views/questionAdministration/subquestionRow.twig
+++ b/application/views/questionAdministration/subquestionRow.twig
@@ -25,7 +25,7 @@
 
         <!-- Code (title) -->
         <td class="code-title" style="vertical-align: middle;">
-            {{ subquestion.title }}
+            {{ subquestion.title|escape('html_attr') }}
         </td>
 
     {# If survey is not activated and first language : move button, code editable   #}
@@ -54,7 +54,7 @@
                     maxlength="20"
                     id='subquestions[{{ subquestion.qid }}][{{ scale_id }}][oldcode]'
                     name='subquestions[{{ subquestion.qid }}][{{ scale_id }}][oldcode]'
-                    value="{{ subquestion.title }}"
+                    value="{{ subquestion.title|escape('html_attr') }}"
                 />
             {% endif %}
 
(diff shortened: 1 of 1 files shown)


Shnoulle:
Why not `|escape` directly ? Why `escape('html_attr')` ?

gabrieljenik:
> Why not `|escape` directly ? Why `escape('html_attr')` ?

As per documentation:
html: escapes a string for the HTML body context.
html_attr: escapes a string for the HTML attribute context.

The latest seem to apply more to the context :)

application/views/questionAdministration/subquestionRow.twig
```diff
@@ -25,7 +25,7 @@
 
         <!-- Code (title) -->
         <td class="code-title" style="vertical-align: middle;">
-            {{ subquestion.title }}
+            {{ subquestion.title|escape('html_attr') }}
```
Shnoulle, line 28: It's html here : `{{ subquestion.title|escape('html') }}` ;)

application/views/questionAdministration/subquestionRow.twig
```diff
@@ -80,7 +80,7 @@
 
         <!-- Code (title) -->
         <td  class="code-title" style="vertical-align: middle;">
-            {{ subquestion.title }}
+            {{ subquestion.title|escape('html_attr') }}
```
Shnoulle, line 83: It's html here : `{{ subquestion.title|escape('html') }}` ;)

Shnoulle (changes requested):
`{{ subquestion.title|escape('html') }}` when it's html (thanks for the [explanation](https://twig.symfony.com/doc/2.x/filters/escape.html))

Shnoulle:
:D

application/views/questionAdministration/subquestionRow.twig
gabrieljenik: done. Thanks

application/views/questionAdministration/subquestionRow.twig
gabrieljenik: done. Thanks

Shnoulle (approved):

olleharstedt:
Thanks for fixing my :hankey: :laughing: :heart:

PR #1687, opened 2020-12-11
Fixed issue #16897: Subquestions are being automatically picked up from previous (sub)questions

gabrieljenik:
Subquestions were listed using Question::getScaledSubquestions(), which queries the question's table by parent_qid = $this->qid.
Since 'qid' is empty until the question was saved, the query returned the data of all the questions in the DB.


olleharstedt:
Good lord. Sorry for that, thanks for the fix!

PR #1689, opened 2020-12-11
Fixed issue #16898: Pop Up Editor does not work inside the Question Editor

gabrieljenik:
There were issues with the triggering of the popup editor, both for answers and for subquestions.
These were mainly related to the identification of the target field, and also the values of SID, GID and QID passed to the editor.
Some problems originated in the twigs, but the updateRowProperties() method (questionEditor.js) was also causing problems.

### DIFF
--- a/assets/scripts/admin/questionEditor.js
+++ b/assets/scripts/admin/questionEditor.js
@@ -91,9 +91,12 @@ $(document).on('ready pjax:scriptcomplete', function () {
    * @return {void}
    */
   function updateRowProperties() {
-    const sID = $('input[name=sid]').val();
-    const gID = $('input[name=gid]').val();
-    const qID = $('input[name=qid]').val();
+    var sID = $('input[name=sid]').val();
+    var gID = $('input[name=gid]').val();
+    var qID = $('input[name=qid]').val();
+    sID = $.isNumeric(sID) ? sID : '';
+    gID = $.isNumeric(gID) ? gID : '';
+    qID = $.isNumeric(qID) ? qID : '';
 
     /**
      * This function adjusts the alternating table rows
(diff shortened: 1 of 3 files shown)


assets/scripts/admin/questionEditor.js
```diff
@@ -91,9 +91,12 @@ $(document).on('ready pjax:scriptcomplete', function () {
...
   function updateRowProperties() {
-    const sID = $('input[name=sid]').val();
-    const gID = $('input[name=gid]').val();
-    const qID = $('input[name=qid]').val();
+    var sID = $('input[name=sid]').val();
+    var gID = $('input[name=gid]').val();
+    var qID = $('input[name=qid]').val();
+    sID = $.isNumeric(sID) ? sID : '';
+    gID = $.isNumeric(gID) ? gID : '';
+    qID = $.isNumeric(qID) ? qID : '';
```
olleharstedt, line 99: Wait, why define these variables globally?

olleharstedt (changes requested):

olleharstedt:
Thanks! Feel free to use `let` instead of `var` for future JS in admin, since it has better scope (block instead of function scope).

gabrieljenik:
Yes, but be carefull about browser compatiblity when using those "strange" syntaxes

olleharstedt:
> Yes, but be carefull about browser compatiblity when using those "strange" syntaxes

We agreed to require ES8 for admin support. All updated browsers support this already. Makes it easier to structure your code.

olleharstedt:
E.g. we can use classes, short-hand function notation, async/await.

olleharstedt:
I even type-checked the question editor JS with Flow _and_ TypeScript :laughing: But, that's another discussion.

PR #1694, opened 2020-12-22
Dev: Move permission on SurveysGroups to SurveysGroupsPermissionController

Shnoulle:


olleharstedt:
Looks good, Denis, will merge when Travis is green. Many thanks!

Shnoulle:
Hope i check all action ;)


I really like to use same view https://github.com/LimeSurvey/LimeSurvey/blob/6325e92cf4b58b6cc1e1a0d16a89dae581fa365f/application/views/SurveysGroupsPermission/subviews/setPermissionForm.php#L16-L59 part for Survey.

But : create a widget seems out of the need , no ?

Start by https://github.com/LimeSurvey/LimeSurvey/tree/6325e92cf4b58b6cc1e1a0d16a89dae581fa365f/application/views/admin/super/permissions is OK ?

olleharstedt:
A widget is nothing more than a class with a render() function and a folder with at least one view file. But... that super/ folder contains a lot of stuff! Both layouts and other random things.

Shnoulle:
> A widget is nothing more than a class with a render() function and a folder with at least one view file. But... that super/ folder contains a lot of stuff! Both layouts and other random things.

Yes, but here : i just need same view. Right : i can create widget but in a scond step maybe …

olleharstedt:
We need more discussions next year on how to organize the views, I think...

Shnoulle:
Yes, i include subviews here … 

If using twig : extend is better

PR #1696, opened 2020-12-23
Fixed issue #15611: No HTML wrapper in public stats

gabrieljenik:
Created new twig views for statistics for using as frame (based on print view)

### DIFF
--- a/application/controllers/Statistics_userController.php
+++ b/application/controllers/Statistics_userController.php
@@ -247,9 +247,12 @@ class Statistics_userController extends SurveyController
             //~ $prb->hide();
         }
 
+        $data['aSurveyInfo'] = getSurveyInfo($iSurveyID);
+        $data['graphUrl'] = Yii::app()->getController()->createUrl("admin/statistics/sa/graph");
+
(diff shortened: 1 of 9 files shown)


Shnoulle:
Can not use only `/subviews/content/statistics_user.twig` ?

PS : this broke all potential theme not based on Vanilla : we must really create a real core theme (no new file in ` themes/survey/vanilla/views/` )

gabrieljenik:
> Can not use only `/subviews/content/statistics_user.twig` ?

You mean dropping `layout_statistics_user.twig`?
How's that better? Wouldn't that still be incompatible with current themes?

Shnoulle:
>  How's that better? Wouldn't that still be incompatible with current themes?

Ah right 👍 

I must report an issue : have a real notemplate system … then : we can really imporve LS core but allow template based on something other :)

eddylackmann (approved):

Shnoulle:
No 3.LST fix ?

gabrieljenik:
@cdorin93 should we create a new ticket for this?

Shnoulle:
> @cdorin93 should we create a new ticket for this?

https://bugs.limesurvey.org/view.php?id=15612 :)

gabrieljenik:
haha my bad! :)

PR #1699, opened 2020-12-30
Fixed issue #16930: HTML Editor inline mode not working in Question Editor View

gabrieljenik:


application/helpers/admin/htmleditor_helper.php
```diff
@@ -221,7 +221,8 @@ function getInlineEditor($fieldtype, $fieldname, $fieldtext, $surveyID = null, $
         $toolbaroption = "";
         $sFileBrowserAvailable = '';
         $htmlformatoption = "";
-        $oCKeditorVarName = "oCKeditor_".str_replace("-", "_", $fieldname);
+        $oCKeditorVarName = "oCKeditor_".preg_replace("/[-\[]/", "_", $fieldname);
```
olleharstedt, line 224: Would be nice with a comment here explaining why this is needed.

gabrieljenik: Added it on another PR: https://github.com/LimeSurvey/LimeSurvey/pull/1701

olleharstedt: Sweet! :star2:

PR #1707, opened 2021-01-05
Fixed issue #16902: publicurl disable attchments

gabrieljenik:


Shnoulle:
My fix was

````
 application/controllers/admin/emailtemplates.php             | 5 +++--
 .../admin/emailtemplates/email_language_template_tab.php     | 2 +-
 2 files changed, 4 insertions(+), 3 deletions(-)
 mode change 100644 => 100755 application/controllers/admin/emailtemplates.php
 mode change 100644 => 100755 application/views/admin/emailtemplates/email_language_template_tab.php

diff --git a/application/controllers/admin/emailtemplates.php b/application/controllers/admin/emailtemplates.php
old mode 100644
new mode 100755
index 21e8b52ad7..06973e94bb
--- a/application/controllers/admin/emailtemplates.php
+++ b/application/controllers/admin/emailtemplates.php
@@ -99,7 +99,7 @@ class emailtemplates extends Survey_Common_Action
      */
     function update($iSurveyId)
     {
-        $uploadUrl = Yii::app()->getBaseUrl(true).substr(Yii::app()->getConfig('uploadurl'), strlen(Yii::app()->getConfig('publicurl')) - 1);
+        $uploadUrl = Yii::app()->getBaseUrl(true).substr(Yii::app()->getConfig('uploadurl'), strlen(rtrim(Yii::app()->getBaseUrl(false))));
         // We need the real path since we check that the resolved file name starts with this path.
         $uploadDir = realpath(Yii::app()->getConfig('uploaddir'));
         $sSaveMethod = Yii::app()->request->getPost('save', '');
@@ -112,7 +112,8 @@ class emailtemplates extends Survey_Common_Action
                     foreach ($_POST['attachments'][$langname] as $template => &$attachments) {
                         foreach ($attachments as  $index => &$attachment) {
                             // We again take the real path.
-                            $localName = realpath(urldecode(str_replace($uploadUrl, $uploadDir, $attachment['url'])));
+							$check = str_replace("/",DIRECTORY_SEPARATOR, urldecode(str_replace($uploadUrl, $uploadDir, $attachment['url'])));
+                            $localName = realpath($check);
                             if ($localName !== false) {
                                 if (strpos($localName, $uploadDir) === 0) {
                                     $attachment['url'] = $localName;
diff --git a/application/views/admin/emailtemplates/email_language_template_tab.php b/application/views/admin/emailtemplates/email_language_template_tab.php
old mode 100644
new mode 100755
index 48f5c42169..430873ce7b
--- a/application/views/admin/emailtemplates/email_language_template_tab.php
+++ b/application/views/admin/emailtemplates/email_language_template_tab.php
@@ -64,7 +64,7 @@ $script = array();
     $hideAttacehemtTable = true;
     if (isset($esrow->attachments[$tab])) {
         foreach ($esrow->attachments[$tab] as $attachment) {
-            $script[] = sprintf("prepEmailTemplates.addAttachment($('#attachments-%s-%s'), %s, %s, %s );", $grouplang, $tab, json_encode($attachment['url']), json_encode($attachment['relevance']), json_encode($attachment['size']));
+            $script[] = sprintf("prepEmailTemplates.addAttachment($('#attachments-%s-%s'), %s, %s, %s );", $grouplang, $tab, '"'.addslashes(addslashes($attachment['url'])).'"', json_encode($attachment['relevance']), json_encode($attachment['size']));
         }
         $hideAttacehemtTable = false;
     }
-- 
````

But i really dislike the `addslashes(addslashes` method for windows …

But use relative URL fix issue for publicurl : 

`$uploadUrl = Yii::app()->getBaseUrl(true).substr(Yii::app()->getConfig('uploadurl'), strlen(rtrim(Yii::app()->getBaseUrl(false)))); `

gabrieljenik:
Sorry, is that a better solution? Is there any downside on the solution on the commit?

Shnoulle:
I still don't understand why you use `publicurl`  public url is not used in admin GUI by default …

gabrieljenik:
> I still don't understand why you use `publicurl` public url is not used in admin GUI by default …

Yes, it is, for example in initKcfinder() inside htmleditor_helper for building upload url.

Shnoulle:
> > I still don't understand why you use `publicurl` public url is not used in admin GUI by default …
> 
> Yes, it is, for example in initKcfinder() inside htmleditor_helper for building upload url.

For image url for html email ? I only this this reason :/

gabrieljenik:
> For image url for html email ? I only this this reason :/

https://github.com/LimeSurvey/LimeSurvey/commit/fd935cd5e2a17119857beafb06131ec3644bd53c

Shnoulle:
OK : only for email : email was sent , then : **need** publicurl, why not save relative path ? 
Or more : save filename : file *must** be in [uploadir]/survey/{sid}/files/

Attachment stay on server : don't need publicurl :)

gabrieljenik:
Sorry, I lost the whole point we are reviewing.
What is your suggestion? To re do the fix? Apply yours? Honestly, don;t understand what you pursuit.

Shnoulle:
I have 3 issue with this system, 

1. `publicurl` this fix : you add publicurl checking without reason. The original issue is using absolute URL : we don't need absolute URL
2. Save without good directory separator broek searching it on wondows
3. Use \ in the current js  broke on windows

Then i think : save only filename and we don't have any issue , with or without publicurl, with or without windows etc …

gabrieljenik:
>  save only filename

Aren;t there other pieces of code which expect a more complete path for the filenames?

Shnoulle:
> > save only filename
> 
> Aren't there other pieces of code which expect a more complete path for the filenames?

This was needed when

1. edit
2. Send : https://github.com/LimeSurvey/LimeSurvey/blob/1a8a2eb730d877f4493aa957e5dbe5ac1fd7d30e/application/core/LimeMailer.php#L712-L714

gabrieljenik:
HI Dennis,

Just double checking something.
We implememted the change suggested by you and tried it.
If we try to do the request from localhost (as opposed of using the publicurl), the patch didn't work.
Is that expected?

Should we push the patch like that?

Shnoulle:
I don't know, if i test whole : i push

Just : since `DIRECTORY_SEPARATOR` cause issue (complete path) with 

- publicurl
- windows server

i think it's good to remove all `DIRECTORY_SEPARATOR` and keep only file name.


>  If we try to do the request from localhost (as opposed of using the publicurl), the patch didn't work.

You mean : `http:\\localhost` and have a totally different `publicurl` ?
I don't know , BUT : you can have a broken `publicurl`  on some network , this public url must be used ONLY out of network.

Usage of `publicurl` must be restricted **only** for really absolute url.


PS : anotgher situation where save all path is a really bad idea : LSA+ressources from one server to another one …

Shnoulle:
I get the other for 3.X (assigned it to me), i have a fork currently, want to remove the fork next time i have to update :).

If OK : i get this one at same time.

PR #1708, opened 2021-01-05
Fixed issue #16937: Easy way to launch CDBException

gabrieljenik:
Porting #16481 to LS4.

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -649,6 +649,7 @@ class SurveyRuntimeHelper
             $this->setPrevStep();
             $this->checkIfFinished();
             $this->setStep();
+            $this->saveStartingValues();
 
             // CHECK UPLOADED FILES
             // TMSW - Move this into LEM::NavigateForwards?
@@ -1886,5 +1887,16 @@ class SurveyRuntimeHelper
         }
 
         return $aQuestionClass;
+    }
+
+    /**
+     * Saves the starting values, if any is set
+     */
+    private function saveStartingValues()
+    {
+        // Don't save the starting values if on Welcome screen
+        if ($this->sSurveyMode != 'survey' && $_SESSION[$this->LEMsessid]['step'] == 0) return;
+        
+        LimeExpressionManager::SaveStartingValues();
     }
 }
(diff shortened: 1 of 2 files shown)


Shnoulle:
Original pull request still not merged in 3.X https://github.com/LimeSurvey/LimeSurvey/pull/1480

And 

> for master : we need a better solution : move after setMoveResult ?

because [prefill value broke survey policy  currently](https://bugs.limesurvey.org/view.php?id=16482).

gabrieljenik:
> for master : we need a better solution : move after setMoveResult ?

We have moved it:
- from `StartSurvey()`, which is run inside `initFirstStep()`
- to after `setMoveResult`

This is the list of steps it does for starting up:
```
    /**
     * Init session/params values depending of user moves
     *
     * - It init the needed variables for navigation: initFirstStep, initTotalAndMaxSteps, setMoveResult
     * - Then perform all the needed checks before moving:
     *   + did the participant used browser navigation?
     *   + did he pressed clear cancel, is he a confirmed quota?
     *   + Is the previous step set?
     *   + Is the survey finished?
     *   + Are all the answer validated? (like: participant didn't answered to a mandatory question)
     */
    private function initMove()
    {
        $this->initFirstStep(); // If it's the first time user load this survey, will init session and LEM
        $this->initTotalAndMaxSteps();
        $this->checkIfUseBrowserNav(); // Check if user used browser navigation, or relaoded page
        if ($this->sMove != 'clearcancel' && $this->sMove != 'confirmquota') {
            $this->checkPrevStep(); // Check if prev step is set, else set it
            $this->setMoveResult();
            $this->checkClearCancel();
            $this->setPrevStep();
            $this->checkIfFinished();
            $this->setStep();
            $this->saveStartingValues();
```

Why do we need to move it after `setMoveResult()`?
Shouldn't it be done before so the EM data is available for setting the move?

cdorin93:
Everything seems ok - I cannot reproduce the issue. However, Travis is red. Could you please check once more the PR, @gabrieljenik ?

gabrieljenik:
The testing error is very strange. Also the merge seems difficult. I will just reaply the patch on top of develop and create a new PR

gabrieljenik:
Check https://github.com/LimeSurvey/LimeSurvey/pull/1726

PR #1709, opened 2021-01-05
Develop permission fixes 14551

Shnoulle:
- [x] Create a service for Permission Management
- [x] Create interfaces and traits for permission 
- [x] Create a widget (views) for Permission Management
- [x] Use it for SurveysGroup and SurveysInGroup
- [x] Use it for Surveys

Final issue fixed

- [x] Fixed issue #14551: user can grant more permissions on a survey than he has himself
- [x] Dev: #16801: Permission function by model review (partial)


application/models/services/PermissionManager.php
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * get the permission data
+     * @param string $modelName the model name
+     * @param interger $modelId the model name
+     * @param integer $userId for this user id
+     * @return array[] 
+     */
+    public function getPermissionData($modelName, $modelId, $userId = null)
+    {
+        $aObjectPermissions = Permission::model()->getEntityBasePermissions($modelName);
```
olleharstedt, line 48: If you want to be able to test this with a database scaffold, you need to either inject the Permission model in the constructor, OR factor out access to permission into separate methods and then mock those functions in the test.

olleharstedt (changes requested):

application/models/services/PermissionManager.php
olleharstedt: _without_ database scaffold, sorry for the typo

Shnoulle: Permission::model()->getEntityBasePermissions didn't use database but model definition. 

See https://github.com/LimeSurvey/LimeSurvey/blob/186bfd8870083724a227d569c13a1c395b01f804/application/models/LSActiveRecord.php#L464 i can use it directly : OK ?

Shnoulle: But still have `Permission::model()->hasPermission($modelId, $modelName, $sPermission, $crud, $this->user->id),` …

Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/1709/commits/c56bb3772602e53b43a931e663f8949da8ded2aa  ?

Shnoulle:
PermissionManager because it's to manage Permission. 

About hasPermssion part : all this function don't use current model , bnut call itself … `$this->findByAttributes` etc …

Shnoulle:
> Exactly, or `Pemission::mode()` instead of new. :)

I work on save an i need a new … 


````
            $oCurrentPermission = $this->permission->find(
                "entity = :entity AND entity_id = :entity_id AND uid = :uid AND permission = :permission",
                array(
                    ":entity" => $modelName,
                    ":entity_id" => $modelId,
                    ":uid" => $userId,
                    ":permission" => $sPermission
                )
            );
            if(empty($oCurrentPermission)) {
                $oCurrentPermission = new $this->permission; // ?????
                $oCurrentPermission->entity = $modelName;
                $oCurrentPermission->entity_id = $modelId;
                $oCurrentPermission->uid = $userId;
                $oCurrentPermission->permission = $sPermission;
            }
```

I need to find a better solution …

olleharstedt:
```php
                $oCurrentPermission = new $this->permission; // ?????
```

2 alternatives:

1. Inject a factory class
2. Move "new" to separate method and mock that method:

```php
public function getNewPermission()
{
  return new Permission();  // Hard-coded dependency but can be mocked by mocking getNewPermission()
}
```

olleharstedt:
> PermissionManager because it's to manage Permission.

That's fine for now, but it leaves it open to grow too big. :) Be ready to add new functions in classes with more specific names later on.

Shnoulle:
> > PermissionManager because it's to manage Permission.
> 
> That's fine for now, but it leaves it open to grow too big. :) Be ready to add new functions in classes with more specific names later on.

My opinion : 2 function here : get the data for a Object, and set the data for an object : manage the permission only, don't do other things. I think we can create another service for : current Permission (current rights).

Another step can be create a Global model to just use the same Service …

Shnoulle:
@olleharstedt about https://github.com/LimeSurvey/LimeSurvey/pull/1709/commits/e83d62af50fce4bff6d11bc26ab23d2e5b24f8f5#diff-f58954e46b07acfed2e04e58e5e252b923d7d7d4ed802f83a36bf72101ca0b45R57


Maybe it's more clean to create a model->getCurrentPermission() ?

I mean :
- `SurveysGroups::model()->getCurrentPermission($sPermission, $crud = 'read', $userId = null)`
- `Survey::model()->getCurrentPermission($sPermission, $crud = 'read', $userId = null)`
-  `SurveysInGroup::model()->getCurrentPermission($sPermission, $crud = 'read', $userId = null)`

Maybe time to rename SurveysGroups to SurveysGroup too …

Shnoulle:
> 
> Maybe it's more clean to create a model->getCurrentPermission() ?
> 

In PermissionManagerService : send the model ($SurveysInGroup, $Surveys, $SurveysGroups …)

application/controllers/SurveysGroupsPermissionController.php
```diff
@@ -339,18 +339,25 @@ public function actionSave($id)
...
-                    $id,
-                    $entity,
-                    $aPermissionSet
-                );
+                $success = $success && !$PermissionManagerService->setPermissions($uid, $entity, $id);
             }
         }
-        App()->setFlashMessage("Surveys groups permissions were successfully updated");
+        if($success) {
+            App()->setFlashMessage("Surveys groups permissions were successfully updated");
```
olleharstedt, line 358: Missing gT()?

olleharstedt (changes requested):

olleharstedt:
> Maybe it's more clean to create a model->getCurrentPermission() ?

I'd prefer not to. "Fat models" mean they grow _really_ fat. Louis even made a model into a singleton, which makes it harder to read and understand. Best case scenario: ActiveRecord model contains nothing except rules and relations. All other more complex database operations are in service classes or in a "ModelRepository". Remember: It's much harder to read code than to write code. No shortcuts. :joy:

olleharstedt:
> Maybe it's more clean to create a model->getCurrentPermission() ?

I'd prefer not to. I think the models are already too big.

Shnoulle:
> > Maybe it's more clean to create a model->getCurrentPermission() ?
> 
> I'd prefer not to. I think the models are already too big.

Maybe : but the Permission on model are linked to model, and the model use Permission.

For example : `$havereponsePermission =Survey::model()->findByPk($iSurveyId)->hasPermission('response','read');`
Seems better than` $havereponsePermission = Permission::model()->hasSurveyPermission($iSurveyId,'response','read');`

We need specific test for each model type . You think a service to replace hasSurveyPermission, hasGlobalPermission, hasSurveysGroupsPermission, hasSurveysInGroupPermission (and hasTemplatePermission, asParticipantPermission, asLabelPermission, or (why not) asQuestionPermission).

If we add a LSActiveRecord->hasPermission we can easily : 

- Add the current check for Survey, SurveysGroups, SurveysInGroup
- Create User, Template, Label … using directly Global permission only
- Create Permission dependance directly (for example : Question->hasPermisson check Survey->hasPermission)

Shnoulle:
> Maybe a trait could work?

Unsure … 

Basically : 
````
function hasMyObjectPermission($objectId, $permission, $crud, $userId) {
    return Permission::hasPermission(0, "global", MyObject", $crud , $userId)
        || Permission::hasPermission($objectId, "MyObject", $permission, $crud , $userId);
}
````

But :  
- $crud can need to be updated
- inherited system can need to be updated (Survey must check SurveysInGroup)
- etc …

**Comment by olleharstedt** [comment]
Maybe create a dev Mantis ticket for this?

**Comment by Shnoulle** [comment]
Like this one : https://bugs.limesurvey.org/view.php?id=16801

**Comment by olleharstedt** [comment]
Oh right. :) Yep.

File: application/models/services/PermissionManager.php [hunk 3]
```diff
@@ -185,4 +183,23 @@ public function getNewPermission()
...
+     * @param string $sPermission the specific permission name
+     * @param string $crud the crud ('create', 'read', 'update', 'delete', 'import', 'export')
+     * @return boolean
+     */
+    public function getCurrentPermission($modelName, $modelId, $sPermission, $crud, $userId)
+    {
+        /* string : function name to use for final Permission */
+        $permissionFunctionName = "has{$modelName}Permission";
+        if(method_exists($this->permission,$permissionFunctionName)) {
+            return $this->permission->$permissionFunctionName($modelId, $sPermission, $crud, $userId);
```
olleharstedt, line 200: Interfaces are better than variable variables.......

olleharstedt: This is impossible for a static analyser to understand.

Shnoulle: I don't have idea on the way to do here. 

I **need** to check hasSurveysGroupsPermission or hasSurveysInGroupPermission (and hasSurveyPermission) …

olleharstedt: Use `instanceof`?

olleharstedt: Or add a new interface for models who have permission. `HasPermissionCheckInterface`.

olleharstedt: If the implementation is the same for all classes implementing the interface, you can use a trait. Otherwise, you will have to add the method inside each class manually.

application/models/services/PermissionManager.php
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * @todo : Save Permission by POST value according to current user permssion
+     * @see Permission::setPermissions
+     * @param mixed $iUserID
+     * @param string $modelName
+     * @param mixed $modelId
+     * @return boolean
+     */
+    public function setPermissions($userId, $modelName, $modelId)
```
olleharstedt, line 108: Maybe it's easier to code if you send in the model class instead of just the name? And instead of `string` it will have type `HasPermissionCheckInterface`?

application/models/services/PermissionManager.php
Shnoulle: > Use `instanceof`?

See https://github.com/LimeSurvey/LimeSurvey/pull/1709#issuecomment-754775785 this need to move checkPermission function in model … create a really simple checkPermission in LSActive record , and replace in Survey, SurveysGroups,SurveysInGroup

> If the implementation is the same for all classes implementing the interface, you can use a trait. Otherwise, you will have to add the method inside each class manually.

- Surveys must update $crud, check Global, check SurveysInGroup, check Own
- SurveysGroups must check global + own
- SurveysInGroup  : update crud + check Global, check Own
- Template : really specific : global + $sTemplateName as object name
- Roles : just need to check global

Then : best seems to send model as LSActiverecord and call $model->getPermission new function …

application/models/services/PermissionManager.php
Shnoulle: See  https://github.com/LimeSurvey/LimeSurvey/pull/1709#issuecomment-754775785 and your comment : https://github.com/LimeSurvey/LimeSurvey/pull/1709#issuecomment-754918857 …

application/models/services/PermissionManager.php
olleharstedt: > create a really simple checkPermission in LSActive record

Yes, but please don't do this without adding an interface, to distinguish models with permission check from those that haven't.

Shnoulle: Arg … without interface : https://github.com/LimeSurvey/LimeSurvey/pull/1709/files#diff-ebf85ba8a486bce5d44fcf65c87a272498d2a841b68fb535e69db3eaf879ade1R485

Shnoulle: I don't understand : all object are already child of LSActiveRecord … you want a LSActivaveRecord without intefcae an one implement a new interface ?

Shnoulle:
Maybe i set 14551: user can grant more permissions on a survey than he has himself to new again … wait for another dev to fix it for Survey …

application/models/services/PermissionManager.php
olleharstedt: It's about _intent_. Not all classes that inherit from LSActiveRecord have permissions attached to them. Also that a class should have a clear purpose, not "contain everything" as LSActiveRecord does.

olleharstedt: > Arg … without interface : https://github.com/LimeSurvey/LimeSurvey/pull/1709/files#diff-ebf85ba8a486bce5d44fcf65c87a272498d2a841b68fb535e69db3eaf879ade1R485

Cool, then you can use interface + trait instead. :)

olleharstedt: You'd have a new interface "HasPermissionInterface" and a new trait that implements it, "HasPermissionTrait".

Shnoulle:
Looking at ci : `Error: Call to a member function getCurrentPermission() on null` … oh yes model can return null …
Then can not work for `create`

application/models/services/PermissionManager.php
Shnoulle: Thank you

Shnoulle:
interface + trait  , name can be reviewed

application/models/Interfaces/PermissionInterface.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+interface PermissionInterface
+{
+    public function getOwnerId();
+    public static function getPermissionData($key = null);
```
olleharstedt, line 6: PHPUnit cannot mock static functions. Avoid if possible.

application/models/Traits/PermissionTrait.php
```diff
@@ -0,0 +1,56 @@
...
+        return array();
+    }
+
+    /**
+     * Get minimal permission name (for read value)
+     * @return null|string
+     */
+    public static function getMinimalPermissionRead()
+    {
+        return null;
```
olleharstedt, line 32: Maybe use:

```php
throw new Exception('Not implemented');
```

application/models/Interfaces/PermissionInterface.php
Shnoulle: It's really a static system fact : it's definition array …

olleharstedt: So it needs no mocking?

application/models/Traits/PermissionTrait.php
Shnoulle: No

SureysInGroup didn't have minimal Permission : it's desired behaviour. 

https://github.com/LimeSurvey/LimeSurvey/blob/4b569fc2e6d55733176f782695dd84fbf4fd8206/application/models/Permission.php#L344

application/models/Interfaces/PermissionInterface.php
Shnoulle: No readon : https://github.com/LimeSurvey/LimeSurvey/blob/4b569fc2e6d55733176f782695dd84fbf4fd8206/application/models/Survey.php#L2142

Shnoulle:
Don't see how to fix 

> IssueId: 45727350  Message: The method ``getPrimaryKey()`` does not exist on ``PermissionInterface``. Since it exists in all sub-types, consider adding an abstract or default implementation to ``PermissionInterface``.  Filename: application/models/services/PermissionManager.php  LineNumber: 74  Link: https://scrutinizer-ci.com/g/LimeSurvey/LimeSurvey/inspections/344603d9-1bf8-45fc-b0e8-1b3b442a695e/issues/files/application/models/services/PermissionManager.php?status=new&orderField=path&order=asc&honorSelectedPaths=0&issueId=45727350

Shnoulle:
travis issue `The command "wget "https://github.com/mozilla/geckodriver/releases/download/v0.27.0/geckodriver-v0.27.0-linux64.tar.gz"" failed and exited with 8 during .` :D

olleharstedt (approved):

cdorin93:
So, shall this one be merged in the next few days :) ?

olleharstedt:
> So, shall this one be merged in the next few days :) ?

Yes, I just want to do a local checkout and read through it first.

Shnoulle:
OK, next : [Add a way to extend Survey list in administration (for Permission)](https://bugs.limesurvey.org/view.php?id=16459) maybe a new Services ? 

Permission need it (if you want to show all survey to user name "AllSurvey" : you need allow update. But maybe other plugins can use it ?

olleharstedt:
See my comment on Discord.

PR #1710, opened 2021-01-06
Bugfix: Register assets in short free text

vitormattos:
How to reproduce:

disable the Limesurvey XSS filter at
Global settings -> Security -> “Filter HTML for XSS” = “No”.

Create a "Short Free Text" question.

Add a script to the question. Example: `console.log (" I'm here ")`

Preview the issue.

Check to see if you’re here in the developer tools console of your browser.

The correction:
The method needs to be refactored but the change resolves the bug.

Dev:

### DIFF
--- a/application/core/QuestionTypes/ShortFreeText/RenderShortFreeText.php
+++ b/application/core/QuestionTypes/ShortFreeText/RenderShortFreeText.php
@@ -31,6 +31,7 @@ class RenderShortFreeText extends QuestionBaseRenderer
     public function render($sCoreClasses = '')
     {
         $result =  @do_shortfreetext($this->aFieldArray);
+        $this->registerAssets();
         return $result;
 
         $answer = '';


olleharstedt:
@Shnoulle Can you have a look?

olleharstedt:
Do we have an issue for this on https://bugs.limesurvey.org/view_all_bug_page.php ?

vitormattos:
I didn't look. I found the bug and found the way to fix it.

Shnoulle:
Why only with short free text ? If asset is badly register on short free text : it must surely badly register on another questions …

My opinion : fix it really and not only for one question type.


Else : i use more my plugin than included script ;) for javascript …

vitormattos:
The method `render` of other fields is implemented differently. I came up with the solution by seeing how the other fields were rendering because in other fields it works.

olleharstedt:
Thank you for your contribution! Let us know if any similar issues pops up.

PR #1711, opened 2021-01-06
Fixed issue #16948: Uploading an image through the theme options fails

gabrieljenik:
Removing call for $checkImageFilename

I think bug was introduced here:
https://github.com/LimeSurvey/LimeSurvey/commit/3376b3ca24d0209015080c48ab1dd33593696a2d


c-schmitz:
Thanks!

Shnoulle:
The original security issue didn't exist anymore ?

> Remote Code Execution (RCE) in template options file uploader

PR #1714, opened 2021-01-07
Fixed issue #16949: Uploading an image through the theme options fails

gabrieljenik:


gabrieljenik:
This error appears when testing.
Not sure it is related to this PR.
```
ls\tests\ChangeQuestionTemplateTest::selectQuestionThemeForQuestion
Facebook\WebDriver\Exception\UnrecognizedExceptionException: Element <select id="question_template" class="form-control" name="question_template"> is not clickable at point (1143,478) because another element <div id="loader-generalSettingsLoader" class=" loader--loaderWidget ls-flex ls-flex-column align-content-center align-items-center"> obscures it
``

There is also a conflict,
Maybe we should pick up later?
Let me know @cdorin93

olleharstedt:
That error can be ignored if it's only one instance that fails (the others are green). It's a timing issue, some popup or whatever is covering a button.

cdorin93:
Ok, then sounds good, @gabrieljenik , need to talk more about the conflict? Or it can be resolved without too much hassle?

gabrieljenik:
Resolved conflict.
Maybe it should be tested again? Not sure what's the protocol  here :)

eddylackmann (approved):

PR #1718, opened 2021-01-14
Fixed issue #16954 & #16237: Themes Variations are saved, but not applied, can't be used

gabrieljenik:
Added new attribute to variation tag: data-mode="replace".
When that is present the related CSS file is saved to the framework_css field. If not, is added to the files_css list.

This is because bootswatch and fruity manage variations differently.

Moved variation selector from Font to Simple options.


cdorin93:
@gabrieljenik , could you please rebase for master?

gabrieljenik:
Here: https://github.com/LimeSurvey/LimeSurvey/pull/1727

PR #1721, opened 2021-01-15
Fixed issue #16853: Inconsistent filter behaviour when create token

gabrieljenik:


gabrieljenik:
@olleharstedt @Shnoulle ping

PR #1733, opened 2021-01-28
Fixed Issue #17026: Custom themes not uploading files

gabrieljenik:
Code for backward compatiblity with custom themes that are expecting to upload images to upload subaction.
That happens as their options.twig are outdated.
Now that the upload subaction doesn't handle all uploads, weneed to dispatch to the proper.

### DIFF
--- a/application/controllers/admin/themes.php
+++ b/application/controllers/admin/themes.php
@@ -178,7 +178,8 @@ class themes extends Survey_Common_Action
         $action = returnGlobal('action');
         if ($action == 'templateuploadimagefile') {
             return $this->templateuploadimagefile();
-        } elseif ($action == 'templateupload') {
+        } 
+        if ($action == 'templateupload') {
             return $this->templateupload();
         }
 
--- a/application/controllers/admin/themes.php
+++ b/application/controllers/admin/themes.php
@@ -172,6 +172,16 @@ class themes extends Survey_Common_Action
      */
     public function upload()
     {
+        // Code for backward compatiblity with custom themes that are expecting to upload images to upload subaction. 
+        // That happens as their options.twig are outdated.
+        // Now that the upload subaction doesn't handle all uploads, weneed to dispatch to the proper.
+        $action = returnGlobal('action');
+        if ($action == 'templateuploadimagefile') {
+            return $this->templateuploadimagefile();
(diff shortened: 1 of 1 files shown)


application/controllers/admin/themes.php
```diff
@@ -172,6 +172,16 @@ public function tmp($id)
      */
     public function upload()
     {
+        // Code for backward compatiblity with custom themes that are expecting to upload images to upload subaction. 
+        // That happens as their options.twig are outdated.
+        // Now that the upload subaction doesn't handle all uploads, weneed to dispatch to the proper.
+        $action = returnGlobal('action');
```
Shnoulle, line 178: `returnGlobal` **must** be deprecated … please …

application/controllers/admin/themes.php
```diff
@@ -172,6 +172,16 @@ public function tmp($id)
...
+        if ($action == 'templateuploadimagefile') {
```
Shnoulle, line 179: Mersonnaly , i prefer this : 
```` 
        if ($action == 'templateuploadimagefile') {
            return $this->templateuploadimagefile();
        }
        if ($action == 'templateupload') {
            return $this->templateupload();
        }
        // else : continue
````
More clear

application/controllers/admin/themes.php
gabrieljenik: I understand, but keep in mind this is used 5 lines below in the same controller. Is this the right time to deprecate it?

application/controllers/admin/themes.php
gabrieljenik: Done

Shnoulle: Thanks

application/controllers/admin/themes.php
Shnoulle: Right … (it was not deprecated before ??)

maziminke:
@cdorin93 can we merge this before the next release? It fixes broken custom themes.

cdorin93:
Sure - @maziminke , did you have time to test it?

maziminke:
@cdorin93 I just made another test and with the fix the upload at custom theme works fine again!

eddylackmann (approved):

PR #1734, opened 2021-02-01
Fixed issue #16756: Error with SUM on array text

gabrieljenik:
Adding total calculations based on relevance


gabrieljenik:
Wrong based

PR #1735, opened 2021-02-01
Fixed issue #16576: only part of the label set is imported

gabrieljenik:
Adding total calculations based on relevance


gabrieljenik:
Wrong based, again. Sorry

PR #1745, opened 2021-02-02
Update SurveyAdministrationController.php

ritapas:
Dev: proposed fix: can the argument string for gT which is split into rows 596 and 597 be marged in one row? Otherwise the translation is messy.


Shnoulle (approved):
:+1: ❤️ 
See
Capture d’écran du 2021-02-02 18-30-39

olleharstedt:
Thank you!

PR #1747, opened 2021-02-03
Fixed issue #17054: Public stats don't use the theme set at survey, but at global settings

gabrieljenik:
Use survey template


Shnoulle (approved):
Seems OK :)

PR #1748, opened 2021-02-04
Display actual model errors if survey import fails

TonisOrmisson:
after 7ece09b8cb737adbac6bc2b49b4d85cdaf6acd98 we we ended up errors while copying/importing surveys but had only a vague general "error" to work with. 

This will add the yii standard model  errorSummary so that one can understand what actually went wrong.


Shnoulle:
Why not 
`$results['error'] = CHtml::errorSummary($newSurvey, gT("Unable to import survey."));` ?

Maybe :)

else :+1:

TonisOrmisson:
@Shnoulle of course, thanks

Shnoulle (approved):

Shnoulle:
Thanks to you :)

eddylackmann (approved):

PR #1750, opened 2021-02-04
fix composer.json license format to match composer standard-V3

TonisOrmisson:
#1454 for V3
composer license has an invalid value 
syncing to master


TonisOrmisson:
otherwise looks like that for packagist:
image

PR #1754, opened 2021-02-08
Fixed issue #16937: Easy way to launch CDBException

gabrieljenik:
SaveStartingValues was moved from StartSurvey to InitMove. InitMove is not being called from the preview mode. So now, we call SaveStartingValues from the preview.

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -1748,6 +1748,9 @@ class SurveyRuntimeHelper
             $this->aMoveResult = LimeExpressionManager::JumpTo($qSec + 1, 'question', false, true);
             $this->aStepInfo = LimeExpressionManager::GetStepIndexInfo($this->aMoveResult['seq']);
         }
+        
+        // "Save" starting values to response. 
+        // In the case of preview will not be "saved"", but will be incorporated in the response handled in memory.
         LimeExpressionManager::SaveStartingValues();
     }
 
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -1748,6 +1748,7 @@ class SurveyRuntimeHelper
             $this->aMoveResult = LimeExpressionManager::JumpTo($qSec + 1, 'question', false, true);
             $this->aStepInfo = LimeExpressionManager::GetStepIndexInfo($this->aMoveResult['seq']);
         }
+        LimeExpressionManager::SaveStartingValues();
     }
 
 
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -642,6 +642,7 @@ class SurveyRuntimeHelper
             $this->setPrevStep();
             $this->checkIfFinished();
             $this->setStep();
+            $this->saveStartingValues();
 
             // CHECK UPLOADED FILES
             // TMSW - Move this into LEM::NavigateForwards?
@@ -1866,5 +1867,16 @@ class SurveyRuntimeHelper
         }
 
         return $aQuestionClass;
+    }
+
+    /**
+     * Saves the starting values, if any is set
+     */
+    private function saveStartingValues()
+    {
+        // Don't save the starting values if on Welcome screen
+        if ($this->sSurveyMode != 'survey' && $_SESSION[$this->LEMsessid]['step'] == 0) return;
+
+        LimeExpressionManager::SaveStartingValues();
     }
 }
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -4619,56 +4619,7 @@ class LimeExpressionManager
         $LEM->indexGseq = array();
         $LEM->indexQseq = array();
         $LEM->qrootVarName2arrayFilter = array();
-        // set seed key if it doesn't exist to be able to pass count of startingValues check at next IF
-        if (array_key_exists('startingValues', $_SESSION[$LEM->sessid]) && !array_key_exists('seed', $_SESSION[$LEM->sessid]['startingValues'])) {
-            $_SESSION[$LEM->sessid]['startingValues']['seed'] = '';
-        }
-            
-        // NOTE: now that we use a seed, count($_SESSION[$LEM->sessid]['startingValues']) start at 1
-        if (isset($_SESSION[$LEM->sessid]['startingValues']) && is_array($_SESSION[$LEM->sessid]['startingValues']) && count($_SESSION[$LEM->sessid]['startingValues']) > 1) {
-            foreach ($_SESSION[$LEM->sessid]['startingValues'] as $k => $value) {
-                if (isset($LEM->knownVars[$k])) {
-                    $knownVar = $LEM->knownVars[$k];
-                } elseif (isset($LEM->qcode2sgqa[$k])) {
-                    $knownVar = $LEM->knownVars[$LEM->qcode2sgqa[$k]];
-                } else {
-                    continue;
-                }
-                if (!isset($knownVar['jsName'])) {
-                    continue;
-                }
-                switch ($knownVar['type']) {
-                    case Question::QT_D_DATE: //DATE
-                        if (trim($value) == "") {
-                            $value = null;
-                        } else {
-                            // We don't really validate date here, anyone can send anything : forced too
-                            $dateformatdatat = getDateFormatData($LEM->surveyOptions['surveyls_dateformat']);
-                            $datetimeobj = new Date_Time_Converter($value, $dateformatdatat['phpdate']);
-                            $value = $datetimeobj->convert("Y-m-d H:i");
-                        }
-                        break;
-                    case Question::QT_N_NUMERICAL: //NUMERICAL QUESTION TYPE
-                    case Question::QT_K_MULTIPLE_NUMERICAL_QUESTION: //MULTIPLE NUMERICAL QUESTION
-                        if (trim($value) == "") {
-                            $value = null;
-                        } else {
-                            $value = sanitize_float($value);
-                        }
-                        break;
-                    case Question::QT_VERTICAL_FILE_UPLOAD: //File Upload
-                        $value = null;  // can't upload a file via GET
-                        break;
-                }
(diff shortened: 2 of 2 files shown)


application/helpers/SurveyRuntimeHelper.php
```diff
@@ -1867,4 +1872,15 @@ public function getCurrentQuestionClasses($iQid)
...
         return $aQuestionClass;
     }
+
+    /**
+     * Saves the starting values, if any is set
+     */
+    private function saveStartingValues()
+    {
+        // Don't save the starting values if on Welcome screen
+        if ($this->sSurveyMode != 'survey' && $_SESSION[$this->LEMsessid]['step'] == 0) return;
```
Shnoulle, line 1882: Question : what happen if welcome page is false ?

gabrieljenik: if welcome page is false, step will be 1

application/helpers/expressions/em_manager_helper.php
```diff
@@ -9800,6 +9752,75 @@ public static function htmlSpecialCharsUserValue($string)
...
+    /**
+     * Save the starting values to the database
+     */
+    public static function SaveStartingValues()
+    {
+        // Init
+        $LEM =& LimeExpressionManager::singleton();
+
+        // If not starting values, retun
+        if (!isset($_SESSION[$LEM->sessid]['startingValues'])) return false;
```
olleharstedt, line 9765: Can you check this code with PSR-12, please?

olleharstedt (changes requested):

application/helpers/expressions/em_manager_helper.php
gabrieljenik: Done. Thanks!

olleharstedt: :partying_face:

application/helpers/expressions/em_manager_helper.php
```diff
@@ -9800,6 +9752,81 @@ public static function htmlSpecialCharsUserValue($string)
         $string = str_replace(["{","}"], ["&#123;","&#125;"], $string);
         return $string;
     }
+
+    /**
+     * Save the starting values to the database
```
olleharstedt, line 9757: Missing @return boolean|void, but can be only @return void, right?

application/helpers/expressions/em_manager_helper.php
```diff
@@ -9800,6 +9752,81 @@ public static function htmlSpecialCharsUserValue($string)
...
+     */
+    public static function SaveStartingValues()
+    {
+        // Init
+        $LEM =& LimeExpressionManager::singleton();
```
olleharstedt, line 9762: `=&` is PHP4 legacy, better to just use `=` these days.

olleharstedt: Sorry, I know you just copied these lines. >< Now you have to pay the price! :laughing:

olleharstedt (changes requested):

application/helpers/expressions/em_manager_helper.php
Shnoulle: My opinion : keep same method in big file like this. Except if really rewrite , no ?

application/helpers/expressions/em_manager_helper.php
```diff
@@ -9800,6 +9752,81 @@ public static function htmlSpecialCharsUserValue($string)
...
+
+        // If not starting values, retun
+        if (!isset($_SESSION[$LEM->sessid]['startingValues'])) {
+            return false;
+        }
+        if (!is_array($_SESSION[$LEM->sessid]['startingValues'])) {
+            return false;
+        }
+        if (!array_key_exists('startingValues', $_SESSION[$LEM->sessid])) {
```
Shnoulle, line 9771: Question : it's not the same than `if (!isset($_SESSION[$LEM->sessid]['startingValues'])) {` ?

We can not remove this tes ?

application/helpers/expressions/em_manager_helper.php
```diff
@@ -9800,6 +9752,81 @@ public static function htmlSpecialCharsUserValue($string)
...
+                        break;
+                }
+                $_SESSION[$LEM->sessid][$knownVar['sgqa']] = $value;
+                $LEM->updatedValues[$knownVar['sgqa']] = array(
+                    'type' => $knownVar['type'],
+                    'value' => $value,
+                );
+                unset($_SESSION[$LEM->sessid]['startingValues'][$k]);
+            }
+            $LEM->_UpdateValuesInDatabase();
```
Shnoulle, line 9827: The issue are not totally fixed : you fix only DATE and NUMBER , but not all other values  need `self::checkValidityAnswer` function 

See https://github.com/LimeSurvey/LimeSurvey/pull/1480/files 

(but you fix the [issue with Survey Policy](https://bugs.limesurvey.org/view.php?id=16482) i think)

Shnoulle (commented):
Lack of self::checkValidityAnswer ? No ?

application/helpers/expressions/em_manager_helper.php
gabrieljenik: You are right. 
checkValidityAnswer got lost in translation.

Will review it.

gabrieljenik: Added.

application/helpers/expressions/em_manager_helper.php
gabrieljenik: @olleharstedt Do you think I should change it only here or everywhere on this file?
I would just do nothing on this case.

application/helpers/expressions/em_manager_helper.php
gabrieljenik: They are very similar. Isset checks for NULL values.
https://www.geeksforgeeks.org/difference-between-isset-and-array_key_exists-function-in-php

I will remove it.

application/helpers/expressions/em_manager_helper.php
olleharstedt: Bwah. Either only here or do nothing. We can create a new dev issue: "Remove PHP4 idioms".

application/helpers/expressions/em_manager_helper.php
olleharstedt: Why return `false` if the return value is never checked, tho?

application/helpers/expressions/em_manager_helper.php
gabrieljenik: Sorry. I say nothing this time and vote for the new issue.

application/helpers/expressions/em_manager_helper.php
gabrieljenik: Well, you never know if that piece of data will be usefull to someone in the future
I would say is a habit

Shnoulle: > Well, you never know if that piece of data will be usefull to someone in the future
> I would say is a habit

Yes, like it :) but not necessary …

application/helpers/expressions/em_manager_helper.php
```diff
@@ -9800,6 +9752,86 @@ public static function htmlSpecialCharsUserValue($string)
...
+                } elseif (isset($LEM->qcode2sgqa[$k])) {
+                    $knownVar = $LEM->knownVars[$LEM->qcode2sgqa[$k]];
+                } else {
+                    continue;
+                }
+                if (!isset($knownVar['jsName'])) {
+                    continue;
+                }
+                switch ($knownVar['type']) {
+                    case Question::QT_D_DATE: //DATE
```
Shnoulle, line 9796: Still needed ? Can not use only self::checkValidityAnswer ?

Shnoulle:
@gabrieljenik no way to set it as default value ? 

Because here : i don't see interest with lighter part already done in 3.X : https://github.com/LimeSurvey/LimeSurvey/commit/6142626ed31031576f6644d25cddad8febde91c8

application/helpers/expressions/em_manager_helper.php
gabrieljenik: Honestly, I don't recall that much.
This code is moved from here:https://github.com/LimeSurvey/LimeSurvey/pull/1754/files/50a5b0c212d5c1c9f041ad0f81224c007ced3108#diff-fa8d35264aa59c85d37ca2d1756472996499f06fb346865594378102f5f7abc0L4640

Don';t want to be rude, but don't think doing re-engineering right now (in the scope of this ticket) will help stability :)

From what I see, checkValidityAnswer don't have much for Dates.
I know the code is messy. Was like that.
Still, I think that if we go in the direction of re-engineering instead of re-organizing, we may create lot of unexpected bugs.

Shnoulle: Right , it's on todo :)

https://github.com/LimeSurvey/LimeSurvey/blob/3ff0af260ca43f09ad87a2ea3171a385e1e074dd/application/helpers/expressions/em_manager_helper.php#L9667-L9669

gabrieljenik:
> @gabrieljenik no way to set it as default value ?
> 
> Because here : i don't see interest with lighter part already done in 3.X : [6142626](https://github.com/LimeSurvey/LimeSurvey/commit/6142626ed31031576f6644d25cddad8febde91c8)

Sorry, I am not following... What do you mean?

Shnoulle:
Same issue in 3.LTS fixed by only this : https://github.com/LimeSurvey/LimeSurvey/commit/6142626ed31031576f6644d25cddad8febde91c8

Your commit are really bigger : i think there are other issue fixed maybe ?

gabrieljenik:
I think this may help fullfill your question:
https://github.com/LimeSurvey/LimeSurvey/pull/1708#issuecomment-754638147

Shnoulle:
OK !
Thank you https://bugs.limesurvey.org/view.php?id=16482

olleharstedt:
Problem with reviewing this change: Really hard to assess risk of moving stuff in runtime helper...

c-schmitz:
Then it should probably go to develop, shouldn't it?

c-schmitz:
Gabriel, can you update the PR, please?

gabrieljenik:
Closing as it is being followed up on https://github.com/LimeSurvey/LimeSurvey/pull/2272

PR #1755, opened 2021-02-10
Fixed issue #17075: User roles : all rights are enabled

Shnoulle:
Dev: clarify functionname in Permission
Dev: use directly DB checking

### DIFF
--- a/application/models/Permission.php
+++ b/application/models/Permission.php
@@ -524,7 +524,7 @@ class Permission extends LSActiveRecord
      * @param string $sEntityName string The entity name
      * @param $sPermission string Name of the permission
      * @param $sCRUD string The permission detail you want to check on: 'create','read','update','delete','import' or 'export'
-     * @param $iUserID integer User ID - if not given the one of the current user is used, 0 is used for role permission (not related to user)
+     * @param $iUserID integer User ID - if empty : use the current user
      * @return bool True if user has the permission
      */
     public function hasPermission($iEntityID, $sEntityName, $sPermission, $sCRUD = 'read', $iUserID = null)
@@ -555,7 +555,7 @@ class Permission extends LSActiveRecord
 
         /* Always return true for CConsoleApplication (before or after plugin ? All other seems better after plugin) */
         // TODO: see above about entry script and superadmin
-        if (is_null($iUserID) && Yii::app() instanceof CConsoleApplication) {
+        if (empty($iUserID) && Yii::app() instanceof CConsoleApplication) {
             return true;
         }
 
@@ -566,16 +566,15 @@ class Permission extends LSActiveRecord
         }
         $sCRUD = $sCRUD . '_p';
 
-        /* Always return false for guests */
-        // TODO: should not be necessary
+        /* Be sure to have an user id */
         $iUserID = $this->getUserId($iUserID);
-        if (!$iUserID && $iUserID !== 0) {
+        /* Always return false for guests */
+        if (!$iUserID) {
             return false;
         }
 
-        /* Always return true if you are the owner : this can be done in core plugin ? */
-        // TODO: give the rights to owner adding line in permissions table, so it will return true with the normal way
-        if ($iUserID && $iUserID == $this->getEntityOwnerId($iEntityID, $sEntityName)) {
+        /* Always return true if user are the owner of entity*/
+        if ($iUserID == $this->getEntityOwnerId($iEntityID, $sEntityName)) {
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


Shnoulle:
- hasGlobalPermission :  Returns true if a user has global permission for a certain action.
- hasSurveyPermission : Checks if a user has a certain permission in the given survey

I think hasRolePermission must return true if user have a certain permission in the given role … but not (before this fix)

application/models/Permission.php
```diff
@@ -575,7 +575,7 @@ public function hasPermission($iEntityID, $sEntityName, $sPermission, $sCRUD = '
 
         /* Always return true if you are the owner : this can be done in core plugin ? */
         // TODO: give the rights to owner adding line in permissions table, so it will return true with the normal way
-        if ($iUserID == $this->getEntityOwnerId($iEntityID, $sEntityName)) {
+        if ($iUserID && $iUserID == $this->getEntityOwnerId($iEntityID, $sEntityName)) {
```
olleharstedt, line 578: Can you be more clear what you test with user id? empty? null? the string "0"?

application/models/Permission.php
```diff
@@ -677,16 +677,26 @@ public function hasSurveyPermission($iSurveyID, $sPermission, $sCRUD = 'read', $
...
-     * @param integer $iUserID integer User ID - if not given the one of the current user is used
-     * @return bool True if user has the permission
+     * Returns true if a role has permission for crud
+     * @param integer $roleId
+     * @param string $sPermission 
+     * @param string $sCRUD The permission detailsyou want to check on: 'create','read','update','delete','import' or 'export'
+     * @return bool allwoedpermssion
      */
-    public function hasRolePermission($iRoleId, $sPermission, $sCRUD = 'read')
+    public function RoleHasPermission($iRoleId, $sPermission, $sCRUD = 'read')
```
olleharstedt, line 686: Not PSR-12 compliant name, should start with lower letter.

olleharstedt (changes requested):

application/models/Permissiontemplates.php
```diff
@@ -291,9 +291,14 @@ public function removeCdataFormat($node)
         return trim(preg_replace($regex_patterns, $regex_replace, $nodeText));
     }
 
+    /**
+     * return if this role GIVE a permission
+     * @param string $sPermission 
+     * @param string $sCRUD 
+     */
     public function getHasPermission($sPermission, $sCRUD)
```
Shnoulle, line 299: did this function must return if current user have permission ?

olleharstedt: It sounds like it should return boolean value...?

application/models/Permission.php
Shnoulle: Oups …

application/models/Permissiontemplates.php
Shnoulle: yes … :) i update phpDoc

olleharstedt: :+1:

application/models/Permission.php
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/1755/commits/5838c394430bcc6093f21eae98eb204c9aad5c1d#diff-10077a3f8791500882779e35a827caf5e6578d5140625094d224de5e3c166780R527

application/models/Permission.php
```diff
@@ -566,16 +566,15 @@ public function hasPermission($iEntityID, $sEntityName, $sPermission, $sCRUD = '
         }
         $sCRUD = $sCRUD . '_p';
 
-        /* Always return false for guests */
-        // TODO: should not be necessary
+        /* Be sure to have an user id */
         $iUserID = $this->getUserId($iUserID);
-        if (!$iUserID && $iUserID !== 0) {
+        /* Always return false for guests */
+        if (!$iUserID) {
```
olleharstedt, line 572: User id === 0 or === null?

olleharstedt (changes requested):

application/models/Permission.php
Shnoulle: empty … === 0 or is null or === "" empmty mean empty :)

See https://github.com/LimeSurvey/LimeSurvey/pull/1755/files#diff-10077a3f8791500882779e35a827caf5e6578d5140625094d224de5e3c166780R732

In 3.X : we **never** send 0 to the function …

olleharstedt: Writing "empty($userId)" is more clear than "!$userId", I think. Or?

application/models/Permissiontemplates.php
```diff
@@ -291,9 +291,16 @@ public function removeCdataFormat($node)
         return trim(preg_replace($regex_patterns, $regex_replace, $nodeText));
     }
 
+    /**
+     * return if this role GIVE a permission
```
olleharstedt, line 295: Return true?

application/models/Permissiontemplates.php
```diff
@@ -291,9 +291,16 @@ public function removeCdataFormat($node)
...
+     * Used in only in compileExportXML
```
olleharstedt, line 296: One "in" too much.

application/models/Permission.php
Shnoulle: K

application/models/Permissiontemplates.php
Shnoulle: `* @return boolean` 2 lines after …

application/models/Permission.php
```diff
@@ -611,7 +610,7 @@ public function hasPermission($iEntityID, $sEntityName, $sPermission, $sCRUD = '
             if (safecount($aRoles) > 0) {
                 $allowed = false;
                 foreach ($aRoles as $role) {
-                    $allowed = $allowed || $this->hasRolePermission($role['ptid'], $sPermission, substr($sCRUD, 0, -2));
+                    $allowed = $allowed || $this->RoleHasPermission($role['ptid'], $sPermission, substr($sCRUD, 0, -2));
```
olleharstedt, line 613: roleHasPermission - R -> r

Shnoulle:
```
$ grep -ri roleHasPermission application/*
models/Permission.php:                    $allowed = $allowed || $this->roleHasPermission($role['ptid'], $sPermission, substr($sCRUD, 0, -2));
models/Permission.php:    public function roleHasPermission($iRoleId, $sPermission, $sCRUD = 'read')
models/Permissiontemplates.php:        return Permission::model()->roleHasPermission($this->ptid, $sPermission, $sCRUD);
views/admin/permissiontemplates/partials/_permissions.php:                                            <?php if(Permission::model()->roleHasPermission($oModel->ptid, $sPermissionKey, $sCRUDKey)):?>

```

olleharstedt:
Thanks, Denis!

PR #1758, opened 2021-02-12
Fixed issue #17040: Question editor buttons dont appear and the button general settings "pull right "

gabrieljenik:
Mapping old values for default html editor to new values


Shnoulle:
This is only for settings in 4.X … 

My opinion : don't care , leave it by default ;)

gabrieljenik:
New PR: https://github.com/LimeSurvey/LimeSurvey/pull/1797

PR #1762, opened 2021-02-12
Fixed issue #16979: Email continue not shown clearly after activation

gabrieljenik:

PR #1763, opened 2021-02-12
Fixed issue #17090: Email continue not shown clearly after activation

gabrieljenik:


application/views/admin/token/emailwarning.php
```diff
@@ -72,6 +77,16 @@
         $('#tokenSendCounter').text(window.countdownTimerTokenSend);
     },1000);
 
+    $('#sendTokenInvitationsNow').on('click', function(evt){
+        clearInterval(intervaltoRenew);
```
Shnoulle, line 81: Didn't test, but i don't like $.css, you can not remive a class + add a class ?

For example if you have a class 'animate-red' addClass('animate-red'); with css

```
.animate-red{color:red;}
.animate-red{animation-duration:1s;}
```

when you set animate-red class : the animation happen.

gabrieljenik: > Didn't test, but i don't like $.css, you can not remove a class + add a class ?

It was already being handled like this before.
I wouldn't update it right now, at least not in this bug.
I know it can be enhanced, I am not sure right now is the best time.

olleharstedt:
@Shnoulle Did you test this already?

Shnoulle:
I think yes , and i think i added screencast ?

Shnoulle:
Ah for 3X : https://github.com/LimeSurvey/LimeSurvey/pull/1762

olleharstedt:
Good enough. ^^ Thanks.

PR #1764, opened 2021-02-12
Fixed issue #17024: Missing string to translate

gabrieljenik:
Porting #16729 to LS4

### DIFF
--- a/application/helpers/twig_translation_helper.php
+++ b/application/helpers/twig_translation_helper.php
@@ -1,0 +1,27 @@
+<?php
+/*
+* LimeSurvey
+* Copyright (C) 2007-2020 The LimeSurvey Project Team / Carsten Schmitz
+* All rights reserved.
+* License: GNU/GPL License v2 or later, see LICENSE.php
+* LimeSurvey is free software. This version may have been modified pursuant
+* to the GNU General Public License, and as distributed it includes or
+* is derivative of works licensed under the GNU General Public License or
+* other free or open source software licenses.
+* See COPYRIGHT.php for copyright notices and details.
+*/
+
+
+/**
+ * Dummy helper intended to facilitate "twig only" strings to be picked by Translations Bot.
+ * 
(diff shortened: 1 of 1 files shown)


olleharstedt:
We didn't already have one of these dummy files...?

Shnoulle:
I really think it's better to add such line near the place it was used, just before the twig call :)

gabrieljenik:
> I really think it's better to add such line near the place it was used, just before the twig call :)

Where? Sorry I don;tunderstand.

Shnoulle:
Here : the string are related to after survey complete ?

Then : aroyund here : https://github.com/LimeSurvey/LimeSurvey/blob/da4b348a1ef832e7b9eaec8c7687ecd6d5480db6/application/helpers/SurveyRuntimeHelper.php#L1335 

```
/* @var string[] language string used in  twig file
$language = array(
     "Your survey responses have not been recorded. This survey is not yet active." => gT("Your survey responses have not been recorded. This survey is not yet active."), // themes/survey/vanilla/views/subviews/content/submit_preview.twig:23  (2020-11-16)
);
$sHtml = Yii::app()->twigRenderer->renderTemplateFromFile("layout_global.twig", array('oSurvey' => $oSurvey, 'aSurveyInfo' => $this->aSurveyInfo, 'language'=>$language), true);
```

gabrieljenik:
Hmmm not sure...
Having them all budled up together seems easier to handle.
Why putting it there? What use case?

Shnoulle:
> Why putting it there? What use case?

For translator : https://translate.limesurvey.org/projects/limesurvey-3/fr/default/ really more easy …

Capture d’écran du 2021-02-17 17-59-50

gabrieljenik:
I don't believe this is the right choice for this kind of messages.
A string that is used in a twig may be used on multiple twigs, but I guess it will be only inserted once on the php files. (Only for having the translator grep to find them).

So the references part may actually turn confusing.

Don't see such a clear benefit.
Actually I see more work in placing them on specific places on the php files, than on a generl file.

PR #1766, opened 2021-02-17
[WIP] Dev: quick test of CI with debug and assets

Shnoulle:
Dev: Just quick test currently

### DIFF
--- a/application/config/config-sample-mysql.php
+++ b/application/config/config-sample-mysql.php
@@ -57,10 +57,10 @@ return array(
         // then please check your error-logs - either in your hosting provider admin panel or in some /logs directory
         // on your webspace.
         // LimeSurvey developers: Set this to 2 to additionally display STRICT PHP error messages and put MySQL in STRICT mode and get full access to standard themes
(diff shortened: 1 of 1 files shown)


olleharstedt:
I saw that the emcache is always disabled when debug is on. Might break.

olleharstedt:
Yeah, if you check the raw Travis log, you can see "emcache is not set".

Shnoulle:
No time to check it currently, just want to see what happen quickly ;)

olleharstedt:
So we can close, or...?

Shnoulle:
[WIP] + draft is OK ?

Else : close it ;)

olleharstedt:
Yep, draft is OK.

Shnoulle:
Maybe try with 

1. `define('YII_DEBUG', true);`
2. `error_reporting(E_ALL);`
3. `App()->setConfig('debug', false);`

Do you think it can be interesting ?

olleharstedt:
I'd go for nr 2 first.

PR #1767, opened 2021-02-17
Fixed issue #16455: Public statistics for Ranking question not shown

gabrieljenik:
Fixed stats for ranking questions.
Also reviewed as requested and fixed the following:
- Array (Numbers): Error 500: query parameter ":scale:id" instead of ":scale_id" on 'createSGQA'
- Array (Numbers): Error 500: wrong fieldname created on 'createSGQA' (X-Scale subquestion not taken into account)
- List with Comments + Pie Chart: Error 500, "Call to a member function gT() on string": $sLanguageCode->gT("Comments")
- Ranking: 'createSGQA' returns null because it looks for subquestions instead of answer options
- Ranking: 'buildOutputList' doesn't implement l10n
- Multiple choice (M and P): 'buildOutputList' doesn't implement l10n

Dev:


Shnoulle:
A big big list here !

olleharstedt:
Great work!

PR #1768, opened 2021-02-17
Fixed issue #15539: queXML PDF shows JS script

gabrieljenik:
Using htmlpurifier instead of strip tags. Strip tags kept the tags internal content.


olleharstedt:
Thank you, Gabriel!

PR #1769, opened 2021-02-18
New Feature #13853: Allow test & basic feedback of the email functionality.

gabrieljenik:
Allow the user to tet the email settings that are set on the global configuration->email settings tab


application/controllers/admin/globalsettings.php
```diff
@@ -480,6 +480,65 @@ public function surveysettingmenues()
...
+        $aData = [
+            'testEmail' => $user->email,
+            'siteadminemail' => Yii::app()->getConfig("siteadminemail"),
+            'siteadminname' => Yii::app()->getConfig("siteadminname"),
+            'emailmethod' => Yii::app()->getConfig("emailmethod"),
+            'emailsmtphost' => Yii::app()->getConfig("emailsmtphost"),
+            'emailsmtpuser' => Yii::app()->getConfig("emailsmtpuser"),
+            'emailsmtppassword' => 'somepassword',
+            'emailsmtpssl' => Yii::app()->getConfig("emailsmtpssl"),
+            'emailsmtpdebug' => Yii::app()->getConfig("emailsmtpdebug"),
```
Shnoulle, line 537: Why not force emailsmtpdebug ? Or by a checkbox in the form ?

application/controllers/admin/globalsettings.php
```diff
@@ -480,6 +480,65 @@ public function surveysettingmenues()
...
+        $body   = array();
+        $body[] = sprintf(gT('This is a test email from %s'), $sSiteName);
+        $body   = implode("\n", $body);
+
+        $this->_sendEmailAndShowResult($body, $sSubject, $sTo, $sFrom);
+    }
+
+    private function _sendEmailAndShowResult($body, $sSubject, $sTo, $sFrom)
+    {
+        $mailer = new \LimeMailer();
```
Shnoulle, line 502: Better to have https://github.com/LimeSurvey/LimeSurvey/blob/91375a58e22334af287e652885599142954a676f/application/core/LimeMailer.php#L49

application/controllers/admin/globalsettings.php
```diff
@@ -480,6 +480,65 @@ public function surveysettingmenues()
...
+            $content = gT('Email sent successfully');
+        } else {
+            $content = gT('Email sending failure.');
+        }
+
+        $data = [];
+        $data['message'] = $content;
+        $data['success'] = $success;
+        $data['maildebug'] = $maildebug;
+
```
Shnoulle, line 521: Wbhy not use : https://github.com/LimeSurvey/LimeSurvey/blob/91375a58e22334af287e652885599142954a676f/application/core/LimeMailer.php#L460-L467 ?

application/views/admin/globalsettings/_emailTestConfirmation.php
```diff
@@ -0,0 +1,38 @@
...
+    </div>
+</div>
+<div class="modal-footer">
+    <a class="btn btn-primary btn-ok" href="<?php echo \Yii::app()->createUrl('admin/globalsettings', array("sa"=>"sendTestEmail"));?>"><span class='fa fa-check'></span>&nbsp;<?php eT("Send Email"); ?></a>
+    <button type="button" class="btn btn-danger" data-dismiss="modal">
+        <?php eT("Close"); ?>
+    </button>
+</div>
+
+<script>
```
Shnoulle, line 18: I don't understand usage of this script ?

gabrieljenik: This scripts identfies if any setting has been changed and not saved. Then it alerts the user about that

application/controllers/admin/globalsettings.php
gabrieljenik: You mean as to show erro details?

application/controllers/admin/globalsettings.php
gabrieljenik: What do you mean?

application/controllers/admin/globalsettings.php
gabrieljenik: The general setting is being used. Why forcing would be better? or why checbox?

application/controllers/admin/globalsettings.php
Shnoulle: This return the error from phpMailer like "can not connect" etc  … the cleanest error

application/controllers/admin/globalsettings.php
Shnoulle: You don't set it, then it stay default … https://manual.limesurvey.org/BeforeEmail

application/controllers/admin/globalsettings.php
Shnoulle: Because: if you do a test : you want more information than in real.

If user set `emailsmtpdebug` to 0 : no debug return.

application/views/admin/globalsettings/_emailTestConfirmation.php
Shnoulle: Oh, it's on **current** global setting … no way to add a dedicated page ? In my opinion : it's really the best way.

gabrieljenik: The idea is to test the settings. How do you picture it? This is already done. The idea was to port it and adapt it. Not re do it.

Shnoulle: > The idea was to port it and adapt it

? It don't exist in 3.X , then it's a new feature , no ?

gabrieljenik: It is a port from a PR already made

Shnoulle:
PS : the hardest part with such system i find what happen when there are error … and need different error : smtp, no mail function … etc …

application/controllers/admin/globalsettings.php
gabrieljenik: Updated

application/controllers/admin/globalsettings.php
gabrieljenik: Updated

Shnoulle: :+1:

olleharstedt:
Travis needs to be checked, and there should be functional test since it's a new feature.

gabrieljenik:
> Travis needs to be checked, and there should be functional test since it's a new feature.

Will email be setup on travis?
How do you picture the test like?

olleharstedt:
> > Travis needs to be checked, and there should be functional test since it's a new feature.
> 
> Will email be setup on travis?
> How do you picture the test like?

Urgh.

Travis says installation step is broken? Can you check it?

For functional test, good question. Something like this? https://michaelthessel.com/mock-mail-setup-for-travis-ci/

gabrieljenik:
> For functional test, good question. Something like this? https://michaelthessel.com/mock-mail-setup-for-travis-ci/

This will send emails or will just mock it?

gabrieljenik:
> Travis says installation step is broken? Can you check it?

I have updated from develop. Let's see if travis turns green.
Installation issue doens;t seem to be from this branch

olleharstedt:
Only one build broke instance broke randomly, so it's OK>

PR #1770, opened 2021-02-18
Fixed issue #16867: Randomization happen in survey logic file and in export

gabrieljenik:


application/controllers/admin/export.php
```diff
@@ -156,6 +155,9 @@ public function exportresults()
         $surveybaselang = $survey->language;
         $exportoutput = "";
 
+        // Avoid randomization of the fieldmap
+        unset($_SESSION['survey_'.$iSurveyID]['fieldmap-'.$iSurveyID.'-randMaster']);
```
Shnoulle, line 159: I really think we need a better way than unset a session var …

Shnoulle: For example check a param or some LimeExpressionManager::State somewher ( sPreviewMode ? for example : we use it for logic for example : logic/export ?)
here
https://github.com/LimeSurvey/LimeSurvey/blob/91375a58e22334af287e652885599142954a676f/application/helpers/common_helper.php#L1897

gabrieljenik: I didn't want to update the createFieldMap function.
Still, could be possible.

Maybe add a new optional param?

Shnoulle: I don't know, maybe you're right.

Why not just checking with `LimeExpressionManager::SetDirtyFlag();` or `killSurveySession($surveyid);` ?

Because : it work in 2.6lts without new params and without hacking  `$_SESSION` directly.

gabrieljenik: That's a good idea as well. Will try it

Shnoulle:
In 2.6lts : 
https://github.com/SondagesPro/LimeSurvey-SondagesPro/blob/bf34293e9538dba87e4ba9e7968190245c5714ed/application/views/admin/expressions/test/survey_logic_file.php#L105

`LimeExpressionManager::SetDirtyFlag();` seems to do the trick ?

gabrieljenik:
We tried with 2.06_sp_1.6.0 and 2.67.3+170728. The issue was there as well.
SetDirtyFlag is not doing the trick.
killSurveySession does it. It is a nicer approach, althugh it is an overkill.
We will do it that way. Uploading the patch in couple of minutes.

olleharstedt:
@Shnoulle You want to test?

Shnoulle:
> We tried with 2.06_sp_1.6.0 and 2.67.3+170728. The issue was there as well.
 
Oh shit ! i really think we have issue and we fixed it … :/ sorry

Shnoulle (approved):
Working , i report another issue , maybe we can fix at same time or just after.

Shnoulle:
[17142: Bad order of VV and database column with randomized question](https://bugs.limesurvey.org/view.php?id=17142)

PR #1773, opened 2021-02-18
Fixed issue #17018: No simple selector for Question type

gabrieljenik:


Shnoulle:
Seems there are issue with Question Theme

Way to reproduce : tested with https://github.com/tpartner/LimeSurvey-Card-Sorting-4.x and https://gitlab.com/SondagesPro/QuestionTheme/EmailInput

1. Before update : question have EmailInput for type
2. Update
3. Edit same question : show short text only
4. Set to EmailInput : save and close : broke (screenshot)
5. Another question : set to cardSort 
6. Save and close : same broke screenshot without debug

Capture d’écran du 2021-02-19 09-08-05
Capture d’écran du 2021-02-19 09-12-36

olleharstedt:
Thanks, everybody!

PR #1780, opened 2021-02-22
Fixed issue #17113: Randomization happen in survey logic file and in export

gabrieljenik:
Killing survey session before running logic file and export


Shnoulle:
Think it's better : it's not really overkill : you don't export each minute (it's related to user session)

olleharstedt:
@Shnoulle Can you test this, maybe?

Shnoulle (approved):
Same than other one :)

PR #1781, opened 2021-02-25
Fixed issue #16449: Error with CLI password reset

gabrieljenik:
Save the model validating only the password, because there may be issues with other attributes (like an invalid value for some setting), which the user cannot fix because he doesn't have access.

### DIFF
--- a/application/commands/ResetPasswordCommand.php
+++ b/application/commands/ResetPasswordCommand.php
@@ -21,8 +21,11 @@ class ResetPasswordCommand extends CConsoleCommand
         if (isset($sArgument) && isset($sArgument[0]) && isset($sArgument[1])) {
             $oUser = User::findByUsername($sArgument[0]);
             if ($oUser) {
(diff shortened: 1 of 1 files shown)


application/commands/ResetPasswordCommand.php
```diff
@@ -21,8 +21,11 @@ public function run($sArgument)
         if (isset($sArgument) && isset($sArgument[0]) && isset($sArgument[1])) {
             $oUser = User::findByUsername($sArgument[0]);
             if ($oUser) {
+                Yii::import('application.helpers.common_helper', true);
```
Shnoulle, line 24: Needed ?

Shnoulle: OK , ` Error: Call to undefined function gT() in /home/usabilit/limesurvey/application/models/User.php:107`

olleharstedt:
Pull Travis config again.

olleharstedt:
Never mind, was trivial. Thanks!

PR #1783, opened 2021-02-25
Fixed issue #16941: Too big file to upload, no error message

gabrieljenik:
Porting patches from #16733

### DIFF
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -1257,7 +1257,8 @@ class SurveyAdministrationController extends LSBaseController
         // Check file size and render JSON on error.
         // This is done before checking the survey permissions because, if the max POST size was exceeded,
         // there is no Survey ID to check for permissions, so the error could be misleading.
-        LSUploadHelper::checkUploadedFileSizeAndRenderJson('file', $debug);
+        $uploadHelper = new LSUploadHelper();
+        $uploadHelper->checkUploadedFileSizeAndRenderJson('file', $debug);
 
         $checkImage = LSYii_ImageValidator::validateImage($_FILES["file"]);
         if ($checkImage['check'] === false) {
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -1,6 +1,7 @@
 <?php
 
 use LimeSurvey\Models\Services\FilterImportedResources;
+use LimeSurvey\Models\Services\UploadHelper;
 
 class SurveyAdministrationController extends LSBaseController
 {
@@ -1257,7 +1258,7 @@ class SurveyAdministrationController extends LSBaseController
         // Check file size and render JSON on error.
         // This is done before checking the survey permissions because, if the max POST size was exceeded,
         // there is no Survey ID to check for permissions, so the error could be misleading.
-        $uploadHelper = new LSUploadHelper();
+        $uploadHelper = new UploadHelper();
         $uploadHelper->checkUploadedFileSizeAndRenderJson('file', $debug);
 
         $checkImage = LSYii_ImageValidator::validateImage($_FILES["file"]);
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -1253,27 +1253,12 @@ class SurveyAdministrationController extends LSBaseController
             );
         }
         $debug[] = $_FILES;
-        if (empty($_FILES)) {
-            $uploadresult = gT("No file was uploaded.");
-            return $this->renderPartial(
-                '/admin/super/_renderJson',
-                array('data' => ['success' => $success, 'message' => $uploadresult, 'debug' => $debug]),
-                false,
-                false
-            );
-        }
-        if ($_FILES['file']['error'] == 1 || $_FILES['file']['error'] == 2) {
-            $uploadresult = sprintf(
-                gT("Sorry, this file is too large. Only files up to %01.2f MB are allowed."),
-                getMaximumFileUploadSize() / 1024 / 1024
-            );
-            return $this->renderPartial(
-                '/admin/super/_renderJson',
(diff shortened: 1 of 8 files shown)


application/core/LSUploadHelper.php
```diff
@@ -0,0 +1,99 @@
...
+{
+    /**
+     * Check uploaded file size
+     * 
+     * @param string $sFileName the name of the posted file
+     * @param mixed $customMaxSize maximum file upload size
+     * 
+     * @throws Exception if the file is too large or no file is found.
+     */
+    public static function checkUploadedFileSize($sFileName, $customMaxSize = null)
```
olleharstedt, line 25: No static functions, please! They cannot be mocked easily.

olleharstedt (changes requested):

olleharstedt:
If this is a bug fix, shouldn't it be to master?

application/core/LSUploadHelper.php
gabrieljenik: OK, will take into account next time.
Can we just leave it like this for now?

gabrieljenik:
> If this is a bug fix, shouldn't it be to master?

Well, I thought to push it to dev as to have this on the system longer before merging to master.
As to get more testing.

application/core/LSUploadHelper.php
olleharstedt: No? Don't add technical debt in new code. :cry:

olleharstedt:
> > If this is a bug fix, shouldn't it be to master?
> 
> Well, I thought to push it to dev as to have this on the system longer before merging to master.
> As to get more testing.

Ah ok, so it's not super urgent, then?

gabrieljenik:
> Ah ok, so it's not super urgent, then?

On one side, it has been like this for a while. Also it is LS4. And at last, is about error handling on file upload on a specific situation.

application/core/LSUploadHelper.php
gabrieljenik: Done! Debt paid :)

application/controllers/SurveyAdministrationController.php
```diff
@@ -1253,27 +1253,13 @@ public function actionUploadimagefile()
...
-                '/admin/super/_renderJson',
-                array('data' => ['success' => $success, 'message' => $uploadresult, 'debug' => $debug]),
-                false,
-                false
-            );
-        }
+
+        // Check file size and render JSON on error.
+        // This is done before checking the survey permissions because, if the max POST size was exceeded,
+        // there is no Survey ID to check for permissions, so the error could be misleading.
```
olleharstedt, line 1259: <3 Comments!

application/controllers/admin/themes.php
```diff
@@ -425,6 +427,20 @@ public function uploadfile()
             $templatename           = returnGlobal('templatename');
             $oEditedTemplate        = Template::getInstance($templatename);
             $screenname             = returnGlobal('screenname');
+
+            // Check file size and redirect on error
+            $uploadHelper = new LSUploadHelper();
+            $uploadHelper->checkUploadedFileSizeAndRedirect(
+                'upload_file', 
+                array(
```
olleharstedt, line 435: For the future, feel free to use the `[]` shorthands instead of `array()` when creating arrays. Looks a bit more modern, I think. ^^

application/core/LSUploadHelper.php
```diff
@@ -0,0 +1,99 @@
...
+ * License: GNU/GPL License v2 or later, see LICENSE.php
+ * LimeSurvey is free software. This version may have been modified pursuant
+ * to the GNU General Public License, and as distributed it includes or
+ * is derivative of works licensed under the GNU General Public License or
+ * other free or open source software licenses.
+ * See COPYRIGHT.php for copyright notices and details.
+ * 
+ */
+
+class LSUploadHelper
```
olleharstedt, line 15: Can we move this to service folder instead? The service folder has additional checks for code quality, would be nice if new code could use those checks.

olleharstedt (changes requested):

application/core/LSUploadHelper.php
olleharstedt: Also, no need to prefix with "LS" since we have a LimeSurvey namespace in the service folder.

gabrieljenik: What about now?

c-schmitz: Travis failed. @gabrieljenik can you check, pelase?

gabrieljenik: I added some space fixes. I think it was that. Let's see...

olleharstedt: You can run all Travis syntax checks locally, with `composer install; composer test`.

olleharstedt: Latest error is namespace error in docblock: `Exception` vs `\Exception`, _or_ `use Exception;` at top of file.

gabrieljenik:
ping @olleharstedt

olleharstedt:
Uhm, upload helper should be in application/models/services/ :stuck_out_tongue_closed_eyes:

olleharstedt:
Sorry for the missing review.

olleharstedt:
About this: "checkUploadedFileSizeAndRenderJson", when you have an "and" in a function name, maybe it should rather be two different functions...?

gabrieljenik:
> About this: "checkUploadedFileSizeAndRenderJson", when you have an "and" in a function name, maybe it should rather be two different functions...? Then just compose the two functions. Or maybe function composition is a bit tricky when the error message is inside the exception...

We were doing code reorganization. Inside the function, you can see it is actually a wrapper for composing 2 functions.
Then we use that on different pieces of the code for the code reorganization.

I agree with you. Not the best name, but it was of great help for the reorg.

olleharstedt:
Yep. For the future, it's better if it looks like this:

```php
$err = $this->checkUploadFileSize();
if ($err) {
  $this->renderJson($err);
}
```

PR #1784, opened 2021-02-25
Fixed issue #15854: Cannot open pop up editor - subquestions (an extra page refresh needed)

gabrieljenik:
For new questions qid is random. Then when sent through the URL, common action executes _addPseudoParams which performs a validation. As the new question is still not valid, 403 is gotten. So now, for new questions. we don't send the param.

### DIFF
--- a/application/views/admin/survey/Question/subquestionsAndAnswers/_subquestion.twig
+++ b/application/views/admin/survey/Question/subquestionsAndAnswers/_subquestion.twig
@@ -108,7 +108,7 @@
             onkeypress=" if(event.keyCode==13) { if (event && event.preventDefault) event.preventDefault(); document.getElementById('save-button').click(); return false;}"
             />
             <span class="input-group-addon">
-                {{ getEditor("editanswer","answer_"~language~"_"~qid~"_"~scale_id, "["~gT("Subquestion:", "js")~"]("~language~")",surveyid,gid,qid,'editanswer') }}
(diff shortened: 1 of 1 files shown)


Shnoulle:
> For new questions qid is random.

But why ? qid must be empty for new question … no ?

gabrieljenik:
> But why ? qid must be empty for new question … no ?

I am not sure about that.
Maybe a mock id is needed for some stuff.

> how do you have a non number qid but not empty ?

a mock id is created and have some letters

> I think the issue to fix is this one qid MUST NOT be a non number (excpt empty)

Maybe that's a new issue... Need to solve the world piece by piece

Shnoulle:
Oups :) sorry : 3.X , not 4.X …

Shnoulle:
> Maybe that's a new issue... Need to solve the world piece by piece

Can not reproduce on 4.X, on 3.X : it's newXXX .

Sorry for my error here.

gabrieljenik:
> Sorry for my error here.

All good! Thanks for keeping an eye on stuf

olleharstedt:
> > Maybe that's a new issue... Need to solve the world piece by piece
> 
> Can not reproduce on 4.X, on 3.X : it's newXXX .
> 
> Sorry for my error here.

Denis, if you tested this PR, feel free to approve it so we don't have to test it again in the office.

olleharstedt (approved):

Shnoulle:
> Denis, if you tested this PR, feel free to approve it so we don't have to test it again in the office.

I didn't test the fix i test the issue ;)

PR #1792, opened 2021-03-02
Fixed issue #15239: Brute-force attack with tokens to enter survey is possible

gabrieljenik:
Adding the already existing FailedLoginAttempt control to the token validation process.

### DIFF
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -1246,25 +1246,31 @@ function finalizeRandomization($fieldmap)
 function testIfTokenIsValid(array $subscenarios, array $thissurvey, array $aEnterTokenData, $clienttoken)
 {
     $FlashError = '';
-    if (!$subscenarios['tokenValid']) {
-
-        //Check if there is a clienttoken set
-        if ((!isset($clienttoken) || $clienttoken == "")) {
-            if (isset($thissurvey) && $thissurvey['allowregister'] == "Y") {
-                $renderToken = 'register';
+    if (FailedLoginAttempt::model()->isLockedOut()) {
+        $FlashError = sprintf(gT('You have exceeded the number of maximum token validation attempts. Please wait %d minutes before trying again.'), App()->getConfig('timeOutTime') / 60);
+        $renderToken = 'main';
+    } else {
+        if (!$subscenarios['tokenValid']) {
+            //Check if there is a clienttoken set
+            if ((!isset($clienttoken) || $clienttoken == "")) {
+                if (isset($thissurvey) && $thissurvey['allowregister'] == "Y") {
+                    $renderToken = 'register';
+                } else {
+                    $renderToken = 'main';
+                }
             } else {
+                //token was wrong
+                $errorMsg    = gT("The token you have provided is either not valid, or has already been used.");
+                $FlashError .= $errorMsg;
                 $renderToken = 'main';
+                FailedLoginAttempt::model()->addAttempt();
             }
         } else {
-            //token was wrong
-            $errorMsg    = gT("The token you have provided is either not valid, or has already been used.");
-            $FlashError .= $errorMsg;
-            $renderToken = 'main';
+            $aEnterTokenData['visibleToken'] = $clienttoken;
+            $aEnterTokenData['token'] = $clienttoken;
+            $renderToken = 'correct';
+            FailedLoginAttempt::model()->deleteAttempts();
         }
-    } else {
-        $aEnterTokenData['visibleToken'] = $clienttoken;
-        $aEnterTokenData['token'] = $clienttoken;
-        $renderToken = 'correct';
     }
 
     return array($renderToken, $FlashError, $aEnterTokenData);


Shnoulle:
Same table for different purpose ??? Really ?

gabrieljenik:
> Same table for different purpose ??? Really ?

Well, it is not a whole different purpose. Is about FailedAttempts, the onlly difference is the login screen used.

Shnoulle:
there are an issue here : this delete admin login with one token connection …

Shnoulle:
> Well, it is not a whole different purpose. Is about FailedAttempts, the onlly difference is the login screen used.

No : you can have token without admin account …

Shnoulle:
I want to get superadmin account on limesurvey.example.org , this limesurvey.example.org have a register survey.

I get a valid token in one survey

Before this «fix» : must find proxyes to try password-list with admin.
After this fix : no need  proxyes , just loop 

1. get new password
2. Try POST connect with admin/newpassword
3. OK => end
4. Send a GET url to survey with my valid token
5. Git to 1

Shnoulle:
Another issue reported related to this commit https://bugs.limesurvey.org/view.php?id=17301

**Need** a different settings and DB for token …

olleharstedt:
Is client token not related to survey id? If participant takes multiple surveys on same machine.

olleharstedt:
Notified Gabriel about it.

olleharstedt:
```✖ 1887     $sClientToken = isset($_SESSION['survey_'.$surveyid]['token']) ? $_SESSION['survey_'.$surveyid]['token'] : "";                                                                                          
```

Shnoulle:
> Is client token not related to survey id? If participant takes multiple surveys on same machine.

Company connexion : https://forums.limesurvey.org/forum/installation-a-update-issues/124479-10-min-lock-out#215920 same IP for all.

olleharstedt:
Ah!

olleharstedt:
Possible to do MAC address instead?

gabrieljenik:
Maybe we could add a whitelist setting on the config?

Shnoulle:
> Same table for different purpose ??? Really ?

And : same config var for 2 different purpose …

All other fix are «quick patch» … 

The 2 ystem don't have the same objectives, don't have the same control (one with a GET param, other with an POST user/pass).

This is **clearly** not the same purpose.

gabrieljenik:
> Same table for different purpose ??? Really ?
> And : same config var for 2 different purpose …

What does thi has to do with the problem?
What would be different if settings and tables would be different?

Shnoulle:
Because : you can want to set 1 seconds for token but 10 minutes for user/pass

My opinion : best is

1. 1second and one try for token
2. 10 minutes and 5 try for user (check pass)

And why need different : because i's **totally unrelated** :  the access are different (more right with admin than with token) and the way to access is different too (again : one single param with GET against 2 related param with POST).


I really don't understand how i can give proof this is different … it seems so clear and obvious to me.

Shnoulle:
And more : here : you broke 2 system (at minima).
If you create a new system (deactivated by default) : you don't broke this 2 user system.

gabrieljenik:
>  the access are different (more right with admin than with token)

But we wanted to protect against DDOS attacks. There settings should be similar.
Still, by setting the tokens delay to 1s, the approach would be to disbale the DDOS protection on the tokens side. right?

Shnoulle:
>  Still, by setting the tokens delay to 1s, the approach would be to disbale the DDOS protection on the tokens side. right?

No : bots don't like to wait 1 seconds …

Token : Bot attack
User pass : Bot attack + Social know attack.

Shnoulle:
Clearly : how can you compare

- My username and the password i choose to get whole data access of the system
- A random 15 caracters to get only one line of one data.

gabrieljenik:
I understand the difference in data, but from a DDOS perspective, there is no difference.
I would add a whitelisting IPs to the config. That should solve it pretty quick

gabrieljenik:
@c-schmitz What do you think?

Shnoulle:
> I understand the difference in data, but from a DDOS perspective, there is no difference.
> I would add a whitelisting IPs to the config. That should solve it pretty quick

Look at forum : 

- before this commit : there are a restriction if a student search to get access of admin GUI, for example a teacher user/pass. Not the best but something was done (and can give alert to administrator : someone try to get admin access)
- You put a whiteliste IP : no restriction to try any user/pass (techer username + password)
=> You broke security.

gabrieljenik:
> You broke security.

Whitelist could be separate for token DDOS and for login DDOS

Shnoulle:
> Whitelist could be separate for token DDOS and for login DDOS

Then separate timeout + try and where it's saved … (new DB)

jelo-ls:
I recommend to take inspiration in a few WordPress plugins which handle brute-force and other login-protections.
There you find UX ideas. Ideas to handle the GDPR impact when storing IP addresses.
E.g. https://wordpress.org/plugins/limit-login-attempts-reloaded/

@olleharstedt MAC adresses are not transmitted via WAN, only LAN.

PR #1785, opened 2021-02-25
Fixed issue #15819: Missing localization for group_name in add_group, set_group_properties, import_group

gabrieljenik:
'import_group' was working fine.
'add_group' was updated as to save to the QuestionGroupL10n model.

'set_group_properties' was updated as to be able to receive an array on 'questiongroupl10ns' param.
The method is backward compatible. If an 'questiongroupl10ns' is not given, then it is composed from the old name, description and language parameters.

Adding test for 'add_group' y 'set_group_properties'.

PR #1793, opened 2021-03-02
Fixed issue #15819: Missing localization for group_name in add_group, set_group_properties, import_group

gabrieljenik:
'import_group' was working fine.
'add_group' was updated as to save to the QuestionGroupL10n model.

'set_group_properties' was updated as to be able to receive an array on 'questiongroupl10ns' param.
The method is backward compatible. If an 'questiongroupl10ns' is not given, then it is composed from the old name, description and language parameters.

Adding test for 'add_group' y 'set_group_properties'.

PRevios PR (https://github.com/LimeSurvey/LimeSurvey/pull/1785) closed. This replaces it.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1005,11 +1005,18 @@ class remotecontrol_handle
 
                 $oGroup = new QuestionGroup();
                 $oGroup->sid = $iSurveyID;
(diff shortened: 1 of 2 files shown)


olleharstedt:
This can be merged? Dunno who would test it otherwise.

gabrieljenik:
From my point of view, yes.
It has the unit testing. It is API. It is develop.
Risk is low

olleharstedt:
Yep, agreed.

PR #1795, opened 2021-03-05
New PHPMailer factory

gabrieljenik:
New factory for getting PHPMailer instances.
Usage modes:
- Default: Factory will create new instances
- Reuse: Factory will reuse previous created instance

For Reuse Mode, need to add the following before SendEmailMessage() call
 ```
 Yii::import('application.helpers.mailHelper');
 mailHelper::setModeForNext(mailHelper::PREVIOUS_INSTANCE_MODE, ['smtpKeepAlive' => true]);
 ```

### DIFF
--- a/application/controllers/admin/tokens.php
+++ b/application/controllers/admin/tokens.php
@@ -1489,6 +1489,8 @@ class tokens extends Survey_Common_Action
                             $maildebug = (string)$event->get('error', $maildebug);
                             $success = $event->get('error') == null;
                         } else {
+                            Yii::import('application.helpers.mailHelper');
+                            mailHelper::setModeForNext(mailHelper::PREVIOUS_INSTANCE_MODE, ['smtpKeepAlive' => true]);
                             $success = SendEmailMessage($modmessage, $modsubject, $to, $from, Yii::app()->getConfig("sitename"), $bHtml, $bounce, $aRelevantAttachments, $customheaders);
                         }
 
--- a/application/helpers/admin/token_helper.php
+++ b/application/helpers/admin/token_helper.php
@@ -153,6 +153,8 @@ function emailTokens($iSurveyID, $aResultTokens, $sType)
                                                 'error'=>'Token not valid anymore');
 
         } else {
+            Yii::import('application.helpers.mailHelper');
+            mailHelper::setModeForNext(mailHelper::PREVIOUS_INSTANCE_MODE, ['smtpKeepAlive' => true]);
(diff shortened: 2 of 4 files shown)


olleharstedt:
Note that Denis already made a mailer class in LS4.

Shnoulle:
Yes, i totally for the concept … 

I use a https://github.com/LimeSurvey/LimeSurvey/blob/4d16745af9a202ff3fd701f4c7ed1697d07337bf/application/core/LimeMailer.php#L227

I do it in develop/4 because: no new feature in 3.X …

c-schmitz:
Yes, this is something developed for customers in LS3 that need it now, and cannot wait for LS4

c-schmitz:
@Shnoulle However I saw that you did not set SMTP Keepalive in your LimeMailer class. That would still needed to be fixed to spee up mass mailing on SMTP  -  see https://github.com/PHPMailer/PHPMailer/blob/master/examples/mailing_list.phps

Shnoulle:
And ? 

We don't have it before, it's a good idea to add it now .

Shnoulle:
Else : maybe move LimeMailer to 3.X ?

Can i add the 2 Expression Manager events in 3.X at same time ?

olleharstedt:
PHP syntax error on PHP 5.6:

> Parse error: syntax error, unexpected 'const' (T_CONST), expecting variable (T_VARIABLE) in application/helpers/mailHelper.php on line 23

olleharstedt:
```
     public const NEW_INSTANCE_MODE = 0;
     ```
     
     Guess PHP 5.6 doesn't support accessor on const?

**Comment by gabrieljenik** [comment]
> ```
>      public const NEW_INSTANCE_MODE = 0;
> ```
> 
> Guess PHP 5.6 doesn't support accessor on const?

Class constant visibility modifiers, as of PHP 7.1.0

Fixed

**Comment by c-schmitz** [comment]
> 
> 
> And ?
> 
> We don't have it before, it's a good idea to add it now .

Sure, it should be a one-liner. If you want I can add it.

**Comment by Shnoulle** [comment]
> > And ?
> > We don't have it before, it's a good idea to add it now .
> 
> Sure, it should be a one-liner. If you want I can add it.

I think it don't need a option to disable this ? Justt here ? 
https://github.com/LimeSurvey/LimeSurvey/blob/4d16745af9a202ff3fd701f4c7ed1697d07337bf/application/core/LimeMailer.php#L164

You see any reason to disable ? 

Maybe add a 

```
public function close() {
    if($this->isSMTP()) {
        $this->SmtpClose();
    }
}

```
And add it after all mail is sent ?

PR #1797, opened 2021-03-09
Fixed issue #17040: Question editor buttons dont appear and the button general settings "pull right "

gabrieljenik:
Mapping old values for default html editor to new values

### DIFF
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -3721,6 +3721,41 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
             $oTransaction->commit();
         }
 
+        if ($iOldDBVersion < 441) {
+            $oTransaction = $oDB->beginTransaction();
+            // Convert old html editor modes if present in global settings
+            $oDB->createCommand()->update(
+                '{{settings_global}}',
+                array(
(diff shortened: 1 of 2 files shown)


application/helpers/update/updatedb_helper.php
```diff
@@ -3721,6 +3721,41 @@ function ($v) {
...
+                "htmleditormode='wysiwyg'"
+            );
+            $oDB->createCommand()->update(
+                '{{users}}',
+                array(
+                    'htmleditormode' => 'none',
+                ),
+                "htmleditormode='source'"
+            );
+            $oDB->createCommand()->update('{{settings_global}}', array('stg_value' => 438), "stg_name='DBVersion'");
```
olleharstedt, line 3756: 438? :O

olleharstedt (changes requested):

application/helpers/update/updatedb_helper.php
gabrieljenik: Typo after moving the code for a new db version

olleharstedt: I've done this 100 times. When we refactor each update into its own class, dbversion will be part of base class and never touched again, making typos like this impossible.

PR #1798, opened 2021-03-10
Fixed issue #15141: Limesurvey uses an out-of-date version of bootstrap.min.js that has security vulnerabilities

gabrieljenik:
Travis test to check if these files are necessary


thedirtypanda:
Works fine. Thanks!

gabrieljenik:
Honestly, this was a test as to double check if those removed files are used or not. 
Not sure if it is the best to remove those files/

thedirtypanda:
Did you update these files too? I only saw that you commented them out. Because if they are just 'deactivated', and never be used anymore we could create a task, and remove them in LS3. If there are no complains we could apply those changes at LS4 too. What do you think @gabrieljenik?

Shnoulle:
You mean remove bootstrap extension ? I think it's a good idea, but really need test.

I think it was used a lot currently in admin views.

gabrieljenik:
I think we should revert this merge

Shnoulle:
Why ?

In my opinion : extensions/bootstrap need bootstrap.js 

Since we already loaded bootstrap.js for admin : we don't need to load this old one.

No ?

thedirtypanda:
@gabrieljenik why do you think, that we should remove the merge?

gabrieljenik:
@Shnoulle mentioned we might need more testing. That's why.
I am not aware of the testing it has been done.

Still, I would remove it first on ls4, then on ls3.

Shnoulle:
> @Shnoulle mentioned we might need more testing. That's why.
> I am not aware of the testing it has been done.
> 
> Still, I would remove it first on ls4, then on ls3.

I just want more testing remove all bootstrap extension, maybe used on grid or i don't know.

But since boostrap extension use bootsrap.js and this one was included by adminbasics package : i don't think we need a second one :)

[bootstrap extension](https://github.com/LimeSurvey/LimeSurvey/tree/master/application/extensions/bootstrap) != [bootstrap.js](https://github.com/LimeSurvey/LimeSurvey/tree/master/assets/packages/bootstrap)

PR #1799, opened 2021-03-10
Fixed bug #15919: Cannot import custom question theme based on an array question

gabrieljenik:
Scanning directories as to import more than one theme
Fetching question type from the path now takes into account types with subtypes, as array/10point

### DIFF
--- a/application/controllers/admin/themes.php
+++ b/application/controllers/admin/themes.php
@@ -379,10 +379,26 @@ class themes extends Survey_Common_Action
                         );
                         $this->getController()->redirect(array("themeOptions/index#questionthemes"));
                     }
-                    $sThemeDirectory = scandir($sPathToThemeDirectory)[2];
-                    // install the Question Theme
-                    $sQuestionThemeTitle = $questionTheme->importManifest($sPathToThemeDirectory . DIRECTORY_SEPARATOR . $sThemeDirectory);
-                    if (empty($sQuestionThemeTitle)) {
+                    // Question themes that apply to more than one question type, are technically different themes but can be distributed
+                    // in the same ZIP. So we must try to install all the available themes in the folder.
+                    $importedThemes = 0;
+                    $directory = new RecursiveDirectoryIterator($sPathToThemeDirectory);
+                    $iterator = new RecursiveIteratorIterator($directory);
+                    foreach ($iterator as $info) {
+                        if ($info->isFile() && $info->getBasename() == 'config.xml') {
+                            $questionConfigFilePath = dirname($info->getPathname());
+                            $sQuestionThemeTitle = null;
+                            try {
+                                $sQuestionThemeTitle = $questionTheme->importManifest($questionConfigFilePath);
+                            } catch (Exception $e) {
+                                // Should probably push the error to an array an display the contents on a results page
+                            }
+                            if (!empty($sQuestionThemeTitle)) {
+                                $importedThemes++;
+                            }
+                        }
+                    }
+                    if ($importedThemes == 0) {
                         rmdirr($destdir);
                         App()->setFlashMessage(
                             gT("An error occured while generating the Question theme"),
--- a/application/controllers/admin/themes.php
+++ b/application/controllers/admin/themes.php
@@ -368,7 +368,6 @@ class themes extends Survey_Common_Action
                     TemplateManifest::importManifest($sNewDirectoryName, ['extends' => $destdir]);
                 }
                 if ($themeType == 'question') {
-                    $questionTheme = new QuestionTheme();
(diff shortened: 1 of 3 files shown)


olleharstedt:
I can test this one.

application/controllers/admin/themes.php
```diff
@@ -379,10 +379,26 @@ protected function uploadTemplate()
...
+                    $directory = new RecursiveDirectoryIterator($sPathToThemeDirectory);
+                    $iterator = new RecursiveIteratorIterator($directory);
+                    foreach ($iterator as $info) {
+                        if ($info->isFile() && $info->getBasename() == 'config.xml') {
+                            $questionConfigFilePath = dirname($info->getPathname());
+                            $sQuestionThemeTitle = null;
+                            try {
+                                $sQuestionThemeTitle = $questionTheme->importManifest($questionConfigFilePath);
+                            } catch (Exception $e) {
+                                // Should probably push the error to an array an display the contents on a results page
```
olleharstedt, line 394: This won't cause inconsistencies in the installation?

olleharstedt:
This seems to not respect uploaddir config, in my test.

application/controllers/admin/themes.php
gabrieljenik: It was already like this on line 384.
https://github.com/LimeSurvey/LimeSurvey/pull/1799/files#diff-ac2f5aae9a7760b49ff438811a452a7c0dccf72e49ac2bc6b4c5a128582a8139L384

What kind of inconsistencies? How?

gabrieljenik:
> This seems to not respect uploaddir config, in my test.
> 
> ```
>  63         'uploaddir' => '/tmp/upload/',  
> ```
> 
> But it ended up in upload/ folder from web root instead.

Will review...

application/controllers/admin/themes.php
olleharstedt: Maybe I didn't read the code carefully enough, but if some files are copied and others are not?

olleharstedt:
> Will review. From a first glance, it doesn't seem to be something introduced by the fix. Still an issue.

Nah, can be a separate issue.

application/controllers/admin/themes.php
gabrieljenik: Actually, yes, we took the opportunity of the fix to allow multiple themes. 
If not, if there are more than 1 theme in the zip, which one should we import?

Yes, the part for showing results for the multiple import, is pending.
Wanted to keep the ticket slim.
We can do another ticket for that?

On the other side, if some import fails, they are left as uninstalled, so they can be picked up later.

Shnoulle: >  If not, if there are more than 1 theme in the zip, which one should we import?

Maybe we can have a "mailIinput" for short text + multiple text (OK) + **array of text** , this mean 3 different config here for "same functionality for different question type".

Is not allowed currently right ?

olleharstedt: We don't want a transaction here, then? Either everything works or it's aborted, to avoid any inconsistencies.

gabrieljenik: Previously, as this line, only the first directory on the zip file was added as theme.
https://github.com/LimeSurvey/LimeSurvey/pull/1799/files#diff-ac2f5aae9a7760b49ff438811a452a7c0dccf72e49ac2bc6b4c5a128582a8139L382

Now, as all directories are scanned, all themes are to be imported.

> Is not allowed currently right ?

So,  now should be allowed.

Shnoulle: > We don't want a transaction here, then? Either everything works or it's aborted, to avoid any inconsistencies.

Yes, seems best : if one error happen : show the error and delete whole files and data imported. 

**BUT** : it's not the issue in [15919](https://bugs.limesurvey.org/view.php?id=15919) 

Muts create another bug or feature request

Shnoulle: > So, now should be allowed.

Just need to find some time to test (still have some issue to report before)

gabrieljenik: > We don't want a transaction here, then? Either everything works or it's aborted, to avoid any inconsistencies.

I think transaction is overcomplicating stuff.
I would say, wrap up the multiple upproach and then add transactions, if you like.

Maybe I should just remove the multiple directory scanning.
Is not related to the ticket and is not letting the ticket move forward.

What do you think?

gabrieljenik: > Maybe I didn't read the code carefully enough, but if some files are copied and others are not?

We will show a toast message saying some of the themes couldn't be exported.
We will try to include some expcetion message as to guide the user (from the current empty `catch` statement).

That will be a wrap. After we can enhance stuff.

olleharstedt: It's what you like. I think the ability to update a theme is the most important next step here.

gabrieljenik: Updated it

PR #1800, opened 2021-03-14
Fixed issue #16903: Windows server : unable to save attachements

Shnoulle:
Fixed issue #16905: Windows server : good attachment throw JS error (and broke all JS action)
Dev: save only basename : JS don't care of separator
Dev: save only basename, find files in survey upload dir (always)

### DIFF
--- a/application/controllers/admin/emailtemplates.php
+++ b/application/controllers/admin/emailtemplates.php
@@ -99,38 +99,22 @@ class emailtemplates extends Survey_Common_Action
      */
     function update($iSurveyId)
     {
-        $sBaseUrl = Yii::app()->getBaseUrl();
-        $uploadUrl = Yii::app()->getConfig('uploadurl');
-        if (substr($uploadUrl, 0, strlen($sBaseUrl)) == $sBaseUrl) {
-            $uploadUrl = substr($uploadUrl, strlen($sBaseUrl));
-        }
-        $sBaseAbsoluteUrl = Yii::app()->getBaseUrl(true);
-        $sPublicUrl = Yii::app()->getConfig("publicurl");
-        $aPublicUrl = parse_url($sPublicUrl);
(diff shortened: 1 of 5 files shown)


olleharstedt:
Why is this fix so big? :cry:

application/controllers/admin/emailtemplates.php
```diff
@@ -99,38 +99,22 @@ function index($iSurveyId)
      */
     function update($iSurveyId)
     {
-        $sBaseUrl = Yii::app()->getBaseUrl();
-        $uploadUrl = Yii::app()->getConfig('uploadurl');
-        if (substr($uploadUrl, 0, strlen($sBaseUrl)) == $sBaseUrl) {
-            $uploadUrl = substr($uploadUrl, strlen($sBaseUrl));
-        }
-        $sBaseAbsoluteUrl = Yii::app()->getBaseUrl(true);
-        $sPublicUrl = Yii::app()->getConfig("publicurl");
```
olleharstedt, line 108: Remove use of publicurl?

Shnoulle:
> Why is this fix so big? cry

Because there are no helper or model system to get and fix attachements : 

The «only fix» was using `$baseName = pathinfo($attachment['url'], PATHINFO_BASENAME);` but it must be done each time we try to find the file , then each time we send a message … with attachment.

Shnoulle:
2 mores ! https://github.com/LimeSurvey/LimeSurvey/pull/1800/commits/9dcf41e2b7c649f65d90c01b7150e36de5b06cb9

I lost admin notification on 1st commit

Shnoulle:
Create a getter SurveyLanguageSetting->getAttachement return the arry is surely the better solution , but it's rewrite, not fix …

Shnoulle:
Oh,

I got aniother idea : manage : use only filename, but save with complete directory . Then less update.

And create a clean getter in 4X

PR #1803, opened 2021-03-17
Fixed issue #16904: Check attributes are escaped properly

gabrieljenik:
Answer options are escaped before render as to avoid broken html


Shnoulle:
unsure : https://twig.symfony.com/doc/2.x/filters/escape.html `|escape('html_attr')` ?

Strangely not escaped in 3.X ????

gabrieljenik:
> unsure

What do you mean?

Shnoulle:
> > unsure
> 
> What do you mean?

I just mean i'm not sure, but seems twig have a settings for escape :  `html_attr`: escapes a string for the HTML attribute context.

Value is an attribute , right ?

Shnoulle:
Argl … my bad ¿ you use  `html_attr` sorry again (me tired …)

c-schmitz (approved):

PR #1807, opened 2021-03-18
Fixed issue: wrong port and TLS description in bounce email/smtp settings

ViliusS:
Dev: As per imap_open() manual {hostname/ssl} could be actually both SSL or implicit TLS connection depending on PHP configuration/version, which almost always uses standard IMAPS/POP3S ports. And {hostname/tls} means that a client requires a server to allow STARTTLS which usually runs on the same port as plain text IMAP/POP3. Make sure we won't set wrong port here if an administrator didn't specify one.

Dev: The same goes for SMTP connection and PHPMailer library. If "tls" is specified the SMTPSecure option in the library initiates STARTTLS support not the implicit TLS connection, see https://github.com/PHPMailer/PHPMailer/blob/master/src/PHPMailer.php#L2024

Dev: While at it, inform administrators that not using SSL/TLS or StartTLS makes a connection very insecure.

Dev: Also, make values lowercase as most of their default options in the code are lowercase.

Backport of PR #1639 to LTS-3.x branch.

### DIFF
--- a/application/views/admin/global_settings/_bounce.php
+++ b/application/views/admin/global_settings/_bounce.php
@@ -59,7 +59,7 @@
                 'name' => 'bounceencryption',
                 'value'=> strtolower(getGlobalSetting('bounceencryption')),
                 'selectOptions'=>array(
-                "off"=>gT("Off (insecure)",'unescaped'),
+                "off"=>gT("Off (unsafe)",'unescaped'),
                 "ssl"=>"SSL/TLS",
                 "tls"=>"StartTLS"
                 )
--- a/application/views/admin/global_settings/_bounce.php
+++ b/application/views/admin/global_settings/_bounce.php
@@ -57,11 +57,11 @@
     <div class="">
         <?php $this->widget('yiiwheels.widgets.buttongroup.WhButtonGroup', array(
                 'name' => 'bounceencryption',
-                'value'=> getGlobalSetting('bounceencryption') ,
+                'value'=> strtolower(getGlobalSetting('bounceencryption')),
                 'selectOptions'=>array(
-                "off"=>gT("Off",'unescaped'),
-                "SSL"=>"SSL",
-                "TLS"=>"TLS"
+                "off"=>gT("Off (insecure)",'unescaped'),
+                "ssl"=>"SSL/TLS",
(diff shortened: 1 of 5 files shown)


application/views/admin/global_settings/_email.php
```diff
@@ -61,9 +61,9 @@
             'name' => 'emailsmtpssl',
             'value'=> getGlobalSetting('emailsmtpssl') ,
             'selectOptions'=>array(
-                ""=>gT("Off",'unescaped'),
-                "ssl"=>gT("SSL",'unescaped'),
-                "tls"=>gT("TLS",'unescaped')
+                ""=>gT("Off (insecure)",'unescaped'),
```
c-schmitz, line 64: "insecure" is wrong - should be 'unsafe' -same for other occurences.

application/views/admin/global_settings/_bounce.php
```diff
@@ -57,11 +57,11 @@
     <div class="">
         <?php $this->widget('yiiwheels.widgets.buttongroup.WhButtonGroup', array(
                 'name' => 'bounceencryption',
-                'value'=> getGlobalSetting('bounceencryption') ,
+                'value'=> strtolower(getGlobalSetting('bounceencryption')),
                 'selectOptions'=>array(
-                "off"=>gT("Off",'unescaped'),
-                "SSL"=>"SSL",
-                "TLS"=>"TLS"
+                "off"=>gT("Off (insecure)",'unescaped'),
```
c-schmitz, line 62: Same here

c-schmitz (changes requested):

application/views/admin/global_settings/_email.php
ViliusS: I understand that 'unsafe' could be more precise language wise, but term 'insecure' is now considered acceptable in IT security, see https://en.wikipedia.org/wiki/Transport_Layer_Security#Cipher

Also, this is inline how it looks in LS4. If you still think it is worth changing, I will make additional commit.

c-schmitz: I fixed it for LS4 already, please change it here, too.

ViliusS: Done.

PR #1809, opened 2021-03-22
Fixed issue #16462: Relavance Equation not in Sync with Condition Designed

gabrieljenik:
Fixing stuff on GeneralOptionWidget. Attributes were not being implemented in the view.


c-schmitz:
Patch creates a bug. You cannot navigate to a question anymore. See issue in Zoho projects,

gabrieljenik:
I see. Will do some initialzing just in case attributes are not defined. Thanks

PR #1812, opened 2021-03-22
Fixed issue #17190: Update Theme Permission Language

gabrieljenik:


application/views/admin/user/setusertemplates.php
```diff
@@ -1,4 +1,9 @@
 <div class="pagetitle h3"><?php printf(gT('Edit theme permissions for user %s'),"<em>".\CHtml::encode($oUser->users_name)."</em>");?></div>
+<div class="row">
+    <div class="col-lg-12">
+        <p><?php eT("If the user doesn't have view/read global permission, please select the themes he would be able to assign to surveys."); ?></p>
```
Shnoulle, line 4: _If the user doesn't have view/read global permission **on theme**_ or something like this ?

PR #1821, opened 2021-03-25
Fixed issue #17188: {GROUPNAME} and other variables not rendered

gabrieljenik:
Adding EM processing to group description

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -397,7 +397,6 @@ class SurveyRuntimeHelper
             $gid              = $gl['gid'];
             LimeExpressionManager::updateReplacementFields(array(
                 'GID' => $gid,
-                'GROUPNAME' => $gl['group_name'],
             ));
             $aGroup           = array();
             if ($this->sSurveyMode != 'survey') {
@@ -418,6 +417,9 @@ class SurveyRuntimeHelper
             }
 
             $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, null, 3, 1);
+            LimeExpressionManager::updateReplacementFields(array(
+                'GROUPNAME' => $gl['group_name'],
+            ));
             $aGroup['gseq']        = $_gseq;
             $showgroupinfo_global_ = getGlobalSetting('showgroupinfo');
             $aSurveyinfo           = getSurveyInfo($this->iSurveyid, App()->getLanguage());
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -418,7 +418,7 @@ class SurveyRuntimeHelper
 
             $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, null, 3, 1);
             LimeExpressionManager::updateReplacementFields(array(
-                'GROUPNAME' => $gl['group_name'],
+                'GROUPNAME' => $aGroup['name'],
             ));
             $aGroup['gseq']        = $_gseq;
             $showgroupinfo_global_ = getGlobalSetting('showgroupinfo');
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -417,7 +417,7 @@ class SurveyRuntimeHelper
                 $aGroup['class'] = ' ls-hidden';
             }
 
-            $aGroup['name']        = $gl['group_name'];
+            $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, null, 3, 1);
             $aGroup['gseq']        = $_gseq;
             $showgroupinfo_global_ = getGlobalSetting('showgroupinfo');
             $aSurveyinfo           = getSurveyInfo($this->iSurveyid, App()->getLanguage());
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -397,7 +397,7 @@ class SurveyRuntimeHelper
             $gid              = $gl['gid'];
             LimeExpressionManager::updateReplacementFields(array(
                 'GID' => $gid,
-                'GROUPNAME' => $this->groupname,
+                'GROUPNAME' => $gl['group_name'],
             ));
             $aGroup           = array();
             if ($this->sSurveyMode != 'survey') {
@@ -433,7 +433,7 @@ class SurveyRuntimeHelper
 
             $aGroup['showgroupinfo'] = $showgroupinfo_;
             $aGroup['showdescription']  = (!$this->previewquestion && trim($gl['description']) != "" && $showgroupdesc_);
-            $aGroup['description']      = $gl['description'];
+            $aGroup['description']      = LimeExpressionManager::ProcessString($gl['description'], null, null, 3, 1);
 
             // one entry per QID
             foreach ($qanda as $qa) {


application/helpers/SurveyRuntimeHelper.php
```diff
@@ -433,7 +433,7 @@ public function run($surveyid, $args)
 
             $aGroup['showgroupinfo'] = $showgroupinfo_;
             $aGroup['showdescription']  = (!$this->previewquestion && trim($gl['description']) != "" && $showgroupdesc_);
-            $aGroup['description']      = $gl['description'];
+            $aGroup['description']      = LimeExpressionManager::ProcessString($gl['description'], null, null, 3, 1);
```
Shnoulle, line 436: We process 2 times then ? 

Seems GROUPNAME have an issue, but process description is another issue (and i don't think we need it).

Shnoulle: https://github.com/LimeSurvey/LimeSurvey/blob/435badbd8f103673ece463a6c2bea8a4296ff25f/themes/survey/vanilla/views/subviews/survey/group_subviews/group_desc.twig#L25

Shnoulle (commented):
I don't lime to process string 2 times. And if some theme manager use it differently ? It broke.

application/helpers/SurveyRuntimeHelper.php
c-schmitz, line 436: I agree.

c-schmitz (changes requested):

Shnoulle:
I update the mantis : https://bugs.limesurvey.org/view.php?id=17188

With some test  on 3.25.19 : GID is OK, always, GROUNAME is OK in group by group, not in all in one.

application/helpers/SurveyRuntimeHelper.php
gabrieljenik: Yes, I agree.

gabrieljenik:
I think GID and Groupname are not rendered because of this.
 https://github.com/gabrieljenik/LimeSurvey/blob/62000f0f910761b305c39ec7e2b0875bb2c16f68/application/helpers/SurveyRuntimeHelper.php#L513

Any idea why is that there?

Shnoulle:
1. GID is rendered in 3.X
2. GID and GROUPNAME is rendered in Grou by group mode

GID and GROUPNAME must be set to empty when we are not inside a group.
I think it was added for this a long time ago.

Capture d’écran_2021-03-26_14-07-07

gabrieljenik:
> GID and GROUPNAME must be set to empty when we are not inside a group.

Maybe we are setting it this all time? That could be the issue?

Shnoulle:
See my screenshot : issue **only** in all in one page.

gabrieljenik:
But that's for LS3 or LS4?

Shnoulle:
See mantis : 4.X OK on group by group mode

gabrieljenik:
We will debug, thanks

Shnoulle:
> Any idea why is that there?

If you remove it : in 4.X, you have GID of group 2 in group 1

Capture d’écran_2021-03-27_09-08-30

Shnoulle:
The only solution to get in in groupdescription : it's your solution  … 
https://github.com/LimeSurvey/LimeSurvey/pull/1821/files#diff-f2c0610f506dc0397d2efd8d8203a78dc44f2825e26eb0ac70efd87205d2e36eR436

I don't see other way.

Must be done in GROUPNAME too ? if you use for title of group : Group {GID} ?

gabrieljenik:
I would just remove the EM Processing for the group on the twig.

Afetr some research we found that when going in "all in one" mode, twig only has EM as setup initially. 
Later when looping through the groups, EM instance is not refreshed, so I think there is the issue. Also, I don't think EM has the info from all groups loaded.

So removing processString from twig could be good.
The issue is the backward compatibility.

Another thought, going deeper it may be conceptually wrong that twig process the EM for that.
Does twig (a presentation layer) has all the info needed to call the EM?

olleharstedt:
Hm.

maziminke:
"processString()" is used at themes quite often so let's not change that or make sure we guarantee backward compatiblitly.

Shnoulle:
> "processString()" is used at themes quite often so let's not change that or make sure we guarantee backward compatiblitly.

I make some test : even with the fix of @gabrieljenik : there are no issue to processtring again.

A solution : update group_desc.twig `{{processString(aGroup.description) }}`

by
```
            {{ processString(
                aGroup.description,
                0,
                3,
                ({
                    'GROUPNAME' : aGroup.name,
                    'QID' : '',
                    'GID' : aGroup.gid
                })
            ) }}
```

It work in 3.X except aGroup.gid is empty

@gabrieljenik maybe you can fix it like this ? @maziminke you can update your template now :)

gabrieljenik:
Hi All,

I believe a vaible solution oving forward:
- Keep my fix
- removing processString from twig, as moving forward.
- As for backward compatibility shouldn't be an issue. Themes that double process, shouldn't throw errors according to Denis.

Will check if GroupName accepts EM in LS2

gabrieljenik:
> > "processString()" is used at themes quite often so let's not change that or make sure we guarantee backward compatiblitly.
> 
> I make some test : even with the fix of @gabrieljenik : there are no issue to processtring again.
> 
> A solution : update group_desc.twig `{{processString(aGroup.description) }}`
> 
> by
> 
> ```
>             {{ processString(
>                 aGroup.description,
>                 0,
>                 3,
>                 ({
>                     'GROUPNAME' : aGroup.name,
>                     'QID' : '',
>                     'GID' : aGroup.gid
>                 })
>             ) }}
> ```
> 
> It work in 3.X except aGroup.gid is empty
> 
> @gabrieljenik maybe you can fix it like this ? @maziminke you can update your template now :)

Why would you reprocess with all that information, which could be faulty sometimes at twig level (case All in one)?

Shnoulle:
> 
> Why would you reprocess with all that information, which could be faulty sometimes at twig level (case All in one)?


I didn't reprocess any information : i use https://github.com/LimeSurvey/LimeSurvey/blob/6437c998731e1e79da24c394ef205444cfa75cdf/application/helpers/expressions/em_manager_helper.php#L4143 

I sent the array. I have group.name, but not group.gid in 3.X

Adding gid in aGroup fix the issue

gabrieljenik:
> I didn't reprocess any information : i use
Ok, but it requires to update all themes.

I prefer to apply central fix, out of twig, and then if someone wants to update their themes to not reprocess (although its harmless) they can.

Shnoulle:
> I prefer to apply central fix, out of twig, and then if someone wants to update their themes to not reprocess (although its harmless) they can.

Yes, +1 , remind to apply to group name too :) It can contain GID ;)

olleharstedt:
Are the discussions resolved?

Shnoulle:
@maziminke : you confirm ?

gabrieljenik:
Added a new commit to the PR as to deal with replacements on the group name

gabrieljenik:
This is all implementes.
As soon as @maziminke confirms, this is ready for testing cycle.

Shnoulle:
> Added a new commit to the PR as to deal with replacements on the group name

Maybe : must not send GROUPNAME inside groupname replacement ? 

(someone must test set group-name to `This is {GROUPNAME}` 😱 🤣

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -417,7 +417,7 @@ public function run($surveyid, $args)
                 $aGroup['class'] = ' ls-hidden';
             }
 
-            $aGroup['name']        = $gl['group_name'];
+            $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, null, 3, 1);
```
c-schmitz, line 420: You will need to strop {GROUPNAME} here.

Using the text " This is {GROUPNAME} inception" in the groupname leads to

"This is This is This is This is Group 2 inception inception inception inception "

and will probably create unnecessary load.

c-schmitz (changes requested):

application/helpers/SurveyRuntimeHelper.php
Shnoulle: Yes, but it's a Survey manager issue.
Still : easy fix : send only {GID}

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -418,7 +418,7 @@ public function run($surveyid, $args)
 
             $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, null, 3, 1);
             LimeExpressionManager::updateReplacementFields(array(
-                'GROUPNAME' => $gl['group_name'],
+                'GROUPNAME' => $aGroup['name'],
```
gabrieljenik, line 421: Using the processed string as replacement value to avoid recursion.

PR #1822, opened 2021-03-26
BF_remote_control_and_rpc_client

arnaudbore:
Hi !

While playing with LS, we end up finding two small issues and here are some suggestions to fix them.

1. **Remote control**
This will return useful information - if saved, you get 'data' on the user - if not, you get the 'errors' (otherwise, with the original code, you never see that).

2. **jsonRPCClient**
The issue here is that $this->debug is never set to begin with.

Hope this will help. 
Arnaud

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1782,9 +1782,9 @@ class remotecontrol_handle
                     $token->generateToken();
                 }
                 if ($token->save()) {
-                    $return[] = $token->getAttributes();
+                    $return = $token->getAttributes();
                 } else {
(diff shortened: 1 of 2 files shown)


c-schmitz:
Thank you!

arnaudbore:
Thank you @c-schmitz , I just wanted to add that "we" included @Apps-n-Add-Ons who found and fix this issue. Do you know when you'll be releasing a new version with this fix ?

c-schmitz:
Actually I found out that the fix in add_participants is bogus and reverted it.
The errors are written and give back for each entry, that was written, already in the old code because
$aParticipant is used as a pointer here:

  foreach ($aParticipantData as &$aParticipant) {

PR #1823, opened 2021-03-26
Fixed issue #16782: Survey groups : theme options: same theme name appear multiple time in list

gabrieljenik:
Fixing issue and deleting wrongly generated records: duplicate template configurations. Keeping the first one.


Shnoulle:
Original issue was fixed ? No way to get multiple one ?

gabrieljenik:
> Original issue was fixed ? No way to get multiple one ?

We think issue is solved and duplicates shouldn't be created (in relation to this issue/situation).
Why...? Have found a crack?

application/helpers/update/updatedb_helper.php
```diff
@@ -3783,12 +3783,29 @@ function ($v) {
...
+            $deleteQuery = "DELETE FROM {{template_configuration}}
+                WHERE id NOT IN (
+                    SELECT id FROM (
+                        SELECT MIN(id) as id
+                            FROM {{template_configuration}} t 
+                            GROUP BY t.template_name, t.sid, t.gsid, t.uid
+                    ) x
+                )";
+            $oDB->createCommand($deleteQuery)->execute();
+            $oDB->createCommand()->update('{{settings_global}}', array('stg_value' => 442), "stg_name='DBVersion'");
```
c-schmitz, line 3806: Wrong DBVersion here

c-schmitz (changes requested):

application/helpers/update/updatedb_helper.php
gabrieljenik: Fixed. Thanks!

PR #1829, opened 2021-03-29
Fixed issue #16414: Question editor is not loaded when config.xml attribute has empty string as default

gabrieljenik:
All empty nodes are turned into null values, instead of an empty std objec / array. (As in LS4)


olleharstedt:
I think I fixed this exact issue in LS4 already.

PR #1833, opened 2021-03-31
SPSS SAV export plugin

adamzammit:
Dev: Preview feature - SPSS SAV export plugin

I know new features should go to develop/master but I currently only use the 3.x-LTS branch and I think this plugin should basically work on develop/master.

I'm putting this feature out there to see if there is interest in testing / mainlining it - and this is a good way to test it on an existing 3.x-LTS install.

This feature allows you to export an SPSS SAV (binary) file using the same export functions as for CSV/Excel/etc and as based on the Stata export plugin along with using the tiamo/spss library for export.

I have tested it with a simple survey along with reading the file in to PSPP - but if anyone out there wants to give it a go please report back on if it works for you. Will certainly save some time having to read in syntax and locate where your data file is/etc.


Shnoulle:
Great , great , great !

Do you have this plugin as a separate one ? 

Next step : remove all SPSS code from core for 5.X , right ?

adamzammit:
I've tested it on 4.x and it appears to be working.

Also have it as a separate plugin here:

https://github.com/adamzammit/ExportSPSSsav

PR #1834, opened 2021-04-01
New Feature: Export to SPSS sav core plugin (for develop branch)

adamzammit:
Dev: Updates database version to 444


olleharstedt:
Do we have a Mantis ticket for this feature, where we can discuss it?

adamzammit:
@olleharstedt - here it is: https://bugs.limesurvey.org/view.php?id=17226

c-schmitz (approved):

olleharstedt:
As I mentioned in the ticket, I'm not sure about the fact that we have two vendor/ folders in the system now. How would we check this for third-party security updates, for example?

Shnoulle:
Maybe add a way for "library" by and for plugin ?

See https://gitlab.com/SondagesPro/coreAndTools/limeMpdf for example.

This plugin do … nothing… just offer a librarie for other plugin.

Then in 5.X : upload/lib/ or something like that .

Core lib are in app/core/lib/

Shnoulle:
Related feature request : https://bugs.limesurvey.org/view.php?id=14567

See https://programmer.spip.net/Installing-external-libraries for a usage sample on another tool.

c-schmitz:
> As I mentioned in the ticket, I'm not sure about the fact that we have two vendor/ folders in the system now. How would we check this for third-party security updates, for example?

I agree. So if this is a core plugin it would need to use core folders, because we need to maintain any libraryies . If it isn't a core plugin then it doesn't need to, because the original author needs to take care of that.

adamzammit:
Hi again,

It looks like the twofactor authentication plugin has it's own vendor directory too, following the same pattern as this feature request:

https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/plugins/TwoFactorAdminLogin/vendor

Do both of these have to change or should this be ok to proceed as is?

olleharstedt:
Yeah, 2FA plugin should also not have its own composer folder, since it's core.

Shnoulle:
remind to update the [Minor Semantic Versioning number](https://semver.org/) for plugin, theme and other :)

adamzammit:
Ok I've make a new pull request with the dependencies being put in core: https://github.com/LimeSurvey/LimeSurvey/pull/1877

PR #1835, opened 2021-04-01
Fixed issue #16958: User with roles can not have more rights

gabrieljenik:
Added a warning message to the assignment of permissions. Modified the controller so that it effectively cleans up the roles when you assign individual permissions.

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -417,7 +417,7 @@ class UserManagementController extends LSBaseController
         $userId = Yii::app()->request->getPost('userid');
         $aPermissions = Yii::app()->request->getPost('Permission', []);
 
-        $clearUser = Permissiontemplates::model()->clearUser($userId);
+        Permissiontemplates::model()->clearUser($userId);
 
         $results = $this->applyPermissionFromArray($userId, $aPermissions);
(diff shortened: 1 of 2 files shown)


application/controllers/UserManagementController.php
```diff
@@ -416,6 +416,9 @@ public function actionSaveUserPermissions()
         }
         $userId = Yii::app()->request->getPost('userid');
         $aPermissions = Yii::app()->request->getPost('Permission', []);
+
+        $clearUser = Permissiontemplates::model()->clearUser($userId);
```
olleharstedt, line 420: Is this variable used anywhere?

gabrieljenik: Removed

olleharstedt: Is it removed? Or am I stupid?

Shnoulle:
Why not disable all checkbox ?

More clear  (_user didn't read_)

application/controllers/UserManagementController.php
gabrieljenik: I am. Code commited wasn't pushed

olleharstedt:
Thanks!

PR #1836, opened 2021-04-05
Fixed issue #17195: When attaching PDF to email invitation, full path is kept as value of a hidden input

gabrieljenik:
Changing from full path to relative path on that input
Reviewing URL for windows


Shnoulle:
Fix windows issue too :)

- [16905](https://bugs.limesurvey.org/view.php?id=16905): Windows server : good attachment throw JS error (and broke all JS action)
- [16903](https://bugs.limesurvey.org/view.php?id=16903): Windows server : unable to save attachements

PR #1837, opened 2021-04-07
Fixed issue #15892: use same logic for validation and unanswered for QT_O

febsn:
Dev: Fixes a bug where type O questions are marked valid but unanswered if
an answer is chosen, but comment left empty.

Fixed issue #15892: use same logic for validation and unanswered for QT_O

This is the fix in the unanswered state logic as discussed in https://bugs.limesurvey.org/view.php?id=15892. Unfortunately I'm not experienced enough in LS development (and php in general) to provide the requested tests; I'd be glad if someone could help me out on that.

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -6235,7 +6235,14 @@ class LimeExpressionManager
                     }
                     // what about optional vs. mandatory comment and 'other' fields?
                     break;
-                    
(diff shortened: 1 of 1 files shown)


c-schmitz:
Thank you very much. I also added a hint to the manual that the comment field cannot be configured to be mandatory, at all.

Shnoulle:
About comment mandatory on manual : it can be done using Validation.

I can update to put a sample of validation if you're OK @c-schmitz

PR #1838, opened 2021-04-07
Fixed issue #17204: {GROUPNAME} and other variables not rendered

gabrieljenik:
Adding EM replacements for group name and description.

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -317,9 +317,9 @@ class SurveyRuntimeHelper
             // Set basic replacements to use on group texts
             $aStandardsReplacementFields = getStandardsReplacementFields($this->aSurveyInfo);
             $aStandardsReplacementFields['GID'] = $gid;
-            $aStandardsReplacementFields['GROUPNAME'] = $gl['group_name'];
 
             $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, $aStandardsReplacementFields, 3, 1);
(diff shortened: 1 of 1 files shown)


application/helpers/SurveyRuntimeHelper.php
```diff
@@ -314,7 +314,12 @@ public function run($surveyid, $args)
                 $aGroup['class'] = ' ls-hidden';
             }
 
-            $aGroup['name']        = $gl['group_name'];
+            // Set basic replacements to use on group texts
+            $aStandardsReplacementFields = getStandardsReplacementFields($this->aSurveyInfo);
+            $aStandardsReplacementFields['GID'] = $gid;
+            $aStandardsReplacementFields['GROUPNAME'] = $gl['group_name'];
+
+            $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, $aStandardsReplacementFields, 3, 1);
```
c-schmitz, line 322: Strip {GROUPNAME} to avoid recursion.

c-schmitz (changes requested):

application/helpers/SurveyRuntimeHelper.php
gabrieljenik: Wouldn;t that happen with any other field replacement?

gabrieljenik: On this new commit, we are using, the processed group name to use a replacement value, and not the original expression. Is that what you meant?

PR #1842, opened 2021-04-07
Fixed issue #17120: After picking a new question theme, the settings are not properly updated

gabrieljenik:
Code reorganization.
- Added some comments.
- Have created helper functions as to be able to have a better outlook of what some function do.

Solution Summary.
The key of the solution is to be able to pickup theme attributes from a theme name and not from a the attribute set on the question. Then added a theme override parameter which will set which theme attributes should be picked up.

So now, question attributes may contain the attributes of a temp theme, which is not set on the DB.

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -2292,28 +2292,13 @@ class QuestionAdministrationController extends LSBaseController
         $advancedSettings = $oQuestion->getAdvancedSettingsWithValues(null, $sQuestionTheme);
 
         // Group the array in categories
-        $advancedSettings = self::groupAttributesByCategory($advancedSettings);
+        $questionAttributeHelper = new LimeSurvey\Models\Services\QuestionAttributeHelper();
+        $advancedSettings = $questionAttributeHelper->groupAttributesByCategory($advancedSettings);
 
         // This category is "general setting".
         unset($advancedSettings['Attribute']);
 
         return $advancedSettings;
-    }
-
-    /**
-     * Receives an array of question attributes and groups them by category.
-     * Used by advanced settings widget.
-     *
-     * @param array $aAttributes
-     * @return array Grouped question attributes, with category as array key
-     */
-    private static function groupAttributesByCategory($aAttributes)
-    {
-        $aByCategory = [];
-        foreach ($aAttributes as $aAttribute) {
-            $aByCategory[$aAttribute['category']][] = $aAttribute;
-        }
-        return $aByCategory;
     }
 
     /**
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -2278,29 +2278,42 @@ class QuestionAdministrationController extends LSBaseController
      *
      * @param int $iQuestionId
      * @param string $sQuestionType
-     * @param string $question_template
+     * @param string $sQuestionTheme
      * @return array
      * @throws CException
      * @throws Exception
      */
-    private function getAdvancedOptions($iQuestionId = null, $sQuestionType = null, $question_template = 'core')
+    private function getAdvancedOptions($iQuestionId = null, $sQuestionType = null, $sQuestionTheme = 'core')
     {
         //here we get a Question object (also if question is new --> QuestionCreate)
-
         $oQuestion = $this->getQuestionObject($iQuestionId, $sQuestionType);
-        $advancedSettings = $oQuestion->getAdvancedSettingsWithValuesByCategory(null);
-        // TODO: Why can empty array be saved as value?
-        foreach ($advancedSettings as &$category) {
-            foreach ($category as &$setting) {
-                if ($setting['value'] === []) {
-                    $setting['value'] = null;
-                }
-            }
-        }
+
+        // Get the advanced settings array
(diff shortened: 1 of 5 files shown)


application/controllers/QuestionAdministrationController.php
```diff
@@ -2278,31 +2278,44 @@ function ($oAnswerOption) {
...
     }
 
+    /**
+     * Receives an array of question attributes and groups them by category.
+     * Used by advanced settings widget.
+     *
+     * @param array $aAttributes
+     * @return array Grouped question attributes, with category as array key
+     */
+    private static function groupAttributesByCategory($aAttributes)
```
olleharstedt, line 2310: Doesn't need to be static? Static functions can't be mocked and should be avoided in case they're really needed (like for factory methods).

olleharstedt (changes requested):

olleharstedt:
Any idea how this solution will interact with the newQuestionAttributes event?

https://bugs.limesurvey.org/view.php?id=17049

application/controllers/QuestionAdministrationController.php
gabrieljenik: Honestly, it was non-static initially. But it felt very wrong, as this function don't use anything from the object. Is a helper method.
Anyway, will change it

gabrieljenik:
> Any idea how this solution will interact with the newQuestionAttributes event?
> 
> https://bugs.limesurvey.org/view.php?id=17049

That's the next step. I believe that having done some tidying on the code, now should be much easier to implement it.
Actually, we were waiting for this commit to be reviewed and debated and then move onto issue 17409 and 16669

application/controllers/QuestionAdministrationController.php
olleharstedt: Yeah, if you have no "$this", another alternative is to move it to a normal function (if it has no side-effects, at least; for a discussion of what is considered a side-effect, see: https://manual.limesurvey.org/Code_quality_guide#Push_side-effects_up_in_the_stack_trace).

olleharstedt:
> > Any idea how this solution will interact with the newQuestionAttributes event?
> > https://bugs.limesurvey.org/view.php?id=17049
> 
> That's the next step. I believe that having done some tidying on the code, now should be much easier to implement it.
> Actually, we were waiting for this commit to be reviewed and debated and then move onto issue 17409 and 16669

Great, I will read and merge then.

olleharstedt:
One thing: Please never add static methods in ActiveRecord classes. In the end, we want ActiveRecord to _only_ contain the database relation. Can you move them to a service class, please? Thank you. Reasons are as before:

1) One responsibility per class (database interaction is one)

2) Testability (static methods can't easily be mocked)

Shnoulle:
Then here for 

-  public static function getMergedQuestionAttributes https://github.com/LimeSurvey/LimeSurvey/pull/1842/files#diff-ac521af6038093a1c4c5c35c0e901c23f2f4dc0572163912ad047d4f12a4574bR646
- public static function getSanitizedQuestionAttributes https://github.com/LimeSurvey/LimeSurvey/pull/1842/files#diff-ac521af6038093a1c4c5c35c0e901c23f2f4dc0572163912ad047d4f12a4574bR685 
- public static function fillAttributesWithValues https://github.com/LimeSurvey/LimeSurvey/pull/1842/files#diff-ac521af6038093a1c4c5c35c0e901c23f2f4dc0572163912ad047d4f12a4574bR712 

Where the new Class must be ? 
https://github.com/LimeSurvey/LimeSurvey/tree/master/application/models/services  ?

application/controllers/QuestionAdministrationController.php
gabrieljenik: > One thing: Please never add static methods in ActiveRecord classes. In the end, we want ActiveRecord to _only_ contain the database relation. Can you move them to a service class, please? Thank you. Reasons are as before:
> 
> 1. One responsibility per class (database interaction is one)
> 2. Testability (static methods can't easily be mocked)
> 3. Readability (big classes with lots of different behaviour are harder to get a conception of)

Yes, we were focused on unwinding the code/
Probably the best is to create a service class.

olleharstedt:
@Shnoulle I'm missing line numbers for those links. :|

Yes, should be a service class. Best is if it has a clear purpose and not only being a "helper".

Shnoulle:
Uility Class !
Or Utils !

Maybe best ?

olleharstedt:
Thanks!

olleharstedt:
FooBuilder, FooFetcher, FooIDon'tKnow. It's hard. :cry:

olleharstedt:
In some cases easy:

activate_survey --> SurveyActivator

createFieldMap --> FieldmapBuilder or FieldmapCreator

Shnoulle:
I just talk of this new static function : clearly seems utilities, not for old and really long function

olleharstedt:
Oh ok. Yes, sure. Although "utility" classes have a risk of growing forever and ever and...

Shnoulle:
QuestionAttributeUtilities::getMerged ?
QuestionAttributeUtilities::getSanitized ?
QuestionAttributeUtilities::fillWithValues ?
QuestionAttributeUtilities::groupByCategory ?

Or ?

olleharstedt:
Yes, hard to think of a better name or a way to group them together. Hopefully it will be obvious if the class grows a lot. :)

PR #1843, opened 2021-04-09
Fixed issue #16669: getQuestionAttributes function don't get the plugins attribute

gabrieljenik:
Picking up plugin attributes and merging them with theme and std ones.

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -309,6 +309,10 @@ class Question extends LSActiveRecord
         $questionAttributeHelper = new LimeSurvey\Models\Services\QuestionAttributeHelper();
         $aAttributes = $questionAttributeHelper->mergeQuestionAttributes($aQuestionTypeAttributes, $aThemeAttributes);
 
+        // Get question attributes from plugins ('newQuestionAttributes' event)
+        $aPluginAttributes = $questionAttributeHelper->getAttributesFromPlugin($this->type);
+        $aAttributes = $questionAttributeHelper->mergeQuestionAttributes($aAttributes, $aPluginAttributes);
+
         uasort($aAttributes, 'categorySort');
 
         // Fill attributes with values
--- a/application/models/QuestionTheme.php
+++ b/application/models/QuestionTheme.php
@@ -1049,6 +1049,7 @@ class QuestionTheme extends LSActiveRecord
         }
 
         $questionAttributeHelper = new LimeSurvey\Models\Services\QuestionAttributeHelper();
+        $additionalAttributes = $questionAttributeHelper->fillMissingCategory($additionalAttributes, gT('Template'));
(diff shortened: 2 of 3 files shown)


olleharstedt:
This has the same issue as before - the plugin is run from two different places. Maybe it's the only way. Still I wonder if it's possible to use ONLY the method "getQuestionThemeAttributeValues" to fetch question theme attributes.

gabrieljenik:
> Or, is it possible to have ONE function to get question theme attributes (together with their values, if any)? And fire the event there?

Why do you wan't fir it from there? I was actually thinking about removing it from there.
Let's review on a call?

olleharstedt:
Sure, ping me tomorrow.

gabrieljenik:
After call, this ticket was created: https://bugs.limesurvey.org/view.php?id=17242

olleharstedt:
Merged after deciding to continue with more clean-up PRs.

Ticket for unit-test: https://bugs.limesurvey.org/view.php?id=17241

Duplicate ticket for clean-up: https://bugs.limesurvey.org/view.php?id=17240

Shnoulle:
@olleharstedt About issue with https://manual.limesurvey.org/NewQuestionAttributes 

For example : 
````
        'inputtype' => 'switch',
        'options'   => array(
          0 => gT('No'),
          1 => gT('Yes'),
        ),
````
Nothing is shown

````
        'inputtype'=>'singleselect',
        'options'=>array(
          'afteranswer'=>$this->_translate("The script is inserted just after answer part."),
          CClientScript::POS_HEAD=>$this->_translate("The script is inserted in the head section right before the title element (POS_HEAD)."),
          CClientScript::POS_BEGIN=>$this->_translate("The script is inserted at the beginning of the body section (POS_BEGIN)."),
          CClientScript::POS_END=>$this->_translate("The script is inserted at the end of the body section (POS_END)."),
          CClientScript::POS_LOAD=>$this->_translate("The script is inserted in the window.onload() function (POS_LOAD)."),
          CClientScript::POS_READY=>$this->_translate("The script is inserted in the jQuery's ready function (POS_READY)."),
        ),
````
Show a dropdown with 'array' (and break with debug)

Report as issue or API upgrade need updating manual ?

Shnoulle:
Oh no … ok : before 4.5.0 : need 
````
          'option'=> array(
            array('value'=>0, 'text'=> gT("No")),
            array('value'=>1, 'text'=> gT("Yes")),
          ),
````

for 4.5 : come back with compatibility of 3.X …

Still some issue with help and with switch

olleharstedt:
Report everything, assign to Gabriel. :) :partying_face:

olleharstedt:
Or tell me which function that needs unit-test to stay correct.

PR #1845, opened 2021-04-15
Fixed issue #17235: Token additional attributes are blank after editing

gabrieljenik:
- Fixed upgradeTokens176() function of updatedb_helper: it was not deserializing the 'attributedescriptions' field, so all the surveys ended up without attribute descriptions. Also added a small fix to initialize the 'cpdbmap' field when missing.
- Fixed getTokenAttributes() method of Survey model: it was already merging the actual extra attributes from the token's table with the attributes defined in 'attributedescriptions' field, but only when the later wasn't empty. Now it always does the merge.


c-schmitz (changes requested):
I don't agree with the changes in  upgradeTokens176().
DBVersion 177 is doing the conversion from serialization to JSON.

gabrieljenik:
Fixed as spoken

PR #1846, opened 2021-04-15
Fixed issue #17235: Token additional attributes are blank after editing

gabrieljenik:
- Fixed upgradeTokens176() function of updatedb_helper: it was not deserializing the 'attributedescriptions' field, so all the surveys ended up without attribute descriptions. Also added a small fix to initialize the 'cpdbmap' field when missing.
- Fixed getTokenAttributes() method of Survey model: it was already merging the actual extra attributes from the token's table with the attributes defined in 'attributedescriptions' field, but only when the later wasn't empty. Now it always does the merge.

### DIFF
--- a/application/helpers/update/updatedb_helper.php
+++ b/application/helpers/update/updatedb_helper.php
@@ -3504,7 +3504,7 @@ function upgradeTokens176()
                     $aAttribute['cpdbmap'] = '';
                 }
             }
-            $oDB->createCommand()->update('{{surveys}}',array('attributedescriptions'=>json_encode($aAttributes)),"sid=".$arSurvey['sid']);
+            $oDB->createCommand()->update('{{surveys}}',array('attributedescriptions'=>serialize($aAttributes)),"sid=".$arSurvey['sid']);
         }
     }
     unset($arSurveys);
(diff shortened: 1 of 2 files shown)


c-schmitz (changes requested):
I don't agree with the changes in  upgradeTokens176().
DBVersion 177 is doing the conversion from serialization to JSON.

gabrieljenik:
> I don't agree with the changes in upgradeTokens176().
> DBVersion 177 is doing the conversion from serialization to JSON.

I think the attributes get blanked after 176.
As this line receives info in an unexpected form, which results in blanking.
https://github.com/LimeSurvey/LimeSurvey/pull/1846/files#diff-fac7e21d70e934b24d9330f951f13c5df1badba6f41fa52c894364fb09db4d8cL3501

application/helpers/update/updatedb_helper.php
```diff
@@ -3498,7 +3498,13 @@ function upgradeTokens176()
...
                 }
             }
-            $oDB->createCommand()->update('{{surveys}}',array('attributedescriptions'=>serialize($aAttributes)),"sid=".$arSurvey['sid']);
+            // Add 'cpdbmap' if missing
+            foreach ($aAttributes as &$aAttribute) {
+                if (!isset($aAttribute['cpdbmap'])) {
+                    $aAttribute['cpdbmap'] = '';
+                }
+            }
+            $oDB->createCommand()->update('{{surveys}}',array('attributedescriptions'=>json_encode($aAttributes)),"sid=".$arSurvey['sid']);
```
gabrieljenik, line 3507: After speaking with @c-schmitz this should be serialize, not json.

gabrieljenik: Done

PR #1847, opened 2021-04-15
Fixed issue #17100: New question will be added to the wrong group

gabrieljenik:
Reassigning group after part of the screen s refreshed

### DIFF
--- a/assets/scripts/admin/questionEditor.js
+++ b/assets/scripts/admin/questionEditor.js
@@ -1514,7 +1514,9 @@ $(document).on('ready pjax:scriptcomplete', function () {
       });
       try {
         const [html, html2] = await Promise.all([generalSettingsPromise, advancedSettingsPromise]);
+        var currentGroup = $('#gid').children("option:selected").val();
         $('#general-settings').replaceWith(html);
+        $('#gid').val(currentGroup);
         // TODO: Double check HTML injected here. Extra div?
         $('#advanced-options-container').replaceWith(html2);
         $('.question-option-help').hide();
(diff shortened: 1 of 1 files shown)


assets/scripts/admin/questionEditor.js
```diff
@@ -1514,7 +1514,9 @@ $(document).on('ready pjax:scriptcomplete', function () {
       });
       try {
         const [html, html2] = await Promise.all([generalSettingsPromise, advancedSettingsPromise]);
+        var currentGroup = $('#gid').children("option:selected").val();
```
olleharstedt, line 1517: Can you use let or const instead, please?

olleharstedt (changes requested):

assets/scripts/admin/questionEditor.js
gabrieljenik: Done

olleharstedt: Sorry, I should have said: Use const (the value is never changed...)

gabrieljenik: Well, I disagree with you on this one. From its conception is a variable, not a constant... I would keep with let... But if you insist, of course I will change it :)

olleharstedt: `const` should be used for variables that do not change. Compare with the `const` keyword in C++. Thanks, though.

olleharstedt: https://docs.microsoft.com/en-us/cpp/cpp/const-cpp?view=msvc-160

olleharstedt:
Thanks!

PR #1849, opened 2021-04-16
16334: Condition editor: previous question not available

gabrieljenik:
As data integrity checl about order was already there, we just reinforce the message :)


c-schmitz:
@olleharstedt Commit message is not proper - please make sure it is correct on merge.

olleharstedt:
I saw, sorry.

gabrieljenik:
I am sorry

PR #1850, opened 2021-04-19
Fixed issue #16669: getQuestionAttributes function don't get the plugins attribute

gabrieljenik:
Moving attribute-value-fecthing-code to service class.
Reusing that from different places, as to return values for advanced attributes sometimes and some other return all attribute and values.

### DIFF
--- a/application/models/services/QuestionAttributeHelper.php
+++ b/application/models/services/QuestionAttributeHelper.php
@@ -247,18 +247,17 @@ class QuestionAttributeHelper
         $aThemeAttributes = \QuestionTheme::getAdditionalAttrFromExtendedTheme($sQuestionTheme, $oQuestion->type);
 
         // Merge the attributes with the question theme ones
-        $questionAttributeHelper = new self();
-        $aAttributes = $questionAttributeHelper->mergeQuestionAttributes($aQuestionTypeAttributes, $aThemeAttributes);
+        $aAttributes = $this->mergeQuestionAttributes($aQuestionTypeAttributes, $aThemeAttributes);
 
         // Get question attributes from plugins ('newQuestionAttributes' event)
-        $aPluginAttributes = $questionAttributeHelper->getAttributesFromPlugin($oQuestion->type);
-        $aAttributes = $questionAttributeHelper->mergeQuestionAttributes($aAttributes, $aPluginAttributes);
+        $aPluginAttributes = $this->getAttributesFromPlugin($oQuestion->type);
+        $aAttributes = $this->mergeQuestionAttributes($aAttributes, $aPluginAttributes);
 
         uasort($aAttributes, 'categorySort');
 
         // Fill attributes with values
         $aLanguages = is_null($sLanguage) ? $oSurvey->allLanguages : [$sLanguage];
-        $aAttributes = $questionAttributeHelper->fillAttributesWithValues($aAttributes, $aAttributeValues, $aLanguages);
+        $aAttributes = $this->fillAttributesWithValues($aAttributes, $aAttributeValues, $aLanguages);
 
         return $aAttributes;
     }
--- a/application/models/services/QuestionAttributeHelper.php
+++ b/application/models/services/QuestionAttributeHelper.php
@@ -53,14 +53,13 @@ class QuestionAttributeHelper
      */
     public function sanitizeQuestionAttributes($aAttributes)
     {
-        foreach ($aAttributes as $key => &$aAttribute) {
-            // Make sure the attribute has a name. If it doesn't try to use the array key as name,
-            // or remove the attribute if the key is numeric.
-            if (!isset($aAttribute['name'])) {
-                if (!is_numeric($key)) {
-                    $aAttribute['name'] = $key;
-                } else {
-                    unset($aAttributes[$key]);
+        $aSanitizedAttributes = [];
+        foreach ($aAttributes as $key => $aAttribute) {
+            // Make sure the attribute has a name.
+            if (!is_numeric($key)) {
+                $aAttribute['name'] = $key;
+            } else {
+                if (!isset($aAttribute['name'])) {
                     continue;
                 }
             }
@@ -71,8 +70,15 @@ class QuestionAttributeHelper
                     $aAttribute[$propertyName] = null;
                 }
             }
+
+            // Make sure "options" have the expected structure
+            if (isset($aAttribute['options']['option']) && is_array($aAttribute['options']['option'])) {
+                $aAttribute['options'] = $aAttribute['options']['option'];
+            }
+
+            $aSanitizedAttributes[$aAttribute['name']] = $aAttribute;
         }
-        return $aAttributes;
+        return $aSanitizedAttributes;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


application/models/services/QuestionAttributeHelper.php
```diff
@@ -209,4 +215,51 @@ public function fillMissingCategory($aAttributes, $sCategoryName)
...
+        $sQuestionTheme = !empty($sQuestionThemeOverride) ? $sQuestionThemeOverride : $sQuestionTheme;
+
+        // Get advanced attribute definitions for the question type
+        $aQuestionTypeAttributes = \QuestionAttribute::getQuestionAttributesSettings($oQuestion->type, $advancedOnly);
+
+        // Get question theme attribute definitions
+        $aThemeAttributes = \QuestionTheme::getAdditionalAttrFromExtendedTheme($sQuestionTheme, $oQuestion->type);
+
+        // Merge the attributes with the question theme ones
+        $questionAttributeHelper = new self();
```
olleharstedt, line 250: Use `this` instead?

gabrieljenik:
Yes, and I believe i18n is missing from the output

olleharstedt:
There's a design challenge in the  application/models/services/QuestionAttributeHelper.php  now. All the previous methods have no dependencies, but the new method depend on (static) methods from model Question Attribute. So should they be in the same class? Or should the previous methods be moved outside a class instead, but in a namespace? (Might be OK _if_ they have neither state nor dependencies.)

olleharstedt:
Or you need a new helper: QuestionAttributeValuesHelper. :|

gabrieljenik:
> on (static) methods from model Questio

hmm I need to read it carefully. But maybe that can go to a reorganization ticket?

gabrieljenik:
> Yes, and I believe i18n is missing from the output

Done

gabrieljenik:
> There's a design challenge in the application/models/services/QuestionAttributeHelper.php now. All the previous methods have no dependencies, but the new method depends on (static) methods from model QuestionAttribute and QuestionTheme. So should they be in the same class? Or should the previous methods be moved outside a class instead, but in a namespace? (Might be OK _if_ they have neither state nor dependencies, contradicting what I said earlier.)
> 
> If you do constructor injection, you have to inject the dependencies even if they are not used (or do nullable dependencies).

I would try to follow the rule:
- DB Access, XML as a Source: Model
- Handling information: Service / Helper
- XML handling, like installation or scanning: Service / Helper

If I had to do these again, I would make the Attribute to be an abstract class with 3 implementation classes depending on the source: DB, XML and Plugin Events. I think all of them could be model-like classes (all sources of information) sharing part of their interfaces.

Then a QuestionAttributes Collection-Like class coordinating all of them (probably holding some of the code which is now on the service class).

The Service class would be then much slimer.

Shnoulle:
> Or you need a new helper: QuestionAttributeValuesHelper. :|

🤣
https://github.com/LimeSurvey/LimeSurvey/blob/3.x-LTS/application/helpers/questionHelper.php

application/models/services/QuestionAttributeHelper.php
```diff
@@ -209,4 +215,50 @@ public function fillMissingCategory($aAttributes, $sCategoryName)
...
+     * @param string|null $sLanguage If you give a language then only the attributes for that language are returned
+     * @param string|null $sQuestionThemeOverride   Name of the question theme to use instead of the question's current theme
+     * @param boolean $advancedOnly If set to true, only the advanced attributes will be returned
+     * @return array
+     */
+    public function getQuestionAttributesWithValues($oQuestion, $sLanguage = null, $sQuestionThemeOverride = null, $advancedOnly = false)
+    {
+        $oSurvey = $oQuestion->survey;
+        if (empty($oSurvey)) {
+            throw new \Exception('This question has no survey - qid = ' . json_encode($oQuestion->qid));
```
Shnoulle, line 233: json_encode ?

olleharstedt: json_encode will make null = "null" instead of empty string.

olleharstedt:
> If I had to do these again, I would make the Attribute to be an abstract class with 3 implementation classes depending on the source: DB, XML and Plugin Events. I think all of them could be model-like classes (all sources of information) sharing part of their interfaces.

Yes, this is the problem. We have one QuestionAttribute class for db, but as you say, since LS4, question attributes have multiple storage points, not only database.

application/models/services/QuestionAttributeHelper.php
Shnoulle: But : $oQuestion have always a qid (it's the pk). Maybe null value for sid ?

Maybe 

\Exception('Question " . $oQuestion->qid . ' survey  ' . json_encode($oQuestion->sid) . 'not found');

olleharstedt:
Thing is, I don't want to merge anything that I know will have to be refactored to be testable. If you get my point. :)

Shnoulle:
> > If I had to do these again, I would make the Attribute to be an abstract class with 3 implementation classes depending on the source: DB, XML and Plugin Events. I think all of them could be model-like classes (all sources of information) sharing part of their interfaces.
> 
> Yes, this is the problem. We have one QuestionAttribute class for db, but as you say, since LS4, question attributes have multiple storage points, not only database.

This make it really complex ! 😩 ;)

olleharstedt:
> > > If I had to do these again, I would make the Attribute to be an abstract class with 3 implementation classes depending on the source: DB, XML and Plugin Events. I think all of them could be model-like classes (all sources of information) sharing part of their interfaces.
> > 
> > 
> > Yes, this is the problem. We have one QuestionAttribute class for db, but as you say, since LS4, question attributes have multiple storage points, not only database.
> 
> This make it really complex ! weary ;)

No top-down design. :/

application/models/services/QuestionAttributeHelper.php
olleharstedt: + sprintf ^^

Shnoulle:
> Another thought, related to function composition: If we already have the question attributes, maybe it would be enough to do "populateValues(getQuestionAttributes());"?

But we need to be sure to have all attribute for current question with current theme and plugin.

No ?

olleharstedt:
> But we need to be sure to have all attribute for current question with current theme and plugin.

```php
$qf = new QuestionAttributeFetcher();
$qf->setTheme($theme);
$qf->setQuestionId($qid);
$attrs = $qf->fetch();
$attrsWithValues = $qf->populateValues($attrs);
```

?

Shnoulle:
> ?

Not exactly

> If we already have the question attributes

then need 2 function : one to get **whole** attributes definition + one for values (from DB ?)

olleharstedt:
> then need 2 function : one to get **whole** attributes definition + one for values (from DB ?)

Uh, what's the difference between whole definition and partial definition...?

Shnoulle:
whole :
 
- core attributes
- theme attribute
- plugin attribute
- (future) survey theme attribute
- (future) extend em attribute
- (future) i don't klnow attribute

partial : see all issue questiontheme and plugin with question attribute since 4.0.0 : we need to be sure to have whole attribute … and it's not the case in 4.0, 4.1, 4.2, 4.3, 4.4 …

olleharstedt:
Yes, and if `setTheme` is used, it would get attributes from core and theme (and plugin, by default). Can you amend my code with the changes you need?

Shnoulle:
> Yes, and if `setTheme` is used, it would get attributes from core and theme (and plugin, by default). Can you amend my code with the changes you need?

I don't want to amend anytghing , i just answer about: 

> Another thought, related to function composition: If we already have the question attributes, maybe it would be enough to do "populateValues(getQuestionAttributes());"?

Why not, but if we have a `populateValues` we need a `getAllAttributesDefinition` for a qid.

Else , about : 

```
$qf = new QuestionAttributeFetcher();
$qf->setTheme($theme);
$qf->setQuestionId($qid);
$attrs = $qf->fetch();  // Calls plugin event and gets XML attributes
$attrsWithValues = $qf->populateValues($attrs);
```

Since theme are  not in question : 

```
$qf = new QuestionAttributeFetcher();
$qf->setQuestionId($qid);
$qf->setTheme();
$attrs = $qf->fetch();  // Calls plugin event and gets XML attributes
$attrsWithValues = $qf->populateValues($attrs);
```

maybe 

```
/**
* set the theme for current qid
* @param string  $theme, if null get current one by database
* @return void
*/
public function setTheme($theme = null) {

}
```

olleharstedt:
> Why not, but if we have a `populateValues` we need a `getAllAttributesDefinition` for a qid.

Yes yes, this should be the _only_ behaviour. :) There's no use-case for excluding any question attributes (that I can think of...).

application/models/services/QuestionAttributeHelper.php
```diff
@@ -209,4 +215,50 @@ public function fillMissingCategory($aAttributes, $sCategoryName)
...
+        }
+
+        // Get attribute values
+        $aAttributeValues = \QuestionAttribute::getAttributesAsArrayFromDB($oQuestion->qid);
```
olleharstedt, line 237: Move to separate method in this class, so it can be mocked in the unit-test (no need for database setup/teardown).

olleharstedt (changes requested):

olleharstedt:
If we have _one_ domain model with _two_ storage strategies, we should consider not using ActiveRecord, but maybe in fact the repository pattern? https://designpatternsphp.readthedocs.io/en/latest/More/Repository/README.html

Maybe it's too complex and overkill, but the point is to separate the domain model from whatever storage strategy is used. ActiveRecord is not adapted for that, since it assumes a database storage (and not file/XML).

gabrieljenik:
After speaking with Olle, the following tickets were created:
https://bugs.limesurvey.org/view.php?id=17261
https://bugs.limesurvey.org/view.php?id=17261

olleharstedt:
Hm, I think I will merge this today and apply the small fixes I asked for above, should be fast. Might remove Hungarian notation while at it. :innocent:

gabrieljenik:
@olleharstedt Working on those. You can merge this. Will apply fixes on different PR

olleharstedt:
Oh, if you're working on it I won't touch it.

PR #1851, opened 2021-04-21
Correctly handle the case of no files having been uploaded

mrpg:
In a survey with optional uploads, the error "count(): Argument #1 ($var) must be of type Countable|array, null given" results if no file was uploaded.

This is fixed by checking for $aFiles being null, and setting $iSize to 0 if it is.

I am filing this bug under the "smaller internal changes" rule without a Mantis issue.

Fixed issue #NA
Dev: MAX R. P. GROSSMANN


c-schmitz:
Thank you!

PR #1853, opened 2021-04-23
Fixed issue #16638: EXPIRY keyword does not work in Survey description

gabrieljenik:
Forcing refresh at the start.

### DIFF
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -1891,7 +1891,7 @@ function display_first_page($thissurvey, $aSurveyInfo)
     $thissurvey                 = $aSurveyInfo;
     $thissurvey['aNavigator']   = getNavigatorDatas();
     LimeExpressionManager::StartProcessingPage();
-    LimeExpressionManager::StartProcessingGroup(-1, false, $surveyid); // start on welcome page
+    LimeExpressionManager::StartProcessingGroup(-1, false, $surveyid, true); // start on welcome page
 
     // WHY HERE ?????
     $_SESSION['survey_' . $surveyid]['LEMpostKey'] = mt_rand();


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/blob/37758c5000f37ef325a4b2f2684295d4a52ab4df/application/helpers/expressions/em_manager_helper.php#L6728

Then : the issue happen only for tester ? The newtest=Y fix the issue ?

If yes : Since there are a lot more issue than this single one : maybe don't fix ?

See this one : https://bugs.limesurvey.org/view.php?id=17107

gabrieljenik:
> Then : the issue happen only for tester ? The newtest=Y fix the issue ?

No, newTest is not related.

Seems tempvars are removed by StartProcessingPage
https://github.com/LimeSurvey/LimeSurvey/pull/1853/files#diff-4b5f6522204521a4e24e45dd60c31295cdc7298b0174dd96cf6c1e7d0334f4ffR1893

Then setVariableAndTokenMappingsForExpressionManager doesn't understand that they don't need to be refreshed (maybe wrongly? ).
That's why we force it.
https://github.com/LimeSurvey/LimeSurvey/blob/cd05811a0089341d4385bed83439d1401776588a/application/helpers/expressions/em_manager_helper.php#L3279

Maybe, the problem is that the refresh should happen also if tempvars is empty? But that would be a more general update than just updating this for first page (as we did)
@olleharstedt will probably say risk vs reward balance doesn't worth messing with it.

gabrieljenik:
> See this one : https://bugs.limesurvey.org/view.php?id=17107

I think it is something else. Not related from my point of view.

Shnoulle:
OK : right tempvars is not related to whole EM 

A question more : EXPIRY didn't work with newtest=Y too ?

gabrieljenik:
> A question more : EXPIRY didn't work with newtest=Y too ?

Nope. Didn't work

olleharstedt:
EXPIRY? Expire?

PR #1857, opened 2021-04-23
Dev 17262: Create new getQuestionAttributes() endpoint for Limesurvey API which is used by plugins

gabrieljenik:


Shnoulle:
New feature ? And need an API version update => for dev.

New API feature => need to be know with a simple float API checker.

Shnoulle:
Else : great ! 

:+1:  :heart:

gabrieljenik:
> New feature ? And need an API version update => for dev.
> 
> New API feature => need to be know with a simple float API checker.

Sorry, can you explain both a little bit further?

Shnoulle:
I mean : it's an API.

If a plugin want to check this API function : need to to test API when activate.

Then : App()->getconfig('versionnumber') > X.Y See https://semver.org/ .

I know i currently use verison_compare in my plugin because some «PATCH» broke API or add feature to API. 
But maybe in 5.Y or 6.Y or … we can use floatval … 

Or better : have a LimeSurvey::API constant to check …

gabrieljenik:
oh ok. I get it.
We can add something on the versions.php file.
Then the API can read it and show it out on method

Shnoulle:
@olleharstedt : nee version minor(feature) update here. In my opinion.

Alternative solution : create multiple API version :+1: 

- SurveyTheme API verison
- QuestionTheme API version
- Plugin API version

but : i think it's best if we use only a single API nulber

olleharstedt:
phpunit just aborts after green bar test. Weird.

olleharstedt:
Maybe try to run only the newly added test with debug = 2?

olleharstedt:
> @olleharstedt : nee version minor(feature) update here. In my opinion.

Sure.

Shnoulle:
* need

PR #1858, opened 2021-04-27
Fixed issue #16339: pluginhelper fullpagewrapper doesnt show content

gabrieljenik:
To clarify the problem: the fullpagewrapper works fine, unless one of the parameters in the URL is the survey id.

In that case the side menu appears because the layout_insurvey.php is used.

The _renderWrappedTemplate function of Survey_Common_Action.php decides that this layout should be used when the data in the view ($ aData) includes the "surveyid".

Until version 3.17.15, when you used the fullpagewrapper it worked fine even if you put a survey id in the URL. The problem is that from the commit de7707d700d1304110eca1e12fd22b3aa1d011b7 some parameters such as the survey id are included in the $ aData (it is not included by the helper plugin, but by Survey_Common_Action itself).

The solution then is for the helper plugin to explicitly specify that the layout_main.php must be used.

### DIFF
--- a/application/controllers/admin/PluginHelper.php
+++ b/application/controllers/admin/PluginHelper.php
@@ -62,7 +62,7 @@ class PluginHelper extends Survey_Common_Action
         $content = $this->getContent(null, $plugin, $method);
 
         $aData['content'] = $content;
-        $this->_renderWrappedTemplate(null, 'super/dummy', $aData);
+        $this->_renderWrappedTemplate(null, 'super/dummy', $aData, 'layout_main.php');
     }
 
     /**


Shnoulle:
Seems strange `$params['surveyid']` is not set if `Yii::app()->getRequest()->getParam('surveyid')`

Because $params start by https://github.com/LimeSurvey/LimeSurvey/blob/73bc85298a5237e8f46577e1d9ef7e3ca848a170/framework/web/CController.php#L325 

(and we add some)

Maybe need to update to getQuery ?

olleharstedt:
Same issue in LS4?

Shnoulle:
@gabrieljenik : do you have an idea why surveyid is not in $params before the XSS fix ? I really don't understand … 😩

gabrieljenik:
> @gabrieljenik : do you have an idea why surveyid is not in $params before the XSS fix ? I really don't understand …

Hi Dennis. Sorry. I don't get it. Can you elaborate a bit more?

I think `$params['surveyid']` is not set as this is not set by the plugin helper. As a side note, this may not be part of a plugin helper request.

Shnoulle:
> > @gabrieljenik : do you have an idea why surveyid is not in $params before the XSS fix ? I really don't understand …
> 
> Hi Dennis. Sorry. I don't get it. Can you elaborate a bit more?
> 
> I think `$params['surveyid']` is not set as this is not set by the plugin helper. As a side note, this may not be part of a plugin helper request.

If you follow code : finally $param came from this function [public function getActionParams()](https://github.com/LimeSurvey/LimeSurvey/blob/73bc85298a5237e8f46577e1d9ef7e3ca848a170/framework/web/CController.php#L323)

and this function is `return $_GET;`

Then if sureyid are in url : it was in $params too … no ?

Then i don't understand how https://github.com/LimeSurvey/LimeSurvey/commit/de7707d700d1304110eca1e12fd22b3aa1d011b7 can broke this ?

Shnoulle:
> Before [de7707d](https://github.com/LimeSurvey/LimeSurvey/commit/de7707d700d1304110eca1e12fd22b3aa1d011b7)
> 
>     * addPseudoParams only replicated (cloned, propogated) values from one key in the `params` array to other keys.

Yes, but still $params is filled in runWithParams parameters , then here : https://github.com/LimeSurvey/LimeSurvey/blob/4d9e77960a1a25436678c50a148896fd84e1364d/framework/yiilite.php#L3720 

Filled by $this->getActionParams() === retuirn $_GET[]

It's not important, it just to understand … why …

gabrieljenik:
Nothing from that part of the code gets to `fullpagewrapper()` and its call to `_renderWrappedTemplate`

PR #1859, opened 2021-04-29
Fixed issue #17256: No warning about adjusted question codes is shown when importing a group which holds questions with already used question codes

gabrieljenik:


Shnoulle:
If i don't make error : when import old survey : we have the complete list of update. It's great

application/views/questionGroupsAdministration/import_view.php
```diff
@@ -37,6 +37,20 @@
...
+                    <!-- Warnings -->
+                    <?php if (count($aImportResults['importwarnings'])>0): ?>
+                        <h2 class="warning"><?php eT("Warnings");?>:</h2>
+                        <ul  class="list-unstyled">
+                            <?php
+                                foreach ($aImportResults['importwarnings'] as $warning)
+                                { ?>
+                                <li><?php echo $warning; ?></li>
+                                <?php
+                            } ?>
```
olleharstedt, line 50: endforeach; instead? You're using endif; below already. :)

application/helpers/admin/import_helper.php
```diff
@@ -587,6 +587,10 @@ function XMLImportGroup($sFullFilePath, $iNewSID, $bTranslateLinksFields)
     LimeExpressionManager::RevertUpgradeConditionsToRelevance($iNewSID);
     LimeExpressionManager::UpgradeConditionsToRelevance($iNewSID);
 
+    if (count($aQuestionCodeReplacements)) {
+        array_unshift($results['importwarnings'], "<span class='warningtitle'>" . gT('Attention: Several question codes were updated. Please check these carefully as the update  may not be perfect with customized expressions.') . '</span>');
```
olleharstedt, line 591: Feel free to split call into multiple lines.

```php
array_unshift(
    $results['importwarnings'],
    "<span class='warningtitle'>"
    . gT('Attention: Several question codes were updated. Please check these carefully as the update  may not be perfect with customized expressions.')
    . '</span>'
);
```

olleharstedt (changes requested):

application/views/questionGroupsAdministration/import_view.php
gabrieljenik: Yes, we were just reusing the same style which was on the file :)

Shnoulle:
Not possible to add the list of updated code ?

See with old lss : 

Capture d’écran_2021-05-03_15-23-05

PR #1861, opened 2021-04-29
Fixed issue #16834: When importing a theme bigger in size than the allowed PHP.ini settings, there is no proper description of the error

gabrieljenik:
Created UploadValidator service helper.

Fix applied to:
- Theme import (Themes list and Theme editor)
- File upload in Theme editor (uploading files to a theme)
- Image upload (in global theme options)
- Image upload (in survey theme options
- CPDB (Import CSV)
- Labelset import
- Token import
- Question import (had to add 'action' and 'sid' to the URL)

There are few which could be pending

These are pending to be fixed:
- UploaderController
- questiongroups
- surveyadmin - Import survey - Need to split the method (used for copy and import), or move the 'action' to the URL.
- frontend_helper, checkUploadedFileValidity ?
- dataentry - Import VV file (need to split the method or assume an upload action when the request is POST)

### DIFF
--- a/application/models/services/UploadValidator.php
+++ b/application/models/services/UploadValidator.php
@@ -10,9 +10,9 @@ class UploadValidator
      * @param string $fileName the name of the posted file
      * @param mixed $customMaxSize maximum file upload size
      *
-     * @throws \Exception if the file is too large or no file is found.
+     * @return string|null the error message or null if all checks are ok
      */
-    public function checkUploadedFileSize($fileName, $customMaxSize = null)
+    public function getError($fileName, $customMaxSize = null)
     {
         if (is_null($customMaxSize)) {
             $maximumSize = getMaximumFileUploadSize();
@@ -24,26 +24,22 @@ class UploadValidator
         // There is no way to confirm if the superglobals are empty because 'post_max_size' was
         // exceeded, or because nothing was posted.
         if (empty($_POST) && empty($_FILES)) {
-            throw new \Exception(
-                sprintf(
-                    gT("No file was uploaded or the request exceeded %01.2f MB."),
-                    convertPHPSizeToBytes(ini_get('post_max_size')) / 1024 / 1024
-                )
+            return sprintf(
+                gT("No file was uploaded or the request exceeded %01.2f MB."),
+                convertPHPSizeToBytes(ini_get('post_max_size')) / 1024 / 1024
             );
         }
 
         if (!isset($_FILES[$fileName])) {
-            throw new \Exception(gT("File not found."));
+            return gT("File not found.");
         }
 
         $fileSize = $_FILES[$fileName]['size'];
 
         if ($fileSize > $maximumSize || $_FILES[$fileName]['error'] == 1 || $_FILES[$fileName]['error'] == 2) {
-            throw new \Exception(
-                sprintf(
-                    gT("Sorry, this file is too large. Only files up to %01.2f MB are allowed."),
-                    $maximumSize / 1024 / 1024
-                )
+            return sprintf(
+                gT("Sorry, this file is too large. Only files up to %01.2f MB are allowed."),
+                $maximumSize / 1024 / 1024
             );
         }
     }
@@ -55,12 +51,11 @@ class UploadValidator
      * @param mixed $redirectUrl the URL to redirect on failure
      * @param mixed $customMaxSize maximum file upload size
      */
-    public function checkUploadedFileSizeAndRedirect($fileName, $redirectUrl, $customMaxSize = null)
+    public function redirectOnError($fileName, $redirectUrl, $customMaxSize = null)
     {
-        try {
-            $this->checkUploadedFileSize($fileName, $customMaxSize);
-        } catch (\Exception $ex) {
-            \Yii::app()->setFlashMessage($ex->getMessage(), 'error');
+        $error = $this->getError($fileName, $customMaxSize);
+        if (!is_null($error)) {
+            \Yii::app()->setFlashMessage($error, 'error');
             \Yii::app()->getController()->redirect($redirectUrl);
         }
     }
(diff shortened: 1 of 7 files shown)


Shnoulle:
Related dev feature https://github.com/LimeSurvey/LimeSurvey/pull/1248 

I think we can check baseDir in the new Class !

Thank !

application/models/services/UploadValidator.php
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Check uploaded file size. Redirects to the specified URL on failure.
+     *
+     * @param string $fileName the name of the posted file
+     * @param mixed $redirectUrl the URL to redirect on failure
+     * @param mixed $customMaxSize maximum file upload size
+     */
+    public function checkUploadedFileSizeAndRedirect($fileName, $redirectUrl, $customMaxSize = null)
```
olleharstedt, line 58: Can you try to avoid using "and" in function names, but rather use function composition?

1) checkUploadedFileSize - return string, the error message

2) Compose functions like `redirect(checkUploadedFileSize());` or `$msg = checkUploadedFileSize(); redirect($msg);` Redirect can do thing if checkUploadedFileSize() returns null. Maybe rename to `redirectIfNotNull()`?

3) Same with JSON

olleharstedt (changes requested):

application/models/services/UploadValidator.php
```diff
@@ -0,0 +1,82 @@
...
+        if (is_null($customMaxSize)) {
+            $maximumSize = getMaximumFileUploadSize();
+        } else {
+            $maximumSize = min((int) $customMaxSize, getMaximumFileUploadSize());
+        }
+
+        // When 'post_max_size' is exceeded $_POST and $_FILES are empty.
+        // There is no way to confirm if the superglobals are empty because 'post_max_size' was
+        // exceeded, or because nothing was posted.
+        if (empty($_POST) && empty($_FILES)) {
```
olleharstedt, line 26: Please inject $_POST and $_FILES instead of accessing them directly. E.g., you can add two attributes to the class, `private $post; private $files` and inject them in the constructor. This will make it possible to unit-test the function later.

olleharstedt (changes requested):

application/models/services/UploadValidator.php
gabrieljenik: Check now please

application/models/services/UploadValidator.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * UploadValidator constructor.
+     *
+     * @param array<string,mixed>|null  $post   HTTP POST variables. If null, $_POST is used.
+     * @param array<string,mixed>|null  $post   HTTP File Upload variables. If null, $_FILES is used.
+     */
+    public function __construct($post = null, $files = null)
+    {
+        $this->post = is_null($post) ? $_POST : $post;
```
olleharstedt, line 22: Please remove this shortcut. ;) The dependency on super-globals should be explicit.

olleharstedt: Same with line below.

olleharstedt (changes requested):

olleharstedt:
For Yii 3, we will pass around request objects instead, based on PSR.

application/models/services/UploadValidator.php
olleharstedt: Ease of reading > ease of use.

olleharstedt: Let me read again

olleharstedt: Maybe you want to inject this instead?
```Yii::app()->request```

gabrieljenik: I like it

Shnoulle: > Maybe you want to inject this instead?
> `Yii::app()->request`

I'm not sure you can get $_POST array via `Yii::app()->request` …

Shnoulle: No : https://www.yiiframework.com/doc/api/1.1/CHttpRequest#getPost-detail

`$name` param is mandatory

Shnoulle: PS : I already check if yii have something ;) .

And same : Yii (1) don't have setPost or setParam :)

olleharstedt: Good! $_POST and $_GET should be read-only. ^^

olleharstedt: Yes, imagine when we have unit-tests for all those functions, Denis. Phew. Champagne!

olleharstedt: > How ! Need to hire hundred of developer ! rofl

Each big method could be a separate class (command object pattern). We inject all dependencies. Then we take it one method at a time.

gabrieljenik: > @gabrieljenik Maybe I can just merge and apply the changes I wanted. ^^

Go for it! Sure!

c-schmitz:
Ehm... this seems to have been implemented already in dev branch a month ago in #1783  ?

@gabrieljenik Gabriel, can you check that please?

olleharstedt:
Crap, you're right. Now there are merge conflicts from master into dev.

gabrieljenik:
> Ehm... this seems to have been implemented already in dev branch a month ago in #1783 ?
> 
> @gabrieljenik Gabriel, can you check that please and merge it one way or the other to /develop?
> 
> I am wondering if work was done here twice..... please give feedback.

I will review.
I don't think the work has been done twice. I would recall it I guess. I am old, but not that much yet :)

@olleharstedt I see a commit. Anything left to do?

olleharstedt:
No, I'll fix it.

PR #1862, opened 2021-04-30
Dev: avoid bad helper load

Shnoulle:
Dev: sometimes need to be sure a class is loaded, Yii does the job, let hime do the job

### DIFF
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -195,6 +195,10 @@ class LSYii_Application extends CWebApplication
      */
     public function loadHelper($helper)
     {
+        if (strpos($helper, "/")) {
+            Yii::log("Invalid helper string " . $helper, \CLogger::LEVEL_WARNING, 'application.loadHelper');
+            $helper = str_replace("/", ".", $helper);
+        }
         Yii::import('application.helpers.'.$helper.'_helper', true);
     }
 
@@ -207,6 +211,10 @@ class LSYii_Application extends CWebApplication
      */
     public function loadLibrary($library)
     {
+        if (strpos($library, "/")) {
+            Yii::log("Invalid helper string " . $library, \CLogger::LEVEL_WARNING, 'application.loadLibrary');
+            $library = str_replace("/", ".", $loadLibrary);
+        }
         Yii::import('application.libraries.'.$library, true);
     }
 


olleharstedt:
Better: Throw exception if slash is found?

olleharstedt:
Actually, could be an assert (turned off on production). And then fix all cases in the code which is wrong.

olleharstedt:
Wait, this is LS3?? Nah, come on.

Shnoulle:
> Better: Throw exception if slash is found?

Then : need to fix whole current error ? 
Why not .

DEBUG + forcedsuperadmin : throw exception

Shnoulle:
> Wait, this is LS3?? Nah, come on.

😎 🕺

Shnoulle:
> Wait, this is LS3?? Nah, come on.

Reason : since you remove it, maybe it's needed in some specific view :D

olleharstedt:
It might be. If you want to fix it you can revert my fix and then replace the '/' with '.'. And also test quick translation after.

olleharstedt:
> DEBUG + forcedsuperadmin : throw exception

`assert()` is even better, because it's not even executed in production mode (in PHP 7+).

olleharstedt:
But we would have to configure assert_options etc, might be too much. https://www.php.net/manual/en/function.assert.php

Shnoulle:
Maybe some user with external hoster want to help …

But : why not : to be in manual ?

olleharstedt:
> Maybe some user with external hoster want to help …

debug = 2 should enable all assertions in the system.

olleharstedt:
> But : why not : to be in manual ?

Which part? :thinking:

Shnoulle:
> > Maybe some user with external hoster want to help …
> 
> debug = 2 should enable all assertions in the system.

It's the case currently ?


> > But : why not : to be in manual ?
> 
> Which part? thinking

Use assert :)

olleharstedt:
> It's the case currently ?

No, it's not. No logic added for assert() yet. It's in the new code quality guideline as a way to check internal invariants. Have to check the Yii 3 approach too, if they did something.

olleharstedt:
https://manual.limesurvey.org/Code_quality_guide#Assertions

olleharstedt:
But this will be added to LS4, not LS3. :joy:

Shnoulle:
:)

PR #1864, opened 2021-05-04
Dev #T638b: After delete group behaviour

gabrieljenik:
- When a user DELETES a question group, He should stay in STRUCTRURE and not be navigated back to SETTINGS.
- The button DELETE CURRENT GROUP should be DELETE GROUP to keep the design consistent
- Don't allow users to delete the last group in the survey.


Shnoulle:
> Don't allow users to delete the last group in the survey.

Why not ? If you allow to delete non empty group : you must allow to delete last one too (in my opinion).

olleharstedt:
> Why not ? If you allow to delete non empty group : you must allow to delete last one too (in my opinion).

I think the reason is you always need at least one group in a survey to work.

Shnoulle:
Yes, but it's not a good reason.

If you want survey still work : you must check this group have questions.

Here : if you want to delete last group : you have to create a group before. It's one step more with clean reason.

It's not understandable in my opinion.

olleharstedt:
Good point.

gabrieljenik:
I believe is a different issue and a more general one with the side bar.
Example: When you add a group, the new group is NOT expanded NEITHER highlighted in dark green. 
The previous shown group is still expanded and shown on light green

assets/packages/adminsidepanel/build/css/adminsidepanel.rtl.css
```diff
@@ -362,181 +384,220 @@ ol.breadcrumb.title-bar-breadcrumb {
...
+          flex: 1; }
   .ls-flex-item .grow-1 {
-    flex-grow: 1; }
+    -webkit-box-flex: 1;
+            flex-grow: 1; }
   .ls-flex-item .col-1 {
     width: 8.33333%; }
   .ls-flex-item .grow-2 {
-    flex-grow: 2; }
+    -webkit-box-flex: 2;
```
Shnoulle, line 534: Not related to the commit or pull request

? Why `-webkit` here ? Outdated dependance ?

https://caniuse.com/mdn-css_properties_flex-grow

olleharstedt:
Tested ok! Can you please fix the conflict?

PR #1877, opened 2021-05-13
New feature #17226: Ability to export to SPSS .sav format

adamzammit:
New feature #17226

This is a new branch to replace the pull request: https://github.com/LimeSurvey/LimeSurvey/pull/1834

It has been updated to pull dependencies using composer instead of supplying the files as part of the plugin.


olleharstedt:
Hi ho. What's the reason for the last switch here? `Dev: composer update --no-dev --ignore-platform-reqs`

864 files changed is a bit much... Twig is also updated?

adamzammit:
I'm not sure why twig updated but for me it seemed to create a new folder with lowercase "twig" - is there something on your end making it "Twig" that isn't reflected in the source?

The extra switch --ignore-platform-reqs is because tiamo/spss requires bcmath php extension, which is provided by  phpseclib/bcmath_compat  if it isn't installed, but this doesn't resolve the platform dependency for composer which will fail without that switch.

olleharstedt:
> I'm not sure why twig updated but for me it seemed to create a new folder with lowercase "twig" - is there something on your end making it "Twig" that isn't reflected in the source?

Yeah, we had a problem with upper-case Twig folder, but fixed it recently in another branch. Gah. Have to check it again.

olleharstedt:
OK, only 400 files changed after merged with upstream. :)

olleharstedt:
@adamzammit If I now run `composer update --no-dev`, will it mess up the composer dependencies?

adamzammit:
No but it might fail if your environment doesn't have the bcmath PHP extension. If it does fail - run with:  --ignore-platform-reqs

olleharstedt:
Question is which command to run and add to the repo...

PR #1878, opened 2021-05-13
Dev #T635: Land in the Suvey Builder (Structure) when creating a survey

gabrieljenik:
1. When a user creates a survey he should always land on the first question of the first question group in the STRUCTURE tab. If no question group exists a default one (the demo one) is created, and he lands on that first question group.
2. By doing so we don't need the option in create survey: "Create example question group and question?". Get rid of that option.

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -82,6 +82,7 @@ $config['allowunblacklist']          = 'N'; // Allow participant to unblacklist 
 $config['userideditable']            = 'N'; // Allow editing of user IDs
 
 $config['defaulttheme']              = 'fruity'; // This setting specifys the default theme used for the 'public list' of surveys
+$config['createsample']              = true;
 $config['customassetversionnumber']  = 1;        // Used to generate the path of tmp assets (see: LSYii_AssetManager::generatePath()  )
 
 // Please be very careful if you want to allow SVG files - there are several XSS dangerous security issues
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -486,7 +486,10 @@ class SurveyAdministrationController extends LSBaseController
             // This will force the generation of the entry for survey group
             TemplateConfiguration::checkAndcreateSurveyConfig($iNewSurveyid);
 
-            $createSample = Yii::app()->getConfig('createsample');
+            $createSample = SettingsUser::getUserSettingValue('createsample');
+            if ($createSample === null || $createSample === 'default') {
+                $createSample = Yii::app()->getConfig('createsample');
+            }
 
             // Figure out destination
             if ($createSample) {
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -486,18 +486,34 @@ class SurveyAdministrationController extends LSBaseController
             // This will force the generation of the entry for survey group
             TemplateConfiguration::checkAndcreateSurveyConfig($iNewSurveyid);
 
-            // Create sample group and question
-            $iNewGroupID = $this->createSampleGroup($iNewSurveyid);
-            $iNewQuestionID = $this->createSampleQuestion($iNewSurveyid, $iNewGroupID);
+            $createSample = Yii::app()->getConfig('createsample');
 
-            Yii::app()->setFlashMessage(gT("Your new survey was created. 
-            We also created a first question group and an example question for you."), 'info');
-            $redirecturl = $this->getSurveyAndSidemenueDirectionURL(
-                $iNewSurveyid,
-                $iNewGroupID,
-                $iNewQuestionID,
-                'structure'
-            );
+            // Figure out destination
+            if ($createSample) {
+                $iNewGroupID = $this->createSampleGroup($iNewSurveyid);
+                $iNewQuestionID = $this->createSampleQuestion($iNewSurveyid, $iNewGroupID);
+
+                Yii::app()->setFlashMessage(gT("Your new survey was created. 
+                We also created a first question group and an example question for you."), 'info');
+                $redirecturl = $this->getSurveyAndSidemenueDirectionURL(
+                    $iNewSurveyid,
+                    $iNewGroupID,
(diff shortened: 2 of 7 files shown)


Shnoulle:
> If no question group exists a default one (the demo one) is created

But why ? 

I already ask to remove the forced demo question, not to ask again to disable it …

ViliusS:
I support @Shnoulle here. The purpose of demo question/question group should be to introduce new LimeSurvey users how to create a question. Hence, demo question should be an option **and** _disabled by default_. I would hate to delete that question every time when I create a survey even though I have created hundreds of them in the past.

gabrieljenik:
This was requested by Jory Nigel. I agree with him on this one.
Having the first groups and question already cerated saves some clicks and also it is easier for newcomers.
For pro-users is it really a big difference? Instead of deleting, you just edit.

ViliusS:
Newcomers would use this functionality _once_. Pro-users would need to use workaround (edit or delete) _every time_.

Shnoulle:
> This was requested by Jory Nigel. I agree with him on this one.

2 VS 2 ;)

> Having the first groups and question already cerated saves some clicks and also it is easier for newcomers.
> For pro-users is it really a big difference? Instead of deleting, you just edit.

New user settings ?

- autocreate question and group when create a survey
- autocreate question and group when a survey is empty.

gabrieljenik:
> New user settings ?

Why? That would make things more complex, not simpler.
I mean, what is so bad about going in the first group and edit it instead of just creating the first group?

> PS : a lot pof user have a 1st group with variables, specific question etc… the 1st things they do : it's import this group.
They can still do that.

ViliusS:
> > PS : a lot pof user have a 1st group with variables, specific question etc… the 1st things they do : it's import this group.
> 
> They can still do that.

I think the question is not if they can do that, but the question of convenience. Pro users could do a lot of things, but it doesn't mean that LimeSurvey needs to put them into this position and force to do unnecessary things.

Don't get me wrong, I support making LimeSurvey more friendly for novice users, but I feel that one-time wizard or better UX in general would be a better way to achieve this.

Now you are making experience better for one group of people and making it worse for other group.

gabrieljenik:
> Don't get me wrong, I support making LimeSurvey more friendly for novice users, but I feel that one-time wizard or better UI in general would be a better way to achieve this.

We understand this. And that is the way to go.
This change was just a small adjustment.

> Now you are making experience better for one group of people and making it worse for other group.

Worse? How much worse?

ViliusS:
> > Don't get me wrong, I support making LimeSurvey more friendly for novice users, but I feel that one-time wizard or better UI in general would be a better way to achieve this.
> 
> We understand this. And that is the way to go.
> This change was just a small adjustment.

This needs to be done on a global scale. You cannot fix UX with small changes. 

> > Now you are making experience better for one group of people and making it worse for other group.
> 
> Worse? How much worse?

I can speak only for myself here, I would _hate_ to do additional steps (delete or edit). LimeSurvey configuration is enough complicated as it is. Also I _hate_ software which thinks that it knows better than me and autocompletes/autocreates/autoreboots/auto-does something on its own.

Shnoulle:
> > Now you are making experience better for one group of people and making it worse for other group.
> 
> Worse? How much worse?

3 click against 1 is worse when you do it 150 times a day …

And : since LimeSurvey admin are not really speed : 2 click more mean 5 second more …

Shnoulle:
> Worse? How much worse?

And why auto create is better ? 
For new user : understand they have to create a group (or more) to put question in it.

Else : we have sometime on forum some user using only one group : they don't know really understand they can create group.

Shnoulle:
User settings ? ;)

olleharstedt:
Who takes the decision here?

ViliusS:
> User settings ? ;)

I think global setting is fine. It could be even more general, and named like "Tutorial interface" > "On"/"Off". This one switch could enable automatic question creation, display of various hints in the interface, etc.

Shnoulle:
> > User settings ? ;)
> 
> I think global setting is fine. It could be even more general, and named like "Tutorial interface" > "On"/"Off". This one switch could enable automatic question creation, display of various hints in the interface, etc.

Yes, but for company : it's a user settings needed. Only some user need an help during a time. 

It a user choice, not a global choice.

maziminke:
Many long term users of Limesurvey will be **very** annoyed if they have to edit an auto-generated group **and** an auto-generated question every time they create a new survey. I will surely be one of them!

Please do **not** add this to the core. The benefit for new users is really minor and for experienced Limesurvey users this makes things worse. We already lost a lot of reputation, this will add to it.

Shnoulle:
@maziminke : why not like other user settings ?

1. Global settings by default (default came from DB)
2. User settings

I know HTML editor stay Inline a lot, and maybe muts be popup by default, but this allow superadmin : 

1. Choose the way for new user
2. Some admin to allow update own settings
3. Have a default default for new installation of CE or at SAAS.

gabrieljenik:
> @maziminke : why not like other user settings ?
> 
> 1. Global settings by default (default came from DB)
> 2. User settings
> 
> I know HTML editor stay Inline a lot, and maybe muts be popup by default, but this allow superadmin :
> 
> 1. Choose the way for new user
> 2. Some admin to allow update own settings
> 3. Have a default default for new installation of CE or at SAAS.

I like having this as user settings. Fair simple to implement.

gabrieljenik:
Now we have a global and a user setting for controlling whether to create sample questions. :)

olleharstedt:
Before merge - has this feature been documented?

gabrieljenik:
Actually, I have not detected anything else to update, nor not sure if there is anything to add.
Maybe we could add the new global setting to the manual, but there is no personal setting section

olleharstedt:
https://manual.limesurvey.org/First_login_-_your_user_preferences#My_personal_settings ?

gabrieljenik:
Added here:
https://manual.limesurvey.org/Global_settings#General
https://manual.limesurvey.org/First_login_-_your_user_preferences#My_personal_settings

olleharstedt:
Nice!

PR #1881, opened 2021-05-14
Bug/17301  tokens  minute lock out

gabrieljenik:
Added a whitelist config setting as to be able to whitelist certain ips or regexs either for token or login attempts.
This is a quick solution which to solve this issue. Further discussions to the blocking feature can be added on top of this later.


olleharstedt:
Can you change title to "New feature: Bla bla bla"?

c-schmitz:
We also need a PR for 4.x

gabrieljenik:
> We also need a PR for 4.x

https://github.com/LimeSurvey/LimeSurvey/pull/1886
Sorry, forgot to  send the PR

PR #1886, opened 2021-05-18
Fixed issue #17301: tokens > minute lock out

gabrieljenik:
Added a whitelist as global security settings as to be able to whitelist certain ips either for token or login attempts.

### DIFF
--- a/application/helpers/sanitize_helper.php
+++ b/application/helpers/sanitize_helper.php
@@ -504,35 +504,14 @@ function sanitize_signedint($integer, $min = '', $max = '')
 };
 
 /**
- * Sanitize IPv4 address $string, accepting octet wildcards if $allowWildcards is true.
+ * Checks the validity of IP address $ip
  *
- * @param string $string to sanitize
- * @param boolean $allowWildcards whether to accept octect wildcards as valid or not
+ * @param string $ip to check
  *
- * @return string the IP address if it's valid, or an empty string
+ * @return boolean true if the $ip is a valid IP address
  */
-function sanitize_ip_address($string, $allowWildcards = false)
+function check_ip_address($ip)
 {
-    if ($allowWildcards) {
-        $regex = '/^((([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5])|\*)\.){3}(([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5])|\*)$/';
-    } else {
-        $regex = '/^((([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5]))\.){3}(([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5]))$/';
-    }
-    return preg_match($regex, $string, $m) ? $string : '';
-}
-
-/**
- * Checks the validity of IPv4 address $input, accepting octet wildcards if $allowWildcards is true.
- *
- * @param string $input to check
- * @param boolean $allowWildcards whether to accept octect wildcards as valid or not
- *
- * @return boolean true if the $input is a valid IP address
- */
-function check_ip_address($input, $allowWildcards = false)
-{
-    if ($input != sanitize_ip_address($input, $allowWildcards)) {
-        return false;
-    }
-    return true;
+    // Leave the wrapper in case we need to enhance the checks later
+    return filter_var($ip, FILTER_VALIDATE_IP);
 }
--- a/application/helpers/sanitize_helper.php
+++ b/application/helpers/sanitize_helper.php
@@ -504,14 +504,14 @@ function sanitize_signedint($integer, $min = '', $max = '')
 };
 
 /**
- * Sanitize an IPv4 address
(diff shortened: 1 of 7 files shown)


application/models/FailedLoginAttempt.php
```diff
@@ -126,4 +137,35 @@ public function addAttempt()
...
+     *
+     * @param string $ip
+     * @param string $attemptType   'login' or 'token'
+     *
+     * @return boolean
+     */
+    private function isWhitelisted($ip, $attemptType)
+    {
+        if ($attemptType != self::TYPE_LOGIN && $attemptType != self::TYPE_TOKEN) {
+            return false;
```
olleharstedt, line 152: Either `ThrowInvalidArgument` or `assert`

olleharstedt (changes requested):

Shnoulle:
Not related to this commit : time for ipv6 ?

/me must report the feature request …

application/models/FailedLoginAttempt.php
gabrieljenik: Done

olleharstedt: Hooray!

application/helpers/sanitize_helper.php
```diff
@@ -502,3 +502,22 @@ function sanitize_signedint($integer, $min = '', $max = '')
...
+{
+    if ($allowWildcards) {
+        $regex = '/^((([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5])|\*)\.){3}(([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5])|\*)$/';
+    } else {
+        $regex = '/^((([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5]))\.){3}(([0-9]{1,2})|(1[0-9]{2,2})|(2[0-4][0-9])|(25[0-5]))$/';
+    }
+    return preg_match($regex, $string, $m) ? $string : '';
+}
+
+function check_ip_address($input, $allowWildcards = true)
```
olleharstedt, line 517: Missing docblock :( And above

olleharstedt (changes requested):

application/models/FailedLoginAttempt.php
```diff
@@ -126,4 +137,34 @@ public function addAttempt()
...
+    /**
+     * Returns true if the specified IP is whitelisted
+     *
+     * @param string $ip
+     * @param string $attemptType   'login' or 'token'
+     *
+     * @throws InvalidArgumentException if an invalid attempt type is specified
+     * @return boolean
+     */
+    private function isWhitelisted($ip, $attemptType)
```
olleharstedt, line 150: NB: This method does not contain `this`, so maybe it doesn't belong in a class...? But currently, there's no good system to put functions together with a domain entity, or "module".

application/helpers/sanitize_helper.php
olleharstedt: To be a little anal, can you maybe do this as well?

> The first line in the docblock should be one sentence describing the relation between function arguments and its output

https://manual.limesurvey.org/Code_quality_guide#Contract

olleharstedt: > Checks the validity of an IPv4 address 

would be

> Checks the validity of an IPv4 address $input

Or maybe, with renaming the argument:

> Checks the validity of IPV4 address $ip

olleharstedt: Boolean arguments that changes the "mode" of the function can be hard to fit in a one-liner like that. :)

application/models/FailedLoginAttempt.php
Shnoulle: static function ?

olleharstedt: Static functions cannot be mocked. But, sometimes they don't need to be mocked. :) (When they have no side-effects.)

This functions has one side-effect: It reads from app config. This should be lifted out if you want to unit-test this function.

Shnoulle: read App->getConfig are side effect ???

Easy test : 
```
App()->setConfig("myconfigtotest","testvalue");
/* My test #1 */
App()->setConfig("myconfigtotest","testvalue2");
/* My test #2 */
```

olleharstedt: Accessing globals are considered side-effects.

You can do setConfig, but then you have setup/teardown in the tests. Easy to forget, and tests can affect each other.

olleharstedt: Oh, and it might be hard to run tests in parallel when you rely on global scope like that.

c-schmitz (changes requested):
Please also consider the changes I requested in the 3.x PR for this issue

application/helpers/sanitize_helper.php
gabrieljenik: > Boolean arguments that changes the "mode" of the function can be hard to fit in a one-liner like that

What do you mean? Sorry! Description for that argument would be too long?

c-schmitz (approved):

PR #1887, opened 2021-05-18
Fixed issue #17309: Date question always shows error when using AR language - can't proceed at survey

gabrieljenik:
The sample survey has two languages: EN and AR.
At the survey level, the date format for EN is dd.mm.yyyy for AR is dd-mm-yyyy (dots vs. dashes).
At the question level it does not have a specific format.

In LS4:
When the question is rendered, it tries to use the format of the question, and if it does not have it, it uses the one of the survey. But use the base language.

In the data validation, if the question does not have a format, use the survey format, but with the corresponding language.

In this case they are different a dump is produced. 
The fix is considering the language on the rendering.

### DIFF
--- a/application/core/QuestionTypes/Date/RenderDate.php
+++ b/application/core/QuestionTypes/Date/RenderDate.php
@@ -398,7 +398,7 @@ class RenderDate extends QuestionBaseRenderer
     {
         $answer = '';
         $inputnames = [];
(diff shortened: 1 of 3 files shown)


c-schmitz (changes requested):
Please include also the unicode change in EM as suggested in the other pull request.

gabrieljenik:
> Please include also the unicode change in EM as suggested in the other pull request.

Done. Thanks

PR #1888, opened 2021-05-18
Fixed issue #17309: Date question always shows error when using AR language - can't proceed at survey

gabrieljenik:
The sample survey has two languages: EN and AR.
At the survey level, the date format for EN is dd.mm.yyyy for AR is dd-mm-yyyy (dots vs. dashes).
At the question level it does not have a specific format.

In LS3:
The rendering part is fine.
But in validation, it fails when converting the Arabic characters. 
Ported the rutine from LS4 method and works fine now.

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -10218,22 +10218,17 @@ report~numKids > 0~message~{name}, you said you are {age} and that you have {num
         {
             $result = $str;
 
-            $standard = array("0","1","2","3","4","5","6","7","8","9");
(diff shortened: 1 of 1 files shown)


Shnoulle:
Seems to be an error when saving the PHP files ? Someone save it in ISO and not in utf8 ?

gabrieljenik:
> Seems to be an error when saving the PHP files ? Someone save it in ISO and not in utf8 ?

Yes. For master we can use something like "\u{0669}", (requires PHP7). We can do json_encode similar for LTS.
I guess, for master would be enough. 

@olleharstedt  What do you think?

c-schmitz:
Please use for master the unicode notation as suggested.

PR #1889, opened 2021-05-19
New Feature #17301: Allow whitelisting IPs to not be locked out

gabrieljenik:
- Added whitelist ip as a global security settings
- Throw InvalidArgumentException if invalid attempt type is specified on isWhitelisted()

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -59,8 +59,8 @@ $config['adminthemeiconsize'] = 32; // This settings describes the icon size for
 // If the user enters password incorrectly
 $config['maxLoginAttempt']    = 3; // Lock them out after 3 attempts
 $config['timeOutTime']        = 60 * 10; // Lock them out for 10 minutes.
-$config['loginIpWhitelist']   = []; // Array of regular expressions for IPs that should be excluded from the max login attemps check. Ex: '192.168.0.5|192.168.0.6' or '192.168.0.\d+'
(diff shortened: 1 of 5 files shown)


c-schmitz (changes requested):
I don't think it is a good idea to use wildcard notation for IP addresses.
As you can see you need to introduce your own validation function.

Also your approach does not seem to handle IPv6.

It would be better to allow/use CIDR notation (https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) on top of normal IP addresses, see https://gist.github.com/pavinjosdev/cb1d636ea9dc2bd201d54107d10650c5 for a working function that also handles IPv6 (you might need to adjust that function a bit to also allow for single IP addresses)

In global settings when I enter an invalid IP address there is no error. It just vanishes on saving. It should show a proper error after saving.

gabrieljenik:
> I enter an invalid IP a

As per discussed with @c-schmitz:
- On this story we will allow single IPv4 addresses, no wildcards, no CIDR
- On a new story to be created we will allow CIDR and wildcards.

Will review the other changes requested and implement them

Thanks!

c-schmitz (changes requested):
validateIpAddresses is still not capable checking IPv6.
Also you should not roll your own solution but use filter_var instead.

gabrieljenik:
> validateIpAddresses is still not capable checking IPv6.
> Also you should not roll your own solution but use filter_var instead.

OK, will ammend. Thanks!

c-schmitz (approved):

c-schmitz:
Please do not forget to update the manual, too.

gabrieljenik:
> Please do not forget to update the manual, too.

Done: https://manual.limesurvey.org/Global_settings#Security

PR #1892, opened 2021-05-26
Fixed issue #17306: Some update theme can broke page in 3.26.4

gabrieljenik:
Added the "tag" method as allowed on the config internal.


olleharstedt:
Thanks. I cherry-picked this into master too.

PR #1893, opened 2021-05-26
Dev #17261: Create new code artifacts to handle QuestionAttributes fetching (& saving)

gabrieljenik:


application/models/Question.php
```diff
@@ -1525,4 +1526,15 @@ public function getScaledSubquestions()
         }
         return $results;
     }
+
+    /**
+     * Get the question theme name
+     *
+     * @return string
```
olleharstedt, line 1533: Myeh. This should probably return question theme object, no?

Later it should be possible to do `$question->question_theme`, too. But impossible now, since there's no database column for it. :(

application/models/services/CoreQuestionAttributeProvider.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        /** @var array<string,array> An array of question attributes */
+        $attributes = [];
+
+        if (file_exists($xmlFilePath)) {
+            // load xml file
+            if (\PHP_VERSION_ID < 80000) {
+                libxml_disable_entity_loader(false);
+            }
+            $xmlConfig = simplexml_load_file($xmlFilePath);
```
olleharstedt, line 71: How about factor out the simplexml_load_file to a separate method, together with disable entity loader and json decode/encode thingy? Some code duplication present now.

application/models/services/QuestionAttributeFetcher.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /** @var array<string,mixed> array of options to pass to the providers */
+    private $options = [];
+
+    /** @var array<QuestionAttributeProvider> array of question attribute providers */
+    private $providers = [];
+
+    public function __construct()
+    {
+        $this->providers = [
```
olleharstedt, line 22: Should be injected in constructor as arguments? Then they can be mocked.

application/models/services/QuestionAttributeFetcher.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Returns the question attribute definitions according to the specified filters,
+     * from all available sources.
+     *
+     * @return array<string,array> array of question attribute definitions
+     */
+    public function fetch()
+    {
+        if (empty($this->question)) {
```
olleharstedt, line 37: Valid case? Or exception/assert failure?

application/models/services/ThemeQuestionAttributeProvider.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        /** @var array<string,array> An array of question attributes */
+        $attributes = array();
+
+        if (\PHP_VERSION_ID < 80000) {
+            libxml_disable_entity_loader(false);
+        }
+        $questionTheme = \QuestionTheme::model()->findByAttributes([], 'name = :name AND extends = :extends', ['name' => $questionThemeName, 'extends' => $questionType]);
+        if ($questionTheme !== null) {
+            $xmlConfig = simplexml_load_file(App()->getConfig('rootdir') . '/' . $questionTheme['xml_path'] . '/config.xml');
```
olleharstedt, line 47: Same thing here with code duplication.

olleharstedt (changes requested):

olleharstedt:
Nice work :)

application/models/Question.php
gabrieljenik: Yes,  having the question_theme as attribute is important. Maybe we can add a story for that?
We can always update the methods later (or add new ones).

application/models/services/CoreQuestionAttributeProvider.php
gabrieljenik: Do you mean creating something like an xml reader service?

application/models/services/QuestionAttributeFetcher.php
gabrieljenik: You mean an optional argument as so they can be overwritter, right?

application/models/services/CoreQuestionAttributeProvider.php
gabrieljenik: I agree, but maybe on another story? I just want to wrap this. The later we merge, more chances of conflict. :)

application/models/Question.php
olleharstedt: I'll add it.

application/models/services/CoreQuestionAttributeProvider.php
olleharstedt: > Do you mean creating something like an xml reader service?

Just a simple function wrapping those lines would be enough.

olleharstedt: > I just want to wrap this

I've heard that before, too many times (not from you). :) The end result: On to the next thing, and the next thing, and the next...

application/models/services/QuestionAttributeFetcher.php
olleharstedt: I mean this: https://en.wikipedia.org/wiki/Dependency_injection#Constructor_injection

application/models/services/QuestionAttributeFetcher.php
olleharstedt: So you can throw InvalidArgumentException, or, if missing $this->question always is a bug, you can use assert(!empty($this->question), "question attribute should not be empty);

application/models/Question.php
olleharstedt: Still, can you change this to getQuestionThemeName? To be clear it's not a domain object you get.

application/models/services/CoreQuestionAttributeProvider.php
olleharstedt: Wait, I'm retarded. I only wrote this some year ago: https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/ExtensionConfig.php#L149

application/models/ExtensionConfig.php
```diff
@@ -198,4 +198,19 @@ public function createVersionFetchers()
...
         return $fetchers;
     }
+
+    /**
+     * Returns the $nodeName XML node as an array
+     *
+     * @param string $nodeName the name of the node to retrieve
+     * @return array<mixed> the node contents as an array
+     */
+    public function getNodeAsArray($nodeName)
```
olleharstedt, line 208: Thanks Gabriel, sorry for wasting your time there.

application/models/services/CoreQuestionAttributeProvider.php
gabrieljenik: Ready for your re-review

application/models/services/QuestionAttributeFetcher.php
gabrieljenik: Ready for review

olleharstedt:
Pushed this to dev instead. Thank you!

PR #1894, opened 2021-05-26
Dev #17301: remove unneeded configs

gabrieljenik:


Shnoulle:
Why ? It's the default when install. In my opinion it's more clear to have it here.

Like (near) all other one

Shnoulle:
> Well, it is not needed anymore as there is a config screen now.

Yes : like 

- https://github.com/LimeSurvey/LimeSurvey/blob/edad104a866891ab07f1efd5f6fdf59b03cda756/application/config/config-defaults.php#L72
- https://github.com/LimeSurvey/LimeSurvey/blob/edad104a866891ab07f1efd5f6fdf59b03cda756/application/config/config-defaults.php#L417
- https://github.com/LimeSurvey/LimeSurvey/blob/edad104a866891ab07f1efd5f6fdf59b03cda756/application/config/config-defaults.php#L418
- etc 
Other false/empty config with GUI :) 

I just mean : seems clearer to have it :)

But : looking at the list seems we must remove some …

c-schmitz:
I agree that some settings could be removed..

IMHO Config-default is only for:

- Settings for which you want to set a different default for all administration users
- Settings that cannot/should not be changed using the UI (like sessions)

Shnoulle:
About to be removed : i mean outdated config :)

- https://github.com/LimeSurvey/LimeSurvey/blob/0c2d6ad3d3614ab66612dd15b6574cf38cf128ff/application/config/config-defaults.php#L294 
- https://github.com/LimeSurvey/LimeSurvey/blob/0c2d6ad3d3614ab66612dd15b6574cf38cf128ff/application/config/config-defaults.php#L312

for example

PR #1896, opened 2021-05-26
Fixed issue #14039: Captcha prevent panel integration. GET URL parameter not captured

gabrieljenik:
Added querystring parameters to the captcha form action, so they are included in the next request.
That way, prefill values are forwarded.

### DIFF
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -2219,6 +2219,8 @@ function cookieConsentLocalization()
 /**
  * Returns an array of URL parameters that can be forwarded
  *
+ * @param LSHttpRequest $request the HTTP request
+ *
  * @return array<string,mixed>
  */
 function getForwardParameters($request)
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -884,7 +884,7 @@ function prefillFromCommandLine($surveyid)
     } else {
         $startingValues = $_SESSION['survey_'.$surveyid]['startingValues'];
     }
-    if (Yii::app()->getRequest()->getRequestType()=='GET') {
+    if (in_array(Yii::app()->getRequest()->getRequestType(), ['GET', 'POST'])) {
         $getValues = array_diff_key($_GET,array_combine($reservedGetValues, $reservedGetValues));
         if(!empty($getValues)) {
             $qcode2sgqa = array();
@@ -1320,7 +1320,7 @@ function renderRenderWayForm($renderWay, array $scenarios, $sTemplateViewPath, $
             // Rendering layout_user_forms.twig
             $thissurvey                     = $oSurvey->attributes;
             $thissurvey["aForm"]            = $aForm;
-            $thissurvey['surveyUrl']        = App()->createUrl("/survey/index", array("sid"=>$surveyid));
+            $thissurvey['surveyUrl']        = App()->createUrl("/survey/index", array_merge(["sid"=>$surveyid], getForwardParameters()));
             $thissurvey['include_content']  = 'userforms';
 
             Yii::app()->clientScript->registerScriptFile(Yii::app()->getConfig("generalscripts").'nojs.js', CClientScript::POS_HEAD);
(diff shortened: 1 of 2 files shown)


application/helpers/frontend_helper.php
```diff
@@ -2214,3 +2214,29 @@ function cookieConsentLocalization()
         gT('Please be patient until you are forwarded to the final URL.')
     );
 }
+
+/**
+ * Returns an array of URL parameters that can be forwarded
+ *
+ * @return array<string,mixed>
+ */
+function getForwardParameters()
```
olleharstedt, line 2223: Hm, this always returns `[]` if it's not GET call?

gabrieljenik: Check now please.

application/helpers/frontend_helper.php
```diff
@@ -2214,3 +2214,29 @@ function cookieConsentLocalization()
...
+        'qid',
+        'lang',
+        'newtest',
+        'action',
+        'seed',
+        'index.php',
+    );
+
+    $parameters = [];
+    if (in_array(Yii::app()->getRequest()->getRequestType(), ['GET', 'POST'])) {
```
olleharstedt, line 2238: Gonna be a bit anal here and ask you to supply `$request` as input argument to the function. This will make it testable. Also replace $_GET below with $request->getQueryParams()... Or not! Since Yii 1 does not have that function. D:

olleharstedt (changes requested):

application/helpers/frontend_helper.php
olleharstedt: https://github.com/yiisoft/yii2/blob/master/framework/web/Request.php#L659

olleharstedt: Not reading globals is part of this guideline: https://manual.limesurvey.org/Code_quality_guide#Push_side-effects_up_in_the_stack_trace

Shnoulle: > https://github.com/yiisoft/yii2/blob/master/framework/web/Request.php#L659

yii**2** none in Yii 1, must be created.

olleharstedt: Why, @Shnoulle?

olleharstedt: Maybe we should ask how it looks in Yii 3 instead :d

Shnoulle: > Why, @Shnoulle?

? I don't understand the question. We use Yii**1**, if need a function not in Yii1 : we need to create it.

olleharstedt: > We use Yii**1**, if need a function not in Yii1 : we need to create it.

Yes. And in this case, it already existed in Yii 2. :) So we can just copy it?

olleharstedt: Oh wait, I misunderstood. I think.

olleharstedt: I know I linked to Yii 2.

> Since Yii 1 does not have that function

olleharstedt: getQueryParams is part of PSR, so it's future proof for Yii 3. @gabrieljenik Can you add that method to the LS request object, and the refactor this function to use it instead of accessing $_GET?

olleharstedt (changes requested):

application/helpers/frontend_helper.php
Shnoulle: Great idea to copy/paste some code from Yii3

gabrieljenik:
Ready for reviewal

application/helpers/frontend_helper.php
```diff
@@ -2214,3 +2215,30 @@ function cookieConsentLocalization()
...
+        'gid',
+        'qid',
+        'lang',
+        'newtest',
+        'action',
+        'seed',
+        'index.php',
+    );
+
+    $request = Yii::app()->getRequest();
```
olleharstedt, line 2238: Can you move this as an argument to the function? Then we're gold!

gabrieljenik: Ready for reviewal

application/helpers/frontend_helper.php
```diff
@@ -2214,3 +2215,29 @@ function cookieConsentLocalization()
         gT('Please be patient until you are forwarded to the final URL.')
     );
 }
+
+/**
+ * Returns an array of URL parameters that can be forwarded
+ *
```
olleharstedt, line 2221: Missing @param LSRequest (or whatever) $request

olleharstedt (changes requested):

PR #1897, opened 2021-05-27
Dev #T656 : As a user i dont understand where i can enter the details of my subquestion when i select an array question

gabrieljenik:
Review question editor screen organization


Shnoulle:
1. I think quetsin type must be just after or before question code
2. Width is not best : on a 1920 screen : code of answer are not shown
3. With a big list , always need to move down + move top, and it can be really annoying
Peek 29-05-2021 20-13




But more clear than current.

For with maybe use https://getbootstrap.com/docs/3.3/javascript/#collapse-example-accordion ?

1. Question text + type + code + relevance etc …
2. Sub question Axys Y (if exist)
3. Sub question Axys X (if exist)
4. Answers  (if exist)
5. Other advanced settings

application/controllers/QuestionAdministrationController.php
```diff
@@ -1730,6 +1730,42 @@ public function actionGetAdvancedSettingsHTML(int $surveyId, string $questionTyp
...
+     * Called with Ajax after question type is selected.
+     *
+     * @param int $surveyId
+     * @param string $questionType One-char string
+     * @param int $questionId Null or 0 if new question is being created.
+     * @return void
+     */
+    public function actionGetExtraOptionsHTML(int $surveyId, string $questionType, $questionId = null)
+    {
+        if (empty($questionType)) {
```
olleharstedt, line 1744: I _think_ this check might be superfluous? Since Yii checks the arguments to the action in the request handling.

application/controllers/QuestionAdministrationController.php
```diff
@@ -1730,6 +1730,42 @@ public function actionGetAdvancedSettingsHTML(int $surveyId, string $questionTyp
...
+            // NB: Could happen if user manipulates request.
+            if (!Permission::model()->hasSurveyPermission($question->sid, 'surveycontent', 'update')) {
+                throw new CHttpException(403, gT('No permission'));
+            }
+        }
+        $this->renderPartial(
+            "extraOptions",
+            [
+                'question'         => $question,
+                'oSurvey'           => $question->survey,
```
olleharstedt, line 1764: If you're touching it anyway, maybe change `oSurvey` to `survey`?

olleharstedt:
Is it by design that question type is at the bottom?

olleharstedt:
Another alternative would be to put "Subquestions" and "Answer options" tabs next to "Question" and "Help". (Can't upload image to stupid GitHub right now.)

Shnoulle:
> 
> Is it by design that question type is at the bottom?

For this oen : i really think it muts be at top, before or after title

Shnoulle:
> Another alternative would be to put "Subquestions" and "Answer options" tabs next to "Question" and "Help".

This don't fix the width, at viewport to 1858px : i think we must show 5 characters at subquestioon title. In 3.X : 77px by default (with the left menu)

20px with this commit …

Capture d’écran_2021-05-31_16-19-19
Capture d’écran_2021-05-31_16-20-22

olleharstedt:
I might prefer something like this:

image

Shnoulle:
> I might prefer something like this. Will discuss with Jory too.
> 

Yes but just need some place for  : mandatory + relevance + group.

olleharstedt:
> mandatory + relevance + group.

"General" tab in settings tabs? Or should be more visible? It can be opened by default, maybe.

Shnoulle:
Ask to Jory or to @maziminke : i think you update more Expression (relevance) or mandatory than text (personnaly : more relevance than text sure at 100%).

Maybe a settings / 1st open by default , but unsure here. 

Else : need script too (a tab after help).

And : when you save (withiut close) : best is to open the current tab (but not when you goes to another question)

olleharstedt:
Save without close is an Ajax call now (except first time).

Shnoulle:
> Save without close is an Ajax call now (except first time).

And update only the current tab ❤️ ?

olleharstedt:
> > Save without close is an Ajax call now (except first time).
> 
> And update only the current tab heart ?

Nope. It's stupid, just collects everything in the <form> and posts it.

Shnoulle:
> Nope. It's stupid, just collects everything in the `<form>` and posts it.

And then : why doing ajax ? Only for the menu in fact ?
And if you reload all form : you need to get to current tab.  
Edit help, save : see help.


> Next step would be edit-in-place instead, I think. Option to always save when leaving focus, and little checkmark to tell you success.

Yes : but some data are related to other one (all settings for example)

olleharstedt:
> And then : why doing ajax ? Only for the menu in fact ?

Because it's faster, obviously.

Shnoulle:
For information : i have to use 2.6lts sometime, and : it's really quick than 3.X or 5.X …

olleharstedt:
> For information : i have to use 2.6lts sometime, and : it's really quick than 3.X or 5.X …

Yeah, it's sad. I'm just happy we got rid of the slowest parts in LS4...

application/controllers/QuestionAdministrationController.php
gabrieljenik: You are saying Yii check's for empty question types?

This was taken from actionGetExtraOptionsHTML.
This was supposed to be a quick win stuff, do you want to review this and update it now?

olleharstedt: Quick win is illegal if it adds technical debt. ;)

olleharstedt: > You are saying Yii check's for empty question types?

I'm saying Yii is checking that the arguments are correct to the controller action. Try it.

olleharstedt: Since the type is not nullable, I mean.

olleharstedt: Oh, but empty checks for empty string too, perhaps... Gah.

gabrieljenik: I would just leave it like this.

> Quick win is illegal if it adds technical debt. ;)

I don't think we are doing that, but won't go into that discussion :) I understand what you mean.

gabrieljenik: image
This is like when calling it with empty questionType (like this http://localhost:8000/limesurvey/index.php/questionAdministration/getExtraOptionsHTML?surveyId=586669&questionId=5700&YII_CSRF_TOKEN=MGxBVTZxaF9tZ3FNdm5fTmJZSXhmQ3BnRnE2eTQyaGS_eBGxzqzMo3r17jyEXq_KS19jJTj37hscsov6Tu9Qbg%3D%3D&questionType=)

Shnoulle: What happen if questionType is a string but no related questionType ? 

We know it's a string (can be empty) but we need to check if it's not a know questiontype (and if not : it's a 404 ?)

Shnoulle: Must fix diffarently : /questionAdministration/getExtraOptionsHTML?surveyId=753953&questionId=7865&questionType=AA

Capture d’écran_2021-06-01_14-37-24

empty page with debug = 0

questionAdministration/getGeneralSettingsHTML?surveyId=753953&questionId=7865&&questionType=EE&questionTheme=bootstrap_buttons
Capture d’écran_2021-06-01_14-39-04

application/controllers/QuestionAdministrationController.php
gabrieljenik: Done

application/controllers/QuestionAdministrationController.php
olleharstedt: Stop hacking the URL, Denis. ><

olleharstedt: > So I guess we need it

Yep, I was wrong.

Shnoulle: We need more, not only empty :) 

> Stop hacking the URL, Denis. ><

1. Never trust user
2. JS always broke

olleharstedt:
Save Dual Scale Answer Options (ls\tests\SaveDualScaleAnswerOptions)

This test fails, btw. As soon as you move stuff around, some tests might need to be updated.

Shnoulle:
Idea : based on the last version 

Move "Settings" at the same place than sub questin adn answers.
Capture d’écran_2021-06-02_09-47-44


We have

- **Edit question** : code + type + language dropdown 
- 1st tabs with : Question + help + script (
- 2 tabs with "Settings" [+ Sub questiion[scale X]] [+ sub question Y] + [Answer [scale 1]]+ [Answer scale 2]
    - Settings tabs Accordion open by default to general

@maziminke @tpartner @tammoterhark ? Your idea here ?

tpartner:
Yes, that might make sense giving more room for the editor(s). But, I think we really need a UI expert.

Shnoulle:
> Yes, that might make sense giving more room for the editor(s). But, I think we really need a UI expert.

perfect world :)

maziminke:
> Yes, that might make sense giving more room for the editor(s). But, I think we really need a UI expert.

I absolutely second this!

Wasn't there once an UI expet hired who started working on the website instead if the LS interface?

I also wonder why we start editing the question editor interface now instead if releasing this with the recently published 5.x version.
So question is: Will the new interface be included at one of the next 5.x releases or are you planning a 6.x version with several UI adjustments?

olleharstedt:
> Wasn't there once an UI expet hired who started working on the website instead if the LS interface?

We had a web designer for a while, but no UX expert.

> Will the new interface be included at one of the next 5.x releases or are you planning a 6.x version with several UI adjustments?

Everything that does not break BC will be included in 5.x, as per semantic versioning. :)

PR #1899, opened 2021-05-27
Dev #T752: Database error when editing a quota

gabrieljenik:
Added max length vaildations to fields with string types on the DB.


olleharstedt:
These weird PR titles won't get included in our changelog. Is there no Mantis ticket?

Shnoulle:
Start by Dev => not in changelog

olleharstedt:
> Start by Dev => not in changelog

Yes, but it's obviously a bug fix. We should "market" it. :)

Anyway, I can change it to "Fixed issue: ..." when merging, no problem.

PR #1911, opened 2021-06-07
Fixed issue #17347: 'Class Twig_Environment not found' when uploading an image

gabrieljenik:
Using autloader from composer

### DIFF
--- a/third_party/yiiext/twig-renderer/ETwigViewRenderer.php
+++ b/third_party/yiiext/twig-renderer/ETwigViewRenderer.php
@@ -67,8 +67,8 @@ class ETwigViewRenderer extends CApplicationComponent implements IViewRenderer
 
     function init()
     {
-        require_once Yii::getPathOfAlias($this->twigPathAlias).'/Autoloader.php';
-        Yii::registerAutoloader(array('Twig_Autoloader', 'autoload'), true);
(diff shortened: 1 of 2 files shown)


olleharstedt:
Reverted the merge, sorry.

third_party/yiiext/twig-renderer/ETwigViewRenderer.php
```diff
@@ -67,8 +67,8 @@ class ETwigViewRenderer extends CApplicationComponent implements IViewRenderer
 
     function init()
     {
-        require_once Yii::getPathOfAlias($this->twigPathAlias).'/Autoloader.php';
-        Yii::registerAutoloader(array('Twig_Autoloader', 'autoload'), true);
+        //require_once Yii::getPathOfAlias($this->twigPathAlias).'/Autoloader.php';
```
olleharstedt, line 70: Why did you have to change this?

olleharstedt: Long-term goal is to rename third_party to vendor, as per composer convention, and then include non of it in the repository. Instead, `composer install` should be part of release script. That means we won't be able to touch any of those files. We're not there yet, however, but for yiiext, "composer update" will probably remove any change you did to it.

gabrieljenik:
New version: https://github.com/LimeSurvey/LimeSurvey/pull/1913

PR #1916, opened 2021-06-10
Fixed issue #17276: 500: Internal Server Error - Call to a member function format() on bool

gabrieljenik:
- Revert "Dev Removed obsolete Date_Time_Converter helper class" and fix conflicts
- This reverts commit 0f8dcc2ec477125117c775db6a63bc5bd110683f.

### DIFF
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -2309,11 +2309,13 @@ class SurveyAdministrationController extends LSBaseController
         $aResults = array();
         $expires = App()->request->getPost('expires');
         $formatdata = getDateFormatData(Yii::app()->session['dateformat']);
+        Yii::import('application.libraries.Date_Time_Converter', true);
         if (trim($expires) == "") {
             $expires = null;
         } else {
-            $datetimeobj = DateTime::createFromFormat($formatdata['phpdate'] . ' H:i', $expires);
-            $expires = $datetimeobj->format("Y-m-d H:i:s");
+            //new Date_Time_Converter($expires, $formatdata['phpdate'].' H:i');
+            $datetimeobj = new date_time_converter($expires, $formatdata['phpdate'] . ' H:i');
+            $expires = $datetimeobj->convert("Y-m-d H:i:s");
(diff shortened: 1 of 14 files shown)


application/controllers/SurveyAdministrationController.php
```diff
@@ -2309,11 +2309,13 @@ public function actionExpireMultipleSurveys()
...
         $expires = App()->request->getPost('expires');
         $formatdata = getDateFormatData(Yii::app()->session['dateformat']);
+        Yii::import('application.libraries.Date_Time_Converter', true);
         if (trim($expires) == "") {
             $expires = null;
         } else {
-            $datetimeobj = DateTime::createFromFormat($formatdata['phpdate'] . ' H:i', $expires);
-            $expires = $datetimeobj->format("Y-m-d H:i:s");
+            //new Date_Time_Converter($expires, $formatdata['phpdate'].' H:i');
+            $datetimeobj = new date_time_converter($expires, $formatdata['phpdate'] . ' H:i');
```
olleharstedt, line 2317: Class spelling is different, but maybe it doesn't matter? date_time_converter vs Date_Time_Converter.

olleharstedt:
Was this tested on PHP 8 too?

gabrieljenik:
> Was this tested on PHP 8 too?

Yes, we did a basic text on PHP8. Haven't tested all revertion points, just a few.

olleharstedt:
Great!

application/controllers/SurveyAdministrationController.php
olleharstedt: https://stackoverflow.com/questions/5260168/capital-letters-in-class-name-php

olleharstedt: Not case sensitive. But still.

gabrieljenik: Let us make things straight, just in case

Shnoulle: Autoloading on linux can broke (but no autoload here)

PR #1923, opened 2021-06-15
Fixed issue #17046: When copy question : can not update text and help

gabrieljenik:
Reviewed Copy Question screen as to match new layout.
Code reorganziation on the question editor as to support editing and also copying screens


olleharstedt:
See Zoho for details.

PR #1924, opened 2021-06-15
Fixed issue #17038: templateurl placeholder not being replaced

gabrieljenik:
Updating some template screens which were not migrated properly from legacy.


olleharstedt:
I see only one file changed? Correct?

gabrieljenik:
Correct. That's the only template we found is still using templateurl EM variable

olleharstedt:
Alright

PR #1925, opened 2021-06-15
Fixed issue #17038: templateurl placeholder not being replaced

gabrieljenik:
Updating some template screens which were not migrated properly from legacy.


olleharstedt:
All I see is that templateurl is removed? :)

gabrieljenik:
> All I see is that templateurl is removed? :)

It was removed in the past. We are removing a leftover now.

PR #1927, opened 2021-06-16
Fixed issue #16480: Empty logo causes under specific circumstances loops on the welcome page

gabrieljenik:


application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+    }
+
+    /**
+     * Sanitizes the theme options making sure that paths are valid.
+     * Options that match a file will be cleared if the file is not valid,
+     * or replaced with the virtual path if the file is valid.
+     */
+    public function sanitizeImagePathsOnJson($attribute, $params)
+    {
+        $decodedOptions = json_decode($this->$attribute, true);
```
gabrieljenik, line 1523: // Validates all options of the theme. Not only classic ones which are expected to hold a path but all attributes, as in the future, new attributes may hold a path as well

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+        if (is_array($decodedOptions)) {
+            foreach($decodedOptions as &$value) {
+                if (empty($value) || $value == 'inherit') {
+                    continue;
+                }
+                $virtualPath = $this->getVirtualThemeFilePath($value);
```
gabrieljenik, line 1529: // Step 1 - If option value is a path that matches a virtual path, transform the value to the virtual path

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+                if (!empty($virtualPath)) {
+                    $value = $virtualPath;
+                } elseif ($this->isVirtualPath($value) || realpath($value) !== false || realpath(Yii::app()->getConfig('rootdir') . '/' . $value) !== false) {
```
gabrieljenik, line 1532: // Alternative B - If any of the following:
- option value matches a virtual path format or 
- option value matches a real existing path to a file either relative to the root LS installation or to the current workgin dir or absolute

Mark the value as invalid, as that's not allowed

application/models/TemplateConfiguration.php
```diff
@@ -838,27 +842,46 @@ private function _getImageInfo($file)
...
-        foreach($categoryList as $category) {
-            $fileList = Template::getOtherFiles($category['path']);
+        $categoryList = $this->getFileCategories();
+
+        // Compose list of image files for each category
+        foreach ($categoryList as $category) {
+            // Get base path for category
+            $pathPrefix = empty($category->pathPrefix) ? '' : $category->pathPrefix;
+            if ($category->name == 'theme') {
+                if (!empty($this->template)) {
```
gabrieljenik, line 861: Make a getRelatedTemplate method.
Add it in getTemplateConfigurationForAttribute

That will simplify this block

application/models/TemplateConfiguration.php
```diff
@@ -838,27 +842,46 @@ private function _getImageInfo($file)
...
-        }
-        foreach($categoryList as $category) {
-            $fileList = Template::getOtherFiles($category['path']);
+        $categoryList = $this->getFileCategories();
+
+        // Compose list of image files for each category
+        foreach ($categoryList as $category) {
+            // Get base path for category
+            $pathPrefix = empty($category->pathPrefix) ? '' : $category->pathPrefix;
+            if ($category->name == 'theme') {
```
gabrieljenik, line 860: // If the category is theme, add the "files folder" to the base path, as that's the directory to scan for files

gabrieljenik, line 860: // If the category is theme, add the "files folder" to the base path, as that's the directory to scan for files

application/models/TemplateConfiguration.php
gabrieljenik: Wrap all in a method named getFilesFolder

Shnoulle:
just a confirmation : here you «just» try to disable empty source for img, but if (foir example)

an user put in welcome text (without XSS) `` we can not really avoid the issue : right ?

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
             $this->options = json_encode($aOptions);
         }
     }
+
+    /**
+     * Returns the list theme's file categories
```
gabrieljenik, line 1385: Returns a list of file categories for the theme.
Each category is related to a directory which holds files for the theme.
This files are usually listed to be selected as values for options.

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+     *
+     * @return LimeSurvey\Datavalueobjects\ThemeFileCategory[]
+     */
+    public function getFileCategories()
+    {
+        // We need to determine the paths first. They may already be set, but if they're not, we need to get them from the template.
+        // Note that the template cannot be accessed as relation until the model is saved.
+        $template = !empty($this->template) ? $this->template : Template::model()->findByAttributes(['name' => $this->template_name]);
```
gabrieljenik, line 1393: getRelatedTemplate

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+        $categoryList[] = new ThemeFileCategory('generalfiles', gT("Global"), $generalFilesPath, 'image::generalfiles::');
+        $categoryList[] = new ThemeFileCategory('theme', gT("Theme"), $path, 'image::theme::');
+        if (!empty($this->sid)) {
+            $categoryList[] = new ThemeFileCategory('survey', gT("Survey"), Yii::app()->getConfig('uploaddir') . '/surveys/' . $this->sid . '/images/', 'image::survey::');
+        }
+        return $categoryList;
+    }
+
+    /**
+     * Returns the virtual path of $path if it's valid.
```
gabrieljenik, line 1415: If the path is valid, returns the matching virtual path

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+
+        if (empty($fileInfo)) {
+            return null;
+        }
+
+        return $fileInfo->virtualPath;
+    }
+
+    /**
+     * Returns the real absolute path of $path if it's valid.
```
gabrieljenik, line 1434: Returns the real aboslute path of $path
If $path is not valid, returns null.

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+            return null;
+        }
+
+        /** @var LimeSurvey\Datavalueobjects\ThemeFileCategory[] */
+        $categoryList = $this->getFileCategories();
+
+        // Check if the path matches the virtual path category format
+        $prefix = $this->getVirtualPathPrefix($path);
+        if (!empty($prefix)) {
+            $filteredCategories = array_filter($categoryList, function ($v) use ($prefix) { return $v->pathPrefix == $prefix; });
```
gabrieljenik, line 1472: // Find category that matches the prefix

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+
+        if (empty($fileInfo)) {
+            return null;
+        }
+
+        return $fileInfo->realPath;
+    }
+
+    /**
+     * Returns the theme file information if the path is valid.
```
gabrieljenik, line 1453: Validates $path and returns its file info
Rules:
- valid virtual path 
- real path for an available category

application/models/TemplateConfiguration.php
```diff
@@ -1354,4 +1380,188 @@ public function setOptionKeysToInherit(){
...
+                $virtualPath = str_replace($category->path, $category->pathPrefix, $realPath);
+                return new ThemeFileInfo($realPath, $virtualPath, $category);
+            } else {
+                return null;
+            }
+        }
+
+        // Path doesn't match the virtual path category format, so we try the determine if it belongs to a category.
+        foreach ($categoryList as $category) {
+            // Check for paths relative to the theme, relative to the rootdir, or absolute
```
gabrieljenik, line 1493: Get the realpath for the file. Test whether it is relative to the category, rootdir or absolute

tests/models/TemplateConfigurationTest.php
```diff
@@ -22,4 +22,94 @@ public function testCopyMinimalTemplate()
...
+            'realGeneralFile' => 'upload/themes/survey/generalfiles/index.html',
+            'realRelativeThemeFile' => './files/logo.png',
+            'virtualThemeFile' => 'image::theme::files/logo.png',
+            'virtualGeneralFile' => 'image::generalfiles::index.html',
+            'nonExistingFile' => 'test.png',
+            'existingInvalidFile' => 'themes/admin/Apple_Blossom/images/Limesurvey_logo.png',
+            'virtualPathWithTraversalInsideTheme' => 'image::theme::../fruity/files/logo.png',
+            'virtualPathWithTraversalOutsideTheme' => 'image::theme::../vanilla/files/logo.png',
+        ];
+        $templateConfiguration->options = json_encode($options);
```
gabrieljenik, line 45: // Test

tests/models/TemplateConfigurationTest.php
```diff
@@ -22,4 +22,94 @@ public function testCopyMinimalTemplate()
...
+
+    public function testOptionsSanitization()
+    {
+        // Import lss
+        $surveyFile = self::$surveysFolder . '/limesurvey_survey_666368.lss';
+        self::importSurvey($surveyFile);
+
+        $templateConfiguration = \TemplateConfiguration::checkAndcreateSurveyConfig(self::$surveyId);
+
+        $options = [
```
gabrieljenik, line 34: // Prepare dataset

tests/models/TemplateConfigurationTest.php
```diff
@@ -22,4 +22,94 @@ public function testCopyMinimalTemplate()
...
+        $output = \Yii::app()->twigRenderer->convertTwigToHtml('{{ imageSrc(existingInvalidFile) }}', $options, $templateConfiguration);
+        $this->assertEmpty($output);
+
+        $output = \Yii::app()->twigRenderer->convertTwigToHtml('{{ imageSrc(virtualPathWithTraversalInsideTheme) }}', $options, $templateConfiguration);
+        $this->assertNotEmpty($output);
+
+        $output = \Yii::app()->twigRenderer->convertTwigToHtml('{{ imageSrc(virtualPathWithTraversalOutsideTheme) }}', $options, $templateConfiguration);
+        $this->assertEmpty($output);
+
+        // There is no general image file uploaded but, when setting a valid default,
```
gabrieljenik, line 108: This is hard to test as general files is empty at time of running the test.
So, we use the "default" parameter of imageSrc.

tests/models/TemplateConfigurationTest.php
```diff
@@ -22,4 +22,94 @@ public function testCopyMinimalTemplate()
...
+        // imageSrc will only return false if the file is found and is not a valid image.
+        // We are passing an 'index.html' file here so, imageSrc returning false means that
```
gabrieljenik, line 110: If the file is not found at all, it returns the default value given by parameter.

tests/models/TemplateConfigurationTest.php
```diff
@@ -22,4 +22,94 @@ public function testCopyMinimalTemplate()
...
+        // the file is found but not an image.
```
gabrieljenik, line 111: On this case, the path exists but is not an image, so imageSrc is expected to return false.
We can't assert false as `convertTwigToHtml` transforms it into an empty string.
As it, we reaplce the "false" xexpected by imageSrc to an "OK".

third_party/composer/autoload_psr4.php
```diff
@@ -6,9 +6,43 @@
 $baseDir = dirname($vendorDir);
 
 return array(
-    'Twig\\' => array($vendorDir . '/twig/twig/src', $vendorDir . '/twig/twig/src'),
```
gabrieljenik, line 9: Should this be tracked?

gabrieljenik: ping @olleharstedt

olleharstedt: This file should be whatever "composer install --no-dev" generates.

application/helpers/update/updatedb_helper.php
```diff
@@ -2542,6 +2542,23 @@ function db_upgrade_all($iOldDBVersion, $bSilent = false)
             $oTransaction->commit();
         }
 
+        /**
+         * Sanitize theme option paths
+         */
+        if ($iOldDBVersion < 366) {
+            $oTransaction = $oDB->beginTransaction();
+            
+            $templateConfigurations = TemplateConfiguration::model()->findAll();
```
gabrieljenik, line 2551: I know model are not supposed to be handled here, but we have not many other ways to do it.

Shnoulle: SQL ? No ?

` $templateConfigurations = "SELECT *  from {{template_configuration}} WHERE id > 0";`

id > 0 because i have an issue with WHERE 1 on MSSQL (plugin)

Shnoulle: OK, got the reason … clearly : doing something with model inside update broke some upgrade part …

It can be done at the 1st load ? Adding a "version" in the model ? If version is 2 : it's done  ?

gabrieljenik: I am sorry. I don't understand. Can you provide a bit more context please?

Shnoulle: If you start with a version without TemplateConfiguration : you broke update …

**Need** another solution …

gabrieljenik: > you broke update

How I breake it?

Shnoulle: Please : setup a 2.6 and update. It broke at this step.

It's the reason you must not use model in update.

gabrieljenik: And that would be solved by using this?
`$templateConfigurations = "SELECT * from {{template_configuration}} WHERE id > 0";`

There is a lot of logic in the update that is in the model.
Want to avoid code duplication.

Shnoulle: 1st look :, second one is here : https://github.com/LimeSurvey/LimeSurvey/pull/1927#discussion_r653246334

gabrieljenik: > Please : setup a 2.6 and update. It broke at this step.

We have tried. No error happened.
Can I ask you to elaborate more? I am sorry but I can't understand what's wrong with short answers.

Shnoulle: > We have tried. No error happened.

? Strange and great then :)

third_party/twig/twig/src/Node/IncludeNode.php
```diff
@@ -22,7 +22,7 @@
  */
 class IncludeNode extends Node implements NodeOutputInterface
 {
-    public function __construct(AbstractExpression $expr, AbstractExpression $variables = null, $only = false, $ignoreMissing = false, $lineno = null, $tag = null)
```
olleharstedt, line 25: Is this change by you or composer update?

olleharstedt:
Need to pull upstream again, sorry.

third_party/twig/twig/src/Node/IncludeNode.php
olleharstedt: You can patch things in third_party/ folder, it will be overwritten by composer. Why do you need this?

gabrieljenik: > Is this change by you or composer update?

The file "third_party/twig/twig/src/Node/IncludeNode.php" was updated in commit https://github.com/LimeSurvey/LimeSurvey/commit/5fa36197f5717588925a7efac24df7f54897ed08#diff-be0d7b384824e9eb33de24b7b4f85a48fd05a897acd1087df34052ac4e22d6a1.

Later that file was updated when I ran the composer install surely, returning to its original state.
Now, I reapplied the same commit.

It is good to ask ourselves: what is the correct way when you have to update something in the composer.json.
Because there are files (the autoload ones that are generated) that depend on the version of composer you use. And it is not clear to me if you have to use a specific one, or always use the last one. It sounds to me that varying the composer version can bring unexpected problems, right?

olleharstedt: Gah

olleharstedt: Yes, composer 2 might behave differently than composer 1. There's a dependency that breaks. See this thread: https://forum.yiiframework.com/t/problems-with-composer-2-and-bower-asset/130825/11

c-schmitz:
When applying  the patch  I get 

"Property "TemplateConfiguration.files_folder" is not defined. "

gabrieljenik:
> When applying the patch I get
> 
> "Property "TemplateConfiguration.files_folder" is not defined. "

Fixed! Thanks!

c-schmitz:
Test  failed. See Projects task for details.

gabrieljenik:
Had that test crashed because of us? I think the errors are before the test scripts start to run.

application/helpers/surveyThemeHelper.php
```diff
@@ -0,0 +1,431 @@
...
+*
+*/
+
+use LimeSurvey\Datavalueobjects\ThemeFileCategory;
+use LimeSurvey\Datavalueobjects\ThemeFileInfo;
+
+/**
+ * General helper class for survey themes
+ */
+class surveyThemeHelper
```
olleharstedt, line 23: Capital letter as first letter, please.

application/helpers/surveyThemeHelper.php
```diff
@@ -0,0 +1,431 @@
...
+{
+    /** @var array $aTemplatesInUploadDir cache for the method getTemplateInUpload */
+    public static $aTemplatesInUploadDir = null;
```
olleharstedt, line 26: No Hungarian notation in new files.

Also, feel free to be more specific than "array", e.g. "string[]" for a list of strings.

application/helpers/surveyThemeHelper.php
```diff
@@ -0,0 +1,431 @@
...
+
+    /** @var array $aTemplatesInStandardDir cache for the method getTemplateInStandard */
+    public static $aTemplatesInStandardDir = null;
```
olleharstedt, line 29: Cache can also be static variable inside the function, but it doesn't matter that much.

application/helpers/surveyThemeHelper.php
```diff
@@ -0,0 +1,431 @@
...
+
+    /**
+     * Returns the virtual path prefix of $virtualPath.
```
olleharstedt, line 32: What's "virtual"? Not relative or absolute, but virtual? :)

application/helpers/surveyThemeHelper.php
```diff
@@ -0,0 +1,431 @@
+<?php if (!defined('BASEPATH')) {
```
olleharstedt, line 1: This check is not needed.

application/helpers/surveyThemeHelper.php
olleharstedt: Same with file name.

application/models/Template.php
```diff
@@ -578,10 +578,15 @@ public function deleteAssetVersion()
...
      * @throws Exception
+     * @deprecated Use surveyThemeHelper::getStandardTemplateList() instead.
      */
     public static function getStandardTemplateList()
     {
-        return array_keys(self::getTemplateInStandard());
+        // Refactored into surveyThemeHelper. Replaced the code here
+        // by a call to the helper to avoid code duplication while keeping
+        // backwards compatibility.
+        Yii::import('application.helpers.surveyThemeHelper');
```
olleharstedt, line 588: You know, you can put the file as a service class and have autoload load the file instead of manually import it everywhere. :)

olleharstedt: Or maybe not, since this is LS3...

olleharstedt:
Is this fix ported to master, too?

gabrieljenik:
I think this was merged on cloud, right? @olleharstedt 
If so, what should happen to this PR?

olleharstedt:
Dunno, Carsten knows this better than me, ping him on Cliq.

gabrieljenik:
Merged another PR into master. (https://github.com/LimeSurvey/LimeSurvey/pull/2058)
I think this can be closed. @c-schmitz what do you think?

c-schmitz:
The patch in 5.x is fine. Do we still want to merge this one?

gabrieljenik:
> The patch in 5.x is fine. Do we still want to merge this one?

I wouldn't.

PR #1926, opened 2021-06-16
[ENH] Add RPC function to delete a response using its ID

arnaudbore:
Add RPC function to delete a response depending on the ID of the response. @olleharstedt 
This PR followed this original one #78

PR #1928, opened 2021-06-17
[ENH ] develop - Add RPC function to delete a response using its ID

arnaudbore:
Add RPC function to delete a response depending on the ID of the response.
This PR followed this original one #78 and #1926


arnaudbore:
Test is still missing though.

olleharstedt:
You need any help with adding test?

arnaudbore:
Definitively if you have some time to push this "fast" I'll take it :+1:

olleharstedt:
I have vacation tomorrow, but I can help you next Monday, if time allows. Otherwise, read the tests in the folder I linked before. Should be pretty self-explanatory.

arnaudbore:
Not sure 100% but it seems to be working :+1:  @olleharstedt  @Shnoulle :tada:

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -2988,6 +2988,49 @@ public function update_response($sSessionKey, $iSurveyID, $aResponseData)
...
+    		    if (!isset($oSurvey)){
+    			      return array('status' => 'Error: Invalid survey ID');
+            }
+
+            if (hasSurveyPermission($iSurveyID, 'responses', 'delete')){
+                // get response id from response table using ID
+                $Response = Response::model($iSurveyID)->findByPk($iResponseID);
+                if ($Response){
+                    // delete the files and timings and row
+                    $Response->delete();
```
Shnoulle, line 3018: Sorry again :) . No reason to have this currently but maybe 

```
if ($Response->delete()) {
	return array($iResponseID=>'deleted');
}
return array('status' => 'Response not deleted for unknow reason');
```

application/models/Response.php
```diff
@@ -11,6 +11,7 @@ public function beforeDelete()
     {
         if (parent::beforeDelete()) {
             $this->deleteFiles();
+            $this->deleteTimings();
```
Shnoulle, line 14: :+1:

tests/unit/helpers/RemoteControlTest.php
```diff
@@ -176,6 +176,89 @@ public function testAddResponse()
...
+            $plugin->active = 1;
+            $plugin->save();
+        }
+        App()->getPluginManager()->loadPlugin('Authdb', $plugin->id);
+        // Clear login attempts.
+        $query = sprintf('DELETE FROM {{failed_login_attempts}}');
+        $dbo->createCommand($query)->execute();
+
+        // Import survey
+        $filename = self::$surveysFolder . '/limesurvey_survey_666368.lss';
```
Shnoulle, line 207: Maybe more simple to import a LSA with two response ?

Then no need to test add response for a second time :) : one function => one test

Shnoulle (approved):
OK like this, maybe some improvement

tests/unit/helpers/RemoteControlTest.php
arnaudbore: Can you point me to a data survey which has 2 responses already ? Does this exists ? Should I create one ?

Shnoulle: Just import 666368 survey, launch it manually in your instance, export as Survey archive limesurvey_survey_add_delete_RPC_function_develop.lsa, and put the good name here.

If you want to use an existing one : https://github.com/LimeSurvey/LimeSurvey/tree/master/tests/data/surveys survey_archive_statCountFunctionsTest.lsa have some reponse.

arnaudbore:
Ok I would need a little bit of help here I'm not able to see/understand the error in the test.

olleharstedt:
Travis is green, that's the test we care about. Scrutinizer fails randomly.

olleharstedt:
@Shnoulle Do we have to update some manual entry for this? Otherwise I'll merge.

Shnoulle:
> @Shnoulle Do we have to update some manual entry for this? Otherwise I'll merge.

All api function goes to autodoc if i don't make error

arnaudbore:
Now next step would be to put this in LS 3.x :+1: How do we proceed should I do a new PR ?

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -2988,6 +2988,51 @@ public function update_response($sSessionKey, $iSurveyID, $aResponseData)
...
+     * Returns array
+     *
+     * @access public
+     * @param string $sSessionKey Auth credentials
+     * @param int $iSurveyID Id of the survey that participants belong
+     * @param string $sToken unique token id of specific participant
+     * @param string $sLanguageCode The language to be used
+     * @return array Result of the change action
+     */
+     public function delete_responses($sSessionKey, $iSurveyID, $iResponseID, $sLanguageCode='')
```
olleharstedt, line 3004: Hm, looks like language code is unused? Can you delete it, please?

olleharstedt (changes requested):

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -3001,7 +3001,7 @@ public function update_response($sSessionKey, $iSurveyID, $aResponseData)
      * @param string $sLanguageCode The language to be used
```
olleharstedt, line 3001: And here :point_up:

olleharstedt (changes requested):

application/helpers/remotecontrol/remotecontrol_handle.php
arnaudbore: :facepalm: so sorry

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -2988,6 +2988,50 @@ public function update_response($sSessionKey, $iSurveyID, $aResponseData)
...
+    /**
+     * Delete a response in a given survey using its ID
+     *
+     * RPC Routine to delete responses of particular id in a survey.
+     * Returns array
+     *
+     * @access public
+     * @param string $sSessionKey Auth credentials
+     * @param int $iSurveyID Id of the survey that participants belong
+     * @param string $sToken unique token id of specific participant
```
olleharstedt, line 3000: Sorry, this seems wrong too. Token??

olleharstedt (changes requested):

application/helpers/remotecontrol/remotecontrol_handle.php
arnaudbore: :facepalm: :facepalm:

olleharstedt:
Travis is red, but that's my error from yesterday.

olleharstedt:
Fun fact: Travis was faulty green, so your test was never run. \o/

olleharstedt:
Fixing it now.

olleharstedt:
I think the reason why it was faulty green is because import can do safeDie() instead of throwing exception. I might change that.

PR #1929, opened 2021-06-20
Fixed issue #17367: Survey policy error not show in vanilla

Shnoulle:
Dev: remove all JS , keep the function for compatibility
Dev: add aria-describedby and move error in privacy part.

Dev:


Shnoulle:
oh no !

datasecurity_accepted is not controlled in fact : NO srevr control …

PR #1933, opened 2021-06-23
Dev #17350: Question theme should be a separate database column, not an attribute

gabrieljenik:
Added question_theme_name as a column for questions: Porting the question attribute (on a key-value table) to an attribute of the class itself
Removed places where attributes where queried for fetching question_template (as now it is fetched straightforward)
Also tried to do some name normalization in between question_temlate and question_theme when using the term for variables.


application/controllers/QuestionAdministrationController.php
```diff
@@ -595,11 +582,11 @@ public function actionGetAdvancedOptions(
         $iQuestionId = null,
         $sQuestionType = null,
         $returnArray = false, //todo see were this ajaxrequest is done and take out the parameter there and here
-        $question_template = 'core'
+        $themeName = 'core'
```
gabrieljenik, line 585: Shouldn't this be questionThemeName?

application/controllers/QuestionAdministrationController.php
```diff
@@ -2243,12 +2234,12 @@ private function getGeneralOptions(
         $iQuestionId = null,
         $sQuestionType = null,
         $gid = null,
-        $question_template = 'core'
+        $themeName = 'core'
```
gabrieljenik, line 2237: Shouldn't this be questionThemeName?

application/models/Question.php
```diff
@@ -214,6 +215,8 @@ public function rules()
                 'message' => sprintf(gT("Code: '%s' is a reserved word."), $this->title), // Usage of {attribute} need attributeLabels, {value} never exist in message
                 'except' => 'archiveimport'
             );
+            /* Don't save empty or 'core' question theme name */
```
gabrieljenik, line 218: Why "core" is not a valid one? What is used instead?

application/models/QuestionTheme.php
```diff
@@ -82,6 +82,17 @@ public function relations()
         return array();
     }
 
+    /** @inheritdoc */
+    public function scopes()
+    {
+        return array(
+            'core' => array(
```
gabrieljenik, line 89: Woulnd't be good to have a scope for user ones?

application/models/Question.php
Shnoulle: empty / null (default)

Maybe related ? https://bugs.limesurvey.org/view.php?id=15982

application/controllers/QuestionAdministrationController.php
olleharstedt: "core" is not a theme name, it's a theme _type_.

olleharstedt:
Can you get in a relation like $question->question_theme too? Or maybe $question->questionTheme, where questionTheme is a QuestionTheme object. Then maybe add a tiny test to show that the relation works.

installer/create-database.php
```diff
@@ -374,6 +374,7 @@ function populateDatabase($oDB)
             'scale_id' =>  "integer NOT NULL default '0'",
             'same_default' =>  "integer NOT NULL default '0'",
             'relevance' =>  "text",
+            'question_theme_name' => "string(150) NULL",
```
Shnoulle, line 377: Personally i prefer to have $oQuestion->theme_name or $oQuestion->themename since we are in question object.

application/controllers/QuestionAdministrationController.php
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/1933/files#r657195896 themeName (in my opinion) everywhere

installer/create-database.php
olleharstedt: Well. We have many types of themes. This is question theme.

olleharstedt:
Can you add a small test to show-case the relation between question and question theme?

application/controllers/QuestionAdministrationController.php
gabrieljenik: > "core" is not a theme name, it's a theme _type_.

This issue that you mention is something that it is carried from prior implementations.
On this change we just focus on porting the question attribute (on a key-value table) to an attribute of the class itself.
The contents of the attribute should remain the same.

We can think of normalizing that later maybe?

application/models/QuestionTheme.php
gabrieljenik: Rename to "base"?

installer/create-database.php
gabrieljenik: As this will be used as foreign key for fetching question_theme_names, maybe it is more explicit (although redundant) to levae it as it is now.

Shnoulle: But we already are in question … like relevance in question are question relevance …

$oQuestion->themename are foreing key for $theme->name : this is clear enough.

application/controllers/QuestionAdministrationController.php
gabrieljenik: We will review when "core" is expected for the theme_name attribute.
We believe it is not necessary and that it is legacy code, but want to double check it

application/models/Question.php
gabrieljenik: > empty / null (default)
> 
> Maybe related ? https://bugs.limesurvey.org/view.php?id=15982

Hmm not sure...

installer/create-database.php
gabrieljenik: But there is no relevance object or table.

Shnoulle: > $oQuestion->themename are foreing key for $theme->name : this is clear enough.

We are in question, why add question again …

application/controllers/QuestionAdministrationController.php
olleharstedt: > The contents of the attribute should remain the same.

If this is possible, then sure. :)

application/models/Question.php
```diff
@@ -89,6 +89,9 @@ class Question extends LSActiveRecord
     /** Set defaut relevance **/
     public $relevance ='';
 
+    /** @var QuestionTheme cached question theme*/
+    private $relatedQuestionTheme;
```
olleharstedt, line 93: Isn't the relation already cached by Yii...? o0

olleharstedt: https://www.yiiframework.com/doc/guide/1.1/en/caching.data

application/models/Question.php
```diff
@@ -1573,6 +1576,9 @@ public function getQuestionTheme()
         if (empty($questionThemeName) || $questionThemeName == 'core') {
             $questionThemeName = QuestionTheme::model()->getBaseThemeNameForQuestionType($this->type);
         }
-        return QuestionTheme::model()->findByAttributes(['question_type' => $this->type, 'name' => $questionThemeName]);
+        if (empty($this->relatedQuestionTheme) || $this->relatedQuestionTheme->name != $questionThemeName) {
+            $this->relatedQuestionTheme = QuestionTheme::model()->findByAttributes(['question_type' => $this->type, 'name' => $questionThemeName]);
```
olleharstedt, line 1580: This should use the new relation instead?

application/controllers/QuestionAdministrationController.php
olleharstedt: My vote is on questionThemeName

application/models/QuestionTheme.php
olleharstedt: Why rename? Is there a benefit?

installer/create-database.php
gabrieljenik: because the foreign table is named like that

application/models/Question.php
gabrieljenik: Maybe we should not say cache. Maybe a temporary storage for avoiding round trips.

> Isn't the relation already cached by Yii...? o0

We are not using the relation as when creating questions, that relation is not loaded.
So we are using a method that mimics it.

olleharstedt: Technically correct term: https://en.wikipedia.org/wiki/Memoization

application/models/Question.php
gabrieljenik: We are thinking of not using the relation and discouraging its usage here.
Use this method getQuestionTheme() instead.

More explained here:
https://github.com/LimeSurvey/LimeSurvey/blob/5d5c2a732d0765bd61adeedc60e483a7d7778e18/application/models/Question.php#L130

olleharstedt: Shouldn't $question->questionTheme rather return a default theme if $question is not yet saved? Is that possible?

gabrieljenik: hmm we could make something about it.
We could have the method to return the theme Associated to the question type. if no question type is set we could use the default question type or the default theme

Shall we?

olleharstedt: Or, it's not enough for the relation to set question_type and question_theme_name before using the relation?

```php
$question = new Question(); // Empty object
$question->question_type = 'L';
$question->question_theme_name = 'blaha';
$theme = $question->question_theme;
```

gabrieljenik: When adding a new record the relationship returns null
image

olleharstedt: Interesting.

olleharstedt: So it _would_ be possible to use the relation if we use the `getRelated` method and set `$refresh` to `true`?

olleharstedt: Instead of `$question->question_theme`, it would have to be `$question->getRelated('question_theme', true);`

olleharstedt: No reason to add a new method if one already exist.

application/models/QuestionTheme.php
```diff
@@ -504,34 +515,40 @@ public static function uninstall($oQuestionTheme)
...
-                }
+        if (!empty($aQuestions)) {
+            // There are questions using this theme. Don't delete it
+            $bDeleteTheme = false;
+        }
+
+        // Just in case, if this is a core (base) theme we also check if there are any questions without theme name (this shouldn't happen)
+        if (empty($oQuestionTheme->extends) && $bDeleteTheme !== false) {
+            $aQuestions = Question::model()->findAll(
+                'type = :type AND (question_theme_name = :empty OR question_theme_name IS NULL) AND parent_qid = :parent_qid',
```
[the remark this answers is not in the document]
olleharstedt: Agree.

application/models/Question.php
gabrieljenik: Great!

application/models/QuestionTheme.php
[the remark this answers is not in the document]
gabrieljenik: Good. I think we already changed it.

application/models/Question.php
olleharstedt: Alternatively, getQuestionTheme() just calls $this->getRelated("question_theme", true);

gabrieljenik: Eventually, we could, but not sure it is worth it right now.

gabrieljenik:
> Can you add a small test to show-case the relation between question and question theme?

Done! Thanks!

ptelu (dismissed):

application/models/QuestionTemplate.php
```diff
@@ -140,8 +140,10 @@ public function getTemplatePath()
     public function getQuestionTemplateFolderName()
     {
         if ($this->sTemplateFolderName === null) {
-            $aQuestionAttributes       = QuestionAttribute::model()->getQuestionAttributes($this->oQuestion->qid);
-            $this->sTemplateFolderName = (isset($aQuestionAttributes['question_template']) && $aQuestionAttributes['question_template'] != 'core') ? $aQuestionAttributes['question_template'] : false;
+            $aQuestionAttributes = QuestionAttribute::model()->getQuestionAttributes($this->oQuestion->qid);
+            /** @var string|null */
+            $questionThemeName = $this->oQuestion->question_theme_name;
+            $this->sTemplateFolderName = (!empty($questionThemeName) && $questionThemeName != 'core') ? $questionThemeName : false;
```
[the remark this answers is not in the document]
Shnoulle: `$questionThemeName != 'core'`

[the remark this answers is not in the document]
ptelu: oh wow my bad was reading to much code

application/models/Question.php
```diff
@@ -124,6 +127,11 @@ public function relations()
             'subquestions' => array(self::HAS_MANY, 'Question', array('parent_qid' => 'qid'), 'order' => App()->getDb()->quoteColumnName('subquestions.question_order') . ' ASC'),
             'conditions' => array(self::HAS_MANY, 'Condition', 'qid'),
             'answers' => array(self::HAS_MANY, 'Answer', 'qid', 'order' => App()->getDb()->quoteColumnName('answers.sortorder') . ' ASC'),
+            // The 'question_theme' relation is possible but commented on purpose as to discourage its ussage.
+            // This relation will fail for non saved questions, which is often the case
+            // when using question editor on create mode.
+            // So better use getQuestionTheme()
+            // 'question_theme' => [self::HAS_ONE, 'QuestionTheme', ['type' => 'question_type', 'question_theme_name' => 'name']],
```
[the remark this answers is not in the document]
gabrieljenik: Because the current object is not present as a record in the DB yet.

application/models/QuestionCreate.php
```diff
@@ -23,6 +23,14 @@ public static function getInstance($iSurveyId, $type = null)
         } else {
             $questionType = SettingsUser::getUserSettingValue('preselectquestiontype', null, null, null, Yii::app()->getConfig('preselectquestiontype'));
         }
+        if (isset($themeName) && !empty($themeName)) {
+            $questionThemeName = $themeName;
+        } else {
+            $questionThemeName = SettingsUser::getUserSettingValue('preselectquestiontheme', null, null, null, Yii::app()->getConfig('preselectquestiontheme'));
+        }
+        if (empty($questionThemeName)) {
+            $questionThemeName = QuestionTheme::model()->getBaseThemeNameForQuestionType($this->type);
```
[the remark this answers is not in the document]
gabrieljenik: I think you are right. Will review.

application/models/Question.php
olleharstedt: Why not worth it? It's one line. :)

application/models/Question.php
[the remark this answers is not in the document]
olleharstedt: See discussion above and 

> 

Alternatively, getQuestionTheme() just calls $this->getRelated("question_theme", true);

application/models/Question.php
gabrieljenik: We are reviewing and analyzing. Will be back soon. Thanks

application/models/Question.php
```diff
@@ -1565,20 +1563,11 @@ public function questionThemeNameValidator()
...
-        $questionThemeName = $this->question_theme_name;
-        // If the model's question_theme_name attribute is empty or 'core', we get the name for the base theme
-        if (empty($questionThemeName) || $questionThemeName == 'core') {
-            $questionThemeName = QuestionTheme::model()->getBaseThemeNameForQuestionType($this->type);
-        }
-        if (empty($this->relatedQuestionTheme) || $this->relatedQuestionTheme->name != $questionThemeName) {
-            $this->relatedQuestionTheme = QuestionTheme::model()->findByAttributes(['question_type' => $this->type, 'name' => $questionThemeName]);
-        }
-        return $this->relatedQuestionTheme;
+        return $this->getRelated("question_theme", $this->isNewRecord);
```
olleharstedt, line 1571: Would it make sense to have second argument always false? That is, always fetch the relation, but throw an exception/return null if object did not set question_theme_name and question_type yet?

gabrieljenik: Hmm not sure... need to think about it.
Why? In which cases do you see an advantage?

application/models/Question.php
```diff
@@ -1565,20 +1563,11 @@ public function questionThemeNameValidator()
      * Returns the QuestionTheme related to this question.
      * It's not implemented as a relation because relations only work on
      * persisted models.
-     * It also returns the proper theme when 'question_theme_name' is empty or 'core'.
      *
      * @return QuestionTheme|null
      */
     public function getQuestionTheme()
     {
-        $questionThemeName = $this->question_theme_name;
```
gabrieljenik: Some extra comments about this last commit.

- This method used to do more than returning the related theme. Also used to derived it in some specific cases where the question_theme_names was not explicit (core, or empty question_theme_names). This is not done anymore. Just as a side note.

application/models/Question.php
olleharstedt: I was thinking we can use $question->getQuestionTheme() in the views even when it's not saved yet.

gabrieljenik: Maybe add a force_refresh argument?
- if force_reresh, always refresh
- if not forece_refresh, use $this->isNewRecord, as it is now

olleharstedt: Why? I'm assuming we always know both question theme and question type when entering the view files, both when question is saved and not saved.

gabrieljenik: I say adding an argument to the method as to not always refresh.
Avoid unnecesary refreshes, just that. I am missing something?

olleharstedt: Ah, ok. Yeah, try it. :+1:

gabrieljenik: > I was thinking we can use $question->getQuestionTheme() in the views even when it's not saved yet. So the logic in the view files always look the same.

FYI, the method is being used as $question->questionTheme. Not sure what you mean by " logic in the view files always look the same", but I believe is already working fine.

The parameter with a default value is harmless still, but wanted to double check before jumpng into it

olleharstedt: > Not sure what you mean by " logic in the view files always look the same", but I believe is already working fine.

I meant it should look the same for saved and not-yet-saved questions. If that's already the case, great!

gabrieljenik:
Removing a test which used to verify the deriving theme name from type.
As the deriving is gone now, the test should be gone.

olleharstedt:
These conflicts in db updater are tiresome. Should be refactored.

gabrieljenik:
> These conflicts in db updater are tiresome. Should be refactored into command objects, put each update in a separate file with timestamp as id instead of a simple number.

I agree on refactoring. Not sure that will be enough. Myabe something like Laravel migrations? Still, I don't like the idea of having to mantain a list of applied patches.

olleharstedt:
@gabrieljenik Draft here: https://bugs.limesurvey.org/view.php?id=17492

PR #1936, opened 2021-06-25
Add App->params to allowed Twig attributes

TonisOrmisson:
Dev:
Added the App->params as allowed attributes that can be fetched from twig for templating. 

Twig code has had a major upgrade with v5.0. in v4 the whole App is accessible in twig. With v5 no custom data can be pushed to twig theme. App->params does not contain any sensitive information by default LS config. This change will allow to add custom data to be pushed to params (eg via plug-ins) and then be displayed in themes via twig.

### DIFF
--- a/application/config/internal.php
+++ b/application/config/internal.php
@@ -349,7 +349,7 @@ $internalConfig = array(
                 ),
                 'properties' =>  array(
                     'ETwigViewRendererYiiCoreStaticClassesProxy' => array("Html"),
-                    'LSYii_Application'                          => array("request"),
+                    'LSYii_Application'                          => array("request", "params"),
                     'TemplateConfiguration'             =>  array("sTemplateurl"),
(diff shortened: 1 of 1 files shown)


c-schmitz:
We are currently not saving anything sensitive in the params, but that is mostly because we are not using it properly.
We made up our own configuration  variable  named 'config' in LSYii_Application.

TonisOrmisson:
I'm not sure what you are saying. Would you prefer not exposing this to twig due to any possible sensitive data in the future @c-schmitz? I am not looking to get a defined app config param to twig. I need a vehicle to transport custom data from a plugin to theme.  

Until V4 app->params was a way to do it. If there is another alternative way to do it I am happy to try, but I have not found. 

If the app->params seems like a bad idea, I would suggest creating some custom array that could act as a vehicle for such purposes. Eg create something like `viewParams` for twig conf that could be written to somehow and displayed at the twig theme.

Shnoulle:
>  I need a vehicle to transport custom data from a plugin to theme.

```
<?php
App()->setConfig("myparam",'myvalue');
```

`{{ getConfig("myparam") }}`
https://github.com/LimeSurvey/LimeSurvey/blob/71f2465e66ed00a0aea95f0ba647dd3a819f7dff/application/core/LS_Twig_Extension.php#L645

TonisOrmisson:
@Shnoulle indeed! Perfect! Discussed this some time ago in discord, but then this did not pop up. Thanks!

PR #1938, opened 2021-06-26
Fixed issue #17363: Survey data policy error message not displayed

Shnoulle:
Fixed issue : Privacy was not shown in all in one survey
Dev: html only solution + some js for improvement


Shnoulle:
test link 

- inline : https://master.sondages.pro/256771?newtest=Y&lang=en
- collapsible : https://master.sondages.pro/389748?newtest=Y&lang=en

Still an issue with resume latter at top in all in one.

Shnoulle:
OK : privacy system didn't work in all in one mode : response are created without condition.

Shnoulle:
> OK : privacy system didn't work in all in one mode : response are created without condition.

https://bugs.limesurvey.org/view.php?id=17405

PR #1941, opened 2021-07-01
Fixed issue #17330: Ranking Question Issue

gabrieljenik:
Mandatory Validation
- If min_answers or max_answers is set, there must be at least one response. Failing message is "Please rank the items.".
- If neither limit is set, all options must be ranked. Failing message is as usual: "Please rank all items."


c-schmitz:
Still need path for master branch,

PR #1942, opened 2021-07-01
Fixed issue #17223: With mandatory attribute : lost all current value when create a token

gabrieljenik:
Adding JS validation for required token attributes


assets/scripts/admin/tokens.js
```diff
@@ -174,6 +174,50 @@ function submitEditToken(){
...
 
+/**
+ * Validates that mandatory additional attributes are filled
+ */
+function validateAdditionalAttributes() {
+    const reportFieldValidity = function (field) {
+        // Pick the message from the field attributes, or use a default one if not set
+        var msg = field.attr('data-empty-msg');
+        if (!msg) {
+            msg = "Please fill out this field";
```
Shnoulle, line 185: ? Why ? Browser have own language string here.

Shnoulle:
My opinion : better to use required attribute and when try to submit : go the 1st tab with error.

assets/scripts/admin/tokens.js
gabrieljenik: Updated

gabrieljenik:
Now validation and the way of showing the validation mssage has been simplified using a toast message. 

HTML5 standard validation message  or field highlighting is not used anymore as to simplify the code (could be error prone and complex to do that on hidden tabs)

This is a starter. If we want to make it more complex or enhance it, it can be done from here.

The validation has also been added for the edit process, which didn't perform any validation at all

olleharstedt:
Please pull latest upstream to fix Github Actions.

gabrieljenik:
Ready!

assets/scripts/admin/tokens.js
```diff
@@ -174,6 +177,33 @@ function submitEditToken(){
...
+    $('#edittoken-error-container').show();
+}
+
+/**
+ * Validates that mandatory additional attributes are filled
+ */
+function validateAdditionalAttributes() {
+    const validationErrorMsg = $('#edittoken').attr('data-validation-error');
+
+    $valid = true;
```
olleharstedt, line 191: Hm, why is `$valid` with a dollar sign here? It's JS, not PHP. :)

olleharstedt (changes requested):

assets/scripts/admin/tokens.js
olleharstedt: Guess you want to use "let" or "var" as well.

PR #1946, opened 2021-07-05
Fixed issue #17381: Missing evaluation of EM in CSS script field

gabrieljenik:
Added EM evaluation when showing question summary

### DIFF
--- a/application/views/admin/survey/Question/question_view.php
+++ b/application/views/admin/survey/Question/question_view.php
@@ -200,8 +200,12 @@
                             </td>
                             <td>
                                 <?php
-                                    if (isset($aAdvancedSetting['expression']) && $aAdvancedSetting['expression']==2){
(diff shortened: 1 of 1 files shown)


application/views/admin/survey/Question/question_view.php
```diff
@@ -200,8 +200,12 @@
                             </td>
                             <td>
                                 <?php
-                                    if (isset($aAdvancedSetting['expression']) && $aAdvancedSetting['expression']==2){
-                                        LimeExpressionManager::ProcessString('{' . $aAdvancedSetting['value'] . '}', $qid);
+                                    if (isset($aAdvancedSetting['expression']) && $aAdvancedSetting['expression'] > 0) {
```
Shnoulle, line 203: `if ($aAdvancedSetting['expression']) `?

1. It must be always set : if not set : someone broke system ;) then it's cool if we have broken CI
2. More clear

PR #1949, opened 2021-07-06
Dev #17070: Replacement Fields on Popup Editor for SubQuestions

gabrieljenik:
GID and QID were not sent to editor on subquestions, imapcting on replacement fields.
Issue detected while reviewing https://bugs.limesurvey.org/view.php?id=17070

### DIFF
--- a/assets/scripts/admin/questionEditor.js
+++ b/assets/scripts/admin/questionEditor.js
@@ -95,8 +95,8 @@ $(document).on('ready pjax:scriptcomplete', function () {
    */
   function updateRowProperties() {
     var sID = $('input[name=sid]').val();
-    var gID = $('input[name=gid]').val();
-    var qID = $('input[name=qid]').val();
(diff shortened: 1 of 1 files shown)


assets/scripts/admin/questionEditor.js
```diff
@@ -95,8 +95,8 @@ $(document).on('ready pjax:scriptcomplete', function () {
    */
   function updateRowProperties() {
     var sID = $('input[name=sid]').val();
-    var gID = $('input[name=gid]').val();
-    var qID = $('input[name=qid]').val();
+    var gID = $('[name=question\\[gid\\]]').val();
```
Shnoulle, line 98: Maybe 

```
var qID = false;
if($('[name=question\\[gid\\]]').length) {
    var qID =$('[name=question\\[gid\\]]').val();
}
```

Else : no way to set it with a better solution ? Why not use data-qid or data-gid for example ?

PR #1951, opened 2021-07-06
Dev #T864: Sticky Bar

gabrieljenik:
- Get sticky topbar in on SETTINGS of survey on all 'Survey settings' and 'Survey menu'
- If easy also get sticky topbar in all all pages which have topbar


application/helpers/LayoutHelper.php
```diff
@@ -445,6 +445,11 @@ protected function beforeSideMenuRender(array $aData)
      */
     public static function renderTopbar(array $aData)
     {
+        App()->getClientScript()->registerScriptFile(
+            App()->getConfig('adminscripts') . 'topbar.js',
```
Shnoulle, line 449: Question : why we don't use assetmanager for adminscripts ?

gabrieljenik:
The automated test is failing. 
```
Question Group Editor (ls\tests\QuestionGroupEditor)
 ✘ Edit question group
   │
   │ Facebook\WebDriver\Exception\UnrecognizedExceptionException: Element <a href="#editgrp_de"> is not clickable at point (635,78) because another element <div id="surveybarid" class="menubar surveybar sticky"> obscures it
```

Seems there is some overlapping on the "testing" screen view. Is there any screenshot we can look?
Maybe saved at tests/tmp/screenshots? Although no sure if it is kept.

olleharstedt:
!    │ Facebook\WebDriver\Exception\NoSuchElementException: Unable to locate element: #execute_survey_button

Did this button change?

olleharstedt:
> Maybe saved at tests/tmp/screenshots? Although no sure if it is kept.

No, they're only saved when running tests locally. It's usually just easier to do the test manually step by step, imitating the code in the test class.

gabrieljenik:
I think we are ready to go! automated testing is green

application/helpers/LayoutHelper.php
olleharstedt: You mean a new package? @Shnoulle

assets/scripts/admin/topbar.js
```diff
@@ -0,0 +1,42 @@
+$(document).on('ready pjax:scriptcomplete', function () {
+    function makeTopbarSticky() {
+        const topbar = $('.menubar:not(.surveymanagerbar)');
+        if (topbar.length == 0) {
+            return;
+        }
+        const topbarOffset = topbar.offset().top;
+
+        // When the topbar is floated, the space it occupies is released, so the rest of the content "jumps" up.
```
olleharstedt, line 9: Nice with comments. :+1:

olleharstedt:
Good stuff, guys!

application/helpers/LayoutHelper.php
Shnoulle: No package needed : it's a question and i really thing it's always better to use assetmanager.

`App()->getClientScript()->registerScriptFile(App()->assetManager->publish(App()->getConfig('adminscripts')) . '/' . 'topbar.js');`

For a quick way

olleharstedt: It's not automatically applied for regsiterScriptFile()?

olleharstedt: See source in application/core/LSYii_ClientScript.php

Shnoulle: Arg ! But it's not the best ! It create different asset for all files ! 

Must confirm : but i think it create a copy of all `adminscripts` for each file !

olleharstedt: OK, well, let me know. :)

Shnoulle: And report as issue …

assets/scripts/admin/topbar.js
gabrieljenik: I like comments :) Sometimes more important than code.
Code reflects what I think is the solution.
Comment tells what I think is the solution.
If code fails, comments allow to still see which was the intention

PR #1952, opened 2021-07-07
Add new CI pipeline for 3.x-LTS

MiroslavRepka:
The new pipeline will test the code on all branches that will have
this workflow file. The tests will run with PHP 5.6. and 7.4.
Both jobs will run, even if some fail. The results of the tests can
be found in the action tab, on a repo page.

Changes made to PHP code were necessary, and were discussed when the first
version of the CI pipeline (for the master branch) was developed.

Dev: Add new CI pipeline for 3.x-LTS


MiroslavRepka:
@c-schmitz @olleharstedt @bernardhalas

PR #1955, opened 2021-07-07
Fixed issue #T929 - When the user navigates to Survey List page and clicks on the Action icon 'Add new Question' the users gets a 404

gabrieljenik:


Shnoulle:
Mantis issue number please ?

PR #1957, opened 2021-07-07
Fixed issue #T876 - Seperator - Chrome - When on back screens

gabrieljenik:
The problem was that it had a min-height of 1px, which came from the col class. I did not want to put display: none because it would have to change many things.


Shnoulle:
Mantis issue number please ?

gabrieljenik:
Hi, No mantis issue I believe. Just a dev patch

PR #1963, opened 2021-07-13
Dev #T976: Survey list - Survey Groups - 2 Tabs each one uses a different buttons which should be shown depening on where user is

gabrieljenik:


Shnoulle:
mantis issue ?

PR #1967, opened 2021-07-14
Fixed issue #16792: Delete permission on Survey : no CRSF control

gabrieljenik:
The fix is applied only for the case of removal of survey permissions.
More to come later, after this is approved.

Created JS script for turning links (used to be sent as GET) into POST forms.
This can be later extended to other situations.


application/controllers/admin/surveypermission.php
```diff
@@ -541,6 +541,9 @@ function set($surveyid)
      */
     function delete($surveyid)
     {
+        if(!Yii::app()->getRequest()->isPostRequest) {
```
gabrieljenik, line 544: Maybe we can place this in a some common library or base controller method?
something like $this->requirePost()

Shnoulle:
In master : Group deletion + Question deletion use https://github.com/LimeSurvey/LimeSurvey/blob/master/assets/packages/adminbasics/src/parts/confirmationModal.js
But no CRSF control.

Shnoulle:
In 3.X : https://github.com/LimeSurvey/LimeSurvey/blob/db67654da6e9ad311ff863273711301dfb96fd64/assets/packages/adminbasics/src/parts/confirmationModal.js#L55

c-schmitz:
That does not seem like a most minimal fix.

gabrieljenik:
> @gabrieljenik That does not seem like a most minimal fix. I would prefer to use something existing.

You are very right. We missed the sendPost function.

If it is fine, we can go ahead an apply same solution to other places as well

Shnoulle:
We can not copy function from Group and Question deletion ? 
Use same confirm box seems better (in my opinion).

application/controllers/admin/surveypermission.php
```diff
@@ -129,7 +129,7 @@ public function index($iSurveyID)
                         'uid' => $PermissionRow['uid']
                     ));
                     $deleteConfirmMessage = gT("Are you sure you want to delete this entry?");
-                    $surveysecurity .= "<a data-target='#confirmation-modal' data-toggle='modal' data-message='{$deleteConfirmMessage}' data-onclick='(function() {window.LS.sendPost(\"{$deleteUrl}\");})' type='submit' class='btn-xs btn btn-default'>
+                    $surveysecurity .= "<a data-target='#confirmation-modal' data-toggle='modal' data-message='{$deleteConfirmMessage}' data-onclick='(function() {" . convertGETtoPOST($deleteUrl) . "})' type='submit' class='btn-xs btn btn-default'>
```
Shnoulle, line 132: `convertGETtoPOST` is not deprecated ?

Shnoulle: > We can not copy function from Group and Question deletion ?
> Use same confirm box seems better (in my opinion).

Oh no : it's really an awfull code ...
https://github.com/LimeSurvey/LimeSurvey/blob/db67654da6e9ad311ff863273711301dfb96fd64/application/views/admin/survey/Question/questionbar_view.php#L128

gabrieljenik: > `convertGETtoPOST` is not deprecated ?

No, still in use.

application/controllers/admin/surveypermission.php
Shnoulle: yep

gabrieljenik:
Let me know if it is OK so we can move the solution to other places as well. Thanks!

Shnoulle:
Personally : i prefer a function … using some data-action or something like this.
I really dislike to see onlick=function i think it's hard to manage , fix and update.

See https://github.com/LimeSurvey/LimeSurvey/blob/f8ee40d663eb346b620091b1253266948b73db69/assets/packages/adminbasics/src/components/gridAction.js#L5

gabrieljenik:
> Personally : i prefer a function … using some data-action or something like this.
> I really dislike to see onlick=function i think it's hard to manage , fix and update.
> 
> See
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/f8ee40d663eb346b620091b1253266948b73db69/assets/packages/adminbasics/src/components/gridAction.js#L5

I think it was like that on the very begining, but later we change it to a more minimum-change approach.

Shnoulle:
Maybe by error : 

> And use POST : we can not use same function directly ?

from https://github.com/LimeSurvey/LimeSurvey/pull/1967#issuecomment-880553590

@c-schmitz create a new js file with a complete function ?

gabrieljenik:
@Shnoulle You are referring to this approach?
https://github.com/LimeSurvey/LimeSurvey/commit/89f478b04e63391b4a4b04a5cc31433c5f6c384c

Shnoulle:
> @Shnoulle You are referring to this approach?
> [89f478b](https://github.com/LimeSurvey/LimeSurvey/commit/89f478b04e63391b4a4b04a5cc31433c5f6c384c)

I didn't test but just`data-post-url='{$deleteUrl}'` work ? If yes : great !

gabrieljenik:
@c-schmitz what do you think?

olleharstedt:
Please have a look at the Yii framework support for limiting action access: https://www.yiiframework.com/doc/guide/1.1/en/basics.controller#filter

gabrieljenik:
> Please have a look at the Yii framework support for limiting action access: https://www.yiiframework.com/doc/guide/1.1/en/basics.controller#filter

OK, that would be good for the PHP part. Nice. 
Still, need to decide what to do with the JS part. 
We the 2 options implemented :)

gabrieljenik:
> Please have a look at the Yii framework support for limiting action access: https://www.yiiframework.com/doc/guide/1.1/en/basics.controller#filter

Would we implement this on Master only or LTS as well?

olleharstedt:
> Would we implement this on Master only or LTS as well?

Good question. What would the impact be in LTS? Same as for master? It might not work for SurveyCommonAction controller classes.

gabrieljenik:
> > Would we implement this on Master only or LTS as well?
> 
> Good question. What would the impact be in LTS? Same as for master? It might not work for SurveyCommonAction controller classes.

I wouldn't innovate too much on LTS. Would apply filters only on Master

olleharstedt:
> I wouldn't innovate too much on LTS. Would apply filters only on Master

Yeah, alright. Assuming we don't have to fix bugs in the new code.

gabrieljenik:
> > I wouldn't innovate too much on LTS. Would apply filters only on Master
> 
> Yeah, alright. Assuming we don't have to fix bugs in the new code.

Yes, always assuming that :) Your call :)

olleharstedt:
If it's my call I'd prefer the bug to fixed identically in both branches (unless it's a lot of extra work).

gabrieljenik:
> If it's my call I'd prefer the bug to fixed identically in both branches (unless it's a lot of extra work).

Let's try then applying filters in LTS and master. Let's see how it goes.

olleharstedt:
Thanks!

gabrieljenik:
As we are talking about filtering on subaction, it will not be straightforward to implement the filtering on LTS using the Yii Filter feature. I suggest to keep on going with `$this->requirePostRequest()` at least on LTS.

Still, need some decision about the JS part.
We have 2 approaches:
Approach #1: https://github.com/LimeSurvey/LimeSurvey/pull/1967/commits/89f478b04e63391b4a4b04a5cc31433c5f6c384c
Approach #2:  https://github.com/LimeSurvey/LimeSurvey/pull/1967/commits/9a95eedf89a45d1143a07fae49b805cdffad5ebc

Any direction so we can move forward to the other screens and apply the same recipe?

olleharstedt:
>  I suggest to keep on going with $this->requirePostRequest() at least on LTS.

Go ahead. For master you can write a "todo" in the code, so Patricia will see it when she refactors the code into a normal Yii controller.

Will check the approaches tomorrow.

olleharstedt:
Is JS part a question of where to put the CSRF token? It's already included in the jQuery system from here: https://github.com/LimeSurvey/LimeSurvey/commit/d3c45064d5cab7067d7ecf5632d40b25fa142a90#diff-e505c3478e5568e02d9dbd28f4bbbca65b1afb6806330a820731da977583f25aR31

gabrieljenik:
> Is JS part a question of where to put the CSRF token? It's already included in the jQuery system from here: [d3c4506#diff-e505c3478e5568e02d9dbd28f4bbbca65b1afb6806330a820731da977583f25aR31](https://github.com/LimeSurvey/LimeSurvey/commit/d3c45064d5cab7067d7ecf5632d40b25fa142a90#diff-e505c3478e5568e02d9dbd28f4bbbca65b1afb6806330a820731da977583f25aR31)

If I understand your question correctly, no.
The dilema is if:
A) Add function on adminbasics to allow `data-post-url` attribute on links (https://github.com/LimeSurvey/LimeSurvey/commit/89f478b04e63391b4a4b04a5cc31433c5f6c384c#diff-61f89174ddcfdb449dec14f8797eac378992ec6e4159dc8ac362a529c43e1f73R28275)
B) Use `data-onclick='(function() {" . convertGETtoPOST($deleteUrl) . "})'`  (https://github.com/LimeSurvey/LimeSurvey/pull/1967/commits/147a2dd78a2ccd4f8ecf7fdb2d2bf7ac4edab6c3)

olleharstedt:
Uh, what's the trade-off?

olleharstedt:
You don't want to add a tiny <form> to create a POST request, but hack it with JS instead?

gabrieljenik:
> You don't want to add a tiny <form> to create a POST request, but hack it with JS instead?

Well, the whole issue is to convert some links to POST.
There are couple of them. Not only this one.
So looking for a solution that can be generalized.

I guess we could also add the forms.

olleharstedt:
You can do POST with Ajax too, if you want. Might be easier than to inject an HTML form.

olleharstedt:
>  So looking for a solution that can be generalized.

Please look for a long-term solution, that is, a solution that's easy to read and change, not something that's easy to write. ;)

gabrieljenik:
> Ping, don't forget.

# Current Situation:
ConfirmationModal is currently sending GET regular requests (when used through data-href parameter).
We need to send POST regular requests.

# Alternative Solution I
Have some JS to inject a form in the HTML to turn the GET into a POST.
This is what has been done so far. Needs to be spreaded to other places as well

# Alternative Solution II
Sending $.post() instead of injecting a FORM.

## Cons:
AJAX can't be a viable solution as that would require:
- the current processing controllers to return an ajax based-response 
- the current view should be adapted for reacting to that ajax response. 

It would be like transferring the "end part" of the processing controllers to the view.
Why do we need to do all that movement?

assets/packages/adminbasics/build/adminbasics.js
```diff
@@ -28275,6 +28272,32 @@
...
+            value = options.postDatas[key].value;
+            htmlClass = options.postDatas[key].class;
+          }
+
+          formObject.append('<input name="' + key + '" value="' + value + '" type="' + type + '" ' + (htmlClass ? 'class="' + htmlClass + '"' : '') + ' />');
+        }
+
+        formObject.append('<input name="' + LS.data.csrfTokenName + '" value="' + LS.data.csrfToken + '" type="hidden" />');
+        modalObject.find('.modal-body').append(formObject);
+        formObject.submit();
```
olleharstedt: https://stackoverflow.com/a/5533477/2138090

olleharstedt:
Didn't find any proper way to do a JavaScript POST without an injected form, so go ahead. :) See my stackoverflow link for some alternative on how to inject the form input elements without concatenating strings.

assets/packages/adminbasics/build/adminbasics.js
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/blob/9aa4fd79dbde1a9a31ba186c5def5387f70b87ae/assets/packages/template-core/template-core.js#L45-L51

PR #1968, opened 2021-07-15
Dev #T907: Workflow optimization - Activation - Screen 1

gabrieljenik:


Shnoulle:
Can we have some detail on the new workflow without checkout and test ?

thedirtypanda:
This task is about rewording activation screen and adding some tooltips @Shnoulle

application/views/surveyAdministration/activateSurvey_view.php
```diff
@@ -255,10 +254,21 @@ function alertPrivacy()
                     </div>
                 </div>
             </div>
-
+            
+            <!-- Tip -->
             <div class='row'>
                 <div class='col-sm-8 col-sm-offset-2'>
-                    <div class="alert alert-info"><?php eT("Tip: Please note that you need to deactivate a survey if you want to add or delete groups/questions or change any of the settings above. The changes will cause all collected data from respondents to be moved and archived."); ?></div>
+                    <div class="alert alert-info">
```
olleharstedt, line 261: Yeah, this won't work. :) Not all languages have the same order of verbs, nouns, etc etc.

PR #1980, opened 2021-07-23
Fixed issue #17423: Fix database custom port not being used

grayot:
Fixed issue #17423:
Dev: Add DB port to the session in order to use it in case of custom port.


olleharstedt:
Thank you!

PR #1981, opened 2021-07-26
Fixed-issue #17474: Unable to update token value of response with remote control

arnaudbore:
Now when using update_response (remote control) with an array containing only token and id, token is updated without the need for the survey to be alloweditaftercompletion.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2676,7 +2676,7 @@ class remotecontrol_handle
      * Routine is only applicable for active surveys with alloweditaftercompletion = Y.
      *
      * If only token and id are provided then token is updated
-     * In this condition routine will be applicable whether or not survey is alloweditaftercompletion
+     * In this condition, routine will be applicable whether or not survey is alloweditaftercompletion
      *
      * @access public
      * @param string $sSessionKey Auth credentials
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2675,6 +2675,9 @@ class remotecontrol_handle
      * Response to update will be identified either by the response id, or the token if response id is missing.
      * Routine is only applicable for active surveys with alloweditaftercompletion = Y.
      *
+     * If only token and id are provided then token is updated
+     * In this condition routine will be applicable whether or not survey is alloweditaftercompletion
+     *
      * @access public
      * @param string $sSessionKey Auth credentials
      * @param int $iSurveyID Id of the Survey to update response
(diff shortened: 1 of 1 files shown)


application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -2675,6 +2675,9 @@ public function add_response($sSessionKey, $iSurveyID, $aResponseData)
      * Response to update will be identified either by the response id, or the token if response id is missing.
      * Routine is only applicable for active surveys with alloweditaftercompletion = Y.
      *
+     * If only token and id are provided then token is updated
```
Shnoulle, line 2678: I think i prefer to have : 

- if id is provided : this response was updated (token included)
- if only token is provided : only single response with this token was updated.

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -2699,6 +2702,25 @@ public function update_response($sSessionKey, $iSurveyID, $aResponseData)
             return 'Error: Survey does not allow edit after completion.';
         }
 
+        if (count($aResponseData)==2 && isset($aResponseData['id']) && isset($aResponseData['token']))
```
Shnoulle, line 2705: Totally different function then : best is to create a new function : `update_token_in_response($sSessionKey, $iSurveyID, $iResponseId,$sToken)`

Shnoulle (commented):
2 choice (in my opinion)

- review to not unset token if id is set `if(empty($aResponseData['id'])) { unset($aResponseData['token']) }`
- create a separate function

Really personal opinion : 1st option is better.

arnaudbore:
If I go with first option I also want to get rid of the validation alloweditaftercompletion and that's why I'm more interested in having a separate function. What do you think ?

Shnoulle:
If you send aResponseData as `['id => $id, 'submitdate' => null, 'token' => $newtoken] `?

arnaudbore:
Maybe I should know which fields are part of the survey and which ones are part of a response (like default fields) before moving forward. In that case I would be able to create a new function to update these specific fields. What do you think ? Do you know what are these fields ?

arnaudbore:
In your opinion what should be the name of this new function ?

olleharstedt:
@arnaudbore You should probably bring this up during a community meeting on Discord. Tuesday 14:30, CET.

c-schmitz:
IMHO you should add new function 
set_response_properties

set_response_properties(string $sSessionKey,integer $iSurveyID,array|integer $aResponseQueryProperties,array $aResponseData): array

You can probably copy most of the existing code. 
The problem is the implicit search in the existing update_response. (which should be marked as obsolete when your new function works).

glimz:
Since we did not get any update from you in the last 10 days, we are closing the PR. Feel free to create a new PR with the changes suggested and we will test and merge if everything is ok with the PR.

PR #1982, opened 2021-07-26
Dev #17375: Unify upper/lower case in question type

gabrieljenik:


olleharstedt:
Sorry, due to a security fix you'll have to bump the dbversion here. Maybe use 452? 451 is already duplicated between master and dev.

olleharstedt:
And again another fix was needed. :(

thedirtypanda:
Please resolve the issues.

thedirtypanda (approved):

PR #1983, opened 2021-07-27
3XLTS isfile use

Shnoulle:
Dev: Log as warning for path traversable check

### DIFF
--- a/application/controllers/admin/responses.php
+++ b/application/controllers/admin/responses.php
@@ -636,27 +636,26 @@ class responses extends Survey_Common_Action
             if (isset($aQuestionFiles[$iIndex])) {
                 $aFile = $aQuestionFiles[$iIndex];
                 // Real path check from here: https://stackoverflow.com/questions/4205141/preventing-directory-traversal-in-php-but-allowing-paths
-                $sDir = Yii::app()->getConfig('uploaddir') . "/surveys/" . $iSurveyId . "/files/";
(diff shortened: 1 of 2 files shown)


olleharstedt:
No reason to touch 3.x, is it? Unless we can cherry-pick it to master.

Shnoulle:
No warning log, and **your** fix add more lines than mine.

See original : https://github.com/LimeSurvey/LimeSurvey/blob/a90933d7c1c1084d69a94e2ff60c005e1d9510e7/application/controllers/admin/responses.php#L638 

If you want i can make a one line fix comparing with 3.27.8+210721

olleharstedt:
OK, but we need to test it before merge.

Shnoulle:
> Log as warning for path traversable check

Have for sysadmin a way to check if there are path traversal try. 

If i remind : _done just after_ @olleharstedt fix a security issue .

When fixing security : i really think we **must** add log (on webserver is better (403/400) but on limesurvey log at minimum)

But 7 month after … hard to remind …

c-schmitz:
Denis, do you still want to merge it? I would agree but it first needs to be updated.

Shnoulle:
> When fixing security : i really think we must add log (on webserver is better (403/400) but on limesurvey log at minimum)

Can we remind of this next time ?
Because doing a second pull request to fix a just fixed commit and have to fix it again month after …
I'm not interested.

PR #1984, opened 2021-07-27
Fixed issue #17148: Publication & access control settings - Start Date / Expiry Date Picker inaccessible in resized browser window

gabrieljenik:
Automatically enable overflow of container class when datepicker is shown.


olleharstedt:
Which commands did you run to build adminbasics?

PR #1988, opened 2021-08-02
Fixed issue #17285: "Resume later" not working after "Load unfinished survey"

gabrieljenik:
It seems to me that there is a confusion between the "step" of the session and the value of $ thisstep.

When you are in group 2 and you save the survey for the first time, the saved table is saved_thisstep = 2. (group number, not group sequence)
That comes out of $ thisstep, which in that context has the same as the "step" of the session (the current group number).

When you save the survey for the second time (the save screen no longer appears, it is saved directly), execute _UpdateValuesInDatabase and set there $ thisstep = $ this-> currentGroupSeq (sequence, that is, 1 instead of 2) and then the SavedControl is updated with that value. At that point, $ thisstep is different from the session step.

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -5648,7 +5648,7 @@
                     }
                     elseif ($this->surveyOptions['allowsave'] && isset($_SESSION[$this->sessid]['scid']))
                     {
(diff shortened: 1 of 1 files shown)


maziminke:
@gabrieljenik What does "not working" actually mean in this case? I remember a similar problem where a green "false" note was shown on top for whatever reason. Is this the same issue?

PR #1993, opened 2021-08-05
Dev #T1062: When user clicks the reset button the screen freezes while should reset to new default

gabrieljenik:
Added "reset" buttons for the specific screens on FullPagebar.

### DIFF
--- a/application/controllers/admin/SurveymenuController.php
+++ b/application/controllers/admin/SurveymenuController.php
@@ -391,7 +391,8 @@ class SurveymenuController extends Survey_Common_Action
                 'buttons' => [
                     'addMenu' => true,
                     'addMenuEntry' => true,
-                    'reset' => Permission::model()->hasGlobalPermission('superadmin', 'read'),
+                    'resetMenus' => Permission::model()->hasGlobalPermission('superadmin', 'read'),
+                    'resetMenuEntries' => true, // TODO: Is this OK? It wasn't checking any permission on SurveymenuEntryController
                     'reorder' => true,
                 ],
             ],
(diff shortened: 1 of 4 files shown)


application/controllers/admin/SurveymenuController.php
```diff
@@ -391,7 +391,8 @@ public function view()
                 'buttons' => [
                     'addMenu' => true,
                     'addMenuEntry' => true,
-                    'reset' => Permission::model()->hasGlobalPermission('superadmin', 'read'),
+                    'resetMenus' => Permission::model()->hasGlobalPermission('superadmin', 'read'),
+                    'resetMenuEntries' => true, // TODO: Is this OK? It wasn't checking any permission on SurveymenuEntryController
```
Shnoulle, line 395: Unsure ;  Permission::model()->hasGlobalPermission('superadmin', 'read'), ?

But tottaly unsure here … there are user menu entries ?

application/views/admin/super/fullpagebar_view.php
```diff
@@ -99,6 +99,29 @@ class='btn btn-default'
                 </a>
             <?php endif; ?>
 
+            <!-- Reset Survey Menus -->
+            <?php if (isset($fullpagebar['menus']['buttons']['resetMenus']) && $fullpagebar['menus']['buttons']['resetMenus']) :?>
```
Shnoulle, line 103: I always prefer : `if (!empty($fullpagebar['menus']['buttons']['resetMenus'])`

shorter, and allow 0/""/null/false/not set for false

Like other dev want :)

Shnoulle (commented):
Code is OK

Shnoulle (changes requested):
Code OK, some remark not linked to code + conflict

gabrieljenik:
I think this can be closed.
Seems this has already been solved differently on master.
Zoho task is marked as done.

PR #1994, opened 2021-08-06
Bug/16720  old invalid question codes adjusted during lss import do not get replaced

gabrieljenik:


gabrieljenik:
Will apply n master as well

PR #1996, opened 2021-08-06
DEV: Fix failing QuestionGroupEditorTest tests

gabrieljenik:
- Sometimes the tests fail because language tabs remain under the sticky topbar. Added a delay after scrolling to top to give it some time.

### DIFF
--- a/assets/scripts/admin/topbar.js
+++ b/assets/scripts/admin/topbar.js
@@ -6,6 +6,11 @@ $(document).on('ready pjax:scriptcomplete', function () {
         }
         const topbarOffset = topbar.offset().top;
 
+        // Don't make the topbar sticky if the window height is to small
+        if (window.innerHeight < topbarOffset + topbar.outerHeight(true)) {
+            return;
+        }
+
         // When the topbar is floated, the space it occupies is released, so the rest of the content "jumps" up.
         // To avoid that, a "placeholder" div is added, with the same height as the topbar.
         // The placeholder is only visible (by css) when the topbar has the 'sticky' class.
--- a/tests/functional/backend/QuestionGroupEditorTest.php
+++ b/tests/functional/backend/QuestionGroupEditorTest.php
@@ -7,6 +7,7 @@ use Facebook\WebDriver\WebDriverExpectedCondition;
 use Facebook\WebDriver\WebDriverDimension;
 use Facebook\WebDriver\Exception\NoSuchElementException;
 use Facebook\WebDriver\Exception\TimeOutException;
+use Facebook\WebDriver\Exception\UnrecognizedExceptionException;
 
 /**
  * Create and edit a question group.
@@ -89,7 +90,6 @@ class QuestionGroupEditorTest extends TestBaseClassWeb
 
             // Switch to German tab.
             self::$webDriver->executeScript("window.scrollTo(0, 0);");  // Scroll to top because otherwise the tabs may be hidden under the topbar
-            sleep(2);
             $germanTab = self::$webDriver->findElement(WebDriverBy::partialLinkText('German'));
             $germanTab->click();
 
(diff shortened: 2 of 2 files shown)


olleharstedt:
Delay didn't fix it?

gabrieljenik:
> Delay didn't fix it?

Seems not and very hard to reproduce.
So, not big fan of suggesting this, but think will make our life easier.
What if we add a rule for not enabling the stickybar on certain screen widths? or even better with some browser agents?

olleharstedt:
Hm. I wonder what the screenwidth is for headless Firefox...

olleharstedt:
Could this help? https://github.com/mozilla/geckodriver/issues/1354

olleharstedt:
> 1366x768 is the default resolution for headless mode. You can change this by setting the following environment variables: MOZ_HEADLESS_WIDTH and MOZ_HEADLESS_HEIGHT.

olleharstedt:
Not needed anymore?

gabrieljenik:
This is still needed. Strange s that some PR test succesfully and some others not.
We will increase sleep and try again

olleharstedt:
>  We will increase sleep and try again

No, remove the sleep and test set a width instead. :) See my comment.

olleharstedt:
You still didn't try this?

> 1366x768 is the default resolution for headless mode. You can change this by setting the following environment variables: MOZ_HEADLESS_WIDTH and MOZ_HEADLESS_HEIGHT.

You just hate the thought of it? :) You'd have to add it in the Github Action workflow file.

gabrieljenik:
> You still didn't try this?
> 
> > 1366x768 is the default resolution for headless mode. You can change this by setting the following environment variables: MOZ_HEADLESS_WIDTH and MOZ_HEADLESS_HEIGHT.
> 
> You just hate the thought of it? :) You'd have to add it in the Github Action workflow file.

haha :) I just hate this story which drives me crazy!

In our local tests we saw that the effect of MOZ_HEADLESS_HEIGHT is limited by the display height of the X server, and that the default (768) is far enough so that the field is not covered. So, resolutions were good enough, not sure what to change there? Make it even bigger? Does it worth it?

The issue (we believe) was on the window size, which we believe was nearly zero.
We couldn't get to prove it on the CI pipeline as now checks are coming out good!

We tried a workaround, disable the stickyness when the height is very small, as to mitigate risks.

olleharstedt:
> In our local tests we saw that the effect of MOZ_HEADLESS_HEIGHT is limited by the display height of the X server, and that the default (768) is far enough so that the field is not covered. So, resolutions were good enough, not sure what to change there? Make it even bigger? Does it worth it?

Was this still run headless?

gabrieljenik:
> Why did you need an X server?

You are right, we don't need it!
Thanks for the comment!
Still, the comments apply I think.

olleharstedt:
For us it seems like ResizeObserver is not defined when running Firefox in headless (or even not headless, just normally).

olleharstedt:
> Sometimes the tests fail because language tabs remain under the sticky topbar

Which exact error are you getting when you think this happens?

gabrieljenik:
> For us it seems like ResizeObserver is not defined when running Firefox in headless (or even not headless, just normally).

What do you mean. If understand correctly, than it shouldn't work even normally and as far as I know, it works, right?

gabrieljenik:
> > Sometimes the tests fail because language tabs remain under the sticky topbar
> 
> Which exact error are you getting when you think this happens?

https://github.com/LimeSurvey/LimeSurvey/pull/1996/checks?sha=86eeaadbd4d3ee963acc52dce2e601b2f4d2c130#step:13:287

```
Facebook\WebDriver\Exception\UnrecognizedExceptionException: Element <a href="#editgrp_de"> is not clickable at point (635,99) because another element <div id="surveybarid" class="menubar surveybar sticky"> obscures it
```

olleharstedt:
Me and Patrick could fix this by applying this:

> 1366x768 is the default resolution for headless mode. You can change this by setting the following environment variables: MOZ_HEADLESS_WIDTH and MOZ_HEADLESS_HEIGHT.

Like I said one month ago :facepalm:

gabrieljenik:
That is something that you applied to master or on the testing script?

olleharstedt:
Sorry, we didn't merge it yet, but here's the fixing commit: https://github.com/LimeSurvey/LimeSurvey/commit/75e929f895da795633215af977750147cfccbcab

gabrieljenik:
As you can see on the check, some commits are green and some are red.
So, hope that's the good solution

olleharstedt:
You mean that it randomly succeeds, and we just get lucky? :) That might be the case, yes... Let's see.

gabrieljenik:
Lucky is not a word I would use. :) Maybe you nailed it. I am saying is hard to know if a solution is final when is failing interemitently.

That's why we disabled the stickyness when the height is very small, as to mitigate risks.
We can afford not to have stickyness on testing

olleharstedt:
>  We can afford not to have stickyness on testing

Maybe, but let's try other possible solutions before starting to change the code to behave differently in the test env.

PR #1999, opened 2021-08-09
Fixed issue #17350: Update fails on pgsql and mssql 

gabrieljenik:
- Updated query used in update according to engine


olleharstedt:
Could this be simplified with the query builder? :thinking:

PR #2004, opened 2021-08-13
Fixed issue #17421: When upgrading from LS3 to LS4 / 5, plugin Manager can't be used if non-compatible plugins are found

gabrieljenik:


gabrieljenik (commented):
> Also see this issue: https://bugs.limesurvey.org/view.php?id=17516

You mean to focus on this https://bugs.limesurvey.org/view.php?id=17516#c66011?

gabrieljenik:
To highlight: If plugin is missing or is not compatible, it will be uninstalled

olleharstedt:
> To highlight: If plugin is missing or is not compatible, it will be uninstalled

Right, but this procedure might need to happen after db update too, not only when entering the plugin manager.

application/controllers/admin/PluginManagerController.php
```diff
@@ -37,7 +37,19 @@ public function index()
 
         $aoPlugins = Plugin::model()->findAll(array('order' => 'name'));
         $data      = [];
-        foreach ($aoPlugins as $oPlugin) {
+        foreach ($aoPlugins as $key => $oPlugin) {
```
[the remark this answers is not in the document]
olleharstedt: The plugin system has its own folder. Maybe a new helper class there?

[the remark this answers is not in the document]
olleharstedt: application/libraries/PluginManager

application/libraries/PluginManager/PluginCleaner.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Sets the flash messages corresponding to each removed plugin.
+     */
+    public function showFlashMessages()
+    {
+        foreach ($this->removedPlugins as $reason => $pluginNames) {
+            $baseMessage = array_key_exists($reason, $this->messages) ? $this->messages[$reason] : "Plugin '%s' was uninstalled.";
+            foreach ($pluginNames as $pluginName) {
```
gabrieljenik, line 80: Not too much inlove with this. Maybe we can output anarray of messages and have a helper make them flash?

application/libraries/PluginManager/PluginCleaner.php
```diff
@@ -0,0 +1,85 @@
...
+ * Helper class used to clean up the list of installed plugins
+ */
+class PluginCleaner
+{
+    /** @var array<string,string[]> list of removed plugin names, grouped by reason of removal */
+    protected $removedPlugins = [];
+
+    /** @var array<string,string> texts to use on flash messages depending on the reason of removal */
+    protected $messages = [
+        'incompatible' => "Plugin '%s' was uninstalled because it is not compatible with your LimeSurvey version.",
```
olleharstedt, line 15: I think this has to be set i  the constructor, so it can be wrapped by gT().

application/models/Plugin.php
```diff
@@ -87,8 +87,12 @@ public function setLoadError(array $error)
      */
     public function isCompatible()
     {
-        $config = $this->getExtensionConfig();
-        return $config->isCompatible();
+        try {
+            $config = $this->getExtensionConfig();
+            return $config->isCompatible();
+        } catch (\Exception $ex) {
```
olleharstedt, line 93: Which exception are you catching here?

gabrieljenik: Any exception from the getExtensionConfig or the isCompatible method.
Example: 'Missing configuration file for plugin %s, looked in location %s',

olleharstedt: Hm, do we want to silence that? Try-catch usually makes better sense higher up in the stacktrace, e.g. in a controller or such.

olleharstedt: By returning `false` we lose important information.

application/controllers/admin/PluginManagerController.php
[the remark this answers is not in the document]
gabrieljenik: Done

application/models/Plugin.php
gabrieljenik: There could be lot of malformations in config.xml files, missing config.xml files and even other situations not thought.
As so, any of those situations can be considered as a non compatibility.
By that is that we return false.

If not returning false, we should look up every usage of ->isCompatible() and add try/catch.
Still, not sure if that hadling would be much different.

Maybe, to inform the error we could add an optional parameter with the exception caught?

olleharstedt:
Ahmed argues that this fix is not needed, it's not an issue. Can you still reproduce it, @gabrieljenik? Without this fix.

olleharstedt:
Sorry, I'll close this for now.

PR #2005, opened 2021-08-13
Dev #T907b: Fix activation screen translations

gabrieljenik:


thedirtypanda (approved):

c-schmitz:
Sorry, but the texts are not proper. Text should not contain HTML.
If you need HTML in there use sprint and %s tags.

c-schmitz:
@gabrieljenik Can you fix that, please?

gabrieljenik:
> @gabrieljenik Can you fix that, please?

Yes, that's something we discussed with @olleharstedt 
This was one of the alternatives.

As the tags here are not links but very tied to the word being emphasized, I personally don't think doing the sprint alternative may be a great fit here, but add confusion or problematic situations when, for example few tags are needed. Still, I understand, content and presentation should be separate as much as possible.

We will review the sprint alternative and be back.

Thanks

olleharstedt:
> Text should not contain HTML. If you need HTML in there use sprint and %s tags.

I'll add this to the code guide. It was one of the things we were missing. :)

olleharstedt:
Wouldn't it be possible to translate HTML tags wrong, tho? E.g. if you're wrapping bold text in a link? Or maybe that should be split rather?

olleharstedt:
Code quality guide: https://manual.limesurvey.org/Code_quality_guide#Localization

This is copied from the old guide with some new styling for code syntax.

gabrieljenik:
OK, but there's nothing about mingling html there.
Can you post an example on how would you do it?

Something like this?
https://stackoverflow.com/questions/2758599/localization-with-html-tags/2759039#2759039

gabrieljenik:
> Wouldn't it be possible to translate HTML tags wrong, tho, if they only see `%s`? E.g. if you're wrapping bold text in a link? Or maybe that should be split rather?

That's what I meant as one possible confussion error.
Spliting it wouldn't be an overkill?

Including <strong> in the language file I think is acceptable. 
Including <a href="%s"> also acceptable 
Including <a href="/surveyController/%s"> not very good.
That's what I picture as border in between good and not good.

What do you think?

gabrieljenik:
Why not creating a gTR function that takes the same parameters as gt() and also an array with named parameters to be replaced?
As parameters are named, there shouldn't be errors that appear because of changing positions of strings across translations neither because translating wrongly %s.

gTR("The :bold_start: cat :bold_end: is under the table", ['bold_start'=>'<strong>', ...])

Right now the named parameters are being sent empty to the Yii:t() function. We just need to start to use them
```
function gT($sToTranslate, $sEscapeMode = 'html', $sLanguage = NULL)
{
    return quoteText(Yii::t('',$sToTranslate,array(),null,$sLanguage),$sEscapeMode);
}
```

gabrieljenik:
Kept on thinking: the only different in using ":bold_start:" and `"<strong>"` is where the html is implemented.
Can;t we just implement meta tags that are later rendered into html?

olleharstedt:
The problem is that any of the translators that work for free need to be able to understand it at once.

PR #2006, opened 2021-08-13
Fixed issue #17152: exporting list question with 'other' to SPSS does not read all data

gabrieljenik:
Porting this from master:
https://bugs.limesurvey.org/view.php?id=16939

### DIFF
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -16,12 +16,15 @@
 * Strips html tags and replaces new lines
 *
 * @param $string
+* @param $removeOther   if 'true', removes '-oth-' from the string.
 * @return string
 */
-function stripTagsFull($string)
+function stripTagsFull($string, $removeOther = true)
 {
-    $string = flattenText($string,false,true); // stripo whole + html_entities
-    $string = str_replace('-oth', '', $string);// Why ?
+    $string = flattenText($string, false, true); // stripo whole + html_entities
+    if ($removeOther) {
+        $string = str_replace('-oth-', '', $string);
+    }
     //The backslashes must be escaped twice, once for php, and again for the regexp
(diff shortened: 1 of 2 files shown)


application/helpers/export_helper.php
```diff
@@ -16,12 +16,15 @@
 * Strips html tags and replaces new lines
 *
 * @param $string
+* @param $removeOther   if 'true', removes '-oth-' from the string.
 * @return string
 */
-function stripTagsFull($string)
+function stripTagsFull($string, $removeOther = true)
```
olleharstedt, line 22: Would it be possible to use function composition instead of a new boolean argument? E.g.

stripTagsFull(remoteOther($theThing));

olleharstedt: `     $string = str_replace('-oth', '', $string);// Why ?`

Why indeed? :laughing: Do we need to test with this removed?

gabrieljenik: > Would it be possible to use function composition instead of a new boolean argument? 

Not sure. I believe not. The "removal" happens after flatennig the text, is not done initially.
Also, we did it like that as to introduce the minimum change on a function which, if remmeber correctly, is used widely.
At last, it is like this on master as well, so I would say is better to keep it like this, at least for this one.

gabrieljenik: > ` $string = str_replace('-oth', '', $string);// Why ?`
> 
> Why indeed?  Do we need to test with this removed?

Just to highlight, the "why" comment was removed.

> Do we need to test with this removed?

I think they are used for specific export functions.
If messing with that, we should test, identify errors that should appear and fix them.

olleharstedt: Urgh. Does anyone knows the correct behaviour for "-oth-" here?

PR #2007, opened 2021-08-17
Remove the LDAP URI scheme

mkofahl:
Since ldap_connect is used with the host name and port, specifying the protocol is misleading and incorrect.

Fixed issue #17519


mkofahl:
obselete with 654771867c6fdb6e739bf83c1a75be779e01b489

PR #2008, opened 2021-08-17
Autocreate LDAP users using simple bind

mkofahl:
Give AuthLDAP the possibility to create new users, even if LDAP is configured as "simple bind". In this case, the credentials of the new user are used to query LDAP attributes.

New feature #17520


mkofahl:
This pull request was accidentally created for the wrong branch. Please refer to #2009.

PR #2009, opened 2021-08-17
New feature #17520: Auto-create LDAP users without bind DN

mkofahl:
Give AuthLDAP the possibility to create new users, even if LDAP is configured as "simple bind". In this case, the credentials of the new user are used to query LDAP attributes.

New feature #17520


glimz:
Hello @mkofahl ,

Thank you for creating a PR. However, I cannot find any bug report in mantis for this issue.

For better tracking, please create a bug report so we can test and merge your PR.

Thank you

PR #2010, opened 2021-08-17
Fixed issue #17453: Strings not available for translation in GlotPress

gabrieljenik:
The question attribute names were not being translated in the summary view.

### DIFF
--- a/application/views/questionAdministration/summary.php
+++ b/application/views/questionAdministration/summary.php
@@ -211,7 +211,7 @@
                 <tr>
                     <td>
                         <strong>
-                            <?php echo $setting['caption'];?>:
+                            <?php eT($setting['caption']);?>:
                         </strong>
                     </td>
                     <td>


application/views/questionAdministration/summary.php
```diff
@@ -211,7 +211,7 @@
                 <tr>
                     <td>
                         <strong>
-                            <?php echo $setting['caption'];?>:
+                            <?php eT($setting['caption']);?>:
```
olleharstedt, line 214: This will only work if setting[caption] already is wrapped around gT or eT somewhere else. Is it?

gabrieljenik: Yes, should be already there. $setting['caption'] holds the name of the attribute.
If the literal is not anywhere else in the code and wasn't added manually, we can add it to a special fiel created for holding literals. (and allowing the translation bot to catch them)

olleharstedt: Caption in question helper is already wrapped in gT(). So eT() shouldn't be needed here...?

```
        "caption" => gT('Sort answers alphabetically'));
```

olleharstedt: And description in QuestionType.php (but also saved in database???):

```
                'description' => gT("Multiple numerical input", "html", $language),
```

gabrieljenik: Let me review

gabrieljenik: Not sure if we are on the same page.
The ticket talks about some translations missing.
Ticket already diagnoses that those aren't available for translating.
But they are. Ex: `Use slider layout` is available for translating.

Still, not showing the translations.
The fix is about showing the translations.

olleharstedt: OK, that makes sense.

PR #2011, opened 2021-08-17
Fixed issue #16720: Old/invalid question codes adjusted during LSS import do not get replaced

gabrieljenik:
There was also an issue when replacing codes on general basis.
If the code was used twice in an expression, only the last code mention was replaced.
This is fixed now as well

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -3510,7 +3510,7 @@ function replaceExpressionCodes($iSurveyID, $aCodeMap)
             // Don't search/replace old codes that are too short or were numeric (because they would not have been usable in EM expressions anyway)
             if (strlen($sOldCode) > 1 && !is_numeric($sOldCode)) {
                 $sOldCode = preg_quote($sOldCode, '~');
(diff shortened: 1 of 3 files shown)


application/helpers/common_helper.php
```diff
@@ -3523,21 +3523,47 @@ function replaceExpressionCodes($iSurveyID, $aCodeMap)
...
+                    $arQuestionLS->help = recursive_preg_replace("~{[^}]*\K{$sOldCode}(?=[^}]*?})~", $sNewCode, $arQuestionLS->help, -1, $iCount);
                     $bModified = $bModified || $iCount;
                 }
             }
             if ($bModified) {
                 $arQuestionLS->save();
             }
         }
+        // Also apply on question's default values
+        $defaultValues = DefaultValue::model()->with('defaultvaluel10ns')->findAllByAttributes(['qid' => $arQuestion->qid]);
```
olleharstedt, line 3540: In the future, blocks like these can be put in separate functions instead.

gabrieljenik: Sorry, got lost. What do you mean?
Creating a static function?

olleharstedt: No, there's no class yet, so no place to put a static method. Just a normal function. :) Or we need a new task to create a class instead.

PR #2012, opened 2021-08-17
Fixed issue #17521: HTML is not escaped in timer messages

gabrieljenik:
Escaping text. Adding automatic test.


c-schmitz:
Not sure if just quoting is correct. Should we allow maybe basic HTML here and only XSS-filter?

olleharstedt:
Another alternative could be to include an `<xss>1</xss>` attribute for question attributes in the question config.xml files.

olleharstedt:
This method filters model attributes: https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSYii_Validators.php#L113

gabrieljenik:
> Another alternative could be to include an `<xss>1</xss>` attribute for question attributes in the question config.xml files.

What would this be for? XSS-on/off depends on the attribute or on the user?

olleharstedt:
> What would this be for? XSS-on/off depends on the attribute or on the user?

Depending on the attribute. If the attribute is numeric, you don't need to filter it, for example.

gabrieljenik:
> > What would this be for? XSS-on/off depends on the attribute or on the user?
> 
> Depending on the attribute. If the attribute is numeric, you don't need to filter it, for example.

I mean, why we should trust the XSS filtering setting on the question-theme.

PR #2017, opened 2021-08-23
Fixed issue #17526: Filter html_entity_decode at save

gabrieljenik:
Added html_entity_decode to the XSS filtering process

### DIFF
--- a/application/core/LSYii_Validators.php
+++ b/application/core/LSYii_Validators.php
@@ -117,7 +117,6 @@ class LSYii_Validators extends CValidator
         /** Start to get complete filtered value with  url decode {QCODE} (bug #09300). This allow only question number in url, seems OK with XSS protection **/
         $sFiltered = $filter->purify($value);
         $sFiltered = preg_replace('#%7B([a-zA-Z0-9\.]*)%7D#', '{$1}', $sFiltered);
-        $sFiltered = str_replace("html_entity_decode", "", $sFiltered);
         Yii::import('application.helpers.expressions.em_core_helper', true); // Already imported in em_manager_helper.php ?
         $oExpressionManager = new ExpressionManager();
         /**  We get 2 array : one filtered, other unfiltered **/
@@ -138,6 +137,8 @@ class LSYii_Validators extends CValidator
                         $sNewValue .= "\"" . (string) $filter->purify($aParsedExpression[0]) . "\""; // This disallow complex HTML construction with XSS
                     } elseif ($aParsedExpression[2] == 'SQ_STRING') {
                         $sNewValue .= "'" . (string) $filter->purify($aParsedExpression[0]) . "'";
+                    } elseif ($aParsedExpression[2] == 'WORD') {
+                        $sNewValue .= str_replace("html_entity_decode", "", $aParsedExpression[0]);
                     } else {
                         $sNewValue .= $aParsedExpression[0];
                     }
--- a/application/core/LSYii_Validators.php
+++ b/application/core/LSYii_Validators.php
@@ -115,7 +115,9 @@ class LSYii_Validators extends CValidator
         $filter = LSYii_HtmlPurifier::getXssPurifier();
 
         /** Start to get complete filtered value with  url decode {QCODE} (bug #09300). This allow only question number in url, seems OK with XSS protection **/
-        $sFiltered = preg_replace('#%7B([a-zA-Z0-9\.]*)%7D#', '{$1}', $filter->purify($value));
+        $sFiltered = $filter->purify($value);
+        $sFiltered = preg_replace('#%7B([a-zA-Z0-9\.]*)%7D#', '{$1}', $sFiltered);
+        $sFiltered = str_replace("html_entity_decode", "", $sFiltered);
         Yii::import('application.helpers.expressions.em_core_helper', true); // Already imported in em_manager_helper.php ?
         $oExpressionManager = new ExpressionManager();
         /**  We get 2 array : one filtered, other unfiltered **/
--- a/tests/unit/LSYiiValidatorsTest.php
+++ b/tests/unit/LSYiiValidatorsTest.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace ls\tests;
+
+/**
+ * Test LSYii_Validators class.
+ */
+class LSYiiValidatorsTest extends TestBaseClass
+{
+    /**
+     * Test filtering of html_entity_decode.
+     */
+    public function testHtmlEntityDecodeFilter()
+    {
+        // First, we define the cases to test. Array keys are the strings to filter, and values are the expected result
+        $cases = [
+            "html_entity_decode('&amp;')" => "html_entity_decode('&amp;')", // Not an expression, so it shouldn't be changed.
+            "{html_entity_decode('&amp;')}" => "{('&amp;')}",   // Used as a function in an expression, so it should be removed.
+            "{join(\"&#123;\",'html_entity_decode(\"&amp;amp;\")',\"&#125;\")}" => "{join(\"{\",'html_entity_decode(\"&amp;amp;\")',\"}\")}",   // Inside a function but as a string, so it's not removed.
+        ];
+
+        $validator = new \LSYii_Validators();
+
+        // Test each case
+        foreach ($cases as $string => $expected) {
+            $actual = $validator->xssFilter($string);
+            $this->assertEquals($expected, $actual);
+        }
+    }
+}


Shnoulle:
I think purify can allow `&lt;script&gt;` and then : you decode it to `<script>` 

Did you test like this ?

gabrieljenik:
> I think purify can allow `&lt;script&gt;` and then : you decode it to `<script>`
> 
> Did you test like this ?

Nope. But I think that if you try to store `&lt;script&gt;` it will be saved as `&amp;lt;script&amp;gt;`

Shnoulle:
No `&lt;script&gt;alert('XSS')&lt;/script&gt;` stay `&lt;script&gt;alert('XSS')&lt;/script&gt;`
Proof
Peek 24-08-2021 14-25


I already find and try some hack with this awfull html_entity_decode in EM

Shnoulle:
Then : with this commit : 

1. Allow adding XSS by `&lt;script&gt;alert('XSS')&lt;/script&gt; `
2. Disallow showing `<strong>` adding `&lt;strong&gt;` (use case : quiz on HTML tag)

gabrieljenik:
> Then : with this commit :
> 
> 1. Allow adding XSS by `&lt;script&gt;alert('XSS')&lt;/script&gt; `
> 2. Disallow showing `<strong>` using `&lt;strong&gt;` (use case : quiz on HTML tag)

Sorry, I don't understand. Are those bugs? Were they present before? If so, this should be registered as another bug to allow this to be merged.

Shnoulle:
Oh ! You remove `html_entity_decode`  string from the string :)

Sorry : bad reading commit.

Maybe remove it via regexp : must be between `{` and `}`

Else it's a API version broker ;)

Shnoulle:
Please :

1. comment the commit
2. comment the line to inform why we remove this string

gabrieljenik:
> Else it's a API version broker ;)

What do you mean?

application/core/LSYii_Validators.php
```diff
@@ -115,7 +115,9 @@ public function xssFilter($value)
         $filter = LSYii_HtmlPurifier::getXssPurifier();
 
         /** Start to get complete filtered value with  url decode {QCODE} (bug #09300). This allow only question number in url, seems OK with XSS protection **/
-        $sFiltered = preg_replace('#%7B([a-zA-Z0-9\.]*)%7D#', '{$1}', $filter->purify($value));
+        $sFiltered = $filter->purify($value);
+        $sFiltered = preg_replace('#%7B([a-zA-Z0-9\.]*)%7D#', '{$1}', $sFiltered);
```
Shnoulle, line 119: Move the previous comment here in my opinion

application/core/LSYii_Validators.php
```diff
@@ -115,7 +115,9 @@ public function xssFilter($value)
...
+        $sFiltered = str_replace("html_entity_decode", "", $sFiltered);
```
Shnoulle, line 120: Need comment

Shnoulle (changes requested):
1. Minima : add comment
2. Maybe preg replace to replace it only inside { } ?

Shnoulle:
> > Else it's a API version broker ;)
> 
> What do you mean?

Before this commit : a survey work with `html_entity_decode` and can be edited
After this commit : a survey with `html_entity_decode` can not be edited and can broke

Then : it broke : [you make incompatible API changes](https://semver.org/),

gabrieljenik:
> After this commit : a survey with html_entity_decode can not be edited and can broke

Sorry, can you elaborate more?

Shnoulle:
> > After this commit : a survey with html_entity_decode can not be edited and can broke
> 
> Sorry, can you elaborate more?

You have a survey using {html_entity_decode(EQUATION.NAOK)} : it work even after copying, import, edited etc …

Update : this surver didn't work if XSS is activated  {html_entity_decode(EQUATION.NAOK)} is replaced by  {(EQUATION.NAOK)} after edition with XSS.

The purpose of the commit create the API incompatible API changes.

Shnoulle:
My opinion still : move this function in core plugin : activated when update, not activated when install.

Still API change : but you have a way to keep API ok.

gabrieljenik:
When you say API you mean RemoteControl?
The behaviour will happen not only for API changes but also for changes done using the UI.
So, yes it may affect prior surveys, but not only for API, but for any kind of change.

@olleharstedt @c-schmitz  comments?

olleharstedt:
> The purpose of the commit create the API incompatible API changes.

Yes, since we change which functions are available inside EM.

olleharstedt:
The "correct" thing to do is to dump major version to 6, since can potentially break some surveys.

Guess we need some kind of warning if the removed function is used in a survey?

Shnoulle:
> The "correct" thing to do is to dump major version to 6, since it can potentially break some surveys.
> 
> Guess we need some kind of warning if the removed function is used in a survey?

See my comment 

1. Move it to a plugin : [really easy task.](https://github.com/LimeSurvey/LimeSurvey/tree/master/plugins/Demo/demoAddEmFunction)
2. Activate the plugin when update
3. Don't activate when install

:)

c-schmitz:
We should bind this to the XSS filter and filter the function on editing/saving survey by XSS-filtered users.
So only super admins will be able to use that function.

gabrieljenik:
> We should bind this to the XSS filter and filter the function on editing/saving survey by XSS-filtered users.
> So only super admins will be able to use that function.

That's the current PR like. right?

application/core/LSYii_Validators.php
olleharstedt: Good idea.

application/core/LSYii_Validators.php
olleharstedt: Agree. Feel free to link to the 17526 issue with a @see tag.

olleharstedt: Also, shouldn't this only apply to expressions? It should still be possible to include "html_entity_decode" as a normal word.

olleharstedt:
> > We should bind this to the XSS filter and filter the function on editing/saving survey by XSS-filtered users.
> > So only super admins will be able to use that function.
> 
> That's the current PR like. right?

Uh, right. No change needed. :) Except the review we made.

application/core/LSYii_Validators.php
gabrieljenik: Well, yes, but I believe we don't have EM filtering, do we?
I mean, detecting keywords inside expression is not straightforward. Possible, but not straightforward and could be very error prone.

Taking into account what you say, maybe we should just disable the internal code for html_entity_decode, and have the EM function to return as output what it receives. That is not the fancier, but could be effective.

Thoughts?

Shnoulle: We already work inside Expression manager :  https://github.com/LimeSurvey/LimeSurvey/blob/3d0a6091c9ec0af96a8b5cb97688bff20bb7926f/application/core/LSYii_Validators.php#L142 

Maybe here it's the function ?

gabrieljenik: Yes, probably that would be a starting point. Will review it. 

> Taking into account what you say, maybe we should just disable the internal code for html_entity_decode, and have the EM function to return as output what it receives. That is not the fancier, but could be effective.

This wouldn't work, sorry. It wou;dn't be filtering, but disabling the whole function for everyone.

Shnoulle:
Still my opinion : there are no clean way to fix it …

Completely allow or completely disallow …

olleharstedt:
> Still my opinion : there are no clean way to fix it …
> 
> Completely allow or completely disallow …

Yeah. Let's start with this solution and then see what people say, if anyone is even using it.

gabrieljenik:
> Yeah. Let's start with this solution and then see what people say, if anyone is even using it.

OK, even with the fact that regular text which says "html_entity_decode" will be filtered, right?
Or you are thining of a more complex solution?

olleharstedt:
> OK, even with the fact that regular text which says "html_entity_decode" will be filtered, right?
> Or you are thining of a more complex solution?

I was thinking of only filtering for this word inside expressions. Do we have a way to detect expressions easily? Or just another regexp for { and }?

Shnoulle:
> > OK, even with the fact that regular text which says "html_entity_decode" will be filtered, right?
> > Or you are thining of a more complex solution?
> 
> I was thinking of only filtering for this word inside expressions. Do we have a way to detect expressions easily? Or just another regexp for { and }?

Since we already filter `concat("<strong>",STRING,"</strong>"')` for exemple.

Look here : https://github.com/LimeSurvey/LimeSurvey/blob/3d0a6091c9ec0af96a8b5cb97688bff20bb7926f/application/core/LSYii_Validators.php#L133

I think we can be sure to fix it here : https://github.com/LimeSurvey/LimeSurvey/blob/3d0a6091c9ec0af96a8b5cb97688bff20bb7926f/application/core/LSYii_Validators.php#L142 

I think function are WORD in token

gabrieljenik:
> I think function are WORD in token

Ok, will review and be back

gabrieljenik:
What do you think now?

Shnoulle:
> What do you think now?

Did you test ? What happen for

1.  `html_entity_decode is a PHP function (no expression)`
2. `html_entity_decode is a Expression manager function : for example {html_entity_decode("&amp;"}}`
3. ` html_entity_decodecan be used like it in expression manager : {join("&#123;",'html_entity_decode("&amp;amp;")',"&#125;")}`

about 3 : if it replace it's not very important i think ;)

olleharstedt:
Yeah, blocking use-case 3 is OK.

olleharstedt:
Also @Shnoulle , I tried in a number of ways to check if it's possible to call a string as a function in EM, but couldn't do it. So it's safe to assume stuff like

```
{"html_" + "entity_decode"("moo")}
```

is not possible?

Shnoulle:
> Also @Shnoulle , I tried in a number of ways to check if it's possible to call a string as a function in EM, but couldn't do it. So it's safe to assume stuff like
> 
> ```
> {"html_" + "entity_decode"("<script>alert(1);</script>")}
> ```
> 
> is not possible?

`"<script>alert(1);</script>"` must be filtered (to `alert(1);`) BUT not `"&gt;script&lt;alert(1);&gt;/script&lt"` currently. (I i don't make error).

Then : a admin user can add `{entity_decode("&gt;script&lt;alert(1);&gt;script&lt;")}`

Shnoulle:
Oh better understand

> call a string as a function in EM

No warranty , but i really don't find a way to do it (without plugin :D )

gabrieljenik:
Added a unit test as to review how it behaves on each case.
If it is OK, we make it the same for LTS.

olleharstedt:
> Inside a function but as a string, so it's not removed

This won't be possible to hack round?

gabrieljenik:
> > Inside a function but as a string, so it's not removed
> 
> This won't be possible to hack round?

I wasn't able to, as there is no "eval" function on EM.
Still, not a definite answer. Who can says is 100% safe?

PR #2019, opened 2021-08-24
Fixed issue #17535: Error when logging in with LDAP

wolbernd:
Fixed issue #17535:


Shnoulle:
Need a test to avoid such issue , no ?

olleharstedt:
Back from vacation, Denis? :) We've missed you.

> Need a test to avoid such issue , no ?

Psalm checks the availability of functions, but it's only used on parts of the code-base.

olleharstedt:
Thanks for the fix!

olleharstedt:
Oh, someone already pushed a fix to master before this...

Shnoulle:
> Back from vacation, Denis? :) We've missed you.
> 

Not officially, but have email.

PR #2025, opened 2021-08-31
[#17541] fixed, added ext + functions

pop1989bb:
Fixed issue #17541: Fixed by adding the missing environment variable EXT, adding features to ConsoleApplication.php and the required asset directory.
Dev: pop1989_bb

### DIFF
--- a/application/core/ConsoleApplication.php
+++ b/application/core/ConsoleApplication.php
@@ -149,4 +149,14 @@ class ConsoleApplication extends CConsoleApplication
         return $this->getComponent('pluginManager');
     }
 
+    public function getAssetManager()
+    {
+        return $this->getComponent('assetManager');
+    }
+
+    public function getClientScript()
+    {
+        return $this->getComponent('clientScript');
+    }
+
 }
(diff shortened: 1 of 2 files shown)


application/core/ConsoleApplication.php
```diff
@@ -149,4 +149,14 @@ public function getPluginManager()
         return $this->getComponent('pluginManager');
     }
 
+    public function getAssetManager()
```
olleharstedt, line 152: Not sure these functions are needed, you should always be able to do: Yii::app()->assetManager, etc for other components.

pop1989bb: We are using the ExportSurveyResultsService class from Limesurvey. This class is calling some internal functions which requires the added methods.

olleharstedt: Hm, alright. Can you then please add docblocks? Like:

```
/**
 * @return CComponent
 */
 ```

Shnoulle: asset manager and script are for web construction, then surely not loaded in console.

https://github.com/LimeSurvey/LimeSurvey/blob/3f1f5f17c3d7dd0aa7885d1bcbad583772ae3ccf/framework/web/CWebApplication.php#L149
Nothink in https://github.com/LimeSurvey/LimeSurvey/blob/master/framework/console/CConsoleApplication.php

PR #2027, opened 2021-09-01
Fixed issue #17454: Unable to remove a user with survey permissions with friendly url enabled

gabrieljenik:
Changing the order of params, just in case the username param adds confusion to the querystring


Shnoulle:
Did we need username ?
Else : maybe try with urlencode ?

PR #2028, opened 2021-09-02
Fixed issue #17527: Validate before using updateAll

gabrieljenik:
Reviewed updateAll() invocations. Added xss filtering.

### DIFF
--- a/application/controllers/admin/database.php
+++ b/application/controllers/admin/database.php
@@ -1448,11 +1448,14 @@ class database extends Survey_Common_Action
                     /* Date management */
                     Yii::app()->loadHelper('surveytranslator');
 
-
+                    $oValidator = new LSYii_Validators;
                     foreach ($validAttributes as $validAttribute) {
                         if ($validAttribute['i18n']) {
                             foreach ($aLanguages as $sLanguage) {
                                 $value = Yii::app()->request->getPost($validAttribute['name'].'_'.$sLanguage);
+                                if ($oValidator->xssfilter) {
+                                    $value = $oValidator->xssFilter($value);
+                                }
                                 $iInsertCount = QuestionAttribute::model()->findAllByAttributes(array('attribute'=>$validAttribute['name'], 'qid'=>$this->iQuestionID, 'language'=>$sLanguage));
                                 if (count($iInsertCount) > 0) {
                                     if ($value != '') {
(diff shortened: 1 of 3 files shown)


Shnoulle:
Why not extend directly https://github.com/LimeSurvey/LimeSurvey/blob/2fd0581a95445cc2f5334a07be8be52f5f964a31/framework/db/ar/CActiveRecord.php#L1768 ?
for some specific model ?

gabrieljenik:
> Why not extend directly
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/2fd0581a95445cc2f5334a07be8be52f5f964a31/framework/db/ar/CActiveRecord.php#L1768
> 
> ?
> for some specific model ?

What if we override updateAll for all model.
If the model has a static attribute xssFilterAttributes, then it is used for knowing which attributes should be filtered.

Shnoulle:
> > Why not extend directly
> > https://github.com/LimeSurvey/LimeSurvey/blob/2fd0581a95445cc2f5334a07be8be52f5f964a31/framework/db/ar/CActiveRecord.php#L1768
> > 
> > ?
> > for some specific model ?
> 
> What if we override updateAll for all models.
> If the model has a static attribute xssFilterAttributes, then it is used for knowing which attributes should be filtered.

With one model : less chance to break different things,, but yes it's a good idea.

Start by one or two model, when it work good : update https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/LSActiveRecord.php

gabrieljenik:
OK, following last conversation, I have applied a more general approach.
- Overloaded updateAll()
- Added xssFilterAttributes atribut on LSActiveRecord
- That attirbute is used by updateAll() to know which values should be xssFiltered.

What do you think?

Shnoulle:
Personally : i like it.

I totally understand we can not read rules currently.

Shnoulle:
@glimz this one is merged no ?

See https://bugs.limesurvey.org/view.php?id=17699 issue

Shnoulle (commented):
https://bugs.limesurvey.org/view.php?id=17699

olleharstedt:
Revert needed? Either that or a fix on Monday.

Shnoulle:
See for reverting, but in my opinion : yes

See master not merged : https://github.com/LimeSurvey/LimeSurvey/pull/2086

PR #2029, opened 2021-09-02
Fixed issue #16792: Delete permission on Survey : no CRSF control

gabrieljenik:
In LTS there were 6 places that had to be corrected. In master there are 12.
- Only one of them is an LSBaseController (the rest are Survey_Common_Action). So you can only work on the filter () for the LSBaseController.
- For the rest, we did the same as in LTS.


c-schmitz:
Please resolve the conflicts. Thank you.

olleharstedt:
Thanks!

PR #2030, opened 2021-09-06
Fixed issue #17512: Survey Logic File Syntax Error Summary Displays Incorrect Error Count

gabrieljenik:
Basically what LimeExpressionManager :: ShowSurveyLogicFile () does is process each string and check for errors. That is, for each "field" to validate, execute $ LEM-> ProcessString (...), and then check $ LEM-> em-> HasErrors ().

The problem is that each expression contained in the chain is processed by RDP_Evaluate (), and one of the first things it does is reset the errors.

So if a string has more than one expression, and the last expression has no errors, $ LEM-> em-> HasErrors () returns false.

So HasErrors () reports only about the last execution of RDP_Evaluate (), it does not report about the entire process of ProcessString ().

On this fix we added the possibility (optional) of not reseting the errors.
That is, allow HasErrors() to report about the whole ProcessString() process.

An alternative is to create a separate flag for checking errors in the whole string.
Still, I believe that would more intrusive, involve more changes and also would be more error prone.

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -523,19 +523,19 @@ class ExpressionManager
      * Main entry function
      * @param string $expr
      * @param boolean $onlyparse - if true, then validate the syntax without computing an answer
-     * @param boolean $resetErrors - if true (default), EM errors will be clear before evaluation
+     * @param boolean $resetErrorsAndWarnings - if true (default), EM errors and warnings will be cleared before evaluation
      * @return boolean - true if success, false if any error occurred
      */
-    public function RDP_Evaluate($expr, $onlyparse = false, $resetErrors = true)
+    public function RDP_Evaluate($expr, $onlyparse = false, $resetErrorsAndWarnings = true)
     {
         $this->RDP_expr = $expr;
         $this->RDP_tokens = $this->RDP_Tokenize($expr);
         $this->RDP_count = count($this->RDP_tokens);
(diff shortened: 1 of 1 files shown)


Shnoulle:
Need to reset warning too ?

See dedicated function : https://github.com/LimeSurvey/LimeSurvey/commit/b43c81d737f2f341bb68cb612b305b70c08e5a23#

gabrieljenik:
> Need to reset warning too ?
> 
> See dedicated function : [b43c81d](https://github.com/LimeSurvey/LimeSurvey/commit/b43c81d737f2f341bb68cb612b305b70c08e5a23)

Do you mean here reset warning here as well?
https://github.com/LimeSurvey/LimeSurvey/pull/2030/files#diff-a974f1aca3b64170e04f9ac415aadb10d9108696f0ae4ce07caee89d8a1447d3R538

Shnoulle:
Yes : in same condition :  if question have  

{Q0 = "A1"} for question  text and {if(Q1 == "A3", "a sentence")} : question must show 'this question have a warning" and top must show "A question have a warning"

Shnoulle:
> Yes : in same condition : if question have
> 
> {Q0 = "A1"} for question text and {if(Q1 == "A3", "a sentence")} : question must show 'this question have a warning" and top must show "A question have a warning"

Must be checked , but it take time to check and report.

gabrieljenik:
What about now?

thedirtypanda (approved):

PR #2039, opened 2021-09-16
Fixed issue: 8. Titles Optimize titles of modals - [XLS - Design Inco…

thedirtypanda:
Design inconsistencies from (internal) sheet. It contains modal changes.


olleharstedt:
Seems like there's a lot of code duplication in this widget?

thedirtypanda:
Yeah, I talked toPatricia about that. Its the easiest solution right now.

In the future we need to add parameters for a base-modal. which renders different footers. I added a task for the next sprint (refactoring out code duplication and add parameters, base-modal)

olleharstedt:
> Yeah, I talked toPatricia about that. Its the easiest solution right now.
> 
> In the future we need to add parameters for a base-modal. which renders different footers. I added a task for the next sprint (refactoring out code duplication and add parameters, base-modal)

Alright. Code looks OK, then. Maybe add an internal task to remove code duplication for this widget? You can suggest it for the next sprint.

PR #2040, opened 2021-09-17
Fix (17603): Removed language parameter from getQuestions method

kpko:
Removed language parameter from getQuestions method and query, as this column has been moved to another table (questionl10ns)

Fixed issue #17603

### DIFF
--- a/application/libraries/PluginManager/LimesurveyApi.php
+++ b/application/libraries/PluginManager/LimesurveyApi.php
@@ -397,14 +397,12 @@ class LimesurveyApi
 
     /**
      * @param int $surveyId
-     * @param string $language
      * $param array $conditions
      * @return \Question[]
      */
-    public function getQuestions($surveyId, $language = 'en', $conditions = array())
+    public function getQuestions($surveyId, $conditions = array())
     {
         $conditions['sid'] = $surveyId;
-        $conditions['language'] = $language;
         return \Question::model()->with('subquestions')->findAllByAttributes($conditions);
     }
 


Shnoulle:
Yes, but before : getQuestions return the question object with language,

Maybe a solution to return the object with language part ? Help and Question ?

PS : i don't use any of this function ;)

application/libraries/PluginManager/LimesurveyApi.php
```diff
@@ -397,14 +397,12 @@ public function getParticipant($iParticipantID)
...
-     * @param string $language
      * $param array $conditions
      * @return \Question[]
      */
-    public function getQuestions($surveyId, $language = 'en', $conditions = array())
+    public function getQuestions($surveyId, $conditions = array())
     {
         $conditions['sid'] = $surveyId;
-        $conditions['language'] = $language;
         return \Question::model()->with('subquestions')->findAllByAttributes($conditions);
```
olleharstedt, line 406: Keep language argument, but join with question_l10ns too

olleharstedt (changes requested):

kpko:
Hi, I really don't know that much about the LimeSurvey code base. The suggested fix was the one we used as we only use single language surveys and LimeSurvey was crashing when calling getQuestions before. If there is a better fix, we should go for it.

Maybe it would be better to close this pull request and open up another issue about this to implement a better solution by one of the core contributors? 

Thanks so much for the good work. :)

glimz:
Hello, thank you for your contribution. However, since the PR does not conform to our code guidelines, we would close the issue and leave it open for another contributor to pick it up.

PR #2046, opened 2021-09-21
Fixed issue #17504: Responses Batch deletion modal only opens the first time

gabrieljenik:
Besides response's Batch Delete, this also happens on:

Box delete (homepageSettings/index)
Theme extend (themeOptions/index)
Theme uninstall (themeOptions/index)
Theme reset (themeOptions/index)
Available survey theme delete (themeOptions/index)

This solution should apply to all of those places.


assets/packages/adminbasics/src/components/confirmdeletemodal.js
```diff
@@ -3,7 +3,7 @@ const ConfirmDeleteModal = function (options) {
 
     options.fnOnShown = options.fnOnShown || function () {};
     options.fnOnHide = options.fnOnHide || function () {};
-    options.removeOnClose = options.removeOnClose || function () {};
+    options.removeOnClose = options.removeOnClose || true;
```
olleharstedt, line 6: Hm, isn't `removeOnClose` a function? How does that work with `true`?

gabrieljenik: Is is used as boolean throughout the code.
image

olleharstedt: OK :)

PR #2050, opened 2021-09-22
Fixed issue 17557: corrected erroneous error message concerning SVG file type as allowed file type for upload

gwdgls:
Fixed issue #17557:

### DIFF
--- a/application/config/version.php
+++ b/application/config/version.php
@@ -11,10 +11,10 @@
  * See COPYRIGHT.php for copyright notices and details.
  */
 
-$config['versionnumber'] = '5.1.8';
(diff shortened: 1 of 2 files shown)


Shnoulle:
Hard to fix … because sentence depends on `allowedthemeimageformats`

must work with `$config['allowedthemeimageformats'] = 'gif,ico,jpg,png,svg';` but for  `$config['allowedthemeimageformats'] = 'jpg,png';` too.

Maybe `sprintf(gT("This file is not a supported image - please only upload this type images : %s"),$config['allowedthemeimageformats']);` ?

c-schmitz:
Thanks, we fixed it now for 5.x.

PR #2052, opened 2021-09-23
Menu entries in the left side are not selected anymore

thedirtypanda:
When you navigate on deeper chapters of an Menu Entry e.g. Statistics - Simple mode the left side navigation is not selected anymore - This accounts for allot of Menu Entry's on left side.


assets/packages/adminsidepanel/src/components/sidebar.vue
```diff
@@ -352,22 +354,39 @@ export default {
             }
 
             this.currentTab = tab;
-        }
-    },
```
olleharstedt, line 356: This change here is probably wrong, one `}` too much removed.

PR #2054, opened 2021-09-28
Fixed issue #17569: Unable to edit answer in single choice question after activate survey

gabrieljenik:
Making LS5 to allow answer options updates as it was on LS3.

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -445,10 +445,19 @@ class QuestionAdministrationController extends LSBaseController
                 SettingsUser::deleteUserSetting('question_default_values_' . $questionData['question']['type']);
             }
 
-            // Clean subquestions and answer options before save.
+            // Clean answer options before save.
             // NB: Still inside a database transaction.
+            $question->deleteAllAnswers();
+            // If question type has answeroptions, save them.
+            if ($question->questionType->answerscales > 0) {
+                $this->storeAnswerOptions(
+                    $question,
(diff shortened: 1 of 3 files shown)


olleharstedt:
Hm, is this still behaving correctly when survey is activated? Not possible to hack the POST?

Shnoulle:
> Hm, is this still behaving correctly when survey is activated?

It's the issue 17569 : in 3.X (since 1.40 maybe): it available because it's not an issue : it don't broke database

olleharstedt:
> > Hm, is this still behaving correctly when survey is activated?
> 
> It's the issue 17569 : in 3.X (since 1.40 maybe): it available because it's not an issue : it don't broke database

Oh, so it's possible to add new answer options without problems, even in activated survey? Even for ranking question type?

Shnoulle:
> > > Hm, is this still behaving correctly when survey is activated?
> > 
> > 
> > It's the issue 17569 : in 3.X (since 1.40 maybe): it available because it's not an issue : it don't broke database
> 
> Oh, so it's possible to add new answer options without problems, even in activated survey? Even for ranking question type?

Yes and yes. For ranking it's the reason of this attribute : https://github.com/LimeSurvey/LimeSurvey/blob/7c8357130f627e314ccc493d8a2e694c1ffd1108/application/helpers/questionHelper.php#L526 (`'readonly_when_active'=>true`)

olleharstedt:
Alright, thanks. :)

olleharstedt:
Any PR that removes more code than it adds is a successful PR. :star2:

PR #2055, opened 2021-09-28
Fixed issue #T1380: Central Participant management - Export all participants modal

gabrieljenik:
Adding 'application/x-www-form-urlencoded' as content-type.

Seems it is the content-type that is sent also when using the same method for individual export.
But when exporting all participants, that content-type was missing.


thedirtypanda:
@gabrieljenik You forgot to change the title. I did it for you :)

thedirtypanda (approved):

gabrieljenik:
> @gabrieljenik You forgot to change the title. I did it for you :)

Thanks!

PR #2057, opened 2021-09-29
Fixed issue: List Surveys - Dropdown cut off

thedirtypanda:
In the list survey view the dropdown is broken. There should be other options for example delete. It affects latest develop.
The Dropdown height changes with the amount of surveys inside the list.


Shnoulle:
`overflow-x: auto; ` or `scroll` is not better ?

thedirtypanda:
So `overflow-x: auto;` introduced the bug. I set it to the default behaviour which is `overflow-x: visible;`.

PR #2059, opened 2021-09-29
Fixed issue #17593: Generated answer code is too long

gabrieljenik:


assets/scripts/admin/questionEditor.js
```diff
@@ -929,7 +929,7 @@ $(document).on('ready pjax:scriptcomplete', function () {
     // TODO: Doc answers
     const answers = {};
     const lsrows = $('#quickaddarea').val().split('\n');
-    const allrows = $closestTable.find('tr').length;
+    const allrows = $closestTable.find('tbody tr').length;
```
Shnoulle, line 932: ? needed ? A row is a tr no ?

gabrieljenik: The header rows are also TR. We only need the body rows. That's why we filter the body in the CSS selector

Shnoulle: Oups : it's a selector : totaly right , sorry !

PR #2064, opened 2021-10-01
New feature #17640: Question theme - bootstrap styling of dropdown boxes

adamzammit:
New feature #17640: Question theme - bootstrap styling of dropdown boxes
Dev: Requires flattenText to be applied at template level not code level
Dev: Adjustments in default template made to accommodate this

bootstrap


Shnoulle:
What happen without JS ?
Keyboard navigation action is OK too ?

(I think to do such theme but : without javascript : show a list of radio :)

adamzammit:
Thanks for the suggestion @Shnoulle - fix applied to handle javascript degradation. Also tested keyboard navigation and it worked.

Shnoulle:
Thanks :)

I must test, if you look at ranking we have both : 

1. flat text in <option> : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/survey/questions/answer/ranking/rows/answer_row.twig#L20
2. not flat text in list :https://github.com/LimeSurvey/LimeSurvey/blob/720c3dac0790c78d4623de0e448b684ea1f04d8f/application/views/survey/questions/answer/ranking/answer.twig#L46

Shnoulle:
@adamzammit do you have a 5.X QuestionTheme version for this ?

adamzammit:
> @adamzammit do you have a 5.X QuestionTheme version for this ?

No - I have a 3.x version but even that one requires a modification to the core code as the core code will "flatten" the text of the dropdown box before it sends it to twig to render so all styling will be lost.

Shnoulle:
Doing some test , some cosmetic issue.

1. Empty choice show as «nothing seleted»
2. Width of dropdown is updated each time something is selected : maybe can set it as start ? Or force to be at width:100% ?
3. Long text goes out of screen : max-width work partially , maybe chekc with 2 ?

Video attached

Peek 1
Peek 2
Peek 3

Shnoulle (changes requested):
Find some way to fix cosmectic issue ?

Shnoulle:
Oh … 3 is not an issue : default theme have same issue.
For 2 : maybe just a min width ?

Shnoulle (approved):
Seems great !

Shnoulle:
theme class name is not added to question wrapper (i think it was case).
Adding a specific class name to the wrapper can allow update by custom.css or Survet Theme css.

(i test with :
```
.bootstrap-select .dropdown-toggle .filter-option-inner-inner{min-height: 1em;}
.bootstrap-dropdown-wrapper .dropdown-item{max-width:100%}
.form-inline .bootstrap-select, .form-inline .bootstrap-select.form-control{max-width:100%}
.form-inline .bootstrap-select .dropdown-menu > li > a{white-space:initial}
 ```

but unsure all user want this …
)

Shnoulle (approved):
The last comment are more improvement than need change.

PR #2069, opened 2021-10-01
Fixed issue #T1383: Survey lists - Survey group modal

thedirtypanda:
Change title to "Change survey group" and button to "Change" (green)


olleharstedt:
Same issue as before, increasing code duplication inside this widget. Did we have a task for that?

PR #2072, opened 2021-10-04
Fixed issue: Survey participant data with quotes are not properly exported in CSV

ViliusS:
Dev: Participant's name could contain quotes, for example if the participant is a company. Quotes could also be present in attribute fields. This fix allows to export quote characters in case these field contain them.

Dev: This ONLY fixes quotes, but there are also other characters like new lines, which still need to be properly escaped! The proper way to do CSV export would be to use fputcsv() and then redirect everything to php://output which automatically takes care of everything.

### DIFF
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -2361,8 +2361,8 @@ function tokensExport($iSurveyID)
         }
 
         $tokenoutput .= '"' . trim($brow['tid']) . '",';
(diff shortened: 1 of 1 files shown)


ViliusS:

c-schmitz:
Thank you. 
For small bug fixes it is fine not to have a bug report.
For big fixes please always create a bug report on bugs.limesurvey.org - you can use your normal limesurvey.org login for that site.

PR #2073, opened 2021-10-04
Fixed issue: Survey participant data with quotes are not properly exported in CSV

ViliusS:
Dev: Participant's name could contain quotes, for example if the participant is a company. Quotes could also be present in attribute fields. This fix allows to export quote characters in case these field contain them.

Dev: This ONLY fixes quotes, but there are also other characters like new lines, which still need to be properly escaped! The proper way to do CSV export would be to use fputcsv() and then redirect everything to php://output which automatically takes care of everything.

Dev: This is a backport to LTS-3.x from https://github.com/LimeSurvey/LimeSurvey/pull/2072

### DIFF
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -2241,8 +2241,8 @@ function tokensExport($iSurveyID)
         }
 
         $tokenoutput .= '"'.trim($brow['tid']).'",';
-        $tokenoutput .= '"'.trim($brow['firstname']).'",';
-        $tokenoutput .= '"'.trim($brow['lastname']).'",';
+        $tokenoutput .= '"'.str_replace('"', '""', trim($brow['firstname'])).'",';
+        $tokenoutput .= '"'.str_replace('"', '""', trim($brow['lastname'])).'",';
         $tokenoutput .= '"'.trim($brow['email']).'",';
         $tokenoutput .= '"'.trim($brow['emailstatus']).'",';
         $tokenoutput .= '"'.trim($brow['token']).'",';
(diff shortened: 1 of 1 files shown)


glimz:
@ViliusS I cannot find any Mantis Bug report for this issue. Can you create one so we can test and merge your PR? Thank you

ViliusS:

glimz:
Hello @ViliusS , we cannot directly merge your PR without a bug report. You have to [report a bug](https://bugs.limesurvey.org/) and provide sufficient details so our QA can reproduce the issue.

This will enable us properly track all PRs and bugs reported and improve the workflow for testing and merging new PR.

You can easily create an account and report a bug.

We will pick up the issue directly from mantis.

Thank you

ViliusS:
I understand you want me to create a bug report, but I don't understand WHY you need me to do this. Instead of just a report I have provided you with complete fix. All the information what this PR fixes is in the commit message.

Shnoulle:
Not related to discussion : why we don't use [fputcsv](https://www.php.net/manual/en/function.fputcsv.php) here ?

ViliusS:
> Not related to discussion : why we don't use [fputcsv](https://www.php.net/manual/en/function.fputcsv.php) here ?

As I have mentioned in the commit, yes, fputcsv() is the proper way to do this. However it needs almost full rewrite of the tokensExport() and could need a lot more testing, which is not ideal for LTS branch.

I will leave that to somebody else, who has more time. My patch is a quick fix with minimal risk.

Shnoulle:
> As I have mentioned in the commit, yes, fputcsv() is the proper way to do this. However it needs almost full rewrite of the tokensExport() and could need a lot more testing, which is not ideal for LTS branch.
> 
> I will leave that to somebody else, who has more time. My patch is a quick fix with minimal risk.

:+1: must create a dev issue :).

c-schmitz:
Thank you!

PR #2077, opened 2021-10-04
Fixed issue #17416: Lost of Total and Remaining in multi numeric

gabrieljenik:
- Re-applied the fix for issue 15978

It was resolved on ticket 15978 last year, but then the merge was reversed on commit 293947c5. I don't know if the revert was done for that particular fix, or if it was for something else. Nevertheless, here is the fix.

### DIFF
--- a/application/core/QuestionTypes/MultipleNumericalQuestion/RenderMultipleNumerical.php
+++ b/application/core/QuestionTypes/MultipleNumericalQuestion/RenderMultipleNumerical.php
@@ -318,6 +318,23 @@ class RenderMultipleNumerical extends QuestionBaseRenderer
             $dynamicTemplate = "/survey/questions/answer/multiplenumeric/rows/dynamic_slider.twig";
         }
 
+        $displaytotal     = false;
+        $equals_num_value = false;
+        if (trim($this->getQuestionAttribute('equals_num_value')) != ''
+        || trim($this->getQuestionAttribute('min_num_value')) != ''
+        || trim($this->getQuestionAttribute('max_num_value')) != ''
+        ) {
+            $qinfo = LimeExpressionManager::GetQuestionStatus($this->oQuestion->qid);
+
+            $sumRemainingEqn = LimeExpressionManager::ProcessString('{'.$qinfo['sumRemainingEqn'].'}', $this->oQuestion->qid);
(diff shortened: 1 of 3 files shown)


application/core/QuestionTypes/MultipleNumericalQuestion/RenderMultipleNumerical.php
```diff
@@ -318,6 +318,23 @@ public function render($sCoreClasses = '')
...
 
+        $displaytotal     = false;
+        $equals_num_value = false;
+        if (trim($this->getQuestionAttribute('equals_num_value')) != ''
+        || trim($this->getQuestionAttribute('min_num_value')) != ''
+        || trim($this->getQuestionAttribute('max_num_value')) != ''
+        ) {
+            $qinfo = LimeExpressionManager::GetQuestionStatus($this->oQuestion->qid);
+
+            $sumRemainingEqn = LimeExpressionManager::ProcessString('{'.$qinfo['sumRemainingEqn'].'}', $this->oQuestion->qid);
```
Shnoulle, line 329: Maybe better to use https://github.com/LimeSurvey/LimeSurvey/blob/720c3dac0790c78d4623de0e448b684ea1f04d8f/application/core/LS_Twig_Extension.php#L513 in twig file ?

Then : a QuestionTheme can check if it's fixed number to do «««some action»»» for example ?

Unsure on the best way : but i like to send raw attributes to QuestionTheme , think it's better for adaptation.

application/views/survey/questions/answer/multiplenumeric/rows/dynamic_slider.twig
```diff
@@ -22,7 +22,7 @@
                 </div>
             {% endif %}
             <div id="remainingvalue_{{ id }}" class="form-control-static numeric dynamic-remaining" data-number="1"><!-- alternative class : form-control : display like an input:text -->
-                {{ '{'~sumRemainingEqn~'}' }}
+                {{ sumRemainingEqn }}
```
Shnoulle, line 25: No way to send XSS with XSS activated ?

gabrieljenik: Can you ellaborate more or give an example?

olleharstedt:
Hm, CI was never run for this PR. Strange.

PR #2083, opened 2021-10-05
Fixed issue #17610: Token does not get marked as used if invalid email address exists

gabrieljenik:


application/helpers/frontend_helper.php
```diff
@@ -428,7 +428,9 @@ function submittokens($quotaexit = false)
         }
         $token->usesleft--;
     }
-    $token->save();
+    // We need to specify the attributes we are updating because validation could fail on other (untouched) attributes,
+    // preventing the update (eg.: the participant could have an invalid email).
+    $token->save(true, ['completed', 'usesleft']);
```
Shnoulle, line 433: :+1:  maybe https://www.yiiframework.com/doc/api/1.1/CActiveRecord#saveAttributes-detail is better ?

olleharstedt: saveAttributes does not run validation, it seems.

Shnoulle: Here : we need to don't validate …  we need to disable validation …
We validate only `['completed', 'usesleft']` …

olleharstedt: >  we need to disable validation …

Yes, and better to be explicit about it. Yii 1 is an absolute mess in this regard already. save() validates but update() doesn't? Pfff.

Shnoulle: I mean `$token->save(true, ['completed', 'usesleft']);` : https://www.yiiframework.com/doc/api/1.1/CActiveRecord#save-detail i reamlly unclear about validation : true mean validate, but validate what ?

 `$token->save(false)` is clear : no validation. 

`$token->saveAttributes(['completed', 'usesleft']);` seems really clear (for me) : just save the attributes in the array: no validation is done :)

gabrieljenik: `$token->saveAttributes(['completed', 'usesleft'])` doesn't perform validation on those attributes.

`$token->save(true, ['completed', 'usesleft']);` only valdates and saves those attributes.
What's wrong about that?

PR #2086, opened 2021-10-06
Fixed issue #17527: Validate before using updateAll

gabrieljenik:


Shnoulle (approved):
Must be merged (XSS filter on em_validation_q_tip)

I report the feature about textarea not filtered.

tests/functional/backend/SaveQuestionAttributesTest.php
```diff
@@ -0,0 +1,199 @@
...
+
+        // Upload the file
+        $web->wait(10)->until(WebDriverExpectedCondition::elementToBeClickable(WebDriverBy::cssSelector('#installPluginZipModal #the_file')));
+        $fileInput = $web->findByCss('#installPluginZipModal #the_file');
+        $fileInput->setFileDetector(new LocalFileDetector());
+        $file = BASEPATH . '../tests/data/file_upload/NewQuestionAttributesPlugin.zip';
+        $this->assertTrue(file_exists($file));
+        $fileInput->sendKeys($file)->submit();
+
+        $button = $this->waitForElementShim($web, '[type="submit"][value="Install"]');
```
gabrieljenik, line 181: @olleharstedt We have problem when executing the test on this line. The button never appears. It seems something is wrong with the previous block of lines, the installation process.

We can't reproduce on our env.
Do you have a magical way of seeing what's happening on the CI?

If not, alternatives:
a- install it through PHP code, not the UI.
b- add it as a demo plugin which can be installable.

ideas?

application/models/QuestionAttribute.php
```diff
@@ -104,7 +106,10 @@ public function setQuestionAttributeWithLanguage($iQuestionID, $sAttributeName,
         $oModel = new self();
         $aResult = $oModel->findAll('attribute=:attributeName and qid=:questionID and language=:language', array(':attributeName' => $sAttributeName, ':language' => $sLanguage, ':questionID' => $iQuestionID));
         if (!empty($aResult)) {
-            $oModel->updateAll(array('value' => $sValue), 'attribute=:attributeName and qid=:questionID and language=:language', array(':attributeName' => $sAttributeName, ':language' => $sLanguage, ':questionID' => $iQuestionID));
+            foreach ($aResult as $questionAttribute) {
+                $questionAttribute->value = $sValue;
+                $questionAttribute->save();
```
Shnoulle, line 111: :+1:

application/models/QuestionAttribute.php
```diff
@@ -578,4 +589,39 @@ public static function getOwnQuestionAttributesViaPlugin()
...
+            return;
+        }
+        $questionAttributeFetcher = new \LimeSurvey\Models\Services\QuestionAttributeFetcher();
+        $questionAttributeFetcher->setQuestion($question);
+        $questionAttributeDefinitions = $questionAttributeFetcher->fetch();
+
+        // The value will be filtered unless the attribute definition has the "xssfilter" property set to false
+        $shouldFilter = true;
+        if (isset($questionAttributeDefinitions[$this->attribute])) {
+            $questionAttributeDefinition = $questionAttributeDefinitions[$this->attribute];
```
Shnoulle, line 612: No way to remove the `array_key_exists` check here ? Adding xssFilter => true on https://github.com/LimeSurvey/LimeSurvey/blob/b43948123d3cb6e1a029779df0b346a704af3a5a/application/models/QuestionAttribute.php#L355 didn't work ?

(If not it's really another dev issue)

application/models/QuestionAttribute.php
```diff
@@ -578,4 +589,39 @@ public static function getOwnQuestionAttributesViaPlugin()
...
+            if (array_key_exists("xssfilter", $questionAttributeDefinition) && $questionAttributeDefinition['xssfilter'] == false) {
+                $shouldFilter = false;
+            }
+        }
+
+        if (!$shouldFilter) {
```
Shnoulle, line 618: And then directly `!$questionAttributeDefinition['xssfilter']` or `$questionAttributeDefinition['xssfilter'] === false` : less code, less test, less variable creation : good for speed and comprehension (with getDefaultSettings)

Shnoulle (commented):
Wait for final code, but before can be tested with a fixed https://github.com/LimeSurvey/LimeSurvey/blob/b43948123d3cb6e1a029779df0b346a704af3a5a/application/models/QuestionAttribute.php#L355 ?

tests/functional/backend/SaveQuestionAttributesTest.php
olleharstedt: >  Do you have a magical way of seeing what's happening on the CI?

No, that's the sad part of CI. It wasn't possible for Travis, but maybe you can google around for Github Actions?

olleharstedt: Is it a file/folder permission problem, perhaps?

application/models/QuestionAttribute.php
gabrieljenik: OK, we will add the xssFilter => TRUE on `getDefaultSettings() `.
Still, I think we should still check, just in case, to avoid problems with old plugins, question themes ... or not expected situations...

tests/functional/backend/SaveQuestionAttributesTest.php
gabrieljenik: > Is it a file/folder permission problem, perhaps?

I think so. We will do a dump and pick up from there.

application/models/QuestionAttribute.php
Shnoulle: >  to avoid problems with old plugins, question themes

Reason to have default settings, it' added for this : be sure to have it set.

> or not expected situations...

a not expected situation means a potential issue : i prefer to have an error at debug > 0 in this case.

gabrieljenik: The problem is that with debug = 0 the error will not show and there is the possibility that the attribute doesn't get  filtered when supposed to. We added the xssFilter on defaultSettings, but prefer to leave the `array_key_exists`, if it is ok. Thanks!

tests/functional/backend/SaveQuestionAttributesTest.php
```diff
@@ -0,0 +1,216 @@
...
+    public static function setupBeforeClass(): void
+    {
+        parent::setupBeforeClass();
+
+        // Permission to everything.
+        \Yii::app()->session['loginID'] = 1;
+
+        $oUser = self::createUserWithPermissions(
+            [
+                "users_name" => "surveyadmin",
```
Shnoulle, line 29: What happen if user already exist (do this test only) : maybe try with random sring ?

tests/functional/backend/SaveQuestionAttributesTest.php
```diff
@@ -0,0 +1,216 @@
...
+use Facebook\WebDriver\WebDriverExpectedCondition;
+use Facebook\WebDriver\Remote\LocalFileDetector;
+
+/**
+ * @since 2021-11-09
+ * @group questionattribute
+ */
+class SaveQuestionAttributesTest extends TestBaseClassWeb
+{
+    public static $newUserId = null;
```
Shnoulle, line 15: Add a publi static username ?

Shnoulle (commented):
Just in case : when testing single Test if user exist .
Create a 42 char of random string with uniqID ?

tests/functional/backend/SaveQuestionAttributesTest.php
gabrieljenik: This is tested on a balnk environment.

tests/functional/backend/SaveQuestionAttributesTest.php
gabrieljenik: why?

Shnoulle (changes requested):
Can not merge before updating existing attribute where xssFilter  break

tests/functional/backend/SaveQuestionAttributesTest.php
Shnoulle: > This is tested on a blank environment.

By our CI, but each test can be tested on existing environment (i use it a lot when construct test).

I think it's better, but not mandatory :).

Shnoulle: @olleharstedt did we construct the test globally or only for our CI ?

olleharstedt: Hm, globally what? The test can make assumption on the data, like "populated database" or "empty database". I'm currently working on a script to generate random test data.

olleharstedt: We don't really have a good system for database fixtures right now. So all tests more or less work on incidental data created by previous tests. Not so good, really.

Shnoulle (approved):

Shnoulle:
Fixed code OK , wit for other needed attrubutes ;)

PR #2090, opened 2021-10-08
Fixing wrong sort order in statisstics output (# 17647)

steffen-moser:
Fixed issue #17647: Strange sort order of question groups in statistics


c-schmitz:
Thank you!

PR #2099, opened 2021-10-15
Fix saving location components in map.js

f107:
Fixed issue #17663:

Fix incorrect behavior of the localization component settings.


gabrieljenik (commented):
There are couple of trailing spaces at the end of the lines. 
Not sure if that's something to be removed or not.
@olleharstedt Can those be kept or those should be removed?

olleharstedt:
> @olleharstedt Can those be kept or those should be removed?

Baaah, these lines do not even have curly braces, so no standard is applied to this file currently. What we want is probably AirBNB eslint on all JS files in this folder.

gabrieljenik (approved):

PR #2105, opened 2021-10-19
New feature [PluginEvent]: Added a beforeUserGroupPermissionSetSave event to surveypermissions controller

danguyf:
Dev: Added a new PluginEvent that my customer needs.

### DIFF
--- a/application/controllers/admin/surveypermission.php
+++ b/application/controllers/admin/surveypermission.php
@@ -660,7 +660,7 @@ class surveypermission extends Survey_Common_Action
             }
 
             if (isset($postusergroupid) && $postusergroupid > 0) {
-                $event = new PluginEvent('beforeUserGroupPermissionSetSave');
+                $event = new PluginEvent('afterUserGroupPermissionSetSave');
                 $event->set('survey', $surveyid);
                 $event->set('ugid', $postusergroupid);
                 $event->set('permissions', $aPermissions);


olleharstedt:
Same here, no new features in LS3. Feel free to open a PR on the development branch instead.

danguyf:
Last time I submitted a new PluginEvent for LS3 you closed the PR saying there are no new features for LS3.

Then months later in the community team meeting someone else needed a new PluginEvent in LS3 and it was accepted.  I pointed out that mine had been denied because supposedly there are no new features for LS3.  The response was that simple PluginEvents were allowable and that I should resubmit my PR.

I finally got around to resubmitting it and you've closed it all over again despite telling me that it would be allowed.

Shnoulle:
Not related here (i'm not the one who don't want new feature in LTS) : why not use existing event in your plugin ?

BeforeUserInGroupSave and AfterUserInGroupSave

danguyf:
That is an excellent suggestion and betrays my lack of understanding of/comfort with Model Events.  I need to hook into when the permissions are set, not when users are added to a group, so perhaps I would use BeforePermissionSave and AfterPermissionSave instead?

Shnoulle:
You want to give permission to user in same group ?

https://github.com/LimeSurvey/LimeSurvey/blob/fbfebff6708dc21626c86d2bc7d894b91ab5731a/application/models/Permission.php#L312 seems a great idea.

maziminke:
I second the approach to use existing plugin event whenever possible.

...but I am also all for having new plugin events at LS3.x. Such events can't really break any existing code and I remember that for our projects it was also possible to add new plugin events for LS 3.x if everything was well coded and codumented.

PR #2107, opened 2021-10-19
New feature [PluginEvent]: Added a beforeResumeEmail plugin event to Save library

danguyf:
Dev: Added a beforeResumeEmail plugin event to Save library.

I should have included this with PR 2106. (This is my last PluginEvent PR related to email, honest!)

The same LS3 customer that I have who needs to be able to selectively alter the from and bounce of outgoing emails needs to intercept the emails sent when a survey is saved and the user is emailed how to resume it.

### DIFF
--- a/application/libraries/Save.php
+++ b/application/libraries/Save.php
@@ -287,13 +287,29 @@ class Save
                 $message .= Yii::app()->getController()->createAbsoluteUrl("/survey/index/sid/{$surveyid}/loadall/reload", $aParams);
                 $from     = "{$thissurvey['adminname']} <{$thissurvey['adminemail']}>";
 
-                if (SendEmailMessage($message, $subject, $_POST['saveemail'], $from, $sitename, false, getBounceEmail($surveyid))) {
-                    $emailsent = "Y";
-                } else {
-
-                    $errormsg .= gT('Error: Email failed, this may indicate a PHP Mail Setup problem on the server. Your survey details have still been saved, however you will not get an email with the details. You should note the "name" and "password" you just used for future reference.');
-                    if (trim($thissurvey['adminemail']) == '') {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Yes, but if you need something in 3.X : i think best is to create a compatible event.

Else : currently i think i start forking 3.X, in december 5.X can be the official stable release and 3.X is the old stable. Since i need some event (for example : i need Expression event since month now …) : i add i can not move to 5./X in a 3.X fork …

olleharstedt:
Hm, can you move this to dev branch, please? Normally we don't accept new features in an LTS version.

Shnoulle:
> **Hm, can you move this to dev branch, please? Normally we don't accept new features in an LTS version.**

> Hm, can you move this to dev branch, please? Normally we don't accept new features in an LTS version.

Then : can close : master already have this event

olleharstedt:
Thanks, Denis.

danguyf:
No, the ENTIRE point of this PR is back-porting the event from LS5 to LS3 so that LS3 customers can make use of it and so that plugins can target LS3 and LS5 without complicated version-switching logic.

That this is submitted to 3.x-LTS is ENTIRELY INTENTIONAL.

Please reconsider.

danguyf:
This adds a PluginEvent to LS3 that is found in LS4+, enabling plugins to target LS3+.  That seems worth adding to LS3, no?

maziminke:
+1 for adding this to LS 3.x.

DenisChenu:
My answer is for «Hm, can you move this to dev branch, please? Normally we don't accept new features in an LTS version.» only :).

Else my other comments are to be more plugin compatible from 3.X to 5.X (but register to unexisting plugin don't broke anything)

PR #2114, opened 2021-10-26
Fixed issue #17525: VVImport: "Exclude record IDs" ignores existing e…

Shnoulle:
Fixed issue #17525: VVImport: "Exclude record IDs" ignores existing entries while it should force "renumber"
Dev: unset any existing id in case of ignore
Dev: default to renumber


olleharstedt:
Exactly the same? Can we cherry-pick?

Shnoulle:
? You want to cherry-pick master to 3.X ?

It's issue : then i do both (always)

gabrieljenik (approved):
Same as in master

PR #2119, opened 2021-10-26
Fixed issue #14955: Question relevance is not reassessed after group …

Shnoulle:
Fixed issue #14955: Question relevance is not reassessed after group is shown
Dev: test of condition of question must be checked if relevance of his group was updated.


gabrieljenik (approved):
All good. Nothing caught my attention.

PR #2123, opened 2021-10-27
Fixed issue #T1358: Label Sets: Export multiple label sets - Export Button disabled state

thedirtypanda:
Export button. Should have a disabled and enabled state. By doing so we dont have to show the error message on export.


application/views/admin/labels/exportmulti_view.php
```diff
@@ -33,3 +33,22 @@
...
 </div>
 
+<script type='text/javascript'>
+    /** If one or more elements are selected, then the export button will be enabled.
+     *  otherwise its disabled by default. */
+    let isSelected    = false;
+    let exportButton  = document.getElementById('exportButton');
+
+    $('select').change(function() {
+        $('#labelsets option:selected').each(function() {
```
Shnoulle, line 43: `isSelected = $('#labelsets option:selected').length` > 0 ? think it's more quick :)

application/views/admin/labels/exportmulti_view.php
```diff
@@ -33,3 +33,22 @@
     </div>
 </div>
 
+<script type='text/javascript'>
+    /** If one or more elements are selected, then the export button will be enabled.
+     *  otherwise its disabled by default. */
+    let isSelected    = false;
+    let exportButton  = document.getElementById('exportButton');
```
Shnoulle, line 40: why not `$("#exportButton")` ? Didn't broke JS if export button is not there.

Shnoulle (commented):
Still same questions on javascript part

PR #2125, opened 2021-10-27
Dev #9629: Global opt out link for CPDB missing

gabrieljenik:


Shnoulle:
emailstatus are not directly checked in https://github.com/LimeSurvey/LimeSurvey/blob/17be6865489fd9de9b33497d4e2a5a01e369759a/application/models/TokenDynamic.php#L167

Only as parameter

Maybe need a parameters for blacklist ?

And why not disable send (by default) any email to blacklisted email ?
Directly in https://github.com/LimeSurvey/LimeSurvey/blob/895b5ac62ac0efe955d5a5c825364dda3c4dd0ce/application/core/LimeMailer.php#L558 

Same : maybe with option `LimeMailer->$sendToGlobalBlacklist = false;`

gabrieljenik:
Thanks for the comments. Yes, solution to the whole feature could be partial. 
We have spoted some blanks but decided to start with this.

> emailstatus are not directly checked in

Sorry, got lost here. What do you mean?
Isn't it checked here? https://github.com/LimeSurvey/LimeSurvey/blob/17be6865489fd9de9b33497d4e2a5a01e369759a/application/models/TokenDynamic.php#L186

> Maybe need a parameters for blacklist ?

How that parameter would work? What are you picturing for it?
Still talking about the global opt out?

Thanks!

Shnoulle:
Hi,

I mean : `findUninvited($aTokenIds = false, $iMaxEmails = 0, $bEmail = true, $SQLemailstatuscondition = '', $SQLremindercountcondition = '', $SQLreminderdelaycondition = '')`

In this function there are no condition on emailstatus by default.

I check of usage : 
```
application/helpers/remotecontrol/remotecontrol_handle.php:            $aResultTokens = $oTokens->findUninvited($aTokenIds, $iMaxEmails, $bEmail, $SQLemailstatuscondition);
application/helpers/remotecontrol/remotecontrol_handle.php:            $aAllTokens = $oTokens->findUninvitedIDs(false, 0, true, $SQLemailstatuscondition);
application/helpers/remotecontrol/remotecontrol_handle.php:            $aAllTokens = $oTokens->findUninvitedIDs(false, 0, false, $SQLemailstatuscondition, $SQLremindercountcondition, $SQLreminderdelaycondition);
application/helpers/remotecontrol/remotecontrol_handle.php:            $aResultTokens = $oTokens->findUninvited($aTokenIds, $iMaxEmails, false, $SQLemailstatuscondition, $SQLremindercountcondition, $SQLreminderdelaycondition);
application/controllers/admin/tokens.php:            $ctresult = TokenDynamic::model($iSurveyId)->findUninvitedIDs($aTokenIds, 0, $bIsInvitation, $SQLemailstatuscondition, $SQLremindercountcondition, $SQLreminderdelaycondition);
application/controllers/admin/tokens.php:            $emresult = TokenDynamic::model($iSurveyId)->findUninvited($aTokenIds, $iMaxEmails, $bIsInvitation, $SQLemailstatuscondition, $SQLremindercountcondition, $SQLreminderdelaycondition);
application/controllers/admin/tokens.php:            $aTokens = TokenDynamic::model($iSurveyId)->findUninvitedIDs($aTokenIds, 0, $bEmail, $SQLemailstatuscondition);
```

Seems this function is always used with an  `SQLemailstatuscondition`
Then add a parameters 

`findUninvited($aTokenIds = false, $iMaxEmails = 0, $bEmail = true, $SQLemailstatuscondition = '', $SQLremindercountcondition = '', $SQLreminderdelaycondition = '', $globalOptOuted = false )`


**And here : it's for invitation, not for reminder or register or confirmation.**

Global must be done for all token email (at minimum) and potentially for admin email too … 
My opinion : best is to 
1. Add control system in mailer, and don't send (by default) if it's a token email (or ?)
2. Add filter in token admin management when send to all
3. Add control in register (and return "This email ask to never receive email …")
4. Think of what to do for admin email …

Shnoulle:
This code didn't have issue.

The issue is this functionality is totally broken ! Don't add link by default to a broken functionality … 

**Global optout didn't work** 

1. _If you do not want to participate in any future survey and don't want to receive any more invitations please click the following link:_ this sentence is FALSE, the true sentence : If you are a know user, we don't use pour existing panel user for future survey, but maybe we use your email on any survey at any time .
2. When user click on optout/participants but not in participant DB : hje don't see any error information.

application/controllers/admin/ExpressionValidate.php
```diff
@@ -195,6 +195,7 @@ public function email($iSurveyId, $lang)
                 $aReplacement["LASTNAME"] = gT("Participant - Last name");
                 $aReplacement["LANGUAGE"] = gT("Participant - Language");
                 $aReplacement["OPTOUTURL"] = gT("URL for a respondent to opt-out of this survey");
+                $aReplacement["GLOBALOPTOUTURL"] = gT("URL for a respondent to opt-out of surveys in this server");
```
Shnoulle, line 198: This sentence is false, it's «You have been removed from the central participants list for this site » the result 

not more : https://github.com/LimeSurvey/LimeSurvey/blob/ef2b9e003080274edcc888f8f012ce94b2fa74ce/application/controllers/OptoutController.php#L183

Shnoulle (changes requested):
Remove it from default.

gabrieljenik:
> The issue is this functionality is totally broken !

Agree. We need some specifications and directions.

> Don't add link by default to a broken functionality …

We are rebuilding it slowly.
I will review the messages again to make sure they make sense.

> If you do not want to participate in any future survey and don't want to receive any more invitations please click the following link: this sentence is FALSE, the true sentence : If you are a know user, we don't use pour existing panel user for future survey, but maybe we use your email on any survey at any time 

That messages is what is supposed to be. You are saying that doesn't hold right now?
Maybe because of still missing features?

Sorry got a bit lost. Maybe we could discuss a bit on discord.

Shnoulle:
See mantis about the final reason.
https://bugs.limesurvey.org/view.php?id=9629#c67586

Your totally right : this fix m=ust stau only for Global opt out link , then 

1. Adding a «Email Placeholders»
2. Adding it in editor (but clearly set "for participant from CPDB only»).

But not in default email : i think CPDB is not used by a lot of our user.

gabrieljenik:
> Adding a «Email Placeholders»

Agree and done. It is {GLOBALOPTOUT}.
We need to add it to the manual yet.
https://manual.limesurvey.org/Email_templates#Email_Placeholders

> Adding it in editor (but clearly set "for participant from CPDB only»).

Added to the editor.
Need to review the description of the replacement variable.

> But not in default email : i think CPDB is not used by a lot of our user.

Agree.
I suggest to also add a new request feature for creating blacklists, besides a survey manager uses or not the CPDB.
That new feature can be included in the default emails (and I think it should as to not be considered in spam)

I think this may be a good way to wrap this up, at least temporarily.

What do you think?

application/controllers/admin/ExpressionValidate.php
gabrieljenik: ok, I think we understood where you are going. Will review this description.

Shnoulle:
> I suggest to also add a new request feature for creating blacklists, besides a survey manager uses or not the CPDB.
> That new feature can be included in the default emails (and I think it should as to not be considered in spam)

Core plugin can be great here :)

application/models/TokenDynamic.php
```diff
@@ -223,44 +229,9 @@ public function findUninvited($aTokenIds = false, $iMaxEmails = 0, $bEmail = tru
...
-
-        $command->order = 'tid';
-
-        $oResult = $this->getCommandBuilder()
-            ->createFindCommand($this->getTableSchema(), $command)
-            ->select('tid')
-            ->queryColumn();
-        return $oResult;
+        $tokens = $this->findUninvited($aTokenIds, $iMaxEmails, $bEmail, $SQLemailstatuscondition, $SQLremindercountcondition, $SQLreminderdelaycondition);
+        $ids = array_map(function ($item) { return $item->tid; }, $tokens);
```
Shnoulle, line 233: PSR12 / 5. Control Structures :)

Shnoulle (changes requested):
Just some line feed on anonymous function please :)

Shnoulle (approved):
:+1:

olleharstedt:
Shouldn't this be in dev branch, tho...?

Shnoulle:
More a bug fix here … manual show solution … that don't exist in real.

> we tried to limit it as much as possible

:+1:

If move to dev version : please with a complete email adress block list, not a fix, a real solution.

glimz:
I will merge this for now. I will create a new task to extend this task and implement what is mentioned in the manual

PR #2128, opened 2021-10-31
Fixed issue: Show full name of a deleted user correctly

ViliusS:
Dev: If user's full name contain double quotes it is not displayed correctly when this user is deleted. Instead of passing full name back and forth in the session let's just use the data from the database.

Other user actions probably have the same bug, this patch is only for delete action.

### DIFF
--- a/application/controllers/admin/useraction.php
+++ b/application/controllers/admin/useraction.php
@@ -192,7 +192,6 @@ class UserAction extends Survey_Common_Action
             $oInitialAdmin = User::model()->findByAttributes(array('parent_id' => 0));
 
             $postuserid = $this->_getPostOrParam("uid");
-            $postuser = flattenText($this->_getPostOrParam("user"));
 
             if ($oInitialAdmin && $oInitialAdmin->uid == $postuserid) {
 // it's the original superadmin !!!
@@ -239,7 +238,6 @@ class UserAction extends Survey_Common_Action
                     $this->deleteFinalUser($ownerUser, $transfer_surveys_to);
                 } else {
                     $aData['postuserid'] = $postuserid;
-                    $aData['postuser'] = $postuser;
                     $aData['current_user'] = $current_user;
 
                     $aViewUrls['deluser'][] = $aData;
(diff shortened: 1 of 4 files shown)


application/controllers/admin/useraction.php
```diff
@@ -305,7 +302,7 @@ public function deleteFinalUser($result, $transfer_surveys_to)
             die();
         }
 
-        $extra = "<br />".sprintf(gT("User '%s' was successfully deleted."), $postuser)."<br /><br />\n";
+        $extra = "<br />".sprintf(gT("User '%s' was successfully deleted."), flattenText($fields['full_name']))."<br /><br />\n";
```
Shnoulle, line 305: Why not Chtml::encode ?

ViliusS: I tried to do the changes as minimal invasive as possible, i.e. I kept flattenText() from the removed lines above. If you think Chtml::encode() is better in this case, I can change it.

To tell the truth I'm not sure how LimeSurvey uses flattenText() and that is the difference because in one place I found it used, other similar places uses hardcoded htmlspecialchars() and others use Chtml::encode().

Shnoulle: `Chtml::encode` == `htmlspecialchars` ;)

https://www.yiiframework.com/doc/api/1.1/CHtml#encode-detail

I totally understand your pov about « changes as minimally» but clearly here : a Chtml::encode seems better.

ViliusS: Done.

Shnoulle (approved):

ViliusS:
I went ahead and removed 'user' parameter from other actions too. Not sure what was the intention but looking from the code it appeared with https://github.com/LimeSurvey/LimeSurvey/commit/4bfbbba7613306f470c97b578ff86023d98c4226 and never materialized. If you think this last change is too much for LTS branch, please merge this pull request without this last commit.

Shnoulle (approved):

glimz:
Hello ViliusS,

Thank you for creating a PR. However, I cannot find any bug report in mantis for this issue.

For better tracking, please create a bug report so we can test and merge your PR.

Thank you

ViliusS:

glimz:
> Sorry I don't understand this requirement, nor do I have a Mantis account.

Please see the note on how to [contribute to the project.](https://github.com/LimeSurvey/LimeSurvey/blob/master/CONTRIBUTING.md)

c-schmitz:
Mantis is our bug tracker at https://bugs.limesurvey.org

PR #2129, opened 2021-11-01
Fixed issue #T136: Error Message Needs Rewording

gabrieljenik:


application/controllers/SurveyAdministrationController.php
```diff
@@ -2350,10 +2361,16 @@ public function actionExpireMultipleSurveys()
...
-            } else {
+            if (!Permission::model()->hasGlobalPermission('superadmin')) {
                 $aResults[$survey->primaryKey]['result'] = false;
+                $aResults[$survey->primaryKey]['error'] = gT("User does not have valid permissions");
+            } else {
+                if ($survey->save()) {
+                    $aResults[$survey->primaryKey]['result'] = true;
+                } else {
+                    $aResults[$survey->primaryKey]['result'] = false;
+                    $aResults[$survey->primaryKey]['error'] = gT("Survey update failed");
```
Shnoulle, line 2372: Maybe $survey->getErrors() ?

application/controllers/SurveyAdministrationController.php
```diff
@@ -2618,6 +2635,7 @@ public function changeTemplate($iSurveyID, $template, $aResults = null, $bReturn
             if (!empty($bReturn)) {
                 $aResults[$iSurveyID]['title'] = $survey->correct_relation_defaultlanguage->surveyls_title;
                 $aResults[$iSurveyID]['result'] = false;
+                $aResults[$iSurveyID]['error'] = gT("User does not have valid permissions");
```
Shnoulle, line 2638: No : permission is already checked at line https://github.com/LimeSurvey/LimeSurvey/blob/743d8b74d16892dfee04fa32194c902813d8f586/application/controllers/SurveyAdministrationController.php#L2634

`errorSummary($osSurvey)` maybe ?

(and why we check `surveyactivation` here ?)

Shnoulle (commented):
Usage of errorSummary directly ? Unsure it can be used in returned party …

Shnoulle (changes requested):
See : https://github.com/LimeSurvey/LimeSurvey/pull/2129/files#r740345573 

Then "Unkonw error" or errorSummary

application/controllers/SurveyAdministrationController.php
gabrieljenik: I don't think those errors are descriptive, but too technical. If there would be a log or an "advanced results" section, would be good idea. But here, not sure if it is a good fit. We are not doing that anywhere else I think.

gabrieljenik: `errorSummary($osSurvey)`, not sure how this could be used when the output is for multiple surveys. Maybe too much information for the grid. I would say to leave it for another ticket?

Shnoulle: I put a **?**

OK for stay (must check what error can be done ;).

Maybe log it then ?

application/controllers/SurveyAdministrationController.php
gabrieljenik: > (and why we check `surveyactivation` here ?)

This was like this before this commit. Probably they are just reusing a permission :)

Shnoulle: > No : permission is already checked at line 

This one is dirfferent : _unknow reason_ ?

gabrieljenik: No, when you reach that point (L2638), you are lacking permissions.

Shnoulle: Oh yes … in the { }

Sorry

Shnoulle (approved):

PR #2131, opened 2021-11-03
Fixed issue #17632: Basic and detailed admin notification email fails

gabrieljenik:
The problem is somehow introduced while fixing #17512.

#17512 error was about string processing errors not being shown as they errors were reset after evaluating each subexpression. 
So, when a string had 2 expressions, if the latter expression had no errors, the Survey Logic File was not showing the errors of the first expression. Fix: Stop resetting errors.

The problem we see now is that every time a subexpression is processed, the first thing done is checking for errors. If any error is found, no subexpression processing is done. 

As after the previous fix, the errors are not being reset, EM considers that after a subexpression has errors, all subexpressions that come after it have errors too.

Here, we need a deeper review on the handling of errors for subexpressions, as we are missing errors at expression level and subexpression level.

As for now, the fix we belive is ok is to:
- somehow revert prior fix
- and re apply it by narrowing down the fix to only the survey logic file process.

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -75,6 +75,7 @@ class ExpressionManager
     private $RDP_result; // final result of evaluating the expression;
     private $RDP_evalStatus; // true if $RDP_result is a valid result, and  there are no serious errors
     private $varsUsed; // list of variables referenced in the equation
+    public $resetErrorsAndWarningsOnEachPart = true;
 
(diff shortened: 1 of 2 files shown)


Shnoulle:
Great : we need to keep it only for admin part !

Shnoulle:
No ,

I mean : we really don't need clean error count except on some specific admin part (logic files and _maybe_ some other part).

Code OK ;)

PR #2133, opened 2021-11-04
Fixed issue #14091: Filenames of uploads starting with special charac…

Shnoulle:
Fixed issue #14091: Filenames of uploads starting with special characters truncated with invalid setlocale
Dev: basename according to OK, avoid usage of basename due to setlocale

### DIFF
--- a/application/controllers/UploaderController.php
+++ b/application/controllers/UploaderController.php
@@ -141,8 +141,9 @@ class UploaderController extends SurveyController
                 mkdir($sTempUploadDir);
             }
             $filename = $_FILES['uploadfile']['name'];
-            // Do we filter file name ? It's used on displaying only , but not save like that.
-            //$filename = sanitize_filename($_FILES['uploadfile']['name']);// This remove all non alpha numeric characters and replaced by _ . Leave only one dot .
+            /* Fix file name : need to get only the basename url encoded for display . Avoid usage of basename : need valid UTF8 setlocale */
+            $aFilename = explode(DIRECTORY_SEPARATOR, rtrim( $filename, DIRECTORY_SEPARATOR));
+            $fixedFilename = rawurlencode(end($aFilename));
             $size = $_FILES['uploadfile']['size'] / 1024;
             $preview = Yii::app()->session['preview'];
             $aFieldMap = createFieldMap($oSurvey, 'short', false, false, $sLanguage);
@@ -246,7 +247,7 @@ class UploaderController extends SurveyController
                                 "success"       => true,
                                 "file_index"    => $filecount,
                                 "size"          => $size,
-                                "name"          => rawurlencode(basename($filename)),
+                                "name"          => $fixedFilename,
                                 "ext"           => $cleanExt,
                                 "filename"      => $randfilename,
                                 "msg"           => gT("The file has been successfully uploaded.")
@@ -274,7 +275,7 @@ class UploaderController extends SurveyController
                     $return = array(
                         "success" => true,
                         "size"    => $size,
-                        "name"    => rawurlencode(basename($filename)),
+                        "name"    => $fixedFilename,
                         "ext"     => $cleanExt,
                         "filename"      => $randfilename,
                         "msg"     => gT("The file has been successfully uploaded.")


olleharstedt:
FILE: ...Survey/LimeSurvey/application/helpers/update/updatedb_helper.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 4170 | ERROR | [x] Whitespace found at end of line

Shnoulle:
> Survey/LimeSurvey/application/helpers/update/updatedb_helper.php

This file is not part of this pull request

olleharstedt:
Pull upstream?

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2133/commits/77dfd23c5b4547fbb85bcb26e00840cf73b3cc3e …

Shnoulle:
But i see a +2/-2 when i merge …

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/commit/495b9426383b44d5c6ebb8692a1a3ef94e917c9e i mean …

olleharstedt:
Sorry, our fault.

Shnoulle:
No problem :)

Shnoulle:
PS : must check security issue here.

application/controllers/UploaderController.php
```diff
@@ -141,8 +141,9 @@ public function run($actionID)
                 mkdir($sTempUploadDir);
             }
             $filename = $_FILES['uploadfile']['name'];
-            // Do we filter file name ? It's used on displaying only , but not save like that.
-            //$filename = sanitize_filename($_FILES['uploadfile']['name']);// This remove all non alpha numeric characters and replaced by _ . Leave only one dot .
+            /* Fix file name : need to get only the basename url encoded for display . Avoid usage of basename : need valid UTF8 setlocale */
+            $aFilename = explode(DIRECTORY_SEPARATOR, rtrim( $filename, DIRECTORY_SEPARATOR));
```
c-schmitz, line 145: I don't think that is safe.
See https://www.php.net/manual/en/features.file-upload.post-method.php
"$_FILES['userfile']['name'] : The original name of the file on the client machine."
Why would the server DIRECTORY_SEPARATOR matter here?

c-schmitz (changes requested):

application/controllers/UploaderController.php
Shnoulle: To replace basename($filename) only. 
I don't know if we need it and when it was added.

c-schmitz: Then it is a bad replacement. 
$_FILES['userfile']['name'] is not to be trusted at all, since it is user content. If you want to trim it of any dangerous content you will have to consider all kind of directory separators and also consider other characters (dots, quotes. etc)

Shnoulle: basename($filename) do only that : no other filter except last part.

I can use https://www.php.net/manual/fr/function.basename.php#121405 if you think it's more secure. Then if a linux user send bad\name.png : we save only name.png

Else: without valid setlocale : i don't have any solution :), except remove any filter on name (we don't really care on tghe server).

c-schmitz: I would use sanitize_filename(), not sure if it also handles file paths - please check.

Shnoulle: > I would use sanitize_filename(), not sure if it also handles file paths - please check.

[sanitize_filename](https://github.com/LimeSurvey/LimeSurvey/blob/f98ca8b9697e819449a4d7c79f224ca8ce4b2b19/application/helpers/sanitize_helper.php#L119) broke with same issue : österreich.png > sterreich.png 
with `setlocale(LC_CTYPE,"C");`
 maybe usage of pathinfo ?

Shnoulle: Oups … bad host for testing …

Shnoulle: No : confirmed : our current sanitize_filename broke with same issue : österreich.png > sterreich.png

Shnoulle: PS : even if fix the way of saving : https://github.com/LimeSurvey/LimeSurvey/blob/f98ca8b9697e819449a4d7c79f224ca8ce4b2b19/application/controllers/ResponsesController.php#L645

We sanitize here : maybe lets close issue as "Server configuration issue" ?

c-schmitz: Hmm.. can you try to figure out what function removes the umlaut in sanitize_filename ?

Shnoulle: I try ,

Else : i don't like current sanitizer : didn't accept `[``]` , but why ? There are no reason here ?

tgoeg: I'd very much advise against making this a configuration issue. LANG=C is the default on all Debian based Apache installations, so the majority of installations would have a configuration issue:
https://salsa.debian.org/apache-team/apache2/-/blob/master/debian/config-dir/envvars
That sounds unrealistic.

Shnoulle: Related issue : 17718: sanitize_filename didn't really fix filename for all system

Shnoulle: > Hmm.. can you try to figure out what function removes the umlaut in sanitize_filename ?

for sanitize_filename : `mb_strcut` usage remove the 1st utf8 character (with C locale), 
https://github.com/LimeSurvey/LimeSurvey/blob/b43948123d3cb6e1a029779df0b346a704af3a5a/application/helpers/sanitize_helper.php#L143

c-schmitz:
@Shnoulle can you check if the master patch is still needed when using #2225

Shnoulle:
> @Shnoulle can you check if the master patch is still needed when using #2225

No, no : clearly not and your fix seems really better :)

tgoeg:
Sorry if I can't seem to determine this myself: Will this be part of the 3.x tree as well (and which version)?
Then I can verify if it fixed the issue at my installation.

Shnoulle:
> Sorry if I can't seem to determine this myself: Will this be part of the 3.x tree as well (and which version)? Then I can verify if it fixed the issue at my installation.

No you're right issue 14091 is not fixed with https://github.com/LimeSurvey/LimeSurvey/pull/2225 but ?
This one is not the good way to fix …

@c-schmitz can you take https://bugs.limesurvey.org/view.php?id=14091 after merge in master is done ?

PR #2134, opened 2021-11-04
 Fixed issue #14091: Filenames of uploads starting with special charac…

Shnoulle:
Fixed issue #14091: Filenames of uploads starting with special characters truncated with invalid setlocale
Dev: basename according to OK, avoid usage of basename due to setlocale

### DIFF
--- a/application/controllers/UploaderController.php
+++ b/application/controllers/UploaderController.php
@@ -124,7 +124,7 @@ class UploaderController extends SurveyController
             $filename = $_FILES['uploadfile']['name'];
             // Do we filter file name ? It's used on displaying only , but not save like that.
             //$filename = sanitize_filename($_FILES['uploadfile']['name']);// This remove all non alpha numeric characters and replaced by _ . Leave only one dot .
-            /* Fix file name : need to get only the basename url encoded for display . Acvoid usage of basename , need valid UTF8 setlocale*/
+            /* Fix file name : need to get only the basename url encoded for display . Avoid usage of basename , need valid UTF8 setlocale*/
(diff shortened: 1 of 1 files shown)


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/commit/495b9426383b44d5c6ebb8692a1a3ef94e917c9e

c-schmitz:
PR for master: #2133

application/controllers/UploaderController.php
```diff
@@ -124,6 +124,9 @@ public function run($actionID)
             $filename = $_FILES['uploadfile']['name'];
             // Do we filter file name ? It's used on displaying only , but not save like that.
             //$filename = sanitize_filename($_FILES['uploadfile']['name']);// This remove all non alpha numeric characters and replaced by _ . Leave only one dot .
+            /* Fix file name : need to get only the basename url encoded for display . Avoid usage of basename , need valid UTF8 setlocale*/
+            $aFilename = explode(DIRECTORY_SEPARATOR, rtrim( $filename, DIRECTORY_SEPARATOR));
```
c-schmitz, line 128: This will not work. You can fake the file name to anything you want on upload, even have it contain XSS.
Instead maybe backport my fixed sanitize_filename function from LS5 and use it here.

c-schmitz (changes requested):

application/controllers/UploaderController.php
Shnoulle: You're right : i close this one.

Previous behaviour are ONLY usage of `basename`

PR #2136, opened 2021-11-05
Fixed issue #17674: Impossible to add new Menu or New Menu Entry 

gabrieljenik:
Recently, seems it has been changed how the token is set on AJAX calls.
Before, the ajaxSetup() methos was used.
Now, the beforeSend callback is used.

The problem of this bug is with the contentType, which is set to text/plain.
Why? Because at the time that is set, during ajaxSetup, the there was no payload, defaulting to text/plain.

Solution, is a bit wide.

Use $.ajaxPrefilter() instead of $.ajaxSetup({beforeSend: ...}) to add the CSRF token because beforeSend is
executed after the content type is determined. So, if the request had no data when beforeSend is executed,
the content type is 'text/plain', which is wrong.


Shnoulle (approved):
Code OK

PR #2141, opened 2021-11-11
Fixed issue #17717: Saving survey group permissions fails with error 500

vits:
Method getPermissionSet() expects camelcase entity name.

Fixed issue #17717

### DIFF
--- a/application/core/plugins/AuditLog/AuditLog.php
+++ b/application/core/plugins/AuditLog/AuditLog.php
@@ -197,7 +197,7 @@ class AuditLog extends \LimeSurvey\PluginManager\PluginBase
         $iSurveyID = $event->get('iSurveyID');
         $iUserID = $event->get('iUserID');
         $oCurrentUser = $this->api->getCurrentUser();
-        $oOldPermission = $this->api->getPermissionSet($iUserID, $iSurveyID, 'survey');
+        $oOldPermission = $this->api->getPermissionSet($iUserID, $iSurveyID, 'Survey');
         $sAction = 'update';   // Permissions are in general only updated (either you have a permission or you don't)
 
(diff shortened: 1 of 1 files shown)


Shnoulle:
@olleharstedt : it's about the `Method getPermissionSet() expects camelcase entity name.` : currently seems survey !== Survey (on myql8).
But : do you think we must check without camelcase by default ? Have survey == Survey when checking and search permission ?

Shnoulle:
@glimz code is OK ;)

olleharstedt:
>  But : do you think we must check without camelcase by default ? Have survey == Survey when checking and search permission ?

I guess this is a good use-case for `assert`. https://manual.limesurvey.org/Code_quality_guide#Assertions

Shnoulle:
3.X (mysql) https://github.com/LimeSurvey/LimeSurvey/blob/ada1d9ce496471dbb8c6380f3aa7737fb02f5932/application/controllers/InstallerController.php#L417 `DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci`

Not set in master, and not set with pgsql and msql

c-schmitz:
> 
> 
> 3.X (mysql)
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/ada1d9ce496471dbb8c6380f3aa7737fb02f5932/application/controllers/InstallerController.php#L417
> `DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci`
> 
> Not set in master, and not set with pgsql and msql

Sure, this is also set in master.

Shnoulle:
> Sure, this is also set in master.

Then must fix `$this->api->getPermissionSet($iUserID, $iSurveyID, 'survey')` !=` $this->api->getPermissionSet($iUserID, $iSurveyID, 'Survey')` ?
No ?

PR #2143, opened 2021-11-12
Fixed issue #17647: Strange sort order of question groups in statistics

Shnoulle:
Dev: order is already set in request for run
Dev: fix request for simple statistics
Dev: check where function is used to add it in comment

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -500,16 +500,13 @@ class Question extends LSActiveRecord
      */
     public function getQuestionList($surveyid)
     {
-        $db                  = Yii::app()->db;
-        $quotedGroup         = $db->quoteTableName('group');
-        $quotedGrouporder    = $db->quoteColumnName('group_order');
-        $quotedQuestionorder = $db->quoteColumnName('question_order');
         return Question::model()
             ->with('group')
             ->findAll(
                 array(
-                    'condition' => 't.sid=' . $surveyid,
-                    'order'     => $quotedGroup .  '.' . $quotedGrouporder . ' DESC, ' . $quotedQuestionorder
+                    'condition' => 't.sid=:sid',
+                    'order'     => 'group.group_order,question_order',
+                    'params'    => array(':sid' => $surveyid)
                 )
             );
     }
@@ -1023,6 +1020,14 @@ class Question extends LSActiveRecord
             ),
         ));
         return $dataProvider;
+    }
+
+    /** @inheritdoc */
+    public function scopes()
+    {
+        return array(
+            'primary' => array('condition' => "parent_qid = 0"),
+        );
     }
 
     /**
--- a/application/controllers/admin/statistics.php
+++ b/application/controllers/admin/statistics.php
@@ -167,13 +167,13 @@ class statistics extends Survey_Common_Action
          * b) "groups" -> group_name + group_order *
          */
 
-
-        $rows = Question::model()
-            ->with(array('group' => array('alias' => 'g')))
-            ->findAll(array('condition' => 'parent_qid = 0 AND g.sid=' . $surveyid, 'order' => 'group_order,question_order'));
-
-        //SORT IN NATURAL ORDER!
-        usort($rows, 'groupOrderThenQuestionOrder');
+         $rows = Question::model()
+             ->with('group')
+             ->findAll(array(
+                'condition' => 'parent_qid = 0 AND t.sid=:sid',
+                'order' => 'group_order,question_order',
+                'params' => array(":sid" => $surveyid)
+            ));
 
         //put the question information into the filter array
         $filters = array();
@@ -693,10 +693,13 @@ class statistics extends Survey_Common_Action
         $summary[4] = "idL";
 
         // 1: Get list of questions from survey
-        $rows = Question::model()->getQuestionList($surveyid);
-
-        //SORT IN NATURAL ORDER!
-        usort($rows, 'groupOrderThenQuestionOrder');
+        $rows = Question::model()
+            ->with('group')
+            ->findAll(array(
+                'condition' => 'parent_qid = 0 AND t.sid=:sid',
+                'order' => 'group_order,question_order',
+                'params' => array(":sid" => $surveyid)
+            ));
 
         // The questions to display (all question)
         foreach ($rows as $row) {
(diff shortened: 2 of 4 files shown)


gabrieljenik:
Sorry, I don't understand something.
We use `groupOrderThenQuestionOrder` for a given set of screens, but we use SQL ordering for other?

If that makes sense, wouldn't be better to have an extra function for the extra sort criteria?

Shnoulle:
`$rows = Question::model()->getQuestionList($surveyid);` return ALL question wityh or without parent_id 
With bad ordering (group DESC and parent_id > 0)

Then `groupOrderThenQuestionOrder` is needed to reorder by group

In other part : the group and the questions attributes was merged in same array : https://github.com/LimeSurvey/LimeSurvey/blob/b43948123d3cb6e1a029779df0b346a704af3a5a/application/controllers/admin/conditionsaction.php#L1320 

Then  `groupOrderThenQuestionOrder` work.

Here : no merging.

I check code and do a test : https://github.com/LimeSurvey/LimeSurvey/blob/5b7edbe43bd0b3bcf2a7c31047e96133fbc1eb62/application/controllers/admin/statistics.php#L710 load parent_qid.

My opinion : must deprecate broken function `Question::model()->getQuestionList` and `groupOrderThenQuestionOrder` usage BUT it's not needed to fix issue 17647 and it work (partially, quick test) on other usage.

gabrieljenik:
> My opinion : must deprecate broken function Question::model()->getQuestionList and groupOrderThenQuestionOrder usage BUT it's not needed to fix issue 17647 and it work (partially, quick test) on other usage.

I actually like the sorting function `groupOrderThenQuestionOrder`. :) If custom sortingis needed, seems to be a good solution, that's why I mention about creating  a new sorting function for this sorting criteria is needed.

I don't like to have that kind-of-query there in the code, but prefer to have it wrapped in a funciton for further reusing and creating a concept.

Another alternative would be to:
- fix sorting on Question::getQuestionList() which is only used on the stats and on the remote_control. (or maybe add a scope for that as well)
- add a scope (ex: primary) to the question model so that it only fetches the ones with parent_qid = 0.

What do you think?

Shnoulle:
> I don't like to have that kind-of-query there in the code, but prefer to have it wrapped in a funciton for further reusing and creating a concept.

I don't understand : we already use activerecords : it's already a wrapper … you want a function for this request : https://github.com/LimeSurvey/LimeSurvey/blob/5b7edbe43bd0b3bcf2a7c31047e96133fbc1eb62/application/controllers/admin/statistics.php#L170 

> fix sorting on Question::getQuestionList() which is only used on the stats and on the remote_control. (or maybe add a scope for that as well)
> add a scope (ex: primary) to the question model so that it only fetches the ones with parent_qid = 0.

Then : it's a rewrite part ? Not a bug fixing :). Need confirmation since it was used at a lot of place and other function …

1. Review this awfull  `Question::getQuestionList() `
2. Create a function to get the parent questionList (with or without langauge ?) with the good order
3. Create a function to get whole question with subquestion  (?)
4. etc …

It's clearly a rewrite not a bug fix.

gabrieljenik:
> It's clearly a rewrite not a bug fix.

well, don't agree that much, but we don't have to agree all time :)

You are adding code to those functions/scripts. I am just saying we could wrap it in functions/methods to unify criteria and to make it easier to reuse and mantain.

Shnoulle:
> > It's clearly a rewrite not a bug fix.
> 
> well, don't agree that much, but we don't have to agree all time :)
> 
> You are adding code to those functions/scripts. I am just saying we could wrap it in functions/methods to unify criteria and to make it easier to reuse and mantain.

Then this lines ?

https://github.com/LimeSurvey/LimeSurvey/blob/5b7edbe43bd0b3bcf2a7c31047e96133fbc1eb62/application/controllers/admin/statistics.php#L170-L176

getQuestionsByOrder($sid) ?

gabrieljenik:
> getQuestionsByOrder($sid) ?

I was thinking about creating a new scope on the model, instead of creating a new function

Shnoulle:
> > getQuestionsByOrder($sid) ?
> 
> I was thinking about creating a new scope on the model, instead of creating a new function

Used only for stat currently ?

gabrieljenik:
Later could be applied to other stuff. It is the start of something

Shnoulle:
Really good idea !

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/blob/5423939de0c28666b666145a2255cca9a5bb08a2/application/models/Survey.php#L1869 :/

gabrieljenik:
Nice

olleharstedt:
No strong opinion from me, I can go with either solution.

gabrieljenik:
> https://github.com/LimeSurvey/LimeSurvey/blob/5423939de0c28666b666145a2255cca9a5bb08a2/application/models/Survey.php#L1869
> 
> :/

Sorry, seeing this again.
I wouldn't use that function. Not widely used and not sure is what we need.
That sorting criteria is different depending on the db engine? weird.

Shnoulle:
> > https://github.com/LimeSurvey/LimeSurvey/blob/5423939de0c28666b666145a2255cca9a5bb08a2/application/models/Survey.php#L1869
> > 
> > :/
> 
> Sorry, seeing this again. I wouldn't use that function. Not widely used and not sure is what we need. That sorting criteria is different depending on the db engine? weird.

Oh i don't think to use it :). I just look at survey if we have a getPrimaryQuestionByGoodOrder nut only find this functioin :)

application/models/Question.php
```diff
@@ -1025,6 +1022,14 @@ public function search()
         return $dataProvider;
     }
 
+    /** @inheritdoc */
+    public function scopes()
+    {
+        return array(
+            'primary' => array('condition' => "parent_qid = 0"),
```
olleharstedt, line 1029: When is this scope applied? It won't break any other queries?

Shnoulle: > When is this scope applied? It won't break any other queries?

On demand only : Question::model()->primary()

Shnoulle: > When is this scope applied? It won't break any other queries?

On demand only : `Question::model()->primary() `

olleharstedt: Alright, great :+1:

PR #2144, opened 2021-11-16
Fixed issue #17696: Mass changing expiration date fails (except for superadmin)

gabrieljenik:


application/controllers/SurveyAdministrationController.php
```diff
@@ -2632,7 +2615,7 @@ public function changeTemplate($iSurveyID, $template, $aResults = null, $bReturn
 
         $survey = Survey::model()->findByPk($iSurveyID);
 
-        if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveyactivation', 'update')) {
+        if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveysettings', 'update')) {
```
Shnoulle, line 2618: User need read permission (usage) on template name too

application/controllers/SurveyAdministrationController.php
```diff
@@ -2632,7 +2615,7 @@ public function changeTemplate($iSurveyID, $template, $aResults = null, $bReturn
...
             if (!empty($bReturn)) {
```
Shnoulle, line 2619: See https://github.com/LimeSurvey/LimeSurvey/blob/0b3a1c0bcab8c5359595b027eafa2e0238318126/application/controllers/SurveyAdministrationController.php#L789 it's an empty array, then break if 1st survey don't have valid permission.

`!is_array` maybe ? Or isnull ?

Shnoulle (changes requested):
See permission on new template (read) (maybe at start ?) 
See !empty for return.

application/controllers/SurveyAdministrationController.php
gabrieljenik: why? not following. Please notice this was already working... we only changed the permission checking...

Shnoulle: Yes but before user have all rights (it was superadmin) then permission return always true this part was never used .

`$aResults` start by an empty array, then if user don't have permission on the 1st surey : it die … https://github.com/LimeSurvey/LimeSurvey/blob/0b3a1c0bcab8c5359595b027eafa2e0238318126/application/controllers/SurveyAdministrationController.php#L2625

gabrieljenik: No, it is not die-ing :)
You can see `$bReturn` is sent as true sometimes, which returns the errors on an array.

Shnoulle (approved):
Code OK
Sorry for error about aResult != breturn …

PR #2151, opened 2021-11-23
new feature #17754:  Inject questionAttributes to short free question twig views

TonisOrmisson:
New feature #17754: Inject questionAttributes to short free question twig views

### DIFF
--- a/application/helpers/qanda_helper.php
+++ b/application/helpers/qanda_helper.php
@@ -2479,7 +2479,8 @@ function do_shortfreetext($ia)
             'suffix'                 => $suffix,
             'inputsize'              => $inputsize,
             'placeholder'            => $placeholder,
-            'withColumn'             => $withColumn
+            'withColumn'             => $withColumn,
+            'questionAttributes'    => $aQuestionAttributes
             ), true);
     } elseif ((int) ($aQuestionAttributes['location_mapservice']) == 1) {
         $coreClass       = "ls-answers map-item geoloc-item";
@@ -2647,7 +2648,8 @@ function do_shortfreetext($ia)
             'maxlength' => $maxlength,
             'inputsize'              => $inputsize,
             'placeholder'            => $placeholder,
-            'withColumn'             => $withColumn
+            'withColumn'             => $withColumn,
+            'questionAttributes'    => $aQuestionAttributes
         );
         $answer = doRender('/survey/questions/answer/shortfreetext/text/item', $itemDatas, true);
     }
--- a/application/views/survey/questions/answer/shortfreetext/text/item.twig
+++ b/application/views/survey/questions/answer/shortfreetext/text/item.twig
@@ -12,6 +12,7 @@
  * $placeholder
  * $maxlength
  * $checkconditionFunction
+ * $questionAttributes
  */
 #}
 
--- a/application/views/survey/questions/answer/shortfreetext/textarea/item.twig
+++ b/application/views/survey/questions/answer/shortfreetext/textarea/item.twig
@@ -12,6 +12,7 @@
  * @var $checkconditionFunction      $checkconditionFunction.'(this.value, this.name, this.type)
  * @var $dispVal
  * @var $placeholder
+ * @var $questionAttributes
  */
 #}
 


Shnoulle:
Tested in 5.X ?
https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/QuestionTypes/ShortFreeText ?

TonisOrmisson:
@glimz What exactly you mean by "needs testing". 
Yes its tested manually on master branch, the data added to the array ends up to accessible in the twig view as the rest of the data in that array. There is no test written for that part of the code that I can adjust with the changes.

glimz:
Hello @TonisOrmisson, the labels are mostly for internal use. We label all PR, so we can easily filter to know the next line of action. We also need to test all PR internally before merging.

Shnoulle:
Else about `$aQuestionAttributes` and utf8 : we have an issue with it currently …

If you need `em_validation_q_tip` in a theme : how do you do ? @tpartner have an issue related to utf8 attributes reported in mantis

Can we (minimum) send  `$aQuestionAttributes`  for current language `$aQuestionAttributes['em_validation_q_tip']` return the tip for current language.

Else 
1. It's really hard to get `em_validation_q_tip` in twig 
2. we need to break API when fix it
3. Or we need to send more and more data to theme

Shnoulle:
I mean : 

```
  'prefix' =>  string '' (length=0)
  'suffix' =>  string '' (length=0)
```

And not

```
 'prefix' => 
    array (size=1)
      'en' => string '' (length=0)
  'suffix' => 
    array (size=1)
      'en' => string '' (length=0)`

```

TonisOrmisson:
This PR does is nothing about the contents of `$aQuestionAttributes`. It's just about passing the data to twig. 

The contents should not be a part of this issue. 

In my specific case I only need is_numeric attribute, but it seemed logical to have any possible question attribute in the twig to  decide how the question needs to be displayed. 

@glimz Why the wont-merge tag? I currently do not understand why not.

Shnoulle:
But : if you don't fix questionAttribute before update : you create a new issue …  Some workaround to get the attrbute in current language. Some other don't do this.
And when we fix : this broke API.

Without this : you just give more work for other dev … you already do a dev choice, since short(text receive `questionAttributes` like this : other quetsion type muts get it same way.

You already know we have a lot of old and deprecated code … please don't add again  and again and again new code without make it simple .

TonisOrmisson:
@Shnoulle I'm not sure In understand what is the issue this PR will instantly create?

The fact that some data inside the aQuestionAttributes is badly formatted or not well accessible is an entirely different issue than the principal decision whether question attributes should be accessible in the twig rendering the question or not. The issue of getting language-specific values from that array is not something this PR creates, this is something already existing. 

Yes, LS has awfully lot of bad code, but one can not fix all related issues with one PR. This PR is just using the code we have and addresses just one problem of not having the appropriate data (at all) in the view where it should be. Im not creating anything new, Im just passing what is already there to a place where this type of info should be accessible.

Shnoulle:
The point is API update : if you add it with 

```
 'prefix' => 
    array (size=1)
      'en' => string '' (length=0)
  'suffix' => 
    array (size=1)
      'en' => string '' (length=0)`

```

then any other dev can create a theme (survey or question) or plugin using this system. Then : we can  not improve it after without broke API compatibility.


> aQuestionAttributes is badly formatted or not well accessible is an entirely different issue than the principal decision

No because currently aQuestionAttributes is not inside API of theme or plugin.

> The issue of getting language-specific values from that array is not something this PR creates

Since it not send to twig : it something PR create.

>  but one can not fix all related issues with one PR

Yes, but here : you add a feature, this feature must be added in the good way. Because all bad way create more issue some month ago for other dev.

TonisOrmisson:
@Shnoulle Im not sure I understand what you mean by API here. 

QuestionAttribute is a core Model. Various attributes any question type has are documented, if I understand correctly you can find your suffix documented here:

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/survey/questions/answer/shortfreetext/config.xml#L75-L86

Any dev can look for what attributes are used for what question type and use them respectively. 

You mean have all attributes defined separately in here?

https://github.com/LimeSurvey/LimeSurvey/pull/2151/files#diff-4462db9ca8b48f96002561e9130e4305b349de87f84138318489c217f119716aR2483

Or where you mean to have the code you suggest?

glimz:
Hello @TonisOrmisson, the reason why I am not merging your PR is because it only addresses part of the problem. There is another feature request we will work on that will make the attributes available in all or most of the question types and not just free text question type. If I push this, it might affect the other development plan

TonisOrmisson:
@glimz thanks for the info. Sounds logical and good. Can you please refer the request? I did a search in mantis beforehand, but did not find any that would address this, but maybe I was just not thorough enough :)

glimz:
@TonisOrmisson we would sort out the specification for the 2 feature request soon [14228](https://bugs.limesurvey.org/view.php?id=14228) and [14248](https://bugs.limesurvey.org/view.php?id=14248). It is better to implement them all at once than pushing for just one question type.

Shnoulle:
> Any dev can look for what attributes are used for what question type and use them respectively.
> You mean have all attributes defined separately in here?

It's only related to i18n attributes.

Currently : no i18n attribute can be used like this after your commit `{{ questionAttribute.hide_tip  }}` then any orther attribute can be used like this : `{{ questionAttribute.suffix  }}`  but not … 
With you current implementation : `{{ questionAttribute.suffix  }}` didn't return the current suffix …

TonisOrmisson:
Again, its about the contents and formatting of the questionAttribute object, not about the principal passing of the object, a different matter. Hopefully the referred feature requests will offer a full solution making it all possible in a good way soon :)

PR #2152, opened 2021-11-23
Fixed issue #17698: Change of email address for survey administrator confusing

gabrieljenik:


application/controllers/admin/useraction.php
```diff
@@ -670,6 +667,25 @@ public function personalsettings()
                 }
             }
 
+            if (Yii::app()->request->getPost('newemailshown') == "1") {
+                if (Yii::app()->getConfig('demoMode')) {
+                    Yii::app()->setFlashMessage(gT("You can't change email if demo mode is active."), 'error');
+                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
```
Shnoulle, line 673: My opinion : it's better to not redirect NOW, but allow other update (else all are lost, not good for web quality) .
Best is to disable saveandclose
https://github.com/LimeSurvey/LimeSurvey/blob/41780805861f2dd88bf5513d9d12583593aeac02/application/controllers/admin/useraction.php#L725

something like 
`if($noerror && Yii::app()->request->getPost("saveandclose")) {`

Shnoulle (changes requested):
No redirect after error with old password. 

Save other and come back to current form with alert.

application/controllers/admin/useraction.php
gabrieljenik: > It's better to not redirect NOW, but allow other update (else all are lost, not good for web quality) .

This would be complex as then 2 alerts would appear. One saying "no passowrd changes are allowed on demo" and another one saying "some changes couldn't be done".

What we could do is to disable the ability to change some fields if on demo mode.
That would be the easiest.

application/controllers/admin/useraction.php
```diff
@@ -670,6 +667,25 @@ public function personalsettings()
...
+                }
+
+                $oldPassword = Yii::app()->request->getPost('oldpassword');
+
+                // Check the current password
+                $currentPasswordOk = $oUserModel->checkPassword($oldPassword);
+                if (!$currentPasswordOk) {
+                    Yii::app()->setFlashMessage(gT('The current password is not correct.'), 'error');
+                    $this->getController()->redirect(array("admin/user/sa/personalsettings"));
```
Shnoulle, line 682: Same here : else all other update are lost

application/controllers/admin/useraction.php
Shnoulle: > This would be complex as then 2 alerts would appear. One saying "no password changes are allowed on demo" and another one saying "some changes couldn't be done".

No : if you disable other save : user lost is update and did not know what is updated or not …

And her e: i put example on demo mode, but there are 2 point

1. Use password only to save new password and email, not the other things 
2. Then you must disallow update email (and show an error) but save other part 
3. If there are an error somewhere (demo, password, and anything else) : you must get back to the form.

Shnoulle (changes requested):
1. Use password only to save new password and email, not the other things 
2. Then you must disallow update email (and show an error) but save other part 
3. If there are an error somewhere (demo, password, and anything else) : you must get back to the form.

Shnoulle (approved):
OK, i report another mantis when it was merged …

gabrieljenik:
> 2\. Then you must disallow update email (and show an error) but save other part

Sorry, I made a typo before.
Instead of saying
> This would be complex as then 2 alerts would appear. One saying "no password changes are allowed on demo" and another one saying "some changes couldn't be done".

I was supposed to say
> This would be complex as then 2 alerts would appear. One saying "no email or password changes are allowed on demo" and another one saying "Your personal settings were successfully saved.".

Two toast messages, seems confusing to me.

PR #2158, opened 2021-11-23
Fixed issue #17618: Conversion failed when converting the nvarchar value 'NaN' to data type int

gabrieljenik:


Shnoulle:
Question : we can not use new1, new2 etc ?

Just increment a nomber ?

Shnoulle:
But if one day we have an user want to add `100000` answer (big computer and big server) 🤣

Else : when added it was a function, now seems we are inside a each : https://github.com/LimeSurvey/LimeSurvey/blob/c4d13bde9f4a237067e21f04d6f6dc6ac01c3540/assets/scripts/admin/questionEditor.js#L803
Can you check what happen if use new${i} ?

Shnoulle (approved):

PR #2161, opened 2021-11-25
Fixed issue #17762: vulnerable verson of jQuery included in qtip2

Shnoulle:
Dev: remove unusued qtip2
Dev: included in third_party, but never registered


gabrieljenik (approved):
Seems ok. I added @olleharstedt 
Extra  pair of eyes seems beetter.

Shnoulle:
Yep , qtip2 is use before, i think it's @olleharstedt that move to own system.

application/controllers/admin/translate.php
```diff
@@ -320,72 +320,6 @@ private function showTranslateAdminmenu($iSurveyID, $survey_title, $tolang)
         return $this->_getLanguageList($iSurveyID, $tolang);
     }
 
-    /*
-    * _getSurveyButton() returns test / execute survey button
-    * @param string $iSurveyID Survey id
-    * @param string $menuitem_url Menu item url
-    */
-    private function _getSurveyButton($iSurveyID, $menuitem_url)
```
olleharstedt, line 328: This function is unused, or?

Shnoulle: Private function is easy to check if unused : it's unused :)

PR #2162, opened 2021-11-25
Dev #15750: Send emails only to participants with incomplete responses

gabrieljenik:


application/models/TokenDynamic.php
```diff
@@ -174,7 +174,7 @@ public function findUninvited($aTokenIds = false, $iMaxEmails = 0, $bEmail = tru
         $command = new CDbCriteria();
         $command->condition = '';
         $command->addCondition("(completed ='N') or (completed='')");
-        $command->addCondition("token <> ''");
+        $command->addCondition("`t`.token <> ''");
```
Shnoulle, line 177: SQL issue here : t.token , no ``t``

application/controllers/admin/tokens.php
```diff
@@ -1366,10 +1366,21 @@ public function email(int $iSurveyId)
...
+                        'condition' => 'responses.submitdate IS NULL',
+                    ],
+                ]);
+            }
+
+            $ctresult = $tokenDynamic->findUninvited($aTokenIds, 0, $bIsInvitation, $SQLemailstatuscondition, $SQLremindercountcondition, $SQLreminderdelaycondition);
             $ctcount = count($ctresult);
 
-            $emresult = TokenDynamic::model($iSurveyId)->findUninvited($aTokenIds, $iMaxEmails, $bIsInvitation, $SQLemailstatuscondition, $SQLremindercountcondition, $SQLreminderdelaycondition);
+            $emresult = array_slice($ctresult, 0, $iMaxEmails);
```
Shnoulle, line 1383: Why this update here ? Seems not related

Shnoulle (changes requested):
SQL issue with pg or ms

application/controllers/admin/tokens.php
gabrieljenik: Previoulsy a query was done to get the number of uninvited tokens and then a new one to pickup the first 50.
Now we query all, count the uninvited tokens and slice to get 50 ones.

olleharstedt: Hm, will this scale with very large number of tokens?

Shnoulle (approved):

application/controllers/admin/tokens.php
gabrieljenik: Well ctresult was always there.
We are avoding the extra query for getting $emresult

olleharstedt: Ah ok, so no extra work, then.

Shnoulle: Maybe some optimisation to do ? No ?

1.  $ctcount = TokenDynamic::model($iSurveyId)->count
2. $emresult = TokenDynamic::model($iSurveyId)->findUninvited($aTokenIds, $iMaxEmails

then we don't get wole to use only 50, we count whole (return one value) and get onkly 50 after ?

olleharstedt:
Thanks!

application/controllers/admin/tokens.php
gabrieljenik: ctcount doesn't count all, only uninvited.
https://github.com/LimeSurvey/LimeSurvey/pull/2162/files/9d556f4175bdd4b21ed4ced583999f233ae3f207#diff-1ef75a5722a7a05aed05e9f54dedb0ec1571871a1996c5ade9789e35a095e2ddR1380

Still, yes, that way we would only count and not load all uninvited.
2 queries vs (load + manage it in memory).

I like 2 queries. What do you think?

Shnoulle: I think 1 queries to count + one query with only 50 token are less memory consuming than one query to have whole tokens (eventually 10000 tokens).

With 2 quesris : we are sure memory is limited. Maybe sometimes it's not needed (for example : have only 50 participant) but the na double call don't broke a lot.

**This is not related here** since request for all came from previous code . It's a code improvment :)

olleharstedt:
> SQL issue with pg or ms

What what?

Shnoulle:
> > SQL issue with pg or ms
> 
> What what?

Fixed in https://github.com/LimeSurvey/LimeSurvey/pull/2162/commits/ed91d74c8b7a541040458c975a73873007c326fd

PR #2163, opened 2021-11-26
17492: Refactor updatedb helper into multiple command classes

olleharstedt:
Info here: https://bugs.limesurvey.org/view.php?id=17492


application/helpers/update/DatabaseUpdateBase.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    private function updateVersion()
+    {
+        $this->db
+            ->createCommand()
+            ->update('{{settings_global}}', ['stg_value' => $this->newVersion], "stg_name='DBVersion'");
+    }
+
+    /**
+     * Get db version number based on class name, e.g. 123 for Update_123
```
Shnoulle, line 68: Why not for `Update_` ?

olleharstedt: Well. Maybe. Two things to think of:

1) Future proof: at some point we'll have Update_XXXX
2) Hacker proof? We might want to introduce timestamps or "Update_123b" or other things. That can be added later, of course.

But sure, you can throw exception if first part of explode() is not "Update".

Shnoulle: Unsure here … but think it's always best to start by more restriction than needed. And remove restriction after.

I read about number or date or … in mantis. No good advice …

olleharstedt: Pushed a check for $nameParts[0]

Shnoulle: But i din't ask for ;)

olleharstedt: Probably a good idea.

Shnoulle (approved):
I like the update … and don't see code issue.

Maybe some part of manual to be updated after.

olleharstedt:
Yes, it can be adjusted later on. I'll merge this now, then.

Shnoulle:
> Yes, it can be adjusted later on. I'll merge this now, then.

Oh yes ! Sure !

PR #2165, opened 2021-11-29
Fixed issue #17765: The order of the answer options not correct when exporting to printable survey or queXML

c-schmitz:


olleharstedt:
That was quick :)

Shnoulle:
No default scope on answer ?

c-schmitz:
I was afraid of setting a default scope. Did not want to compromise other stuff like alpha_sort

Shnoulle:
> I was afraid of setting a default scope. Did not want to compromise other stuff like alpha_sort

Yes, i think default_scope order are only when there are no other order :). Maybe develop ?

PR #2167, opened 2021-12-01
Fixed issue #17732: Direct access to response browsing break

gabrieljenik:


Shnoulle:
Since it('s surely needed in other controller : in LSBaseController constructor ?

gabrieljenik:
> Since it('s surely needed in other controller : in LSBaseController constructor ?

Something like a new method?
How would you name it?

Shnoulle:
Unsure … 

My question is more because before it was somewhere in admin global controller, and we surely need it on other admin related controller.

PR #2169, opened 2021-12-02
Dev #T207: Make it possible to change the base language of a survey

gabrieljenik:


application/extensions/admin/survey/LanguagesWidget/LanguagesWidget.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+class LanguagesWidget extends WhSelect2
```
olleharstedt, line 3: Hm, why is a new widget needed for a simple select box? Or did I miss something?

olleharstedt: And why is that needed?

olleharstedt: Ah ok. That's a good answer. :)

gabrieljenik: > Ah ok. That's a good answer. :)

I have my moments :)

PR #2175, opened 2021-12-08
Fixed issue #17787: open_base_dir + debug can throw error when load s…

Shnoulle:
Fixed issue #17787: open_base_dir + debug can throw error when load some question type.
Dev: try to load any files in any directory are always an error

### DIFF
--- a/application/models/QuestionTemplate.php
+++ b/application/models/QuestionTemplate.php
@@ -103,7 +103,7 @@ class QuestionTemplate extends CFormModel
     {
         if (!isset($this->aViews[$sView])) {
             $sTemplatePath = $this->getTemplatePath();
-            if (is_file("$sTemplatePath/$sView.twig")) {
+            if (!empty($sTemplatePath) && is_file("$sTemplatePath/$sView.twig")) {
                 $this->aViews[$sView] = true;
             } else {
                 $this->aViews[$sView] = false;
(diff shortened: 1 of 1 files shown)


Shnoulle:
@olleharstedt see https://github.com/LimeSurvey/LimeSurvey/commit/cb8043fee0ac1b3ac1513c571984a1c313e3ba2b

I don't understand here this can never work except for core question theme … 

1. This don't take application/view/surveys/questions theme
2. directory for uploaded theme is broken : always (here). since it's not related to rootdir.

Maybe we need to remove whole system if it's unused ?

olleharstedt:
Hm, can never work, but does it work? o0

olleharstedt:
Can you write testing instructions for this commit, please?

Shnoulle:
1. Set open_base_dir to /var/:/tmp/:/home/ for example
2. Activate debug mode
3. Import included survey
4. preview question

https://bugs.limesurvey.org/view.php?id=17787

Shnoulle:
> What I can see, it's using rootdir. But you mean it shouldn't?

1. https://github.com/LimeSurvey/LimeSurvey/blob/0bb1de73246e9dd913420ba7bc20cbb2ec9520c5/application/models/QuestionTemplate.php#L124
2. https://github.com/LimeSurvey/LimeSurvey/blob/0bb1de73246e9dd913420ba7bc20cbb2ec9520c5/application/config/config-defaults.php#L733

no rootdir …

Just put a log or tracevar at https://github.com/LimeSurvey/LimeSurvey/blob/0bb1de73246e9dd913420ba7bc20cbb2ec9520c5/application/models/QuestionTemplate.php#L125

c-schmitz:
Thank you!

PR #2177, opened 2021-12-10
Fixed issue #17677: Valid TSV file can not be used in VV import

Shnoulle:
Dev: allow line feed, checked with latin1 file too.

Dev:

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -1891,34 +1891,30 @@ function CSVImportResponses($sFullFilePath, $iSurveyId, $aOptions = array())
     if (!isset($aOptions['bDeleteFistLine'])) {$aOptions['bDeleteFistLine'] = true; } // By default delete first line (vvimport)
     if (!isset($aOptions['sExistingId'])) {$aOptions['sExistingId'] = "ignore"; } // By default exclude existing id
     if (!isset($aOptions['bNotFinalized'])) {$aOptions['bNotFinalized'] = false; } // By default don't change finalized part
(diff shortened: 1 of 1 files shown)


application/helpers/admin/import_helper.php
```diff
@@ -1891,34 +1891,30 @@ function CSVImportResponses($sFullFilePath, $iSurveyId, $aOptions = array())
...
-        $aOptions['sCharset'] = "utf8";
-    }
 
     // Prepare an array of sentence for result
     $CSVImportResult = array();
-    // Read the file
-    $handle = fopen($sFullFilePath, "r"); // Need to be adapted for Mac ? in options ?
-    if ($handle === false) {
-        safeDie("Can't open file");
+    $tmpVVFile = fileCsvToUtf8($sFullFilePath, $aOptions['sCharset']);
```
gabrieljenik, line 1903: If this function already reads the file, and this function is only used here (guessing), why don't this funciton just returns an array?

gabrieljenik (approved):
Code OK from structure point of view.
Haven't tested it

application/helpers/admin/import_helper.php
Shnoulle: Think it can be used elsewhere after : in TXT import/export for example.

Else : function does one think : return an utf8 file :)

Shnoulle: Else : 2 places : here + TSV import (taken from TSVImportSurvey function)

https://github.com/LimeSurvey/LimeSurvey/pull/2177/files#diff-4033daa1797e130ed555f4b345ed46387aa79c5eb457adf216ea049b648e5976R2222

Shnoulle:
master version : https://github.com/LimeSurvey/LimeSurvey/pull/2273

PR #2178, opened 2021-12-10
Dev #T206: Add 'Default answers' as a tab in advanced settings

gabrieljenik:
## Release Notes

* **New Features**
  * Added "Default Answers" tab to question administration for applicable question types
  * Enabled default answer configuration for 5-point choice questions
  * Added default value editors for text, date, scale, and yes/no question types
  * Implemented automatic synchronization of default answers with available options


Shnoulle:
No access on Dev #T206 : can we have a mantis issue ?

gabrieljenik:
> No access on Dev #T206 : can we have a mantis issue ?

https://bugs.limesurvey.org/view.php?id=18251

Shnoulle:
Need to delete current controller + views

ptelu:
What is the status here? Doesn't seem to work properly. Can not even see the Default Answers tab or button anywhere. And creating questions is resulting in an error since the new implementation expects that the question type already exists, which it can't.

gabrieljenik:
> What is the status here? Doesn't seem to work properly. Can not even see the Default Answers tab or button anywhere. And creating questions is resulting in an error since the new implementation expects that the question type already exists, which it can't.

@ptelu 

Done:
- Moved default values to main screen
- Set different user interface for different question types.

Pending:
- User interface for Date question type (not far away from it)

gabrieljenik:
Updated info: 
- The PR also adds support for 5 point choice question type 
- Besides date question type, there could be other question types pending to be implemented.

@ptelu Let me know if we should finish it (make the screens for the pending question types) or if you will take it from here on. Thansk

c-schmitz:
Any chance of finishing this up or should we close this?

gabrieljenik:
From what I recall, this was to be followed up by ptelu. (This was 3 years ago -- so my memory could fail on me) Not sure the current status (maybe there is another branch of this which I don't know). Do you want me to pick it up?

c-schmitz:
@ptelu ?

c-schmitz:
@coderabbitai review

gabrieljenik:
> Any chance of finishing this up or should we close this?

From what we reviewed, the best would be to kind of port the changes of this PR (which is based on v5) to a new PR v6-based. Shall we?

c-schmitz:
@gabrieljenik A new possible implementation would only go into the new question editor. Do you feel you could do that?

PR #2179, opened 2021-12-14
Fixed issue #17711: Cannot export selected responses (only range)

gabrieljenik:
This was done as there was some issue of variable context when evaluating postUrl.
https://github.com/LimeSurvey/LimeSurvey/compare/master...gabrieljenik:bug/17711--Cannot-export-selected-responses---only-range?expand=1#diff-ebcc738a58028c4733ffa427dbbf7b683f9b6f6513f97ce50414c7902dd48c79L71

The global postUrl variables was ignored.
It seems, somehow JS was defining postUrl as local from the very beggining, overwritting the global one. Even if `var postUrl` wasn't evaluated yet.


Shnoulle (approved):
Don't see issue on this commit.

PR #2183, opened 2021-12-16
Dev #14878: Fixed order for 'Question attribute Category'

gabrieljenik:


application/models/services/QuestionAttributeHelper.php
```diff
@@ -218,4 +218,62 @@ public function getQuestionAttributesWithValues($question, $language = null, $qu
...
+        return $result;
+    }
+
+    /**
+     * Returns the array of categories with their assigned order.
+     * The array doesn't contain all the posible categories, only those with an order assigned.
+     *
+     * @return array<string,int>
+     */
+    public function getCategoryOrders()
```
Shnoulle, line 251: And when a plugin or a QuestionTheme add an new category ?

gabrieljenik: If is not listed there (ex: custom category), then it will be sent to the bottom.
https://github.com/LimeSurvey/LimeSurvey/pull/2183/files#diff-e2134989f5e14268a9ac8f96c947c6dd63dc3770523fef3a3cf16178e5b7975cR232

Shnoulle: You have an idea to get it a better place than after statistics or Time ?

gabrieljenik: Maybe a new plugin event getCategorySettings ?
We could overwrite the std category settings and give settings to custom ones.

Shnoulle: But for QuestionTheme ? A category order in XML ?

Shnoulle: PS : why not in same order than the XML ? Or the declaration ?

gabrieljenik: > But for QuestionTheme ? A category order in XML ?

We could add some section on the XML of the question themes, where they set the attributes of categories they are including.

> PS : why not in same order than the XML ? Or the declaration ?

What do you mean? Can you provide an example?

Shnoulle: See : https://github.com/LimeSurvey/LimeSurvey/blob/998fa2a5123b373ba0815f1551540b5952267fc5/application/views/survey/questions/answer/arrays/increasesamedecrease/config.xml

Display is 1st : https://github.com/LimeSurvey/LimeSurvey/blob/998fa2a5123b373ba0815f1551540b5952267fc5/application/views/survey/questions/answer/arrays/increasesamedecrease/config.xml#L64
Logic : https://github.com/LimeSurvey/LimeSurvey/blob/998fa2a5123b373ba0815f1551540b5952267fc5/application/views/survey/questions/answer/arrays/increasesamedecrease/config.xml#L174
Other : https://github.com/LimeSurvey/LimeSurvey/blob/998fa2a5123b373ba0815f1551540b5952267fc5/application/views/survey/questions/answer/arrays/increasesamedecrease/config.xml#L296
Statistics : https://github.com/LimeSurvey/LimeSurvey/blob/998fa2a5123b373ba0815f1551540b5952267fc5/application/views/survey/questions/answer/arrays/increasesamedecrease/config.xml#L352

And the a new theme can do (for example here)
```
    <!-- Question attributes -->
    <attributes>
        <attribute>
            <category>Display</category>
        </attribute>
        <attribute>
            <category>Logic</category>
        </attribute>
        <attribute>
            <name>my_attribute</name>
            <category>My category</category>
            <sortorder>11</sortorder>
            <inputtype>text</inputtype>
            <expression>2</expression>
            <help>helpo</help>
            <caption>Caption</caption>
            <i18n></i18n>
            <readonly></readonly>
            <readonly_when_active></readonly_when_active>
        </attribute>
    </attributes>
```

Or maybe directly in XML of the question ?

```
<categories>
<category>Display</category>
<category>Logic</category>
<category>Other</category>
<category>Statistics</category>
</categories>
```

Get parent order if not exist (need to be created in core XML)

gabrieljenik: > Or maybe directly in XML of the question ?

This is interesting. I was thinking something like that.
But we can add more details (child nodes) to each category.
Will be usefull for the future.

Still, I think all this goes inside a new feature ticket

Shnoulle: OK … you create the mantis issue for discussion ?

Shnoulle (approved):

application/models/services/QuestionAttributeHelper.php
gabrieljenik: https://bugs.limesurvey.org/view.php?id=17814

PR #2188, opened 2021-12-21
Fixed issue #17751: Date stamped throw php notice when data edit and debug

Shnoulle:


Shnoulle:
1. Move to boostrap widget with current user date settings like Survey settings and token edit
2. Add required attribute
3. Set a default value in all condition

gabrieljenik (approved):

Shnoulle:
> Code OK. Seems similar to the one pushed for v5, right? Not tested.

Yes : near same fix

PR #2191, opened 2021-12-23
Dev #7367: 'Other answer' at a certain position in the list

gabrieljenik:


application/views/survey/questions/answer/list_dropdown/config.xml
```diff
@@ -142,6 +142,49 @@
             <readonly></readonly>
             <readonly_when_active></readonly_when_active>
         </attribute>
+        <attribute>
+            <name>other_position</name>
+            <category>Display</category>
+            <sortorder>150</sortorder>
+            <inputtype>singleselect</inputtype>
```
Shnoulle, line 149: Why use integer ? Why not simple string ?

```
           <options>
                <option>
                    <value>begin</value>
                    <text>At beginning</text>
                </option>
                <option>
                    <value>default</value>
                    <text>Before No Answer</text>
                </option>
                <option>
                    <value>end</value>
                    <text>At end</text>
                </option>
                <option>
                    <value>specific</value>
                    <text>After specific answer option</text>
                </option>
           </options>
           <default>default</default>
```

application/core/QuestionTypes/ListDropdown/RenderListDropdown.php
```diff
@@ -21,9 +21,26 @@ class RenderListDropdown extends QuestionBaseRenderer
...
+
+    /** @var string the code of the answer after which the 'Other' option should be placed (if $otherPosition == 3) */
+    protected $answerBeforeOther;
+
+    /** @var boolean indicates if 'Other' option has already been rendered */
+    protected $otherRendered = false;
     
     private $iRowNum = 0;
 
+    const OTHER_POS_BEFORE_NOANSWER = 0;
```
Shnoulle, line 39: `const OTHER_POS_DEFAULT = 'default';` etc …

Shnoulle (changes requested):
For discussion : 

1. I prefer string than integer 
2. No test survey ? New functionality : new test. And here a sample lss with the 3 question type + some option set.

gabrieljenik:
Test added

Shnoulle:
Thank you for string in place of integer .
You're OK too ?

Shnoulle (approved):

application/core/QuestionTypes/ListDropdown/RenderListDropdown.php
```diff
@@ -21,9 +21,26 @@ class RenderListDropdown extends QuestionBaseRenderer
...
+
+    /** @var string the code of the answer after which the 'Other' option should be placed (if $otherPosition == 3) */
+    protected $answerBeforeOther;
+
+    /** @var boolean indicates if 'Other' option has already been rendered */
+    protected $otherRendered = false;
     
     private $iRowNum = 0;
 
+    const OTHER_POS_BEFORE_NOANSWER = 'default';
```
olleharstedt, line 39: These and other lines seems duplicated in lots of classes? Can we move it out to a trait instead?

gabrieljenik: I will create a new DEV ticket for that

gabrieljenik: https://bugs.limesurvey.org/view.php?id=18489

PR #2193, opened 2021-12-27
Fixed issue #17518: Non-ascii characters in email addresses

gabrieljenik:
When loading a token by the add participant option, the HTML5 validation is run and doesn't allow to add emails with "non-ascii characters". That's Chrome.

Still, going further, the previous code uses the idna_convert class to convert from unicode to punycode during validation. That created two problems:

1) Conversion was applied to the entire address, not just the domain. That is why unicode characters were being accepted on the local part of the address, which is not supposed to be accepted.
2) The idna_convert class does the conversion always, while PHPMailer requires the intl extension to be installed to support IDN addresses. Therefore, some addresses could be accepted which later PHPMailer was not going to be able to handle.

This fix leaves out the idna_convert class (does it have to be removed?) And uses PHPMailer's punyencodeAddress () method, for better consistency.


application/core/LimeMailer.php
```diff
@@ -767,17 +767,19 @@ public static function validateAddress($address, $patternselect = null)
...
-        $oIdnConverter = new idna_convert();
-        $sEmailAddress = $oIdnConverter->encode($address);
-        $bResult = filter_var($sEmailAddress, FILTER_VALIDATE_EMAIL);
-        if ($bResult !== false) {
-            return true;
-        }
-        return false;
+        if ($patternselect == 'php-idna') {
+            /**
+             * PHPMailer has support for IDN, but it relies on `intl` and `mbstring`.
```
Shnoulle, line 772: We have `intl` and `mbstring` no ?

I create this when move from old system to LimeMailer  to use same fuinction, but maybeit's bet to use PJPMailer directly ?

application/core/LimeMailer.php
```diff
@@ -767,17 +767,19 @@ public static function validateAddress($address, $patternselect = null)
...
+             * If we use 'idna_convert' as we did, we are not ensuring that PHPMailer
+             * can handle the address later. So instead of using 'idna_convert' we use
+             * PHPMailer's punyencodeAddress() method.
+             */
+            $mailer = new \PHPMailer\PHPMailer\PHPMailer(); // Can't use an instance of LimeMailer because of a recursion problem
```
Shnoulle, line 777: No way to use parent::punyencodeAddress ? or this->punyencodeAddress ?

Shnoulle (commented):
I need to understand why we use php-idna before .
I think it's best if we can use only PHPMailer system ? Then we can update more easily.

application/core/LimeMailer.php
gabrieljenik: Nope, there is a recursion problem

gabrieljenik:
> I need to understand why we use php-idna before . I think it's best if we can use only PHPMailer system ? Then we can update more easily.

Now we are using it. :)

> If we can : remove and use only PHPMailer

Will remove idna_convert class

application/core/LimeMailer.php
Shnoulle: Strange  ?

Shnoulle:
Maybe need to check https://github.com/LimeSurvey/LimeSurvey/blob/a3dda2b515babc7fae9a349ae500a03dd8e9f484/application/third_party/phpmailer/src/PHPMailer.php#L1412 ?

And validate with html5 : https://github.com/LimeSurvey/LimeSurvey/blob/a3dda2b515babc7fae9a349ae500a03dd8e9f484/application/third_party/phpmailer/src/PHPMailer.php#L1389 if not available ?

PR #2194, opened 2021-12-28
Fixed issue #17685: When adding a new participant, user is able to Save without adding any information and a blank record is save

gabrieljenik:
Added confirmation modal when blank participant is created.

### DIFF
--- a/assets/scripts/admin/tokens.js
+++ b/assets/scripts/admin/tokens.js
@@ -208,6 +208,18 @@ function validateAdditionalAttributes() {
     return valid;
 }
 
+function validateEmptyTokenForm() {
+    if ($('#edittoken').find('[name="subaction"]').val() != 'inserttoken') {
+        return true;
+    }
+    var isFormEmpty = $('#email').val() == '' && $('#firstname').val() == '' && $('#lastname').val() == '';
+    if (isFormEmpty) {
+        $('#emptyTokenConfirmationModal').modal('show');
+        return false;
+    }
+    return true;
+}
+
 /**
  * Scroll the pager and the footer when scrolling horizontally
  */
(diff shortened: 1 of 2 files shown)


olleharstedt (changes requested):

assets/scripts/admin/tokens.js
```diff
@@ -208,6 +208,18 @@ function validateAdditionalAttributes() {
     return valid;
 }
 
+function validateEmptyTokenForm() {
+    if ($('#edittoken').find('[name="subaction"]').val() != 'inserttoken') {
+        return true;
+    }
+    var isFormEmpty = $('#email').val() == '' && $('#firstname').val() == '' && $('#lastname').val() == '';
```
Shnoulle, line 215: Need to check attribute too : (for example) client id set.

Shnoulle (changes requested):
attribute must be checked too.

assets/scripts/admin/tokens.js
gabrieljenik: Sorry, don't understand. What is "client id"?
Is that a custom attribute?

We can't check emptyness with attributes which are not always going to be used.

Shnoulle: attribute_1 for example 

isFormEmpty must check if any value is entered : 

1. can be default one (email, firstname, lastname) : OK
2. attribute_X if exist
3. token 
4. others ? mailsatus (!OK), usesleft (!1)? etc …

1 + 2 + 3 mlust be checked. for 4 : we can leave it like this.

Shnoulle (changes requested):
Must check if attribute_X or token is filled too.

assets/scripts/admin/tokens.js
gabrieljenik: This is just a warning for preventing blank records. Not blocking anything.
Checking for token and all possible attribute isn't it going too far?

Shnoulle: It's not only a warning… it take time … clearly enter anything in attribute or token must not be confirmed.

olleharstedt:
> Must check if attribute_X or token is filled too.

Is this a blocking issue? Or can I merge?

Shnoulle:
> > Must check if attribute_X or token is filled too.
> 
> Is this a blocking issue? Or can I merge?

It's clearly a Ergonomy + Usuability issue … 

Do waht you want, don't care … Ergonomy of LS 5 is totally broken … nobody care …

olleharstedt:
@gabrieljenik Any opinion? Otherwise, just merge. You can also create a separate issue for Denis' concern.

gabrieljenik:
> Any opinion? 

@glimz or someone else should review the scope of the task.

> Otherwise, just merge. You can also create a separate issue for Denis' concern.

Yes, after merging, we could complement it as well.

PR #2195, opened 2021-12-28
Fixed issue #17801: No way to go to submit button in create group or question form with keyboard

gabrieljenik:
Added tabindex to save buttons


Shnoulle:
https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/

I didn't check, but unsure it's the good solution here.

gabrieljenik:
What's your suggestion (without major chaning how the screen looks like)?
We had to add the tabindex to get to the button with the keyboard. Not sure why we were not getting there.

Shnoulle:
> What's your suggestion (without major chaning how the screen looks like)?

Add the action button at bottom, always ;)
I don't have any idea, but in general : adding tabindex can break easily when update HTML code …


> Not sure why we were not getting there.

The issue is here , it's a link (a) then there are no reason to add tabindex.

gabrieljenik:
> The issue is here , it's a link (a) then there are no reason to add tabindex.

It is a link but with no href. That's why we need a tab index.
https://stackoverflow.com/a/30740305/1392917

Maybe the best solution would be to change the tags from A to Button... will try..!

Shnoulle:
Yes : button seems better :)
If it work : :+1: 
Thank you for checking !

gabrieljenik:
What about now?

Shnoulle:
Can be done :+1:

Lot of issue with keyboard, but can be done the n it's OK since «Accessibility of the administration interface is currently not a planned feature.»

Shnoulle (approved):
Access can be done, still awfull but need more to fix all a11y issue in admin page.

Shnoulle:
@glimz already reviewed : «Access can be done, still awfull but need more to fix all a11y issue in admin page.» according to @c-schmitz on issue 17748 : «we would accept third-party contributions if someone wants to take on this task» ;)

PR #2197, opened 2021-12-29
Fixed issue #17802: Error shown when try to load a saved survey

gabrieljenik:


gabrieljenik:
@olleharstedt This is failing as it says a method is not used.
https://github.com/LimeSurvey/LimeSurvey/runs/4660622797?check_suite_focus=true#step:9:18

Still, is used here as reflection.
https://github.com/LimeSurvey/LimeSurvey/commit/f4d4c1b65b4eb3ca5240a74052eb42f0d8848790#diff-e2134989f5e14268a9ac8f96c947c6dd63dc3770523fef3a3cf16178e5b7975cR273

Maybe we could by pass the rule on this test?
Not sure how to handle it.

olleharstedt:
Hm, let me check. Maybe it shouldn't be a private method at all, just an anonymous function instead.

olleharstedt:
Oh, you already fixed this one? :) Sorry, I was on vacation.

gabrieljenik:
> Still, is used here as reflection.

Sorry, callback for the sort method.

Nope, haven't fixed this yet.
I think the syntax checker rule is wrong, as it doesn't consider dynamic function calling.

olleharstedt:
>  I think the syntax checker rule is wrong, as it doesn't consider dynamic function calling.

Yes and no. There is no state here, so no class is needed at all, in fact. But maybe setting it to "protected" and add a "todo" about it is enough right now.

olleharstedt:
Pushed a "fix"

olleharstedt:
Sorry, the correct rule-of-thumb should be "no state and no side-effects" for using a function instead of class method.

Shnoulle:
Maybe add an annotation https://phpmd.org/documentation/suppress-warnings.html ?

olleharstedt:
@Shnoulle Yes, warning can also be suppressed when needed. :)

gabrieljenik:
Does the checker throws yellow lights?
For warnings...

olleharstedt:
No, warnings are fatal errors for PHP Mess Detector.

Shnoulle:
> Does the checker throws yellow lights? For warnings...

😁 … i made an error : can try @phpcs:ignore UnusedPrivateMethod ?

olleharstedt:
@Shnoulle phpcs is code sniffer, but the error for private function is thrown by PHP Mess Detector. Two different checkers.

olleharstedt:
Yep

Shnoulle (approved):

application/controllers/survey/index.php
```diff
@@ -425,11 +425,16 @@ public function action()
...
-                    Yii::app()->setConfig('move', 'reload');
-                    $move = "reload"; // veyRunTimeHelper use $move in $arg
+                // If no name and pass are set in the request, don't try to load an answer. Just show the loading form.
+                if (empty(Yii::app()->request->getParam('loadname')) && empty(Yii::app()->request->getParam('loadpass'))) {
+                    Yii::app()->setConfig('move', "loadall"); // Show loading form
                 } else {
-                    $aLoadErrorMsg['matching'] = gT("There is no matching saved response.");
+                    if (loadanswers()) {
+                        Yii::app()->setConfig('move', 'reload');
+                        $move = "reload"; // veyRunTimeHelper use $move in $arg
```
olleharstedt, line 434: veyRunTimeHelper - tiny typo here?

gabrieljenik: Fixed

PR #2198, opened 2021-12-30
Fixed issue #17776: New security settings

gabrieljenik:


olleharstedt:
New feature, should be in dev, or?

gabrieljenik:
Well, it is a security issue.
That's why I set it on master.

olleharstedt:
:thinking:

application/config/config-defaults.php
```diff
@@ -804,6 +804,8 @@
 
 $config['pluginWhitelist'] = [];
 
+$config['disablePluginUpload'] = false;
```
Shnoulle, line 807: Add comment please :)

application/config/config-defaults.php
```diff
@@ -804,6 +804,8 @@
 
 $config['pluginWhitelist'] = [];
```
Shnoulle, line 805: Can you add comment on this one ? For https://github.com/LimeSurvey/LimeSurvey/blob/64fb578c5f0fddc0ea89fe4ede27cd44a3f29d82/application/config/config-defaults.php#L805 something like 

> User plugin list allowed to be installed. This don't disallow core plugin, but disable already installed plugin action.

Shnoulle (changes requested):
Need comments

Else : App()->getConfig('pluginCoreList') : really ? In my opibnion : it must be a fixed static variables here

olleharstedt:
For the failures in CI, you can merge master into this branch again. Should be fixed.

application/libraries/PluginManager/PluginManager.php
```diff
@@ -634,17 +638,38 @@ protected function getPluginName(string $class, \ExtensionConfig $extensionConfi
...
+            'AuditLog',
+            'ExportR',
+            'ExportSTATAxml',
+            'ExportSPSSsav',
+            'extendedStartPage',
+            'oldUrlCompat',
+            'AuthLDAP',
+            'Authdb',
+            'Authwebserver'
+        ];
```
Shnoulle, line 673: Lack of some core plugin : different ways to fix 

- don't allow some core plugins (maybe disallow Authwebserver for example)
- fix the list adding all core plugins
- get the list from directory (and cache it)

application/views/admin/super/fullpagebar_view.php
```diff
@@ -17,7 +17,8 @@
                 <!-- Install Plugin Zip -->
                 <?php if (
                     isset($fullpagebar['pluginManager']['buttons']['installPluginZipModal']['hasConfigDemoMode']) &&
-                    !$fullpagebar['pluginManager']['buttons']['installPluginZipModal']['hasConfigDemoMode']
+                    !$fullpagebar['pluginManager']['buttons']['installPluginZipModal']['hasConfigDemoMode'] &&
+                    !Yii::app()->getConfig('disablePluginUpload')
```
Shnoulle, line 21: Adding a boolean `shownUpload` in controller maybe ? 

PS : i really think this isset is directy …

Shnoulle (changes requested):
More remark than change needed.

But need to be sure for list of core plugins

application/libraries/PluginManager/PluginManager.php
Shnoulle: Sorry for the list i give … (copy pasted from yours)

gabrieljenik: Sorry, what do you mean by `Lack of some core plugin : different ways to fix` ?

application/views/admin/super/fullpagebar_view.php
gabrieljenik: You mean for not showing the button at all if uploads are disabled?

application/libraries/PluginManager/PluginManager.php
Shnoulle: Lack of some plugins : https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/plugins 
All core plugins not included here :)

Shnoulle: Then the 3 point is 3 way to fix it :) Why allow AuthWebsrevr (that can broke access) and disallow expressionFixedDbVar ?

Shnoulle: In fact your 1st idea to have a config array for this is not a bad solution … but name it 'allowedCorePlugin' and always force AuthDB (currently)

gabrieljenik: I am not sure about changing the name.
If this list is `allowedCorePlugins`, would there be a `notAllowedCorePlugins`? would there be any core plugin not allowed?

gabrieljenik: I suggest that disabling core plugins should be another story.

Shnoulle: > I suggest that disabling core plugins should be another story.

Then you need to fix the list : expressionFixedDbVar, expressionQuestionHelp … etc … :).

By «fix the list adding all core plugins» or «get the list from directory (and cache it)»

gabrieljenik: > fix the list adding all core plugin

Agree.

application/views/admin/super/fullpagebar_view.php
Shnoulle: Thank you

Shnoulle:
Seems action can be done in demoMode ? : another issue or not ?

gabrieljenik:
> Seems action can be done in demoMode ? : another issue or not ?

I think that's another issue.

Shnoulle (approved):

Shnoulle:
> > Seems action can be done in demoMode ? : another issue or not ?
> 
> I think that's another issue.

Yes : same, you report it after this one is merged ?

PR #2201, opened 2022-01-06
Fixed issue #17773: Code name change while we updated from Edit Quest…

gabrieljenik:
Made the labelsets long enough for hlding answer codes (5 chars) and subquestion codes (>= 5).
If a labelset thought for a subquestion code is imported on answer options, codes are trimed and alerts are shown


Shnoulle (approved):
No code issue

PR #2202, opened 2022-01-06
Dev #T266: Implement CPDB blacklist functions

gabrieljenik:
WIP
Implementing cpdb: optout, optin and other cpdb specific features


application/controllers/OptinController.php
```diff
@@ -27,53 +27,183 @@ class OptinController extends LSYii_Controller
...
+        $surveyId = Yii::app()->request->getQuery('surveyid');
+        $languageCode = Yii::app()->request->getQuery('langcode');
+        $accessToken = Token::sanitizeToken(Yii::app()->request->getQuery('token'));
+
+        //IF there is no survey id, redirect back to the default public page
+        if (!$surveyId) {
+            $this->redirect(['/']);
+        }
+
+        $surveyId = (int) $surveyId; //Make sure it's an integer (protect from SQL injects)
```
Shnoulle, line 98: No : just try the findByPk : no need to cast as int. Return a 404 if not exist

application/controllers/OptinController.php
```diff
@@ -27,53 +27,183 @@ class OptinController extends LSYii_Controller
...
+        if (!isset($languageCode) || $languageCode == "" || !$languageCode) {
+            $baseLanguage = $survey->language;
+        } else {
+            $baseLanguage = sanitize_languagecode($languageCode);
+        }
+
+        Yii::app()->setLanguage($baseLanguage);
+
+        if (empty($survey) || !$survey->hasTokensTable) {
+            throw new CHttpException(404, "This survey does not seem to exist. It may have been deleted or the link you were given is outdated or incorrect.");
```
Shnoulle, line 112: Can we register to an notb activated survey ? Or not valid by date survey ?

My opinion : must check active + open 

https://github.com/LimeSurvey/LimeSurvey/blob/14990d854b7ba2298a04965b74264c0fad6377d7/application/models/Survey.php#L457-L458

application/controllers/OptinController.php
```diff
@@ -27,53 +27,183 @@ class OptinController extends LSYii_Controller
...
+        } else {
+            LimeExpressionManager::singleton()->loadTokenInformation($surveyId, $accessToken, false);
+            $token = Token::model($surveyId)->findByAttributes(['token' => $accessToken]);
+
+            if (!isset($token)) {
+                $message = "<p>" . gT('You are not a participant of this survey.') . "</p>";
```
Shnoulle, line 118: tag in PHP : must be theme twig file

Shnoulle: See https://github.com/LimeSurvey/LimeSurvey/blob/master/themes/survey/vanilla/views/subviews/registration/register_message.twig for example

application/controllers/OptinController.php
```diff
@@ -27,53 +27,183 @@ class OptinController extends LSYii_Controller
...
+                    $token->save();
+                    $message = gT('You have been successfully added back to this survey.');
+                } elseif ($token->emailstatus == 'OK') {
+                    $message = gT('You are already a participant of this survey.');
                 } else {
-                    $sMessage = gT('You have been already removed from this survey.');
+                    $message = gT('You have been already removed from this survey.');
+                }
+                // If the $global param is true and 'allowunblacklist' is enabled, remove from the blacklist
+                if ($global && Yii::app()->getConfig('allowunblacklist') == "Y") {
```
Shnoulle, line 194: Yii::app()->getConfig('allowunblacklist') == "Y" ?

We use true/false or 1/0 for all other boolean settings …

Please : use boolean

application/controllers/OptinController.php
```diff
@@ -27,53 +27,183 @@ class OptinController extends LSYii_Controller
...
+                        foreach ($blacklistResult->getMessages() as $blacklistMessage) {
+                            $message .= "<br>" . $blacklistMessage;
+                        }
+                    }
                 }
             }
         }
 
-        $this->renderHtml($sMessage, $oSurvey);
+        $this->renderHtml($message, $survey);
```
Shnoulle, line 206: Same : need a twig file

application/controllers/OptoutController.php
```diff
@@ -68,118 +71,122 @@ public function actiontokens()
...
+            $participant = $blacklistHandler->getCentralParticipantFromToken($oToken);
+
+            if (!empty($participant) && $participant->blacklisted != 'Y') {
+                $message = "<p>" . gT('Please confirm that you want to be removed from the central participants list for this site.') . "</p>";
+                $message .= '<p><a href="' . Yii::app()->createUrl('optout/removetokens', array('surveyid' => $surveyId, 'langcode' => $baseLanguage, 'token' => $accessToken, 'global' => true)) . '" class="btn btn-default btn-lg">' . gT("I confirm") . '</a><p>';
+            } elseif (!$optedOutFromSurvey) {
+                $message = "<p>" . gT('Please confirm that you want to opt out of this survey by clicking the button below.') . '<br>' . gT("After confirmation you won't receive any invitations or reminders for this survey anymore.") . "</p>";
+                $message .= '<p><a href="' . Yii::app()->createUrl('optout/removetokens', array('surveyid' => $surveyId, 'langcode' => $baseLanguage, 'token' => $accessToken)) . '" class="btn btn-default btn-lg">' . gT("I confirm") . '</a><p>';
```
Shnoulle, line 118: twig file

application/controllers/admin/ParticipantsAction.php
```diff
@@ -333,6 +333,9 @@ public function displayParticipants()
             $iTotalRecords = Participant::model()->getParticipantsOwnerCount($iUserId);
         }
         $model = new Participant();
+        if (Yii::app()->getConfig('hideblacklisted') == "Y") {
```
Shnoulle, line 336: boolean

application/controllers/OptinController.php
```diff
@@ -27,53 +27,183 @@ class OptinController extends LSYii_Controller
...
+            $token = Token::model($surveyId)->findByAttributes(['token' => $accessToken]);
+
+            if (!isset($token)) {
+                $message = "<p>" . gT('You are not a participant of this survey.') . "</p>";
+            } else {
+                $optedOutFromSurvey = substr($token->emailstatus, 0, strlen('OptOut')) == 'OptOut';
+
+                $blacklistHandler = new LimeSurvey\Models\Services\ParticipantBlacklistHandler();
+                $participant = $blacklistHandler->getCentralParticipantFromToken($token);
+                $isBlacklisted = !empty($participant) && $participant->blacklisted == 'Y';
```
Shnoulle, line 124: `$participant->blacklisted == 'Y' `? We can not use a boolean ? Or create a `$participant->isBlacklisted()`;

application/datavalueobjects/BlacklistResult.php
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Class BlacklistResult
+ *
+ * This class represents the result of a blacklist operation
+ *
+ * @package LimeSurvey\Datavalueobjects
+ */
+class BlacklistResult
+{
+    /** @var bool the basic result of the operation */
```
Shnoulle, line 14: Great a bool :+1:

application/models/Participant.php
```diff
@@ -24,7 +24,7 @@
  * @property string $lastname
  * @property string $email
  * @property string $language
- * @property integer $blacklisted
+ * @property string $blacklisted
```
Shnoulle, line 27: integer or string but use it at boolean  : 

- 0 == null == "" : not blacklisted. 
- 1 == "Y": blacklisted

application/models/Token.php
```diff
@@ -514,4 +514,16 @@ public static function getDefaultEncryptionOptions()
                 )
         );
     }
+
+    public function onBeforeSave($event)
+    {
+        // Mark token as "OptOut" if globally blacklisted and 'blacklistnewsurveys' is enabled
+        if (Yii::app()->getConfig('blacklistnewsurveys') == "Y" && $this->getIsNewRecord()) {
```
Shnoulle, line 521: blacklistnewsurveys as boolean

Shnoulle (changes requested):
- Usage of boolean for new config
- More twig file for all return message
- blacklisted as boolean seem better

application/controllers/OptinController.php
gabrieljenik: It is done below, we will move it to the bottom.

application/controllers/OptinController.php
gabrieljenik: This was carried over code... but I get your point.
This is saved on the DB as Y/N.
Changing it could be very painfull.

application/controllers/OptinController.php
Shnoulle: But if it's a global optin : need a surveyid ? Even without …

application/controllers/OptinController.php
Shnoulle: Since when its saved as Y/N ? 

In 3.X : it was 0/1
Capture d’écran du 2022-02-08 16-18-25

And it's clearly better

application/models/Participant.php
Shnoulle: See 3.X
Capture d’écran du 2022-02-08 16-20-19

False is save as `0` , and true save as `Y` . Then  we can cast it as bookean 

`boolval("0") === boolval(0) === false` , boolval("Y") === true

No issue 
https://www.php.net/manual/fr/function.boolval.php

application/controllers/OptinController.php
Shnoulle: Else : it can be saved as Y for true, since it was saved as 0 for false : we can use boolval :)

Shnoulle: Argls … show as 0, saved a N ...

**awfull** 😵

I close

application/controllers/admin/ParticipantsAction.php
Shnoulle: Ok bad saved … 😵

application/models/Participant.php
Shnoulle: 😵

application/models/Token.php
Shnoulle: 😵

Shnoulle:
Just more twig then.

gabrieljenik:
> Just more twig then.
What about now?

application/controllers/OptinController.php
```diff
@@ -27,68 +27,199 @@ class OptinController extends LSYii_Controller
...
+                $message = gT('Please confirm that you want to be added back to the central participants list for this site.');
+                $link = Yii::app()->createUrl('optin/addtokens', ['surveyid' => $surveyId, 'langcode' => $baseLanguage, 'token' => $accessToken, 'global' => true]);
+            } elseif ($optedOutFromSurvey) {
+                $message = gT('Please confirm that you want to be added back to this survey by clicking the button below.') . '<br>' . gT("After confirmation you may start receiving invitations and reminders for this survey.");
+                $link = Yii::app()->createUrl('optin/addtokens', ['surveyid' => $surveyId, 'langcode' => $baseLanguage, 'token' => $accessToken]);
+            } elseif (empty($participant)) {
+                $message = gT('You are already a participant of this survey.');
+            } else {
+                $message = gT('You are already part of the central participants list for this site.');
+            }
```
Shnoulle, line 137: Can we send data about situation in view ?

Something like `['participant' => $participant->getAttributes(), 'token' => $token->getAttributes()]` etc …

Else : twig edition don't know anything on the reason  , best is to put gT in twig but take more time.

application/controllers/OptinController.php
gabrieljenik: Not related, but let me tell you something I don't love that much about this twig.
Twig idea is for sperating controller/logic and view.
But in that twig the status is kind of decided on the view and not the controller/logic.
(just as some feedback, not to make a discussion out of it) :)

Shnoulle: It's currently the only solution to allow user to do such things : https://forums.limesurvey.org/forum/can-i-do-this-with-limesurvey/126630-access-code-page#224661 

I really think it's something needed and great for user.

gabrieljenik: > allow user to do such things

You mean "Override messages from theme"?

Shnoulle: For example ,
Or rediect in some condition or use different color or … etc …

Shnoulle: >  But in that twig the status is kind of decided on the view and not the controller/logic.

Showin specific string is view task : not logic task. But create a twig file for each message seems overkiil

gabrieljenik: Redirect doesn't sound like a "view" task.

> Showin specific string is view task

Display message is view task. 
Deciding which message to display is "view" task also?
I understand we have to give in on purity sometimes. Not complaining. Just sending feedback :)

application/controllers/OptinController.php
gabrieljenik: > Can we send data about situation in view ?

Why?

application/controllers/OptinController.php
Shnoulle: >  Deciding which message to display is "view" task also?

A possible «««good»»» way is to load a specific twig for each message type. But create a twig file for each message seems overkiil :)

> Redirect doesn't sound like a "view" task.

Yeag, right, but user like it i think (see the number of forum post about way to redirect Survey list page for example).

>  Deciding which message to display is "view" task also?

No : but have "error type" and choose message according to type ? Error information/detail .

When try to construct real solution by theme : it's really hard to get information. Currently sometimes i use ExpressionManager to get core information (for example {{ processString("{TOKEN}") }}seems really a workaround and not a good way to do.
If we are inside Participant management (for example), seems legit to have Participant information.

application/controllers/OptinController.php
Shnoulle: To allow Survey theme to do different job according to situation. And show own string with own class and action …

application/controllers/OptinController.php
Shnoulle: @gabrieljenik : today we can optin/optout on a inactive or date out survey (closed by date) ? I can check but if you already done.

application/controllers/OptinController.php
Shnoulle: Find a better solution than br in https://github.com/LimeSurvey/LimeSurvey/pull/2202/files#diff-b5c799a27b238495f0849ffe1a47677968dde9b20817563805cb573c1b031fccR131 

br tag is clearly a view … ;)

Alternative solution : participant.attribute and token.attribute : you say it's logic, i say it's between ;).

I'm sure someone ask/need a solution to do something according to «situation» in this page. Have a solution nee to have information ;).

application/controllers/OptinController.php
gabrieljenik: > today we can optin/optout on a inactive or date out survey (closed by date)

Yes, we have looked through the code and think so.

application/controllers/OptinController.php
Shnoulle: Ok, i report the issue tomorrow :).

Close this one

application/controllers/OptinController.php
```diff
@@ -27,68 +27,219 @@ class OptinController extends LSYii_Controller
...
-        if (!$iSurveyID) {
-            $this->redirect(array('/'));
+
+        $surveyId = Yii::app()->request->getQuery('surveyid');
+        $languageCode = Yii::app()->request->getQuery('langcode');
+        $accessToken = Token::sanitizeToken(Yii::app()->request->getQuery('token'));
+
+        //IF there is no survey id, redirect back to the default public page
+        if (!$surveyId) {
+            $this->redirect(['/']);
```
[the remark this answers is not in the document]
gabrieljenik: Current PR code is not failing for us.
We are testing from subfolder.

How did you get the error? Can you provide more detail?

Shnoulle (approved):
Code OK ! Reviewed a lot !

PR #2208, opened 2022-01-13
17719: Question list not updated when adding/changing questions (jest)

thedirtypanda:
Switch into _assests/packages/adminsidepanel/_ and run **yarn** after that, run **yarn test** and the test should be green.


gabrieljenik (commented):
Not sure I am the best to review this.
I don't have a context of the issue nor the solution.

Doesn't seem to be a small fix.
Sounds like a lot of vue stuff is added, which caught my attention as sounds like an enhacement more than a fix. But again, I don't have the context.

PR #2210, opened 2022-01-14
Fixed issue #17798: Set authwebserver as default didn't deactivate AuthDB usage

gabrieljenik:


application/core/plugins/Authwebserver/Authwebserver.php
```diff
@@ -83,9 +83,9 @@ public function beforeLogin()
                 $this->setAuthPlugin(); // This plugin handles authentication, halt further execution of auth plugins
                 return;
             }
-            if ($this->get('is_default', null, null, $this->settings['is_default']['default'])) {
-                throw new CHttpException(401, 'Wrong credentials for LimeSurvey administration.');
-            }
+        }
```
Shnoulle, line 86: This must be checked only if $this->get('serverkey') is not empty.

Shnoulle (changes requested):
The issue is in `isset($_SERVER[$serverKey])` and maybe we need to show an alert to the user if `!isset($_SERVER[$serverKey])` in Plugin settings .

1. Don't disable access : show an alert «The $_SERVER[$serverKey] is not set, plugin are not activated»
2. Disable access : show a bigger alert : «Danger : The $_SERVER[$serverKey] is not set, you can not connect»

And , maybe a config['authwebserver_checkserverkey'] to true by default, not updatable via GUI (or updatable via GUI)

application/core/plugins/Authwebserver/Authwebserver.php
gabrieljenik: That was the issue! It was checked only if the server received the proper information.
If the server didn't the regular login was used, not respecting the "default checkbox".

Now, if no login possible with this plugin and the plugin is the defualt, an exception is thrown.

Shnoulle: Yes but tjis must not happen if `$this->get('serverkey')` is empty : empty not configured :)

Shnoulle:
If admin make an error : we need to show it clearly in plugin config : 

in getPluginSettings function something like this : 

```
if(!isset($_SERVER[$serverKey])) {
$this->settings['serverkey']['help'] = "<p class='alert alert-danger'>" . gT("The current server key are not set in current access, if you set this plugin as default : you can noty log in again") . "<p>";
}
```

application/core/plugins/Authwebserver/Authwebserver.php
gabrieljenik: > Yes but tjis must not happen if $this->get('serverkey') is empty : empty not configured :)

I understand.  But why not?

gabrieljenik:
I would suggest a general config setting on the config files for swhitching on emergnecy to regular login, just in case someone gets locked out. Ex: Set the AithWebserver but the webserver is not properly setup, how do I login again to deactivate the plugin?

Shnoulle (approved):

PR #2176, opened 2021-12-09
Fixed issue #17790: KCFinder Image Browse Server not working due to i…

Shnoulle:
Fixed issue #17790: KCFinder Image Browse Server not working due to incorrect cookieDomain
Dev: this need to set it in config.

PR #2212, opened 2022-01-19
Fixed issue #17790: KCFinder Image Browse Server not working due to incorrect cookieDomain

Shnoulle:
Dev: this need to set it in config.
Dev: but take config to fix KCFINDER cookieDomain
Dev: same fix from master


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2176

PR #2213, opened 2022-01-19
New feature #14782: Possibility to override single translations by database

Shnoulle:
Dev: working solution for user with DB access, no GUI


application/core/LSMessageSource.php
```diff
@@ -0,0 +1,181 @@
...
+class LSMessageSource extends CMessageSource
+{
+    public const CACHE_KEY_PREFIX = 'Yii.LSMessageSource.';
+    public const MO_FILE_EXT = '.mo';
+    public const PO_FILE_EXT = '.po';
+    /**
+     * @var string the ID of the database connection application component. Defaults to 'db'.
+     * @see CDbMessageSource::connectionID
+     */
+    public $connectionID = 'db';
```
Shnoulle, line 21: Maybe disable update by config ?

application/core/LSMessageSource.php
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * @var string the name of the source message table. Defaults to 'SourceMessage'.
+     * The DB prefix is added automatically
+     * @see CDbMessageSource::sourceMessageTable
+     */
+    public $sourceMessageTable = 'sourcemessage';
```
Shnoulle, line 27: Same : LimeSurvey create it

application/core/LSMessageSource.php
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * @var string the name of the translated message table. Defaults to 'Message'.
+     * The DB prefix is added automatically
+     * @see CDbMessageSource::translatedMessageTable
+     */
+    public $translatedMessageTable = 'message';
```
Shnoulle, line 33: Same : LimeSurvey create it

c-schmitz:
I  am very hesitant to have this as a core feature, because I am afraid that this will diminish contributions to translate.limesurvey.org, because why correct a wrong translation (and contribute back) if you can just quickly fix it in your own instance, only.

Shnoulle:
See your comment here : https://bugs.limesurvey.org/view.php?id=17163

>  That would be nice. 

seems i made an error in bug report, i update the title

Shnoulle:
@olleharstedt if it('s OK to be included in core, can you help me on the DB update issue ? I update myself (from 479) : it's OK …

Shnoulle:
> @olleharstedt if it('s OK to be included in core, can you help me on the DB update issue ? I update myself (from 479) : it's OK …

OK, seems i have it : https://github.com/LimeSurvey/LimeSurvey/pull/2213/commits/6a87152239e9ab587a1508eb0d2ca011956bc63a

application/core/LSMessageSource.php
```diff
@@ -0,0 +1,182 @@
...
+     * Defaults to 'cache' which refers to the primary cache application component.
+     * @see CDbMessageSource::cacheID
+     * @see CGettextMessageSource::cacheID
+     */
+    public $cacheID = 'cache';
+    /**
+     * @var string the base path for all translated messages. Defaults to null, meaning
+     * @see CGettextMessageSource::basePath
+     */
+    public $basePath;
```
Shnoulle, line 51: Same than db, table name etc … LimeSurvey offer is own system.

Did we allow to update basePath in config , then user can create their own po file.

Shnoulle:
> I am very hesitant to have this as a core feature, because I am afraid that this will diminish contributions to translate.limesurvey.org, because why correct a wrong translation (and contribute back) if you can just quickly fix it in your own instance, only.

Then , for information :. To have own translation, not translated currently. 

**With DB (this commit)** 

1. Search the english string
2. Create the string via DB in sourcemessage table
3. Check the id
4. Create the translated string in message.


** With po (no commit)**

1. Download po file in https://translate.limesurvey.org/projects/
2. Edit via poedit.
3. Create ownlocale directory, update https://github.com/LimeSurvey/LimeSurvey/blob/fca5dc1256faf9c740eb5b0b7efc1daab428ea89/application/config/internal.php#L210 
4. Put your updated version in this directory.

Shnoulle:
The issue in «Compare upgrade and fresh install» is not related to MYISAM against InnoDB ? Since only InnoDB accept foreignKey.

I can not force foreign key and create model, but sonce we don't use this model …

@olleharstedt an idea ?

olleharstedt:
Hm

installer/create-database.php
```diff
@@ -1121,6 +1121,40 @@ function populateDatabase($oDB)
...
+                'translation' => "text",
+            ],
+            $options
+        );
+        $oDB->createCommand()->addPrimaryKey(
+            '{{message_pk}}',
+            '{{message}}',
+            ['id', 'language']
+        );
+        $oDB->createCommand()->addForeignKey(
```
olleharstedt, line 1148: Normally we try to avoid foreign keys, because we never know how they work across MSSQL, Postgres and Maria/MySQL.

application/helpers/update/updates/Update_480.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->db->createCommand()->addForeignKey(
+            '{{FK_Message_SourceMessage}}',
+            '{{message}}',
+            'id',
+            '{{sourcemessage}}',
+            'id',
+            'CASCADE',
+            'RESTRICT'
+        );
+        $this->db->createCommand()->update('{{settings_global}}', array( 'stg_value' => 480), "stg_name = 'DBVersion'");
```
olleharstedt, line 41: Remove this line, it's automatically done based on class name.

application/helpers/update/updates/Update_480.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace LimeSurvey\Helpers\Update;
+
+class Update_480 extends DatabaseUpdateBase
+{
+    public function up()
+    {
+        $this->db->createCommand()->createTable(
+            '{{sourcemessage}}',
```
olleharstedt, line 10: sourcemessage or source_message?

olleharstedt:
I'd like a unit-test of any kind to test this feature, too.

application/helpers/update/updates/Update_480.php
olleharstedt: Or message_source??

application/core/LSMessageSource.php
```diff
@@ -0,0 +1,182 @@
...
+     * @var string the message catalog name. This is the name of the message file (without extension)
+     * that stores the translated messages. Defaults to 'messages'.
+     * see CGettextMessageSource::catalog
+     */
+    public $catalog = 'messages';
+
+    /**
+     * Used as static for get DB
+     */
+    private $messagedb;
```
olleharstedt, line 73: Could this be moved to static variable in the function instead?

installer/create-database.php
Shnoulle: See : https://www.yiiframework.com/doc/api/1.1/CDbMessageSource 

And yes : no constraint with mariadb/myisam

application/helpers/update/updates/Update_480.php
Shnoulle: Yiii use SourceMessage , then : what you want ;)

application/core/LSMessageSource.php
Shnoulle: See https://github.com/LimeSurvey/LimeSurvey/blob/fca5dc1256faf9c740eb5b0b7efc1daab428ea89/framework/i18n/CDbMessageSource.php#L94 

But yes only used inside this function

Shnoulle: PS : maybe just use current DB : not allowing to update ?

Shnoulle:
> I'd like a unit-test of any kind to test this feature, too. If Carsten is willing to merge it, that is.

DB update allowed ?

I'm think of a system : 

1. Create a random string
2. Check translation : return same string ( gT($string) )
3. Create another random string
4. Add the DB translation for string1 to string2 (without model)
5. Check gT($string) again.
6. Remove the translation

Allowed ?

application/core/LSMessageSource.php
Shnoulle: Removed

application/core/LSMessageSource.php
Shnoulle: message

application/core/LSMessageSource.php
Shnoulle: Leave it : didn't broke potential usage

installer/create-database.php
Shnoulle: Removed : use at your own risk for manual

application/core/LSMessageSource.php
Shnoulle: removs : not allowing update by DB (keep it for a new plugin event ;) )

application/core/LSMessageSource.php
Shnoulle: removed

olleharstedt:
If you mock getDbConnection, you won't need to touch the db at all. I can show you how. Alternatively, you can do anything in the db as long as you restore the previous state. The problem is different tests affecting each other, as we've seen before.

Shnoulle:
function is now this :

```
    public function getDbConnection()
    {
        return App()->getDb();
    }
```

I need to mock->createCommand() and queryAll() right. It's already exist in another test ? More quick to copy/paste ;)

olleharstedt:
> I need to mock->createCommand() and queryAll() right. It's already exist in another test ? More quick to copy/paste ;)

Something like that. But what you need to mock differs depending on which function (unit) you test. Let me check...

olleharstedt:
Hm, maybe it's easier to mock getMessagesFromDb() instead.

olleharstedt:
Mock docs for PHPUnit: https://phpunit.readthedocs.io/en/9.5/test-doubles.html?highlight=mock

I don't think we're using version 9.5 tho, you'd have to double check.

application/helpers/update/updates/Update_480.php
olleharstedt: Then probable source_message

Sadly we're not consistent in our db naming scheme. But it's supposed to be snake_case.

olleharstedt:
> Then : no need mockup ;) if it work with DB , why need mockup ?

Well, at some point I'd like to gather all tests that DO NOT use file or db and run them in parallel. That's only possible for tests that are fully mocked.

olleharstedt:
Btw, I'll release 5.3 next sprint, so this feature will probably be in 5.4.

Shnoulle:
> 

No problem, i think it's a really good feature, but not need it currently.

PS : maybe i create a free like a speech but not free like a bear plugin for translation management (when LimeStore is ready).

application/core/LSMessageSource.php
```diff
@@ -0,0 +1,149 @@
...
+                $file = new CGettextPoFile();
+            }
+            $messagesGettext = $file->load($messageFile, $category);
+        } else {
+            $messagesGettext = array();
+        }
+
+        /* Messages by DB */
+        if (App()->getConfig('DBVersion') >= 480) {
+            $messagesDb = $this->loadMessagesFromDb($category, $language);
```
olleharstedt, line 96: Wait, does this mean the there's a db hit for every translation string...?

olleharstedt: Or this only happens once, at app start?

Shnoulle: Happens once at each call if debug>0, like reading po/mo

Else : file caching by default.

olleharstedt: :star:

Shnoulle:
> Hm, maybe it's easier to mock getMessagesFromDb() instead.
> 
> And then test getMessagesFromDb() with real db data (which is deleted after test is run, so called fixture).

My opinion : we must check gT function, then can not really mock : best seems to use db update.

Else Yii:t, but then need to mock Yii …

olleharstedt:
You can assume old code is working and test only new code added.

Shnoulle:
Then : i can just check `loadMessages('','noexit' . $uniqid')` return the array returned by mocked `loadMessagesFromDb`  ?

It's not really the feature ;)

olleharstedt:
Remember, you're not writing test to only make sure it works NOW, you write tests to make it KEEPS working in the future when other devs are changing your code.

Shnoulle:
> 

Then : since we use gT for translation : must test gT whole system …

olleharstedt:
> Then : since we use gT for translation : must test gT whole system …

That would be the so called integrity test. Testing new functions separately is unit testing. Feel free to do both, if you have time. :)

Shnoulle:
About unit test : here i really don't see any interest to test LSMessageSource only. 

I can replace loadMessagesFromDb return a fixed array and test if loadMessages('','invalid') return this array.
But i don't see any interest here.

olleharstedt:
Hm, some discussion on it here: https://stackoverflow.com/questions/249664/best-practices-to-test-protected-methods-with-phpunit

Shnoulle:
? not related … in my opinion …

LSMessageSource depend only from App() and update the way Yii::t work.

olleharstedt:
loadMessages() is a protected method, that's why.

Shnoulle:
> I can replace loadMessagesFromDb return a fixed array and test if loadMessages('','invalid') return this array.
> But i don't see any interest here.

olleharstedt:
> Yes , but used ONLY by LSMessageSource …

Yes yes, but that's the new code that should be checked. Maybe I can write something to show you...

tests/functional/acceptance/14782-override-translation/LSMessageSourceTest.php
```diff
@@ -0,0 +1,54 @@
...
+        ));
+        $result = App()->db->createCommand()
+            ->select("id")
+            ->from("{{source_message}}")
+            ->where(
+                'message=:message',
+                array(':message' => $sourcemessage)
+            )->queryRow();
+        $this->assertCount(1, $result, 'Unable to create source_message row in DB');
+        return;
```
olleharstedt, line 32: :O

olleharstedt:
```php
public function testLsMessageSource()
{
    $messageSource = new LSMessageSource();
    $messageSource->basePath = YII_PATH . '/../locale/';
    $ref = new ReflectionClass('LSMessageSource');
    $met = $ref->getMethod('loadMessages');
    $met->setAccessible(true);
    $messages = $met->invokeArgs($messageSource, ['', 'en']);
    $this->assertNotEmpty($messages);
}
```

One example of calling loadMessages (even tho it's protected) and test the result.

Shnoulle:
> One example of calling loadMessages (even tho it's protected) and test the result.

Interest ? Testing a 2 functions via phpunit …

I can add if you want , but didn't see the interest.

The current test check gT function.

olleharstedt:
Well, optimally, you're supposed to test each possible path in the function - with cache, without cache, with db, without db, etc. That will be hard only using gT().

olleharstedt:
Hm, on another note, I'm not able to repeat your test multiple times locally. It works once, but then I get this:

```
1) ls\tests\LSMessageSourceTest::testUpdatedString
Message translation by DB is broken
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'translatedmessageP8YlHKPocMV98wFEZeolVgWhTYm5FbadrKkzYM~7UX'
+'sourcemessageCgfg5jHVZ5MhS_F_pQ4D7LFi~QRv7SY0s9gL9dqjmn'
```

Related to random string?

Shnoulle:
> Well, optimally, you're supposed to test each possible path in the function - with cache, without cache, with db, without db, etc. That will be hard only using gT(). Just using gT(), you might not get 100% or 50% coverage.

And with your test : what is the coverage ? What do you test ? 



I can not create test for **all** previous feature each time : it's not part of the coverage… i have to test only the new feature.

What is the coverage here : «Possibility to override single translation by db» OK to add caching (translation not updated with DB update only), but don't understand witout db …

Shnoulle:
> 



> Hm, on another note, I'm not able to repeat your test multiple times locally. It works once, but then I get this:
> 
> ```
> 1) ls\tests\LSMessageSourceTest::testUpdatedString
> Message translation by DB is broken
> Failed asserting that two strings are equal.
> --- Expected
> +++ Actual
> @@ @@
> -'translatedmessageP8YlHKPocMV98wFEZeolVgWhTYm5FbadrKkzYM~7UX'
> +'sourcemessageCgfg5jHVZ5MhS_F_pQ4D7LFi~QRv7SY0s9gL9dqjmn'
> ```
> 
> Related to random string?

Related to caching … cache for 1 hour by default.

olleharstedt:
> Related to caching … cache for 1 hour by default.

OK, so I need to disable caching? You see one benefit at once with my approach: just set `$messageSource->cachingDuration = 0` in the test code. :)

Shnoulle:
> > Related to caching … cache for 1 hour by default.
> 
> OK, so I need to disable caching? You see one benefit at once with my approach: just set `$messageSource->cachingDuration = 0` in the test code. :)

I already think we need to check if caching work .

Best seems to 
1. Reset caching at start
2. Create the new translation in DB
3. Call existing po DB string : (Submit ? Yes ? anything can be updated at a time
4. Call a random string : must return new translation
5. Update the random string in DB : must retuen previous one

The lack is update core string , then if i can add a translation for _Submit_ : it's OK.

But in this condition : if user have updated his own language : didn't work   …

I create all of this

olleharstedt:
> I already think we need to check if caching work .

Maaaybe. Caching is part of Yii, and we should assume it's working (Yii has its own test suite).

Shnoulle:
> Well, optimally, you're supposed to test each possible path in the function - with cache, without cache, with db, without db, etc. That will be hard only using gT(). Just using gT(), you might not get 100% or 50% coverage.
> 
> You're also supposed to do _negative_ tests - testing that the function _fails_ correctly.



> > I already think we need to check if caching work .
> 
> Maaaybe. Caching is part of Yii, and we should assume it's working (Yii has its own test suite).

????

olleharstedt:
Oh, I'm contradicting myself. :facepalm:

Shnoulle:
> we should assume

Like : we should assume po/mo file work (or not ?) 

The «we should assume» the default way to get translation use LSMessageSource can not be assumed.

gT whole system can be updated too easily …

olleharstedt:
> Like : we should assume po/mo file work (or not ?)

I think, assuming po/mo files are correct, is the same as mocking the access to mo/po files in the unit test. Which is one way to do it. But it's OK to just access those files, too.

Shnoulle:
> > Like : we should assume po/mo file work (or not ?)
> 
> I think, assuming po/mo files are correct, is the same as mocking the access to mo/po files in the unit test. Which is one way to do it. But it's OK to just access those files, too.

Again : mocking a 2 function call in same class … i don't really understand the point here …

I don't understand what you want …

Shnoulle:
> One example of calling loadMessages (even tho it's protected) and test the result.

Any idea to mock loadMessagesFromDb here ? Because without this : i didn't test this feature …

Shnoulle:
Check if po translation work + check if file cache (by default) work + DB new source + DB replace source.

I don't see what i can do now …

olleharstedt:
Great! Is it now possible to run the test multiple times locally?

Shnoulle:
> Great! Is it now possible to run the test multiple times locally?

I add a resetCache function.
Default behaviour is to reset DB cache only : https://github.com/LimeSurvey/LimeSurvey/blob/30d8f21316cafc0b06a2e8198f682d4feea132d8/tests/TestBaseClass.php#L40 

I add  https://github.com/LimeSurvey/LimeSurvey/pull/2213/files#diff-3cc131b4d608fd992d775790fbc8c5f5b04a49347a85a54390af3291c6b504e0R44
(PS : maybe we need an API function for this)

olleharstedt:
OK, gonna test locally again...

Shnoulle:
> LSMessage Source (ls\tests\LSMessageSource)
>  ✔ Updated string

olleharstedt:
Works multiple times for me locally :+1: :star2:

olleharstedt (approved):

PR #2218, opened 2022-01-25
New feature #8239: Allow for user friendly survey URLS

gabrieljenik:
- [x] Set an alias to a survey [Must Have]
- [x] Run the survey using the alias [Must Have]
- [x] Uniqueness Validation [Must Have]
- [x] Add short url display in overview [Must Have]
- [x] Automatic tests [Must Have]
- [x] Test: Does it work with tokens? [Must Have]
- [x] Test: Does it work with parameter? [Must Have]
- [x] Replacement variable for writing mails. [Could Have]


gabrieljenik:
This is a WIP so you can throw your thoughts in.

gabrieljenik:
@olleharstedt Is this ok?
https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/helpers/update/updatedb_helper.php#L90

This is causing new updates on DB to fail on the automated tests.

olleharstedt:
Hm, let me see.

olleharstedt:
Nope, not OK. :) Check the application/helpers/update/updates/ folder and add a new class called Update_480.php. Check other classes in that folder for inspiration.

olleharstedt:
Each update is a separate class now.

gabrieljenik:
I did that.
https://github.com/LimeSurvey/LimeSurvey/pull/2218/files#diff-d79139a77f31dedfa48f77ec35037910ec097711865fc4d6ed74ee04487c6cde

But automated test fails as this:
https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/helpers/update/updatedb_helper.php#L107

Feel the code I

olleharstedt:
Ah, maybe a faulty merge then. Let me clean that up for you.

olleharstedt:
OK, pull dev into your branch again and see.

Shnoulle:
> @olleharstedt Is this ok? https://github.com/LimeSurvey/LimeSurvey/blob/develop/application/helpers/update/updatedb_helper.php#L90
> 
> This is causing new updates on DB to fail on the automated tests.

See https://github.com/LimeSurvey/LimeSurvey/pull/2213/commits/6a87152239e9ab587a1508eb0d2ca011956bc63a

Merge issue from master to develop surely.

Shnoulle:
About the feature : i'm personnaly for a core plugin here …

application/core/LSYii_Controller.php
```diff
@@ -39,7 +39,7 @@ public function __construct($id, $module = null)
         // Deprecated function
         $this->loadHelper('globalsettings');
         // tracevar function
-        $this->loadHelper('common');
+        //$this->loadHelper('common');
```
Shnoulle, line 42: ?

application/views/admin/survey/editLocalSettings_view.php
```diff
@@ -37,6 +37,27 @@
...
+                        <?php eT("Survey alias:"); ?>
+                    </label>
+                    <div class="">
+                        <?php echo CHtml::textField(
+                                "alias_{$aSurveyLanguageSettings['surveyls_language']}",
+                                $aSurveyLanguageSettings['surveyls_alias'],
+                                [
+                                    'class' => 'form-control',
+                                    'size' => "80",
+                                    'maxlength' => 200,
```
Shnoulle, line 53: 100

gabrieljenik:
> About the feature : i'm personnaly for a core plugin here …

Core plugins don't allow new fields on the edit survey screen.
We would need to add the "alias" as a plugin setting.

Also, the url wouldn't be as short as we are thinking as it would require a valid controller and action.
(or using a dynamic rule)

olleharstedt:
Time to pull upstream?

gabrieljenik:
This is ready for testing! :)

Shnoulle:
> Core plugins don't allow new fields on the edit survey screen.

I can be in Plugin settings and/or Tools menu. Else : this answer is more for LimeSurcvey core «Allow plugin to updfate any form in Survey Screen» ;)

> We would need to add the "alias" as a plugin setting.

I don't understand ?

gabrieljenik:
> > Core plugins don't allow new fields on the edit survey screen.
> 
> It can be in Plugin settings and/or Tools menu. Else : this answer is more for LimeSurcvey core «Allow plugin to updfate any form in Survey Screen» ;)

Well, but that's not available now or wasn't available at the time this started.
Also, besides having those event types, there is the other discussion, why on plugins and not in core?

> > We would need to add the "alias" as a plugin setting.
> 
> I don't understand ?

I was argumenting why we need this as a core enhancement as opposed to creating a plugin.
If we do it as a plugin, we can't add settings on user friendly places.

olleharstedt:
Hey Gabriel. Can you resolve the conflicts, please?

gabrieljenik:
> Hey Gabriel. Can you resolve the conflicts, please? Then we'll try to get this merged...

Done

olleharstedt:
Neat! Let's see if we can review and test this internally, so it gets merged finally!

application/core/LSYii_Application.php
```diff
@@ -102,6 +102,9 @@ public function __construct($aApplicationConfig = null)
         if (!isset($aApplicationConfig['components']['assetManager']['basePath'])) {
             App()->getAssetManager()->setBasePath($this->config['tempdir'] . '/assets');
         }
+
+        // Load common helper
+        $this->loadHelper("common");
```
olleharstedt, line 107: Probably a good idea, but it would mean we can delete lots of other common_helper includes, right?

application/core/LSYii_Application.php
```diff
@@ -508,6 +511,57 @@ public function getTwigCustomExtensionsConfig($sUsertwigextensionrootdir, $aAppl
...
+
+        return $controller;
+    }
+
+    /**
+     * Create controller from short url if the route matches a survey alias.
+     * @param string $route the route of the request.
+     * @return array<mixed>|null
+     */
+    private function createControllerFromShortUrl($route)
```
olleharstedt, line 536: Do we have a unit-test for this function?

application/core/LSYii_Application.php
```diff
@@ -508,6 +511,57 @@ public function getTwigCustomExtensionsConfig($sUsertwigextensionrootdir, $aAppl
...
+    {
+        $controller = parent::createController($route, $owner);
+
+        // If no controller is found by standard ways, check if the route matches
+        // an existing survey's alias.
+        if (is_null($controller)) {
+            $controller = $this->createControllerFromShortUrl($route);
+        }
+
+        return $controller;
```
olleharstedt, line 528: Fun fact, $controller can still be null here. :)

application/models/Survey.php
```diff
@@ -2286,6 +2289,56 @@ function ($s) {
...
 
+    /**
+     * Returns the survey URL with the specified params.
+     * If $preferShortUrl is true (default), and an alias is available, it returns the short
+     * version of the URL.
+     * @param string|null $language
+     * @param array<string,mixed> $params   Optional parameters to include in the URL.
+     * @param bool $preferShortUrl  If true, tries to return the short URL instead of the traditional one.
+     */
+    public function getSurveyUrl($language = null, $params = [], $preferShortUrl = true)
```
olleharstedt, line 2300: Another candidate for a unit-test, perhaps?

application/views/admin/survey/editLocalSettings_view.php
olleharstedt: Oh yeah, this one, @gabrieljenik

tests/functional/frontend/ShortUrlTest.php
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * @group shorturl
+ */
+class ShortUrlTest extends TestBaseClassWeb
+{
+    /**
+     * Test short urls without extra params (except language)
+     * @dataProvider shortUrlDataProvider
+     */
+    public function testShortUrlOnOpenSurvey($alias, $params, $welcomeText, $firstQuestionText)
```
olleharstedt, line 16: Nice, maybe these tests cover the functions I've commented on above.
All good!

olleharstedt (approved):

olleharstedt:
Only thing is the 200 -> 100 comment by Denis that needs fixing.

Shnoulle:
What about 

- /lilurlfr : goes to survey in fr
- /lilurlfr?lang=en : goes to survey in en

I think it's better.

Shnoulle (commented):
See 

- /lilurlfr : goes to survey in fr
- /lilurlfr?lang=en : goes to survey in en

And need help too

application/core/LSYii_Application.php
gabrieljenik: Hmm.. Honestly, don't recall. This was done long ago.
Let me pick up.

gabrieljenik: It is worth trying... I would just set it as separated DEV task.
We would need to load the helper as well in `application/core/ConsoleApplication.php`.

Maybe we can create a helperLoader class to centralize the loading depending on the application type

application/core/LSYii_Application.php
gabrieljenik: Not a unit test, but a functional test that uses it.
We can do a Unit test if needed

application/models/Survey.php
gabrieljenik: Yes, also called from the functional test

application/core/LSYii_Application.php
```diff
@@ -515,4 +518,52 @@ public function getTwigCustomExtensionsConfig($sUsertwigextensionrootdir, $aAppl
...
+            ])->findByAttributes(['surveyls_alias' => $alias]);
+            if (!empty($languageSettings)) {
+                // Override language
+                if (isset($_GET['lang'])) {
+                    $_GET['lang'] = $languageSettings->surveyls_language;
+                }
+                if (isset($_POST['lang'])) {
+                    $_POST['lang'] = $languageSettings->surveyls_language;
+                }
+                return $this->createController("survey/index/sid/" . $languageSettings->surveyls_survey_id);
```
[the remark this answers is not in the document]
gabrieljenik: > I'm more for `$this->createController`

Please take a look at the signature of the method.
This is not a URL creation method, but a controller creation.
Params are not received.

Still, we will review it as we think we should call parent::createController, mainly to avoid recursion.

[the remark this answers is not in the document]
gabrieljenik: > /lilurlfr : goes to survey in fr
> but
> /lilurlfr?lang=en : goes to survey in en

We discussed and thought about this (I admit, in the past).
- Each language can handle a different alias
- Still, they can use the same alias
- When looking for a survey from an alias, if there is a language in the querystring, we use it as criteria to find the survey and the 
- So, your prior case would work if both languages share the same alias.
- If EN doesn't have `lilurlfr` as an alias, survey will not be found.

An enhancement we can do is:
- When searching a survey by alias, don't filter by language, any language settning will be OK.
- At last, we just wan't to know the survey id.

application/views/admin/survey/editLocalSettings_view.php
```diff
@@ -37,6 +37,27 @@
...
+                                    'id' => "alias_{$aSurveyLanguageSettings['surveyls_language']}",
+                                    //'pattern' => '[\w\d-]+'
+                                ]
+                            ); ?>
+                    </div>
```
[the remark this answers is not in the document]
Shnoulle: Yes : it shown, but user need to understand why it's `/index.php?r=aliasurl` and not `/aliasurl` 

I'm sur there are a lot of complaint in forum.

But let it empty for now.

Shnoulle (approved):

application/views/admin/survey/editLocalSettings_view.php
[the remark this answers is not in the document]
gabrieljenik: It should be filed as a different ticket... If not we will never wrap this up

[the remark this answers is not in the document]
Shnoulle: Yes, i resolve the conversation :)

gabrieljenik:
@olleharstedt Ready for re-reviewing

c-schmitz:
@gabrieljenik can you update the PR please? Since it is coming from your own repository I cannot do that.

gabrieljenik:
> Please use source branches in our repository for future patches.

Ok, you mean for developments, right? or for all?

c-schmitz:
This is inherently unsafe. 

I can enter existing 'survey' and 'index.php' as short names , which miserabley fail.
I can also enter "<script>alert(1)</script>" by which I can XSS the administration.

You need to establish proper security filtering and sanitizing here. Also you probably might want to check the alias against existing controller names to avoid clashes.

c-schmitz:
@Shnoulle @olleharstedt The XSS should have been spotted by code review, tbh. Please pay extra attention to these things.

olleharstedt:
> `\w\d` allow `<` and `>` ? I really think not … more care next time … regexp is awfull …

I'm confused... The original rule was `[\w\d]+` and it still allowed `<script>` to be entered? That's not right...

Shnoulle:
> > `\w\d` allow `<` and `>` ? I really think not … more care next time … regexp is awfull …
> 
> I'm confused... The original rule was `[\w\d]+` and it still allowed `<script>` to be entered? That's not right...
> 
> https://regex101.com/r/HAkiJv/1

I test : `<script>alert('XSS');</script>` can be entered and saved …

It must allow é and other ? Since it's OK now in url.

gabrieljenik:
> > `\w\d` allow `<` and `>` ? I really think not … more care next time … regexp is awfull …
> 
> I'm confused... The original rule was `[\w\d]+` and it still allowed `<script>` to be entered? That's not right...
> 
> https://regex101.com/r/HAkiJv/1

The rule allowed partial matches, as it didn't enforce to match the whole string (no start of line or end of line symbol in the pattern). Now it has to start with a letter.

Also has enhanced the validation, validating against current rules and controllers.

olleharstedt:
> The rule allowed partial matches, as it didn't enforce to match the whole string (no start of line or end of line symbol in the pattern).

Ah ok. Bit weird behaviour from Yii, to be honest.

Shnoulle:
> Still, after reviewing the comments would like to review some more cases a bit more closely.

What cases ? You mean accent ? http://example.org/m%C3%A9dical for example ?

application/helpers/update/updates/Update_493.php
```diff
@@ -9,6 +9,7 @@ class Update_493 extends DatabaseUpdateBase
      */
     public function up()
     {
+        $this->db->createCommand()->addColumn('{{surveys_languagesettings}}', 'surveyls_alias', 'string(100)');
```
Shnoulle, line 12: We can not create a 494 for this update ?

I'm really for the minimum update between each number :)

Shnoulle:
You need to disable some specific word too : `assets` , `tmp` etc … ?

Maybe adding [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) to input too ?

Shnoulle (approved):

application/views/admin/survey/editLocalSettings_view.php
Shnoulle: Still this one ;)

Shnoulle:
Thank you for 494 :)

gabrieljenik:
I think it is ready to be re-reviewed.

c-schmitz:
When activating tokens and trying to run a survey using the alias I get grafik

c-schmitz (changes requested):
See comments

c-schmitz:
In general it seems to work.

I still think a warning message is needed if an existing survey alias is changed (so if you edit the exising value in the input box).
"Warning: If you change your survey alias then any links/invitations using the old alias will not work, anymore."
should then be shown below the edit box.

Shnoulle:
> Will work on the warning. For the email variables, is already done. It is called: SURVEYSHORTURL

Carsten want : SURVEYURL use short URL and have a new codename for «old url with survey id»

application/core/LSYii_ShortUrlValidator.php
```diff
@@ -0,0 +1,152 @@
...
+            $className = ucfirst($id) . 'Controller';
+            $classFile = $basePath . DIRECTORY_SEPARATOR . $className . '.php';
+
+            if (isset($owner->controllerNamespace)) {
+                $className = $owner->controllerNamespace . '\\' . str_replace('/', '\\', $controllerID) . $className;
+            }
+
+            if (is_file($classFile)) {
+                if (!class_exists($className, false)) {
+                    require($classFile);
```
c-schmitz, line 102: This should be require_once.

c-schmitz:
> Carsten want : SURVEYURL use short URL and have a new codename for «old url with survey id»

Yes, exactly. If you already have some placeholder for this I think it is easy to revise it for this case.

c-schmitz:
> SURVEYIDURL 

I like this one ;)

gabrieljenik:
Question is: What happens to old surveys with no alias? What if someone want's the regular old URL?

I think it could be good to have:
- SURVEYURL: Alias -> Sid Based
- SURVEYSHORTURL
- SURVEYLONGURL

c-schmitz:
> What happens to old surveys with no alias? What if someone want's the regular old URL?
Nothing happens - they get the old URL.

If no survey alias is set, then the old SID URL will be used, anyway.

I think that,if set, by default the alias URL should be used, when using SURVEYURL. It does not make sense otherwise.

Because, what would you show for 'SURVEYALIASURL' is no alias URL is set?

So SURVEYURL and SURVEYIDURL is enough.

gabrieljenik:
> So SURVEYURL and SURVEYIDURL is enough.

I just feel we may get to a situation where SURVEYALIASURL maybe needed.
If no survey alias is set, SURVEYALIASURL would be blank.

But I will not debate much :) and implement SURVEYURL, SURVEYIDURL

Shnoulle:
> If no survey alias is set, SURVEYALIASURL would be blank.

I like this idea

c-schmitz:
> But I will not debate much :) and implement SURVEYURL, SURVEYIDURL

Please do ;)

gabrieljenik:
> > But I will not debate much :) and implement SURVEYURL, SURVEYIDURL
> 
> Please do ;)

I think it is better to implement `SURVEYALIASURL` now, when we are focused on it, having it just in case, than having to implement it later.

application/models/Survey.php
```diff
@@ -2284,6 +2287,74 @@ function ($s) {
         return $aSurveys;
     }
 
+    /**
+     * Returns the survey URL with the specified params.
```
Shnoulle, line 2291: Returns the **absolute** survey URL with the specified params.

Shnoulle (approved):

gabrieljenik:
Implemented
- SURVEYURL (Alias based first -> ID based if not) 
- SURVEYIDURL (id based, always)

c-schmitz:
Great work, thank you!

PR #2223, opened 2022-01-31
Fixed issue #17833: Multiple choice: 'Text input box size' has no effect.

gabrieljenik:
As said in the bug report, we have removed the attribute.
> If this is not referring to the "other" text box, then it shouldn't be shown for this question type.


Shnoulle (approved):

Shnoulle:
3LTS have the same issue 

:+1:

PR #2224, opened 2022-01-31
Fixed issue #17586: Opening new browser tabs using question index lea…

gabrieljenik:


Shnoulle:
Since it a copy of https://gitlab.com/SondagesPro/SurveyThemes/skelvanilla/-/commit/29e0cd92fdc6147ee7bdf4baeb634d280279f038

Without the role button.
What can i say ?

Shnoulle (approved):

fibl-smart:
hi everyone 
Was this fix pushed only into Version 3?
Versions 5 and 6 seem to be still affected by this issue. Thanks

c-schmitz:
Feel free to create a bug report at https://bugs.limesurvey.org with a detailed description on how to reproduce it.

fibl-smart:
Thanks for the quick reply. 
It is however the same issue still appearing version 5 and 6, with the same description: https://bugs.limesurvey.org/view.php?id=17586 
I was wondering, if / why this was only pushed in version 3... Do we need a bug report for this?

fibl-smart:
Any answer to the question above, would be very appreciated. Thanks!

PR #2225, opened 2022-02-01
Various file upload issues

c-schmitz:
Fixed issue #17718: In file upload question type any file with non-alphanumeric multi-byte characters may not properly show
Fixed issue: Title & comment not shown for file upload question when navigating back and forth
Fixed issue: [security] Possible XSS in file upload question type
Fixed issue: File upload question type popup not showing correct title
Fixed issue: File upload question type preview not working


application/controllers/UploaderController.php
```diff
@@ -274,7 +272,7 @@ public function run($actionID)
                     $return = array(
                         "success" => true,
                         "size"    => $size,
-                        "name"    => rawurlencode(basename($filename)),
+                        "name"    => basename($filename),
```
Shnoulle, line 275: basename still needed ? Since it's sanitised before ?

olleharstedt:
Code looks good otherwise. Do we have instructions to test?

c-schmitz:
It is a mashup of a total of 6 issues that were fixed - 2 of them were reported by a pen test from a user. The rest I fixed while fixing the other two. 
[survey_archive_117565.lsa.zip](https://github.com/LimeSurvey/LimeSurvey/files/7979055/survey_archive_117565.lsa.zip)

Testing is easy. Take a file with Chinese characters only *.png file and try to upload it in the attached survey (rename to .lsa for import).
Navigate back and forth to see what happens in the current master.

- Filename will be cut off/empty / Weird conversion of filenames
- preview will break
- XSS input is possible in description/title etc.

Shnoulle:
PS : 汉语考试服务网.png donwload work on admin part ? When checking related issue : https://bugs.limesurvey.org/view.php?id=14091 even with valid filename in uploader : this part remove the started character.

Since it's a rewrite of sanitize_filename , maybe w can check here too ?

application/controllers/UploaderController.php
c-schmitz: I also think it can be removed, especially considering the  locale problems (which I unfortunately could not reproduce in Windows).
Will do.

c-schmitz:
> > XSS input is possible in description/title etc.
> 
> We don't have already fixed this part before ?

Not this part.

Shnoulle:
> > > XSS input is possible in description/title etc.
> > 
> > 
> > We don't have already fixed this part before ?
> 
> Not this part.

3.X fix needed ?

c-schmitz:
> PS : 汉语考试服务网.png donwload work on admin part ? When checking related issue : https://bugs.limesurvey.org/view.php?id=14091 even with valid filename in uploader : this part remove the started character.
> 
> Since it's a rewrite of sanitize_filename , maybe w can check here too ?

I check admin part only quickly. Single file download was broken there in general.

c-schmitz:
> When checking related issue : https://bugs.limesurvey.org/view.php?id=14091 even with valid filename in uploader : this part remove the started character.

As said, I cannot reproduce it., It is also most likely fixed with this patch.

c-schmitz:
> 3.X fix needed ?

I checked and the security issues do not exist there.

c-schmitz:
> I checked admin part only quickly which worked. Single file download was broken there in general. I see it as a separate issue.

I fixed that  issue, too (already committed because it was a small one-line fix)

c-schmitz:
@glimz Can you please test this today? Thank you.

PR #2226, opened 2022-02-02
Fixed issue #17321: Array question are not split up on mobile devices when using vanilla theme

gabrieljenik:


Shnoulle:
It's a big issue the system between database and XML file … 
No real update can be done or managed using only xml …

gabrieljenik:
About test failed, please make sure the DB is OK as the solution for old installations is based on a DB update.
For new installations and DB, everything should run smoothly.

glimz:
@gabrieljenik Can you update the DB version so I can merge the task? Thanks

Shnoulle:
This broke if someone update advanced settings to (for example) 
`{"add":["css/base.css","css/theme.css","css/custom.css","css/noTablesOnMobile.css","css/hacked.css"]}`

Maybe where must be limited to `{"add":["css/ajaxify.css","css/theme.css","css/custom.css"]}` `{"add":["css/base.css","css/theme.css","css/custom.css"]}` 


And theme.css and custom.css must come after noTablesOnMobile.css .

application/helpers/update/updatedb_helper.php
```diff
@@ -5050,6 +5050,23 @@ function ($v) {
...
+         */
+        if ($iOldDBVersion < 480) {
+            $oTransaction = $oDB->beginTransaction();
+            // Update vanilla config
+            $oDB->createCommand()->update(
+                '{{template_configuration}}',
+                [
+                    'files_css' => '{"add":["css/base.css","css/theme.css","css/custom.css","css/noTablesOnMobile.css"]}',
+                ],
+                "template_name = 'vanilla' AND files_css != 'inherit'"
```
Shnoulle, line 5065: `files_css like '{"add":["css/ajaxify.css","css/theme.css","css/custom.css"]}'` ?

application/helpers/update/updatedb_helper.php
```diff
@@ -5050,6 +5050,23 @@ function ($v) {
...
+         * Add missing noTablesOnMobile.css to vanilla configs again. It was done on 428, but the bug on LsDefaultDataSets remained
+         * causing problems on new installations.
+         */
+        if ($iOldDBVersion < 480) {
+            $oTransaction = $oDB->beginTransaction();
+            // Update vanilla config
+            $oDB->createCommand()->update(
+                '{{template_configuration}}',
+                [
+                    'files_css' => '{"add":["css/base.css","css/theme.css","css/custom.css","css/noTablesOnMobile.css"]}',
```
Shnoulle, line 5063: revoiew order : base,noTablesOnMobile,theme,custom ?

application/helpers/update/updatedb_helper.php
Shnoulle: bootswatch and fruity don't have the same issue ?

They have the `css/ajaxify.css`

Shnoulle (changes requested):
This broke template API and can broke some system.

Hard to do better … but maybe some more restriction and a API version update

application/helpers/update/updatedb_helper.php
Shnoulle: > `files_css like '{"add":["css/ajaxify.css","css/theme.css","css/custom.css"]}'` ?

Alt : loop for all css, json_decode, add noTablesOnMobile if not inside, json_encode.

Shnoulle:
Oh;, i didn't see this new option `:default_files_css`

Else : seems OK , maybe more restriction than needed : but if user update this part : he see the warning ;)

Shnoulle:
@glimz i add again the _Need testing_ : i think the new version need a new test ;)

gabrieljenik:
1) I corrected the new value in the updatedb_helper to match the one in LsDefaultDataSets. That solves the issue of order that Denis mentioned. The order is the same as the files in the XML.

2) On the updatedb_helper, I modified the conditions so that it only applies to records that have the files_css as the defaults, which we changed in LsDefaultDataSets. That is, only records that have the value as it was in LsDefaultDataSets before this PR will be updated.

3) On the other hand, I did a merge with master, so now the updatedb_helper change passed to a single update.

gabrieljenik:
Will update the conflict after it is OK to merge. (as to not have to fix conflict many times)

glimz:
Tested Ok, @gabrieljenik please fix merge conflict and merge

gabrieljenik:
> Tested Ok, @gabrieljenik please fix merge conflict and merge

Merged

PR #2227, opened 2022-02-04
Fixed issue #17869: User list in Survey group permission management is unclear

Shnoulle:
Dev: add full_name in parenthesis
Dev: create a getUserDescription to be used elsewhere

### DIFF
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -961,4 +961,16 @@ class User extends LSActiveRecord
 
         return $this->save();
     }
+
+    /**
+     * Get the decription to be used in list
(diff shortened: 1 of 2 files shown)


gabrieljenik (approved):
Code OK.
Not Tested.

I usually use method name getDisplayName.

`getUserDescription` doesn't sound like a good choice to me.
The word `User` is needed?
Is the word `Description` a good description for method?

Shnoulle:
> I usually use method name getDisplayName.

:+1: seems better.

Let me fix.

PR #2230, opened 2022-02-05
exclude build/devel/test files from export to zip/tarball

computersalat:
There are files exported to zip/tarball that are of no use for production. Hence they should not be exported.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,21 @@
+# exclude build/devel/test files from export to zip/tarball
+.bowerrc                export-ignore
+buildVueComponents.js   export-ignore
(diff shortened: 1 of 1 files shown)


Shnoulle:
Like it ! :+1:

c-schmitz:
Thank you!

Shnoulle:
https://manual.limesurvey.org/index.php?title=Regression_and_unit_tests&type=revision&diff=172046&oldid=169142 :)

computersalat:
@c-schmitz 
I guess you wanted to add `psalm.xml` and not `palm.xml` as you did ;-)

PR #2231, opened 2022-02-07
Zoho/l41 t1659 question administration controller check action params

Trischi80:
Dev:  check action params (GET and POST), to avoid something like gid='someText'


thedirtypanda (approved):

Shnoulle:
If someone try to hack my instance sending 123<script>alert("XSS")</script> : i want to have a log at minimum …

See https://github.com/LimeSurvey/LimeSurvey/pull/2232#issuecomment-1031309986 

PR #2232, opened 2022-02-07
dev: checked incoming params for actions in SurveyAdministrationContr…

Trischi80:
Dev:  check GET and POST params

### DIFF
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -209,7 +209,6 @@ class SurveyAdministrationController extends LSBaseController
 
         if (Permission::model()->hasGlobalPermission('superadmin', 'read')) {
             $aData['issuperadmin'] = true;
-        } else {
         }
         $aData['model'] = new Survey('search');
         $aData['groupModel'] = new SurveysGroups('search');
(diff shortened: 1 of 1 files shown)


thedirtypanda (approved):

Shnoulle:
It's not Check here, it's sanitize or cast. Check is better for security

See opquast : https://checklists.opquast.com/en/web-quality-assurance/if-any-data-entered-in-the-form-are-rejected-the-fields-containing-the-rejected-data-are-highlighted-for-the-user and https://checklists.opquast.com/en/web-quality-assurance/if-any-data-entered-in-the-form-are-rejected-the-reasons-why-are-explained-to-the-user

See : https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures/

Trischi80:
> It's not Check here, it's sanitize or cast. Check is better for security
> 
> See opquast : https://checklists.opquast.com/en/web-quality-assurance/if-any-data-entered-in-the-form-are-rejected-the-fields-containing-the-rejected-data-are-highlighted-for-the-user and https://checklists.opquast.com/en/web-quality-assurance/if-any-data-entered-in-the-form-are-rejected-the-reasons-why-are-explained-to-the-user
> 
> See : https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures/
> 
> > Ensure all login, access control, and server-side input validation failures can be logged with sufficient user context to identify suspicious or malicious accounts and held for enough time to allow delayed forensic analysis.

Dear Deni,

my original task was to santinize incoming parameters. But sure if you want to log something, you can do it. Maybe we can add a task for that too.

Greetings 
Patricia

Shnoulle:
@Trischi80 : before : sending bad request : no action is done. You update the action : i think it's best to directly send error.
 
And you do sometimes , for example `actionGetTokenTopBar` send an error (400 if i don't make error myself) if you send sid=aa in param.
Then seems best to always send 400 error.

https://github.com/LimeSurvey/LimeSurvey/pull/2232/files#diff-f8c970472fc354612651d57fa98027ecd114f95cf42e4594e0b0b61340ec217fR1380

I create a DEV mantis issue about this … since years now.
If we always santitize or cast: we never update to send error: you know this too ;) (don't touch something working).

PR #2233, opened 2022-02-07
Zoho/l41 t1668 check params usermanagementcontroller

Trischi80:
Dev: check GET and POST params


application/controllers/SurveyAdministrationController.php
```diff
@@ -2006,6 +1983,7 @@ public function actionOrganize()
 
         $iSurveyID = $this->getSurveyIdFromGetRequest();
```
thedirtypanda, line 1985: Maybe you could combine those two lines to only one? `$iSurveyID = sanitize_int($this->getSurveyIdFromGetRequest));`
What do you think?

application/controllers/UserManagementController.php
```diff
@@ -378,8 +394,8 @@ public function actionUserPermissions()
             );
         }
 
-        $oRequest = Yii::app()->request;
-        $userId = $oRequest->getParam('userid');
+        $userId = Yii::app()->request->getParam('userid');
+        $userId = sanitize_int($userId);
```
thedirtypanda, line 398: The same for that one. `$userId = sanitize_int(Yii::app()->request->getParam('userid');` Its just an idea but im not sure. Maybe you have another opinion about that.

application/controllers/UserManagementController.php
```diff
@@ -465,9 +481,8 @@ public function actionUserTemplatePermissions(): ?string
             );
         }
         $aTemplateModels = Template::model()->findAll();
-        $oRequest = Yii::app()->request;
-        $userId = $oRequest->getParam('userid');
-        $oUser = User::model()->findByPk($userId);
+        $userId = Yii::app()->request->getParam('userid');
+        $oUser = User::model()->findByPk((int)$userId);
```
thedirtypanda, line 485: See above.

thedirtypanda (changes requested):

Shnoulle:
If someone try to hack my instance sending 123<script>alert("XSS")</script> : i want to have a log at minimum …

See https://github.com/LimeSurvey/LimeSurvey/pull/2232#issuecomment-1031309986 

application/controllers/SurveyAdministrationController.php
Trischi80: Yes sure :). Thanks Jessi

PR #2237, opened 2022-02-09
Fixed issue #17879: Remote Control `delete_question` error message is too generic

edgarrmondragon:
Fixed issue [#17879](https://bugs.limesurvey.org/view.php?id=17879): Remote Control `delete_question` error message is too generic


c-schmitz:
Thank you!

PR #2243, opened 2022-02-14
Zoho/l41 t1672 check params responsescontroller

Trischi80:
Dev: investigation of incoming parameters (typecast, sanitize) and inserted exceptions when params are not integer


Shnoulle:
:heart_eyes:

PR #2244, opened 2022-02-14
Zoho/l41 t1673 check params surveygroupspermissioncontroller

Trischi80:
Dev: investigated action params (sanitize), only added type declaration to function arguments in actions


application/controllers/SurveysGroupsPermissionController.php
```diff
@@ -409,7 +409,7 @@ public function actionSave($id)
      * @param integer $uid user id
      * @return void
      */
-    public function actionDeleteUser($id, $uid)
+    public function actionDeleteUser(int $id, $uid)
```
thedirtypanda, line 412: Please add typehint to the second param $uid.

thedirtypanda (changes requested):
Fix the param hints and then it can be merged :)

thedirtypanda (approved):

PR #2250, opened 2022-02-15
Fixed issue: [security] Update packages and document package building process for js packages

thedirtypanda:
We have currently ~140 security issues in your JS asset libraries which can be seen in your dependabot alerts on Github [https://github.com/LimeSurvey/LimeSurvey/security/dependabot?q=is%3Aopen](https://github.com/LimeSurvey/LimeSurvey/security/dependabot?q=is%3Aopen+)
Also, it seems that there is no documentation on the build/update process of these packages.

These packages need to be either updated or removed (if obsolete), and the process needs to be documented.


Shnoulle:
But : we have same package in 3.X ? No ?
If one package have real security issue in master : it have too in 3.X ?

No ?

thedirtypanda:
@Shnoulle We changed the dependant bot now. Every branch (develop, master and LTS) will be checked. If anything raises up, it will be fixed by me. Thanks for your hint.

Shnoulle:
Thanks to you :)

I didn't know there are difference on package in 3.X and master :). I really think it's the same except some …

PR #2251, opened 2022-02-17
Fixed issue #17322: [security] Separate settings for administration & participants for handling failed login attempts

Trischi80:
Fixed issue #17322: [security] Separate administration and participant "failed login attempts", add a button to reset "failed login attempts" for participants (via access token)


Shnoulle:
Related issue `#17322`: Need different time and count for lock out access for token VS admin user

installer/create-database.php
```diff
@@ -142,6 +142,8 @@ function populateDatabase($oDB)
             'ip' =>  "string(40) NOT NULL",
             'last_attempt' =>  "string(20) NOT NULL",
             'number_attempts' =>  "integer NOT NULL",
+            'is_frontend' => "integer NOT NULL DEFAULT 0"
```
Shnoulle, line 145: We can not use integer(1) ? Unsure it work, but best if :)

Shnoulle: There are no issue before with boolean on MSSQL  or PG ?

c-schmitz:
Please also add separate settings in config-defaults.php for participants and administration.

PR #2253, opened 2022-02-18
Dev #T315: Make xml_path relative for question theme

gabrieljenik:
WIP

### DIFF
--- a/application/models/QuestionTheme.php
+++ b/application/models/QuestionTheme.php
@@ -33,6 +33,10 @@ use LimeSurvey\Helpers\questionHelper;
  */
 class QuestionTheme extends LSActiveRecord
 {
+    const THEME_TYPE_CORE = 'coreQuestion';
+    const THEME_TYPE_CUSTOM = 'customCoreTheme';
+    const THEME_TYPE_USER = 'customUserTheme';
+
     /**
      * @return string the associated database table name
      */
@@ -1086,5 +1090,43 @@ class QuestionTheme extends LSActiveRecord
         $questionTheme->question_type = $questionType;
         $questionTheme->settings = $settings;
         return $questionTheme;
+    }
+
+    /**
+     * Returns the type of question theme (coreQuestion, customCoreTheme, customUserTheme)
+     *
+     * coreQuestion = Themes shipped with Limesurvey that don't extend other theme
+     * customCoreTheme = Themes shipped with Limesurvey that extend other theme
+     * customUserTheme = User provided question themes
+     *
+     * @return string
+     */
+    public function getThemeType()
+    {
+        if ($this->core_theme) {
+            if (empty($this->extends)) {
+                return self::THEME_TYPE_CORE;
+            } else {
+                return self::THEME_TYPE_CUSTOM;
+            }
+        } else {
+            return self::THEME_TYPE_USER;
+        }
+    }
+
+    /**
+     * Returns the XML path relative to the path configured for the question theme type
+     * @return string
+     */
+    public function getRelativeXmlPath()
+    {
+        $type = $this->getThemeType();
+        $directories = self::getQuestionThemeDirectories();
+        $typeDirectory = $directories[$type];
+
+        // xml_path is supposed to contain the type directory, so we extract the rest of the path
+        $relativePath = substr($this->xml_path, strpos($this->xml_path, $typeDirectory) + strlen($typeDirectory));
+        $relativePath = ltrim($relativePath, "\\/");
+        return $relativePath;
     }
 }
--- a/application/models/QuestionTheme.php
+++ b/application/models/QuestionTheme.php
@@ -233,7 +233,8 @@ class QuestionTheme extends LSActiveRecord
     public function getManifestButtons()
     {
         $sLoadLink = CHtml::form(array("themeOptions/importManifest/"), 'post', array('id' => 'forminstallquestiontheme', 'name' => 'forminstallquestiontheme')) .
-            "<input type='hidden' name='templatefolder' value='" . $this->xml_path . "'>
+            "<input type='hidden' name='templatefolder' value='" . $this->getRelativeXmlPath() . "'>
+            <input type='hidden' name='theme_type' value='" . $this->getThemeType() . "'>
             <input type='hidden' name='theme' value='questiontheme'>
             <button id='template_options_link_" . $this->name . "'class='btn btn-default btn-block'>
             <span class='fa fa-download text-warning'></span>
@@ -250,14 +251,21 @@ class QuestionTheme extends LSActiveRecord
      * @param string $sXMLDirectoryPath the relative path to the Question Theme XML directory
      * @param bool   $bSkipConversion   If converting should be skipped
      * @param $bThrowConversionException If true, throws exception instead of redirecting
+     * @param string|null $themeType The theme type. If specified, the path is considered to be relative to the theme type directory.
      * @return bool|string
      * @throws Exception
      * @todo Please never redirect at this level, only from controllers.
      */
-    public function importManifest($sXMLDirectoryPath, $bSkipConversion = false, $bThrowConversionException = false)
+    public function importManifest($sXMLDirectoryPath, $bSkipConversion = false, $bThrowConversionException = false, $themeType = null)
     {
         if (empty($sXMLDirectoryPath)) {
             throw new InvalidArgumentException('$templateFolder cannot be empty');
+        }
+
+        // If a theme type is specified, the $sXMLDirectoryPath is supposed to be relative to the corresponding
+        // theme type's directory. So we need to get the full path.
+        if (!is_null($themeType)) {
+            $sXMLDirectoryPath = self::getQuestionThemeDirectoryForType($themeType) . '/' . $sXMLDirectoryPath;
         }
 
         // convert Question Theme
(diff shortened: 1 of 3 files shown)


olleharstedt:
Action should be green before code review.

Shnoulle:
WIP : i wait

application/models/services/ThemeQuestionAttributeProvider.php
```diff
@@ -41,7 +41,7 @@ protected function getAttributesFromQuestionTheme($questionThemeName, $questionT
 
         $questionTheme = \QuestionTheme::model()->findByAttributes([], 'name = :name AND extends = :extends', ['name' => $questionThemeName, 'extends' => $questionType]);
         if ($questionTheme !== null) {
-            $xmlFilePath = $questionTheme['xml_path'] . '/config.xml';
+            $xmlFilePath = $questionTheme->getRelativeXmlPath() . '/config.xml';
```
olleharstedt, line 44: There must be other locations where xml_path is read??

gabrieljenik: Maybe not needed to replace all other spots.
We could, but would replace it by some other method, like getXmlPath().
I have left a comment on Zoho.

olleharstedt: Hm, are you saying absolute path is OK in some places but not others? :thinking:

gabrieljenik: Yes. Absolute and "LS root based paths" can be ok when used internally. Actually, it is needed as to read the config files.
Still, when exposing the paths, as in hidden files, that's where we want to "split" the absolute path in:
- relative path
- type of plugin

Usign those 2 components, later when processing the hidden fields, the path can be located.

olleharstedt: Well, first task is to make CI green.

gabrieljenik: Let's make LS green again!

olleharstedt: :cow: :notes: 

Happy cow :D

c-schmitz:
This needs top be tested in cloud hosting environment.

1. Install a question theme
2. Move installation to a different instance
3. Check if question themse still works

c-schmitz:
As this is a (urgent) bug fix, I changed the target to master.

olleharstedt:
@gabrieljenik Hm, guess we need testing instructions so that @glimz can check it.

gabrieljenik:
I don't like ending up in the easy answer, but I think we need a full regression testing.
Upload, Install, Uninstall, Use, Render

olleharstedt:
Hm, want me to add another task to add tests for question theme? Dunno if we have any at the moment.

application/models/QuestionTheme.php
```diff
@@ -1087,4 +1093,77 @@ public static function getDummyInstance($questionType)
...
+
+        // xml_path is supposed to contain the type directory, so we extract the rest of the path
+        $relativePath = substr($this->xml_path, strpos($this->xml_path, $typeDirectory) + strlen($typeDirectory));
+        $relativePath = ltrim($relativePath, "\\/");
+        return $relativePath;
+    }
+
+    /**
+     * Returns the XML path
+     * It may be absolute or relative to the Limesurvey root
```
olleharstedt, line 1136: Hm, it should always be relative, but question theme is relative config userquestionthemerootdir, not the LS root. Right?

Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/blob/26349710d4885c58ef542e80b8dae93d7afb9514/application/config/config-defaults.php#L729-L730 : rootdir not used here neither [webroot](https://github.com/LimeSurvey/LimeSurvey/blob/26349710d4885c58ef542e80b8dae93d7afb9514/application/config/config-defaults.php#L745)

application/models/QuestionTheme.php
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/blob/26349710d4885c58ef542e80b8dae93d7afb9514/application/config/config-defaults.php#L737

gabrieljenik:
> Hm, want me to add another task to add tests for question theme? Dunno if we have any at the moment.

Yes, could be a good idea.

gabrieljenik:
@olleharstedt 
> Hm, it should always be relative

In master, if you have customized the userquestionthemerootdir with an absolute path, an absolute path is saved on DB.
Maybe that's a bug we need to solve in here as well?

olleharstedt:
> In master, if you have customized the userquestionthemerootdir with an absolute path, an absolute path is saved on DB.
> Maybe that's a bug we need to solve in here as well?

Hm, nah. I think it's OK when userquestionthemerootdir is relative. Will tell Ahmed to test it like that.

olleharstedt:
The purpose it to make question theme path configurable without breaking the system. For that, xml_path must be relative.

Shnoulle:
Like surveytheme then ( `standardthemerootdir`)

Must allow update directory only after update (old question theme not fixed) or not ?

I mean : 

1. Install a theme1 and theme2 on current directory
2. Update to this version
3. Update `corequestionthemedir`
4. Move theme1 but not theme 2 : 

What is the desired behaviour here ?

Shnoulle:
> What do you mean?

It's for test todo : you start by default directory and update it after

1. Set up QuestionTheme after fix, moving after fix
2. Set up QuestionTheme before fix, moving after fix
3. Set up QuestionTheme before fix, moving before fix

What test must be done according to 

1. Moving question theme too
2. Not moving question theme

application/models/QuestionTheme.php
```diff
@@ -229,7 +233,8 @@ public function getVisibilityButton()
     public function getManifestButtons()
     {
         $sLoadLink = CHtml::form(array("themeOptions/importManifest/"), 'post', array('id' => 'forminstallquestiontheme', 'name' => 'forminstallquestiontheme')) .
-            "<input type='hidden' name='templatefolder' value='" . $this->xml_path . "'>
+            "<input type='hidden' name='templatefolder' value='" . $this->getRelativeXmlPath() . "'>
```
Shnoulle, line 236: It's the only place where you use getRelativeXmlPath function, right ?

Shnoulle (approved):
Code OK, not tested

gabrieljenik:
@olleharstedt @ptelu Maybe good to merge to BS5 branch?

olleharstedt:
No, but maybe current dev branch, which will be 5.5

olleharstedt:
Changed base to develop

lajosarpad:
The notes for the ticket as far as I know were

T315

These are my notes.

Right now, xml_path is

1) Relative to ROOT folder, or:
2) Absolute (user upload)

Should be relative to three different folders

1) Core
2) Custom (also in git repo)
3) User

Preferable solution: Have a method $theme->getXmlPath(); that solves it.


>> Relative to the LS installation directory, right?

Not LS root, but rather the configured installation folder of that particular question theme type. On hosting, LS root is completely different than user data folder.

The implementation has to work on our hosting, where question theme folder is customized.

>> OK, can you show a sample of that customization?

Yep, config-defaults.php contains customizations, e.g.

// Line 728 approx
$config['configdir'] = / Our custom user data folder /

lajosarpad:
Conflicts were solved. Further notes:

1. Sanitized $templateFolder (**recent change**)

2. Used the newly created constants of THEME_TYPE_CORE, THEME_TYPE_CUSTOM, THEME_TYPE_USER in QuestionTheme more consistently (**recent change**)

3. xml_path is at some places referenced as is, instead of using the new getRelativeXmlPath. So we are using absolute path in some cases and relative path at others.

4. According to https://github.com/LimeSurvey/LimeSurvey/pull/2253#issuecomment-1060989160, we need to test this on cloud environment. Should this be part of ls-cloud rather than ls-ce?

5. Question config paths. I will need information about this in order to determine whether it's correct

6. getXmlPath just returns whatever can be found inside $this->xml_path. Is this what we expect?

gabrieljenik:
Thanks @lajosarpad 

> xml_path is at some places referenced as is, instead of using the new getRelativeXmlPath. So we are using absolute path in some cases and relative path at others.

Which ones? Do you have an example? Could be good to review?
I had the idea everything was replaced. We could have missed them or they could be coded after this PR.

> Question config paths. I will need information about this in order to determine whether it's correct

Got lost here.

> getXmlPath just returns whatever can be found inside $this->xml_path. Is this what we expect?

I guess so. If recall correctly, this was just a dummy getter. Any idea? Any expectation?

lajosarpad:
@gabrieljenik 

### 1. xml_path

We have `QuestionTheme::search`

image

as well as usages of `getXmlPath`

image

### 2. config paths

I referred to this comment: https://github.com/LimeSurvey/LimeSurvey/pull/2253#issuecomment-1072276373

I am not sure what the paths should be or whether this is a problem, but left a note to make sure that those who were also involved into this ticket might know.

### 3. The return of getXmlPath

I do not have any idea about what the expectations should be, I left a note to make sure that we clarify this matter.

gabrieljenik:
> 1. xml_path
> We have QuestionTheme::search
> as well as usages of getXmlPath

I guess is fine.

> 2. config paths
> I referred to this comment: https://github.com/LimeSurvey/LimeSurvey/pull/2253#issuecomment-1072276373
> I am not sure what the paths should be or whether this is a problem, but left a note to make sure that those who were also involved into this ticket might know.

I guess they are fine. I believe those are not updated in the PR.
If something is wrong there, it will popup when testing.
right?

> 3. The return of getXmlPath
> I do not have any idea about what the expectations should be, I left a note to make sure that we clarify this matter.

OK, I believe it is OK

lajosarpad:
@gabrieljenik if I'm mistaken, currently we are awaiting for tests to be performed so if anything is missing it will be detected. Can you confirm?

gabrieljenik:
> currently we are awaiting for tests to be performed so if anything is missing it will be detected

Yes

gabrieljenik:
@tiborpacalat Can you help testing this?

PR #2255, opened 2022-02-18
fixing Mantis:17903 , limesurvey not reading ip address from multiple proxies

tassoman:
Fixed issue #17903 (3.x-LTS)


Shnoulle:
Arg … PSR2 is great but not on a fix please …

Only PSR2 or ONLY the fix …

Shnoulle (changes requested):
No other update than the fix please : no PSR2 PSR12 on the other lines.

please ;)

tassoman:
indeed
revoked by #2258

PR #2257, opened 2022-02-21
Fixed issue #17903: Incorrect IP address used when having multiple forward from proxies

tassoman:
Fixed issue Mantis #17903 - branch 5.x (master)

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -4643,7 +4643,9 @@ function getIPAddress()
             // Single forward 
             $sIPAddress = $_SERVER['HTTP_X_FORWARDED_FOR'];
         } else {
-            // Multitple forward
+		// Multitple forward
+		// see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
+		// TODO: RFC7239 full implementation (https://datatracker.ietf.org/doc/html/rfc7239#section-5.2)
             $aForwarded = explode(',', $_SERVER['HTTP_X_FORWARDED_FOR']);
             if (false !== filter_var($aForwarded[0], FILTER_VALIDATE_IP)) {
                 $sIPAddress = $aForwarded[0];
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -4625,24 +4625,37 @@ function ellipsize($sString, $iMaxLength, $fPosition = 1, $sEllipsis = '&hellip;
 }
 
 /**
-* This function tries to returns the 'real' IP address under all configurations
-* Do not rely security-wise on the detected IP address as except for REMOTE_ADDR all fields could be manipulated by the web client
-*/
+ * This function tries to returns the 'real' IP address under all configurations
+ * Do not rely security-wise on the detected IP address as except for REMOTE_ADDR all fields could be manipulated by the web client
+ *
+ * @return	string	Client's IP Address
+ */
 function getIPAddress()
 {
     $sIPAddress = '127.0.0.1';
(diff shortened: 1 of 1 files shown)


application/helpers/common_helper.php
```diff
@@ -4625,25 +4625,38 @@ function ellipsize($sString, $iMaxLength, $fPosition = 1, $sEllipsis = '&hellip;
...
+        //Check IP Address passed from proxy
+        $vComma = strpos($_SERVER['HTTP_X_FORWARDED_FOR'], ',');
+        if (false === $vComma && filter_var($_SERVER['HTTP_X_FORWARDED_FOR'], FILTER_VALIDATE_IP) !== false) {
+            // Single forward 
+            $sIPAddress = $_SERVER['HTTP_X_FORWARDED_FOR'];
+        } else {
+            // Multitple forward
+            $aForwarded = explode(',', $_SERVER['HTTP_X_FORWARDED_FOR']);
+            if (false !== filter_var($aForwarded[0], FILTER_VALIDATE_IP)) {
+                $sIPAddress = $aForwarded[0];
```
Shnoulle, line 4649: Then it take the first one ?
`127.0.0.1,127.0.0.2,127.0.0.3` return `127.0.0.1`

Seems OK with stackoverflow , but a better source can be great
Adding a source for explanation (rfc ; https://datatracker.ietf.org/doc/html/rfc7239#section-5.2 ?)

Shnoulle (approved):
Maybe adding a source for comment ?

application/helpers/common_helper.php
tassoman: I based my contribution on [Mozilla Http Headers documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For). 
The validation goes beyond function behaviour.  I guess is done inside `filter_var($var, FILTER_VALIDATE_IP)`

Shnoulle: Yes, not about `FILTER_VALIDATE_IP` just adding the mozilla link on comment about explode ;)

application/helpers/common_helper.php
```diff
@@ -4625,25 +4625,38 @@ function ellipsize($sString, $iMaxLength, $fPosition = 1, $sEllipsis = '&hellip;
...
-        //Check IP address passed from proxy
-        $sIPAddress = $_SERVER['HTTP_X_FORWARDED_FOR'];
+    } elseif (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
+        //Check IP Address passed from proxy
+        $vComma = strpos($_SERVER['HTTP_X_FORWARDED_FOR'], ',');
+        if (false === $vComma && filter_var($_SERVER['HTTP_X_FORWARDED_FOR'], FILTER_VALIDATE_IP) !== false) {
+            // Single forward 
+            $sIPAddress = $_SERVER['HTTP_X_FORWARDED_FOR'];
+        } else {
+            // Multitple forward
```
Shnoulle, line 4646: `// Multitple forward , see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For`

for example ?

tassoman: can I push another commit to this branch-pr ? :thinking:

Shnoulle: Why not ? I can do if you want ;)

Shnoulle: You can locally, or if you want : you can edit this file online for examle

tassoman: I've pushed both! Thank you

Shnoulle (approved):

tassoman:
@Shnoulle , maybe the `strpos()` part can be archived returning always 1st `explode()` element.
if the separator is not present, [the entire string is returned as array element](https://www.php.net/manual/en/function.explode.php). 🤔

PR #2259, opened 2022-02-21
Fixed issue #15057: Translation screen not working correctly when there are long question text

gabrieljenik:


glimz:
@gabrieljenik We also need a PR for master

Shnoulle:
test failed overflow:auto don't work on cell.
`word-wrap: break-word;` seem best 
Capture d’écran du 2022-02-28 11-39-29

Shnoulle (changes requested):
overflow didn't work.

Maybe update a little html ? Give specific class to each cell ? .code, .text, .input ?

Shnoulle:
I see the change : if we add a div : unsure we need word-wrap …

I don't know here what is the best solution …

gabrieljenik:
I think the word-wrap can help in some cases where we have simple text or formatted text.
It doesn't bother, but it helps.

We forsee some cases that may not be great looking, as below.
But I guess that to tackle those, we may need a new ticket as it needs a bigger intervention and rethinknig this screen (as for example using Flex).

image

Shnoulle (approved):

PR #2262, opened 2022-02-22
Fixed issue #12105: Scroll bar partially broken at browse response screen

gabrieljenik:


gabrieljenik:
Well, what would you do?

Shnoulle:
see phpMyAdmin for example, or 2.6lts (but need fixed header) : https://bugs.limesurvey.org/view.php?id=12105#c68017

The windows scroll are used directly : it's really an improvment.

c-schmitz:
The listpager is moving with it -. that is because the name changed.
replace in listrespone.js

        var $elListPager = $('#ListPager');
with 
        var $elListPager = $('#reponsesListPager');

PR #2266, opened 2022-02-23
Fixed issue #17838: Add javascript in script tab only works for superadmin

gabrieljenik:
- Enable 'Show script field' user setting by default


Shnoulle:
It was false by default ? Argl … :+1: for true by default …

glimz (approved):

PR #2268, opened 2022-02-23
Dev #17859: Disable core plugins from config file

gabrieljenik:


Shnoulle (approved):
Don't see any potential code issue

PR #2270, opened 2022-02-25
use different cache key for QuestionGroup model getTotalGroupsWithQue…

jackrabbithanna:
…stions() query

For issue:
https://bugs.limesurvey.org/view.php?id=17912

Sometimes, when emcache is enabled, an erroneous value is cached for the OptionGroup model, getTotalGroupsWithoutQuestions() function

This results in error taking the survey, when there should not be. See bug issue for more details.


c-schmitz:
Thank you very much!

PR #2271, opened 2022-02-26
DEV: Fix config.xml

gabrieljenik:
- Add missing 'type' on Bootstrap Dropdown config.xml


Shnoulle:
test failed ? I need to do same to check theme page :)

gabrieljenik:
@glimz Can you please remove the test failed tag? This needs to be pushed. Has no story. This is because when using debug a dump happens when entering question editor.

gabrieljenik:
@olleharstedt Can I ask for your reviewal?

Shnoulle:
> test failed ? I need to do same to check theme page :)

:)

Shnoulle:
We need it in master … No way to get theme list with debug> 0 …

PR #2273, opened 2022-02-28
Fixed issue #17677: Valid TSV file can not be used in VV import

Shnoulle:
Dev: allow line feed, checked with latin1 file too.
Dec: see https://github.com/LimeSurvey/LimeSurvey/commit/3197b7ebec3eff206d198da2b161669e6677a4f3


Shnoulle:
Why broke on create question ?

PR #2274, opened 2022-02-28
Dev: Less potential issue with console

Shnoulle:
Dev: using better rootdir

Dev:

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -38,10 +38,13 @@ $config = array();
 
 // FILE LOCATIONS
 
-$config['rootdir'] = getcwd(); //dirname(__FILE__); // This is the physical disk location for your limesurvey installation. Normally you don't have to touch this setting.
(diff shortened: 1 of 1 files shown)


c-schmitz (changes requested):
That does not seem right. getcwd is definitely better then a hardcoded relative path.
If this collides with console then why not change it only for console scripts.

Shnoulle:
If you do php www/htdocs/limesurvey… : getcwd is broken.

realpath(dirname(dirname(dirname(__FILE__)))); is not really hardcoded : it take the directory ./application/config/../../.. then always the limesurvey directory.

The worst part is https://gitlab.com/SondagesPro/mailing/sendMailCron/-/blob/master/sendMailCron.php#L462 , since rootdir is broken : it broke after.

I can create a plugin to check new issue in 5.X maybe.

Shnoulle:
I find a way to set it only for command then …

PR #2278, opened 2022-03-01
Fixed issue: Bootstrap 5 - Admin Theme Generator - Move code into pro…

thedirtypanda:
Moved the admin theme generator tool into the project.


Shnoulle:
:+1: Great thanks !

Why not as a related project ?

thedirtypanda:
@Shnoulle We never touch that project. Only if we are generating a new version of the admin themes ... thats why we thought about it. We want to reduce the smaller projects. But its not final ... so things may change.

Shnoulle:
Yes, but not needed for a lot of user :).
Less file take less place ;)

PR #2284, opened 2022-03-04
Dev #8436: Enable / disable users and fields to manage users

gabrieljenik:
From these requirements:
- a flag to activate/deactivate the user (so I could reset the user when I start a new class)
- a field to set an expiry date for the password
- a field to set an expiry date for the user
- a field with "last login date" to display the last date they logged in the system

We believe it is ok to start with these:
- [x] a field to set an expiry date for the user
- [x] a field with "last login date" to display the last date they logged in the system


c-schmitz:
Expiry date for the password is something that maybe  should be part of the password requirment plugin, imho.

gabrieljenik:
> Expiry date for the password is something that maybe should be part of the password requirment plugin, imho.

Agree. A plugin setting.

gabrieljenik:
BTW, "a field to set an expiry date for the user", could be on another branch/PR. So I guess this is could be OK to merge after testing

Shnoulle:
> Expiry date for the password is something that maybe should be part of the password requirment plugin, imho.

Warning : all password system are linkend to AuthDB only . Don't send warning etc … to user with only LDAP account (for example).

c-schmitz:
> a field to set an expiry date for the user

I would not implement this.
However, a simple field/toggle to enable/disable a user would be of value.

gabrieljenik:
@c-schmitz Why would you prefer an on/off setting rather than an expiration date? (would be kind similar to how surveys are expired) ?

glimz:
Tested failed: The action buttons when clicked and closed, there is still a modal displayed. Seems there are two modals opened.
Edit user does not work. A modal is displayed and it is loading infinitely.

With these errors, I cannot verify the dev

gabrieljenik:
I wasn't able to reproduce the error.
Can you send me a recording or something similar?
Are you sure assets were properly cleared and that cache wasn't an issue? Sometimes those things are cache issues as button and modals rely on JS.

gabrieljenik:
Retested. :)

gabrieljenik:
@olleharstedt Want to merge it?

olleharstedt:
Because of the difference in datetime picker implementation in Bootstrap 5, this is now broken in one of our branches. :(

olleharstedt:
Where does "#8436" point to? Which project manager? o0

olleharstedt:
Should say "New feature #xxx" no? Now it's hidden in the changelog.

gabrieljenik:
> Because of the difference in datetime picker implementation in Bootstrap 5, this is now broken in one of our branches. :(

Hmm hard matter. Maybe we could do sme kind of overloading to the datepicker method to be able to work in BS5 ?
How are you handling that matter?

gabrieljenik:
> Should say "New feature #xxx" no? Now it's hidden in the changelog.

Sorry. Not sure why, but thought DEV was a valid option.
Should we create a dummy PR? :)

gabrieljenik:
> Also, new features require an acceptance test, but now this wasn't written? The epic branch CI is green despite being broken in user management. :(

Not sure about the approach I looked for this.
It has been a long time.
I wish I would have thought that if this development was liked, I would write the test after...

I will do the test

olleharstedt:
Yeah, also not good that we (in Hamburg) leave PRs for so long, we're looking for more people to help there.

> I will do the test

Awesome, I was about to move the Mantis ticket to "new" but then we had 3 bots hugging the server :skull:

olleharstedt:
image

"Expire date" should be enough? Why "date/time"?

Shnoulle:
> Is the user expiring, or the password?

User acces via AuthDB … not via other authentication method …

olleharstedt:
"Start date/time:" - oh! I see this is how it looks in access control settings, too. Sorry. Heh.

PR #2285, opened 2022-03-07
Fixed issue [security] : Missing password policy when saving a survey

Shnoulle:
Dev: new event validateSaveSurveyForm allowing check password and more
Dev: use it in existing PasswordRequirement plugin

### DIFF
--- a/application/core/plugins/PasswordRequirement/PasswordRequirement.php
+++ b/application/core/plugins/PasswordRequirement/PasswordRequirement.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copre plugin for LimeSurvey
+ * Core plugin for LimeSurvey : password requirement settings
  * @version 1.1.0
  */
 
@@ -38,7 +38,7 @@ class PasswordRequirement extends \LimeSurvey\PluginManager\PluginBase
         ),
         'surveySaveActive' => array(
             'type' => 'boolean',
-            'label' => 'Check password when save survey',
+            'label' => 'Check password when saving survey',
             'default' => false,
         ),
         'surveySaveNeedsNumber' => array(
@@ -70,7 +70,7 @@ class PasswordRequirement extends \LimeSurvey\PluginManager\PluginBase
         $this->subscribe('checkPasswordRequirement');
         $this->subscribe('createRandomPassword');
 
-        $this->subscribe('SaveSurveyForm', 'validateSaveSurveyForm');
+        $this->subscribe('saveSurveyForm', 'validateSaveSurveyForm');
     }
 
     public function checkPasswordRequirement()
@@ -105,14 +105,14 @@ class PasswordRequirement extends \LimeSurvey\PluginManager\PluginBase
             // Action only when validate
             return;
         }
-        $aSaveDatas = $event->get('aSaveDatas');
+        $saveData = $event->get('saveData');
         $aSaveErrors = $event->get('aSaveErrors');
-        if (empty($aSaveDatas['clearpassword'])) {
+        if (empty($saveData['clearpassword'])) {
             // No need to check password if empty : core disallow it
             return;
         }
 
-        $password = $aSaveDatas['clearpassword'];
+        $password = $saveData['clearpassword'];
         $errors = $this->checkValidityOfPassword(
             $password,
             $this->get('surveySaveNeedsNumber', null, null, false),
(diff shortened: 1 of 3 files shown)


Shnoulle:
Better event : move to `SaveFormEvent` with `state` (get, validate)

application/core/plugins/PasswordRequirement/PasswordRequirement.php
```diff
@@ -1,4 +1,8 @@
 <?php
+/**
+ * Copre plugin for LimeSurvey
```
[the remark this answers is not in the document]
Shnoulle: core … core plugin 😩

application/libraries/Save.php
```diff
@@ -351,4 +285,21 @@ function set_answer_time()
...
     }
+
+    /**
+     * Launch the event SaveForm
+     * @param $state
+     * @return void
+     */
+    private function launchSaveFormEvent($surveyid, $state = 'show')
+    {
+        $saveSurveyEvent = new PluginEvent('SaveSurveyForm');
```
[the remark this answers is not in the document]
Shnoulle: Previous one , yes . Didn't know best one. wikimedia set it to uppercase ;)

Shnoulle:
Need test ?

glimz:
@Shnoulle Did you already make the changes?

Shnoulle:
String update ? Yes.

I mean : did we need to add a [frontend test](https://github.com/LimeSurvey/LimeSurvey/tree/master/tests/functional/frontend)?  I can do it (need some more time). 

But unsure current plugin have one ?

Shnoulle:
> @Shnoulle Did you already make the changes?

Another update to do ?

application/libraries/Save.php
```diff
@@ -182,26 +119,29 @@ function saveSurvey()
 
         $aSaveForm  = array();
         $timeadjust = getGlobalSetting('timeadjust');
+        $this->saveData = array(
```
olleharstedt, line 122: Does this have to be an object property? What about passing around a DTO instead?

olleharstedt: DTO = data transfer object

Shnoulle: Why ? It's clearly an array of value filled by POST, allowed to be updated in `PluginEvent('saveSurveyForm')`

Why it can be used as an object ?

I really think it's not an object. It's an array of value sent by user, updatable by plugin. Moved `$_POST` to `$this->saveData`

olleharstedt: I don't like when devs use object properties just because it's easy to write. Sending DTOs as argument to functions more clearly communicates what the function needs to do a task. IMO. :) 

Basically never use associative arrays: https://steemit.com/php/@crell/php-use-associative-arrays-basically-never

Shnoulle: $_POST is an associative array.

Dynamic form by Yii use associative array too (because came from POST value).

Shnoulle: I don't want to make it complex for plugin dev .
It's here to allow plugin to update : not more.

If needed i come back to App()->request->getPost …

olleharstedt: Yeah, never use $_POST, always the request object.

Class is less complex than associative array, always, because it has will-defined properties.

olleharstedt: I can checkout this PR and check more closely, hard to see on github...

Shnoulle: About $_POST : see the commit : saveData replace $_POST to allow update by plugin.

I can 4 class variables and send it to plugin event (and allow update)
I can create a class and send it to plugin event and allow update

But i really think it's more clear if it's a array of value here .

olleharstedt: Bah. I'd prefer if launchSaveFormEvent() returned values instead of setting object properties, but fine. I'll merge.

Shnoulle: You mean (sample)

`$saveSurveyEvent->set('identifier', $identifier);
$saveSurveyEvent->set('email', $email);`

It's what i done at 1st step, but after review : i think a plugin can ADD some input in sabe system and have it in twig view with saveData.myvalue
I didn't update twig file here, but i do it when testing :/

:)

Shnoulle: PS : don't merge new event if it can be improved please.

olleharstedt: Was the manual updated with new event data, btw?

Shnoulle: Currently not, wait for merge ;)

Shnoulle: > Was the manual updated with new event data, btw?

https://manual.limesurvey.org/SaveSurveyForm

PR #2288, opened 2022-03-10
Fixed issue #17940: Array filter in combination with 'equals sum value' feature does not sum up values correctly

gabrieljenik:
There is a missmatch in the questionId used by the twig processString.
That missmatch produces unexpected behaviour on the EM.

The "twig processString" just assumes the current question is the last queston of the group.
So, we moved out the processString usage from the fileupload twig implementation.
Now the processString is done on PHP side.

### DIFF
--- a/application/core/QuestionTypes/MultipleNumericalQuestion/RenderMultipleNumerical.php
+++ b/application/core/QuestionTypes/MultipleNumericalQuestion/RenderMultipleNumerical.php
@@ -326,11 +326,16 @@ class RenderMultipleNumerical extends QuestionBaseRenderer
         ) {
             $qinfo = LimeExpressionManager::GetQuestionStatus($this->oQuestion->qid);
 
+            $sumRemainingEqn = LimeExpressionManager::ProcessString('{'.$qinfo['sumRemainingEqn'].'}', $this->oQuestion->qid);
+            $sumEqn = LimeExpressionManager::ProcessString('{'.$qinfo['sumEqn'].'}', $this->oQuestion->qid);
+
             if (trim($this->getQuestionAttribute('equals_num_value')) != '') {
                 $equals_num_value = true;
             }
             $displaytotal = true;
         }
+
+
 
         $answer .= Yii::app()->twigRenderer->renderQuestion(
             $this->getMainView() . '/answer',
@@ -341,10 +346,10 @@ class RenderMultipleNumerical extends QuestionBaseRenderer
                 'rowTemplate' => $rowTemplate,
                 'dynamicTemplate' => $dynamicTemplate,
                 'id' => $this->oQuestion->qid,
-                'sumRemainingEqn' => $equals_num_value ? $qinfo['sumRemainingEqn'] : '',
+                'sumRemainingEqn' => $equals_num_value ? $sumRemainingEqn : '',
                 'equals_num_value' => $equals_num_value,
                 'displaytotal' => $displaytotal,
-                'sumEqn' => $displaytotal ? $qinfo['sumEqn'] : '',
+                'sumEqn' => $displaytotal ? $sumEqn : '',
                 'sLabelWidth' => $this->widthArray['sLabelWidth'],
                 'sInputContainerWidth' => $this->widthArray['sInputContainerWidth'],
                 'prefix' => $this->prefix,
--- a/application/views/survey/questions/answer/multiplenumeric/rows/dynamic.twig
+++ b/application/views/survey/questions/answer/multiplenumeric/rows/dynamic.twig
@@ -23,7 +23,7 @@
             {% endif %}
             <div id="remainingvalue_{{id}}" class="form-control-static numeric dynamic-remaining" data-number="1">
             <!-- alternative class : form-control : display like an input:text -->
-                {{ processString('{'~sumRemainingEqn~'}') }}
+                {{ sumRemainingEqn }}
             </div>
             {% if suffix  %}
                 <div class="ls-input-group-extra suffix-text suffix">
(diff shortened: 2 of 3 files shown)


Shnoulle:
You mean the `sumRemainingEqn` set to twig file is not always this one of this question ?

If yes : this must be fixed, in my opinion : allow QuestionTheme to get current value to so something is **needed** for other theme. A way to get the current attributes of the question.

Else must check all other question : just fo rmultiple text :+1: 

https://github.com/LimeSurvey/LimeSurvey/blob/1e04d703e926f5c9cf22335f9c7cb9f00b81b761/application/views/survey/questions/answer/multipleshorttext/rows/answer_row_inputtext.twig#L42
https://github.com/LimeSurvey/LimeSurvey/blob/1e04d703e926f5c9cf22335f9c7cb9f00b81b761/application/views/survey/questions/answer/multipleshorttext/rows/answer_row_inputtext.twig#L33
https://github.com/LimeSurvey/LimeSurvey/blob/1e04d703e926f5c9cf22335f9c7cb9f00b81b761/application/views/survey/questions/answer/multipleshorttext/rows/answer_row_inputtext.twig#L49

Maybe it's the way of using it ?

Shnoulle:
Need a better lss to make a better fix …
Because the original issue are not always reproducible ...

gabrieljenik:
> Need a better lss to make a better fix … Because the original issue are not always reproducible ...

This helps
[limesurvey_survey_144143_(1).lss.txt](https://github.com/LimeSurvey/LimeSurvey/files/8232206/limesurvey_survey_144143_.1.lss.txt)
?

gabrieljenik:
> If yes : this must be fixed, in my opinion : allow QuestionTheme to get current value to so something is needed for other theme. A way to get the current attributes of the question.

I agree. This problem could mess with other themes as well.

> Maybe it's the way of using it ?

What do you mean?
We need more time to research it.

> twig processString use processStepString : no question id inside.

Yes and that's a problem if you want to use EM as the code generated by the EM heavily relies on the question id to name the replacement variables.

Shnoulle:
> > Need a better lss to make a better fix … Because the original issue are not always reproducible ...
> 
> This helps [limesurvey_survey_144143_(1).lss.txt](https://github.com/LimeSurvey/LimeSurvey/files/8232206/limesurvey_survey_144143_.1.lss.txt) ?

Need condition on last question : ok, realy more clear :)


> Yes and that's a problem if you want to use EM as the code generated by the EM heavily relies on the question id to name the replacement variables.

If it's about order of question : what happen if you put 2 questions between the Multiple numeric : and no way to fix ?

We don't have another fix some week ago about order of question ?

Shnoulle:
This issue : https://bugs.limesurvey.org/view.php?id=13529 reappear …  «Group title and description + Survey name and description are not updated vi EM javascript something change in expression manager between 3.6 and current .

Shnoulle:
Some idea :

1. Update twig function to allow qid usage (and use it here)
2. Report again the issue about Group title and description are not updated via EM javascript (because it work at a time : processString and processStepString are created to fix THIS issue.
3. Check the other usage of processString

With 1 : you don't broke possible hacked theme  (Question or Survey) : there still the same issue, but data sent is the same.

gabrieljenik:
With 1. the issue is we need to send the QID to the twig, which is not sent always.
I would say to open a new ticket, and then we review approaches.

I like the idea of using a static variable for QID.
That would be set by the renderer.
Could be read inside twig, when needed.

Shnoulle:
> With 1. the issue is we need to send the QID to the twig, which is not sent always. I would say to open a new ticket, and then we review approaches.

Ah shit !

> I like the idea of using a static variable for QID. That would be set by the renderer. Could be read inside twig, when needed.


Related : https://bugs.limesurvey.org/view.php?id=14228 ?

gabrieljenik:
> sumRemainingEqnDynamic

What is this? Didn't found it in the code.

Shnoulle:
> > sumRemainingEqnDynamic
> 
> What is this? Didn't found it in the code.

Adding a new var to be used in twig : then : old twig work (except if there are an hidden question wait same condition at end of the group).

- sumRemainingEqn : not updated attribute
- sumRemainingEqnDynamic : updated attribute

gabrieljenik:
> or a quick fix without broke hacked them

On `sumRemainingEqn` we are sending an already processed string. If that gets reprocessed (say by a not-yet-updated custom theme), `sumRemainingEqn` will not get updated, as there is nothing to process there.

In which situation do you forsee the theme being broken?

Shnoulle:
Ah right :).

BUT : i really think QuestionTheme must receive always RAW value 

The issue in this bug are here : https://bugs.limesurvey.org/file_download.php?file_id=15712&type=bug
Not elsewhere.

gabrieljenik:
> i really think QuestionTheme must receive always RAW value

Why?

> The issue in this bug are here : https://bugs.limesurvey.org/file_download.php?file_id=15712&type=bug

You mean in version 3.25.20 ?

Shnoulle:
> > i really think QuestionTheme must receive always RAW value
> 
> Why?

To allow QuestionTheme or SurveyTheme maker to do exactly what they want. Maybe get is as static or get it as dynamic even if core use is as static

> > The issue in this bug are here : https://bugs.limesurvey.org/file_download.php?file_id=15712&type=bug
> 
> You mean in version 3.25.20 ?

I mean : issue [17940](https://bugs.limesurvey.org/view.php?id=17940) happen between  between 3.25.20 and 3.25.22

It's OK in 3.25.20 with the exact same twig code (usage of `processString`), but broke in 3.25.22.
The issue is somewhere here, a commit broke `processString` function.

Shnoulle:
> To allow QuestionTheme or SurveyTheme maker to do exactly what they want. Maybe get is as static or get it as dynamic even if core use is as static

Leave it for globally : https://bugs.limesurvey.org/view.php?id=14228

Shnoulle (changes requested):
:+1:

Shnoulle (approved):

PR #2289, opened 2022-03-11
Fixed issue: cache sqlexception when update-db 481 has not been done …

Trischi80:
Fixed issue: cache sql-exception when update-db 481 has not been done and field 'is_frontend' is not there, this could happen if participant takes survey before db-update has been done

### DIFF
--- a/application/models/FailedLoginAttempt.php
+++ b/application/models/FailedLoginAttempt.php
@@ -22,6 +22,8 @@
  * @property integer $number_attempts
  * @property int $is_frontend  from frontend(=1) or from backend (=0)
  */
+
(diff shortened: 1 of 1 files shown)


Shnoulle:
can not check with `App()->getConfig('DBVersion') >= 481` ? since 

1. Update of DB version is done after adding the column
2. check a static var is more quick than anytyhing else

See https://github.com/LimeSurvey/LimeSurvey/blob/62c1585acca81e65dbb597f3319f00d9ec2a7321/application/models/AssetVersion.php#L65 for example

c-schmitz:
I agree, preventing this by just checking for the current DBVersion would be enough, and cleaner than using exceptions.

PR #2290, opened 2022-03-14
Task/zoho l41 t1700 #17919 add new class for menu button

twilligls:
New Feature zoho/L41-T1700: 
dev: Added new menu class for button item


application/libraries/MenuObjects/Menu.php
```diff
@@ -131,4 +147,20 @@ public function getTooltip()
...
+    {
+        return $this->isPrepended;
+    }
+
+    /**
+     * @return boolean
+     */
+    public function isButton()
+    {
+        return $this->isButton;
```
olleharstedt, line 164: You can return false here and true in the subclass, then you don't need a property.

Or use instanceof, but I see in the code that I didn't do that either.

application/libraries/MenuObjects/MenuButton.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Returns the label string.
+     * if a iconClass is set it will be put before the label
+     * @return string
+     */
+    public function getLabel()
+    {
+        $label = $this->label;
+        if ($this->iconClass !== '') {
+            $label = "<i class='" . $this->iconClass . "'></i>&nbsp;" . $this->label;
```
olleharstedt, line 54: I would prefer if all HTML was in the view files, but no big deal.

olleharstedt (changes requested):

olleharstedt (approved):

PR #2291, opened 2022-03-14
Fixed issue #17950: is_int(false) returns true

c-schmitz:
See https://bugs.limesurvey.org/view.php?id=17950


Shnoulle:
It's OK on PHP ?

Shnoulle (changes requested):
No issue in js, issue in PHP

Shnoulle (changes requested):
Unsure we don't need `is_int('23') == true`

PR #2295, opened 2022-03-14
Fixed issue #17955: newtest/Y in URL is not clearing session

Shnoulle:
Dev: move newtest check before all other action


gabrieljenik (commented):
Haven't tested it. Code looks OK.

gabrieljenik (approved):

PR #2305, opened 2022-03-18
Fixed issue #17963: Numbers on text question does not detect blanks

Shnoulle:
Dev: Number() remove white space then check if not update string

### DIFF
--- a/assets/scripts/expressions/em_javascript.js
+++ b/assets/scripts/expressions/em_javascript.js
@@ -245,17 +245,15 @@ function LEMis_int(mixed_var)
 }
 /**
  * Test if mixed_var is a PHP numeric value
- * From: http://phpjs.org/functions/is_numeric/
+ * Do not use locutus because of LEMradix
  */
 function LEMis_numeric(mixed_var)
 {
     var isNumericRegex = new RegExp(/^(-)?[0-9]*(\.)[0-9]*$/);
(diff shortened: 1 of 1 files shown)


assets/scripts/expressions/em_javascript.js
```diff
@@ -1021,7 +1021,10 @@ function LEMval(alias)
                     return "";
                 }
 
-                var checkNumericRegex = new RegExp(/^(-)?[0-9]*(,|\.)[0-9]*$/);
+                var isNumericRegex = new RegExp(/^(-)?[0-9]*(\.)[0-9]*$/);
```
gabrieljenik, line 1024: This is used more than once. Maybe we can put it on a function named like `getNumericValidatioRegex`?

gabrieljenik: Not sure if this is a must for merging... Will leave it up to you @Shnoulle

Shnoulle: > This is used more than once. Maybe we can put it on a function named like `getNumericValidatioRegex`?

I just fix here, not rewrite.

Shnoulle: > Not sure if this is a must for merging... Will leave it up to you @Shnoulle

?

olleharstedt:
Thanks!

PR #2306, opened 2022-03-18
Fixed issue #17963: Numbers on text question does not detect white space

Shnoulle:
Dev: Number() remove white space then check if not update string
Dev: review some numeric test

### DIFF
--- a/assets/packages/expressions/em_javascript.js
+++ b/assets/packages/expressions/em_javascript.js
@@ -457,14 +457,15 @@ function LEMis_int(mixed_var)
 }
 /**
  * Test if mixed_var is a PHP numeric value
- * From: http://phpjs.org/functions/is_numeric/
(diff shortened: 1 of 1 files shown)


assets/packages/expressions/em_javascript.js
```diff
@@ -1233,7 +1233,10 @@ function LEMval(alias)
                     return "";
                 }
 
-                var checkNumericRegex = new RegExp(/^(-)?[0-9]*(,|\.)[0-9]*$/);
+                var isNumericRegex = new RegExp(/^(-)?[0-9]*(\.)[0-9]*$/);
```
gabrieljenik, line 1236: Same as in the other branch. Maybe create a function? Up to you :) Let me know. thanks!

Shnoulle: Remove unrelated fix :)

PR #2307, opened 2022-03-19
New feature #13815: Allowing to change language before log in

arsieg:
New feature #13815: Allowing to change language before log in

### DIFF
--- a/application/views/admin/authentication/login.php
+++ b/application/views/admin/authentication/login.php
@@ -107,6 +107,9 @@ echo viewHelper::getViewTestTag('login');
                                     'name' => 'loginlang',
                                     'data' => $languageData,
                                     'value' => $language,
+                                    'events' => array(
+                                        'change' => 'function() {this.form.submit();}'
+                                    ),
                                     'pluginOptions' => array(
                                         'htmlOptions' => array(
                                             'id' => 'loginlang'


Shnoulle:
My opinion : when changing language you must send another params, to disable the error.

Capture d’écran du 2022-03-20 11-39-20

Think best is to add a submit button with value="changelang", and allow user to click on it.

And don't add an autosubmit : you can loose current  value entered.

1. Enter username
2. Change language (auto submit)
3. loose username entered.

arsieg:
Then maybe necessary to add id to Login button and then: 
'events' => array(
    'change' => 'function() {$("#login_submit").click();}'
 ),
 If your name and pass are correct, then form logs in.
 If not, they will not be kept anyway.

application/views/admin/authentication/login.php
```diff
@@ -151,6 +154,7 @@
 <!-- Set focus on user input -->
 <script type='text/javascript'>
 $( document ).ready(function() {
+    $('#user').val('<?php echo addslashes($_POST['user'])?>');
```
Shnoulle, line 157: 1. Never us $_POST : https://www.yiiframework.com/doc/api/1.1/CHttpRequest#getPost-detail
2. Why in JS ? Add it directly in PHP 
3. this don't fix «when changing language you must send another params, to disable the error.»

arsieg: Removed from JS, added directly in PHP using getPost().
Changing language does not pop up this alert. No need to disable it.
Alert pops up when you press 'Log in' button and name or password is incorrect.

Shnoulle: > Alert pops up when you press 'Log in' button and name or password is incorrect.

Or is both empty , i just check

arsieg: Yes, if both empty then also pops up (empty is incorrect)
But this is not related with changing language. 
Changing language works without pressing 'Log in' button.

Shnoulle: Ah right …

application/core/plugins/Authdb/Authdb.php
```diff
@@ -125,7 +125,7 @@ public function newLoginForm()
         }
 
         $this->getEvent()->getContent($this)
-                ->addContent(CHtml::tag('span', array(), "<label for='user'>" . gT("Username") . "</label>" . CHtml::textField('user', $sUserName, array('size' => 240, 'maxlength' => 240, 'class' => "form-control"))))
+                ->addContent(CHtml::tag('span', array(), "<label for='user'>" . gT("Username") . "</label>" . CHtml::textField('user', $sUserName ? : App()->request->getPost('user'), array('size' => 240, 'maxlength' => 240, 'class' => "form-control"))))
```
Shnoulle, line 128: Fill $sUserName in plugin

application/core/plugins/Authdb/Authdb.php
```diff
@@ -125,7 +125,7 @@ public function newLoginForm()
...
                 ->addContent(CHtml::tag('span', array(), "<label for='password'>" . gT("Password") . "</label>" . CHtml::passwordField('password', $sPassword, array('size' => 240, 'maxlength' => 240, 'class' => "form-control"))));
```
Shnoulle, line 129: Fill $sPassword in plugin event

application/views/admin/authentication/login.php
```diff
@@ -107,6 +107,9 @@
                                     'name' => 'loginlang',
                                     'data' => $languageData,
                                     'value' => $language,
+                                    'events' => array(
+                                        'change' => 'function() {this.form.submit();}'
```
Shnoulle, line 111: And for other plugin ?
And it's better if it can be easily remplaced by plugin OR Admin theme.

It must be in external js file.

Shnoulle (changes requested):
1. `$sUserName` and `$sPassword` must be prefilled in https://github.com/LimeSurvey/LimeSurvey/blob/3745f544e1dfeb01194a12cda5c3ae1481ef6e27/application/core/plugins/Authdb/Authdb.php#L118
2. JS action must be only for Authdb and came from an external file

application/core/plugins/Authdb/Authdb.php
arsieg: Password fill added

Shnoulle: No : fill it here : https://github.com/LimeSurvey/LimeSurvey/blob/3745f544e1dfeb01194a12cda5c3ae1481ef6e27/application/core/plugins/Authdb/Authdb.php#L118

PR #2312, opened 2022-03-24
Fixed issue #17458: Check integrity broke with SQL error

Shnoulle:
Dev: add a try/cast
Dev: happen if subquestion title are number only


Shnoulle:
Same fix than master (to close the mantis issue)

PR #2313, opened 2022-03-24
Fixed issue #17458: Check integrty broke for subquestion with only number

Shnoulle:
Dev: Try cast and don't throw error


Shnoulle:
@olleharstedt issue introduce in https://github.com/LimeSurvey/LimeSurvey/commit/6a829601ce2e58160f47c965af8993c20c31598c#r49200600 

Since it can happen with clean survey …

olleharstedt:
OK, looks good.

PR #2315, opened 2022-03-25
Fixed issue #17966: twig processString assumes wrong question Id

Shnoulle:
Fixed issue #17967: Group description are not updated by javascript

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -420,7 +420,7 @@ class SurveyRuntimeHelper
                 $aGroup['class'] = ' ls-hidden';
             }
 
-            $aGroup['name']        = LimeExpressionManager::ProcessString($gl['group_name'], null, null, 3, 1);
+            $aGroup['name']        = $gl['group_name'];
             LimeExpressionManager::updateReplacementFields(array(
                 'GROUPNAME' => $aGroup['name'],
             ));
@@ -439,7 +439,7 @@ class SurveyRuntimeHelper
 
             $aGroup['showgroupinfo'] = $showgroupinfo_;
             $aGroup['showdescription']  = (!$this->previewquestion && trim($gl['description']) != "" && $showgroupdesc_);
-            $aGroup['description']      = LimeExpressionManager::ProcessString($gl['description'], null, null, 3, 1);
+            $aGroup['description']      = $gl['description'];
 
             // one entry per QID
             foreach ($qanda as $qa) {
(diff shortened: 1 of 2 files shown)


Shnoulle:
Oh, i think you ask for a test :)

olleharstedt:
For everything touching EM, a test would be very nice, yes. :) Because of the complexity of that system.

Shnoulle:
Yes, but before adding a test, i like a review … to be sure to don't do the test without need …

olleharstedt:
Alright.

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -439,7 +439,7 @@ public function run($surveyid, $args)
 
             $aGroup['showgroupinfo'] = $showgroupinfo_;
             $aGroup['showdescription']  = (!$this->previewquestion && trim($gl['description']) != "" && $showgroupdesc_);
-            $aGroup['description']      = LimeExpressionManager::ProcessString($gl['description'], null, null, 3, 1);
+            $aGroup['description']      = $gl['description'];
```
gabrieljenik, line 442: Just over viewing this.
Shouldn't we add description to the `LimeExpressionManager::updateReplacementFields`?

Shnoulle: I test with , it can be done
But looking at manual : it's not included.

Seems is lost in 3.X since 3.0.

Then it's a new feature :)

application/helpers/expressions/em_manager_helper.php
```diff
@@ -6770,7 +6790,7 @@ public static function StartProcessingGroup($gseq = null, $anonymized = false, $
         $LEM->groupRelevanceInfo = [];
         if (!is_null($gseq)) {
             $LEM->currentGroupSeq = $gseq;
-
+            self::InitGroupRelevanceInfo();
```
gabrieljenik, line 6793: Is it correct to add this here taking into accoutn the StartProcessingGroup may have added asomething already? (haven't looked into much detail really)

Shnoulle: You have this line https://github.com/Shnoulle/LimeSurvey-Shnoulle/blob/1b3fa0255eaf8ba1c5d75dc1e06c69e490a69423/application/helpers/expressions/em_manager_helper.php#L4761 each time i set it agian.

And more : you NEED some not related to question id : group description are used a lot with Expresssion.

gabrieljenik: Sorry, not that into EM to follow

PR #2317, opened 2022-03-29
Fixed issue #17902: Question theme error: Missing metadata (Type: M)

gabrieljenik:


Shnoulle:
Something i don't understand : we need or don't need metadata type ?

Here : still no `type` metadata ? Right ?

application/config/version.php
```diff
@@ -12,7 +12,7 @@
  */
 
 $config['versionnumber'] = '5.3.7';
-$config['dbversionnumber'] = 485;
+$config['dbversionnumber'] = 486;
```
olleharstedt, line 15: Missing the update file?

application/helpers/update/updates/Update_486.php
```diff
@@ -0,0 +1,20 @@
...
+
+class Update_486 extends DatabaseUpdateBase
+{
+    public function up()
+    {
+        $this->db->createCommand()->update(
+            "{{question_themes}}",
+            [
+                'xml_path' => 'themes/question/bootstrap_buttons_multi/survey/questions/answer/multiplechoice',
+                'image_path' => '/themes/question/bootstrap_buttons_multi/survey/questions/answer/multiplechoice/assets/bootstrap_buttons_multiplechoice.png',
```
olleharstedt, line 15: That's correct, xml_path starts with letter, image_path starts with slash?

gabrieljenik: shouldn't we just straitght that (in another story)?
Maybe add a function which will allow any kind of path and then converted to the required type of path?

gabrieljenik:
> Something i don't understand : we need or don't need metadata type ?
> 
> Here : still no `type` metadata ? Right ?

That's here https://github.com/LimeSurvey/LimeSurvey/pull/2271

Shnoulle:
> > Something i don't understand : we need or don't need metadata type ?
> > Here : still no `type` metadata ? Right ?
> 
> That's here #2271

But not there ;: https://github.com/LimeSurvey/LimeSurvey/blob/256ad585f78f2cbd044ea667386c3800faa7374f/themes/question/bootstrap_buttons_multi/survey/questions/answer/multiplechoice/config.xml

You're fixing [17902 Question theme error: Missing metadata (Type: M)](https://bugs.limesurvey.org/view.php?id=17902) 
or another issue ?

gabrieljenik:
> You're fixing 17902 Question theme error: Missing metadata (Type: M) or another issue ?
I am fixing 17902 here, that to 2271 PR, although it's connected to bootstrap buttons, was done before and it was not related specifically to this ticket.

olleharstedt:
Feel free to merge when CI is green

PR #2319, opened 2022-04-01
Fixed issue #18006: Unable to use statCountIf to have a "unicity" control

Shnoulle:
Dev: add optional self parameters to false (no function API update)


gabrieljenik (approved):

PR #2326, opened 2022-04-04
Fixed issue #18008: Broken condition using self.sgqa (static)

Shnoulle:
Fixed issue #18012: Usage of static suffix in condition can broke JS
Dev: return static var for JS using aRDP_regexpStaticAttribute


gabrieljenik (approved):

PR #2327, opened 2022-04-04
New feature #18017: Have a solution to force expression as static

Shnoulle:
Add a getStatic to force any expression as static var

### DIFF
--- a/application/core/plugins/FunctionStatic/FunctionStatic.php
+++ b/application/core/plugins/FunctionStatic/FunctionStatic.php
@@ -1,0 +1,99 @@
+<?php
+
+/**
+ * @author Denis Chenu <denis@sondages.pro>
+ * @copyright 2022 Denis Chenu <https://www.sondages.pro>
+ * @license GPL version 3
+ * @version 0.1.0
+ *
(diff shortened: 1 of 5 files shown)


application/core/plugins/FunctionStatic/FunctionStatic.php
```diff
@@ -0,0 +1,99 @@
...
+            ),
+        );
+        $this->getEvent()->append('functions', $newFunctions);
+    }
+
+    /**
+     * @inheritdoc
+     * Update the information content
+     */
+    public function getPluginSettings($getValues = true)
```
gabrieljenik, line 69: Why do you prefer these to using `protected $settings`?

gabrieljenik (approved):
Haven't tested it. Code looks OK.

application/core/plugins/FunctionStatic/FunctionStatic.php
Shnoulle: Here : twig + gt, i return HTML

I think twig is better to produce conetnet, and need a gT.

Not necessary

PR #2329, opened 2022-04-05
Fixed issue #17895: Error in multiple-numeric "Equals sum value" setting

Shnoulle:
Dev: remove default precision system (12 is really bigger than any float value)


gabrieljenik (approved):

PR #2330, opened 2022-04-05
Fixed issue #17895: Error in multiple-numeric "Equals sum value" setting

Shnoulle:
Dev: remove broken default round


Shnoulle:
Still not perfect : for remaining
Capture d’écran du 2022-04-05 15-27-47

gabrieljenik (approved):

PR #2334, opened 2022-04-06
Fixed issue #17972: events are dispatched if the plugin is deactivated

Shnoulle:
Dev: Add param to load init
Dev: set it to false in PluginManagerController in all case
Dev: active plugin are loaded via PluginManager->loadPlugins


application/libraries/PluginManager/PluginManager.php
```diff
@@ -83,7 +83,7 @@ public function init()
     }
     /**
      * Return a list of installed plugins, but only if the files are still there
-     *
+     * @deprecated unused in 5.3.8
```
gabrieljenik, line 86: Why do you mark it as deprecated?

gabrieljenik (approved):
Haven't tested it. Code looks good!

PR #2335, opened 2022-04-06
Fixed issue #17972: events are dispatched if the plugin is deactivated

Shnoulle:
Dev: Add param to load init
Dev: set it to false in PluginManagerController in all case
Dev: active plugin are loaded via PluginManager->loadPlugins
Dev: use plugin model active status in PluginManager


Shnoulle:
Same than merged in 5.X ;)

gabrieljenik (approved):
Code looks good. Haven't tested it

gabrieljenik:
Master PR has already been pushed. I think this could be as well, right?

olleharstedt:
> Master PR has already been pushed. I think this could be as well, right? @olleharstedt @c-schmitz ?

If it's the same patch? Sure.

PR #2336, opened 2022-04-07
Fixed issue #17811: Theme options for logo and background images are not saved

gabrieljenik:
- Using DIRECTORY_SEPARATOR.
- Making logo paths on base theme config.xml file to be local, instead of relative to the LS root


Shnoulle (approved):
Clearly needed !

PR #2343, opened 2022-04-12
Fixed issue #18019: 500 error in question edition (bad code on question)

gabrieljenik:
Adding an error handler for the case when the save is not done using ajax.
Although it now shows the error in an alert, its content is still ugly (the print_r is still there).

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -548,15 +548,36 @@ class QuestionAdministrationController extends LSBaseController
             }
         } catch (CException $ex) {
             $transaction->rollback();
-            throw new LSJsonException(
-                500,
-                gT('An error happened:') . "\n" . $ex->getMessage() . PHP_EOL,
-                0,
-                App()->createUrl(
-                    'surveyAdministration/view/',
-                    ["surveyid" => $iSurveyId]
-                )
-            );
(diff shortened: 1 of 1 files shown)


Shnoulle (approved):

Shnoulle:
Not totally related , but  https://github.com/LimeSurvey/LimeSurvey/blob/60d001798f3dabc13ed1700a906d74c62220d391/application/controllers/QuestionAdministrationController.php#L379 is badly set…

Any reason we don't use https://www.yiiframework.com/doc/api/1.1/CHttpRequest#getIsAjaxRequest-detail ?

gabrieljenik:
> is badly set…

Why badly set? Maybe should be cast to bool?

> Any reason we don't use https://www.yiiframework.com/doc/api/1.1/CHttpRequest#getIsAjaxRequest-detail ?

Not sure. It was like this before this fix. :)
But definitely different.
One is based on a form value.
The other is based on headers.

Shnoulle:
> Why badly set? Maybe should be cast to bool?

Because it's not because you have ajax=1 in POST it's really an ajax request :)

>  One is based on a form value.

Yes , then it's unsure it was really an ajax request (i think we have it somewhere when you can send ajax=1 bit it's not ajax).

PS : i approve the fix.

PR #2349, opened 2022-04-13
Fixed issue #17991: Importing a question from the "List questions" menu fails with 400 error

gabrieljenik:
This is just a workaround.

We still need to review if the change below (making the groupid mandatory) is still necesarry to review a more integral approach. https://github.com/LimeSurvey/LimeSurvey/commit/e877a504bcad71dbc0266d830acb0752f168f477#diff-8ba52d76c0ab5c25746214e40fd926cfdb0ae2024bd7bcdd90f1035014f5be12R141

If not necessary, the urls that point to the screen may need some reviewal in the order of the parameters and the ones being sent.

### DIFF
--- a/application/controllers/admin/questions.php
+++ b/application/controllers/admin/questions.php
@@ -141,7 +141,7 @@ class questions extends Survey_Common_Action
     /**
      * Display import view
      */
-    public function importView($groupid , $surveyid)
+    public function importView($surveyid, $groupid = null)
     {
         $iSurveyID = (int) $surveyid;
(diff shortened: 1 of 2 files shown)


Shnoulle:
`public function importView($groupid = 0, $surveyid)`

work without issue, seem better.

And the best :  `public function importView($surveyid, $groupid = null)` if it can be done without updating other function

Shnoulle (changes requested):
Seem better to update function to allow empty groupid

Shnoulle:
Sorry for delay …

PR #2352, opened 2022-04-14
Fixed issue #17648: Email received by participant is empty (LDAP)

Shnoulle:
Fixed issue #18034: Date from token with invalid language show error and date not saved
Dev: LimeMailer already fix language for automatic raw content
Dev: forced raw content (from POST) are not fixed

Dev:

### DIFF
--- a/application/controllers/admin/Tokens.php
+++ b/application/controllers/admin/Tokens.php
@@ -1399,9 +1399,12 @@ class Tokens extends SurveyCommonAction
                 $mail->emailType = $sSubAction;
                 $mail->replaceTokenAttributes = true;
                 foreach ($emresult as $emrow) {
-                    if (empty($emrow['language'])) {
-                        $emrow['language'] = $sBaseLanguage;
+                    $mailLanguage = $emrow['language'];
+                    if (!in_array($mailLanguage, Survey::model()->findByPk($iSurveyId)->getAllLanguages())) {
+                        $mailLanguage = $sBaseLanguage;
+                        $tokenoutput .= CHtml::tag("div", array('class' => 'text-warning'), sprintf(gT("Invalid language %s for token ID: %s."), $emrow['language'], $emrow['tid']));
                     }
+
                     $mail = \LimeMailer::getInstance();
                     if ($this->tokenIsSetInEmailCache($iSurveyId, $emrow['tid'], $bIsInvitation)) {
                         // The email has already been send this session, skip.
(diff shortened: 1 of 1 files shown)


olleharstedt:
LDAP... Who's testing LDAP?

Shnoulle:
No LDAP : when LDAP import is done : language can be false
Check the issue 18034 too.

Shnoulle:
PS : i can create and check the old LDAP import too.
Here : i fix the send email only :)

application/controllers/admin/Tokens.php
```diff
@@ -1438,7 +1441,16 @@ public function email(int $iSurveyId)
                                 $oToken->remindercount++;
                             }
                             $tokenSaveError = "";
-                            if (!$oToken->encryptSave(true)) {
+                            if (
```
olleharstedt, line 1444: No encryption needed?

Shnoulle: We save only not encrypted attributes, then no encryption needed :) 

Interesting because we don't update other attribute (we don't encrypt it)

application/controllers/admin/Tokens.php
```diff
@@ -1438,7 +1441,16 @@ public function email(int $iSurveyId)
...
+                                !$oToken->save(
+                                    true,
+                                    array(
```
Shnoulle, line 1447: We save only this attribute in the request https://github.com/yiisoft/yii/blob/b031205cf4ec20131639653910a13ca61747da32/framework/db/ar/CActiveRecord.php#L1121

olleharstedt: Got it

Shnoulle:
Maybe another way to test  : 

Create a multilingual survey, french for example
Import or create Token with French language
Delete the french language from survey
Send email

olleharstedt:
Big conflict in Tokens.php :(

Shnoulle:
> Big conflict in Tokens.php :(

Yes … commit from April … 6 month ago …

It can be tested checked and tested without merging

olleharstedt:
Code looks fine. Who's the customer who reported it? Can they test?

Shnoulle:
> Code looks fine. Who's the customer who reported it? Can they test?

https://forums.limesurvey.org/forum/french-forum/125314-mail-d-invitation-vide#219607

6 month after : they fix themself the issue …

OK , i merge …

olleharstedt:
> 6 month after : they fix themself the issue …

By same code as in this PR?

Shnoulle:
> > 6 month after : they fix themself the issue …
> 
> By same code as in this PR?

By fix the token language value « J'ai renseigné le champ langue avec fr pour tous les participants et toutes les invitations ont été correctement envoyées.» 

I check if with a lsa impor we can reproduce the empty language value.

Shnoulle:
Follow up on https://github.com/LimeSurvey/LimeSurvey/pull/2691

PR #2355, opened 2022-04-16
 Fixed issue #18035: Conditions does not work for all in one surveys 

Shnoulle:
Dev: add not related to question relevance js only one time
Dev: disable Group always relevant for qid==0
Dev: TODO test

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -4754,7 +4754,6 @@ class LimeExpressionManager
                     if ($LEM->currentGroupSeq > $LEM->maxGroupSeq) { // Did we need it ?
                         $LEM->maxGroupSeq = $LEM->currentGroupSeq;
                     }
-                    self::InitGroupRelevanceInfo();
                     $LEM->ProcessAllNeededRelevance($LEM->currentQuestionSeq);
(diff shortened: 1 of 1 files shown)


Shnoulle:
@gabrieljenik i set it to draft, it's ready for commit but NEED a test survey (2 maybe)
1. one with group title update
2. one with all in one and survey

application/helpers/expressions/em_manager_helper.php
```diff
@@ -6985,15 +6953,26 @@ public static function GetRelevanceAndTailoringJavaScript($bReturnArray = false)
                     }
                 }
             }
+            $pageRelevanceInfo[] = array(
```
Shnoulle, line 6956: the «global» page relevance info is set at end with last group seq. Maybe out of group seq too ?

gabrieljenik: Why is this record needed? It is not related to any question id. Is it for handling EM at group level?

Shnoulle:
3 test done, lot of assertions.
Have some other issue to report

c-schmitz:
I would need the patch for 3.x also, urgently.

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2367

Shnoulle:
Oups : https://github.com/LimeSurvey/LimeSurvey/pull/2366

PR #2357, opened 2022-04-19
FIX: 17881: Question code / ID on quick translation screen showing question text not ID making it impossible to translate surveys

gabrieljenik:


glimz:
The quick translation tool is still not working correctly ... the Question and Sub Question Tabs are now showing the Question ID correctly BUT the answer options tab is still replicating the English Text in the column for the Question ID

PR #2359, opened 2022-04-21
Fixed issue #17300: better show selected option difference

Shnoulle:
Just really show a difference … not only via color.
I use it since month now, and clearly better 

Capture d’écran du 2022-04-21 12-45-49


glimz:
This is nice, but I think there should be a design decision. The improvement is a huge change to all buttons. @c-schmitz what do you think?

Shnoulle:
It can be done with a [plugin](https://gitlab.com/SondagesPro/coreAndTools/select2ForAdmin) or [AdminTheme](https://gitlab.com/SondagesPro/sondagesme/SondagesPro)  , but since it still really unclear if a value is checked or not …

gabrieljenik (approved):
Only CSS changes.

Shnoulle:
> This is nice, but I think there should be a design decision. The improvement is a huge change to all buttons. @c-schmitz what do you think?

> Only CSS changes.

No code issue (just to redo whole) …

PR #2360, opened 2022-04-21
Fixed issue #17488: Cannot view survey after updating to version 3.27.9 with custom twig extension (breaks survey)

gabrieljenik:


Shnoulle (approved):
seems great ! :+1:

PR #2361, opened 2022-04-21
Fixed issue #17900: Response date is incorrect when there is a mis-match in date format

gabrieljenik:


gabrieljenik:
Tested it. Pending to be merged

Shnoulle (approved):

PR #2362, opened 2022-04-21
Fixed issue #17994: Comments at question type "List with comments" are missing at print answer overview

gabrieljenik:


maziminke:
@glimz Can you review so we can finally bring this into the next 3.x and 5.x release?

c-schmitz:
@maziminke You can speed things up if you test this patch for functionality and let us know if it works alright.

PR #2367, opened 2022-04-23
Fixed issue #18047: Conditional filters attachments for confirmation mail doesn't work

Shnoulle:
Dev: adding lost checking of relevance …
Dev: clearly view condition don't need {}


Shnoulle:
> Dev: clearly view condition don't need {}

Added because i test test test again with `{TOKEN:ATTRIBUTE_1}` (and not `TOKEN:ATTRIBUTE_1`) and never get the file …

c-schmitz (approved):

Shnoulle:
Tested by reporter : https://bugs.limesurvey.org/view.php?id=18047#c69317

PR #2378, opened 2022-04-28
 Fixed issue #17980: Txt Import Crashing with just exported survey 

Shnoulle:
Dev: Integrity constraint violation: 1062 Duplicate entry
Dev: exported survey have same qid for different question, check if not exist
Dev: need warning ?
Dev: add rules on QuestionL10n and AnswerL10n model (unicity of qid/language)

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -2809,7 +2809,7 @@ function TSVImportSurvey($sFullFilePath)
                     $group['group_order'] = $ginfo[$sGroupseq]['group_order'];
                 } else {
                     /* Get the new gid from file if it's number and not already set*/
-                    if (!empty($row['id']) && ctype_digit($row['id']) && !in_array($row['id'],$groupIds)) {
+                    if (!empty($row['id']) && ctype_digit($row['id']) && !in_array($row['id'], $groupIds)) {
                         $gid = $row['id'];
                     } else {
                         $gidNew += 1;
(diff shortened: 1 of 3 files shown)


gabrieljenik:
This needs specific testing of multi-language in both questions and answer

Shnoulle:
> This needs specific testing of multi-language in both questions and answer

I didn't fix the export part here. Only included lss in bug report.

Seems the original issue was in export part …

Shnoulle:
> > It's not related to multilanguage survey
> 
> I was mentioning about this https://github.com/LimeSurvey/LimeSurvey/pull/2378/files#diff-ecb7e30aaef42eddc7292256c9667b199d1f4ccc0b0b33a70e833c2177cc4bc4R70

See https://github.com/LimeSurvey/LimeSurvey/blob/f10aa458beb1edf3a6c2140cba56141a10fa0869/application/models/Question.php#L101
and
https://github.com/LimeSurvey/LimeSurvey/blob/f10aa458beb1edf3a6c2140cba56141a10fa0869/application/models/Answer.php#L79

for 3.X :)

I have a mantis issue in dev part with this …

Shnoulle:
> I am sorry. I don't follow. You added some rules which relates to multiple languages. As that, I am commenting this should be tested with multiple languages.

You're right … i added this rules when testing, to see what happen … and don't remove it because it's needed.
But still not related to multilingual :) 

related to existing index : 

- https://github.com/LimeSurvey/LimeSurvey/blob/79d79294805fa3361bd53265a56d7c4d0ae262d7/installer/create-database.php#L61
- https://github.com/LimeSurvey/LimeSurvey/blob/79d79294805fa3361bd53265a56d7c4d0ae262d7/installer/create-database.php#L403

Finally we need this rules for all i10n ;) 
To allow usage of ->save without send a SQL error :)

gabrieljenik:
> But still not related to multilingual :)

Well, the rule is using language, right?
That's why I was mentioning about the multi-lingual

PR #2379, opened 2022-04-28
Dev #18040: Add buttons submit the whole labelset instead of only add…

gabrieljenik:


gabrieljenik:
This is ready for reviewal.

application/controllers/admin/Labels.php
```diff
@@ -740,4 +741,64 @@ protected function renderWrappedTemplate($sAction = 'labels', $aViewUrls = array
...
+     * Called by ajax from the label set editor.
+     * @param int $lid
+     * @param string $newId
+     * @param string $code
+     * @param int $assessmentValue
+     * @param string $title
+     */
+    public function getLabelRowForAllLanguages($lid, $newId, $code, $assessmentValue = 0, $title = '')
+    {
+        if (!Permission::model()->hasGlobalPermission('labelsets', 'update')) {
```
olleharstedt, line 757: Read permission should be enough, if there's no write in the method?

assets/scripts/admin/labels.js
```diff
@@ -227,60 +244,75 @@ function get_next_code(event) {
...
+            // arrayofhtml is a json object containing the different HTML row by language
+            // eg: {"en":"{the html of the en row}", "fr":{the html of the fr row}}
+            $defer.resolve({ arrayofhtml: arrayofhtml });
+        },
+        error(html, status) {
+            alert('Internal error retrieving row html: ' + errormessage);
+            $defer.reject([html, status, errormessage]);
+        },
+    });
+    return $defer.promise();
```
olleharstedt, line 273: I think you can just `return $ajax({ ...` too

application/controllers/admin/Labels.php
gabrieljenik: This is because this function is called in the process of editing labels.

olleharstedt: You shouldn't lock a function to its context, tho, you might prevent reuse that way.

gabrieljenik: But this is an action in the controller. Not expecting to reuse this.

Shnoulle: > But this is an action in the controller. Not expecting to reuse this.


You can get the json just to show it to user . It's a public function. 
Why disallow show lable if user have read rights ?

This public function didn't update label : then need only read.

olleharstedt: You're not expecting to reuse it _now_, but who knows what will happen 2-5 years from now.

gabrieljenik: Do you expect to reuse controller actions?
Anyway, already changed the permission to read :)

olleharstedt: "getLabelRowForAllLanguages" sounds like something that could be used in, say, question editor. Na ja. Thanks!

application/controllers/admin/Labels.php
```diff
@@ -740,4 +741,64 @@ protected function renderWrappedTemplate($sAction = 'labels', $aViewUrls = array
...
+     * @param string $title
+     * @return string   The html of the row
+     */
+    private function getLabelRow($language, $first, $newId, $code, $assessmentValue, $title)
+    {
+        $view = 'labelRow.twig';
+        $aData = array(
+            'language' => $language,
+            'first' => $first,
+            'sortOrder' => $newId,
```
olleharstedt, line 792: sort order = new id? Variable naming mismatch?

olleharstedt:
Meh. The broken CI comes from dev branch, right?

gabrieljenik:
> 25 changed files?

Ooops. Files from T206 got into here. Will clean up

gabrieljenik:
Conflicts solved. Ready for re-re-review.

Shnoulle:
re-re-re-re-re-re-re-re-review ;)

olleharstedt:
Dev branch should be green now, can you merge upstream?

gabrieljenik:
> can you merge upstream

What do you mean? Sorry, didn't get it. From where to where?

olleharstedt:
Merge develop branch into this again, to get a green CI.

gabrieljenik:
> Merge develop branch into this again, to get a green CI.

What about now?

application/controllers/admin/Labels.php
```diff
@@ -737,4 +738,64 @@ protected function renderWrappedTemplate($sAction = 'labels', $aViewUrls = array
...
+     * @param string $title
+     * @return string   The html of the row
+     */
+    private function getLabelRow($language, $first, $newId, $code, $assessmentValue, $title)
+    {
+        $view = 'labelRow.twig';
+        $aData = array(
+            'language' => $language,
+            'first' => $first,
+            'rowId' => $newId,
```
olleharstedt, line 789: rowId or newId?

assets/scripts/admin/labels.js
```diff
@@ -110,6 +111,9 @@ function quickaddfunction() {
     }
 
     console.log('lsrows', lsrows);
+
+    // parsedRows is an array of objects with the parsed row data (ie: [{code:"code", titles:{"en":"Title"}}])
+    let parsedRows = [];
```
olleharstedt, line 116: Nice with example data in comment. Maybe we can even create types here one day (TypeScript or Flow, for example).

application/controllers/admin/Labels.php
gabrieljenik: $newId is the name of the argument. Should we change it you think?

gabrieljenik:
The "need update by author" is because of this?
https://github.com/LimeSurvey/LimeSurvey/pull/2379#discussion_r1093453153

Not sure what am I suppose to focus on :)

PR #2384, opened 2022-04-30
Fixed issue #18066: RPC method `upload_file` does not honor `$sFileName` argument in response

edgarrmondragon:
Fixed issue [#18066](https://bugs.limesurvey.org/view.php?id=18066): Method upload_file does not honor $FileName argument in response


c-schmitz:
Thank you!

PR #2386, opened 2022-05-02
Fixed issue #18060: 'Save as default values' question setting not working properly

gabrieljenik:
1) The value of the front_end switch depended on use_defaults, which has nothing to do with it. Now it always appears disabled, as this a one way switch: You can only set it to true and then defaults will be snapshot and saved. There is no link in between the default values sanpshot and the current question attributes. Question attributes can later be changed, and the default wouldn't be impacted.

2) Defaults were saved but not used for populating new questions. Now they do.

### DIFF
--- a/application/extensions/GeneralOptionWidget/settings/SaveAsDefaultGeneralOption.php
+++ b/application/extensions/GeneralOptionWidget/settings/SaveAsDefaultGeneralOption.php
@@ -19,7 +19,7 @@ class SaveAsDefaultGeneralOption extends GeneralOption
             'save_as_default',
             null,
             gT('All attribute values for this question type will be saved as default'),
-            $question->same_default == 1 ? 'Y' : 'N',
+            'N',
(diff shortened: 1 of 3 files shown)


Shnoulle:
> Defaults were saved but not used for populating new questions. Now they do.

We can not accept anyone set the default value for all survey of anybody …

gabrieljenik:
> We can not accept anyone set the default value for all survey of anybody …

They are saved under personal settings.

Shnoulle:
> > We can not accept anyone set the default value for all survey of anybody …
> 
> They are saved under personal settings.

Not tested, sorry !!!

c-schmitz:
@gabrieljenik 
Just tested. How are the default values populated on a new question?
I mean, you would have to select the question type first  when you create a new question for that to work, woudn't you?
Because on question type change you usuall want to preserve the settings.

PR #2387, opened 2022-05-02
Fixed issue #16907: Importing Question through Remote Control doesn't set new title for all languages

gabrieljenik:
There are several problems with importing questions.

1) The new "title" is only set in the first language (it is the problem of the ticket).
2) The new question text is also set to the first language only.
3) The new "title" is set after importing. In other words, if you want to import a question, whose title already exists, but set it at the moment (through the API call), you can't, because before setting the new one, the process will fail because of the duplication.

This particular ticket is simple, because the title has to always be the same. The text of the question is more complicated because setting the same text in all languages does not make sense. This ticket is only limited to item 1.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1457,7 +1457,7 @@ class remotecontrol_handle
                     fixLanguageConsistency($iSurveyID);
                     $iNewqid = $aImportResults['newqid'];
 
-                    $oQuestion = Question::model()->findByAttributes(array('sid' => $iSurveyID, 'gid' => $iGroupID, 'qid' => $iNewqid));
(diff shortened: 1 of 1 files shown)


Shnoulle:
I completely rewrite all system on https://github.com/LimeSurvey/LimeSurvey/pull/2405

And adding potential issue with invalid title ( try to set title to `1234` for example : erreo happen but  nit shown)

Shnoulle:
We can not merge both : https://github.com/LimeSurvey/LimeSurvey/pull/2405/files#diff-22d09ac4b2de7e29721c3fc14eb239f215498a280b8c37f4b718074646f01f5dR1482

Shnoulle (changes requested):
The https://github.com/LimeSurvey/LimeSurvey/pull/2405 fix it too and 2 other bug.

OK to close this one ?

gabrieljenik:
> The #2405 fix it too and 2 other bug.
> 
> OK to close this one ?

Sure, go ahead!

PR #2388, opened 2022-05-03
Fixed issue #18063: Expression manager show false error when editing question

Shnoulle:
Dev: SetDirtyFlag for EM after save question by controller


Shnoulle:
Other solution : LimeExpressionManager::SetDirtyFlag(); in model after save ?

olleharstedt:
Small and tidy fix :)

PR #2390, opened 2022-05-03
Fixd issue #18031: Broken theme broke with 500 error

Shnoulle:
Partial need a way to inform user there are issue with deleted them


Shnoulle:
Don't close related issue, there still more work : 

1. check if theme is broken (and show it somewhere), 
2. add a forced uninstall

PR #2391, opened 2022-05-04
Lsc 30 bs06 grid system

ptelu:


application/views/admin/update/_updateContainer.php
```diff
@@ -16,40 +16,38 @@
...
-        <?php if(YII_DEBUG):?>
-            <small>Server:<em class="text-warning"> <?php echo Yii::app()->getConfig("comfort_update_server_url");?></em></small>
-        <?php endif;?>
-    </h3>
+        <div class="row">
+            <div class="col-lg-12 content-right">
+                <div id="updaterWrap">
+                    <div id="preUpdaterContainer">
+                        <!-- The check buttons : render by ajax only if no step is required by url or post -->
+                        <?php // $this->renderPartial("./update/check_updates/_checkButtons", array( "thisupdatecheckperiod"=>$thisupdatecheckperiod, "updatelastcheck"=>$updatelastcheck,"UpdateNotificationForBranch"=>$UpdateNotificationForBranch )); ?>
```
thedirtypanda, line 35: Is this an left over? If this would be an left over, please remove it. Otherwise write it into the comment, that you should enable this (if its needed). Otherwise it looks like disabled code.

application/views/userManagement/index.php
```diff
@@ -21,7 +21,6 @@
 
 <?php $this->renderPartial('partial/_menubar'); ?>
 
-<div class="row">
     <?php
```
thedirtypanda: Do we need such comments like "Example: Datepicker etc..." inside the codebase?

thedirtypanda:
Code looks fine!

application/views/userManagement/index.php
ptelu: no we don't, but not really part of the task to remove it. Probably would have removed it if i saw it :)

application/views/admin/update/_updateContainer.php
ptelu: I can check what this is, but haven't touched that part of the code

olleharstedt:
@ptelu We can close this one?

ptelu:
> @ptelu We can close this one?
yeah

PR #2393, opened 2022-05-04
Fixed issue #17987: disable XSS on equation attribute

Shnoulle:


gabrieljenik (approved):

Shnoulle:
Micro fix … and since this was not in any view, and since it was saved in DB with lot of filer …

But make equation using if condition really more hard …

Shnoulle:
> I don't recall the attribute that well, but would this allow any ser (not only superamdins) to set mailicious html?

Oh !You're right !
I think it was encdoded, but it's false :)

PR #2396, opened 2022-05-04
Fixed issue #17987 : quota url need to be decoded

Shnoulle:
Dev: same decode than survey url


Shnoulle:
Alternative solution : decode in https://github.com/LimeSurvey/LimeSurvey/blob/8d8420a4efcf8e71c4fccbb6708648ace263ca80/application/core/LSYii_Validators.php#L63
But then : no XSS can broke ? Unsure on the good fix …

gabrieljenik (approved):

PR #2398, opened 2022-05-05
Fixed Issue: #18029 Removed html tags in emails

fsay2604:
Fixed issue #<Mantis issue number>: 18029
Dev:
Dev: CKEditor configuration file must enable fullPage config in order to keep every html tags.

### DIFF
--- a/assets/packages/ckeditor/config.js
+++ b/assets/packages/ckeditor/config.js
@@ -21,6 +21,7 @@ CKEDITOR.editorConfig = function (a) {
         a.basicEntities = !1;
         a.entities = !1;
         a.uiColor = "#f1f1f1";
+        a.fullPage = true;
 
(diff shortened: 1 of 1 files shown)


c-schmitz:
Contrary to the PR description I can only see that the Fullpage attribute is set and that is the onyl change. Any reason?

fsay2604:
Since AllowedContent was already set to true, I did not have to change the setting. I forgot to remove the line in the description. 

Also, I see that this patch doesn't pass the code analysis, probably because those settings affect all ckeditors instance in the app, particularly the one used to create questions. The setting `FullPage=true` add HTML, Head, Body tags if they are missing, but the behavior we are looking for is to keep the tags if they are present, and not adding them if they are not.

I will probably submit another PR that will fix that later. (Either by changing setting for CKEditor instance only for the one used in email templates, or by finding another workaround.)

c-schmitz:
Duplicate of #2411

PR #2399, opened 2022-05-05
Fixed Issue: #18029 Removed html tags in emails

fsay2604:
Fixed issue #<Mantis issue number>: 18029
Dev:
Dev: CKEditor configuration file must enable fullPage config in order to keep every html tags.
Dev: CKEditor configuration file must enable allowedContent config so the contentFiltering is disabled.


c-schmitz:
There is something very wrong with this patch. Can you try to submit a new clean PR based on the latest revision of the master branch? Otherwise it is imposssible to review or merge

PR #2404, opened 2022-05-05
Fixed issue #17992: unable to reimport tokens from old table

gabrieljenik:


Shnoulle:
We can not do a find , and if null return json empty array (or [unknow])

Shnoulle:
We can not do a find , and if null return json empty array (or [unknow])

gabrieljenik:
Not sure I follow. An update was missing for some users.
So querying for the missing column.
If actually missing, we add it.

Shnoulle:
Oh yes, better look : lack of a column.

We do the `foreach ($archivedTableSettings as $archivedTableSetting) {` in original push too ?

gabrieljenik:
Yes, that is kept the same as it was on the original db update.

Shnoulle (approved):

Shnoulle (approved):

Shnoulle (approved):

Shnoulle (approved):

Shnoulle:
@gabrieljenik since i don't see any potential issue, please fix the conflict and merge :)

Shnoulle (approved):

gabrieljenik:
@c_schmitz, this seem a major impact stuff. Should we fix (again) and merge?

application/helpers/update/updates/Update_489.php
```diff
@@ -0,0 +1,23 @@
...
+class Update_489 extends DatabaseUpdateBase
+{
+    public function up()
+    {
+        $columnSchema = $this->db->getSchema()->getTable('{{archived_table_settings}}')->getColumn('attributes');
+        if ($columnSchema === null) {
+            $this->db->createCommand()->addColumn('{{archived_table_settings}}', 'attributes', 'text NULL');
+            $archivedTableSettings = \Yii::app()->db->createCommand("SELECT * FROM {{archived_table_settings}}")->queryAll();
+            foreach ($archivedTableSettings as $archivedTableSetting) {
+                if ($archivedTableSetting['tbl_type'] === 'token') {
```
c-schmitz, line 17: Why not put this condition in the query?

application/helpers/update/updates/Update_489.php
```diff
@@ -0,0 +1,23 @@
...
+                    $this->db->createCommand()->update('{{archived_table_settings}}', ['attributes' => json_encode(['unknown'])], 'id = :id', ['id' => $archivedTableSetting['id']]);
```
c-schmitz, line 18: Why is the field 'attributes' filled with the value 'unknown' ? Is this value actually used somewhere?

c-schmitz: To be able to fix this for 5.x and 6.x this needs to be put into different DBVersions.
Could you actually see that this was a cause by a bug we released in the past?

c-schmitz (changes requested):

application/helpers/update/updates/Update_489.php
gabrieljenik: Let me recall about this and be back :) Thanks!

application/helpers/update/updates/Update_489.php
gabrieljenik: That's how it is laid out on the update 450 and 453

application/helpers/update/updates/Update_489.php
gabrieljenik: > Why is the field 'attributes' filled with the value 'unknown' ? Is this value actually used somewhere?

Honestly, don;t recall now. Should look further.
That's how it is laid out on the update 450 and 453. As we are re-applying a previous recipe, wanted to keep it consistent.

tiborpacalat:
Please make 2 patches out of it, one for 5.x and one for 6.x.

gabrieljenik:
Done for master. Let's merge this first and then create a new one for v5

PR #2407, opened 2022-05-09
Fixed issue #18087: Focus is jumping in the survey structure

gabrieljenik:


c-schmitz:
That actually made it worse. The Problem is that the tool menu disappears. It should be visible, imho. The user owns the survey, so according to permissions it should be visible, unless I am mistaken.

c-schmitz (changes requested):
now the tools menu is not visible at all.

Shnoulle:
Maybe all button must be show in editor view too …

gabrieljenik:
Will continue discussion on Mantis

c-schmitz:
Thanks, great work.

PR #2408, opened 2022-05-09
Fixed issue #18076: Input on demand question type not working when setting multiple rows

gabrieljenik:


gabrieljenik:
Tested it.

Shnoulle (approved):
If it work, not tested

PR #2409, opened 2022-05-09
Fixed issue #16154: Not all files imported with custom question theme

gabrieljenik:


gabrieljenik:
Tested it OK!
image

Shnoulle:
The question is : did we need allow upload such file ? It's sample, to be imported separately.

Shnoulle:
> The question is : did we need allow upload such file ? It's sample, to be imported separately.

@c-schmitz you advice here ? It's not really an issue , no ?

gabrieljenik:
Why shouldn't we?

Shnoulle:
> Why shouldn't we?

Not needed, not used by QuestionTheme, not usuable by QuestionTheme, no reason to have it uploaded.

gabrieljenik:
No reason to limit it either... maybe there will come a QuestionTheme that uses it?
It is part of a deliverable and usually the deliverable is what it is uploaded as it is simple for the user and hence or the growth.

Shnoulle:
> maybe there will come a QuestionTheme that uses it?

tell me how … 

> It is part of a deliverable and usually the deliverable

More part of manual, documentation in my opinion.

gabrieljenik:
> More part of manual, documentation in my opinion.

Isn't that part of a delivberable?
What's wrong in keeping documentation close to source? (or levaing that as an option to the developer)

Shnoulle:
OK : answered by Tony : via Li!meStore : no way to distribute lss except like that.

PR #2411, opened 2022-05-09
Fixed issue #18029: Removed html tags in emails

fsay2604:
Fixed issue #18029: Removed html tags in emails
Dev: Add js script to change configuration of loaded instances of CKEDITOR in the email templates view to prevent losing html tags when switching from wysiwyg to source mode. The fullpage config: "With these settings in place, CKEditor 4 will output the entire HTML page, including the elements outside the <body> section."

### DIFF
--- a/application/views/admin/emailtemplates/email_language_template_tab.php
+++ b/application/views/admin/emailtemplates/email_language_template_tab.php
@@ -126,14 +126,3 @@ App()->getClientScript()->registerScript("ScriptEmailTemplateLanguageTemplate_<?
     {close: '".gT('Close')."', save: '".gT('Save')."'}, '".App()->getController()->createUrl('admin/emailtemplates/getTemplateOfType', array('type' => $tab, 'language' => $grouplang, 'survey' => $surveyid ))."');\n
     ".implode("\n", $script), LSYii_ClientScript::POS_POSTSCRIPT);
 ?>
-
-
-
-    <script >
-        $(document).ready(function(){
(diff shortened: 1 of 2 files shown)


Shnoulle:
In https://github.com/LimeSurvey/LimeSurvey/blob/9e9d3b4658352cda3b37d202e0f0aded0be6ab35/application/views/admin/survey/prepareEditorScript_view.php#L30 seems better in my opinion.

And document ready didn't work with popup editor

fsay2604:
You are right, I move the code to this file but it doesn't affect the popup editor configuration neither.

fsay2604:
I pushed a new commit that changes the configuration of the popup editor if the action is related to edit email templates. Let me know if I need to change something.

c-schmitz:
Is this only a problem in v3, or also in v5?

c-schmitz:
@fsay2604 Does this issue only affect v3 ? Or is it also present in v5 ?

fsay2604:
I have to install a v5 version on my machine before I can confirm that. I'll get back to you asap.

fsay2604:
@c-schmitz I confirm that this issue is also in v5. But the fix in the v3 wouldn't work in v5 since the name of the ckeditor div has changed from "email_..." to "cke_email_...". I can make another PR for the V5 to fix the issue if this PR is accepted (this is my first contribution, I will get familiar with the procedure and standards over time! ). 

Let me know if I can do something else to help.

c-schmitz:
@fsay2604 Thank you for your contribution! Yes, please create another PR for v5.

PR #2415, opened 2022-05-11
Fixed issue #17709: DB session lifetime ignore GUI settings

Shnoulle:
Dev: reset session when application start


gabrieljenik (approved):

PR #2416, opened 2022-05-13
Fixed issue #17709: Session timeout ignoring settings

Shnoulle:
Dev: use iSessionExpirationTime only after load DB


gabrieljenik (approved):
Code looks OK.
Haven't tested it yet.

PR #2420, opened 2022-05-16
Fixed issue #18073: Can't change reminder / invitation date

gabrieljenik:


application/controllers/admin/Tokens.php
```diff
@@ -680,24 +680,29 @@ public function edit($iSurveyId, $iTokenId, $ajax = false)
...
             }
 
             //sent
-            if (trim($request->getPost('sent')) === 'N') {
+            if (trim($request->getPost('sent')) == 'N') {
                 $_POST['sent'] = 'N';
             } else {
-                 $_POST['sent'] = 'Y';
+                $datetimeobj = new Date_Time_Converter(trim($request->getPost('sent')), $dateformatdetails['phpdate'] . ' H:i');
+                $_POST['sent'] = $datetimeobj->convert('Y-m-d H:i');
```
Shnoulle, line 697: invalid date set to false . Need to check data validity

application/controllers/admin/Tokens.php
```diff
@@ -680,24 +680,29 @@ public function edit($iSurveyId, $iTokenId, $ajax = false)
...
             }
 
             // remindersent
-            if (trim($request->getPost('remindersent')) === 'N') {
+            if (trim($request->getPost('remindersent')) == 'N') {
                 $_POST['remindersent'] = 'N';
             } else {
-                $_POST['remindersent'] = 'Y';
+                $datetimeobj = new Date_Time_Converter(trim($request->getPost('remindersent')), $dateformatdetails['phpdate'] . ' H:i');
+                $_POST['remindersent'] = $datetimeobj->convert('Y-m-d H:i');
```
Shnoulle, line 705: Possible to avoid to set $_POST value ?

Shnoulle (changes requested):
We don't use `$request->getPost('sent-date')` etc …

Looking at HTML we have sent-date , remind-date , completed-date

something like this : 

```
$completed = "N";
if ($request->getPost('completed-switch')) {
    $completed = "Y"; // or $completed = date("Y-m-d h:i");
    if(!empty($request->getPost('completed-date'))) {
            $completed = "Y"; // or $completed = date("Y-m-d h:i");
            $datetimeobj = new Date_Time_Converter(trim($request->getPost('completed-date')), $dateformatdetails['phpdate'] . ' H:i');
            if($datetimeobj->convert('Y-m-d H:i')) {
                $completed = datetimeobj->convert('Y-m-d H:i');
            }
    }
}
```

And if we can avoid usage of setting $_POST value

application/controllers/admin/Tokens.php
gabrieljenik: This is just like it is on LTS.
If there is a new bug, maybe we should just create a new ticket.
I know is not your preferred way, but it is hard to streamline the process if not.

Shnoulle (approved):

PR #2424, opened 2022-05-17
Fixed Issue #18120: Server Error 500 when editing global settings: alias "application.controllers.admin.GlobalSettings" invalid

hydroid7:
Fixed issue #18120.
Correct naming of the components fixes the issue.


gabrieljenik:
I wasn't able to reproduce it. But I have docker on top of windows, so that could be tricky with the filenames.
Will let @Shnoulle review it. Seems OK

Shnoulle:
> I wasn't able to reproduce it. But I have docker on top of windows, so that could be tricky with the filenames. Will let @Shnoulle review it. Seems OK

Can not reproduce too, but nginx, let me try on debian/apache

Shnoulle:
Capture d’écran du 2022-05-18 09-56-05

With fix …

Shnoulle:
Fixed both on https://github.com/LimeSurvey/LimeSurvey/pull/2430

Great thanks to you

PR #2434, opened 2022-05-20
Fixed issue #18129 : potential 500 error in file upload

Shnoulle:
Dev: json_decode return null, true, false.
Dev: here we need array of object


gabrieljenik (approved):

Shnoulle:
@maziminke can you check with PHP8 ?

PR #2437, opened 2022-05-20
Fixed issue #17872: Answerfilters reset after pagination changed

gabrieljenik:


Shnoulle:
tested OK
There are no existing solution by Yii grid ?
https://forum.yiiframework.com/search?q=CGridView%20%23yii-1-1-x%20pagesize

Maybe we must have a solution for other grid (token for example, CPDB too …)
Add a solution here :https://github.com/LimeSurvey/LimeSurvey/tree/master/application/extensions/LimeGridView ?

Shnoulle:
If there are a Yii solution or a global solution : it's better to do it at 1st commit

Else : 

1. Less reason to create a global solution 
2. when you create the global solution : you have to remove and update all the local solution.

gabrieljenik:
What about now? @Shnoulle

Shnoulle (approved):

Shnoulle:
> What about now? @Shnoulle

You create the solution , **thank you** ! (i ask if Yii have a solution)

PR #2438, opened 2022-05-20
Fixed issue #18094: Broken view in uploader

gabrieljenik:


Shnoulle:
conflict ;)
I can test on firefox / chrome / midori easily. You do the test on edge ?

gabrieljenik:
We removed the Flex.
Still notice it wasn't introduced on this PR, but previously (aa00a0a5). 
Maybe for having file title to the right?

Shnoulle:
> Maybe for having file title to the right?

Theme task, not from core .
core must do the minimum :)

Shnoulle (approved):

Shnoulle:
PS : why we must use bs and not specific css 

Capture d’écran du 2022-06-20 17-58-19

PR #2439, opened 2022-05-24
Dev #18033: Model errors need friendlier messages

gabrieljenik:
Created a UserException class which can be friednly shown as a flash message or rendered as json message.
This is good for triggering exceptions directly from the model, for example, and ease its handling by the controller.


application/core/LSYii_Application.php
```diff
@@ -514,6 +508,81 @@ public function getTwigCustomExtensionsConfig($sUsertwigextensionrootdir, $aAppl
...
+     * @param array $error
+     * @param LSUserException $exception
+     * @return void
+     */
+    private function handleFriendlyException($exception)
+    {
+        $message = $exception->getMessage();
+        $errors = $exception->getDetailedErrors();
+        if (!empty($errors)) {
+            $message .= "<br>" . implode("<br>", $errors);
```
Shnoulle, line 542: 🤔

application/core/LSYii_Application.php
```diff
@@ -514,6 +508,81 @@ public function getTwigCustomExtensionsConfig($sUsertwigextensionrootdir, $aAppl
...
+            if ($exception->getRedirectUrl() != null) {
+                $outputData['redirectTo'] = $exception->getRedirectUrl();
+            }
+            if ($exception->getNoReload() != null) {
+                $outputData['noReload'] = $exception->getNoReload();
+            }
+            $errors = $exception->getDetailedErrors();
+
+            // Add the detailed errors to the message, so simple handlers can just show it.
+            $outputData['message'] .= "<br>" . implode("<br>", $errors);
```
Shnoulle, line 575: 🤔

Shnoulle (commented):
Why not https://www.yiiframework.com/doc/api/1.1/CHtml#errorSummary-detail usage ?

Shnoulle (changes requested):
Usage of CHtml::errorSummary seems really the best

gabrieljenik:
ErrorSummary is just for models. We are doing something similar, but for more general usage.

Although it started for displaying erros from the model, could be used for errors within methods nested in a controller or a helper. Ex: MassiveActions.

gabrieljenik:
> I really think replace setDetailedErrorsFromModel($oQuestion); by CHtml::errorSummmary($oQuestion) is better.

`CHtml::errorSummmary` renders the errors.
I prefer to have someone else to render the errors.

> And really : the HTML produced here is … i have no word.

Not sure I understand. What do you mean?

When rendering the errors we are doing a basic formatting.
Of course could be better. Not sure it is worth it.

We could have used CHtml::errorSummmary, but now we have something more general which can be used for displaying more than model errors.

Shnoulle:
About html : 

- the attribute is already included in default Yii message , no need to add it again.
- it's a list of error … then produce an html list

And if you can add an header too …

Something like … [this](https://github.com/yiisoft/yii/blob/e1665bdc506cc8e054139f6dd91649e8e12feb5b/framework/web/helpers/CHtml.php#L2236)

Shnoulle:
Thanks you :).

Did you have a clean reproduce way ?

Enter 111 in a new question ?

gabrieljenik:
> Did you have a clean reproduce way ?
> 
> Enter 111 in a new question ?

The easiest is to put an invalid character in the question title and use save and close.

Shnoulle:
big difference between  develop and master before merging (Save and close part)
questionCode-error-develop
questionCode-error-master

application/core/LSYii_Application.php
```diff
@@ -514,6 +508,78 @@ public function getTwigCustomExtensionsConfig($sUsertwigextensionrootdir, $aAppl
...
```
Shnoulle, line 572: Possible to use same HTML than https://github.com/LimeSurvey/LimeSurvey/pull/2439/files#diff-830594dcfebb3cf724c6c04a8b9791bf841a954d95a84614481007be4568a060R100 ?

Shnoulle:
Like to have same clean HTML when save and close and save.

Not the wrapper (it's another issue) but the content
Peek 15-06-2022 09-33

Shnoulle (changes requested):
Same HTML (ul > li) with save and save and close.
If possible.

application/core/LSYii_Application.php
Shnoulle: Oh … thing need JS improvment here …

gabrieljenik:
> big difference between develop and master before merging (Save and close part)

Big difference good?
We will do an update for rendering jsonExceptions and be back

Shnoulle:
> Big difference good?

develop throw a 500 error …

Shnoulle:
I create the issue about usage of Boostrap :)

PR #2440, opened 2022-05-24
Fixed issue #18108: Unable to save existing question with button

Shnoulle:
Dev: really validate whole title using validate
Dev: rename badly named function

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -1618,7 +1618,6 @@ class QuestionAdministrationController extends LSBaseController
         $aData['oSurvey'] = $oSurvey;
         $aData['oQuestionGroup'] = $oQuestionGroup;
         $aData['oQuestion'] = $oQuestion;
-        $aData['action'] = 'copy';
 
         //save the copy ...savecopy (submitbtn pressed ...)
         $savePressed = Yii::app()->request->getParam('savecopy');
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -3022,6 +3022,15 @@ class QuestionAdministrationController extends LSBaseController
     }
 
     /**
+     * @deprecated in 5.3.17
+     * replaced by better name actionValidateQuestionTitle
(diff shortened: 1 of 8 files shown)


Shnoulle:
Oh shit …
`Facebook\WebDriver\Exception\UnrecognizedExceptionException: Element <div id="heading_single_choice_questions" class="panel-heading"> could not be scrolled into view`

gabrieljenik:
Tested OK

olleharstedt:
Not so sure about the "on-blur" check... Can't get the test to work on my machine, but not sure when it broke either.

olleharstedt:
Yep, another "click()" fixed it.

Shnoulle:
> I _think_ the problem is that the test code sends keys to the input field _without_ causing an _unblur_ event.

See https://github.com/LimeSurvey/LimeSurvey/pull/2440/files#diff-738d78a37df841dff8d9b693de92387fca60d44d94488875b8376d8d1b3985d3R81

i click on relevance to send a blurout.

olleharstedt:
> i click on relevance to send a blurout.

Yes, but that only works if there's already a blur, which there wasn't. :) Fixed now, anyway.

Shnoulle:
> Yes, but that only works if there's already a blur, which there wasn't. :) Fixed now, anyway.

Still don't understand ? I mean : sometimes it work without any issue, and sometimes it don't work. It's not a timing issue ?

Shnoulle:
See:

1. Broken 1st time : https://github.com/LimeSurvey/LimeSurvey/runs/7264357838?check_suite_focus=true#step:14:144
2. No update, just launch : https://github.com/LimeSurvey/LimeSurvey/actions/runs/2641412964

olleharstedt:
I could reproduce the issue on my local machine.

Shnoulle:
> I could reproduce the issue on my local machine.

Maybe but it work sometimes : https://github.com/LimeSurvey/LimeSurvey/runs/7278683887?check_suite_focus=true OK ?

olleharstedt:
No idea. Now it works anyway. :)

PR #2443, opened 2022-05-26
Fixed issue #18146: Unable to use "other" as subquestion code

Shnoulle:
Dev: reset other to N when save by rules
Dev: in 3.X : done in database, getPost is empty when not submitted
Dev: inverse must be done too when try to update to single choice with other

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -1512,10 +1512,11 @@ class Question extends LSActiveRecord
 
     /**
      * Remove subquestion if needed when update question type
+     * @return void
      */
     protected function removeInvalidSubquestions()
     {
-        if (!$this->getAllowSubquestions()) {
+        if ($this->getAllowSubquestions()) {
             return;
         }
 
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -138,18 +138,13 @@ class Question extends LSActiveRecord
      */
     public function rules()
     {
+        /* Basic rules */
         $aRules = array(
             array('title', 'required', 'on' => 'update, insert', 'message' => gT('The question code is mandatory.', 'unescaped')),
             array('title', 'length', 'min' => 1, 'max' => 20, 'on' => 'update, insert'),
             array('qid,sid,gid,parent_qid', 'numerical', 'integerOnly' => true),
             array('qid', 'unique','message' => sprintf(gT("Question id (qid) : '%s' is already in use."), $this->qid)),// Still needed ?
             array('other', 'in', 'range' => array('Y', 'N'), 'allowEmpty' => true),
-            array('other', 'filter', 'filter' => function ($value) {
-                if ($this->getAllowOther()) {
-                    return $value;
-                }
-                return 'N';
-            }),
             array('mandatory', 'in', 'range' => array('Y', 'S', 'N'), 'allowEmpty' => true),
             array('encrypted', 'in', 'range' => array('Y', 'N'), 'allowEmpty' => true),
             array('question_order', 'numerical', 'integerOnly' => true, 'allowEmpty' => true),
@@ -159,8 +154,20 @@ class Question extends LSActiveRecord
             array('preg,relevance', 'safe'),
             array('modulename', 'length', 'max' => 255),
         );
-        // Always enforce unicity on Sub question code (DB issue).
+        /* Filtering */
+        /* other must be no when other is not allowed */
+        $aRules[] = array('other', 'filter', 'filter' => function ($value) {
+            if ($this->getAllowOther()) {
+                return $value;
+            }
+            return 'N';
+        });
+        /* Don't save empty or 'core' question theme name */
+        $aRules[] = ['question_theme_name', 'questionThemeNameValidator'];
+        /* Specific rules to avoid collapse with column name in database */
         if ($this->parent_qid) {
+            /* Subquestion specific rules */
+            /* unicity of title by scale */
             $aRules[] = array('title', 'unique', 'caseSensitive' => false,
                 'criteria' => array(
                     'condition' => 'sid=:sid AND parent_qid=:parent_qid and scale_id=:scale_id',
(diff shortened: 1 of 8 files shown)


Shnoulle:
Maybe set to null when not applicable ?

Shnoulle:
Usage of questiontype : set it as draft temporary

application/models/Question.php
```diff
@@ -144,6 +144,12 @@ public function rules()
             array('qid,sid,gid,parent_qid', 'numerical', 'integerOnly' => true),
             array('qid', 'unique','message' => sprintf(gT("Question id (qid) : '%s' is already in use."), $this->qid)),// Still needed ?
             array('other', 'in', 'range' => array('Y', 'N'), 'allowEmpty' => true),
+            array('other', 'filter', 'filter' => function ($value) {
```
gabrieljenik, line 147: I believe this shall have a bigger sign. Like a separate block of lines inside the rules() method, as for highlighting this is a filter. (which is not frequently used)

application/models/Question.php
```diff
@@ -176,10 +182,12 @@ public function rules()
                 $aRules[] = array('title', 'match', 'pattern' => '/comment$/', 'not' => true, 'message' => gT("'comment' suffix can not be used with multiple choice with comments."));
             }
         } else {
-            // Disallow other if sub question have 'other' for title
-            $oSubquestionOther = Question::model()->find("parent_qid=:parent_qid and LOWER(title)='other'", array("parent_qid" => $this->qid));
-            if ($oSubquestionOther) {
-                $aRules[] = array('other', 'compare', 'compareValue' => 'Y', 'operator' => '!=', 'message' => sprintf(gT("'%s' can not be used if the 'Other' option for this question is activated."), 'other'), 'except' => 'archiveimport');
+            if ($this->getAsSubquestions()) {
+                // Disallow other if sub question have 'other' for title
+                $oSubquestionOther = Question::model()->find("parent_qid=:parent_qid and LOWER(title)='other'", array("parent_qid" => $this->qid));
```
gabrieljenik, line 187: Maybe we can wrap this into a method?

application/models/Question.php
```diff
@@ -176,10 +182,12 @@ public function rules()
...
+                if ($oSubquestionOther) {
+                    $aRules[] = array('other', 'compare', 'compareValue' => 'Y', 'operator' => '!=', 'message' => sprintf(gT("'%s' can not be used if the 'Other' option for this question is activated."), 'other'), 'except' => 'archiveimport');
```
gabrieljenik, line 189: Can you remind me why the archiveImport is excepted?
Maybe add it as a comment?

gabrieljenik (commented):
Left a few comments! Nice PR!

application/models/Question.php
Shnoulle: I think it's for LSA

application/models/Question.php
Shnoulle: Just separate in same function ? Can be an idea :)

Keep it at anonymous function or not ? Here , i think it's cleaner enough at anonymous function (and usage only for other)

application/models/Question.php
Shnoulle: you mean something like `$oSubquestionOther = $this->getAsOtherSubquestion()` ?

Can be better, right.

application/models/Question.php
Shnoulle: > I think it's for LSA

but unsure … let's grep

application/models/Question.php
gabrieljenik: >> Just separate in same function ? Can be an idea :)

Yes, just as a separate block of lines inside the function

>> Keep it at anonymous function or not ? Here , i think it's cleaner enough at anonymous function (and usage only for other)

Maybe a setOther method?

application/models/Question.php
gabrieljenik: why are there methods names like `getAsxxxx`?
What do getAsxxx means?

application/models/Question.php
gabrieljenik: I would add a comment about it.

application/models/Question.php
Shnoulle: > Maybe a `setOther()` method?

Think it less clear what it's done if not in existing rule …

application/models/Question.php
gabrieljenik: `hasSubquestionsOther` ?
Do we need the `get` prefix?

application/models/Question.php
Shnoulle: And here : i need to remove archiveimport allowed : it broke DB :)

application/models/Question.php
Shnoulle: `$question->hasubquestionother` == `$question->getHasOtherSubquestion()` no ?

application/models/Question.php
gabrieljenik: > > Maybe a `setOther()` method?
> 
> Think it less clear what it's done if not in existing rule …

I meant: call setOther from the rules

Shnoulle: ? We have 'filter' rules, then adding a function like `addRulesFilterOther` ? don't like.

I prefer to create
`addRulesSubQuestiion` and `haveRulesPrimaryQuestion` in this case.

application/models/Question.php
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/2443/files#diff-75d3428e7aa3f16b9dc93ef54dfa4d0556c04431cb53510749ded531d549a148R1488

Shnoulle:
@c-schmitz why you remove bug fix ? There are a clearly a bug in 5.X , this bug didn't exist in 3.X

application/models/Question.php
Shnoulle: Check please :)

gabrieljenik: I mean you could have like this:
```
function setOther($value) {
 if ($this->getAllowOther()) {
          return $value;
      }
      return 'N';
  }
}
```

and then from within rules
```
array('other', 'filter', 'filter' => function ($value) { return $this->SetOther($value); }
```

application/models/Question.php
gabrieljenik: You could have like this 
```
function getSubquestionOther() {
   return Question::model()->find("parent_qid=:parent_qid and LOWER(title)='other'", array("parent_qid" => $this->qid));
}
```

then use `getSubquestionOther()` in the code.

application/models/Question.php
Shnoulle: > Not sure is the best, but thats what I meant originally.

Compare current filter on other and filter on theme : and tell me the simple one … I think anonymous function is clearer here.

Maybe @olleharstedt or @c-schmitz advice ?

gabrieljenik: Just keep it as it is.
No big deal!

gabrieljenik (approved):

application/models/Question.php
Shnoulle: vote ;) 

I'm OK to create a separate function, hard to know what is the better solution. 

Create 2 function : addSubquestionsRules and addPrimaryQuestionRules seems better to do before.

gabrieljenik:
https://bugs.limesurvey.org/view.php?id=18146#c70337

application/models/Question.php
olleharstedt: You can also write a function that produces functions. :)

Shnoulle: The question is : did we really need to construct some thing more complex than 
```

        $aRules[] = array('other', 'filter', 'filter' => function ($value) {
            if ($this->getAllowOther()) {
                return $value;
            }
            return 'N';
        });
```

When we don't have a better separation of rules for subquestions and questions for example.

gabrieljenik:
Tested OK.  Haven't done regression testing.
Still considering the strange reproduction steps needed and the code updated, as to avoid regressions, I would send it to DEV as to have the code be sinked and lived for a while.

Shnoulle:
>  the strange reproduction steps 

? Just set other to a question, never resetted to no when updating the question.

Mazi have the issue in real survey … and can not fix without hacking DB … 

It's a big issue, really a big issue …
And we don't have this issue in 3.X …

PR #2444, opened 2022-05-29
Fixed issue #18158: `Property "CDbCriteria.qid" is not defined` status returned by delete_question

edgarrmondragon:
Fixed issue [#18158](https://bugs.limesurvey.org/view.php?id=18158): `Property "CDbCriteria.qid" is not defined` status returned by `delete_question`

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1444,7 +1444,7 @@ class remotecontrol_handle
                     Question::model()->deleteAll($sCriteria);
 
                     // delete defaultvalues and defaultvalueL10ns
-                    $oDefaultValues = DefaultValue::model()->findAll(array('qid' => $iQuestionID));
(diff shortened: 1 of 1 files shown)


Shnoulle:
Seems unneeded when use ->delete : https://github.com/LimeSurvey/LimeSurvey/blob/2cced37bb82839792c2e4ecfbcf5d7a0e51163de/application/models/Question.php#L394

Best is to move to $questtion->findByPk($qid)->delete after other control is done.

Shnoulle (changes requested):
At https://github.com/LimeSurvey/LimeSurvey/blob/74db1e4d8dde21f49ccd212eb1eff925d79c8c7a/application/helpers/remotecontrol/remotecontrol_handle.php#L1435

replace by $question->delete please :)

edgarrmondragon:
@Shnoulle thanks for reviewing, I've updated the method to use `->delete()`

Shnoulle:
I do the test , thanks to you !

PR #2446, opened 2022-05-30
Fixed issue: #18029 - Removed html tags in emails for v5

fsay2604:
Fixed issue #18029:
Dev:  Add js script to change configuration of loaded instances of CKEDITOR in the email templates view to prevent losing html tags when switching from wysiwyg to source mode. The fullpage config: "With these settings in place, CKEditor 4 will output the entire HTML page, including the elements outside the section.". This patch was previously merged in 3.x LTS, this PR is for v5.


gabrieljenik (approved):
Code looks good

PR #2170, opened 2021-12-06
Fixed issue #17654: spurious error "Incorrect username and/or passwor…

Shnoulle:
Fixed issue #17654: spurious error "Incorrect username and/or password!" on auth_webserver autocreating user
Dev: minimal fix are fixing plugin

PR #2448, opened 2022-05-31
Fixed issue #spurious error "Incorrect username …" in webserver Auth

Shnoulle:
Dev: Add optional param \LimeSurvey\PluginManager\PluginEvent when potentially needed

### DIFF
--- a/application/libraries/PluginManager/AuthPluginBase.php
+++ b/application/libraries/PluginManager/AuthPluginBase.php
@@ -85,17 +85,19 @@ abstract class AuthPluginBase extends PluginBase
      * Set authentication result to success for the given user object.
      *
      * @param User $user
+     * @param \LimeSurvey\PluginManager\PluginEvent, current event if not set
      * @return AuthPluginBase
      */
-    public function setAuthSuccess(User $user)
+    public function setAuthSuccess(User $user, \LimeSurvey\PluginManager\PluginEvent $event = null)
     {
-        $event = $this->getEvent();
-        $identity = $this->getEvent()->get('identity');
+        if (empty($event)) {
+            $event = $this->getEvent();
+        }
+        $identity = $event->get('identity');
         $identity->id = $user->uid;
         $identity->user = $user;
-        $this->getEvent()->set('identity', $identity);
+        $event->set('identity', $identity);
         $event->set('result', new LSAuthResult(self::ERROR_NONE));
-
         return $this;
     }
 
@@ -104,30 +106,34 @@ abstract class AuthPluginBase extends PluginBase
      *
      * @param int $code Any of the constants defined in this class
      * @param string $message An optional message to return about the failure
+     * @param \LimeSurvey\PluginManager\PluginEvent, current event if not set
      * @return AuthPluginBase
      */
-    public function setAuthFailure($code = self::ERROR_UNKNOWN_IDENTITY, $message = '')
+    public function setAuthFailure($code = self::ERROR_UNKNOWN_IDENTITY, $message = '', \LimeSurvey\PluginManager\PluginEvent $event = null)
     {
-        $event = $this->getEvent();
+        if (empty($event)) {
+            $event = $this->getEvent();
+        }
         $identity = $this->getEvent()->get('identity');
         $identity->id = null;
         $event->set('result', new LSAuthResult($code, $message));
-
         return $this;
     }
 
     /**
      * Set this plugin to handle the authentication
      *
+     * @param \LimeSurvey\PluginManager\PluginEvent, current event if not set
      * @return AuthPluginBase
      */
-    public function setAuthPlugin()
+    public function setAuthPlugin(\LimeSurvey\PluginManager\PluginEvent $event = null)
     {
-        $this->getEvent();
-        $identity = $this->getEvent()->get('identity');
+        if (empty($event)) {
+            $event = $this->getEvent();
+        }
+        $identity = $event->get('identity');
         $identity->plugin = get_class($this);
-        $this->getEvent()->stop();
-
+        $event->stop();
         return $this;
     }
 
--- a/application/core/plugins/Authwebserver/Authwebserver.php
+++ b/application/core/plugins/Authwebserver/Authwebserver.php
@@ -3,7 +3,7 @@
 class Authwebserver extends LimeSurvey\PluginManager\AuthPluginBase
 {
     protected $storage = 'DbStorage';
-    
+
     protected static $description = 'Core: Webserver authentication';
     protected static $name = 'Webserver';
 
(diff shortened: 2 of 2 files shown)


Shnoulle:
Update API version … new feature in API : 5.4

Shnoulle:
`Error: '/home/runner/work/LimeSurvey/LimeSurvey/third_party/composer/tmp-39795d923d59d80ad603e95578ec9345' is a corrupted zip archive (0 bytes), try again.` thi nk it's  not related with the commit ;)

gabrieljenik (approved):
Code structure looks OK. Haven't tested it

application/libraries/PluginManager/AuthPluginBase.php
```diff
@@ -85,17 +85,19 @@ public function remoteControlLogin()
      * Set authentication result to success for the given user object.
      *
      * @param User $user
+     * @param \LimeSurvey\PluginManager\PluginEvent, current event if not set
      * @return AuthPluginBase
      */
-    public function setAuthSuccess(User $user)
+    public function setAuthSuccess(User $user, \LimeSurvey\PluginManager\PluginEvent $event = null)
```
olleharstedt, line 91: Can you add a `use` statement on top of file, so we don't have to write the full namespace? Thanks.

olleharstedt (changes requested):

application/libraries/PluginManager/AuthPluginBase.php
Shnoulle: Or namespace\ ?

olleharstedt:
@Shnoulle Did you test your own PR? o0

Shnoulle:
I check again after last commit,

Else : https://bugs.limesurvey.org/view.php?id=17654#c67723

Shnoulle:
> Really, a developer testing his/her own stuff is not a valid test...

prigaux test the previous one https://github.com/LimeSurvey/LimeSurvey/pull/2170 
But let's go to have another test with this one

Shnoulle:
> Connect as superadmin (and stay connected)
> Create an user TEST for login , just give him right to connect via AuthDB
> configure Authwebserver to get user by $_SERVER['TEST_USER'];
> update config.php and add $_SERVER['TEST_USER'] = 'TEST'; at start
> Open another browser, go to limesurvey admin page 

Ad check with removing TEST user with autocreate and without.

https://bugs.limesurvey.org/view.php?id=18169

brenard:
Hello, in case of an upgrade if a limesurvey instance using Authserver, I reproduce this bug : my users haven't the _auth_webserver_ permission and have a loop during their authentication.

My diagnostic was complexified because in the admin interface, when I try to edit their permissions, the corresponding permission is checked (even if I can't find it in database). By saving permissions of this user, without changing anything, the permission is effectively added in database and the user is enabled to connect. Do you think it could be also fixed ?

Futhermore, in my case, I didn't find a proper solution to bulk add the missing permission to my users. When trying to fix it, I found [this post](https://forums.limesurvey.org/forum/installation-a-update-issues/120225-use-web-server-authentication-bulk-setting) suggering to add the `$config['auth_webserver'] = true;` in _config.php_ to make new users to automatically have this permission. This parameter already exists in _config-defaults.php_ and the [documentation](https://manual.limesurvey.org/Optional_settings#Authentication_delegation_to_the_webserver) specify that it is to do, but I thought I'd understood that all parameters have been [migrating in the plugin configuration](https://help.limesurvey.org/portal/en/kb/articles/plugin-manager-22-4-2021#Webserver). Moreover, in the [plugin code](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/plugins/Authwebserver/Authwebserver.php#L135), the permission is automatically added to all users. I think to documentaion and help center isn't uptodate.

To finish, if you have any advice to how to add the missing permission after upgrading, it could be useful for me (and idealy [documented in upgrade process](https://manual.limesurvey.org/Upgrading_from_a_previous_version)).

Shnoulle:
> Hello, in case of an upgrade if a limesurvey instance using Authserver, I reproduce this bug : my users haven't the auth_webserver permission and have a loop during their authentication.

Can you test with this commit ?
This commit must fix this issue …

brenard:
> > Hello, in case of an upgrade if a limesurvey instance using Authserver, I reproduce this bug : my users haven't the auth_webserver permission and have a loop during their authentication.
> 
> Can you test with this commit ? This commit must fix this issue …

I just test your commits by patching the two files and it seem working great : existing users without the _auth_webserver_ permission could login. I'm not sure to understand how it's work because, in `beforeLogin()` method, you seem authorized only existing users with the _auth_webserver_ permission or inexisting user when autocreate is enabled.

Shnoulle:
> > > Hello, in case of an upgrade if a limesurvey instance using Authserver, I reproduce this bug : my users haven't the auth_webserver permission and have a loop during their authentication.
> > 
> > 
> > Can you test with this commit ? This commit must fix this issue …
> 
> I just test your commits by patching the two files and it seem working great : existing users without the _auth_webserver_ permission could login. I'm not sure to understand how it's work because, in `beforeLogin()` method, you seem authorized only existing users with the _auth_webserver_ permission or inexisting user when autocreate is enabled.

Yes it's the case. And if you set AuthWeb by default : throw a clean 401. Else : leave AuthDB doing his own task.

brenard:
> Yes it's the case. And if you set AuthWeb by default : throw a clean 401. Else : leave AuthDB doing his own task.

OK, so it seem not working as expected unless the _auth_webserver_ permission is defaulty enabled to all users.

In my case, with an user without the _auth_webserver_ permission (check in DB) and AuthWeb set by default, I haven't the expected 401 error.

By adding some debug in your code, I see that the `Permission::model()->hasGlobalPermission()` check return `True`, so this suggests that the permission is true by default, but is this the expected behavior? If so, I don't understand the interest of this permission because I don't see how we could disable it.

Shnoulle:
If i don't make error `Permission::model()->hasGlobalPermission('auth_webserver', 'read', $oUser->uid)` return true for superadmin user.

It's the case here ?

brenard:
> If i don't make error `Permission::model()->hasGlobalPermission('auth_webserver', 'read', $oUser->uid)` return true for superadmin user.
> 
> It's the case here ?

You're right ! I try with another user (not superadmin) and I have the 401 error. So it's seem work as expected.

Do you have any advice to how to bulk adding the missing permission to my users ? As said before, it could be great to document a good method in the upgrade process.

Shnoulle:
> Do you have any advice to how to bulk adding the missing permission to my users ? As said before, it could be great to document a good method in the upgrade process.

No … except db update via direct SQL

Shnoulle:
@olleharstedt i remove the need testing part …

olleharstedt:
Should be merged to develop?

Shnoulle:
> Should be merged to develop?

There are real issue currently with WebAuth … with or without WebAuth set as default.

This one for example : https://bugs.limesurvey.org/view.php?id=18169 if user is set by SERVER : always get a loop (if it don't have the right to connect to AuthWeb)

PR #2449, opened 2022-06-01
Fixed issue #17655: Launch survey via token link : unclear icon

Shnoulle:
Dev: Use same icon than global view, review string


Shnoulle:
Peek 01-06-2022 10-24

maybe "Execute this survey with this token" since global view is "Execute this survey".

Shnoulle:
I can use fa-play always, but then : maybe best to use fa-play for global preview button too ?

PR #2455, opened 2022-06-03
Fixed issue #17976: Bad place for pencil in edit question

gabrieljenik:


Shnoulle:
I just think it's great to have a form-group class : form-html-editor or scope-contains-ckeditor don't know


Check with HTML inline editor : 
Current
Current : width:100%

withfix
With fix

gabrieljenik:
> I just think it's great to have a form-group class : form-html-editor or scope-contains-ckeditor don't know

What do you mean? Sorry, don't understand. Those classes were removed.

> Check with HTML inline editor :

You have created a different bug for that, right?
https://bugs.limesurvey.org/view.php?id=18175

Shnoulle:
> > I just think it's great to have a form-group class : form-html-editor or scope-contains-ckeditor don't know
> 
> What do you mean? Sorry, don't understand. Those classes were removed.
> 
> > Check with HTML inline editor :

Check the screenshot done on https://github.com/LimeSurvey/LimeSurvey/pull/2455#issuecomment-1147129499

1st one is done before merge, 2nd one after.

Just reloading same page


> 
> You have created a different bug for that, right? https://bugs.limesurvey.org/view.php?id=18175

No , it's a different issue here.

Shnoulle:
col-12 ?

Shnoulle (approved):
tested 6 options : OK

PR #2456, opened 2022-06-03
Fixed issue #18155: Translation missing in question editor

gabrieljenik:


Shnoulle:
Need to be in https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/AdvancedSettingWidget/AdvancedSettingWidget.php#L40

Not in view.

See issue with empty help.

Shnoulle (changes requested):
Best is to move translation in https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/AdvancedSettingWidget/AdvancedSettingWidget.php

Not a view task

gabrieljenik:
> Best is to move translation in https://github.com/LimeSurvey/LimeSurvey/blob/master/application/extensions/AdvancedSettingWidget/AdvancedSettingWidget.php

There is a place already for placing strings to be caught by translation-robot.
https://github.com/LimeSurvey/LimeSurvey/pull/2456/files#diff-2a53578ef545195e185f082a9a75bc01df2f9dc49f5b7e7e105f6c91f1b70724

> Not a view task

Not sure I understand

Shnoulle:
> There is a place already for placing strings to be caught by translation-robot.

Yes , and ? See the issue with help : https://github.com/LimeSurvey/LimeSurvey/commit/c28e7d06a2fb195f548a9d3c5999ccca61137e10

Best is to move translation in controller.

> Not a view task


It's not a view task to translate a variable. View have to use fixed string for translation.

gabrieljenik:
And what about this?
https://github.com/LimeSurvey/LimeSurvey/commit/c28e7d06a2fb195f548a9d3c5999ccca61137e10#diff-497bd29b2a6f30017ce4e096ae5c7d8ba3d4e838aa4fbc5c5a575e5dc71d79c1R5

Also, the views updated here are for specific types.
The widget is for all types. We would be applying the translation to all types?

If this kind of reoganization would be on a separate ticket, I am sure it would have very low prio.
So, do we need to keep spending time on this?
We try to do fixing + touching a bit more for making things better.
But I think we have passed that point here.

Shnoulle:
>And what about this?

You're right : i didn't update this lines when i fix help (and other) issue … 

> The widget is for all types. We would be applying the translation to all types?

It's the original issue : it's done in other views but not in this 2 views … next time a dev add a view : potential issue about translation again …
The translation must be done for ALL type , and here : it's not done for 2 types …


> If this kind of reoganization would be on a separate ticket, I am sure it would have very low prio.

It's clearly not a big reorganisation here … we can create hundred of tickets then and never improve code …

gabrieljenik:
@Shnoulle Can you please re-review? Thanks

Shnoulle:
Can you check `   │ Facebook\WebDriver\Exception\UnexpectedAlertOpenException: Dismissed user prompt dialog: Internal error in updateQuestionAttributes: [object Object]: ` ?

Shnoulle:
> Can you check ` │ Facebook\WebDriver\Exception\UnexpectedAlertOpenException: Dismissed user prompt dialog: Internal error in updateQuestionAttributes: [object Object]:` ?

Same issue this morning

gabrieljenik:
Finally, green CI

Shnoulle:
double translation ?

PR #2459, opened 2022-06-07
Fixed issue #17700: Custom logo disappears in extended themes

gabrieljenik:
Fixing:
- Settings at global level are not sticking (not saving).
- Need to enter multiple times to theme options or to a survey (participant mode) to have settings being picked up correctly.


Shnoulle:
Hard to review here … ;)

Shnoulle:
It don't fix the current issue about JS error ?
Nothing work on theme management today …

Shnoulle:
I think it's better, code is OK, but unsure it fix the related issue …

PR #2461, opened 2022-06-08
[WIP] 18170: Database error when clicking entry to edit and survey was deactivated

Shnoulle:
Dev: throw 404 if response deleted (not response table) and try to edit
Dev: move from App()->session['flashmessage'] to App()->setFlashMessage (done for this)
Dev: reorder testing (test access before and out)

### DIFF
--- a/application/controllers/ResponsesController.php
+++ b/application/controllers/ResponsesController.php
@@ -126,241 +126,242 @@ class ResponsesController extends LSBaseController
         }
         $survey = Survey::model()->findByPk($surveyId);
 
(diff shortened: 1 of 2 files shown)


Shnoulle:
There less real reorganization than visually

Reverse check system for permission : `if (NoPermission) {quit} DoTheAction on lot of line`  not `if(Permissioon) {DoTheAction on lot of line} else {quit}` 
Then all tab updated …

gabrieljenik:
> Reverse check system for permission

Yes yes Notice that.
I like that way of coding.

PR #2464, opened 2022-06-13
Fixed issue #18184: cpd_importParticipants with encryption turned on, fails to save the participant encrypted in CPDB, causing 500 error to CPDB

gabrieljenik:


Shnoulle (approved):
Seems OK,

But need big testing in my opinion …

related pull request : https://github.com/LimeSurvey/LimeSurvey/pull/2149/files

gabrieljenik:
@olleharstedt As this seems to require lot of testing, should we push to develop?

gabrieljenik:
Will be creating a ticket for this.

gabrieljenik:
Automatic test added.

Shnoulle (approved):

olleharstedt:
> @olleharstedt As this seems to require lot of testing, should we push to develop?

This is fixed now? Master is ok?

gabrieljenik:
> This is fixed now? Master is ok?

This PR holds the fix and also an automatic test.

PR #2465, opened 2022-06-13
Fixed issue #18180: Calling set_participant_properties with an invalid token id throws an internal server error

gabrieljenik:


olleharstedt:
Uh, code checked or PR tested...? Which one was done? Both?

Shnoulle:
> Uh, code checked or PR tested...? Which one was done? Both?

Oups … code checked, not tested sorry.

gabrieljenik:
@olleharstedt Maybe we can think of pushing to develop as to have mature it and get more testing time?

olleharstedt:
"Testing time" only works if someone is actually testing. :D Else, we need to write unit tests for encrypted/non-encrypted cases.

gabrieljenik:
> "Testing time" only works if someone is actually testing. :D Else, we need to write unit tests for encrypted/non-encrypted cases.

Here you are :)

olleharstedt:
Neat!

PR #2468, opened 2022-06-17
Fixed issue #18193: Question theme error: Missing metadata (Type: X)

Shnoulle:
Dev: questionThemeNameValidator in afterFind Yii event

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -94,6 +94,24 @@ class Question extends LSActiveRecord
 
     /**
      * @inheritdoc
+     */
+    public function init()
+    {
+        $this->attachEventHandler("onAfterFind", array($this, 'afterFindQuestion'));
+    }
+
+    /**
+     * Action to do after find a question
+     * @see https://www.yiiframework.com/doc/api/1.1/CActiveRecordBehavior#afterFind-detail
+     * - fix theme name after find
+     */
+    public function afterFindQuestion()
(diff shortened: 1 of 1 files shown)


Shnoulle:
Some old lss/lsa didn't have valid `question_theme_name` then fix it when load if needed.

Did you try with https://forums.limesurvey.org/forum/german-forum/127651-fragebogen-bricht-ab (see related mantis issue)

c-schmitz:
I have the same doubts regarding performance. Could this be done more performance-friendly?

Shnoulle:
> I have the same doubts regarding performance. Could this be done more performance-friendly?

On save ? Save happen lesser.

And fix on update for older value ?

olleharstedt:
Is the event inside a loop?

Shnoulle:
> Is the event inside a loop?

? All question are find at a time.

Shnoulle:
> > I have the same doubts regarding performance. Could this be done more performance-friendly?
> 
> On save ? Save happen lesser.
> 
> And fix on update for older value ?

This ?

tiborpacalat:
Fixing it on save sounds fine, please update the patch.

Shnoulle:
5 and master , right ?

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/3252

PR #2470, opened 2022-06-17
Fixed issue #18186: Language changer does not work in registration page

seccanj:
Dev:

### DIFF
--- a/assets/packages/limesurvey/survey.js
+++ b/assets/packages/limesurvey/survey.js
@@ -213,7 +213,7 @@ function activateLanguageChanger(){
         }
     });
     /* Language changer dropdown */
-    $('.form-change-lang [name="lang"] option:not(selected)').on('click', function(event) {
+    $('.form-change-lang [name="lang"]').on('change', function(event) {
         var closestForm = $(this).closest('form');
         var newLang = $(this).val();
         if (!closestForm.length) {


Shnoulle:
No,

Else user can not use Keyboard navigation … 
See https://bugs.limesurvey.org/view.php?id=14211

We mist find a better way … eventually : leave whole JS outside ?

seccanj:
Ok, anyway i noticed this is the same code as on the master branch.
How about responding only to the language change button being pushed, instead of the select box being changed?

Shnoulle:
> 

Yes, think it's the good solution.

And check if we can add it to top menu too (in JS then).

Token survey : no register is OK, not register… We have the top menu working in JS 
Let me update the mantis issue :)

Shnoulle:
You can apply same workaround then master : https://github.com/LimeSurvey/LimeSurvey/blob/88137f01377c6db10aebf809e761cc5c17bd2a2d/assets/packages/limesurvey/survey.js#L228

For the 1st step.

The other issue about language menu can be done after (and in master only)

seccanj:
OK Denis, found a better solution that also restores the functionality of the submit button, now broken.
Please take a look at the new pull request.
Thanks.

Shnoulle:
Thank you !

PR #2472, opened 2022-06-20
Fixed issue #18201: Unable to import LSA file with MSSQL and debug set

Shnoulle:
Dev: check if $result['CONTRAINT_NAME'] exist before using it


gabrieljenik (approved):
Code looks OK from a structural point of view.
Haven't tested it

PR #2474, opened 2022-06-21
Fixed issue #18186: Language changer does not work in registration page

seccanj:
Fixed issue #18186: Language changer does not work in registration page

Avoid same accessibility issues in https://bugs.limesurvey.org/view.php?id=14211 and applied workaround in https://github.com/LimeSurvey/LimeSurvey/blob/88137f01377c6db10aebf809e761cc5c17bd2a2d/assets/packages/limesurvey/survey.js#L228

Also restored the language change submit button functionality.

Tested with and without a wrapping "form".

### DIFF
--- a/assets/packages/limesurvey/survey.js
+++ b/assets/packages/limesurvey/survey.js
@@ -212,48 +212,75 @@ function activateLanguageChanger(){
             $(this).closest('.ls-language-changer-item').find(":submit").click();
         }
     });
+
     /* Language changer dropdown */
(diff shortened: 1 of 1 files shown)


Shnoulle:
Tested on FF+chrome+IE11+Edge with KB and mouse : OK (vanilla theme) same without JS

Shnoulle:
@gabrieljenik : you want to put an eye ?

seccanj:
Thanks

Shnoulle:
Your welcome : thanks to **you** !

PR #2476, opened 2022-06-21
New feature #18206: Add an Expression Engine function

leonjanzen-work:
New feature #18206: Add an Expression Engine function for extracting a value from a JSON encoded string
Dev: Added a function to the list of built-in Expression Engine functions.


application/helpers/expressions/em_core_helper.php
```diff
@@ -3250,3 +3251,74 @@ function exprmgr_unique($args)
...
+	$json_array = json_decode($json_encoded_string,true);
+
+	if(is_null($json_array)) { //Check if the string was not valid JSON, and could not be decoded.
+		//I'm not sure which is preferable, to return an error message, or throw an error, or do nothing and just return null.
+		
+		//Return an error message as the value.
+		//$json_element = "Invalid JSON (json_decode() failed).";
+
+		//Throw an error.
+		throw new \ErrorException("Invalid JSON (json_decode() failed).");
```
Shnoulle, line 3292: 😵 there are a system to get errors https://github.com/LimeSurvey/LimeSurvey/blob/297841a92f4e0e52e5a9362d73665eb5a00df4b6/application/helpers/expressions/em_core_helper.php#L2229

leonjanzen-work: Thanks for taking the time to review my code. I wasn't aware that this functionality can be accomplished through a plugin, so I'll pursue that option instead.

PR #2477, opened 2022-06-22
Fixed issue #18171: Timings are not saved on MSSQL DB

Shnoulle:
Dev: need switchMSSQLIdentityInsert
Dev: Throw Exception if debug set
Dev: remove uneeded and unused survey object (memory usage)
Dev: refresh only if really needed

### DIFF
--- a/application/models/SurveyTimingDynamic.php
+++ b/application/models/SurveyTimingDynamic.php
@@ -22,9 +22,6 @@ class SurveyTimingDynamic extends LSActiveRecord
     /** @var int $sid Survey id */
     protected static $sid = 0;
 
-    /** @var Survey $survey*/
-    protected static $survey;
-
     /**
      * @inheritdoc
      * @param string $sid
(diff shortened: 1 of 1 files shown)


Shnoulle:
tested by reporter : https://bugs.limesurvey.org/view.php?id=18171#c70476
(and by me in real)

application/models/SurveyTimingDynamic.php
```diff
@@ -141,10 +139,22 @@ public function insertRecords($data)
             $record->$k = $v;
         }
 
+        if (isset($data['id'])) {
+             switchMSSQLIdentityInsert(trim($this->tableName(), "{}"), true);
+        }
         try {
             $record->save();
+            if (isset($data['id'])) {
+                 switchMSSQLIdentityInsert(trim($this->tableName(), "{}"), false);
```
gabrieljenik, line 148: Out of curiosity: Why are these needed?

Shnoulle: Search on internet : `SQLSTATE[23000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Cannot insert explicit value for identity column in table ** when IDENTITY_INSERT is set to OFF.`

See the report here : https://bugs.limesurvey.org/view.php?id=18172

You have the exception.

Then each time you have a request with pk : you need to set IDENTITY_INSERT to ON

More information : https://stackoverflow.com/a/1334032/2239406

We have a global function : https://github.com/LimeSurvey/LimeSurvey/blob/1857de3ff170a9525cb2f87fdc10fd7bf38260d1/application/helpers/common_helper.php#L5042

I think we can improve it and move it to an automatic set beforeSave but really more complex :)

PR #2478, opened 2022-06-22
Fixed issue #17300: better show selected option difference

Shnoulle:
Dev: real difference, not only by color.
Dev: font-awesome radio


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2359#issuecomment-1113118535 > This is nice, but I think there should be a design decision. The improvement is a huge change to all buttons. @c-schmitz what do you think?

 https://github.com/LimeSurvey/LimeSurvey/pull/2359#pullrequestreview-973872680 > Only CSS changes.

Shnoulle:
It fix too an issue when you force color.
But i review it to add a diff to switch too
Capture d’écran du 2022-07-06 17-42-25

I report the issue (accessibility issue)

Shnoulle:
Accessibility issue : LimeSurey admin is unusuable with forced color

Time to fix conflict ?

Capture d’écran du 2022-10-18 19-08-39

Shnoulle:
What do you want to discuss ?

A difference without color is needed …

olleharstedt:
Hm, the default Bootstrap 5 buttongroup is not accessible enough?

Shnoulle:
> Hm, the default Bootstrap 5 buttongroup is not accessible enough?

Here it's for current master … 

Seems it's same for BS5 

Level A criterion : https://www.w3.org/TR/WCAG21/#use-of-color

> Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.

Shnoulle:
BS 5 have same issue

Capture d’écran du 2022-10-24 11-38-24

olleharstedt:
Ping @ptelu

Shnoulle:
Reported for BS5 : https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use#w_custom-font-color

ptelu:
> 

If i get this whole discussion right, you want to add an indicator that button groups are selected when no colors are used?
We will then have to discuss this internally and how it could fit into the new design.

Shnoulle:
It's only a criterion level A … not a AA …

olleharstedt:
> It's only a criterion level A … not a AA …

What does this mean? You have a link to explain it?

Shnoulle:
Even with color :  green, or red or blue for checked or not … maybe OK when you have 3 radio, but what for 2 ?

ptelu:
We haven't done an accessibility overhaul yet but this is something that should then come up as well, so I'm not sure if this is the right time to introduce it.

Shnoulle:
> > It's only a criterion level A … not a AA …
> 
> What does this mean? You have a link to explain it?

https://www.w3.org/WAI/WCAG2A-Conformance

About WCAG : https://www.w3.org/WAI/standards-guidelines/wcag/

For Europe : https://en.wikipedia.org/wiki/EN_301_549 and https://digital-strategy.ec.europa.eu/en/policies/web-accessibility

WCAG is used in company, in France : we have RGAA , it's a partial WCAG but this is required for all website and tool for government, affiliated to government or big company by law.

Shnoulle:
> We haven't done an accessibility overhaul yet but this is something that should then come up as well, so I'm not sure if this is the right time to introduce it.

This is really a microfix here, only css …

Shnoulle:
@olleharstedt or @ptelu i keep it only for User admin template or i fix conflict ?

ptelu:
I think this is still a good idea, brought it up if the design is okay. Also it would have to be redone for admin_theme as well since we are using different files/html. Will let you know once i know the result.

Shnoulle:
> I think this is still a good idea, brought it up if the design is okay. Also it would have to be redone for admin_theme as well since we are using different files/html. Will let you know once i know the result.

Still issue in 6.dev
Capture d’écran du 2023-03-17 09-22-51

olleharstedt:
@ptelu Can you create an internal task for this perhaps?

ptelu:
created a ticket internally so we dont forget it CR-1200

olleharstedt:
Hooray!

Shnoulle:
PS : i use this one in my admin theme https://github.com/LimeSurvey/LimeSurvey/pull/2359
And it's really better

c-schmitz:
This will be picked up again, very soon.

PR #2480, opened 2022-06-27

Shnoulle:
Dev: inline style solution
Dev: force in one column by theme
Dev: work for any column number


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2481

PR #2483, opened 2022-06-28
Fixed issue #18203: Return arrays instead of strings from `update_response`

edgarrmondragon:
Fixed issue #18203: https://bugs.limesurvey.org/view.php?id=18203


Shnoulle (approved):

Shnoulle:
Maybe we need a get_limesurvey_version function in API https://api.limesurvey.org/classes/remotecontrol_handle.html 

For future update …

Shnoulle:
> > Maybe we need a get_limesurvey_version function in API
> 
> @Shnoulle there's [`get_site_settings("versionnumber")`](https://api.limesurvey.org/classes/remotecontrol_handle.html#method_get_site_settings). Did you have something different in mind.

No … you're right :)

olleharstedt:
Thanks!

PR #2486, opened 2022-06-30
Task/l41 t1767 panels

twilligls:
Dev:


olleharstedt:
```css
input[type="file"].form-control {
    border: none;
    padding-left: 0px;
}
```

This breaks input style everywhere. When back from vacation, can you tell me why you needed it? I'll remove it for now.

twilligls:
I bet there was a reason, but I cannot remember at all. If it still looks ok after you removed it, I'd say we keep it removed then.

PR #2487, opened 2022-06-30
Lsc 25 ccu21 wells

twilligls:
Dev:


application/views/admin/survey/editDataSecurityLocalSettings_view.php
```diff
@@ -29,6 +29,7 @@
 <div id="editdatasecele-<?php echo $i; ?>" class="tab-pane fade <?php if ($i == 0) {
     echo "show active";
 } ?> center-box">
+<div class="row">
```
Shnoulle, line 32: started not closed ?

twilligls: It was closed but not started (accidentally deleted, I guess), So now it's started and closed.

Greetings from Hamburg

PR #2489, opened 2022-06-30
Dev #18099: Automatic testing for navigation buttons

gabrieljenik:


tests/functional/frontend/SurveyNavigationTest.php
```diff
@@ -0,0 +1,164 @@
...
+
+        try {
+            // Open survey.
+            $web->get($url);
+
+            // Move from Welcome to first group
+            // Click next.
+            $web->clickButton('ls-button-submit');
+
+            // Wait max 10 second to find the first group title
```
Shnoulle, line 46: 10 seconds ???? It's a maximum ? If we get it before ?

Shnoulle:
Just comment, code is OK.

Adding index to 1, then we don't have to update lss when we have to add index navigation check.

PR #2491, opened 2022-06-30
Fixed issue #18200: User with only reponse export can export Survey + token (survey archive)

gabrieljenik:


application/views/admin/survey/surveybar_displayexport.php
```diff
@@ -22,7 +22,11 @@
         "href" => $this->createUrl("admin/export/sa/survey/action/exportstructurexml/surveyid/".$oSurvey->sid),
         "download" => true
     ];
-    if(($respstatsread && $surveyexport)) {
+    if(
+        $respstatsread
+        && $surveyexport
+        && (!$oSurvey->hasTokensTable || Permission::model()->hasSurveyPermission($oSurvey->sid, 'tokens', 'export'))
```
Shnoulle, line 28: :+1:

application/controllers/admin/export.php
```diff
@@ -963,6 +969,13 @@ private function _exportarchive($iSurveyID, $bSendToBrowser = true)
     {
         $survey = Survey::model()->findByPk($iSurveyID);
 
+        if (
+            ($survey->hasTokensTable && !Permission::model()->hasSurveyPermission($iSurveyID, 'tokens', 'export'))
```
Shnoulle, line 973: Private function, but if you add Permission control here, you must control surveycontent/export too :)

Shnoulle (approved):
Since you add Permisiion control on the private function : add surveycontent check too.

application/controllers/admin/export.php
Shnoulle: OK done their : https://github.com/LimeSurvey/LimeSurvey/blob/24d240d3a8a94086b4ff8fe00c32e507c72dcaf0/application/controllers/admin/export.php#L41

this file need a rewrite …

application/controllers/admin/export.php
```diff
@@ -867,32 +867,38 @@ public function exportMultipleSurveys($sSurveys, $sExportType)
...
-                                $bArchiveIsEmpty                = false;
-                                $archiveFile                    = $archiveName;
-                                $newArchiveFileFullName         = 'survey_archive_'.$iSurveyID.'.lsa';
-                                $this->_addToZip($zip, $archiveFile, $newArchiveFileFullName);
-                                unlink($archiveFile);
+                        if (
+                            ($oSurvey->hasTokensTable && !Permission::model()->hasSurveyPermission($iSurveyID, 'tokens', 'export'))
+                            || !Permission::model()->hasSurveyPermission($iSurveyID, 'responses', 'export')
+                        ) {
+                            $aResults[$iSurveyID]['error'] = gT("We are sorry but you don't have permissions to do this.");
```
Shnoulle, line 883: I think a single work are better : No permission
Capture d’écran du 2022-07-01 17-27-31

But i merge it like that (security issue)

PR #2493, opened 2022-06-30
Fixed issue #18183: Date filter at statistics is not applied correctly

gabrieljenik:
Making the equal filter to use a date picker (just date).
Later on the PHP side, we query by columnanme in between the start and midnight of the filter date.

### DIFF
--- a/application/views/admin/export/statistics_subviews/_question.php
+++ b/application/views/admin/export/statistics_subviews/_question.php
@@ -277,7 +277,7 @@
                     'id' => $myfield3,
                     'value' => isset($_POST[$myfield3]) ? $_POST[$myfield3] : '',
                     'pluginOptions' => array(
-                        'format' => $dateformatdetails['jsdate'] . " HH:mm",
+                        'format' => $dateformatdetails['jsdate'],
                         'allowInputToggle' =>true,
                         'showClear' => true,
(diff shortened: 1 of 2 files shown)


application/views/admin/export/statistics_subviews/_question.php
```diff
@@ -277,7 +277,7 @@
                     'id' => $myfield3,
                     'value' => isset($_POST[$myfield3]) ? $_POST[$myfield3] : '',
                     'pluginOptions' => array(
-                        'format' => $dateformatdetails['jsdate'] . " HH:mm",
```
Shnoulle, line 280: Keep it ?

Shnoulle:
Fixed for eq but font for less and more ? https://github.com/LimeSurvey/LimeSurvey/blob/415fd87e020cfb156c736bb4dc06c9e22fac25c1/application/helpers/admin/statistics_helper.php#L473-L480

@maziminke issue was : « E.g. if I define to show data sets with a date between January 1st, 2021 and December 31st, 2021»

Shnoulle (approved):

Shnoulle:
@maziminke time for testing ?

maziminke:
Looks good to me!

PR #2494, opened 2022-06-30
Fixed issue #18143: cannot import resources for imported (lsa) survey with uploaded resources in responses

gabrieljenik:
Getting the extension from the mime type was not that complicated (for this case).

### DIFF
--- a/application/core/Survey_Common_Action.php
+++ b/application/core/Survey_Common_Action.php
@@ -1220,8 +1220,18 @@ class Survey_Common_Action extends CAction
         while ($direntry = readdir($dh)) {
             if ($direntry != "." && $direntry != "..") {
                 if (is_file($extractdir."/".$direntry)) {
-                    // is  a file
+                    // Current entry is a file. Check if extension is allowed.
+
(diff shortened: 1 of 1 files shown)


application/core/Survey_Common_Action.php
```diff
@@ -1220,8 +1220,18 @@ protected function _filterImportedResources($extractdir, $destdir)
...
                 if (is_file($extractdir."/".$direntry)) {
-                    // is  a file
+                    // Current entry is a file. Check if extension is allowed.
+
+                    // Get extension from file name
                     $extfile = (string) substr(strrchr($direntry, '.'), 1);
+
+                    // Workaround for files comming from file upload questions, which don't have an extension.
+                    // If the extension couldn't be determined, and the folder is "files", try to get the extension
+                    // from mime type.
```
Shnoulle, line 1230: fu_***

c-schmitz:
> I don't think it's the way to fix : fu_ file is NOT part of survey ressources … it's part of survey responses.

I agree. I don't think this should be allowed.

PR #2497, opened 2022-07-01
Fixed issue #18220: Question image theme don't seems to work

gabrieljenik:


Shnoulle:
It's a file rename issue ?
Something to be done for other QuetsionTheme too ? for example for @tpartner QuestionTheme ?

API version update ?

gabrieljenik:
The name of the directory where they were located didn't match the name of the question theme.

tpartner:
If the file name changes are only for the image-select themes, it will not affect any of my question themes.

Shnoulle (approved):

Shnoulle:
@gabrieljenik : can you put some dev comments on your commit ? Here : i added the rename directory :)

I think it's best when you look at git blame and history

c-schmitz:
We had to revert this patch. It was causing existing multiple choice question to turn into image-select and image-select to not work anymore, accroding to customers.

mwehr:
Hi,

I think what's missing here is to change the path in the DB. 
Table "question_theme" field xml_path needs also to be changed.

Regards
Mario

gabrieljenik:
Will review! Thanks

PR #2500, opened 2022-07-05
Fixed issue #18141: changing display mode from extended to compact in survey responses windows causes freeze

gabrieljenik:


Shnoulle:
I don't really understand the code … i dislike to use off/on …

But since it was already there …

Shnoulle (approved):

PR #2502, opened 2022-07-07
Fix bug  Remote Control export statistic for certain type of fields

prob6:
Fixed issue #<Mantis issue number>:18049
New feature #<Mantis issue number>:18049
Dev:

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -1183,7 +1183,7 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
 
                 //go through all the (multiple) answers
                 foreach ($result as $row) {
-                    $myfield2 = $flt['type'] . $myfield . reset($row);
+                    $myfield2 = $flt['type'].$myfield .$row['title'];
                     $allfields[] = $myfield2;
                 }
                 break;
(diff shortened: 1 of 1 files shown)


gabrieljenik:
Shouldn't this solution be applied for Dual Scale as well?

gabrieljenik:
Please, review the style guidelines as to follow PRS-12.
(spaces after dot)

olleharstedt:
Is createCompleteSGQA only used by remote API?

gabrieljenik:
> Is createCompleteSGQA only used by remote API?

It is.

Still, If I am right the root cause of the problem is that the parameter fields is never used here.
https://github.com/prob6/LimeSurvey/blob/a34c39ecf400599f25806db2e239053bc29af4ac/application/models/Question.php#L512

So, if this patch gets applied and after that `getQuestionsForStatistics()` starts using the $field parameter inside the code, the code will be broken again, as the code will look for fields, which are not available.

gabrieljenik:
Part of this was already solved here on 18814: https://github.com/LimeSurvey/LimeSurvey/commit/3fb81e51289b6ce90cd2d3749bcc62b2bbf699b5

For some question types is pending as there is no automated test yet.

c-schmitz:
Closing due to lack of feedback from author. Also will most likely be obsolete with field name changes in v7.

PR #2503, opened 2022-07-09
Fixed issue #18229: Faulty message on numeric questions

Shnoulle:
Dev: add the DB validity message only if question is not valid

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -6378,18 +6378,20 @@ class LimeExpressionManager
         foreach ($sgqas as $sgqa) {
             $validityString = self::getValidityString($sgqa);
             if ($validityString && $qrel && !$qhidden) {
-                /* Add the string to be showned , no js error or another class ? */
-                $stringToParse .= App()->twigRenderer->renderPartial(
-                    '/survey/questions/question_help/error_tip.twig',
-                    [
-                        'qid'       => $qid,
-                        'coreId'    => "vmsg_{$qid}_dberror",
-                        'vclass'    => 'dberror',
-                        'coreClass' => 'ls-em-tip em_dberror',
-                        'vtip'      => $validityString,
-                    ]
-                );
-                /* Set this question invalid (only if move next due to $force) */
+                /* Add the string if current question is valid , see #18229: Faulty message on numeric questions */
+                if ($qvalid) {
+                    $stringToParse .= App()->twigRenderer->renderPartial(
+                        '/survey/questions/question_help/error_tip.twig',
+                        [
+                            'qid'       => $qid,
+                            'coreId'    => "vmsg_{$qid}_dberror",
+                            'vclass'    => 'dberror',
+                            'coreClass' => 'ls-em-tip em_dberror',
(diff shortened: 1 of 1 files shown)


Shnoulle:
`   │ PHPUnit\Framework\ExpectationFailedException: Title validation in question-title-warning are not empty on success, get “The question code is mandatory.”` ?

Shnoulle:
Must test

- [x] numeric question with AA : OK show "number error"
- [x] numeric question with with 1.1234567890123456789 : show validity error
- [ ] index > 1  (see forced) : unsure on needed result
- [ ] date question with invalid date (2022-02-31)
- [ ] Another idea ?

gabrieljenik:
I don't understand this fix. Sorry.
I thought the error_tip should always be shown from PHP side. Then JS makes it red or green. Right?
Why now only when is valid?

Also the fix is said to be only for numeric, but the change is done on something which looks general, right?

Code change is OK from structure point of view.

Shnoulle:
> Also the fix is said to be only for numeric, but the change is done on something which looks general, right?

Yes, but numeric is the only way to have an issue without try to hack HTML (currently)

The other is more, for example : when you try to send a bad value (for example 10 on a 5 choice question) etc … but sibnce question already have error : no need to show the validity error (it's : `You enter an invalid value for such data`)

If question is not valid: user can not go to next group and have an error : no real reason to show error.

See the original issue : https://bugs.limesurvey.org/view.php?id=10827 and fix : https://github.com/LimeSurvey/LimeSurvey/pull/533

In general : i dislike to disable by exception (without params). 

>  I thought the error_tip should always be shown from PHP side. Then JS makes it red or green. Right?

It's totally unrelated to javascript here : validity error is inside `alert alert-danger` block, not managed by Expression Manager.

Shnoulle:
The important think is : show something to user if can not move to next page. Expression error if there are. Validity error if not. I really think we don't need to show both.

Shnoulle:
@gabrieljenik i don't know how to explain differently …

After expression manager check validity of an answer, there are a final validation to check if the value is OK to send in database (and we didn't save it in DB in such case). It's happen for single choice (but then user need to hack HTML) but can happen to numercial too without hacking 

Currently if there are an EM error + a DB validty error : we show bith.
With this fix : if there are an EM error, we don't show DB validty error

gabrieljenik:
OK.
`qValid` = TRUE if EM Validation is OK.

`$validityString = self::getValidityString($sgqa);`
Will have something if question didn't pass core validation rules.

You are enforcing Validation core rule messages to be shown only if EM Validations have been passed OK.
Makes sense.

gabrieljenik (approved):
Code OK
Haven't tested it

Shnoulle:
> You are enforcing Validation core rule messages to be shown only if EM Validations have been passed OK, right?

No : i always validate (else it brake with 500 DB error when save to DB) 

but i show it only if there are no other errors.

gabrieljenik:
> but i show it only if there are no other errors.

That's what I was saying :)

Shnoulle:
> That's what I was saying :)

Read too quick :)

PR #2504, opened 2022-07-11

gabrieljenik:


Shnoulle:
Don't ask mle to review something i'm not OK.

Else : if you have a aria-labelledby : you need a role=group

Shnoulle:
Need more work 

I test a solution with role="presentation" currently : 

But with this one 
_Error: The role attribute must not be used on a tr element which has a table ancestor with no role attribute, or with a role attribute whose value is table, grid, or treegrid._
then : remove the role="radiogroup" on each line.
Then : need fieldset for a11Y …

Shnoulle (commented):
validity issue and it's not a tabular information

See : https://www.w3.org/WAI/WCAG21/Techniques/html/H51

Closed by https://github.com/LimeSurvey/LimeSurvey/pull/2512

Shnoulle:
Capture d’écran du 2022-07-26 16-42-23

gabrieljenik:
> Capture d’écran du 2022-07-26 16-42-23

If I am right, the issues reported on that image are not items updated on this ticket.

Shnoulle:
> If I am right, the issues reported on that image are not items updated on this ticket.

Without this fix : there are no issue because table ancestor have role="group" attribute :)

**Only appear after this fix** : because you directly remove the role="group"

I replace by role="presentation" in https://github.com/LimeSurvey/LimeSurvey/pull/2512

gabrieljenik:
> > If I am right, the issues reported on that image are not items updated on this ticket.
> 
> Without this fix : there are no issue because table ancestor have role="group" attribute :)
> 
> **Only appear after this fix** : because you directly remove the role="group"

OK, I got it.

> I replace by role="presentation" in #2512

OK... I think you did more than that  :)
I would have just stopped there on replacing with role = presentation
Was the other part needed right now?

Shnoulle:
> OK... I think you did more than that :)
> I would have just stopped there on replacing with role = presentation
> Was the other part needed right now?

Yes … because with role="presentation"  : you must remove th and thead for WCAG (the issue about script and sureytheme cc @tpartner ) . Validator throw error about invalid th…

(and for RGAA (french) : you must remove col too … but i realy can not remove col (for width))

gabrieljenik:
> you must remove th and thead for WCAG

Can that be done a on a question theme?

tpartner:
> > you must remove th and thead for WCAG
> 
> Can that be done a on a question theme?

Or in an "accessible" survey theme. (at least until 6.x)

Shnoulle:
Find the issue : https://www.w3.org/TR/WCAG20-TECHS/F92.html

> > you must remove th and thead for WCAG
> 
> Can that be done a on a question theme?

Yes and no … it can be done in SurveyTheme : but then : next time a fix was done in core : theme is broken.
It's the reason i don't want to do it for SkelVanilla but on user theme OR core …

Shnoulle:
> Or in an "accessible" survey theme. (at least until 6.x)

It's clearly already awfull to maintain skelvanilla to have 3.X and 5.X compatibility . I don't do it for skelvanilla (except on a specific fork)

I show to Gabriel and Oll where to update the theme.

gabrieljenik:
ping @olleharstedt as to rule on this :) Thanks

gabrieljenik:
Closed as this was replaced by #2512

PR #2509, opened 2022-07-13
Fixed issue #18106: Missing survey ID replacements at registration sc…

gabrieljenik:


Shnoulle (approved):
:+1:

Shnoulle:
No need to StartPage/endpage ?

Shnoulle:
> You are the EM master.... I was looking for that kind of advice... :)

I think it must work without , but ? 
I already make commit with big issue on Expression … :/

PR #2058, opened 2021-09-29
Fixed issue #16480: Empty logo causes under specific circumstances loops on the welcome page

gabrieljenik:

PR #2510, opened 2022-07-13
Fixed issue #18246: The setting in the theme variation field (global level) is not shown

gabrieljenik:
The dropdown value was blank because the selected value had a prefix and the options did not.
Removes the prefix to the saved value.

cssframework (variation) shouldn't be prefixed as logos do, for example.
They are always relative to the theme.

### DIFF
--- a/application/models/TemplateConfiguration.php
+++ b/application/models/TemplateConfiguration.php
@@ -1593,12 +1593,18 @@ class TemplateConfiguration extends TemplateConfig
      */
     public function sanitizeImagePathsOnJson($attribute, $params)
     {
+        $excludedOptions = [
+            'cssframework'
+        ];
         // Validates all options of the theme. Not only classic ones which are expected to hold a path,
         // as other options may hold a path as well (eg. custom theme options)
         $decodedOptions = json_decode($this->$attribute, true);
         if (is_array($decodedOptions)) {
             Yii::import('application.helpers.SurveyThemeHelper');
-            foreach ($decodedOptions as &$value) {
+            foreach ($decodedOptions as $option => &$value) {
+                if (in_array($option, $excludedOptions)) {
+                    continue;
+                }
                 $value = SurveyThemeHelper::sanitizePathInOption($value, $this->template_name, $this->sid);
             }
             $this->$attribute = json_encode($decodedOptions);


Shnoulle:
> cssframework (variation) shouldn't be prefixed as logos do, for example.

And for other ? How user theme can disable prefix ?

See : https://gitlab.com/SondagesPro/SurveyThemes/skelvanilla/-/blob/master/views/subviews/header/theme.twig#L6

(i don't understand why we don't save ./ for prefix …)

gabrieljenik:
> And for other ? How user theme can disable prefix ?

The prefix is added only to the `options` attribute.

> See : https://gitlab.com/SondagesPro/SurveyThemes/skelvanilla/-/blob/master/views/subviews/header/theme.twig#L6
> (i don't understand why we don't save ./ for prefix …)

This was a mechanism requested by @c-schmitz to not store paths that access directories outside of the LS Root reach.
https://bugs.limesurvey.org/view.php?id=16480#c65206
https://github.com/LimeSurvey/LimeSurvey/pull/2058

Shnoulle:
> > And for other ? How user theme can disable prefix ?
> 
> The prefix is added only to the `options` attribute.

I don't understand ? All SurveyTheme are in `options` attribute no ?
It apply to each dropdown? 

Here : you disable for `cssframework` , i ask how to diable for other dropdown 
Or : how to enable ONLY for specific dropdown.

> 
> > See : https://gitlab.com/SondagesPro/SurveyThemes/skelvanilla/-/blob/master/views/subviews/header/theme.twig#L6
> > (i don't understand why we don't save ./ for prefix …)
> 
> This was a mechanism requested by @c-schmitz to not store paths that access directories outside of the LS Root reach. https://bugs.limesurvey.org/view.php?id=16480#c65206 #2058

« Fixed issue #16480: Empty logo causes under specific circumstances loops on the welcome page» ?

But it's not the question here :)

To disable access out of limesurvey directory : it's more easy to check if realpath($file) start by config("uploadidr")

`if(!is_dir($dirPath) || substr($dirPath, 0, strlen($baseDir)) !== App()-getConfig("uploaddir"))`

gabrieljenik:
> Here : you disable for cssframework , i ask how to diable for other dropdown

cssframework  is not stored on the options attribute, but is kind of set on the include files if remember correctly.

application/models/TemplateConfiguration.php
```diff
@@ -1593,12 +1593,18 @@ public function setOptionKeysToInherit()
      */
     public function sanitizeImagePathsOnJson($attribute, $params)
     {
+        $excludedOptions = [
+            'cssframework'
```
Shnoulle, line 1597: You add it here : how add other ?

**Need** a global solution. Something than can be done by config.xml or HTML options.js

Shnoulle:
Ok,

> > Here : you disable for cssframework , i ask how to diable for other dropdown
> 
> cssframework is not stored on the options attribute, but is kind of set on the include files if remember correctly.

? You nadd it hardly here : i ask for an option, a settings somewhere.

Maybe in config.xml or something like that …

Shnoulle (changes requested):
Need to be in config.xml or elsewhere

application/models/TemplateConfiguration.php
gabrieljenik: It could be a global config setting?
I would say to add a new ticket so we can discuss it there.

Shnoulle: Again a ,new ticket …

Hacking is really more the quickest to improve LS

gabrieljenik: Hacking is the quickest way to loose track of changes, arguments and decisions. Also not very helpfull for coordinating in a team.

gabrieljenik:
> See SQL : need a way to disable (or enable if not enabkle by default) on any dropdown.

Why do you need to disable it?
Is a security measure.

Shnoulle:
> > See SQL : need a way to disable (or enable if not enabkle by default) on any dropdown.
> 
> Why do you need to disable it? Is a security measure.

I don't need it on variation for example …
Dropdown is not ONLY used for image … can be used for a lot of other things …

application/models/TemplateConfiguration.php
Shnoulle: I'm OK, but i have fix since month now … and i have to give proof of reach of my sentence.

gabrieljenik:
> I don't need it on variation for example … Dropdown is not ONLY used for image … can be used for a lot of other things …

That dropdown is kind of special as it is holding a path. That's why is impacted.

Shnoulle:
> That dropdown is kind of special as it is holding a path. That's why is impacted.

And ? Still need a way to disable for specific usage … no ?

gabrieljenik:
> > That dropdown is kind of special as it is holding a path. That's why is impacted.
> 
> And ?

I was replying why that dropdown is being impacted.

> Still need a way to disable for specific usage … no ?

If the option holds a path, it is being prefixed.
If a surveytheme can disable that then it can expose internal paths in the server that the admin is not willing to.
That's why what you are requesting needs to be run through @c-schmitz first.

Taking that into account, I would say that config,xml is not at the admin level to decide, but at user level (as anyone can upload a theme). So, exceptions could be set as config.php (or similar) settings

gabrieljenik:
> Need to be in config.xml or elsewhere

I believe this change request is not related to the bug resolution but to the path prefix system itself.

Shnoulle:
>  If a surveytheme can disable that then it can expose internal paths in the server that the admin is not willing to.

**A theme allowed user can alway put anything in path of image source :  imageSrc("../../../myimage.png")** not restricted to the dropdown …

Even if ../../../myimage.png exist and can be readed by web user it give a link. ` `

Same for registerScriptFile : it create a link to a file: `<script src="/../../../password.ini" defer="defer"></script>` : no security issue
Or registerCss (didn't exit in twig) `<link rel="stylesheet" href="/../../../php.ini"  />`

There are **NO security issue to create link** … (except for 404 error).

The potential issue can be with [include](https://github.com/LimeSurvey/LimeSurvey/blob/master/themes/survey/vanilla/views/layout_global.twig#L115)

Shnoulle:
> > Need to be in config.xml or elsewhere
> 
> I believe this change request is not related to the bug resolution but to the path prefix system itself.

It's **totally** related … core theme need it , then User theme can need it for other dropdown.

gabrieljenik:
Without the prefix mechanism, if uploads directory is outside of LS root, the paths saved could include traversal across directories outside LS root. That could expose some not intended information.

I am not against having an exception mechanism but let's just wait for Carsten to throw his comments here.
He asked for that mechanism.

olleharstedt:
What's going on here? :sweat_smile: Such a long discussion.

olleharstedt:
Can I merge this, and then we can discuss further improvements in a new ticket? Or?

Shnoulle:
Do what you want …

But again : 

1. give me proof there are security issue about img src ``
2. it **must** be allowed to disable the rewriting when needed … (my opinion : it must be disable by default for all dropdown)

application/models/TemplateConfiguration.php
c-schmitz: > I'm OK, but i have fix since month now … and i have to give proof of reach of my sentence.

which fix?

Shnoulle: Still have some, don't remind of the one related here.

PR #2515, opened 2022-07-14
Fixed issue #18145: Data integrity: Cannot press button when only participant table is selected

gabrieljenik:


Shnoulle:
Personnaly : i'm for diable all JS system here … don't set the button to disable state at start.

If nothing is checked : nothing is done.

Like 3.X : why adding uneeded JS system. 3.X is OK here 
integrity-3X

gabrieljenik:
@c-schmitz @olleharstedt Thoughts?

Shnoulle:
> @c-schmitz @olleharstedt Thoughts?

See : https://github.com/LimeSurvey/LimeSurvey/pull/2533

olleharstedt:
Ehm, I think disabling buttons that do nothing is a bit more UX friendly.

Shnoulle:
> Ehm, I think disabling buttons that do nothing is a bit more UX friendly.

I really don't think since the button is after the selection. Maybe adding customValidity then. Because disable without explain is worst.

Else, if we still disable/enable : since we already have a function : seems better to use a` $("#data-redundancy-list :checkbox").on(…` for example

olleharstedt:
Disabled buttons can have a tooltip to explain why, we do this in some places. Not sure it's worth it here...

Shnoulle:
> Disabled buttons can have a tooltip to explain why, we do this in some places. Not sure it's worth it here...

tooltip is not shown … see https://designsystem.digital.gov/components/tooltip/  don't use tooltip !

> Helpful, non-critical information. Use tooltips to strengthen an existing message.
> Enhance confidence. Use tooltips to increase certainty about an interaction.
> Brief descriptions. Tooltips perform best with succinct helper text.
> Lack of space. Tooltips are useful as a last resort for space-constrained UI. Explore other options for keeping content visible without a tooltip.

The best is to use customValidity and validity form (and not disable button)

Here it work and nobody ask to have an enable/disable system since years !It's not better 

See here for advice https://uxplanet.org/disabled-buttons-in-user-interface-4dafda3e6fe7

Shnoulle:
Fixed in https://github.com/LimeSurvey/LimeSurvey/pull/2533

PR #2518, opened 2022-07-18
Fixed issue #17488: Cannot view survey after updating to version 3.279 with custom twig extension (breaks survey)

gabrieljenik:


gabrieljenik:
Tested by @tpartner .
@Shnoulle Can I ask for your reviewal?

application/core/LSETwigViewRenderer.php
```diff
@@ -812,4 +830,17 @@ public function getLoader()
...
     }
+
+    /**
+     * Adds custom user extensions
+     * @param array $extensions @see self::$user_extensions
+     */
+    public function addUserExtensions($extensions)
+    {
+        foreach ($extensions as $extName) {
+            Yii::setPathOfAlias('extName_' . $extName, Yii::app()->getConfig('usertwigextensionrootdir') . '/' . $extName . '/');
```
Shnoulle, line 841: No need to filter $extName here ?
It's a public function : then can be used by any plugins

Maybe `Yii::setPathOfAlias` is secure :)

Shnoulle (commented):
1. Need to sanitize $extName
2. addUserExtensions must be public ?

application/core/LSETwigViewRenderer.php
Shnoulle: I didn't see it in master …

gabrieljenik: Sanitize against what? What kind of sanitization?

Shnoulle: file/directory trasversal

Shnoulle: setPathOfAlias didn't sanitize directory : https://github.com/yiisoft/yii/blob/master/framework/YiiBase.php#L391

Then extension dev can set '../../../../etc/passwd' to path of alias (addUserExtensions is a public function, and user_extensions is a public var too.)

gabrieljenik: Yes, we could sanitize that. Not much of a hassle. Not sure it is worth it.
Still, after sanitizing, they could do an suspicious include in the new extension.

@olleharstedt Thoughts?

olleharstedt: Probably a good idea to filter, BUT: It also depends on the assumed access to even use this feature. If file system access is needed to use the feature, filtering won't matter. Still, yeah, why not.

Shnoulle: I put discussion … i thihnk we can do it after . Not required update , but seems better in my opionon

Shnoulle (approved):

application/core/LSETwigViewRenderer.php
Shnoulle: It still public function.

1. In controller : we must always filter/check access etc on all public function
2. In [twig function](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LS_Twig_Extension.php) : we must filter all data if we include something (can be used by Template manager, remind the issue with [getAllAnswers](https://github.com/LimeSurvey/LimeSurvey/commit/69784616b81594d4fd79b0e018d46d461161cffc) )
3. Here ? Unsure …

gabrieljenik: I think it doesn't worth it.
This is to include files, right?
If someone want's to include a suspicious file, they can include a valid file here and then include a dangerus one on the valid include.

Shnoulle: Yes : «here ? Unsure … » ;)

PR #2520, opened 2022-07-19
Fixed issue #18247: php version 8 and remote control, error in bigdata.php

gabrieljenik:


Shnoulle (approved):

Shnoulle:
Code seems OK, but need test

gabrieljenik:
@olleharstedt Do you think we should push this to develop to mature and get more testing time?

olleharstedt:
Hmmm, is BigData an external dependency? Carsten has been moving things to composer.json, so that we can finally keep our dependencies up-to-date.

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/blob/49edd10700501807bc08bd7b434f309076992158/application/libraries/BigData.php#L5

If need to set as dependency : need to create a repo ?

olleharstedt:
Ah, by Sam. Yeah, then it's probably not on Packagist. :) (Checked, couldn't find it.)

gabrieljenik:
> Ah, by Sam. Yeah, then it's probably not on Packagist. :) (Checked, couldn't find it.)

So...? not sure where that leaves us

olleharstedt:
If you feel uncertain about the fix, pushing it to dev and wait is not gonna help much, I think. Rather, we need to discuss the different risks, what could break, and how to test it.

gabrieljenik:
> If you feel uncertain about the fix, pushing it to dev and wait is not gonna help much, I think. Rather, we need to discuss the different risks, what could break, and how to test it.

Not uncertain, but as it is hard to test ...

olleharstedt:
Can you write down how it should be tested?

Shnoulle:
When it's used ? :rofl:

gabrieljenik:
Actually, I think that if automated tests on PHP8 went good, I guess everything is OK,
Look at the reproduction steps. Just trying to get session key was triggering the error.

Question is, why haven't that error pop up before as CI runs PHP 8.

Thoughts?

olleharstedt:
> Look at the reproduction steps. Just trying to get session key was triggering the error.

Ah ok!

PHP 8.0 or 8.1?

c-schmitz:
The fix is fine - there is no need to pass by reference here as the value if $result is never manipulated direclty or indirectly in that function.

PR #2522, opened 2022-07-19
Fixed issue #18194: Link after import survey goes to subquestion edit

gabrieljenik:


Shnoulle (approved):
:+1: 

Still think it better to return to Survey management (like 3.X) but fix the issue :)

Shnoulle (approved):
:+1: better fix

PR #2525, opened 2022-07-20
Dev: Update testing dependencies to match v5, enable testing php8.0 for v3

TonisOrmisson:
Dev:

Synced V3 to V5 regarding testing dependencies. 
With this, the tests can now also run on php8.0


olleharstedt:
Ehm, 3.x will be discontinued soon enough, and LS5 will be the new LTS. Not sure this change is needed.

tiborpacalat:
Thank you very much for your contribution, but after a long discussion, we decided to not take this change into our code base based on the fact that Limesurvey 3.x will be discontinued soon.

PR #2533, opened 2022-07-21
Fixed issue #18145: Data integrity: Cannot press button when only participant table is selected

Shnoulle:
Dev: remove the disable and the javascript.
Dev: cleaner, and less complex. No issue if clicked


Shnoulle:
Capture d’écran du 2022-07-21 12-12-31

PR #2534, opened 2022-07-21
Fixed issue #18106: Missing survey ID replacements at registration screen

gabrieljenik:


c-schmitz:
This patch seems to cause a bug in LS3 that the whole response is lost. Do NOT merge for master.

c-schmitz:
Calling 
  $LEM =& LimeExpressionManager::singleton();
        $LEM->setVariableAndTokenMappingsForExpressionManager($iSurvey

only works in  a clean way if the survey is already running.
If not (this happens on a survey register screen) it is causing to create an incomplete session - which then in turn does not save any response data eventhough the participants can navigate the whole survey

Shnoulle:
Sh … https://github.com/LimeSurvey/LimeSurvey/pull/2509#issuecomment-1183282669

StarSurvey/Startpage needed …

gabrieljenik:
Reverted on LTS. So closing on Master

PR #2539, opened 2022-07-25
Failed Email Notifications

ptelu:
new feature failed email notifications, catches sending emails to the mail server fails.


application/models/FailedEmail.php
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /**
+     * @return string
+     * @throws CException
+     * @throws CHttpException
+     */
+    public function getRawMailBody(): string
+    {
+        $mailer = \LimeMailer::getInstance(true);
```
olleharstedt, line 199: This line is not correct, expects int, 0 1 or 2, not boolean; I think you can leave it empty and it would default to 1, right?

application/helpers/frontend_helper.php
```diff
@@ -529,30 +535,90 @@ function sendSubmitNotifications($surveyid)
...
     }
+    if ($return) {
+        return [
+            'successfullEmailCount' => $successfullEmailCount,
+            'failedEmailCount'      => $failedEmailCount,
+        ];
+    }
+}
+
+function saveFailedEmail($id, $recipient, $surveyId, $emailType, $language, $errorMessage) {
```
olleharstedt, line 587: Missing docs and return type; some PSR-12 issues.

application/helpers/frontend_helper.php
```diff
@@ -529,30 +535,90 @@ function sendSubmitNotifications($surveyid)
...
+    $failedEmailModel->email_type = $emailType;
+    $failedEmailModel->language = $language;
+    $failedEmailModel->error_message = $errorMessage;
+    $failedEmailModel->created = date('Y-m-d H:i:s');
+    $failedEmailModel->status = 'SEND FAILED';
+    $failedEmailModel->updated = date('Y-m-d H:i:s');
+
+    return $failedEmailModel->save(false);
+
+    Yii::log(\CVarDumper::dumpAsString($model->getErrors()), 'warning', 'application.models.FailedEmail.insertRecords');
```
olleharstedt, line 609: Code not reachable?

application/controllers/FailedEmailController.php
```diff
@@ -0,0 +1,179 @@
...
+        }
+        return $this->renderPartial('/admin/super/_renderJson', [
+            "data" => [
+                'success' => false,
+                'message' => gT('Please select at least one item'),
+            ]
+        ]);
+    }
+
+    public function actionModalContent(): string
```
olleharstedt, line 167: Missing permission check?

application/controllers/FailedEmailController.php
```diff
@@ -0,0 +1,179 @@
...
+    {
+        $contentFile = App()->request->getParam('contentFile');
+        $id = App()->request->getParam('id');
+        $failedEmailModel = new FailedEmail();
+        $failedEmail = $failedEmailModel->findByPk($id);
```
olleharstedt, line 172: Missing check for null, throw 404 website exception or what it's called

application/controllers/FailedEmailController.php
```diff
@@ -0,0 +1,179 @@
...
+            $failedEmails = FailedEmail::model()->findAll($criteria);
+            if (isset($failedEmails)) {
+                foreach ($failedEmails as $failedEmail) {
+                    $emailsByType[$failedEmail->email_type][$failedEmail->id] = $failedEmail->recipient;
+                }
+                if (isset($emailsByType)) {
+                    $result = sendSubmitNotifications($surveyId, $emailsByType, $preserveResend, true);
+                    if (!$preserveResend) {
+                        $criteria->addCondition('status', FailedEmail::STATE_SUCCESS);
+                        $failedEmails->deleteAll($criteria);
```
olleharstedt, line 88: $failedEmails is an array here. Should this be inside a loop? Or FailedEmails::modal()->deleteAll()?

application/controllers/FailedEmailController.php
```diff
@@ -0,0 +1,179 @@
...
+        $item = [App()->request->getParam('item')];
+        $items = json_decode(App()->request->getParam('sItems'));
+        $selectedItems = $items ?? $item;
+        if (!empty($selectedItems)) {
+            $criteria = new CDbCriteria();
+            $criteria->select = 'id, email_type, recipient';
+            $criteria->addCondition('surveyid', $surveyId);
+            $criteria->addInCondition('id', $selectedItems);
+            $failedEmails = FailedEmail::model()->findAll($criteria);
+            if (isset($failedEmails)) {
```
olleharstedt, line 80: This is always set. Did you mean empty or count === 0? (Psalm find)

application/controllers/FailedEmailController.php
```diff
@@ -0,0 +1,179 @@
...
+                foreach ($failedEmails as $failedEmail) {
+                    $emailsByType[$failedEmail->email_type][$failedEmail->id] = $failedEmail->recipient;
```
olleharstedt, line 82: Psalm is not a fan of using variables before they are defined. Maybe define it as empty array above, and then use empty() instead of isset below?

application/controllers/FailedEmailController.php
```diff
@@ -0,0 +1,179 @@
...
+            'failedEmailModel' => $failedEmailModel,
+            'pageSize'         => $pageSize,
+            'massiveAction'    => $massiveAction,
+        ]);
+    }
+
+    /**
+     * @throws CHttpException
+     */
+    public function actionResend()
```
olleharstedt, line 60: 61 lines long, you could probably break out some part to a separate function.

application/controllers/FailedEmailController.php
ptelu: you want an additional check for the content? there is allready one for displaying the button

olleharstedt: Yep, since a malicious user can construct a custom http request to fetch data.

Shnoulle: > you want an additional check for the content? there is allready one for displaying the button

All controller public function need permission check. Else : security issue.
With yii : all controller public function can be call by direct url

olleharstedt:
```
    public function beforeRender($view) {
```

Having curly brace on same line as function signature is not PSR-12. Only if/for/switch statements should have curly brace on same line. Legacy tradition from C, I believe.

application/controllers/FailedEmailController.php
olleharstedt: Marked as resolved, but not change...?

application/controllers/FailedEmailController.php
ptelu: dont have that locally

application/controllers/FailedEmailController.php
```diff
@@ -0,0 +1,213 @@
...
+        }
+        $preserveResend = App()->request->getParam('preserveResend') ?? false;
+        $item = [App()->request->getParam('item')];
+        $items = json_decode(App()->request->getParam('sItems'));
+        $selectedItems = $items ?? $item;
+        $emailsByType = [];
+        if (!empty($selectedItems)) {
+            $criteria = new CDbCriteria();
+            $criteria->select = 'id, email_type, recipient';
+            $criteria->addCondition('surveyid = ' . (int) $surveyId);
```
Shnoulle, line 82: `$criteria->compare('surveyid', $surveyId):`

application/controllers/admin/Authentication.php
```diff
@@ -374,16 +380,68 @@ private function redirectIfLoggedIn()
...
         $returnUrl = App()->user->getReturnUrl(array('/admin'));
         Yii::app()->getController()->redirect($returnUrl);
     }
 
+    /**
+     * Delete all entries from failed_email table which are older then 30days
+     *
+     * @return void
+     */
+    private static function cleanFailedEmailTable()
```
Shnoulle, line 403: Why not in email controller ?

application/controllers/admin/Authentication.php
```diff
@@ -374,16 +380,68 @@ private function redirectIfLoggedIn()
...
+
+        FailedEmail::model()->deleteAll($criteria);
+    }
+
+    /**
+     * Checks failed_email table for entries for this user and creates a UniqueNotification
+     *
+     * @return void
+     */
+    private static function createNewFailedEmailsNotification()
```
Shnoulle, line 424: See https://github.com/LimeSurvey/LimeSurvey/pull/2539/files#r947674831

application/controllers/FailedEmailController.php
olleharstedt: Does it matter? If you suggest a change, feel free to add motivation. :)

Shnoulle: motivation : usage of PDO mean usage of more SQL cache system.

olleharstedt: Isn't PDO always used by Yii internally?

Shnoulle: Sorry : PDO with parameters :)

Else : if `(int) $surveyId != $surveyId` it must throw an error , no ?

Shnoulle: But this one is not important, justthink is better.
Have the 2 new function in Authentication controller are really more strange …

olleharstedt: Oh yeah, prepared statements. Yes, if `$criteria->compare('surveyid', $surveyId):` is doing that, then it's better.

olleharstedt: lulz
https://github.com/yiisoft/yii/blob/master/framework/db/schema/CDbCriteria.php#L446

PR #2545, opened 2022-07-25
Fixed issue #17135: activating Public statistics option causes json to display

gabrieljenik:
The problem was that when the statistics page is loaded by PJAX the Chart.min.js file is not loaded. Then, when you go to the rendersidemenylink pages (also by PJAX), you click a code in statistics.js that depends on Chart, so the code that handles the form submission by AJAX does not get executed.


Shnoulle (approved):

gabrieljenik:
> LSYii_ClientScript::POS_BEGIN work with pajax ?

Seems so. Not sure exactly why, but seems so :)

Shnoulle:
> CClientScript::POS_BEGIN : the script is inserted at the beginning of the body section.

pjax replace whole body ?

Must check with pos_END too

> CClientScript::POS_END : the script is inserted at the end of the body section.

Only POS_HEAD and POS_LOAD and POS_READY didn't work directly then ?

gabrieljenik:
pjax takes scripts from the following elements:
image

When a script is registered by default is sent to HEAD, so it is not caught by PJAX.

Shnoulle:
But more : since it replace HTML : body content are done ther ? No ?

I mean : if pjax load
```

<div id="hideme">Hide me</div>
<script>$("#hideme").hide())</script>

```
It work.

The issue is more for scriptFile i think.

gabrieljenik:
> The issue is more for scriptFile i think.

I believe issue is related to scripts set with registerScript() and registerScriptFile()

PR #2549, opened 2022-07-27
Fixed issue #18288: Can't assign groups to surveys

Shnoulle:
Dev: always use getUserGroupList
Dev: this function do the filter, return all usergroup for superadmin


gabrieljenik (approved):
Code structure looks OK. Haven't tested

Shnoulle:
@olleharstedt : please test and revert if not fix

PR #2551, opened 2022-07-27

gabrieljenik:


Shnoulle (approved):
Code is OK

Must really test if it's OK when return error in GUI and remotecontrol

gabrieljenik:
Master implementation implements front end validations.
#2558

Let's review that before implementing the same on here.

Shnoulle (approved):

gabrieljenik:
[15195-ConParche.webm](https://user-images.githubusercontent.com/4942495/192534081-334f7646-d733-454b-9257-be401e2bcaf7.webm)

Tested OK.

Wasn't tested on remote control. Still taking into account this has already been merged on master, we can safely merge here as well. What do you think?

@Shnoulle Remove the "Discussion" tag if you feel is OK :)

Thanks

Shnoulle:
> Wasn't tested on remote control. Still taking into account this has already been merged on master, we can safely merge here as well. What do you think?

Yep

PR #2553, opened 2022-07-28
Fixed issue #17067: Subquestion Relevance in array by column does not work

gabrieljenik:
- Disable feature for Array by column

### DIFF
--- a/application/views/questionAdministration/subquestionRow.twig
+++ b/application/views/questionAdministration/subquestionRow.twig
@@ -129,18 +129,21 @@
         </div>
     </td>
 
+    {% set relevanceTooltip = supportsRelevance ? gT("Click to expand") : gT("Relevance is not currently supported for columns") %}
+
     <!-- Relevance equation -->
     {% if first and scale_id == 0 %}
         <td class="relevance-equation">
             <input
                 data-toggle="tooltip"
-                data-title="{{ gT("Click to expand") }}"
+                data-title="{{ relevanceTooltip }}"
                 type='text'
                 class='relevance form-control input'
                 id='subquestions[{{ subquestion.qid }}][{{ scale_id }}][relevance]'
                 name='subquestions[{{ subquestion.qid }}][{{ scale_id }}][relevance]'
                 value="{{ subquestion.relevance|escape('html_attr') }}"
                 onkeypress=" if(event.keyCode==13) { if (event && event.preventDefault) event.preventDefault(); document.getElementById('save-button').click(); return false;}"
+                {{ supportsRelevance ? '' : 'readonly="readonly"' }}
             />
         </td>
     {%  else %}
--- a/application/views/questionAdministration/subquestions.twig
+++ b/application/views/questionAdministration/subquestions.twig
@@ -69,6 +69,7 @@
                                 'subquestionl10n': subquestion.questionl10ns[lang],
                                 'language'  : lang,
                                 'oldCode'   : true,
+                                'supportsRelevance' : question.questionType.code != 'H',
                             } %}
 
                             {% set position = position + 1 %}


application/views/questionAdministration/subquestionRow.twig
```diff
@@ -129,18 +129,21 @@
...
+
     <!-- Relevance equation -->
     {% if first and scale_id == 0 %}
         <td class="relevance-equation">
             <input
                 data-toggle="tooltip"
-                data-title="{{ gT("Click to expand") }}"
+                data-title="{{ relevanceTooltip }}"
                 type='text'
                 class='relevance form-control input'
```
Shnoulle, line 141: Why not hide ?

application/views/questionAdministration/subquestionRow.twig
```diff
@@ -129,18 +129,21 @@
         </div>
     </td>
 
+    {% set relevanceTooltip = supportsRelevance ? gT("Click to expand") : gT("Relevance is not currently supported for columns") %}
+
     <!-- Relevance equation -->
     {% if first and scale_id == 0 %}
```
Shnoulle, line 135: `{% if supportsRelevance and first and scale_id == 0 %} ` and same for header ?
(best solution)

application/views/questionAdministration/subquestionRow.twig
```diff
@@ -129,18 +129,21 @@
...
         <td class="relevance-equation">
```
Shnoulle, line 136: If must be shown something like this : <td class="relevance-equation{% if !supportsRelevance %} hidden-td{% endif %}">
Or another class, to set of column to 0

Shnoulle (changes requested):
Find a way to hide column,
Or minimum : hide input (css only)
Capture d’écran du 2022-07-29 11-15-15

application/views/questionAdministration/subquestionRow.twig
gabrieljenik: @c-schmitz Asked for it like this (if I understood correctly)

gabrieljenik:
> Find a way to hide column,

Hiding the column without any other notice could be confusing for users.

application/views/questionAdministration/subquestionRow.twig
Shnoulle: @c-schmitz  write

> Either introduce a question attribute that **hides the relevance column** in the subquestion editor for this question type (and other question types that don't use the relevance properly)

https://bugs.limesurvey.org/view.php?id=17067#c71103

Shnoulle:
> Hiding the column without any other notice could be confusing for users.

See https://github.com/LimeSurvey/LimeSurvey/pull/2553#discussion_r933264172

Personally : i think too is more clear to hide the column directly. 

Else : hide the input and replace the header _relevance is not available for this question type_

application/views/questionAdministration/subquestionRow.twig
gabrieljenik: Hiding is a way to say, "not allow people to use it". 
Making readonly is similar and I believe clearer, easier and less confusing.
If someone was expecting to use ut and suddnely is hidden, they will be confused.

tammoterhark: Great to see this. I think it would help the user (developer of surveys) if a tooltip is added: "Conditionally hiding columns is not (yet) possible, due to technical constraints."

If no help is given, people will look into the manual, ask questions in the forum and add bug reports because this options is greyed out.

tammoterhark: And talking about relevance has been changed to talking about conditions, hasn't it?

Shnoulle: Adding a specific class in td (and in th) then …

@c-schmitz ask to hide column , and we already hide column for Y scale

Capture d’écran du 2022-07-29 15-56-57

gabrieljenik:
> > Hiding the column without any other notice could be confusing for users.
> 
> See [#2553 (comment)](https://github.com/LimeSurvey/LimeSurvey/pull/2553#discussion_r933264172)

I don't think he was meaning that technically or that he wasn't open to other solutions.

I think tooltip can be valuable.
For example: https://github.com/LimeSurvey/LimeSurvey/pull/2553#discussion_r933294317

@c-schmitz @olleharstedt any comment? Thanks!

Shnoulle:
> 
> I think tooltip can be valuable. For example: [#2553 (comment)](https://github.com/LimeSurvey/LimeSurvey/pull/2553#discussion_r933294317)
> 

I thin @tammoterhark speak to show it as readonly or disable with information :).

I think no column is more clear (like X axis).

But : **code is OK !**

olleharstedt:
> Hiding the column without any other notice could be confusing for users.

I agree with this.

Tested and merged by me. Thanks!

PR #2554, opened 2022-08-01
Fixed issue #18261: Simple statistics, 'Array (5 point choice)' dont show subquestions

gabrieljenik:


olleharstedt:
Tested and reviewed by me.

Jerome-Herbinet:
@gabrieljenik @olleharstedt is this PR supposed to solve this ? : https://bugs.limesurvey.org/view.php?id=13955

gabrieljenik:
> @gabrieljenik @olleharstedt is this PR supposed to solve this ? : https://bugs.limesurvey.org/view.php?id=13955

Would have to get into the issue and test it to assure it fixes it completely, but yes, feels it could help (probably a lot)

Why? What do you think?
Want to keep discussion on 13955 ticket?

Jerome-Herbinet:
I've not found how to create an account on the mantis tool ...

gabrieljenik:
I think from here?
https://account.limesurvey.org/sign-up

Jerome-Herbinet:
I already tried ; this is only an account to get access to SASS LimeSurvey :-(

gabrieljenik:
By looking at this should be good for mantis as well.
https://community.limesurvey.org/bug-tracker/

PR #2555, opened 2022-08-01
Fixed issue #17745: Incomplete instructions to export response data to SPSS using Python

gabrieljenik:


application/views/admin/export/spss_view.php
```diff
@@ -103,7 +103,7 @@
         <ol>
             <li><?php eT("Download the data and the syntax file.");?></li>
             <li><?php eT("Open the syntax file in SPSS in Unicode mode.");?></li>
-            <em><?php eT("The next step (editing the 'FILE=' line) is only necessary when you have selected a SPSS version without Python. If you selected the version for SPSS with the Python plugin / Essentials, just save the syntax and the data file in the same folder. The full path will be automatically detected when you run the syntax."); ?></em>
+            <em><?php eT("The next step (editing the 'FILE=' line) is only necessary when you have selected a SPSS version without Python. If you selected the version for SPSS with the Python plugin / Essentials, just save the syntax and the data file in the same folder. If you use Python 3 you need to edit the syntax file: replace the line \"begin program.\" with \"begin program PYTHON3.\". The full path will be automatically detected when you run the syntax."); ?></em>
```
olleharstedt, line 106: Any idea how `\"` will interact with the script that scans for localized strings?

gabrieljenik: No idea... shall we wait or maybe replace by encoded html?

olleharstedt: Can we find another instance in the code base? Else, yeah, the encoded char might be better.

gabrieljenik: Replaced double quotes by single quotes :)

PR #2556, opened 2022-08-02
Fixed issue #18242: Quick translation: end note not shown

gabrieljenik:
"nothing to translate" message will only appear if all elements' text are empty (not just main).

### DIFF
--- a/application/controllers/admin/Translate.php
+++ b/application/controllers/admin/Translate.php
@@ -247,7 +247,9 @@ class Translate extends SurveyCommonAction
                 $textfrom_length = strlen(trim($textfrom));
                 $textfrom2_length = $associated ? strlen(trim($textfrom2)) : 0;
 
-                $all_fields_empty = !($textfrom_length > 0) && !($textfrom2_length > 0);
+                if ($textfrom_length > 0 || $textfrom2_length > 0) {
+                    $all_fields_empty = false;
+                }
 
                 $aData = array_merge($aData, array(
                                 'textfrom' => $this->cleanup($textfrom, array()),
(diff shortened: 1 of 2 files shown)


application/controllers/admin/Translate.php
```diff
@@ -244,9 +244,12 @@ private function displayUntranslatedFields($iSurveyID, $tolang, $baselang, $tab_
                 $gid = ($amTypeOptions["gid"] == true) ? $gid = $aRowfrom['gid'] : null;
                 $qid = ($amTypeOptions["qid"] == true) ? $qid = $aRowfrom['qid'] : null;
 
-                $textform_length = strlen(trim($textfrom));
+                $textfrom_length = strlen(trim($textfrom));
+                $textfrom2_length = $associated ? strlen(trim($textfrom2)) : 0;
 
-                $all_fields_empty = !($textform_length > 0);
+                if ($textfrom_length > 0 || $textfrom2_length > 0) {
+                    $all_fields_empty = false;
```
Shnoulle, line 251: Potential PHP Notice ? 

`$all_fields_empty = ($textform_length == 0 && $textfrom2_length == 0);`

(or set as true before the test)

Shnoulle (changes requested):
Avoid PHP notice (except if var $all_fields_empty is set before)

application/controllers/admin/Translate.php
gabrieljenik: Nope, defined here
https://github.com/LimeSurvey/LimeSurvey/pull/2556/files/9a7ce0f6c43a1dc97fa932b270cc49bac9b68772#diff-14ba4a24e3b50598976f2587fdfff882d528e670c401d97dc945124f5d5da5f6R175

Shnoulle: OK 
Link is bad ?

https://github.com/LimeSurvey/LimeSurvey/blob/9a7ce0f6c43a1dc97fa932b270cc49bac9b68772/application/controllers/admin/Translate.php#L175

Shnoulle (approved):

PR #2558, opened 2022-08-02

gabrieljenik:


assets/scripts/admin/surveysettings.js
```diff
@@ -152,6 +152,38 @@ function guidGenerator() {
...
+            return validateEndDate();
+        default:
+            return true;
+    }
+}
+
+/**
+ * Validates that the expiration date is not lower than the start date
+ */
+function validateEndDate() {
```
olleharstedt, line 170: Do we need to duplicate this logic in javascript? Isn't the model validation enough?

gabrieljenik: Not on the question editor.
As this screen covers so many models, if one model throughs an error, the already inputed data may be lost.
So, we validate as much as possible on the front end to avoid that.

olleharstedt: OK. Well, can you pass in $form to this function too?

olleharstedt: Hm, maybe better would be the IDs of the datetime pickers. Now they are hardcoded, which makes the function not usable in other contexts.

Shnoulle: My opinion: we must check both ...

1. disallow it in model (**must be done**)
2. return an error in model
3. Get error in form and show it in a clean HTML
5. Unsure we don't update sent value (all we can)
6. Add the JS/HTML5 validation to disallow submit
7. get error in remote control and show it as clean HTML

gabrieljenik: > Or better still, send in the $startdateTimepicker and $expiresTimepicker.

Ready for re-review

assets/scripts/admin/surveysettings.js
```diff
@@ -158,29 +158,38 @@ function guidGenerator() {
...
             return true;
     }
 }
 
 /**
- * Validates that the expiration date is not lower than the start date
+ * Validates the "Publication & access" form
  */
-function validateEndDate() {
+ function validatePublicationForm() {
```
olleharstedt, line 170: That's great, _but_ I have one more complaint: validatePublicationForm is not needed. Instead, just using function composition:

```js
return validateEndDateHigherThanStart(
  $('#startdate_datetimepicker').data('DateTimePicker'),
  $('#expires_datetimepicker').data('DateTimePicker'),
  expirationLowerThanStartError
);
```

Now you have composable parts working together!

gabrieljenik:
Ready for re-review @olleharstedt

assets/scripts/admin/surveysettings.js
```diff
@@ -152,6 +152,40 @@ function guidGenerator() {
...
+        return true;
+    }
+    if (!endDatePicker || !endDatePicker.date()) {
+        return true;
+    }
+    const difference = endDatePicker.date().diff(startDatePicker.date());
+    if (difference >= 0) {
+        return true;
+    }
+    LS.LsGlobalNotifier.createFlash(errorMessage, 'alert-danger fade in');
```
olleharstedt, line 185: Preferably, you shouldn't mix pure logic code (validate date) with effectful code (show message), but I'll allow it for now. :) You want to unit test the logic code, but integrity test the effectful code, this is the main reason to keep it separated. Also, it creates higher composability of the code base parts.

PR #2559, opened 2022-08-02
Fixed issue #18298: Saving in question organizer always prompts

gabrieljenik:
Removing confirmation prompt as it is not working well with pjax (ignored).


olleharstedt:
Thanks!

PR #2565, opened 2022-08-09
Fixed issue #18281: Users in group are not deleted

gabrieljenik:
Part of the problems had to do with the fact that the code assumed that if you were the owner you were in the group. In LTS it is like this, the owner is always in the group. But in master no.

Also done:
- Transfer to user 1 if the owner user is deleted
- If the owner of a group no longer exists, instead of the name it shows "(Deleted user)" in the table. (shouldn't happen, but still)
- Fixed inconsistencies in the actionViewGroup(). Things that were set inside IF statements and were always used.


application/controllers/UserGroupController.php
```diff
@@ -100,98 +100,96 @@ public function actionIndex()
      */
     public function actionViewGroup($ugid, bool $header = false)
     {
-        if (!Permission::model()->hasGlobalPermission('usergroups', 'read')) {
+        if (empty(Yii::app()->session['loginID']) || !Permission::model()->hasGlobalPermission('usergroups', 'read')) {
```
Shnoulle, line 103: empty(Yii::app()->session['loginID']) then Permission::model()->hasGlobalPermission return false always.

Why add it ?

application/controllers/UserGroupController.php
```diff
@@ -100,98 +100,96 @@ public function actionIndex()
...
             Yii::app()->session['flashmessage'] = gT('Access denied!');
             $this->redirect(App()->createUrl("/admin"));
         }
 
+        $ugid = (int)$ugid;
+        if (empty($ugid)) {
+            Yii::app()->user->setFlash('error', gT('GroupId missing'));
```
Shnoulle, line 110: Since throw 404 exception after, throw a 400 here maybe ?

application/controllers/UserGroupController.php
```diff
@@ -100,98 +100,96 @@ public function actionIndex()
      */
     public function actionViewGroup($ugid, bool $header = false)
     {
-        if (!Permission::model()->hasGlobalPermission('usergroups', 'read')) {
+        if (empty(Yii::app()->session['loginID']) || !Permission::model()->hasGlobalPermission('usergroups', 'read')) {
             Yii::app()->session['flashmessage'] = gT('Access denied!');
```
Shnoulle, line 104: Since throw 404 exception after, throw a 403 here maybe ?

application/controllers/UserGroupController.php
```diff
@@ -100,98 +100,96 @@ public function actionIndex()
...
+            throw new CHttpException(404, gT("User group not found."));
+        }
+
+        // Only allow access if user is:
+        // - Superadmin
+        // - Owner of the group
+        // - Member of the group
+        if (
+            !(
+                $userGroup->owner_id == Yii::app()->user->id ||
```
Shnoulle, line 125: No usage of `usercontrolSameGroupPolicy` ? Must review according to https://bugs.limesurvey.org/view.php?id=18289 decision

If we use ownerID : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Interfaces/PermissionInterface.php#L14 


```
        /* If have global : return true */
        if (!App()->getConfig('usercontrolSameGroupPolicy') && Permission::model()->hasPermission(0, 'global', 'usergroups', $sCRUD, $iUserID)) {
            return true;
        }
        /* Finally : return specific one */
        /* To check */
    }
```

application/controllers/UserGroupController.php
```diff
@@ -100,98 +100,96 @@ public function actionIndex()
...
-                });
-                // Remove group owner because an owner is automatically member of a group
-                unset($aNewUserListData[$aUserInGroupsResult->owner_id]);
-                $aData["addableUsers"] = array('-1' => gT("Please choose...")) + $aNewUserListData;
-                $aData["useraddurl"] = "";
-            }
+        $aSearchCriteria = new CDbCriteria();
+        $aSearchCriteria->compare("ugid", $ugid);
+        if (!Permission::model()->hasGlobalPermission('superadmin', 'read')) {
+            $aSearchCriteria->compare("owner_id", Yii::app()->session['loginID']);
```
Shnoulle, line 176: Need decision about https://bugs.limesurvey.org/view.php?id=18289

And review about usage of `usercontrolSameGroupPolicy`

Shnoulle (commented):
In my opinion :

- We need to clarify the option about owner_id : use it or not 
- We need to move logic of permission in UserGroup model
- Need to check usage of usercontrolSameGroupPolicy

application/controllers/UserGroupController.php
Shnoulle: in 3.X :` if (Permission::model()->hasGlobalPermission('usergroups', 'delete')) { `

gabrieljenik: > No usage of usercontrolSameGroupPolicy 

So far, this is only used in the dropdown list for assign permission to a survey.
Adding it here seems a bit out of scope.
Maybe it should, but let's analyze it separately, as to be able to wrap up the PR

Shnoulle: Yes

gabrieljenik:
@Shnoulle Ready for a re-reviewal

Shnoulle:
Don't find issue in code.

Still think we must move to `permission in UserGroup model`

Shnoulle:
I mean : there are a lot of user/usergroup issue about permission … Time to create user->hasPermission and usergroupè>hasPermssion ?

Shnoulle:
> > I mean : there are a lot of user/usergroup issue about permission … Time to create user->hasPermission and usergroupè>hasPermssion ?
> 
> Probably that needs some work and review. But I would rather push this now and solve that on another ticket, probably on DEV

Yes, and i approve here.

PR #2571, opened 2022-08-16
17117: 'Send invitations' button does nothing after failed send invitations

gabrieljenik:


Shnoulle (approved):
seems cool and better

Shnoulle (approved):

PR #2574, opened 2022-08-17
Fixed issue #18301: Mass action drop-up menu looks like overflow hidden in question list

gabrieljenik:
Works the same as in Survey List.


Shnoulle (approved):
:+1:

PR #2575, opened 2022-08-18

gabrieljenik:
- Fix how surveysettings.js is loaded (to avoid the bug in ticket #18314)
- Uncommenting the original PR, which was reverted for solving #18314


Shnoulle:
Code seems OK, 
Not tested : alert is shown ?

gabrieljenik:
I have tested it. Will re test it once again, just in case.

Shnoulle (approved):

PR #2576, opened 2022-08-22
Fixed issue #18286: Filter by group name in Group List, does not work

gabrieljenik:


application/controllers/QuestionGroupsAdministrationController.php
```diff
@@ -367,8 +367,8 @@ public function actionListquestiongroups($surveyid)
         $baselang = $survey->language;
         $model    = new QuestionGroup('search');
 
-        if (isset($_GET['QuestionGroup'])) {
-            $model->attributes = $_GET['QuestionGroup'];
+        if (isset($_GET['QuestionGroup']['group_name'])) {
```
Shnoulle, line 370: Can we move to 
```
$QuestionGroup = App()->getRequest()->get('QuestionGroup'); 
if(!empty($QuestionGroup['group_name'])) {
```

Shnoulle (changes requested):
micro update

maybe (unneded) 

```
$QuestionGroupFilter = App()->getRequest()->get('QuestionGroup'); 
foreach($QuestionGroupFilter as $attribute->value) {
```

application/controllers/QuestionGroupsAdministrationController.php
gabrieljenik: why?

Shnoulle: https://stackoverflow.com/questions/3498207/is-using-superglobals-directly-good-or-bad-in-php#3498463
https://phpmd.org/rules/controversial.html

Since we have a framework : use it …

olleharstedt: Maybe not super critical here, but yes, in general, using the request object is much, much better for testability than accessing super-globals.

Shnoulle: See «maybe (unneded)»

Better but uneeded

olleharstedt:
Will merge for now so we're making some progress.

application/controllers/QuestionGroupsAdministrationController.php
gabrieljenik: OK. Great. Yes, code was like that. We didn't refactored

Shnoulle: Yes true :) isset came from previous code.

PR #2577, opened 2022-08-22
Fixed issue #18316: User role cannot be deleted with action button

gabrieljenik:


olleharstedt:
Why are all checks cancelled? No CI was run?

Shnoulle:
Deleting a role MUST be CRSF controlled, must be a POST action.

ALL db update must be CRSF controlled then must be POST action.

application/controllers/UserRoleController.php
```diff
@@ -250,8 +250,6 @@ public function actionRunExport($ptid)
      */
     public function actionDelete()
     {
-        $this->requirePostRequest();
```
Shnoulle, line 253: Disabling CRSF control is a security issue.

Shnoulle (changes requested):

application/controllers/UserRoleController.php
olleharstedt: Yes, agree, this is not OK.

Shnoulle: Oh yes : https://github.com/LimeSurvey/LimeSurvey/blob/62d4b30c3ec38c478551c6e35054e8a85cf38ce0/application/controllers/UserRoleController.php#L257

Right , we have a getPost here :)

resolved

olleharstedt:
Tested by me. Thanks!

PR #2579, opened 2022-08-23
New feature #18321: Add `get_available_site_settings` RPC method

edgarrmondragon:
New feature [#18321](https://bugs.limesurvey.org/view.php?id=18321): "New method: get_available_site_settings"


olleharstedt:
Thanks for contributing, and thanks for adding a new unit test. :)

PR #2580, opened 2022-08-23

Shnoulle:
Dev: set role to presentation and remove thead and th (needed)
Dev: Css part : update SurveyTheme 
Dev: Fix updateRepeatHeading


Shnoulle:
Still the EM JavaScript to update

olleharstedt:
We have a LS6 in the making, using Bootstrap 5, so this PR could maybe go there?

Shnoulle:
Maybe … some week ago it was urgent …

Shnoulle:
> We have an LS6 in the making, using Bootstrap 5, so this PR could maybe go there?

Branch to be used ?

olleharstedt:
> Branch to be used ?

We have an "epic" branch we use internally.

Shnoulle:
Capture d’écran du 2022-08-29 10-36-29

I create on Epic/BS ?

olleharstedt:
I think you can wait a little, we have to discuss too, and get feedback from tpartner regarding the changes.

Shnoulle:
Need to check again and remove the colgroup

Else have other idea … but i want to know the best solution.

Shnoulle:
Any decision on this ? Need to work on alternative solution with less HTML update or not ?

olleharstedt:
@ptelu would be the man to ask.

Shnoulle:
> Shouldnt be a problem as long as it works with the current dev branch. Will write a note to check the changes. Did you check if it potentially breaks any frontend queries for javascript?

Yes it's potentially broke any jquery workaround ,near all using array question type in fact. 

The alternate solution : less update : move rile=group on a wrapper https://www.shnoulle.net/test/arrayWCAG/tableau_role-table-2.html

It's ok on the WCAG and OK usage for accessibility, but it seems less easy for real accessibility.

Shnoulle:
@ptelu alternative solution : https://github.com/LimeSurvey/LimeSurvey/pull/2888

1. Move role=group to a div wrapper of the array with aria-labelledby question->question
2. Put the role="columnheader" on the each header (answer text)
3. Give an id to each header ( answer-{SGQA}-{code})
4. Each line stay role="radiogroup" with aria-labelledby subquestion->question (canbe reol group for other array)
5. Put aria-hidden to label used for labelling and shown on phone
6. Set aria-labelledby to each header for each input (aria-labelledby="answer-{SGQA}-{code}")


Lot of work to do it on all other array.
I add some data like oAnswer to twig views, not needed but can be interesting

Shnoulle:
Here there are still the col to remove. I don't want to fix it if it was never merged …

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2888 is (a lot) more compatible with JS workaround.

PR #2581, opened 2022-08-26
Fixed issue #18294: User was not added to group created automatically

Shnoulle:
Dev: Add the current user adter creation via GUI
Dev: add flash message for information


gabrieljenik (approved):
Code seems OK.
Still, not sure the issue is a bug. Some exchanges can be found in mantis

Shnoulle:
Yes, you found some code using owner_id … 

This can be leave like their, then it work like before (and it's light). 
But clearly : there are issue with owner_id on current 5.X (with `usercontrolSameGroupPolicy`)

PR #2583, opened 2022-08-29
Dev/lsc 312 move yii framework to composer

c-schmitz:
Just moving Yii framework to use composer


olleharstedt:
Should be in dev branch?

Shnoulle:
We don't have any framework manual fix currently ?
All are i official Yii ?

https://github.com/LimeSurvey/LimeSurvey/commits/master/framework

PR #2589, opened 2022-09-06
updated "generate tokens" button text to "access code" to keep the wo…

arjunsriva-woven:
Fixed issue NA: unsure if it needs a mantis ID as it's a minor text change,

Updated the button text for the button used to generate access codes for a close access survey. 

It used to say "Generate Tokens" which is non-obvious to a non-technical user. 
Who wouldn't realize what the issue is by looking at the warning message when sending out invitation emails, which read "....having an access code"


Shnoulle:
:+1: 

Unsure for the mantis : @c-schmitz ?

gabrieljenik:
I would say we need Mantis, at least for tracking purposes.

c-schmitz:
Micro change, but I fixed the upper case wording, which was wrong.

PR #2590, opened 2022-09-09
Fixed issue #17778: Possible log out when preview survey

Shnoulle:
Dev: remove uneeded regenerateID


Shnoulle:
I have more and more issue with bad connection or old server and preview survey. and with questionExtraSurvey and questionSpreadSheetSurvey even on big server).

Shnoulle:
@c-schmitz https://github.com/LimeSurvey/LimeSurvey/commit/47107ee0e033727a009d2e696fe204bde374b996 
why not reser whole `$_SESSION['survey_'.$surveyid]` ?

We do it for token : https://github.com/LimeSurvey/LimeSurvey/blob/c9a1a04479e2c5007d3d46393ebd6c0837d47711/application/helpers/frontend_helper.php#L2048
It's the only way to do it for token …

application/helpers/frontend_helper.php
```diff
@@ -1341,13 +1341,14 @@ function renderRenderWayForm($renderWay, array $scenarios, $sTemplateViewPath, $
...
- * Resets all session variables for this survey
+ * Resets all needed session variables for this survey,
+ * keep related current survey : s_lang, srid, token …
+ * @see killSurveySession to reset ALL session
  * @param int $surveyid
  * @return void
  */
 function resetAllSessionVariables($surveyid)
 {
-    Yii:app()->session->regenerateID(true);
```
olleharstedt, line 1350: Hm, what's the commit message for this line? Before it's removed here, I mean.

Shnoulle: https://github.com/LimeSurvey/LimeSurvey/commit/47107ee0e033727a009d2e696fe204bde374b996

 Fixed issue: [security] Possible session fixation on survey entry with token

Shnoulle: PS : see the comment here : i put the line of the commit in comment …

c-schmitz:
> why not reser whole `$_SESSION['survey_'.$surveyid]` ?

I don't think it is a good idea to remove a security fix without creating a replacement that ensures the same safety.

Shnoulle:
> I don't think it is a good idea to remove a security fix without creating a replacement that ensures the same safety.

But this don't really fix security about token here : need XSS + CSP disable or server compromission etc … and this «fix» broke preview system, not related to token … and a lot of potential session system …

If need to reset sessionId : do it before admin login !  (and after logout).

Shnoulle:
> I don't think it is a good idea to remove a security fix without creating a replacement that ensures the same safety.

Added a better session fixation security : ['sameSite' as  'Strict'](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#strict) for session and crsf.
This is a better security fix and not only for token : but for whole system (admin included)

I can report it as an issue if needed.

Shnoulle:
I close this topic : because it disable a possible loss of data.

Still unrelated to security (for this : i think we must reset after login)/

But : 

1. have 2 big survey
2. Open 1st one in a tab
3. Open second one in the tab 
4. During load of the second tab : move next on fist : you get a session time out due to this session reinit

Without this reinitialisation : you delete all response in the current page … i think we still can have it, but more hard … (less time of issue).
Maybe we need to use 2 sid var in ExpressionManager : on session (current one) and one static : if static one is different than session one : throw a session issue (the session of expression manager was updated during the same process).

PR #2591, opened 2022-09-09
Fixed issue #18337: Placeholders VALIDFROM and VALIDUNTIL at email templates do not use localized dates

gabrieljenik:


Shnoulle:
1. It's a new feature
2. Unsure we need it : because some admin user can test year for example and show something.
3. date_format(VALIDFROM) and date_format(VALIDUNTIL) can be used (if it work)

gabrieljenik:
> It's a new feature

@maziminke thoughts?

Shnoulle:
Else  : it can broke some current usage of VALIDFROM : 

- comparaison (VALIDUNTIL < EXPIRY) for example
- `{date("d/m/Y",strtotime(VALIDFROM))}` usage

We don't update EXPIRY, allow user to use `{date("d/m/Y",strtotime(EXPIRY))}` then ;: i think we MUST not update it …

gabrieljenik:
> PS : if you send it according to token language : user can not use their own format : you remove a solution

Yes, you are right.
Still, now admins can't show dates in the matching language.
If doing a survey for US and Spain, that could lead to many confusions to participants as they use opposte formats: `d/m/y` vs `m/d/y`. Taking into account that, could be considered a bug.

@maziminke I believe that either we introduce new replacement variables formatted according to the current language or survey admins use something like `{date("d/mmm/Y",strtotime(VALIDFROM))}` which could be international.

If you need a quick solution, I suggest the latter.

Shnoulle:
> If doing a survey for US and Spain

`{if(TOKEN;LANGUAGE=="en",date("d/m/Y",strtotime(VALIDFROM)),date("m/d/Y",strtotime(VALIDFROM)))}`

> introduce new replacement variables formatted according to the current language

Yes : like VALIDFROMFORMATTED, unusure really needed.

> introduce new function that format dates according to the current language

You mean add a expression manager function format_date(); ?
Must define the option 

```
@param string $date ISO format
@param booelan $withTime : default to ? false
@param language $language : default to survey or token language
```

A [good core plugin](https://manual.limesurvey.org/ExpressionManagerStart#Function_definition) to do here :+1:

gabrieljenik:
> > If doing a survey for US and Spain
> 
> `{if(TOKEN;LANGUAGE=="en",date("d/m/Y",strtotime(VALIDFROM)),date("m/d/Y",strtotime(VALIDFROM)))}`

Sure. I was providing an example on why I think this is a bug.
Also, on the other side, f I have a 20 language survey, what should I do?

Shnoulle:
> > > If doing a survey for US and Spain
> > 
> > 
> > `{if(TOKEN;LANGUAGE=="en",date("d/m/Y",strtotime(VALIDFROM)),date("m/d/Y",strtotime(VALIDFROM)))}`
> 
> Sure. I was providing an example on why I think this is a bug. Also, on the other side, f I have a 20 language survey, what should I do?

VALIDFROMFORMATTED : **but** still the issue with time or not … 

or
`
{if(TOKEN;LANGUAGE=="en",date("d/m/Y",strtotime(VALIDFROM)))}{if(TOKEN;LANGUAGE=="en",date("m/d/Y",strtotime(VALIDFROM)))}…` easy to do with Calc ;)

Shnoulle:
> > > If doing a survey for US and Spain
> > 
> > 
> > `{if(TOKEN;LANGUAGE=="en",date("d/m/Y",strtotime(VALIDFROM)),date("m/d/Y",strtotime(VALIDFROM)))}`
> 
> Sure. I was providing an example on why I think this is a bug. Also, on the other side, f I have a 20 language survey, what should I do?

Else with new expression manager function `{format_date(VALIDFROM,0,TOKEN:LANGUAGE)}`

application/core/plugins/dateFunctions/EMFunctions.php
```diff
@@ -0,0 +1,44 @@
...
+
+class EMFunctions
+{
+    /**
+     * Formats a date according to the Survey's date format for the specified language
+     * @param string $date : a date in "Y-m-d H:i:s" format. Example: VALIDFROM.
+     * @param string $language : the language used for localization. If the language is not configured in the survey, the base language will be used. Example: TOKEN:LANGUAGE
+     * @return string
+     */
+    public static function localize_date($date, $language)
```
Shnoulle, line 21: In my opinion : 

1. $language must be optional : get it for App()->getLanguage() if null
2. add an opt for time

Shnoulle (commented):
App()->getLanguage() by default and need an option for time.

application/core/plugins/dateFunctions/EMFunctions.php
```diff
@@ -0,0 +1,44 @@
...
+        $surveyId = LimeExpressionManager::getLEMsurveyId();
+        $survey = Survey::model()->findByPk($surveyId);
+
+        if (empty($survey)) {
+            return '';
+        }
+
+        // If the specified language is note one of the survey's languages, fallback to the survey's base language.
+        if (!in_array($language, $survey->getAllLanguages()) || empty($survey->languagesettings[$language])) {
+            $language = $survey->language;
```
Shnoulle, line 36: Then allow null language ;)

application/core/plugins/dateFunctions/EMFunctions.php
```diff
@@ -0,0 +1,57 @@
...
+     * @return string
+     */
+    public static function localize_date($date, $language = null)
+    {
+        if (empty($date)) {
+            return '';
+        }
+
+        // Try to get Survey ID from the EM
+        $surveyId = LimeExpressionManager::getLEMsurveyId();
```
gabrieljenik, line 28: when called from email, this variable is not set on EM. That;s why we keep on trying for other alternatives if no SID found.

Shnoulle: Maybe we can create a Dev mantis issue for this.

It set in `\Yii::app()->session['LEMsid'] ` then seems OK for me.

Shnoulle (approved):
Adding time by option can be done after.

Shnoulle:
@maziminke you have some time to test ?

maziminke:
Sorry, no time for testing currently :-(

gabrieljenik:
This is tested. When using the function on an email, the language need to be explicitly set as parameter, either hard coded or using a replacement variable.

Ex:
> You can start from {localize_date(TOKEN:VALIDFROM, LANGUAGE)} and you have until {localize_date(TOKEN:VALIDUNTIL, TOKEN:LANGUAGE)} to take the survey.

maziminke:
We should definitely add a note at the manual about the new feature and how to use it.
Can you add this? I can update the German version afterwards.

Shnoulle:
Don't update manual if it's not merged

gabrieljenik:
> We should definitely add a note at the manual about the new feature and how to use it. Can you add this? I can update the German version afterwards.

This is a Core plugin. Where should we add it on the manual?

Shnoulle:
> > We should definitely add a note at the manual about the new feature and how to use it. Can you add this? I can update the German version afterwards.
> 
> This is a Core plugin. Where should we add it on the manual?

https://manual.limesurvey.org/Category:Core_Plugins

gabrieljenik:
> We should definitely add a note at the manual about the new feature and how to use it.

https://manual.limesurvey.org/DateFunctions

olleharstedt:
New feature should go to dev branch?

Shnoulle:
Oh yes :)
Not a lot of commit on current dev, sorry :)

gabrieljenik:
> New feature should go to dev branch?

You are right. Still, as it is wrapped in a plugin and solves a current bug, I thought it could be OK to merge into master.
Let me know if I should change that

PR #2594, opened 2022-09-12
Fixed issue #18300: Panel Integration is failing to load

gabrieljenik:
Removing datatables from the panel integration as to make it consistent with other screens


gabrieljenik:
PR should go to dev

gabrieljenik:
Follow up on #2607

PR #2596, opened 2022-09-12
Fixed issue #18240: QuestioinTheme browserdetect, Show Platform information didn't work

gabrieljenik:


Shnoulle (approved):

fosteruk:
Not sure why we have an LTS PR for this. 

The browserdetect QuestioinTheme code does exist in LTS but QuestioinThemes are not (officially) supported in LTS as far as I can see.

Do we really need to apply the fix in LTS?

gabrieljenik:
Well, if the plugin was done by LS and is offered by LS. 
If not supported, the only to blame would be LS.

So I would think yes, on this case, we can fix this

fosteruk:
That makes sense. Thanks for clearing that up. 

Let me figure out how to test it and then I will finalise the merge.

PR #2598, opened 2022-09-13
refactoring table striping with bs table-striped

jack-han-115:
Dev:


Shnoulle:
Unsure table-striped work with hidden line.
(but old system didn't work too i think)

PR #2601, opened 2022-09-15
Fixed issue #18351: Extra token placeholder show as invalid in expression

Shnoulle:
Fixed issue #18351: Extra token placeholder show as invalid in expression
Dev: add completed to fixed columns
Dev: use all in expression validator since Script use all token->attributes

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -2955,10 +2955,11 @@ function getTokenFieldsAndNames($surveyid, $bOnlyAttributes = false)
 {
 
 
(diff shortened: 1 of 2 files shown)


Shnoulle:
Patricia : same fix than master (and lss file id from 3.X in issue), can you check too ?
https://bugs.limesurvey.org/view.php?id=18351

Trischi80 (commented):
Hello Denis,

yes sure this one is the same fix as on master :).
Sure, will merge it now into LS3. Thanks for your work :).

Greetings Patricia

PR #2603, opened 2022-09-15
Fixed issue [security] #18355 user can grant more permissions to any …

Shnoulle:
Fixed issue [security] #18355 user can grant more permissions to any user than he has himself
Dev: Update only current user permission
Dev: check to disallow update of himself
Dev: fix the HTML error return (broken before)

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -448,14 +448,33 @@ class UserManagementController extends LSBaseController
     public function actionSaveUserPermissions(): string
     {
         if (!Permission::model()->hasGlobalPermission('users', 'update')) {
-            return $this->renderPartial(
-                'partial/error',
-                ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
-            );
+            return Yii::app()->getController()->renderPartial('/admin/super/_renderJson', [
+                "data" => [
+                    'success' => true,// With false : nothing was updated
+                    'html'    => $this->renderPartial(
+                        'partial/error',
+                        ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true],
+                        true
+                    )
+                ]
+            ]);
         }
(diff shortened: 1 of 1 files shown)


application/controllers/UserManagementController.php
```diff
@@ -1394,40 +1413,90 @@ protected function getRandomString(): string
...
             ];
         }
         return $results;
     }
 
+    /**
+     * Get user permission (CRUD) as array
+     * @return array
+     */
+    private function getCurrentUserPermission(): array
```
olleharstedt, line 1462: Should be in service class?

Shnoulle: No, we have it here : https://github.com/LimeSurvey/LimeSurvey/blob/9510ce9580e1f85cc3b1b980e5474413789c78b4/application/controllers/UserManagementController.php#L412

I have to apply it here too : same function twice. I make the commit.

About Service class : unsure it must be the same way for Object ? I check at same time.
(but prefer to have it merged monday)

Shnoulle: I move to a single funtcion for both part.

About in Service class : yes we have it , but need a complet GUI rewirite.

We use same Gui (near) for Survey and for SurveyGroup (and hope for user, Template etc …) but here : it's a security fix not a rewrite.

Shnoulle: I Survey we use disable : https://github.com/LimeSurvey/LimeSurvey/blob/9510ce9580e1f85cc3b1b980e5474413789c78b4/application/models/services/PermissionManager.php#L77 
Not hidden

olleharstedt:
Is this the smallest possible change to fix the issue?

Shnoulle:
Then: have issue when save etc …

olleharstedt:
> Then: have issue when save etc …

Yes, but when the commit is small, it's easier to review without getting anxious about possible regressions... :neutral_face:

PR #2605, opened 2022-09-15
Fixed issue #17870: Default value with line feed broke Expression Script

Shnoulle:
Dev: json_encode since it a JS var


Shnoulle:
Oups … fix the title

gabrieljenik (approved):
Change makes sense.
The encoding which was done previously by a str_replace now is done by json_encode.
Haven't tested it

Shnoulle:
Merged in master : https://github.com/LimeSurvey/LimeSurvey/pull/2604

PR #2608, opened 2022-09-16
Fixed issue #18307: A comment string starting with a '<' is not saved using Multiple choice with comments

gabrieljenik:


application/models/SurveyDynamic.php
```diff
@@ -388,12 +388,12 @@ public function getExtendedData($colName, $sLanguage, $base64jsonFieldMap)
...
         $value     = $this->$colName;
 
-        $sFullValue = strip_tags(getExtendedAnswer(self::$sid, $oFieldMap->fieldname, $value, $sLanguage));
+        $sFullValue = getExtendedAnswer(self::$sid, $oFieldMap->fieldname, $value, $sLanguage);
         if (strlen($sFullValue) > 50) {
             $sElipsizedValue = ellipsize($sFullValue, $this->ellipsize_question_value);
             $sValue          = '<span data-toggle="tooltip" data-placement="left" title="' . quoteText($sFullValue) . '">' . $sElipsizedValue . '</span>';
         } else {
-            $sValue          = $sFullValue;
+            $sValue          = CHtml::encode($sFullValue);
```
Shnoulle, line 396: If I don't make error : `getExtendedAnswer` return non stripped value for single choice.

Then if in a single choice question : we have `<strong>Really good</strong>` : it shown `<strong>Really good</strong>` but we need  

- stripped for answer from survey (maybe wit code)
- encoded for user answer

See https://gitlab.com/SondagesPro/coreAndTools/getQuestionInformation/-/blob/master/helpers/surveyColumnsInformation.php#L1004 and https://gitlab.com/SondagesPro/coreAndTools/getQuestionInformation/-/blob/master/helpers/surveyColumnsInformation.php#L1012

gabrieljenik: strip_tags is not a good option here.
It works ok for flatening the answer.
But it doesn't do well when answer options are like  " < 1" .
(strip tags gets confused and produces the reported bug)

Shnoulle: Yes, but it's not the question : it's about the value return must be different between single answer choice and free text.

Shnoulle:
Test failed  : 
Capture d’écran du 2022-09-23 16-01-38

With [survey_archive_TagCheck.lsa](https://bugs.limesurvey.org/file_download.php?file_id=16320&type=bug)

gabrieljenik:
> Test failed 

I am not sure what has been tested as the final solution is yet not implemented.

Shnoulle:
The current commit you ask me to review.

gabrieljenik:
> The current commit you ask me to review.
> 
> This commit fix partially the issue but the solution is not valid.

Oh yes, as we have been discussing it is not finished yet.

Shnoulle:
OK :) sorry :)

gabrieljenik:
What do you think now? @Shnoulle

Shnoulle:
Replacing strip_tag by viewHelper::flatten fix the issue about 1 < 2 ?

Unsure : issue with `<strong>Really good</strong>` still there (for free text answers)

Shnoulle:
You see the idea about getExtendedAnswer ?

```
    if (isset($this_answer)) {
        if($forgrid) {
            return viewHelper::flatten($this_answer) .  " [" . $sValue . "]";
        } else {
            return $this_answer . " [" . $sValue . "]";
        }
    } else {
        return $sValue;
    }
```

And always encode (in place of strip_tags) ?

gabrieljenik:
> You see the idea about getExtendedAnswer 

I understand i although I don't share it.

I think that `getExtendedAnswer()` should be left as it is.
Then whoever calls `getExtendedAnswer()` shall treat data as it needs.
If there is a common pattern, we can make a common function for treating it.

gabrieljenik:
> Replacing strip_tag by viewHelper::flatten fix the issue about 1 < 2 ?
> 
> Unsure : issue with `\<strong\>Really good\</strong\>` still there (for free text answers)

This worked on some tests we did. Need to test it further.

Shnoulle:
Need testing :)

Shnoulle:
OK test file : 
Capture d’écran du 2022-09-30 17-32-00

This fix only the issue about single `<` or `>`, but not real tag :)

My opinion : merge this one and report another one for HTML tag in free text answers.

gabrieljenik:
> My opinion : merge this one and report another one for HTML tag in free text answers.

Just as to refresh, how it should behave for free text on your opinion?

Shnoulle:
mantis is broken but i send 3 picture with  current, 1st fix and desired behaviour (in my option)

We need toi check with export too 

If user enter in Free text, comment or other `<strong>strong</strong>` we must show it like this and not like `strong`

Else : limesurvey can not be used for testing HTML competences …

gabrieljenik:
> If user enter in Free text, comment or other <strong>strong</strong> we must show it like this and not like strong

Agree. free text should be shown as it is typed.

> We need toi check with export too

For the current PR you mean? or for the free text issue?

Shnoulle:
> For the current PR you mean? or for the free text issue?

For the free text with tag  issue to report : sure. If it's needed or not. And report the 2 issue separately (i think it don't use same function …)
But about this issue : i really like to have it by [QuestionTypes](https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/QuestionTypes) : a function in this class maybe ? Or an new class ? ColumnExportValue ? 

For the current PR : another issue, must be tested. maybe reporter can test export ?

gabrieljenik:
> But about this issue : i really like to have it by [QuestionTypes](https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/QuestionTypes) : a function in this class maybe ? Or an new class ? ColumnExportValue ?

This is a DEV task to be handled on another ticket, right?

gabrieljenik:
Follow up ticket created: https://bugs.limesurvey.org/view.php?id=18402

application/controllers/ResponsesController.php
```diff
@@ -330,7 +330,7 @@ public function actionView(int $surveyId, int $id, string $browseLang = ''): voi
                     }
                 } else {
                     $answervalue = htmlspecialchars(
-                        strip_tags(
+                        viewHelper::flatten(
```
olleharstedt, line 333: `flatten` returns an array. The heck is going on here?

olleharstedt: Ah crap, sorry, my mistake. Used jump to definition to wrong function.

PR #2610, opened 2022-09-16
Dev: fix really seacrh Question, not with language

Shnoulle:


Shnoulle:
I don't think need testing, since : you really test on https://bugs.limesurvey.org/view.php?id=17042

gabrieljenik:
> I don't think need testing, since : you really test on https://bugs.limesurvey.org/view.php?id=17042

17042 only happens in 3.x, not in master.
So, no, will not be testing on master. Sorry

gabrieljenik (approved):
Code change looks good.
Haven't tested it

Shnoulle:
>  So, no, will not be testing on master. Sorry

We need a way to broke database to check if it work ;)

PR #2611, opened 2022-09-16
Dev: Allow beforeEmail events to update core replacement

Shnoulle:
Dev: sample : EXPIRY on user language, MESSAGE from user in place of SURVEYDESCRIPTION
Dev: …

### DIFF
--- a/application/core/LimeMailer.php
+++ b/application/core/LimeMailer.php
@@ -863,7 +863,7 @@ class LimeMailer extends \PHPMailer\PHPMailer\PHPMailer
                 }
             }
         }
-        $aReplacements = array_merge($this->aReplacements, $aReplacements);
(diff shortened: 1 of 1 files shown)


Shnoulle:
No impact : no core impact, only external Plugin more easy to write.

gabrieljenik (approved):
Check. Haven't tested.
Seems a good idea

olleharstedt:
Should be in dev tho? Since it's not a bug fix?

Shnoulle:
Then : new plugin can not use it … and need to find a workaround.
This make just more complex for plugin dev., and they have to manage multiple version.

It's one of the reason why i didn't try to improve LimeSurveyAPI : time to be merged : i already publish the plugin

Here : it's really a light update, then i think its good on master.

olleharstedt:
Fine.

PR #2612, opened 2022-09-19
UQT02.5 Array (Numbers)

jack-han-115:
Dev:


application/views/survey/questions/answer/arrays/multiflexi/rows/cells/answer_td.twig
```diff
@@ -23,7 +23,8 @@
 #}
 
 <!-- answer_td -->
-<td class="answer_cell_{{ ld }} answer-item {{ answertypeclass }} {{ extraclass }}{% if error %} has-error{% endif %}">
+ {# {{ answertypeclass }} is removed as dropdown-item class from bootstrap #}
```
Shnoulle, line 26: Why dropdown-item removed ? Need a way to know it's a dropdown item for SurveyTheme, plugin, workaroud etc …

https://manual.limesurvey.org/Theme_editor#Global_classes_for_question_part

Please : readd it ! It's **needed** !
Something that shown maybe 2 or 3 month after (when user goes to update) but **really needed**

Shnoulle: OK, checked BS .

Then **NEED** a `ls-dropdown-item` (and other system like `ls-answer-item` etc …

PR #2613, opened 2022-09-19
Fixed issue #18353: Unable to upload an xml file

Shnoulle:
Fixed issue #18353: Unable to upload an xml file
Fixed issue #18347: Unable to check extension of this file type image/svg
Dev: some CS

### DIFF
--- a/application/core/utils/fileExtensions.php
+++ b/application/core/utils/fileExtensions.php
@@ -17,5 +17,7 @@ return array_merge(
         'image/heic-sequence' => 'heics',
         /* xml as text/xml : #18353 */
         'text/xml' => 'xml',
+        /* svg as image/xml : #18347 */
+        'image/svg' => 'svg',
     )
 );
--- a/application/core/utils/fileExtensions.php
+++ b/application/core/utils/fileExtensions.php
@@ -15,5 +15,7 @@ return array_merge(
         'image/heif-sequence' => 'heifs',
         'image/heic' => 'heic',
         'image/heic-sequence' => 'heics',
+        /* xml as text/xml : #18353 */
+        'text/xml' => 'xml',
     )
 );
--- a/application/core/utils/fileExtensions.php
+++ b/application/core/utils/fileExtensions.php
@@ -1,17 +1,19 @@
 <?php
+
 /**
  * File extensions by MIME types.
  *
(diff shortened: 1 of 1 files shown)


gabrieljenik (approved):
Code OK. Not tested

Shnoulle:
Test for `text/xml` : https://forums.limesurvey.org/forum/french-forum/128133-difficult%C3%A9s-avec-une-question-de-type-t%C3%A9l%C3%A9chargement

application/core/utils/fileExtensions.php
```diff
@@ -1,17 +1,23 @@
...
+    array(
+        /* iphone image : issue #15624 */
+        'image/heif' => 'heif',
+        'image/heif-sequence' => 'heifs',
+        'image/heic' => 'heic',
+        'image/heic-sequence' => 'heics',
+        /* xml as text/xml : #18353 */
+        'text/xml' => 'xml',
+        /* svg as image/xml : #18347 */
+        'image/svg' => 'svg',
```
Trischi80, line 21: Hi Denis,

this task/bug was about text/xml. Is it really needed to also 'image/svg' => 'svg' for that bugfix?
Could you reproduce the bug and did you tested it by yourself after fixing it?
My problem is that i do not have ngingx set up right now for first reprocuding the bug and afterwards test if it still exist.

Greetings Patricia

Shnoulle: 2 issue in same file :  Fixed issue #[18347](https://bugs.limesurvey.org/view.php?id=18347): Unable to check extension of this file type image/svg 


Yes : on my debian 11, the file sent by user show as `image/svg`, i'm on nginx too.

I have exactly the same issue than reporter.
Capture d’écran du 2022-09-30 12-14-19

Shnoulle: PS : it's notb related to nginx, you have the same issue with apache. It's related to mime_type / fileinfo.

If fileinfo/mimetype return image/svg : we don't set it as svg.

Trischi80: Thanks Deni. So testing is ok from my side. Will merge it :)

Shnoulle: You use apache on ? for testing ?

PR #2614, opened 2022-09-19
Uqt01.5 list dropdown

twilligls:
List dropdown to BS5


assets/packages/expressions/em_javascript.js
```diff
@@ -223,7 +223,7 @@ $(document).on("keyup change",".answer-item textarea:not([onkeyup]),.answer-item
     }
 });
 /* select/dropdown item */
-$(document).on("change",".select-item select:not([onchange]),.dropdown-item select:not([onchange])",function(event){
+$(document).on("change",".select-item select:not([onchange]),.dropdown-item select:not([onchange]), select.list-question-select:not([onchange])",function(event){
```
Shnoulle, line 226: Better to move to 
`$(document).on("change",".ls-select-item select:not([onchange]),.ls-dropdown-item select:not([onchange]))",function(event){`

Else we have issue agaian and again if we don't use same class in all question view. Must respect some rules.

olleharstedt: You mean `ls-select-item` instead of `select-item`? Are you sure `ls-dropdown-item` can replace `select.list-question-select`?

PR #2621, opened 2022-09-20
UQT02.7 Array (Yes/No/Uncertain)

jack-han-115:
Dev:


Shnoulle:
> implement table-striped , remove previous ls-odd/even

table-striped didn't work with hidden line … and we have hidden line with array filter …

PR #2624, opened 2022-09-20
Fixed issue [security] : #18355 user can grant more permissions to any user than he has himself

Shnoulle:
Fixed issue : user can delete any permission to other user.
Dev: limit commit to only own permission


olleharstedt:
Thank you!! :heart:

PR #2625, opened 2022-09-20
Fixed issue #18356: User with only user update allowed can set/remove any role to any user

gabrieljenik:
Enforces parent-child user relationship for editing users setting permissions and roles.

As so, at some point, also covers a bit of
> 18355: User with user update right can give any permission to any user (inlude himself)

UserManagerService should be continued to include more behaviour.


application/models/services/UserManager.php
```diff
@@ -0,0 +1,78 @@
...
+    ) {
+        $this->managingUser = $managingUser;
+        $this->targetUser = $targetUser;
+    }
+
+    /**
+     * Returns true if the managing user can assign permissions to the target user.
+     * @return boolean
+     */
+    public function canAssignPermissions()
```
Shnoulle, line 36: ? Not related here …

application/models/services/UserManager.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param LSWebUser $managingUser
+     * @param User|null $targetUser
+     */
+    public function __construct(
+        LSWebUser $managingUser,
+        User $targetUser = null
+    ) {
+        $this->managingUser = $managingUser;
```
Shnoulle, line 28: managingUser must be current user no  ?

Shnoulle:
I don't like to add a new Service since Permission is already here.

I think we can just start with a minimal Permission system : 2 permission, 'user' and 'permission' (maybe themepermission' and never use DB inside. 
Only global + owner + maybe Group

application/models/services/UserManager.php
gabrieljenik: Well, it is wrapping some common behaviours... That's the relationship...

application/models/services/UserManager.php
gabrieljenik: Yes, that's when initialized... 
We could automate it a bit more to have the managinsuer to be set to current user by default

gabrieljenik:
> I don't like to add a new Service since Permission is already here.

What permission service do you refer?

> I think we can just start with a minimal Permission system ...

This is just trying to organize the code and adding the missing checks.
Then it can be transformed in something else.
But as a starter i something to containe the water in an organized matter.

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Interfaces/PermissionInterface.php
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Traits/PermissionTrait.php

See https://github.com/LimeSurvey/LimeSurvey/blob/e4f3ede06d502a176e9d02d3e9d4ac55bb3b76e4/application/models/Traits/PermissionTrait.php#L41 for usage in User

Then your function are 

- canAssignPermissions : `$user->hasPermission('permission', 'update');`
- canAssignRole : `$user->hasPermission('role', 'update');`
- canEdit : `$user->hasPermission('user', 'update');`

See : https://github.com/LimeSurvey/LimeSurvey/blob/e4f3ede06d502a176e9d02d3e9d4ac55bb3b76e4/application/models/Survey.php#L2250 where a lot is done before get the final permission by DB

Shnoulle:
And when we create the Permission manage page : a lot is ready :)

gabrieljenik:
> Then your function are ...

Well the logic about child-parent relationship is not there.
What's wrong in having a logic layer for that?

> And when we create the Permission manage page : a lot is ready :)

Not sure I follow this. Sorry

Shnoulle:
> Well the logic about child-parent relationship is not there.

? see [getOwnerId](https://github.com/LimeSurvey/LimeSurvey/blob/e4f3ede06d502a176e9d02d3e9d4ac55bb3b76e4/application/models/Traits/PermissionTrait.php#L10) … done for this …
You add it in user : where is the issue here ?

We talk for permission : use Permision … more clear for future dev. 

- _Where is the permission function to update an user? _ 
- _Oh right it's in `$user->hasPermission`_

> Not sure I follow this. Sorry

We need a real Permission system for user, like we have it for Survey and SurveyGroup. Then start by using same function …

Shnoulle:
> Well the logic about child-parent relationship is not there. What's wrong in having a logic layer for that?

It's here :  https://github.com/LimeSurvey/LimeSurvey/blob/e4f3ede06d502a176e9d02d3e9d4ac55bb3b76e4/application/models/Permission.php#L576 exactly.

gabrieljenik:
Child parent-relationship were not being checked,
Now I believe there is a central place where that is being checked.
Those checkes are consumed through different controllers.

`getOwnerId` or `getEntityOwnerId` is about permissions specifically.
This is about who can set permissions to who.
I am not sure we are talking about the same thing.
If we are, then in the future, we can merge it. 
It is not definite. Just trying to solve a problem decently in a short period of time.

Maybe @olleharstedt can shread more light.

Shnoulle:
My opinion on this commit : 2 solution 

1. Just fix the issue 18356 : use `Permission::model()->hasGlobalPermission('superadmin', 'read')` in actionSaveRole and actionAddRole (light commit)
2. Do the whole access system using Permission (rewrite)

olleharstedt:
Urgh, I'd like some diagrams to understand how the heck the current permission system is working... :dizzy_face:

application/models/services/UserManager.php
olleharstedt: I wonder if this function wouldn't be clearer if the two users were given as argument instead of dependencies in the constructor. "Returns true if $user1 can assign permissions to $user2".

gabrieljenik: Well.... That would turn things into a helper, and we are trying to move away from helpers.
It was out first choice, but trashed it after thinking you were not gonna like it :)

Also, one of the users is usually the same (the one assigning permissions or performing the actions), so why sending it always as a parameter.

Shnoulle: > I wonder if this function wouldn't be clearer if the two users were given as argument instead of dependencies in the constructor. "Returns true if $user1 can assign permissions to $user2".

Somethink like `hasPermssion` function …

gabrieljenik: I understand `hasPermission` doesn't check if the user applying the permission has access to the target user.
`$this->targetUser->parent_id == $this->managingUser->id`

gabrieljenik:
> The issue is about role, not the other task … adding a complete Class for this ?

The same issue (or similar) issue happened when updating email and passowrd for example. @olleharstedt pushed a commit to fix it. 

So we have similar logics when Editing users, applying permissions, applying roles, theme permissions, ... That's why a class to handle it centrally is usefull, as a starter. Later, it can hold pieces of logic which are not to be in the controller neither in the model.

Shnoulle:
> So we have similar logics when Editing users, applying permissions, applying roles, theme permissions, ... That's why a class to handle it centrally is usefull, as a starter. Later, it can hold pieces of logic which are not to be in the controller neither in the model.

Whole are hasPermission task … 
Currently : owner_id is tested in Permission model, but you can test in User->hasPermission andnever use Permission model.

application/models/services/UserManager.php
olleharstedt: > That would turn things into a helper, and we are trying to move away from helpers. It was out first choice, but trashed it after thinking you were not gonna like it :)
> 
> Also, one of the users is usually the same (the one assigning permissions or performing the actions), so why sending it always as a parameter.

Well. You shouldn't fit a round block into a square hole. :) Helper functions aren't wrong, when they _are_ helper functions, that is, without dependencies, (preferably) without side-effects.

It's easier to read. The function depicts a relationship between two entities. But where are the two entities? Somewhere else in the code, that I now have to find...

olleharstedt:
> Currently : owner_id is tested in Permission model, but you can test in User->hasPermission andnever use Permission model.

How would you write `$user->hasPermission(<to edit this user>)` using the permission trait?

Shnoulle:
> > Currently : owner_id is tested in Permission model, but you can test in User->hasPermission andnever use Permission model.
> 
> How would you write `$user->hasPermission(<to edit target user>)` using the permission trait?

Right, no  getPermissionData() and getMinimalPermissionRead() : empty …
(I have to update phpdoc too on Trait)

application/models/User.php
```diff
@@ -471,18 +471,12 @@ class='btn btn-default btn-sm green-border action_usercontrol_button'>
                     <span class='fa fa-pencil text-success'></span>
                 </button>";
         } else {
-            if (
-                $this->canEdit(Yii::app()->session['loginID'])
-            ) {
+            $userManager = new \LimeSurvey\Models\Services\UserManager(Yii::app()->user, $this);
```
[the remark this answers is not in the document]
gabrieljenik: Added

Shnoulle:
For me : the commit are really complex against a `if (!Permission::model()->hasGlobalPermission('users', 'update')) {` replaced by a `if (!Permission::model()->hasGlobalPermission('superadmin', 'read')) {`

gabrieljenik:
Something is off here.
This commit is not expected on this branch: fa9a309

PR #2627, opened 2022-09-21
Fixed issue: HTML tags are not allowed for admins in participant attributes

Shnoulle:
Dev: rules for register: sanitize
Dev: XSS filter for admin

### DIFF
--- a/application/models/Token.php
+++ b/application/models/Token.php
@@ -438,12 +438,12 @@ abstract class Token extends Dynamic
             $aRules[] = array(
                 $key, 'filter',
                 'filter' => array(self::class, 'sanitizeAttribute'),
-                'on'=>'register'
+                'on' => 'register'
             );
             $aRules[] = array(
                 $key,
                 'LSYii_Validators',
-                'except'=>'finalSubmit,register'
+                'except' => 'finalSubmit,register'
             );
         }
         return $aRules;


application/models/Token.php
```diff
@@ -438,7 +438,12 @@ public function rules()
             $aRules[] = array(
                 $key, 'filter',
                 'filter' => array(self::class, 'sanitizeAttribute'),
-                'except' => 'FinalSubmit'
+                'on' => 'register'
+            );
+            $aRules[] = array(
+                $key,
+                'LSYii_Validators',
+                'except' => 'finalSubmit,register'
```
c-schmitz, line 446: XSS filter needs also to be applied for admins according to XSS security setting, otherwise a malicious admin can embed XSS data into a participant attribute and XSS another admin for access.

c-schmitz (changes requested):

c-schmitz (approved):

application/models/Token.php
c-schmitz: Ignore my comment - I misread.

Shnoulle: I can force xss protection for superamin too, we have tools for this.

Shnoulle:
Remind 3.X version : https://github.com/LimeSurvey/LimeSurvey/pull/2626

Shnoulle:
Test failed on remote_control

Shnoulle:
test failed on

1. Remote control
2. CPDB

Thanks to @gabrieljenik for testing :)

Shnoulle:
Tested : no issue in remote_control or CPDB …

@gabrieljenik can you explain your issue ?

gabrieljenik:
> @gabrieljenik can you explain your issue ?

- Non superadmin
- Enter a participant in CPDB which contains tags (Ex: <strong>) in the first name
- Copy the participant to a survey
- The tag in the participant was removed.
- Same participant, different First Name in survey vs CPDB.

Shnoulle:
> The tag in the participant was removed.

Always for firtsname and lastname …

Tag allowed only on attribute_X

Shnoulle:
CPDB issue are clearly another issue …

PR #2630, opened 2022-09-22
Fixed issue #18343: Problem with interpreting double slash in URL (IIS)

Shnoulle:
Dev: use PHP to construct the url
Dev: check if non empty before adding it


Shnoulle:
I do the master after this one is reviewed

Shnoulle:
Alternative solution move to jquery with data-url and a on('click but finalyy it's same.

application/views/admin/survey/prepareEditorScript_view.php
```diff
@@ -49,7 +49,7 @@ function find_popup_editor(fieldname)
         return null;
     }
 
-    function start_popup_editor(fieldname, fieldtext, sid, gid, qid, fieldtype, action)
+    function start_popup_editor(fieldname, url)
```
gabrieljenik, line 52: param could be called editorURL ? (a bit more descriptive)

gabrieljenik (approved):
Code looks good. Nice fix.

Shnoulle:
> Code looks good. Nice fix.

I really think there are a better way …

application/views/admin/survey/prepareEditorScript_view.php
Shnoulle: Yep, and add some phpdoc

Shnoulle:
Next step : use a dialog box + data-url etc …

c-schmitz:
Only for 3.x?

Shnoulle:
Oups …

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2639

Shnoulle:
Reporter test the commit : fixed the popup editor but have another issue with Placeholder for example.
https://bugs.limesurvey.org/view.php?id=18343#c72044

OK to set it as tested ?

Shnoulle:
Done for LimeReplacementFields

application/helpers/admin/htmleditor_helper.php
```diff
@@ -231,7 +268,27 @@ function getInlineEditor($fieldtype, $fieldname, $fieldtext, $surveyID = null, $
             ,filebrowserImageUploadUrl:'{$sFakeBrowserURL}'
             ,filebrowserFlashUploadUrl:'{$sFakeBrowserURL}'";
         }
-
+        /* @var string[] parameters of the replacementfields url */
```
Shnoulle, line 271: Move to a function in helper ?

PR #2631, opened 2022-09-23
Fixed issue #18330: SurveySetting type date doesnt show datepicker

gabrieljenik:
The solution is a bit wide, so must be tested in many scenarios/screens.
There was a JS issue related to a complex ID (with brackets) which caused JS errors.

### DIFF
--- a/application/extensions/yiiwheels/widgets/datetimepicker/WhDateTimePicker.php
+++ b/application/extensions/yiiwheels/widgets/datetimepicker/WhDateTimePicker.php
@@ -60,7 +60,7 @@ class WhDateTimePicker extends CInputWidget
 	public function init()
 	{
 		$this->attachBehavior('ywplugin', array('class' => 'yiiwheels.behaviors.WhPlugin'));
-		$this->htmlOptions['id'] = TbArray::getValue('id', $this->htmlOptions, $this->getEscapedId());
+		$this->htmlOptions['id'] = TbArray::getValue('id', $this->htmlOptions, $this->getId());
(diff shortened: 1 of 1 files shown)


Shnoulle:
>  There was a JS issue related to a complex ID (with brackets) which caused JS errors.

Unable to fix it ? I k now anything is alwed in HTML5 but Yii have a clean solution

gabrieljenik:
> Unable to fix it ? I k now anything is alwed in HTML5 but Yii have a clean solution

Yes, this is the fix: A proper treatment of the ID

Shnoulle:
> > Unable to fix it ? I k now anything is alwed in HTML5 but Yii have a clean solution
> 
> Yes, this is the fix: A proper treatment of the ID

I don't understand or I don't make myself understood :+1: 
I mean no way to remove the bracket in the id ? Using CHtml::getIdByName for example ?

gabrieljenik:
> I mean no way to remove the bracket in the id ? Using CHtml::getIdByName for example ?

I didn;t want to mess into that. Seemed more complicated.

Shnoulle:
I make a quick test

Shnoulle:
See : https://github.com/LimeSurvey/LimeSurvey/pull/2632

A one line fix seems better than a JS hack.
And more : select or text have alreay good id for example see the `plugin_TerminPlaner_isActive` in this plugin example.

Shnoulle:
And more : the dfefault lable alreay use CHtml::getIdByName

PR #2632, opened 2022-09-24
Fixed issue #18330: SurveySetting type date doesnt show datepicker

Shnoulle:
Dev: fix the id with CHTml
Dev: same than other render part

### DIFF
--- a/application/extensions/SettingsWidget/SettingsWidget.php
+++ b/application/extensions/SettingsWidget/SettingsWidget.php
@@ -614,7 +614,7 @@ class SettingsWidget extends CWidget
         $value = isset($metaData['current']) ? $metaData['current'] : '';
         $html = Yii::app()->getController()->widget('yiiwheels.widgets.datetimepicker.WhDateTimePicker', array(
                 'name' => $name,
-                'id' => $name,
+                'id' => \CHtml::getIdByName($name),
                 'value' => $value,
                 'pluginOptions' => array(
                     'format' => $dateformatdetails['jsdate'] . " HH:mm",


Shnoulle:
See label : https://github.com/LimeSurvey/LimeSurvey/blob/0569c3c326f22927caee459ffe9fef55303218f7/application/extensions/SettingsWidget/SettingsWidget.php#L313

Shnoulle:
tested by reporter
https://bugs.limesurvey.org/view.php?id=18330#c71934

application/extensions/SettingsWidget/SettingsWidget.php
```diff
@@ -614,7 +614,7 @@ public function renderDate($name, array $metaData, $form = null)
         $value = isset($metaData['current']) ? $metaData['current'] : '';
         $html = Yii::app()->getController()->widget('yiiwheels.widgets.datetimepicker.WhDateTimePicker', array(
                 'name' => $name,
-                'id' => $name,
+                'id' => \CHtml::getIdByName($name),
```
gabrieljenik, line 617: I was looking at this. This is the only widget it has the id specifically set. I think that if the id attribute is removed, Yii will use getIdbYName anyway...

Shnoulle: I think it's only since a YII update, removing the id here : we get `yw32`


1. Look at HTML for text, texarea, dropddown …: they have their own id
2. Check label : use `getIdByName` (and it's needed)
3. With this good id : label action work (see peek)
4. 

Peek -NOID
Peek -GOODID

Shnoulle: PS : https://github.com/LimeSurvey/LimeSurvey/commit/45ed548e32c742f6a01bca9641353b09e7925409

Shnoulle: You're right : https://github.com/LimeSurvey/LimeSurvey/blob/e058b839f77043eb10a8d366d4bd69ae74014633/third_party/yiisoft/yii/framework/web/helpers/CHtml.php#L890 

See : https://github.com/yiisoft/yii/commit/0154c5e64872a5e4112f45dacb2f5ff645f1db61

gabrieljenik: OK, let me know if you want to keep it like this or change it. Thanks!

Shnoulle: What do you mean ? 

This fix the issue, t andhis is needed because `'yiiwheels.widgets.datetimepicker.WhDateTimePicker` use a bad id.

Shnoulle: You're right about CHtml fro Yii core, not `yiiwheels` widget :)

gabrieljenik (approved):

PR #2635, opened 2022-09-26
Fixed issue #18358: Statistics filter feature is not working properly

gabrieljenik:
For encrypted attributes, filtergin requires encrypting values for the query.
Only did for single choice and multiple choice.


application/helpers/admin/statistics_helper.php
```diff
@@ -380,7 +382,7 @@ function buildSelects($allfields, $surveyid, $language)
                 $thisquestion = Yii::app()->db->quoteColumnName($pv) . " IN (";
 
                 foreach ($_POST[$pv] as $condition) {
-                    $thisquestion .= "'$condition', ";
+                    $thisquestion .= "'" . getEncryptedCondition($responseModel, $pv, $condition) . "', ";
```
Shnoulle, line 385: usage of https://www.yiiframework.com/doc/api/1.1/CDbConnection#quoteValue-detail seems better, no ?

application/helpers/admin/statistics_helper.php
```diff
@@ -401,7 +403,8 @@ function buildSelects($allfields, $surveyid, $language)
                 foreach ($aresult as $arow) {
                     // only add condition if answer has been chosen
                     if (in_array($arow['title'], $_POST[$pv])) {
-                        $mselects[] = Yii::app()->db->quoteColumnName(substr($pv, 1, strlen($pv)) . $arow['title']) . " = 'Y'";
+                        $fieldname = substr($pv, 1, strlen($pv)) . $arow['title'];
+                        $mselects[] = Yii::app()->db->quoteColumnName($fieldname) . " = '" . getEncryptedCondition($responseModel, $fieldname, 'Y') . "'";
```
Shnoulle, line 407: usage of https://www.yiiframework.com/doc/api/1.1/CDbConnection#quoteValue-detail seems better, no ?

Shnoulle (changes requested):
usage of https://www.yiiframework.com/doc/api/1.1/CDbConnection#quoteValue-detail seems better, no ?

Why only single choice since issue is same for contains etc …(where usage of quoteValue is needed too)

Shnoulle:
> > Why only single choice since issue is same for contains etc …(where usage of quoteValue is needed too)
> 
> We didn't fix the quoteValue issue (want minimal impact), just handled the encryption for the single choice.

And it's not possible for contains or upper/lower thah : i'm silly …

Just the usage of quoteValue https://www.yiiframework.com/doc/api/1.1/CDbConnection#quoteValue-detail ?

application/helpers/admin/statistics_helper.php
c-schmitz: yeah, quoteValue. I agree.

Shnoulle:
> We didn't fix the quoteValue issue (want minimal impact), just handled the encryption for the single choice.

There are a difference between answer value : a lot of restriction are done : can contain only [A-Za-z0-9]

What for encrypted value ? Can not contain strange carater ?
(and quoteValue seems really a good option : always)

Shnoulle:
PS : merge with last master to have the CI Pipeline green too.

Shnoulle (approved):

application/helpers/admin/statistics_helper.php
```diff
@@ -402,7 +404,9 @@ function buildSelects($allfields, $surveyid, $language)
                 foreach ($aresult as $arow) {
                     // only add condition if answer has been chosen
                     if (in_array($arow['title'], $_POST[$pv])) {
-                        $mselects[] = Yii::app()->db->quoteColumnName(substr($pv, 1, strlen($pv)) . $arow['title']) . " = 'Y'";
+                        $fieldname = substr($pv, 1, strlen($pv)) . $arow['title'];
+                        $condition =
```
c-schmitz, line 408: What is going on here?

c-schmitz (changes requested):

gabrieljenik:
Updated.

PR #2639, opened 2022-09-29
Fixed issue #18343: Problem with interpreting double slash in URL (IIS)

Shnoulle:
Dev: use PHP to construct the url
Dev: check if non empty before adding it


gabrieljenik (approved):
Code looks OK. Not tested

Shnoulle:
Same than 3LTS except the contdir (unsure it's used)

Shnoulle:
Reporter test the commit on 3LTS : fixed the popup editor but have another issue with Placeholder for example.
https://bugs.limesurvey.org/view.php?id=18343#c72044

OK to set it as tested ?

Shnoulle:
Need to check and fix https://bugs.limesurvey.org/view.php?id=18811

Shnoulle:
I redo a simple fix … 
See https://bugs.limesurvey.org/view.php?id=18811

PR #2431, opened 2022-05-20
Fixed issue #17950: is_int(false) returns true

Shnoulle:
Dev: false and '' disallowed

PR #2642, opened 2022-09-30
Fixed issue #17950: is_int(false) returns true

Shnoulle:
Dev: false and '' fixed
Dev: test on exprmgr_int PHP function


Shnoulle:
@olleharstedt i can not update https://bugs.limesurvey.org/view.php?id=17950 … seems there are a protection system ?

olleharstedt:
Neat, thanks!

Shnoulle:
We can merge in 3LTS without unit test ? 
https://github.com/LimeSurvey/LimeSurvey/pull/2431

PR #2644, opened 2022-10-04
Fixed issue #18323: When copying a question with sub-questions, the sub-questions are not copied

gabrieljenik:
Setting attributes using setAttributes($safeOnly=false), as to not trigger the process of creating the validators.

### DIFF
--- a/application/models/services/CopyQuestion.php
+++ b/application/models/services/CopyQuestion.php
@@ -93,10 +93,10 @@ class CopyQuestion
     public function createNewCopiedQuestion($questionCode, $groupId, $questionToCopy)
     {
         $this->newQuestion = new \Question();
-        // We need to use setAttributes here with $safeOnly=false to avoid issue #18323, because otherwise the
-        // validators are loaded before setting the attributes, and that means the wrong rules are applied.
-        // By setting $safeOnly=false here, we let the proper validators to be determined while saving, when the
-        // attributes already have their values.
+        // We need to use setAttributes here with $safeOnly=false to avoid issue #18323.
+        // Otherwise validators are loaded before setting the attributes. 
+        // Right now, probably a bug, some rules are added as validators conditionally, depending on the attribute values.
+        // Then the rules set (final validators loaded) may not match the attributes which are later set.
         $this->newQuestion->setAttributes($questionToCopy->attributes, false);
         $this->newQuestion->title = $questionCode;
         $this->newQuestion->gid = $groupId;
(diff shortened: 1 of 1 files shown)


application/models/services/CopyQuestion.php
```diff
@@ -93,7 +93,11 @@ public function copyQuestion($copyOptions)
     public function createNewCopiedQuestion($questionCode, $groupId, $questionToCopy)
     {
         $this->newQuestion = new \Question();
-        $this->newQuestion->attributes = $questionToCopy->attributes;
+        // We need to use setAttributes here with $safeOnly=false to avoid issue #18323.
+        // Otherwise validators are loaded before setting the attributes. 
+        // Right now, probably a bug, some rules are added as validators conditionally, depending on the attribute values.
```
Shnoulle, line 98: Not a really bug since (for example) a subquestion can not have title `other` if` other == "Y"` for parent question.

Since we copy here : if you disable validation : you can set same code than another question (for example).

You have to disable validation for question ? Not only subquestion ?

Shnoulle: Oups … we save with validation after :)

Shnoulle: Maybe need to check isNew ?

Shnoulle (approved):
OK

Shnoulle:
I like t know if we can fix rules ;) but this fix is OK

gabrieljenik:
I have created 3 other tickets caught while reviewing this.

https://bugs.limesurvey.org/view.php?id=18394

https://bugs.limesurvey.org/view.php?id=18396

https://bugs.limesurvey.org/view.php?id=18397

Shnoulle (changes requested):
```

FILE: ...meSurvey/LimeSurvey/application/models/services/CopyQuestion.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
  97 | ERROR | [x] Whitespace found at end of line
 159 | ERROR | [x] Whitespace found at end of line
```

;)

Shnoulle (approved):

Shnoulle:
Tested with Array and Array/text : OK

gabrieljenik:
Should we test it with multiple short text as well?

Shnoulle:
> Should we test it with multiple short text as well? I will

The question : a reviewer can test ? I think it's better if tester and reviewer was different. Some test idea can be dofferent. But here : i don't think we have to test each question type …

PR #2650, opened 2022-10-07
Fixed issue {security] #09056: Plugin settings password shown in HTML source

Shnoulle:
Dev: Add Update password checkbox
Dev: do not send to save action is checkbx is unchecked
Dev: by optin (secure) to don't broke any plugin workaround

### DIFF
--- a/application/controllers/admin/Database.php
+++ b/application/controllers/admin/Database.php
@@ -656,6 +656,13 @@ class Database extends SurveyCommonAction
             $aOldAttributes = $oSurvey->attributes;
             // Save plugin settings : actually leave it before saving core : we are sure core settings is saved in LS way.
             $pluginSettings = App()->request->getPost('plugin', array());
+            /* No need to unset since js is activated, do it by security in case user deactivate JS before save settings (or have a JS error)*/
+            $aUpdate = (array) App()->request->getPost('update', null);
+            foreach($aUpdate as $name => $value) {
+                if(empty($value)) {
+                    unset($pluginSettings[$name]);
+                }
+            }
             foreach ($pluginSettings as $plugin => $settings) {
                 $settingsEvent = new PluginEvent('newSurveySettings');
                 $settingsEvent->set('settings', $settings);
--- a/application/controllers/admin/Database.php
+++ b/application/controllers/admin/Database.php
@@ -658,8 +658,8 @@ class Database extends SurveyCommonAction
             $pluginSettings = App()->request->getPost('plugin', array());
             /* No need to unset since js is activated, do it by security in case user deactivate JS before save settings (or have a JS error)*/
             $aUpdate = (array) App()->request->getPost('update', null);
-            foreach($aUpdate as $name => $value) {
-                if(empty($value)) {
+            foreach ($aUpdate as $name => $value) {
+                if (empty($value)) {
                     unset($pluginSettings[$name]);
                 }
             }
--- a/application/controllers/admin/PluginManagerController.php
+++ b/application/controllers/admin/PluginManagerController.php
@@ -330,7 +330,7 @@ class PluginManagerController extends SurveyCommonAction
             $aSave     = array();
             $aUpdate = (array) App()->request->getPost('update', null);
             foreach (array_keys($aSettings) as $name) {
-                if(!isset($aUpdate[$name]) || $aUpdate[$name]) {
+                if (!isset($aUpdate[$name]) || $aUpdate[$name]) {
                     $aSave[$name] = App()->request->getPost($name, null);
                 } else {
                     // Do not send to update array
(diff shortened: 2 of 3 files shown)


Shnoulle:
@gabrieljenik see the idea. Code can be improved (move to a new widget ?)

Else i really hate this false checkbox … issue with bacground color + issue with indeterminate state (used in Permission)
Capture d’écran du 2022-10-07 19-34-17

Shnoulle:
CI pipeline issue are not related …

`ERROR: InvalidArrayOffset - application/libraries/ExtensionInstaller/FileFetcherUploadZip.php:212:23 - Cannot access value on variable $_FILES using offset value of 'the_file', expecting 'name', 'type', 'size', 'tmp_name' or 'error' (see https://psalm.dev/115)`

Shnoulle:
Have a better idea for the system to check if need update (with API compatibility)

gabrieljenik:
Hi Denis,

I get the general idea.
I thought you were going to use the "some password" approach for identifying if a user has updated a password or not (same approach used now on other places). Althgough I prefer that approach, I am not to say this approach is not OK :) .

https://github.com/LimeSurvey/LimeSurvey/pull/2650/files#diff-6bbdf0ee6329bf62fdeb4326696b98f1a3b4879e075f9a9004a37e0d533ec34cR553

Why do we need to keep on using the old approach?
Is there any break in backwards compatibility I am missing?

Shnoulle:
> Why do we need to keep on using the old approach?
> Is there any break in backwards compatibility I am missing?

1. It's unclear for the user if you send the password or not, have a checkbox is more clear.
2. Because sending any value : you disable this value for password …

Shnoulle:
Alternative way : https://github.com/LimeSurvey/LimeSurvey/pull/2651

Maybe the alternative way (checkbox checked by default) but by option option is the best ?

Shnoulle:
PS : an idea can be to move to a clean widget and use this clean widget in other password setting :)
Disable password : no "Do you want to save password"

gabrieljenik:
> > Why do we need to keep on using the old approach?
> > Is there any break in backwards compatibility I am missing?
> 
> 1. It's unclear for the user if you send the password or not, have a checkbox is more clear.
> 2. Because sending any value : you disable this value for password …

I am sorry you provided reasons for using the "checkbox" approach vs the "some password" approach.
I was asking about the optin thing.

https://github.com/LimeSurvey/LimeSurvey/pull/2650/files#diff-6bbdf0ee6329bf62fdeb4326696b98f1a3b4879e075f9a9004a37e0d533ec34cR553
Why not just enforcing the chages to all password control types in plugin settings? 
Is there any break in backwards compatibility I am missing?

gabrieljenik:
About "checkbox aproach", I would like someone else to review it.
I don't like it, but it is  a personal opinion.

Still, how would you implement a password and confirm password settings using the checkbox approach?

Shnoulle:
>About "checkbox aproach", I would like someone else to review it.

It can be a button to hide/shown the password

> Still, how would you implement a password and confirm password settings using the checkbox approach?

It's not related, it's not a Create Password, it's use an existing password.

With an update password : you need 3 input and a specific form (see github for example)

See too LS account 
Capture d’écran du 2022-10-11 19-15-07

Shnoulle:
> I was asking about the optin thing.

Because : if you don't send password to save part : some plugin can save NULL for password. If a plugin replace saveSettings : you broke it … if a plugin always save all plugin settings in newSurveySettings: you broke the API

Then you need to get the previous password to send it : see https://github.com/LimeSurvey/LimeSurvey/pull/2651

This one send the previous password in saveSettings and newSurveySetting but it broke usage SettingsWidget out of core system

last time we revert a fix because of javascript workaround … here : did usage of SettingsWidget by plugin is part of API or not ?

Shnoulle:
> About "checkbox aproach", I would like someone else to review it.

Same with https://github.com/LimeSurvey/LimeSurvey/pull/2651#issuecomment-1272899902  ?

Else : without disabled : each time you save plugin : Firefox and Chrome show the save password …
Like it does on Global settings … each time you update Global settings…
With a disabled input : no alert

Capture d’écran du 2022-10-11 19-26-14

Shnoulle:
Global settings save
Capture d’écran du 2022-10-11 19-28-16

Shnoulle:
You can not choose never : else it don't save your user account …

gabrieljenik:
> last time we revert a fix because of javascript workaround

Thanks for that! :)
What am I supposed to say about it?

Shnoulle:
> > last time we revert a fix because of javascript workaround
> 
> Thanks for that! :) What am I supposed to say about it?

Nothing :) . It's just because we don't have a clear politic about API version ;)

Shnoulle:
The other one is more clear

PR #2651, opened 2022-10-10
Bug/09056 password plugin potential broke

Shnoulle:
Fixed issue {security] #09056: Plugin settings password shown in HTML source
Dev: Add Keep password checkbox
Dev: sedn an array with keepit => true
Dev: by optout to allow plugin show the real passwprd.

### DIFF
--- a/application/controllers/admin/Database.php
+++ b/application/controllers/admin/Database.php
@@ -656,6 +656,13 @@ class Database extends SurveyCommonAction
             $aOldAttributes = $oSurvey->attributes;
             // Save plugin settings : actually leave it before saving core : we are sure core settings is saved in LS way.
             $pluginSettings = App()->request->getPost('plugin', array());
+            /* No need to unset since js is activated, do it by security in case user deactivate JS before save settings (or have a JS error)*/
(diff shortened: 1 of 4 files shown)


Shnoulle:
Capture d’écran du 2022-10-10 09-36-36

By optout : can broke usage of SettingsWidget out of core system (in sidebody or directRequest for example) 

See here : https://gitlab.com/SondagesPro/TokenManagement/TokenUsersListAndManage/-/blob/master/TokenUsersListAndManage.php#L816 for example.

Shnoulle:
Move to optin 

```
        'checkPasswordNew'=>array(
            'type'=>'password',
            'label' => 'A new password',
            'default' => '',
            'usekeepit' => true,
        ),
```

Then hard to broke other plugin 
(except if a plugin use ['keepit' => true] for some settings …)

gabrieljenik:
Is this PR and #2650 overlaping each other or complementing?

Shnoulle:
Replaced by https://github.com/LimeSurvey/LimeSurvey/pull/2743

Same code (just update to last master)

PR #2652, opened 2022-10-10
Sonarqube fixes

c-schmitz:


application/controllers/SurveyAdministrationController.php
```diff
@@ -1288,7 +1288,7 @@ public function actionUploadimagefile()
         $uploadValidator = new LimeSurvey\Models\Services\UploadValidator();
         $uploadValidator->renderJsonOnError('file', $debug);
 
-        $iSurveyID = Yii::app()->request->getPost('surveyid');
+        $iSurveyID = (int) Yii::app()->request->getPost('surveyid');
```
Shnoulle, line 1291: My opinion : we must send an error in such case.

Since we check Permission : https://github.com/LimeSurvey/LimeSurvey/blob/7637261efd2bb3851bf83e8d87da45d1fad24a91/application/models/Permission.php#L672 we don't need it.

PR #2653, opened 2022-10-11
underline not working when hover question

jack-han-115:
Dev:


assets/bootstrap_5/scss/customizations.scss
```diff
@@ -1,38 +1,43 @@
 html {
-    font-size: $font-size-root;
+  font-size: $font-size-root;
 }
 
 a {
-    text-decoration: none;
+  text-decoration: none;
```
Shnoulle, line 6: I don't understand such customisation : see https://github.com/twbs/bootstrap/blob/d9f8be677325d8060253dd25716764f37dcbd343/scss/_reboot.scss#L245

It's inside variable (and by default get underline)

BS5 is better for accessibility and usage than BS3, why broke it ?

assets/bootstrap_5/build/css/bootstrap_5.css
```diff
@@ -10950,4 +10951,21 @@ a {
 
 .table > :not(:first-child) {
   border-top: 0;
+<<<<<<< HEAD
```
olleharstedt, line 10954: Ah, this is not correct, I think.

olleharstedt (changes requested):

assets/bootstrap_5/scss/customizations.scss
olleharstedt: We don't have accessibility for admin theme tho.

Shnoulle: Not only a11Y , but 

1. BS5 have a variable for this : why we LS don't use variable (more easy to create new theme) why it's not `_variables.scss` here ? Or usage of variable ?
2. BS5 default is underline : https://github.com/twbs/bootstrap/blob/d9f8be677325d8060253dd25716764f37dcbd343/scss/_variables.scss#L450

olleharstedt: Hm, @ptelu it sick, but he'll be back next week. I hope. :)

assets/bootstrap_5/build/css/bootstrap_5.css
Shnoulle: Are you sure ? 🤣

jack-han-115: i will need to create new PR again for this issue.

PR #2657, opened 2022-10-11
Fixed issue #18250: admin css hide logic file part

gabrieljenik:


application/views/admin/expressions/test/survey_logic_file.php
```diff
@@ -32,4 +32,27 @@
             </div>
         </div>
     </div>
-</div>
\ No newline at end of file
+</div>
+
+<style>
```
Shnoulle, line 37: Can not be included in style file ?

application/views/admin/expressions/test/survey_logic_file.php
```diff
@@ -32,4 +32,27 @@
...
+    /**
+     * This whole style block is a workaround for https://bugs.limesurvey.org/view.php?id=18250
+     */
+    .main-content-container {
+        overflow-x: visible;
```
Shnoulle, line 42: Did we really need to `overflow-x: hidden; ` in all other part ? Can not be removed totally ?

application/views/admin/expressions/test/survey_logic_file.php
```diff
@@ -32,4 +32,27 @@
...
+    }
+    .table-responsive {
```
Shnoulle, line 44: Removing table-responsive to the logic table is not better ?

application/views/admin/expressions/test/survey_logic_file.php
```diff
@@ -32,4 +32,27 @@
...
+        overflow-x: visible;
+    }
+
+    #vue-sidebar-container {
```
Shnoulle, line 48: No way to remove it from HTML for logic-file ?

application/views/admin/expressions/test/survey_logic_file.php
```diff
@@ -32,4 +32,27 @@
...
+        display: none;
+    }
+    #pjax-content {
+        max-width: 100%!important;
+    }
+    #vue-apps-main-container > .col-11 {
```
Shnoulle, line 54: No way to set as col-12 for logic file ?

Shnoulle (commented):
See general comment.

application/views/admin/expressions/test/survey_logic_file.php
gabrieljenik: As this is kind of a workaround, we thought this could be a good place.

application/views/admin/expressions/test/survey_logic_file.php
gabrieljenik: Not sure what you mean... Still, this is a workaround for doing a minimum update and avoiding a complete redo.

application/views/admin/expressions/test/survey_logic_file.php
Shnoulle: workaround for bad css ;)

application/views/admin/expressions/test/survey_logic_file.php
Shnoulle: I mean : in general overflow : hidden is a bad solution and must be avoid ;)

application/views/admin/expressions/test/survey_logic_file.php
gabrieljenik: Not sure.

Shnoulle: I give a quick test and send the screenshot ;)

Shnoulle: Oh : like before : adding fix for CSS inline : less update.
Good idea :)

Shnoulle (approved):
Inline CSS to minimal update.

Review whole page can be don after with better solutions

:+1:

Shnoulle:
Test OK too :)

application/views/admin/expressions/test/survey_logic_file.php
olleharstedt, line 37: > BREAKING BAD NEWS for "style in body" lovers: W3C has recently [lost the HTML war](https://techxplore.com/news/2019-06-w3c-whatwg-agreement-version-html.html) against WHATWG, whose versionless HTML "Living Standard" [has now become](https://www.w3.org/2019/04/WHATWG-W3C-MOU.html) the official one, which, alas, [does not allow](https://html.spec.whatwg.org/#the-style-element) STYLE in the BODY. The short-lived happy days are over. ;) The [W3C validator](https://validator.w3.org/nu/about.html) also works by the WHATWG specs now. (Thanks @FrankConijn for the heads-up!)

https://stackoverflow.com/questions/2830296/using-style-tags-in-the-body-with-other-html

Shnoulle: We follow WHATWG now ? God news, really good news… … …

gabrieljenik: Created fix: https://github.com/LimeSurvey/LimeSurvey/pull/2761

PR #2665, opened 2022-10-13
Fixed issue #18413: Cannot authenticate with LDAP when using PHP 8.1

Dennis1993:
After upgrading to PHP 8.1 the login via LDAP isn't possible.

Error message: Cannot use object of type LDAP\Connection as array

This is a fix for it.

Fixed issue #18413 (https://bugs.limesurvey.org/view.php?id=18413)

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -480,7 +480,7 @@ class AuthLDAP extends LimeSurvey\PluginManager\AuthPluginBase
 
         // Try to connect
         $ldapconn = $this->createConnection();
-        if (!is_resource($ldapconn)) {
+        if ($ldapconn === false) {
             $this->setAuthFailure($ldapconn['errorCode'], gT($ldapconn['errorMessage']));
             return;
         }


application/core/plugins/AuthLDAP/AuthLDAP.php
```diff
@@ -480,7 +480,7 @@ public function newUserSession()
 
         // Try to connect
         $ldapconn = $this->createConnection();
```
Shnoulle, line 482: Can return array or ressource 

https://github.com/Dennis1993/LimeSurvey/blob/91006af1e59f8338726dab68cdb8da2915a6372e/application/core/plugins/AuthLDAP/AuthLDAP.php#L340
https://github.com/Dennis1993/LimeSurvey/blob/91006af1e59f8338726dab68cdb8da2915a6372e/application/core/plugins/AuthLDAP/AuthLDAP.php#L351
https://github.com/Dennis1993/LimeSurvey/blob/91006af1e59f8338726dab68cdb8da2915a6372e/application/core/plugins/AuthLDAP/AuthLDAP.php#L361

Shnoulle (commented):
I think there are issue here : $ldapconn  is an array in case of error.

You just avoid the situation here ?

Maybe : if(is_array()) ? 

I don't understand why [is_resource](https://www.php.net/manual/en/function.is-resource.php) didn't work here ? 

Can you send the whole report on mantis please ?

Dennis1993:
I found this on Google: https://php.watch/versions/8.1/LDAP-resource

Shnoulle:
> I found this on Google: https://php.watch/versions/8.1/LDAP-resource @Shnoulle

Yes, but LimeSurvey can return array before return an invalid ressource.

I currently check with invalid user in my test system to report and fix .

Shnoulle:
> I found this on Google: https://php.watch/versions/8.1/LDAP-resource @Shnoulle

See here : https://github.com/Dennis1993/LimeSurvey/blob/91006af1e59f8338726dab68cdb8da2915a6372e/application/core/plugins/AuthLDAP/AuthLDAP.php#L337 

We return an array if `ldap_connect` return false.

Shnoulle:
PS : see this simpel code

```
        if ($ldapconn === false) {
            $this->setAuthFailure($ldapconn['errorCode'], gT($ldapconn['errorMessage']));
```

if $ldapconn === false then $ldapconn['errorCode'] is not set …

andreas-bulling:
I checked out the latest version from it and the respective line looks different (problem isn't fixed for me with PHP 8.1):
`
 if (isset($ldapconn['errorCode'])) {`

andreas-bulling:
I fixed those errors and created a pull request:
https://github.com/LimeSurvey/LimeSurvey/pull/2684

Shnoulle:
> **EDIT:** This error also occurs in other places. I think this needs a much more fundamental fix. The problem is that the type of $ldapconn is not unique, it is "misused" as an array - sometimes but not always. Part of the code expects an array, other parts don't.


$ldapconn is array when error. Let me fix it again

andreas-bulling:
Fixed by #2684 and #2685

PR #2666, opened 2022-10-13
Fixed issue #18310: Inconsistent GUI: Sort alphabetically vs. random

gabrieljenik:
- Added new attribute `answer_order` to replace `alpha_sort` and 'random_sort`

Pending
1) Add translations to the twig_translation_helper for the new texts.
2) _
  2a) Update process to update the settings of existing questions. (Derive the new attribute from the current ones)
  2b) What should we do with the old attributes data? Keep it just in case?
3) Revew import process: Derive the new attribute from the current ones.


application/models/Question.php
```diff
@@ -726,54 +726,94 @@ public function getOrderedAnswers($scale_id = null)
...
+     * Returns true if the answer options should be ordered randomly.
+     * @return bool
+     */
+    private function shouldOrderAnswersRandomly()
+    {
+        // Question types supporting both Random Order and Alphabetical Order should
+        // implement the 'answer_order' attribute instead of using separate attributes.
+        $answerOrder = $this->getQuestionAttribute('answer_order');
+        if (!is_null($answerOrder)) {
+            return $answerOrder == 'random';
```
Shnoulle, line 799: Not used after

gabrieljenik: Sorry, don't understand

Shnoulle: `$answerOrder` variable is not used in function 

The function do ` return $this->getQuestionAttribute('random_order') == 1 && $this->getQuestionType()->subquestions == 0;`

gabrieljenik: > $answerOrder variable is not used in function

It is used in that snippet that you comment. Don't understand.

> The function do return $this->getQuestionAttribute('random_order') == 1 && $this->getQuestionType()->subquestions == 0;

That's later, in case the answerOrder approach doesn't go well (ex: questions which don't have the new attribute)

Shnoulle: Argl … didn't see the "return"

I muts understand the reason of this update :).

Shnoulle:
Sorry for my mistake :)

To avoid such situtaion : no way to set answer_order attribute in an upgrade step ?

```
$qidsToSetrandom = Chtml::listData("qid","qid",
    QuestionAttribue::model()->find("attribute = :attribute and value = :value"),
    [':attribute' => 'random', ':value' => 1]
);
$criteria = new CDBcriteria:
$criteria->addInCondition("qid",$qidsToSetrandom);
…
```

Or , maybe better (for imported survey) an autofixed when load
```
if($this->getQuestionAttribute('random_order') == 1) {
    if($this->getQuestionType()->subquestions == 0) {
        // Here : set the attribute answer_order to random
    }
    // Here delete the attribute random_order
}
```

I don't like to keep old attributes for years …

gabrieljenik:
- Import process patched
- Update process patched
Ready for re-review

application/views/admin/super/layout_insurvey2.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+/* @deprecated moved to layouts/layout_questioneditor */
```
Shnoulle, line 3: Why we have this new file ?

Shnoulle (approved):

application/helpers/admin/import_helper.php
```diff
@@ -2160,6 +2174,34 @@ function checkWrongQuestionAttributes($questionId)
...
+function upgradeAnswerOrderQuestionAttributes($questionId)
+{
+    $questionId = (int) $questionId;
+    $db = Yii::app()->db;
+    $db->createCommand(
+        "INSERT INTO {{question_attributes}} (qid, " . $db->quoteColumnName("attribute") . ", " . $db->quoteColumnName("value") . ")
+         SELECT qa.qid, 'answer_order' " . $db->quoteColumnName("attribute") . ", 'random' " . $db->quoteColumnName("value") . "
+         FROM {{question_attributes}} qa
+         JOIN {{questions}} q ON qa.qid = q.qid
+         WHERE " . $db->quoteColumnName("attribute") . " = 'random_order' AND " . $db->quoteColumnName("value") . " = '1' AND q.type IN ('!', 'L', 'O', 'R')
```
Shnoulle, line 2190: Must test if random_order is not set in all DB

gabrieljenik: Sorry, not following... :)

Shnoulle: In the lss imported : `random_order` can not exist. If it not exist : no issue in mariadb or mysql, surely pgsql but what with mssql :)

Shnoulle: And more : 2 request for each question more … see https://bugs.limesurvey.org/view.php?id=18479

Must update without DB reques
t
Something like https://github.com/LimeSurvey/LimeSurvey/blob/174a12431a6149272d3bb52772993d49fd610642/application/helpers/admin/import_helper.php#L435 


```
    if ($insertdata['attribute'] == 'random_order' && !empty($insertdata['value']) ) {
        $insertdata['attribute'] = 'answer_order';
        $insertdata['value'] = 'random';
    }
    if ($insertdata['attribute'] == 'alphasort' && !empty($insertdata['value']) ) {
        $insertdata['attribute'] = 'answer_order';
        $insertdata['value'] = 'alphabetical';
    }
```

Shnoulle (changes requested):
No more DB request when import please

application/helpers/admin/import_helper.php
gabrieljenik: > In the lss imported : `random_order` can not exist. If it not exist : no issue in mariadb or mysql, surely pgsql but what with mssql :)

I don't see why this would fail. See this.
https://stackoverflow.com/a/45174111/1392917

gabrieljenik: > Must update without DB request

The whole ticket is about combining 2 attributes into a new one.
Attribute questions are processed one by one and we don't know the order on which they are processed.
So, what you say would work for almost all cases. 

But if Alpha = ON and Random = ON (which I think is a possible comibination right now), the current behaviour is order=Random. We must keep the compatibility for that.

If alpha is processed after random, with the approach you sent, the order would be set to alpha, which doesn't match the current behaviour.

I would like to avoid querying, but that would involve a huge code reorganziation:
- Have atributes to be saved to be stored temp on an array 
- Review the Alpha + Random transformation on memory
- Save all at once

Not sure if it is worth that reorganization for 1 extra db query per question.

Shnoulle: > I would like to avoid querying, but that would involve a huge code reorganziation:

Maybe , but doing 2 request when upload : you surely broke import LSS.

> If alpha is processed after random, with the approach you sent, the order would be set to alpha, which doesn't match the current behaviour.

You can use a var and check it, quick sample

```
    $attribute_answer_order = null;
    foreach ($xml->question_attributes->rows->row as $row) {
    …
    …
        /* Fix answer_order */
        if ($insertdata['attribute'] == 'random_order' && !empty($insertdata['value']) ) {
            $insertdata['attribute'] = 'answer_order';
            $attribute_answer_order = $insertdata['value'] = 'random';
        }
        if ($attribute_answer_order != "random" && $insertdata['attribute'] == 'alphasort' && !empty($insertdata['value']) ) {
            $insertdata['attribute'] = 'answer_order';
            $attribute_answer_order = $insertdata['value'] = 'alphabetical';
        }
    …
    …
    }
```

gabrieljenik: > Maybe , but doing 2 request when upload : you surely broke import LSS.

I believe you are being a bit too dramatic. But let me review workarounds.

> You can use a var and check it, quick sample

Is not that simple. If not, I would have already done it.

Shnoulle: > 2 extra queries per question will broke the import process?

If www and DB in in same server : no. But on different server … possible … see : https://bugs.limesurvey.org/view.php?id=18472

> Is not as simple as you wrote.

Oh yes need an array `$attribute_answer_order = [];` `$attribute_answer_order[$insertdata['qid']] = $insertdata['value'] = 'random';`
Maybe ?

Memory is easy to add, move DB is less easy.

gabrieljenik: I will come up with something

gabrieljenik:
Workaround available now

Shnoulle (approved):

kevin-foster-uk:
I have tested the affected question types, the database update patch and also the import-helper changes. I would be happy to merge this but I see this PR is pending approval by Carsten.

This .zip contains some surveys I created to test:
[ls_answer_order_test_lss.zip](https://github.com/LimeSurvey/LimeSurvey/files/10206953/ls_answer_order_test_lss.zip)

PR #2670, opened 2022-10-14
GC03 Tooltips

jack-han-115:
Dev:


Shnoulle:
Why we need IBM Sans fonts ?

PR #2672, opened 2022-10-14
Dev: better fix #18413: PHP 8.1 und LDAP funtkionieren

Shnoulle:
Dev: createConnection return array in case of error, never false (test it

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -238,8 +238,7 @@ class AuthLDAP extends LimeSurvey\PluginManager\AuthPluginBase
         if (empty($ldapmode) || $ldapmode == 'simplebind') {
             // Use the user's account for LDAP search
             $ldapbindsearch = @ldap_bind($ldapconn, $prefix . $new_user . $suffix, $password);
-        }
-        else if (empty($binddn)) {
+        } elseif (empty($binddn)) {
             // There is no account defined to do the LDAP search,
             // let's use anonymous bind instead
             $ldapbindsearch = @ldap_bind($ldapconn);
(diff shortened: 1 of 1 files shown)


application/core/plugins/AuthLDAP/AuthLDAP.php
```diff
@@ -478,9 +477,9 @@ public function newUserSession()
         $groupsearchbase        = $this->get('groupsearchbase');
         $groupsearchfilter      = $this->get('groupsearchfilter');
 
-        // Try to connect
+        /* Get the conexion, createConnection return an error in array, never return false */
         $ldapconn = $this->createConnection();
-        if ($ldapconn === false) {
+        if (isset($ldapconn['errorCode'])) {
```
Shnoulle, line 482: We can check is_array, like you want.

gabrieljenik: > We can check is_array, like you want.

Seems a good idea.

If the connection is OK and if the connection is rejected, the objectes returned may collide a bit?
I mean they are not as different as a scalar and an array as the used to be. Now they are both non-scalar.
Will that create problems?

Shnoulle: > Seems a good idea.
>
> If the connection is OK and if the connection is rejected, the objectes returned may collide a bit?
> I mean they are not as different as a scalar and an array as the used to be. Now they are both non-scalar.
> Will that create problems?

I do not understand …

in 7.4 : createConnection() return array or ressource
in 8.1 : createConnection() return array or Class

But it return array always with 'errorCode' set, never without (currently)

gabrieljenik (approved):
Code structure looks OK
Haven't tested

Shnoulle:
@c-schmitz this is needed to fix the fix : https://github.com/LimeSurvey/LimeSurvey/commit/a26b5a8231cf6b82f25bc54ea7554fe29aa99df7 

Looking at code : you see the issue

Shnoulle:
@c-schmitz this is the real fix … the other are broken …

PR #2674, opened 2022-10-17
Fixed issue #18414: 'Add line' button label can not be translated

gabrieljenik:


Shnoulle (approved):

Shnoulle:
> No risk seen. What do you think?

Maybe we need a "No risk merge" tag ?

But yes : totally right.

c-schmitz (changes requested):
Garbriel, this is a hack, not a solution. 
I asked for the proper solution in the bug report.

gabrieljenik:
Replaced by #2678, which is a broader and more general solution

PR #2677, opened 2022-10-18
UQT04.4 Gender

jack-han-115:
Dev:

### DIFF
--- a/application/views/survey/questions/answer/gender/buttons/answer.twig
+++ b/application/views/survey/questions/answer/gender/buttons/answer.twig
@@ -13,9 +13,9 @@
 
 <!--Gender question, buttons display -->
 <!-- answer -->
-<ul class="{{ coreClass }} list-unstyled btn-group btn-group-justified" data-toggle="buttons" role="radiogroup" aria-labelledby="ls-question-text-{{basename }}">
+<div class="ls-answers answers-list gender-button btn-group btn-group-justified" role="radiogroup" aria-labelledby="ls-question-text-{{ basename }}">
     <!-- Female -->
-    <li id="javatbd{{ name }}F" class="button-item btn btn-primary {% if fChecked!='' %} active {% endif %}">
+    {# <li id="javatbd{{ name }}F" class="button-item btn btn-primary {% if fChecked!='' %} active {% endif %}">
         <input
             class="radio"
             type="radio"
@@ -27,10 +27,15 @@
         <label for="answer{{ name }}F">
             <span class="fa fa-venus ls-icon" aria-hidden="true"></span> {{ gT('Female') }}
         </label>
-    </li>
+    </li> #}
+
+    <input type="radio" class="btn-check ls-button-radio" name="{{ name }}" id="answer{{ name }}F" value="F" {{ fChecked }}>
+        <label class="btn btn-primary mb-3 answer-item ls-button-label" id="javatbd{{ name }}F" for="answer{{ name }}F">
+        <span class="fa fa-venus ls-icon" aria-hidden="true"></span> {{ gT('Female') }}
+    </label>
 
     <!-- Male -->
-    <li id="javatbd{{ name }}M" class="button-item btn btn-primary  {% if mChecked!='' %} active {% endif %} ">
+    {# <li id="javatbd{{ name }}M" class="button-item btn btn-primary  {% if mChecked!='' %} active {% endif %} ">
         <input
             class="radio"
             type="radio"
@@ -42,11 +47,16 @@
         <label for="answer{{ name }}M">
             <span class="fa fa-mars ls-icon" aria-hidden="true"></span> {{ gT('Male') }}
         </label>
-    </li>
+    </li> #}
+
+    <input type="radio" class="btn-check ls-button-radio" name="{{ name }}" id="answer{{ name }}M" value="M" {{ fChecked }}>
+        <label class="btn btn-primary mb-3 answer-item ls-button-label" id="javatbd{{ name }}M" for="answer{{ name }}M">
+        <span class="fa fa-mars ls-icon" aria-hidden="true"></span> {{ gT('Male') }}
+    </label>
 
 <!-- No answer -->
     {% if noAnswer %}
-    <li id="javatbd{{ name }}" class="button-item btn btn-primary  {% if naChecked!='' %} active {% endif %}">
+    {# <li id="javatbd{{ name }}" class="button-item btn btn-primary  {% if naChecked!='' %} active {% endif %}">
         <input
             class="radio"
             type="radio"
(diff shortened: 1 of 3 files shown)


application/views/survey/questions/answer/gender/buttons/answer.twig
```diff
@@ -13,54 +13,29 @@
 
 <!--Gender question, buttons display -->
 <!-- answer -->
-<ul class="{{ coreClass }} list-unstyled btn-group btn-group-justified" data-toggle="buttons" role="radiogroup" aria-labelledby="ls-question-text-{{basename }}">
+<div class="ls-answers answers-list gender-button btn-group btn-group-justified" role="radiogroup" aria-labelledby="ls-question-text-{{ basename }}">
```
Shnoulle, line 16: Why remove coreClass system ?

The reason to add it it's to try to have something fixed for SurveyTheme developper and workaround. 
Not perfect, but better than nothing.

application/views/survey/questions/answer/gender/buttons/answer.twig
```diff
@@ -13,54 +13,29 @@
...
-            type="radio"
-            name="{{ name }}"
-            id="answer{{ name }}"
-            value=""
-            {{ naChecked }}
-        />
-        <label for="answer{{ name }}">
-            <span class="fa fa-genderless ls-icon" aria-hidden="true"></span> {{ gT('No answer') }}
+        <input type="radio" class="btn-check ls-button-radio" name="{{ name }}" id="answer{{ name }}" value="" {{ naChecked }}>
+        <label class="btn btn-primary mb-3 answer-item ls-button-label" id="javatbd{{ name }}" for="answer{{ name }}">
```
Shnoulle, line 33: Are you sure moving javatbd system to label and NOT tow wrapper is a good idea ?

javatbd is used for hidding by Expression Manager, it's the reaoson of the wrapper …

application/views/survey/questions/answer/gender/buttons/answer.twig
olleharstedt: Hm, seems to be hard-coded in multiple places in qanda helper. Is it the `ls-answer` class you want to be constant?

application/views/survey/questions/answer/gender/buttons/answer.twig
Shnoulle: You hidding only label, not input .

gender didn't have array filter, the no issue, but with radio list : you totally broke accessibility …

I really think #javatbd must wrap input + label for accessibility and clearer javascript process.

application/views/survey/questions/answer/gender/buttons/answer.twig
Shnoulle: > Hm, seems to be hard-coded in multiple places in qanda helper. Is it the `ls-answer` class you want to be constant?

$coreClass yes, since QuestionTheme must use it …

olleharstedt: Question is where to put it. If you want it to be same everywhere, it must be defined in one place only. One suggestion is question model base class, as a property.

olleharstedt: Then put `ls-answer` in $coreClass, and everything else in Twig files?

Shnoulle: > Then put `ls-answer` in $coreClass, and everything else in Twig files?

But no : since twig file came from QuestionTheme … if it's only in twig file : we can not add class to all  dependent theme …

The only reason is to try to have something clear for Theme dev … maybe we must put an information at top of Global classes for question part `deprecated in 6.X` …
This make harder to create global solution but … who really care …

Shnoulle: > > if it's only in twig file : we can not add class to all dependent theme …
> 
> Well, is that possible now? :)

Not always, but if QuestionTheme dev use {{ coreClass }} it's `ls-answers answers-list` currently for gender button here.

But maybe we just need to put https://manual.limesurvey.org/Theme_editor#Global_classes_for_question_part as global rule and hope QuestionTheme dev respect it …

application/views/survey/questions/answer/gender/buttons/answer.twig
olleharstedt: Looks like we're keeping the old structure for radio list, so should work same as before? So if it's only this question type it should be OK?

This is the required HTML structure for Bootstrap 5, so we should stick to it when it's possible, at least.

application/views/survey/questions/answer/gender/buttons/answer.twig
olleharstedt: So maybe the important part is not that we keep the $coreClass variable, but that the manual (https://manual.limesurvey.org/Theme_editor#Global_classes_for_question_part) is correct and up to date and doesn't change more than it has to when moving to Bootstrap 5?

Shnoulle: > So maybe the important part is not that we keep the $coreClass variable, but that the manual (https://manual.limesurvey.org/Theme_editor#Global_classes_for_question_part) is correct and up to date and doesn't change more than it has to when moving to Bootstrap 5?

Yes, sure. But i think have some core part for QuetsionTheme dev make it easy.

application/views/survey/questions/answer/gender/buttons/answer.twig
Shnoulle: > Looks like we're keeping the old structure for radio list, so should work same as before? So if it's only this question type it should be OK? Especially since there's no way to hide answer options anyway.

> This is the required HTML structure for Bootstrap 5, so we should stick to it when it's possible, at least.

I think you can do with different HTML; like BS3 in fact.

If i don't make error, i update Yes/No button before we release LS 2.50 … not really complex but take time …

Shnoulle: Compare https://getbootstrap.com/docs/3.3/components/#btn-groups and https://demo.sondages.me/84134?lang=fr#

CSS class do the job.

Shnoulle: > > Yes, but have same HTML structure is better if we can do it
> 
> I mean, the default must be to do Bootstrap 5 standard, and not hack around to make it look like a previous version. Unless accessibility is affected, then yes.

OK, understandable .

application/views/survey/questions/answer/gender/buttons/answer.twig
Shnoulle: PS : i think with 6.X : we must review the JS action by class ?
https://github.com/LimeSurvey/LimeSurvey/blob/b4695ff55b91d214da64851036772467f5292614/assets/packages/expressions/em_javascript.js#L210

Maybe by data-em on input ?
Adding this class fix a lot of issue about part not updated when value is updated.

ptelu: yes that might be a better solution

Shnoulle: for javatbd : need to rework expression_manager … complex

PR #2684, opened 2022-10-22
Update AuthLDAP.php

andreas-bulling:
Fixed errors with PHP 8.1

I encountered multiple errors when running LimeSurvey with PHP 8.1. I fixed these with this pull request.

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -230,8 +230,8 @@ class AuthLDAP extends LimeSurvey\PluginManager\AuthPluginBase
         if (!is_resource($ldapconn)) {
             $oEvent->set('errorCode', self::ERROR_LDAP_CONNECTION);
             $oEvent->set('errorMessageTitle', '');
-            $oEvent->set('errorMessageBody', $ldapconn['errorMessage']);
-            return null;
+            $oEvent->set('errorMessageBody', gT(ldap_error($ldapconn)));
+	   return null;
         }
 
         // Search email address and full name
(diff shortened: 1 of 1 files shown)


application/core/plugins/AuthLDAP/AuthLDAP.php
```diff
@@ -230,8 +230,8 @@ private function _createNewUser($new_user, $password = null)
         if (!is_resource($ldapconn)) {
```
Shnoulle, line 230: The issue seems more here :)

Shnoulle: $ldacon return ressource in 7.X, Class in 8.1 and array (with error) on both.

You can just replace here by the same test `if (isset($ldapconn['errorCode'])) {`

Shnoulle (changes requested):
Same test must be done when create new user.

Shnoulle:
@andreas-bulling see https://bugs.limesurvey.org/view.php?id=18427

Shnoulle:
Can i ask you to test https://github.com/LimeSurvey/LimeSurvey/pull/2685

andreas-bulling:
I found this message on the forum, which is related:
https://forums.limesurvey.org/forum/plugins/128491-error-when-authenticating-using-authldap-auto-create-user-fails

andreas-bulling:
Got the latest version from git - back to the error.

As mentioned [here](https://github.com/LimeSurvey/LimeSurvey/pull/2665#issuecomment-1287883764)
authentication in multiple other locations needs to be fixed first before the createNewUser workflow can be tested.

```
500: Internal Server Error
Cannot use object of type LDAP\Connection as array
```

Shnoulle:
Please : https://github.com/LimeSurvey/LimeSurvey/pull/2685

andreas-bulling:
works, including user auto create. Thanks.

Shnoulle:
> works, including user auto create. Thanks.

Thank to you :) 

Can you put this information on Pull request ?

I have PHP81 on some server, but not on the LDAP testing … server …

Shnoulle:
Thank you for testing and view the issue :)

PR #2685, opened 2022-10-24
Fixed issue #18427: LDAP plugin error when try to create new user on PHP8.1

Shnoulle:
Dev: same fix than log in
Dev: check if connection is an array

### DIFF
--- a/application/core/plugins/AuthLDAP/AuthLDAP.php
+++ b/application/core/plugins/AuthLDAP/AuthLDAP.php
@@ -227,7 +227,7 @@ class AuthLDAP extends LimeSurvey\PluginManager\AuthPluginBase
 
         // Try to connect
         $ldapconn = $this->createConnection();
-        if (!is_resource($ldapconn)) {
+        if (isset($ldapconn['errorCode'])) {
             $oEvent->set('errorCode', self::ERROR_LDAP_CONNECTION);
             $oEvent->set('errorMessageTitle', '');
             $oEvent->set('errorMessageBody', $ldapconn['errorMessage']);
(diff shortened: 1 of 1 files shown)


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2684#issuecomment-1288606568

gabrieljenik:
I noticed I got assigned to this PR. What I am supposed to do about it? Sorry don't understand :)

Shnoulle:
Just a code review.

See 
- https://github.com/LimeSurvey/LimeSurvey/pull/2684#issuecomment-1288553588 : need to check it's an array (in place of isset) with debug > 1
- https://forums.limesurvey.org/forum/plugins/128491-error-when-authenticating-using-authldap-auto-create-user-fails need a fix for PHP8 for user creation 

Thank you

gabrieljenik:
Already left a few comments :) Thanks

Shnoulle:
Checked OK ? or not ?

gabrieljenik:
I hope you can answer them ....

olleharstedt:
Looks fine to me? Also, I see no comments.

Shnoulle:
> I see a problem of letting the user test - they only test in their new environment. So when a fix is done for PHP 8.1, we should probably check so that it still works for PHP 7.2.

I can for LDAP …

You're right about User test but dev test too …

olleharstedt:
> You're right about User test but dev test too …

Hopefully dev test, yes. ^^

Shnoulle:
> > You're right about User test but dev test too …
> 
> Hopefully dev test, yes. ^^

I mean : any tester test what they think … but rarely : error in config for example, deleted directory, hacked HTML etc …

PR #2689, opened 2022-10-26
Fixed issue #18422: Remove Google font usage

gabrieljenik:
- Moved Open Sans out from the uploads folder.
- Made some fonts local


application/config/fonts.php
```diff
@@ -101,6 +101,15 @@
         ),
     ),
 
+    'font-opensans' => array(
```
Shnoulle, line 104: Why not adding raleway etc … on the same way ?

themes/survey/bootswatch/css/variations/cosmos.min.css
```diff
@@ -1,4 +1,4 @@
-@import url("https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700");/*!
+@import url("../../../../../assets/fonts/source-sans-pro.css");/*!
```
Shnoulle, line 1: Yep … really don't see how to do it differently …
Don't really like it, but no idea on a cleaner way to do …

Shnoulle (approved):
No big issue in code.
Just some discussion.

Shnoulle:
Still OK for SurveyTheme boostrap variation :)

gabrieljenik:
Add web safe fonts a s fallback on the email sent when "Reset pwd with mass action"

PR #2692, opened 2022-10-31
Fixed issue #14108: No error proper error message when LDAP login fails due to case-sensitivity

Shnoulle:
Dev: fix it by get the real username found


Shnoulle:
Checked on mariadb and MS sql server : api search user without case-sensitivity

Check creation on PGSQL
Capture d’écran du 2022-10-31 16-02-37

Broke at https://github.com/LimeSurvey/LimeSurvey/blob/73ef7087b344ea55b44629ba596708e540d15838/application/controllers/UserManagementController.php#L1286

PR #2693, opened 2022-11-01
Add REST API to develop branch

olleharstedt:


application/controllers/rest/V1/OpenApi/Schema/Question/QuestionDetail.php
```diff
@@ -0,0 +1,277 @@
+<?php
+
+/**
+ * @OA\Schema(
```
olleharstedt, line 4: @fosteruk Heyo, here's the PR, answer when time allows, no stress.

Seems like nothing or few things in Swagger can be generated automatically? Or?

application/controllers/rest/V1/QuestionGroupController.php
```diff
@@ -0,0 +1,426 @@
...
+     * @param string $id Question Group Id
+     * @return void
+     */
+    public function actionIndexDelete($id)
+    {
+        $requestData = [
+            'sessionKey' => $this->getAuthToken(),
+            'groupID' => $id
+        ];
+        $commandResponse = (new QuestionGroupDelete())
```
olleharstedt, line 421: @fosteruk Do we need to hardcode the command objects? Would reflection or similar be an alternative?

application/controllers/rest/V1/SurveyController.php
```diff
@@ -0,0 +1,219 @@
...
+    public function actionIndexGet(string $id)
+    {
+        $request = Yii::app()->request;
+        $requestData = array(
+            'sessionKey' => $this->getAuthToken(),
+            'surveyID' => $id,
+            'surveySettings' => $request->getParam('surveySettings')
+        );
+        $commandRequest = new Request($requestData);
+        $commandResponse = (new SurveyPropertiesGet())
```
olleharstedt, line 164: @fosteruk  I think there's some code repetition here that can be removed by either reflection or mapping request name to command object name, like

```php
[
  ['GET', 'survey'] => 'SurveyPropertiesGet'
]
```
Or something like that. :)

application/core/LSHttpRequest.php
```diff
@@ -36,7 +36,7 @@ class LSHttpRequest extends CHttpRequest
 {
 
     private $_pathInfo;
-    
+
```
olleharstedt, line 39: Please remove all white-space-only changes, they kinda pollute the PR and makes it harder to read.

application/core/LSYii_ControllerRest.php
```diff
@@ -0,0 +1,244 @@
...
+                $httpCode = 401;
+                break;
+            case 'error_bad_request':
+                $httpCode = 400;
+                break;
+            case 'error_not_found':
+                $httpCode = 404;
+                break;
+        }
+        return $httpCode;
```
olleharstedt, line 108: 200 should not be the default, I think. Rather throw an exception?

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -46,23 +63,11 @@ public function __construct(AdminController $controller)
...
-            $session->expire = time() + (int) Yii::app()->getConfig('iSessionExpirationTime', ini_get('session.gc_maxlifetime'));
-            $session->data = $username;
-            $session->save();
-            return $sSessionKey;
-        }
-        if (is_string($loginResult)) {
-            return array('status' => $loginResult);
-        }
-        return array('status' => 'Invalid user name or password');
+        return (new SessionKeyCreate)
```
olleharstedt, line 66: Hard-coding command name here is totally fine, I think. We need to keep the old functions anyway.

Shnoulle:
Just looking quickly at this commit.

Time to allow new function by plugin ? Or i have to create [a new workaround again](https://extensions.sondages.pro/development-and-example/extend-remotecontrol-api/documentation.html) ?

Shnoulle:
When i write plugin, it can be something other files in upload/remote/

Workaround work good since 2.05

tests/unit/controllers/UserManagementTest.php
```diff
@@ -110,6 +110,7 @@ public function testUpdateAdminUserFullName() {
 
     public function testUpdateAdminUserTamperproofed()
     {
+        $_SESSION = [];
```
Shnoulle, line 113: We don't have already a test helper for this ?

olleharstedt: To clear session? Maybe?

application/controllers/rest/V1/QuestionController.php
```diff
@@ -0,0 +1,434 @@
...
+     *          response="401",
+     *          description="Unauthorized"
+     *      )
+     * )
+     *
+     * @return void
+     */
+    public function actionIndexPost()
+    {
+        $request = Yii::app()->request;
```
olleharstedt, line 109: Why can't we pass the Yii request object directly to the command?

olleharstedt: Because getParam is not compatible with getRestParams, since request can come as json blob.

tests/unit/controllers/UserManagementTest.php
Shnoulle: Seems not (quick look)

PR #2700, opened 2022-11-03
Fixed issue #18181: "no answer" still shown after turning it off

Shnoulle:
Dev: remove deprecated usage of  App()->getConfig('shownoanswer')
Dev: No GUI to set : no way to update, remve it form config-default too


gabrieljenik:
Is it necesarry to also remove the config-defaults?

Shnoulle:
Yes : 
in 3.X : we have the settings in GUI
in 5.X : no settings in GUI

If you remove usage : no need to keep it config-default : not used anymore.

Alternative solution : add the settings in GUI **AND** disable settings in Survey Settings (and group settings).

application/controllers/admin/Database.php
```diff
@@ -727,9 +727,6 @@ private function actionUpdateSurveyLocaleSettings($iSurveyID)
             //if (getGlobalSetting('showqnumcode') === 'choose'){
                 $oSurvey->showqnumcode = $this->filterEmptyFields($oSurvey, 'showqnumcode');
             //}
-            //if (getGlobalSetting('shownoanswer') == 2){  // Don't do exact comparison because the value could be from global settings table (string) or from config (integer)
```
Shnoulle, line 730: Commented reason ? Need to be removed or not ?

gabrieljenik:
> If you remove usage : no need to keep it config-default : not used anymore.

Not sure I follow. Why is not used anymore?

Ok, is not in the GUI anymore.
Still, if you keep the config-defaults, you can set the config globally through the file.

Maybe it is in the teams plan to remove it from there. 
Still, I haven't read that anywhere.
What if someone uses it?
So, removing it now, on a minor release, seems too much.

I think this is fixable keeping the config-defaults.
Later, we can remove it.

Shnoulle:
There are 3 solutions : 

1. Remove it because it must be removed at some time but someone didn't remove it 
2. Leave it and the,n fix all other part (add global settings, show settings as disable, show help for simple user etc …)
3. Leave it but without allowing to update in global settings (need show settings as disable, show help for simple user etc …)

> Maybe it is in the Olle/Carsten's plan to remove it.
> Still, I haven't read that anywhere.

The number of things updated without clear decision during each big update are very high … 


> What if someone uses it?

See the reported issue : we need 5 month to understand what happen. I have the issue for a friends of mine to understand what happen. Friend ask me to hide it, superadmin user … 3 reporters have the same issue etc …

I set as decision needed  @olleharstedt and @c-schmitz : keep the global settings or remove it ?

We have same issue with 3 other global settings https://bugs.limesurvey.org/view.php?id=18450

Shnoulle:
I think the Global survey settings replace this part.

Shnoulle:
@olleharstedt need to add the global settings (like in 3.X) 
Or Global survey settings replace it ?

olleharstedt:
Dunno, man. We need bi-weekly meeting for stuff like this.

Shnoulle:
Here : issue reported : 2022-06-07 12:24
Find a way to reproduce (by a LS dev) : 2022-11-02

3 user on same mantis with discussion and say "we have same issue" 
3 user report issue mean a lot more have issue ;)

Shnoulle:
Still people who can not deactivate "Show no answer" …

Shnoulle:
> Still people who can not deactivate "Show no answer" …

gabrieljenik (approved):
Code looks ok, not tested

Shnoulle:
Need to be in 5

Shnoulle:
And 3.X too

Shnoulle:
@olleharstedt : what is the decison here ? 

See https://bugs.limesurvey.org/view.php?id=18181 3 user minimum.

The issue is about removing this option from GUI but leave the test. Then it's a specific way for reproduce.

But clealy need a cleaner solution

I need to fx it 5, but if it's another decision : i wait.

olleharstedt:
> The issue is about removing this option from GUI but leave the test

Hm you want to remove the question attribute? GUI = question editor?

olleharstedt:
If both Gabriel and Denis thinks this solution is OK, then it's OK for me. :) Just waiting for testing to happen then?

Shnoulle:
Hm you want to remove the question attribute? GUI = question editor?


NO : Global settings : « Show 'no answer' option for non-mandatory questions: 

Capture d’écran du 2023-06-09 11-27-02


Disappear in 5.X from GUI

Shnoulle:
> If both Gabriel and Denis thinks this solution is OK, then it's OK for me. :) Just waiting for testing to happen then?

I'm unsure if we have same decision; I don't understand oif @gabrieljenik want to keep the global settings «« Show 'no answer' option for non-mandatory questions:» in 5 (and then add it again to GUI) or if it's another solution.

olleharstedt:
> Disappear in 5.X from GUI

On purpose? Or bug? Was a decision made?

Shnoulle:
> > Disappear in 5.X from GUI
> 
> On purpose? Or bug? Was a decision made?

You really ask to me ? 
https://github.com/LimeSurvey/LimeSurvey/commit/07c3a949afe8aacbd65869cb438f086c28f9e54b


Like here : https://github.com/LimeSurvey/LimeSurvey/pull/3157 …

gabrieljenik:
> If both Gabriel and Denis thinks this solution is OK, then it's OK for me. :) Just waiting for testing to happen then?

Honestly, don't recall the issue. But is not that Denis and I got aligned, but more that a decision was done by someone else [https://bugs.limesurvey.org/view.php?id=18181#c75418]. That's why I mvoed forward with the code review.

I am picking this app, so I maynot recall details.
My thought:
- If the setting dissapeared from GUI 
- but is still used by scripts
- Then 
  + reinstante the GUI is better than 
  + removing the setting at global al level and also at config.php level.
- Why?
  + Not loose feature
  + Less impact on the code and on the app

Shnoulle:
I think because the system move to [inherit] by global Survey settings (and SurveyGroup settings).

We don't have global survey settings before. 

And here : it lack a default value.

We can add it again but then need to show it at disable a all survey settings (global + SurveyGroup)

Shnoulle:
> * Less impact on the code and on the app

What ?

The issue still persist about Surey settings : we must show as disable.

**More code to do**

gabrieljenik:
> We can add it again but then need to show it at disable a all survey settings (global + SurveyGroup)

Chain goes: Config -> Global ->Survey Group -> Survey
Right?
Wouldn't the survey settings override the global?

> The issue still persist about Surey settings : we must show as disable.

OK, yes, probably something on the survey groups and survey setting shall be reviewed.
But it would be consistent with other settings, right?

Shnoulle:
> Chain goes: Config -> Global ->Survey Group -> Survey

No : Config > Global survey (direct)
Capture d’écran du 2023-06-09 15-07-42


>  Wouldn't the survey settings override the global?

Inherited value. 

We have issue reported because user CAN NOT update this part. No issue because this settings disappear to force Survey settings.

Shnoulle:
@gabrieljenik if we add it again : why not

 Show "There are X questions in this survey": 
Capture d’écran du 2023-06-09 15-12-00

 
 Show question group title and/or description: 
Capture d’écran du 2023-06-09 15-12-33

 Show question number and/or question code: 
Capture d’écran du 2023-06-09 15-12-59


Maybe because these parameters have been completely removed without leaving out a single part?

gabrieljenik:
OK, we have a mixed situation.
What's the future desired direction? @tiborpacalat ?

Set it up from global settings or global survey?

Shnoulle:
> Set it up from global settings or global survey?

It's currently in global survey, the question is not OR.

The questions are
1.  «Do we want to force user with GlobalSettings permission to force some survey settings»
2. If yes : why only this one and all other (the 3 previous one for started)

gabrieljenik:
I am sorry. I can't follow the discussion.
Seemes there is something I am missing and believe that now that Tibor and Olle are involved, I can just retire from the topic. :)

tiborpacalat:
I don't know why there is still discussion about this. I went through tickets that had "Decision needed" with Carsten and this is why the tag was removed. After that I commented that the PR is fine and should be reviewed and tested. If you want additional things to be handled, please create a new ticket.

tiborpacalat:
This PR points to master, but it is actually for 5.x.
Please update this branch for master and create another one for 5.x.

Shnoulle:
Fixed the save issue

And clearly remove comments about **global settings are off** on `shownoanswer` , `showxquestions` , `showgroupinfo` , `showqnumcode`.

The questions now are  (or @tiborpacalat and @olleharstedt ) 

1. Did i create same PR for master
2. OR did i add shownoanswer choose/forced on global settings like 3.X
3. OR i didn't add it at global settings BUT : show it as not editable on Survey settings and at SurveyGroup settings (with the reason)
5. MUST test for the 3 other settings if it work (and remove it from config-default if not.

tiborpacalat:
All settings are working shownoanswer , showxquestions , showgroupinfo , showqnumcode.

Shnoulle:
> All settings are working shownoanswer , showxquestions , showgroupinfo , showqnumcode.

Yes, but in 3.X : they are in Global settings. In 5 and 6 : they are in global survey settings.

In 3.X : superadmin can force the settings to ON , OFF or let admin choose.

tiborpacalat:
> 

My comment was only regarding testing this PR.

PR #2701, opened 2022-11-03
Don't omit PO and MO files when uploading plugins with transcriptions

BartoszMarganiec:
Fixed issue #18451: PO and MO files are no more ignored when plugin is uploaded and copied to the upload dir.
https://bugs.limesurvey.org/view.php?id=18451


Shnoulle:
Really needed ! 
Thank you !!

Shnoulle (approved):
:+1:

PR #2708, opened 2022-11-08
Fixed issue #18464: Min/max settings doesn't work for question attrib…

Shnoulle:
Fixed issue #18464: Min/max settings doesn't work for question attrib…ute with type of integer
Dev: add the HTML attribute (clearly removed from code)

### DIFF
--- a/application/extensions/AdvancedSettingWidget/views/integer.php
+++ b/application/extensions/AdvancedSettingWidget/views/integer.php
@@ -4,10 +4,6 @@
             <?= $this->setting['aFormElementOptions']['inputGroup']['prefix']; ?>
         </div>
     <?php endif; ?>
-        <!--
-        max=""
-        min=""
-        -->
     <input 
         type="number" 
         class="form-control" 
@@ -15,6 +11,8 @@
         id="<?= CHtml::getIdByName($inputBaseName); ?>"
         <?= ($this->setting['help']) ? 'aria-describedby="help-' . CHtml::getIdByName($inputBaseName) . '"' : "" ?>
         value="<?= CHtml::encode($this->setting['value']); ?>"
+        <?= isset($this->setting['min']) ? 'min="' . $this->setting['min'] . '"' : '' ?>
+        <?= isset($this->setting['max']) ? 'max="' . $this->setting['max'] . '"' : ''  ?>
     />
     <?php if (isset($this->setting['aFormElementOptions']['inputGroup']['suffix'])) : ?>
         <div class="input-group-addon">
--- a/application/models/QuestionAttribute.php
+++ b/application/models/QuestionAttribute.php
@@ -380,8 +380,8 @@ class QuestionAttribute extends LSActiveRecord
             "readonly_when_active" => false,
             "expression" => null,
             "xssfilter" => true,
-            "min" => null,
-            "max" => null,
+            "min" => null, // Used for integer type
+            "max" => null, // Used for integer type
         );
     }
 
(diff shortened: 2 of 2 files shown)


application/models/QuestionAttribute.php
```diff
@@ -380,6 +380,8 @@ public static function getDefaultSettings()
             "readonly_when_active" => false,
             "expression" => null,
             "xssfilter" => true,
+            "min" => null,
```
gabrieljenik, line 383: I am not sure about adding these here. Why?

application/extensions/AdvancedSettingWidget/views/integer.php
```diff
@@ -4,17 +4,15 @@
...
-        min=""
-        -->
     <input 
         type="number" 
         class="form-control" 
         name="<?= $inputBaseName ?>"
         id="<?= CHtml::getIdByName($inputBaseName); ?>"
         <?= ($this->setting['help']) ? 'aria-describedby="help-' . CHtml::getIdByName($inputBaseName) . '"' : "" ?>
         value="<?= CHtml::encode($this->setting['value']); ?>"
+        <?= isset($this->setting['min']) ? 'min="' . $this->setting['min'] . '"' : '' ?>
```
gabrieljenik, line 14: Checking like this is OK or should also check there is content?

gabrieljenik (approved):
Code looks OK generally speaking.
Please also recall to update manual

application/extensions/AdvancedSettingWidget/views/integer.php
Shnoulle: What do you mean ?

Shnoulle:
>  Please also recall to update manual

This work in 3.LTS , not in 5, but right : nothing in manual for min/max.
About set it to null in default value : i really think it's more clean and clear for other dev.

Shnoulle:
>  Please also recall to update manual

Must add XSS (and the complete version number since)

application/models/QuestionAttribute.php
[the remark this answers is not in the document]
gabrieljenik: Yes, that could help

application/extensions/AdvancedSettingWidget/views/integer.php
Shnoulle: > if the min setting contains a space, should the attribute be used or not?

If it's not an integer :). 
It's PHP or config.xml file : it's not user entered data. I think we don't have to fix it.

`value="<?= CHtml::encode($this->setting['value']); ?>"` don't check if it's integer here too :)

Shnoulle:
> > @olleharstedt some core attribute need it i think.
> 
> Sure, but still waiting for test?

Adding unit or functional test do the trick when i really need a fix ? Or i add to hire a tester ?

olleharstedt:
> > > @olleharstedt some core attribute need it i think.
> > 
> > 
> > Sure, but still waiting for test?
> 
> Adding unit or functional test do the trick when i really need a fix ? Or i add to hire a tester ?

Gabriel has a tester and we're trying to hire one too.

Yeah, automated tests could be an alternative, maybe. Do we have test instructions?

application/extensions/AdvancedSettingWidget/views/integer.php
olleharstedt: Use `??` here please.

olleharstedt (changes requested):

application/extensions/AdvancedSettingWidget/views/integer.php
olleharstedt: Ah wait, not possible. Sorry, my bad.

Shnoulle: In 3.X : we have a system to be sure to have something : https://github.com/LimeSurvey/LimeSurvey/blob/06e252130069f697b247446ae52c6b8b18d261d8/application/models/QuestionAttribute.php#L312

But in 5.X … no system

PR #2722, opened 2022-11-14
Fixed issue #18463: List of uploaded files should be sorted alphabetically

gabrieljenik:


Shnoulle (commented):
`strcasecmp` maybe ?

Shnoulle (commented):
Maybe better with strcasecmp ?

PR #2723, opened 2022-11-14
Fixed issue #18461 Survey groups at survey list screen should be listed alphabetically

gabrieljenik:


Shnoulle (approved):

Shnoulle (commented):
It must be default, let me test

Shnoulle (approved):
OK

PR #2727, opened 2022-11-14

gabrieljenik:
- Fix expiration process
- Restore date validation


Shnoulle (approved):
Don't find issue in code, **not tested**

PR #2728, opened 2022-11-16
Optimize PNG files

dilyanpalauzov:
by running
  `for i in $(find -name '*.png'); do optipng -o7 -zm9 $i; done`

This will result less image data transmitted to the browser, and in turn faster load times.

Total size of all png files
before this change: 9 025 455 bytes,
after 7 713 639 bytes


Shnoulle:
Seems some files are on vendor directory. Replaced next time we update

dilyanpalauzov:
Please run `for i in $(find -name '*.png'); do optipng -o7 -zm9 $i; done` whenever that files are updated.

Shnoulle:
Best is to create Pull Request on vendor …

PR #2733, opened 2022-11-16
Fixed issue #18488: Conditions are not properly highlighted if they have additional spaces

Shnoulle:
Dev: trim when view
Dev: trim when save
Dev: uneeded trim in expression

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -3399,9 +3399,9 @@ class LimeExpressionManager
             }
 
             $questionNum = $fielddata['qid'];
-            $relevance = (isset($fielddata['relevance'])) ? $fielddata['relevance'] : 1;
-            $SQrelevance = (isset($fielddata['SQrelevance'])) ? $fielddata['SQrelevance'] : 1;
-            $grelevance = (isset($fielddata['grelevance'])) ? $fielddata['grelevance'] : 1;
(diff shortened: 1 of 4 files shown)


c-schmitz:
I think it solves only part of the problem, as it only trims surrounding spaces of the equation.
It still does not take core of double spaces inside the condition parts, correct?

Shnoulle:
>  It still does not take core of double spaces inside the condition parts, correct?

There are no issue with double or triple space inside relevance equation or line feed etc …
https://github.com/LimeSurvey/LimeSurvey/blob/b10a1b17c3d25d08867fe19cea6915b1b644ffc4/application/helpers/expressions/em_core_helper.php#L2485 don't care of multiple space (don't care of any space)
```Q00.NAOK 
           
== 


"Y"```

===
`Q00.NAOK=="Y"`

There are just the space and line feed after the first `{ ` (and before the last) to allow javascript inside question.

Added it during Expression Manager construction some years ago

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/blob/b10a1b17c3d25d08867fe19cea6915b1b644ffc4/application/helpers/expressions/em_core_helper.php#L2507

c-schmitz (changes requested):
Looks good so far, please check my new information in the bug report.

Shnoulle:
Screen-cast : only one «seems to be a space» character
Peek 17-11-2022 10-36

c-schmitz (approved):

PR #2738, opened 2022-11-21
Fixed issue #18501: Answers with 0 as value was not saved if encrypted

Shnoulle:
Dev: replace !empty by isset && !== ""
Dev: keep not set and null value just in case


Shnoulle:
Need an unit test

Shnoulle:
Need help : don't understand why this test is broken …

Shnoulle:
OK : there are an issue with FailedEmailController somewhere : try to decrypt an empty value, but this value is not a empty string or a null value or 'null' string …

The survey in test don't have any crypted value …

application/models/LSActiveRecord.php
```diff
@@ -424,7 +424,7 @@ public static function decryptSingleOld($value = ''): string
     public static function encryptSingle($value = '')
     {
         // if $value is provided, it would decrypt
-        if (!empty($value)) {
+        if (isset($value) && $value !== "") {
```
gabrieljenik, line 427: If value is "", shouldn't that be encrypted too? Isn't that a piece of data to be encrypted?

I believe that we have 2 problems:
1 - How to encrypt empty data (0, null, "", .. ) everything that matches empty()
2 - On this case I would just add an else.
```
 if (!empty($value)) {
 } else {
   return $value;
}
```

application/models/LSActiveRecord.php
```diff
@@ -399,7 +399,7 @@ public static function decryptSingleOld($value = ''): string
             $sodium = Yii::app()->sodiumOld;
         }
         // if $value is provided, it would decrypt
-        if ($value) {
+        if (isset($value) && $value !== '') {
```
gabrieljenik, line 402: Is this fix on the right function?
By the function name, seems like this function is not supposed to be used.

application/models/LSActiveRecord.php
Shnoulle: before we use empty, the broke with 0 … it's the issue …

Then you 2 cause the issue 18501 … _Answers with 0 as value was not saved if encrypted_

application/models/LSActiveRecord.php
Shnoulle: And ?

application/models/LSActiveRecord.php
olleharstedt: > empty strings should be encrypted as well

This is true.

application/models/LSActiveRecord.php
Shnoulle: > decryptSingle() and decrypt() are still using empty. That was my comment.

Yes : because null OR à or "" are not an encrypted value :) . 

See : https://github.com/LimeSurvey/LimeSurvey/pull/2738#issuecomment-1323242180

application/models/LSActiveRecord.php
Shnoulle: Another issue to report, right ?

olleharstedt: It's a bit complicated as numerical columns _cannot_ be encrypted (nowhere to save the hash), but of course people can write "0" in text questions. :)

Shnoulle: > It's a bit complicated as numerical columns cannot be encrypted (nowhere to save the hash), but of course people can write "0" in text questions. :)

If you have a numerical question encrypted : it was saved in a text column … 

All crypted columns are text columns … date or numérical or single choice.

tests/unit/models/EncryptAttributesTest.php
```diff
@@ -52,6 +52,37 @@ public function testTokenWithoutValidation()
...
+    {
+        // Get our token.
+        $tokens = \TokenDynamic::model(self::$surveyId)->findAll();
+        $this->assertNotEmpty($tokens);
+        $this->assertCount(1, $tokens);
+        $token = $tokens[0];
+        $token->decrypt();
+
+        // Change attribute_1.
+        $token->attribute_1 = '0';
```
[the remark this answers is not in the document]
Shnoulle: Currently ;  not crypted

application/models/LSActiveRecord.php
olleharstedt: > Another issue to report, right ?

Depends on if this fix breaks it or not :sweat_smile:

Shnoulle: It's clearly another issue than «Answers with 0 as value was not saved if encrypted»

And more 4 month after …

Shnoulle:
PS 

> OK : there are an issue with FailedEmailController somewhere : try to decrypt an empty value, but this value is not a empty string or a null value or 'null' string …
> 
> The survey in test don't have any crypted value …


This i totally unrelated to this issue.

application/models/LSActiveRecord.php
olleharstedt: > > It's a bit complicated as numerical columns cannot be encrypted (nowhere to save the hash), but of course people can write "0" in text questions. :)
> 
> If you have a numerical question encrypted : it was saved in a text column …
> 
> All crypted columns are text columns … date or numérical or single choice.

Oh OK, didn't know.

tests/unit/models/EncryptAttributesTest.php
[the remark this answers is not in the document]
olleharstedt: > Currently ; not crypted

OK, and it never was?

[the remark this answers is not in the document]
Shnoulle: No … never …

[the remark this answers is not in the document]
olleharstedt: Ah alright. Well it's not enough to encrypt it, it has to be salted too to not be obvious. But, another issue.

PR #2742, opened 2022-11-22
only one tab should have active class

smohamadabedy:
Fixed issue : application\views\admin\conditions\includes\form_editconditions_header.php all tabs have active class and it's wrong.
all corrected


gabrieljenik:
From the conflict I believe this is already fixed on master.
I will close this.
Please would you reopen it if you feel it should?

Thanks!

PR #2744, opened 2022-11-23
Fixed issue #18498: Questions with custom themes not shown on question list

gabrieljenik:


olleharstedt (approved):

Shnoulle (commented):
See comment about adding a specific relation

application/models/Question.php
```diff
@@ -1060,9 +1060,16 @@ public function search()
...
-            'question_theme' => ['alias' => 'qt', 'condition' => "extends=''"]
         ];
 
+        // The question theme table holds records for question_themes, which can be:
+        // - user created question_themes,
+        // - LS created core question_themes or
+        // - basic question_types (which can be thought as a basic question themes).
+        // As we want to filter by the description of the question type, we perform a join to the question_type record,
+        // not to the question_theme used.
+        $criteria->join = "JOIN {{question_themes}} as qt ON t." . Yii::app()->db->quoteColumnName("type") . " = qt.question_type AND extends = ''";
```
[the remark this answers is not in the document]
gabrieljenik: There is already a getQuestionType() method. Already behaves kind of like a relationship and allows to enhance the returning object.

I think would be confusing to add the relationship as it may lead to sometimes use the method and others to use the relationship.

I would just leave it like this for this fix.

Shnoulle (approved):

c-schmitz (changes requested):
In the question list you can also see the question type. 
With this patch now the question using a custom question type are shown, however instead of naming the question theme, the base type is only shown in the type column.
It should  show either only the name of the question theme or both, base type & question theme.
Please fix.

gabrieljenik:
Nice catch. Thanks!
Sorry we missed it

gabrieljenik:
New commit. We will re test it now. :)

gabrieljenik:
Ready for re-review :)

application/models/Question.php
```diff
@@ -547,6 +547,7 @@ public function getQuestionList($surveyid)
 
     /**
      * @return string
+     * NOTE: Not used anymore. Based on a deprecated method. Should be deprecated.
```
olleharstedt, line 550: Ehm not used = should be removed then? Why keep code that's not used?

olleharstedt (changes requested):

application/models/Question.php
gabrieljenik: Well, it is a public method. I would wait a bit before removing it.
Someone could be using it on a custom development.

Shnoulle: This is not **this** commit that deprecate the function, tight ?

gabrieljenik: No, if recall correctly, we just added the comment as we found that out

Shnoulle: Then : i think it's a good idea to add comments, but not to remove or deprecate on a unrelated commit :)

gabrieljenik: Yes, as we did

Shnoulle: > Yes, as we did

Exactly :)

PR #2745, opened 2022-11-23
Fixed issue #18460: Assigned survey group is gone after survey export/import

gabrieljenik:
- Export: Add survey group details to LSS
- Import: If the LSS includes those details and a group with the same name exists, use it.

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -1217,10 +1217,6 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         $insertdata = array();
 
         foreach ($row as $key => $value) {
-            // Set survey group id to default if not a copy
-            if ($key == 'gsid' & !$isCopying) {
-                $value = 1;
-            }
             if ($key == 'template') {
                 $sTemplateName = (string)$value;
             }
@@ -1272,6 +1268,24 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         if (isset($insertdata['tokenlength']) && $insertdata['tokenlength'] > 35) {
             $insertdata['tokenlength'] = 35;
         }
+
+        // Fix survey group when not copying
+        if (!$isCopying) {
+            // Set to 1 by default
+            $insertdata['gsid'] = 1;
+            // If the xml includes survey group details, try to find the group by name.
+            if (!empty($xml->surveys_groups->rows->row[0]->name)) {
+                $surveyGroupName = (string) $xml->surveys_groups->rows->row[0]->name;
+                $surveyGroup = SurveysGroups::model()->findByAttributes(["name" => $surveyGroupName]);
+                // If a survey group is found with the specified name, assign it to the survey.
+                if (!empty($surveyGroup)) {
+                    $insertdata['gsid'] = $surveyGroup->gsid;
+                } else {
+                    $results['importwarnings'][] = gT("The original survey group couldn't be found.");
+                }
+            }
+        }
+
         /* Remove unknow column */
         $aSurveyModelsColumns = Survey::model()->attributes;
         $aSurveyModelsColumns['wishSID'] = null; // Can not be imported
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -967,6 +967,14 @@ function surveyGetXMLStructure($iSurveyID, $xmlwriter, $exclude = array())
                 . " FROM {{plugin_settings}} as settings JOIN {{plugins}} as plugins ON plugins.id = settings.plugin_id"
                 . " WHERE model='Survey' and model_id=$iSurveyID";
     buildXMLFromQuery($xmlwriter, $slsquery);
+
+    // Survey Group
+    $sgquery = "SELECT sg.*
+    FROM {{surveys_groups}} as sg
+    JOIN {{surveys}} as s ON sg.gsid = s.gsid
+    WHERE s.sid={$iSurveyID}";
+    $excludeFromSurveyGroup = ['sortorder', 'owner_id', 'created', 'modified', 'created_by'];
+    buildXMLFromQuery($xmlwriter, $sgquery, '', $excludeFromSurveyGroup);
 }
 
 /**


application/helpers/admin/import_helper.php
```diff
@@ -1272,6 +1268,24 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         if (isset($insertdata['tokenlength']) && $insertdata['tokenlength'] > 35) {
             $insertdata['tokenlength'] = 35;
         }
+
+        // Fix survey group when not copying
+        if (!$isCopying) {
```
Shnoulle, line 1273: No option ? A checkbox like translinksfields ?

application/helpers/admin/import_helper.php
```diff
@@ -1272,6 +1268,24 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
...
+            $insertdata['gsid'] = 1;
+            // If the xml includes survey group details, try to find the group by name.
+            if (!empty($xml->surveys_groups->rows->row[0]->name)) {
+                $surveyGroupName = (string) $xml->surveys_groups->rows->row[0]->name;
+                $surveyGroup = SurveysGroups::model()->findByAttributes(["name" => $surveyGroupName]);
+                // If a survey group is found with the specified name, assign it to the survey.
+                if (!empty($surveyGroup)) {
+                    $insertdata['gsid'] = $surveyGroup->gsid;
+                } else {
+                    $results['importwarnings'][] = gT("The original survey group couldn't be found.");
```
Shnoulle, line 1284: If there are a checkbox : yes, without : no warning

Shnoulle (commented):
I like to have a checkbox (unchecked by default)

Shnoulle:
Reason to don't import Survey group without confirmation

1. Instance A : Used for Client1 : survey group name was "TemplateSurvey", right to read/export Client1User (With a NDA with restriction to tell you work for Client1)
2. Instance B : Used for Client2 : survey group name was "TemplateSurvey", right to read/export Client2User
3. You have near same survey for Client1 and Client2, you want to copy Client1 survey in instance B, update string to remove all information from Client1 and put information to Client2
4. You export from Instance A  to Instance B : survey is set in "TemplateSurvey" : can be copied by Client2
5. Even if you update it after : Client2 have the information about Client1

Then : 
1. Need a checkbox "Import SurveyGroup"
2. This checkbox must be unchecked by default.

Shnoulle (changes requested):
This clearly need a way to deactivate + inform user on risk OR set default as not import.

gabrieljenik:
> This clearly need a way to deactivate + inform user on risk OR set default as not import.

I am sorry, I don't see this so clearly as you say. 
Actually, from what I understand, isn't it an edge case?

> Even if you update it after : Client2 have the information about Client1

What do you mean?

On general terms, if after importing, got in the wrong survey group, you can always move it right away... right?

Shnoulle:
> > This clearly need a way to deactivate + inform user on risk OR set default as not import.
> 
> I am sorry, I don't see this so clearly as you say. Actually, from what I understand, isn't it an edge case?

Need to have same group on export./import are already an edge case , since Copy is available.

> 
> > Even if you update it after : Client2 have the information about Client1
> 
> What do you mean?
> 
> On general terms, if after importing, got in the wrong survey group, you can always move it right away... right?

Yes but during a time : the survey was shown to any Client2 on the instance. And look at your sentence after.

And : you can write exactly the same for this feature :  **you can always move it right away** then don't create this feature, it's not needed …


> On my opinion with checkboxes for options, people usually bypass them and don't pay much attention to them. So, if someone is slappy, on thse edge cases that you mention, they will get into unwanted situations anyway.

Adding a new feature with potential issue : people usually bypass the new group and think it's on Default group like today …

Shnoulle:
@olleharstedt or @c-schmitz i'm clearly against doing this by default.

Current situation : the new questionnaire is in the default group, which is what is expected. 

Putting it in a group without being sure that it is intended seems (to me) dangerous in terms of confidentiality.

I think 

1. We need a checkbox option
2. The default must be set to Default group when import
3. If not a big warning mus be shown when import (then user can uncheck the checkbox)

Shnoulle (commented):
No advice

olleharstedt:
Yeah, I'm not convinced either.

Shnoulle:
Decision seem already done : no checkbox … i think it's a potential issue …
And it's not a big update to add a checkbox.
(i didn't mind of Permission on SurveyGroup)

olleharstedt:
> Decision seem already done

By whom?

Shnoulle:
> > Decision seem already done
> 
> By whom?

https://bugs.limesurvey.org/view.php?id=18460#c72899

gabrieljenik:
Ready foe re-reviewal

application/helpers/admin/import_helper.php
```diff
@@ -1272,6 +1268,26 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
...
+        // Fix survey group when not copying
+        if (!$isCopying) {
+            // Set to 1 by default
+            $insertdata['gsid'] = 1;
+            // If the xml includes survey group details, try to find the group by name.
+            if (!empty($xml->surveys_groups->rows->row[0]->name)) {
+                $surveyGroupName = (string) $xml->surveys_groups->rows->row[0]->name;
+                $surveyGroupPermissionCriteria = SurveysGroups::getPermissionCriteria();
+                $surveyGroup = SurveysGroups::model()->findByAttributes(["name" => $surveyGroupName], $surveyGroupPermissionCriteria);
+                // If a survey group is found with the specified name, assign it to the survey.
```
Shnoulle, line 1281: This disallow (currently) Permssion event + must check Permission to "import" survey in this group

It's not because you have the right to view SurveyGroup you have the right to import survey in this group

Then must be
```
$surveyGroup = SurveysGroups::model()->findByAttributes(["name" => $surveyGroupName]);
if(!empty($surveyGroup) && SurveysInGroup::hasPermission('survey','import) {
```

application/models/SurveysGroups.php
```diff
@@ -392,10 +392,10 @@ public static function model($className = __CLASS__)
     }
 
     /**
-     * get criteria from Permission
+     * Returns the criteria for assesing user permission over the group.
      * @return CDbCriteria
      */
-    protected static function getPermissionCriteria()
+    public static function getPermissionCriteria()
```
Shnoulle, line 398: No needed if we really check Permission

Shnoulle (changes requested):
Review to really test survey/import permission

application/helpers/admin/import_helper.php
gabrieljenik: You are right. Got confused.

Shnoulle:
About `SurveysInGroup::hasPermission('survey','import')` the function must be fixed here 

https://github.com/LimeSurvey/LimeSurvey/blob/1c19d1c44ed0c806104b7ed005332806fd3a40e3/application/models/SurveysInGroup.php#L74

You're right on this point (related : https://bugs.limesurvey.org/view.php?id=18530)
Seems it's the 1st time we need it :)

application/helpers/admin/import_helper.php
Shnoulle: Still an issue about `SurveysInGroup::hasPermission('survey','import')` this need a fix in hasPermission …

gabrieljenik: Actually, this `getPermissionCriteria` is the same method used by the create survey screen when listing available groups.
That's why we use it. As to try to use the same.

I think that if there is an issue, it is a more general one.

Also, you mention
> It's not because you have the right to view SurveyGroup you have the right to import survey in this group

But then, you also mention this on 18530
> User can add survey in group only if he can see the group. There is no 'Create' for this permission.

I believe there is some kind of problem on permission management for the surveygroups, but for this fix, we are kind of copying what already is available

Shnoulle: > I believe there is some kind of problem on permission management for the surveygroups, but for this fix, we are kind of copying what already is available. So maybe we should just treat that on chat, another ticket or somehting else?

2 fix needed ? 
see https://bugs.limesurvey.org/view.php?id=18530

gabrieljenik: I would just leave this as it is now, which is the same as the create survey.
And then, after 18530, if we need, we update it.
As for simplifying the process and not try to solve all at once

gabrieljenik: Thought that. Was the seed of this approach. But seems better to just update getPermissionCriteria.
Why do I need to fetch all surveygroups if only interested in one?

Shnoulle: > Thought that. Was the seed of this approach. But seems better to just update getPermissionCriteria.

getPermissionCriteria exist on other model, i think i made an error : it need to be in Trait
See related issue : https://bugs.limesurvey.org/view.php?id=16459

And because currently ; it's the only way to get valid list of gsid

> Why do I need to fetch all surveygroups if only interested in one?

because it's already public, see https://github.com/LimeSurvey/LimeSurvey/pull/2752/files?diff=unified&w=0#diff-c2e2b37076b5bae18bcd942f811752f5be549ebcec44fd866ffbba17662ca35eR332
(related issue https://bugs.limesurvey.org/view.php?id=18530)
Here : i add an optional gsid on the list to get only the needed one.

Shnoulle: PS : maybe just try `SurveysInGroup::model()->findByPK($gsid)->hasPermission('surveys', 'import')` ?

Since the issue is related to this ?

gabrieljenik: > Here : i add an optional gsid on the list to get only the needed one.

Actually, we thought about that too. But then the name of the method will not make much sense.
Get a list of survey groups with only a single item?

> getPermissionCriteria exist on other model, i think i made an error : it need to be in Trait
We can create an alias to expose the method.
When you fix the method, it shouldn't impact.

Shnoulle: > We can create an alias to expose the method.
> When you fix the method, it shouldn't impact.

Here : the fix must use `SurveysInGroup::model()->findByPK($gsid)->hasPermission('surveys', 'import')`  right ?

Except current permission system is broken and need to be fixed.
https://github.com/LimeSurvey/LimeSurvey/pull/2752

Since it's another issue : maybe leave with just `SurveysInGroup::model()->findByPK($gsid)->hasPermission('surveys', 'import')` ?

gabrieljenik:
To be picked up after 18350 is finished.

Shnoulle:
> To be picked up after 18350 is finished.

Unsure :) 

This commit can just use current Permission system, and 18350 fix the permission system, no ?

Shnoulle (approved):

olleharstedt:
No unit test for this? Since it has wide security implications, it might be a good idea to make sure the new function lives up to its spec?

application/helpers/admin/import_helper.php
```diff
@@ -1272,6 +1268,30 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         if (isset($insertdata['tokenlength']) && $insertdata['tokenlength'] > 35) {
             $insertdata['tokenlength'] = 35;
         }
+
+        // Fix survey group when not copying
+        if (!$isCopying) {
+            // Set to 1 by default
```
olleharstedt, line 1274: The body of this if-statement could maybe be factored out to a separate function or method, and then that function could be unit-tested.

olleharstedt:
Patrick had a good idea: Survey group should have a UUID to make sure it's not another group that accidentally has the same name + id.

Shnoulle:
> Patrick had a good idea: Survey group should have a UUID to make sure it's not another group that accidentally has the same name + id.

But then :  you disable feature when you want to keep same Group with only GroupName ( Preprod instance and production instance)

gabrieljenik:
> > Patrick had a good idea: Survey group should have a UUID to make sure it's not another group that accidentally has the same name + id.
> 
> But then : you disable feature when you want to keep same Group with only GroupName ( Preprod instance and production instance)

Yes, you would be in a very similar case as if you use the current gsid

gabrieljenik:
Shall we make the automatic test on a new ticket later or prefer not to merge until the test is there?

olleharstedt:
> Yes, you would be in a very similar case as if you use the current gsid

gsid is not unique, so it's not safe.

olleharstedt:
I kinda hate this whole feature. For sure it's not a "bug fix". Maybe we should discuss it in a meeting.

Shnoulle:
> I kinda hate this whole feature. For sure it's not a "bug fix". Maybe we should discuss it in a meeting.

Said it at start …

> Yes, you would be in a very similar case as if you use the current gsid

Using gsid it's worst … 
I send you a lss it was set to any group without any control (if you and me use group)

olleharstedt:
OK so

1) Move this to develop instead of master
2) Make survey groups use UUID instead of only id, and use _that_ UUID for export/import to reduce the risk of assigning wrong survey group by mistake.

gabrieljenik:
> Make survey groups use UUID instead of only id, and use that UUID for export/import to reduce the risk of assigning wrong survey group by mistake.

But if we do UUID for import/export and UUID is kind of unique, the only use case would be importing/exporting on the same installation? Does that makes sense?

What if we calculate a hash based on the main domain name, survey group name?

olleharstedt:
> But if we do UUID for import/export and UUID is kind of unique, the only use case would be importing/exporting on the same installation? Does that makes sense?

What kind of weird use-case is the target here? :joy: 

> What if we calculate a hash based on the main domain name, survey group name?

OK _if_ we also hash with encryptionsecretboxkey in security.php.

Shnoulle:
> Make survey groups use UUID instead of only id, and use that UUID for export/import to reduce the risk of assigning wrong survey group by mistake.

My opinion : leave name as default with a checkbox is a better solution.

For example : we can create a Demo Group, and set all [demo survey](https://github.com/LimeSurvey/LimeSurvey/tree/master/docs/demosurveys) in in Demo group .

Sample plugin survey can be interesting too ;)

**BUT** UUID only is OK (or nothing too)

gabrieljenik:
> What kind of weird use-case is the target here?

Exactly. What I mean is that if we implement UUID, there wouldn't be many other use cases.
All this feature is kind of useles with UUID

olleharstedt:
> All this feature is kind of useles with UUID

I mean, it's obviously a security issue without UUID. Did anyone consider to just have a select box during import to choose which survey group to assign it to? Does it have to be more complicated?

gabrieljenik:
> > All this feature is kind of useles with UUID
> 
> I mean, it's obviously a security issue without UUID. Did anyone consider to just have a select box during import to choose which survey group to assign it to? Does it have to be more complicated?

I don't think is a security issue. Still, dropdown doesn't seem a bad idea if that's something which would be usefull to unlock this situation :)

c-schmitz:
@gabrieljenik Can you fix the conflicts and then ask coderabbitai for a review?

Shnoulle:
PS : we need a new feature request : when export/import allow set group by name :)

c-schmitz:
@coderabbitai review

application/helpers/admin/import_helper.php
```diff
@@ -1272,6 +1268,30 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
...
+                    if (!empty($surveysInGroup) && $surveysInGroup->hasPermission('surveys', 'import')) {
+                        // If a survey group is found with the specified name, and the user has import permission on it, assign it to the survey.
+                        $insertdata['gsid'] = $surveyGroup->gsid;
+                        $results['importwarnings'][] = sprintf(gT("The survey was assigned to the '%s' group."), $surveyGroup->title);
+                    } else {
+                        $results['importwarnings'][] = gT("The original survey group couldn't be found. The survey was assigned to the default group.");
+                    }
+                } else {
+                    $results['importwarnings'][] = gT("The original survey group couldn't be found. The survey was assigned to the default group.");
+                }
```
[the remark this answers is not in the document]
gabrieljenik: That message is because if the user doesn't have permissions for certain piece of data, we don't want to hint anything about it, not even that the piece of data exists.

assets/packages/adminsidepanel/lib/surveysettings.js
```diff
@@ -245,3 +245,11 @@ function guidGenerator() {
...
     return (S4() + S4() + '-' + S4() + '-' + S4() + '-' + S4() + '-' + S4() + S4() + S4());
 }
+
+$(document).on('change', '#importsurvey #surveysgroup', function () {
+    if ($(this).val() == 'from_survey') {
+        $('#survey_group_import_warning').removeClass('hide');
+    } else {
+        $('#survey_group_import_warning').addClass('hide');
+    }
+});
```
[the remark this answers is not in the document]
gabrieljenik: This is ok. surveysgroup is initialized as "default"

gabrieljenik:
@c-schmitz Done

c-schmitz:
@gabrieljenik Please check on the CI error

application/helpers/admin/import_helper.php
```diff
@@ -1248,6 +1248,7 @@ function getTableArchivesAndTimestamps(int $sid, string $baseTable = 'old_survey
  * @param string|null $baselang            Optional base language to set or ensure on the imported survey.
  * @param string|null $targetSurveyGroup   Optional target survey group strategy (e.g., 'from_survey' to attempt using the source survey's group); when null defaults to standard group assignment.
  * @return array|null Result array containing counters, warnings and mappings (including 'newsid' on success), or null for unsupported file extensions.
+ */
```
Shnoulle, line 1251: coderabbit here :)

gabrieljenik: Yes!! haha he made mistakes (finally)!

gabrieljenik:
@tiborpacalat Please test carefully... This is a story which is +3 years old :)

PR #2746, opened 2022-11-25
fixed issue: pageSize dropdown in responses-grid didn't work

twilligls:
Dev:

### DIFF
--- a/assets/scripts/admin/listresponse.js
+++ b/assets/scripts/admin/listresponse.js
@@ -140,6 +140,9 @@ function onDocumentReadyListresponse() {
         $('#change-display-mode-form').find('input[type=submit]').trigger('click');
     });
 
+    $(document).on("change", '#pageSize', function () {
+        $.fn.yiiGridView.update('responses-grid', {data: {pageSize: $(this).val()}});
+    });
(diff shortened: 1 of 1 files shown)


Shnoulle:
We don't update LSGrid to have an helper for this ?

See : https://github.com/LimeSurvey/LimeSurvey/commit/739ac13b5dd72740ceb0b36dc3835bac90f139f6

Why it don't work here ?
If it broke in response : it's surely broke elsewhere ?

assets/scripts/admin/listresponse.js
```diff
@@ -140,6 +140,9 @@ function onDocumentReadyListresponse() {
         $('#change-display-mode-form').find('input[type=submit]').trigger('click');
     });
 
+    $(document).on("change", '#pageSize', function () {
```
Shnoulle, line 143: Just adding changePageSize class ?

Or update https://github.com/LimeSurvey/LimeSurvey/blob/bc809236a180e79f06b02c580b4435b0ad830605/application/extensions/LimeGridView/LimeGridView.php#L79 to use `data-action` or something else ?

PR #2752, opened 2022-11-30
Fixed issue #18530: No permission control for SurveyGroup when update Survey

Shnoulle:
Dev: allow to filter by gsid SurveysGroups::getSurveyGroupsList
Dev: fix SurveysInGroup::hasPermission : add specific control on create and import
Dev: fix SurveysInGroup::hasPermission crud are already the same than globalCrud
Dev: Add a constant for default group (1)

### DIFF
--- a/application/models/SurveysGroups.php
+++ b/application/models/SurveysGroups.php
@@ -25,6 +25,9 @@ class SurveysGroups extends LSActiveRecord implements PermissionInterface
     /* @var boolean|integer alwaysavailable : set default, and set for old DB , usage of integer for DB compatibility */
     public $alwaysavailable = 0;
 
+    /* @var integer default SurvetGroup */
+    public const DEFAULTGROUP = 1;
+
     /**
      * @return string the associated database table name
      */
@@ -46,7 +49,7 @@ class SurveysGroups extends LSActiveRecord implements PermissionInterface
     protected function afterFind()
     {
         parent::afterFind();
-        if ($this->gsid == 1) {
+        if ($this->gsid == self::DEFAULTGROUP) {
             $this->alwaysavailable = 1;
         }
     }
@@ -314,7 +317,7 @@ class SurveysGroups extends LSActiveRecord implements PermissionInterface
             $button .= '<a class="btn btn-sm btn-default" href="' . $sSurveySettingsUrl . '" role="button" data-toggle="tooltip" title="' . gT('Survey settings') . '"><i class="fa fa-cog" aria-hidden="true"></i><span class="sr-only">' . gT('Survey settings') . '</span></a>';
         }
         /* Can not delete group #1 + with survey (or move it to hasPermission function ?) */
-        if ($this->gsid != 1 && !$this->hasSurveys && $this->hasPermission('group', 'delete')) {
+        if ($this->gsid != self::DEFAULTGROUP && !$this->hasSurveys && $this->hasPermission('group', 'delete')) {
             $button .= '<span data-toggle="tooltip" title="' . gT('Delete survey group') . '"><a class="btn btn-sm btn-default" href="#" data-post-url="' . $sDeleteUrl . '" data-target="#confirmation-modal" role="button" data-toggle="modal" data-message="' . gT('Do you want to continue?') . '"><i class="fa fa-trash text-danger " aria-hidden="true"></i></a></span>';
         }
         $button .= "</div>";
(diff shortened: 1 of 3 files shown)


Shnoulle:
Follow/maintenance : https://github.com/LimeSurvey/LimeSurvey/pull/1709

Shnoulle:
@gabrieljenik : i rework on it with your advice : getAccessTo($sird) for the name of function ?

application/models/SurveysGroups.php
```diff
@@ -323,22 +326,22 @@ public function getButtons()
 
     /**
      * Get the group list for current user
+     * @gsid null|integer : allow to restrict to a specigf gif
      * @return array
      */
-    public static function getSurveyGroupsList()
+    public static function getSurveyGroupsList($gsid = null)
```
gabrieljenik, line 332: I am not sure about this.
I understand the usefullness.

Still, name wouldn't match the behaviour when using it with an argument.
Doesn't feel right.

If needed, maye we should:
a) create a new method? What name would it be?
b) create an alias to expose the getPermissionCriteria method?

Right now sounds like something which may need to be unwinded and the more we use it, the hardest will be to unwind it

gabrieljenik:
> @gabrieljenik : i rework on it with your advice : getAccessTo($sird) for the name of function ?

It is an honor to hear that. Thanks!

From my point of view, I think exposing the criteria is better.
Maybe just creating a new scope? (not sure if it fits the current situation).

If not, the new name could be hasReadAccessTo($gsid)

Shnoulle:
> Maybe just creating a new scope? (not sure if it fits the current situation).

It can be a great idea … but seems complex … maybe you have a sample ? 
The only scope is see are really simple 

> If not, the new name could be hasReadAccessTo($gsid)

I think of a findWithPermssion or findByPkWithPermissin
(but scope are better)

application/models/SurveysGroups.php
Shnoulle: Both can be done, but for getPermissionCriteria i think it's more related to https://bugs.limesurvey.org/view.php?id=16459 (and moving it to Trait)

gabrieljenik:
> > Maybe just creating a new scope? (not sure if it fits the current situation).
> 
> It can be a great idea … but seems complex … maybe you have a sample ? The only scope is see are really simple

Something like this may work?
https://www.yiiframework.com/doc/guide/1.1/en/database.ar#parameterized-named-scopes

application/models/SurveysGroups.php
gabrieljenik: We create an alias now, to expose it.
If later we move it to a trait, it will still work. right?

Shnoulle:
> > > Maybe just creating a new scope? (not sure if it fits the current situation).
> > 
> > 
> > It can be a great idea … but seems complex … maybe you have a sample ? The only scope is see are really simple
> 
> Something like this may work? https://www.yiiframework.com/doc/guide/1.1/en/database.ar#parameterized-named-scopes

Yeah! Great thanks .

https://github.com/LimeSurvey/LimeSurvey/pull/2752/commits/dda7ad83ed5e6a6b964c05819e20e79d7dd0a9cc

Shnoulle:
scope name withpermission, maybe better withaccess ? or withminimalpermission ?

application/models/SurveysGroups.php
Shnoulle: Moving to scope

Shnoulle: >  If later we move it to a trait, it will still work. right?

See https://github.com/LimeSurvey/LimeSurvey/pull/2754

application/models/SurveysGroups.php
```diff
@@ -391,6 +389,16 @@ public static function model($className = __CLASS__)
         return $model;
     }
 
+    /**
+     * Scope for permission
+     * @return $this
+     */
+    public function withpermission()
```
gabrieljenik, line 396: name should be withPermission?
or better withReadPermission?

application/models/SurveysInGroup.php
```diff
@@ -73,22 +73,18 @@ public static function getMinimalPermissionRead()
      */
     public function hasPermission($sPermission, $sCRUD = 'read', $iUserID = null)
     {
-        /* If have global for surveys : return true */
-        $sGlobalCRUD = $sCRUD;
-        if (($sCRUD == 'create' || $sCRUD == 'import')) { // Create and import (token, response , question content …) need only allow update surveys
-            $sGlobalCRUD = 'update';
-        }
-        if (($sCRUD == 'delete' && $sPermission != 'survey')) { // Delete (token, response , question content …) need only allow update surveys
```
gabrieljenik, line 81: Is it ok to remove this?

application/models/SurveysInGroup.php
```diff
@@ -73,22 +73,18 @@ public static function getMinimalPermissionRead()
      */
     public function hasPermission($sPermission, $sCRUD = 'read', $iUserID = null)
```
gabrieljenik, line 74: As you are updating so much about this function, I think it would be very helpfull to comment what does this function do and also about the logic inside. To me it was confusing from the very begining how different permissions from different objects interact here.

This will need a bigger test, so this instructions will be usefull for testing it.

gabrieljenik:
I realize now. Should this PR be closed?

application/models/SurveysInGroup.php
Shnoulle: Yes, 

1. the only permission are 'surveys', and there are a clear right about survey
2. create and import are done after in this commit (and fixed)

application/models/SurveysInGroup.php
Shnoulle: This model are used only for survey right inside this group

Permission are call here :  https://github.com/LimeSurvey/LimeSurvey/blob/e54d2f532b363e474f679e6392e59da06efb8000/application/models/Survey.php#L2287

Shnoulle:
> I realize now. Should this PR be closed?

If you think the other is better : yes, but more update …

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2754

PR #2762, opened 2022-12-02
Fixed issue #14310: Google Analytics functionality doesn't work

gabrieljenik:


Shnoulle (approved):
No issue 

Still like to see this in a (core) plugin …

olleharstedt:
> Still like to see this in a (core) plugin …

Or maybe core question themes? It's weird that it's in "short text", should be its own thing. But for later...

olleharstedt:
Oh right, I thought it was the Google Maps question type. Sorry, disregard.

PR #2788, opened 2022-12-13
Fixed issue #18220: Question image theme don't seems to work

gabrieljenik:
DEV: LB-82

Moving question themes to their own folders.
Making the question themes to show the images as options if the answer option is a URL.
If not, the content of the answer option is shown as its parent question theme does.


application/core/LS_Twig_Extension.php
```diff
@@ -692,4 +692,44 @@ public static function getLanguageRTL($sLanguageCode)
         Yii::app()->loadHelper('surveytranslator');
         return getLanguageRTL($sLanguageCode);
     }
+
+    /**
+     * Returns true if the argument is an URL (with scheme and domain).
```
Shnoulle, line 697: I don't understand : why not use `filter_var($url, FILTER_VALIDATE_URL)` and `parse_url($url,PHP_URL_SCHEME)` ?

Shnoulle: OK : must validate `/` too

Then why not 
```
$urlcheck = $url;
if ($url[0] ==="/") {
 $urlcheck = App()->getBaseUrl(true) . $ur;
}
return filter_var($urlcheck, FILTER_VALIDATE_URL);
```

I really think it's best to use PHP included function than hacked solution.

gabrieljenik: > I really think it's best to use PHP included function than hacked solution.

This is from Symfony.

> I don't understand : why not use filter_var($url, FILTER_VALIDATE_URL) and parse_url($url,PHP_URL_SCHEME) ?

FILTER_VALIDATE_URL seems to be having some problems and lacks flexibility.
I believe it may fail with relative URLs, localhost like URLS and I think there are other cases. (read some comments which identify bugs on it).

Here we don't want to validate, but to identify something that can be used as a URL.
Bear in mind we need to be flexible here as to allow user entered (non validated) entries.

The idea is not to find the perfect solution here. Just an acceptable one and then we can enhance it

Shnoulle: >  I believe it may fail with relative URLs, localhost like URLS and I think there are other cases. (read some comments which identify bugs on it).

Yes, i add another solution for string starting with `/`

gabrieljenik: I am sorry, for now, I feel more comfortable with the symphony solution, which we already tested with some edge cases.

When we start adding cases like the `/` you mention, we create a new hacked solution, which wasnt tested.
I don't think it is worth redoing this.

Shnoulle: > When we start adding cases like the `/` you mention, we create a new hacked solution, which wasnt tested. I don't think it is worth redoing this.

the / case are in symphony

Shnoulle (approved):
Don't like using regexp for URL in place of PHP core solution, but don't find issue.

Not tested

Shnoulle (approved):
Lot of HTML issue in previous code,
But no issue found in the update

PR #2791, opened 2022-12-14
Dev: #18556: Possible optimisation when lot of survey

Shnoulle:
Dev: get only needed part for survey listing
Dev: don't search user for all survey found

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -172,28 +172,32 @@ function isStandardTemplate($sTemplateName)
 */
 function getSurveyList($bReturnArray = false)
 {
-    static $cached = null;
+    static $static = null;
     $bCheckIntegrity = false;
     $timeadjust = getGlobalSetting('timeadjust');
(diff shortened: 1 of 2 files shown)


Shnoulle:
I surely used this code in a fork, need a lot optimization.

I create PR for master  when it tested a lot.

application/helpers/common_helper.php
```diff
@@ -172,28 +172,32 @@ function isStandardTemplate($sTemplateName)
...
+    if (is_null($static)) {
+        $criteria = new CDBCriteria();
+        $criteria->select = array(
+            'sid','owner_id','active','expires'
+        );
         $surveyidresult = Survey::model()
             ->permission(Yii::app()->user->getId())
-            ->with('languagesettings')
-            ->findAll();
+            ->with('defaultlanguagetitle')
```
gabrieljenik, line 188: I like this idea.

Shnoulle: With a low quality connection between SQL and WEB server : you really see a big difference.

Must redo for master here …

Shnoulle:
3.X version,

PR #2793, opened 2022-12-15
Fixed issue #18543: Question Type "Array (texts)" crashes when in combination with "array filter", "Array filter style": disabled and "show totals"

gabrieljenik:
There were two problems:

1) When sumTable set the total of the rows and columns, the jquery selector for updating the input field was based only on the "disabled" condition. It did not check that it was the input corresponding to the total, for example by class or id. 

Under normal conditions, the only one disabled is the total, but when you use "Array Filter Style" = "disabled" there are more fields disabled, which created confusions.

It also created loops. 
After the total field is updated, some events are triggered and the checkconditions is executed again.
When that was triggered for the "disabled" filtered fields, it created a loop.

2) In the "relevance:on", it removed the disabled from all the entries in the row, including the total one, which should always be disabled.

The ticket error was caused by point 1, but without fixing point 2 it would still look bad, so I fixed both.


assets/packages/limesurvey/survey.js
```diff
@@ -61,7 +61,7 @@ function triggerEmRelevanceSubQuestion(){
         data = $.extend({style:'hidden'}, data);
         $(this).removeClass("ls-irrelevant ls-"+data.style);
         if(data.style=='disabled'){
-            $(event.target).find('input').each(function(itrt, item ){
+            $(event.target).find('input:not(.total)').each(function(itrt, item ){
```
Shnoulle, line 64: Maybe `$(event.target).find('.answer-item input)')` 
I really dislike to use `total` class only … create a new specific class, or apply only to needed input.

Shnoulle (commented):
Can you check with `.answer-item input` selector ?

Shnoulle:
Seems test broken with multiple and radio 

input stay disabled 
[limesurvey_survey_139438.lss.txt](https://github.com/LimeSurvey/LimeSurvey/files/10261614/limesurvey_survey_139438.lss.txt)

gabrieljenik:
What do you think now?

Shnoulle (approved):
Code review done,
To be tested again

Shnoulle:
Firefox, asset resetted (i validate it's this file) : recursion again :(
[limesurvey_survey_Filter4.lss.txt](https://github.com/LimeSurvey/LimeSurvey/files/10270852/limesurvey_survey_Filter4.lss.txt)
Capture d’écran du 2022-12-20 18-35-27

Shnoulle:
OK !
Tested OK

gabrieljenik:
> OK ! Tested OK

OK, the issue was the minified file was missing.

Shnoulle:
> > OK ! Tested OK
> 
> OK, the issue was the minified file was missing.

debug=0 issue … didn't happen if debug set … ouf …

twilligls:
works for me as well!

PR #2794, opened 2022-12-15
Fixed issue #18507: {ANSWERTABLE} not working in basic admin notification

gabrieljenik:


application/helpers/frontend_helper.php
```diff
@@ -527,12 +527,16 @@ function sendSubmitNotifications($surveyid, array $emails = [], bool $return = f
                 $responseId = $sRecipient['responseId'];
                 $notificationRecipient = $sRecipient['recipient'];
                 $emailLanguage = $sRecipient['language'];
+                $aReplacementVars['ANSWERTABLE'] = getResponseTableReplacement($surveyid, $responseId, $emailLanguage, $bIsHTML);
+                LimeExpressionManager::updateReplacementFields($aReplacementVars);
```
Shnoulle, line 531: If we have one email in aEmailNotificationTo and aEmailResponseTo , we do the action 2 times.

But worst : if we have 10 email : we do it 10 times … 
Best is to 

1. Add `$aReplacementVars['ANSWERTABLE']` at https://github.com/LimeSurvey/LimeSurvey/blob/660ba79064437256f60cc64a4efaa518c1f0a241/application/helpers/frontend_helper.php#L516
2. remove the 4 action … done : 4 times 

```
                $aReplacementVars['ANSWERTABLE'] = getResponseTableReplacement($surveyid, $responseId, $emailLanguage, $bIsHTML);
                LimeExpressionManager::updateReplacementFields($aReplacementVars);
```

Shnoulle (changes requested):
Need improvement … do action one time only and not for every emails …

application/helpers/frontend_helper.php
gabrieljenik: The issue is that the table contents depends on the response id.
Although usually this function is used for a single response, it is coded in a way where it could be used for mutlple recipients.

Check here:
https://github.com/LimeSurvey/LimeSurvey/blob/660ba79064437256f60cc64a4efaa518c1f0a241/application/helpers/frontend_helper.php#L527

Response id is set inside the loop.
I will keep on reviewing.

Shnoulle: Oh shit !
You're right … 

Must be reconstructed to construct table only one time by response id …

Shnoulle (approved):
Original code is really badly optimized …

Shnoulle:
https://bugs.limesurvey.org/view.php?id=18563

PR #2795, opened 2022-12-15
Fixed issue #18544: Update Theme Options: Survey Group permissions are not checked correctly

gabrieljenik:
Changed permission type being checked and object whose is checked


Shnoulle (approved):
:+1:

Shnoulle (approved):
I think we don't need to test if SurveyGroup exist bit :+1:

PR #2796, opened 2022-12-15
Fixed issue #18503: Label sets: When adding a label set e.g. at the edit responses screen, the loaded label sets can not be searched

gabrieljenik:


Shnoulle (approved):

Shnoulle:
We don't have it elsewhere ?

Shnoulle (approved):
Not tested

PR #2797, opened 2022-12-15
Fixed issue #18471: Data Entry: Data entry with multiple languages does not work

gabrieljenik:


application/helpers/common_helper.php
```diff
@@ -2335,9 +2335,8 @@ function languageDropdown($surveyid, $selected)
     $baselang = Survey::model()->findByPk($surveyid)->language;
     array_unshift($slangs, $baselang);
     $html = "<select class='listboxquestions' name='langselect' onchange=\"window.open(this.options[this.selectedIndex].value, '_top')\">\n";
-
     foreach ($slangs as $lang) {
-        $link = Yii::app()->homeUrl . ("/admin/dataentry/sa/view/surveyid/" . $surveyid . "/lang/" . $lang);
+        $link = Yii::app()->createUrl("admin/dataentry/sa/view/surveyid/". $surveyid . "/lang/" . $lang);
```
Shnoulle, line 2339: I personnaly prefer `Yii::app()->createUrl("admin/dataentry/sa/view/surveyid/". $surveyid, ['lang' => $lang])` for me lang is a parameter, i know sureyid too, but less clear

Shnoulle (approved):
OK
Needed :)

application/helpers/common_helper.php
gabrieljenik: These are written with different patterns all over the codebase.... We should set a std

Shnoulle: Yes , i know …

olleharstedt: Something for the code quality guide?

gabrieljenik: Defintely, I will let you set the standard :)

PR #2799, opened 2022-12-16
Fixed issue #18559: Potential 500 error with LimeMailer and PHP8

Shnoulle:
Dev: Use PHPMailer default for all if empty by plugin (or before)
Dev: really needed currently by to only

### DIFF
--- a/application/core/LimeMailer.php
+++ b/application/core/LimeMailer.php
@@ -535,8 +535,9 @@ class LimeMailer extends \PHPMailer\PHPMailer\PHPMailer
             $this->Sender = $event->get('bounce', '');
         }
         $this->eventMessage = $event->get('message');
-        /* Need loose compare (default to true) */
-        if ($event->get('send', true) == false) {
+        /* Need loose compare : if null, return true (default) */
+        /* Plugin can send anything for false : don't break API by move to strict compare */
+        if (!$event->get('send', true)) {
             $this->ErrorInfo = $event->get('error');
             return $event->get('error') == null;
         }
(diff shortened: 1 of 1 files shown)


gabrieljenik (approved):

Shnoulle:
> Who tested it? In production?

me … i add the "Hard to test" tag ;) 

Fix another plugin : https://gitlab.com/SondagesPro/SurveyAccess/surveyChaining/-/commit/bb405f14f3654d548de3b81cfb67accfe7ef72ab

I can send the previous plugin again.

The issue is when a plugin set email to false

application/core/LimeMailer.php
```diff
@@ -515,24 +515,27 @@ private function manageEvent($eventParams = array())
...
-            $this->to = $event->get('to');
+            /* Set default as array to avoid the to to null and broke on PHP8 */
+            $this->to = $event->get('to', array());
         }
         if (empty($updateDisable['bounce'])) {
-            $this->Sender = $event->get('bounce');
+            $this->Sender = $event->get('bounce', '');
         }
         $this->eventMessage = $event->get('message');
+        /* Need loose compare (default to true) */
```
olleharstedt, line 538: Why does it need lose compare?

olleharstedt: OK, can you add that motivation in the comment, please? Alternatively use `=== "" || === 0`.

Shnoulle: Next time i don't add comment … it was loose compare before …

Shnoulle:
```
        /* Need loose compare : if null, return true (default) */
        /* Plugin can send anything for false : don't break API by move to strict compare */
```

olleharstedt: Don't be grumpy :sweat_smile: Thanks.

Shnoulle: > Don't be grumpy 

I can not : i'm french ;) (and  old)

PR #2803, opened 2022-12-20
Gc05 cr 761 buttons revisited

twilligls:
Dev:


application/views/admin/user/personalsettings.php
```diff
@@ -139,11 +139,11 @@
                             </div>
                             <div class="row">
                                 <div class="col-lg-6">
-                                    <button type="button" class="btn btn-outline-secondary btn-warning " id="selector__showChangePassword" style="color: white; outline: none;">
+                                    <button type="button" class="btn btn-warning " id="selector__showChangePassword" style="color: white; outline: none;">
```
Shnoulle, line 142: No way to remove inline style ? with color:#white no way to use different theme easily.

And outline must be **never** removed except if it was replaced (keyboard navigation)

application/views/admin/user/personalsettings.php
```diff
@@ -139,11 +139,11 @@
...
                                         <i class="ri-lock-fill"></i>
                                         <?= gT("Change password") ?>
                                     </button>
-                                    <button type="button" class="btn btn-outline-secondary btn-warning " id="selector__showChangeEmail" style="color: white; outline: none;">
+                                    <button type="button" class="btn btn-warning " id="selector__showChangeEmail" style="color: white; outline: none;">
```
Shnoulle, line 146: Same : remove inline style

application/views/admin/user/personalsettings.php
Shnoulle: https://getbootstrap.com/docs/5.2/components/buttons/ default btn-warning was dark on yellow, with white on yellow : you make it unreadable for a lot of user 

contrast of 1.63 even in A it's broke …

twilligls: Oh, didn't see your comment. Will check the style. Don't know why there is inline style for color anyway, wasn't me. :)

I understand correctly:  style="outline: none;" must stay?

Shnoulle: > Oh, didn't see your comment. Will check the style. Don't know why there is inline style for color anyway, wasn't me. :)

Yes, i know but since you update it :)

> I understand correctly: style="outline: none;" must stay?

No, usage of outline:none is really a bad usage : http://www.outlinenone.com/

the best solution (IMO) it's allow default browser behavior, if not : there are a possible solution (using focus-visible) but must be done with care.

https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html
https://www.w3.org/TR/2015/NOTE-WCAG20-TECHS-20150226/F78 


Else : in general : avoid all inline style … move to css, and if possible to scss in a dedicated variable scss file.

twilligls: I don't see any reason why outline: remove is there, so I will delete the whole inline style.

Yes, we're trying to outsource all components into dedicated scss files. As the legacy code is huge, it's no wonder that some inline styles are overseen.

Thanks for your eyes.

PR #2805, opened 2022-12-20
Fixed issue #LB-60: Fixed JS part for copy functionality

apacheborys:
Fixed issue #18564: fixed handling response from BE

### DIFF
--- a/application/extensions/TopbarWidget/views/includes/copyQuestionTopbarRight_view.php
+++ b/application/extensions/TopbarWidget/views/includes/copyQuestionTopbarRight_view.php
@@ -29,12 +29,12 @@
                 code
               },
               success: (data) => {
-                if (data) {
-                    $('#question-title-warning').text(data);
-                    $('#question-title-warning').removeClass('hidden');
-                } else {
-                    $("#submit-copy-question").click();
-                }
+                  if (data.hasOwnProperty('message') && data.message === null) {
+                      $("#submit-copy-question").click();
+                  } else {
+                      $('#question-title-warning').text(data.hasOwnProperty('message') ? data.message : data);
+                      $('#question-title-warning').removeClass('d-none');
+                  }
               },
               error: (data) => {
                 alert('Internal error: ' + data);


application/extensions/TopbarWidget/views/includes/copyQuestionTopbarRight_view.php
```diff
@@ -29,12 +29,12 @@
                 code
               },
               success: (data) => {
-                if (data) {
-                    $('#question-title-warning').text(data);
-                    $('#question-title-warning').removeClass('hidden');
-                } else {
-                    $("#submit-copy-question").click();
-                }
+                  if (data.hasOwnProperty('message') && data.message === null) {
```
gabrieljenik, line 32: Didn't dig a lot into the issue, but just smell it... Shold the condition be like this maybe?
`!data.hasOwnProperty('message') || data.message === null`

apacheborys: According to changes on BE side, message item is mandatory. That's why I checking for item existence with null value. For me it seems like anti-pattern too. But we already have these changes in BE. WDYT?

application/controllers/QuestionAdministrationController.php:3036

image

gabrieljenik: So if there is no message property, data will be just showed as before. Also, this is already on develop? Seems OK then :)

gabrieljenik (approved):
Not the best pattern, but if it already working on develop, good!
Haven't tested it

application/extensions/TopbarWidget/views/includes/copyQuestionTopbarRight_view.php
apacheborys: I'm not sure I got you correctly. Message property will be still permanently in response. And acceptance criteria to have a null as value (according to BE code). And yeah, it's already fixed in dev branch ✅

PR #2809, opened 2022-12-21
Dev #8436: Enable / disable users and fields to manage users

gabrieljenik:
- Test Update

UserExpired Test was failing when attempting to clear the date picker by clikcing the clear button.


olleharstedt:
Less waiting?

gabrieljenik:
Different kind of clearing the date.
Instead of pressing the button, send keys to the input box

olleharstedt:
Testing this, thank you

PR #2812, opened 2022-12-22
Fixed issue #18516: Create Survey button allows multiple clicking and may create multiple copies

gabrieljenik:


Shnoulle (commented):
It's really an issue ? All web work like that …

Shnoulle (approved):

Shnoulle:
Maybe a global better system must be constructed, we can not add workaround like this again and again …

Have a global clean solution must be constructed.

gabrieljenik:
> Maybe a global better system must be constructed, we can not add workaround like this again and again …

There is already a global mechanism, but it works for most screens as they only have 1 clickable save button.
Here we have 2. That's why a workaround.

I would live like this. If more screens appear, we rview a global mechanism.
Also, before BS5 I wouldn't do much more

Shnoulle:
I already approve the commit :) testing now.

> There is already a global mechanism, but it works for most screens as they only have 1 clickable save button.

Have the real button at bottom is really best , see all forms : have button at end . At top it's a feature ;)

>  Also, before BS5 I wouldn't do much more

Totally right

application/views/surveyAdministration/tabCreate_view.php
```diff
@@ -198,6 +198,9 @@
 
         $('#addnewsurvey').on('submit',  function(event){
             event.preventDefault();
+            // Disable both buttons. Normally there's no need to re-enable them. The 'save-form-button' may already be disabled by it's onclick event.
+            $('#create-survey-submit').prop('disabled', true);
+            $('#save-form-button').addClass('disabled').attr('onclick', 'return false;');
```
Shnoulle, line 203: Question : `$('#save-form-button').addClass('disabled').off('click');` didn't work ?

Shnoulle (commented):
Maybe must move logic to https://github.com/LimeSurvey/LimeSurvey/blob/b4ec7b26c54b2c8a3ab6f16a1e894f8fcf73e8d3/assets/packages/adminbasics/src/parts/save.js#L90 ?

something like actionWhenSubmit function

Ands then we can really fix the loading state system (doing it only when submit …)

gabrieljenik:
> Maybe must move logic to
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/b4ec7b26c54b2c8a3ab6f16a1e894f8fcf73e8d3/assets/packages/adminbasics/src/parts/save.js#L90
> 
> ?
> something like actionWhenSubmit function
> 
> Ands then we can really fix the loading state system (doing it only when submit …)

We could consider it when re approaching it generarly after BS5

application/views/surveyAdministration/tabCreate_view.php
gabrieljenik: Haven't tried it. Not sure it is worth trying.
The button has some JS onthe onclick attribute, so guess would be good to keep on in the same line when handling events.

Shnoulle: > Haven't tried it. Not sure it is worth trying. The button has some JS onthe onclick attribute, so guess would be good to keep on in the same line when handling events.

`$('#save-form-button').addClass('disabled').off('click').removeAttr('onclick');` ?

Not important …

PR #2821, opened 2022-12-30
Fixed issue #18517: Survey editing : Survey creation date is wrong

gabrieljenik:


application/models/Survey.php
```diff
@@ -550,7 +550,7 @@ public function rules()
             array('running', 'safe', 'on' => 'search'),
             array('expires', 'date','format' => ['yyyy-M-d H:m:s.???','yyyy-M-d H:m:s','yyyy-M-d H:m'],'allowEmpty' => true),
             array('startdate', 'date','format' => ['yyyy-M-d H:m:s.???','yyyy-M-d H:m:s','yyyy-M-d H:m'],'allowEmpty' => true),
```
Shnoulle, line 552: I know you want to fix each issue separately, but here seems there are same issue just 2 lines before.

Can we fix it at same time, it take less time to fix directly it than to report an issue and create another PR …

Shnoulle (commented):
See comment

application/models/Survey.php
Shnoulle: I can test i think , but yes : not really needed.

gabrieljenik: I would make a different PR, maybe a direct commit?

Shnoulle: Unsure  …

gabrieljenik: Undid the change. That format validator pattern was ok

Shnoulle: Then array('datecreated', 'default', 'value' => date("Y-m-d H:i:s")), didn't fix issue

application/models/Survey.php
```diff
@@ -493,7 +493,7 @@ public function rules()
             array('sid', 'numerical', 'integerOnly' => true,'min' => 1), // max ?
             array('sid', 'unique'),// Not in pk
             array('gsid', 'numerical', 'integerOnly' => true),
-            array('datecreated', 'default', 'value' => date("Y-m-d H:m:s")),
+            array('datecreated', 'default', 'value' => date("Y-m-d H:i:s")),
```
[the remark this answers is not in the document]
Shnoulle: You give me the issue ;)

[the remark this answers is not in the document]
gabrieljenik: Well, that date default value is set by the date function, which uses "i" for minutes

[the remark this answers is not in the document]
Shnoulle: OK :)

Shnoulle (approved):

Shnoulle:
Code review done

application/models/Survey.php
gabrieljenik: This was hard to test. The issue is not on the copy process, but on the survey creation one, where the default value is given to the datecreated field

application/models/Survey.php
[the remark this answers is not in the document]
gabrieljenik: > You give me the issue ;)

That issue was for the format pattern, which is run by the Yii framework. That's where the confusion lies.

application/models/Survey.php
Shnoulle: Yes, thank you for explanation

PR #2823, opened 2022-12-31
Fixed issue #18542: Plugins: PasswordRequirement plugin is not working properly

gabrieljenik:
ctype_lower doesn't work good when having numbers.


Shnoulle (approved):
:+1:

PR #2825, opened 2023-01-04
Fixed issue #18324: mandSoft POST parameter may be passed to bypass mandatory questions

Shnoulle:
Dev: move mandSoft control to _validateQuestion only


Shnoulle:
@olleharstedt i can assign @gabrielcavalari as reviewer ? 
I think its a big issue here (but i don't use mandatory soft option)

gabrieljenik (approved):
Code structure looks OK. Not sure I understand the solution (don'tneed to :) )
Haven't tested it,

Shnoulle:
> Not sure I understand the solution (don'tneed to :) ) 

:) since maybe you need to fix EM : 

- before : checking of `App()->request->getPost('mandSoft')` on a lot of function and if question are Y or S
- now : checking  `App()->request->getPost('mandSoft')` only in _validateQuestion function and only for mandatory == 'D'.

Fibnally : if a group have one Y and one S and  mandSoft is set :  Y questions is set invalid (and stay invalid) : group is invalid.
Update only Y question : need to set mandSoft transoarently to the user.

Shnoulle:
Added auto test

adamzammit:
This breaks the Mandatory soft function when used more than once over multiple pages/groups.

See here for the bug report: https://bugs.limesurvey.org/view.php?id=18808

PR #2827, opened 2023-01-04
Fixed issue #18578: Internal server error (HTTP 500) in RemoteControl get_participant_properties for removed participant

gabrieljenik:
Very related to #2465


Shnoulle (approved):
:+1:

Shnoulle:
Clearly needed and only way to fix it. Unsure need testing …

gabrieljenik:
> Clearly needed and only way to fix it. Unsure need testing …

Added automatic tests.
Hence, I think we can mark it as tested, right? 
What do you think @olleharstedt ?

PR #2839, opened 2023-01-07
Fixed issue #18243: ARIA Attribute role on Array question not correct.

Shnoulle:
Dev: Use role="presentation"  for table
Dev: remove all thead and th see https://www.w3.org/TR/WCAG20-TECHS/F92.html


application/config/packages.php
```diff
@@ -95,7 +95,8 @@
         ),
         'depends' => array(
             'limesurvey-public',
-            'embeddables'
+            'embeddables',
+            'bootstrap'
```
Shnoulle, line 99: Needed for this : https://github.com/LimeSurvey/LimeSurvey/pull/2839/files#diff-95fab772d475fc6bd9feb24f8a5e7a809747be3a57e932f0927e6dd7f5543fd9R277

Unsure if we need to move to theme.css ?

Shnoulle:
Oups …

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/2580

PR #2846, opened 2023-01-16
Fixed Issue #18581: Blocking users after X failed attempts counts incorrectly (off by 1)

gabrieljenik:


Shnoulle (approved):

olleharstedt:
Failing CI?

gabrieljenik:
It is failing on a test which is not even impacted here.
Maybe re merging master?

BTW, what is the difference in between the push test and the pull_request test ?

tests/functional/acceptance/15532-em-warnings/ExpressionWarningsOnLogicTest.php
```diff
@@ -53,7 +53,7 @@ public function testCheckInvalid()
         );
         try {
             self::$webDriver->get($url);
-            sleep(1);
+            sleep(5);
```
olleharstedt, line 56: ???

olleharstedt (changes requested):

tests/functional/acceptance/15532-em-warnings/ExpressionWarningsOnLogicTest.php
gabrieljenik: The test is failing. Making sure it wasn't a speed issue.

Shnoulle:
> Updated the FailedLoginAttempt test. This still fails at ls\tests\ExpressionWarningsOnLogic --> Check invalid Will be back

Sometimes fail are really strange …

gabrieljenik:
> > Updated the FailedLoginAttempt test. This still fails at ls\tests\ExpressionWarningsOnLogic --> Check invalid Will be back
> 
> Sometimes fail are really strange …

Will push a solution about it.

olleharstedt:
First step should probably be to try to run the test steps manually, in the browser.

gabrieljenik:
The issue is about some modal boxes that appear. We are working on it.
image
image

Maybe best is to set some test mode where we just don't show some stuff which can mess tests?
(modals, alerts)

Shnoulle:
> Maybe best is to set some test mode where we just don't show some stuff which can mess tests?

Why this appear only now ? This 2 alert shown before ?
SSL alert can be disable by config

Shnoulle:
😩

PR #2847, opened 2023-01-16
Fixed issue #18542: Plugins: PasswordRequirement plugin is not working properly

gabrieljenik:
Found bug while creating automated test on issue 18370

PR #2753, opened 2022-12-01
Fixed issue #18381: PluginSetting type date must be saved as a common datetime string

Shnoulle:
Dev: use saveformat option
Dev: Can be set by default to "Y-m-d H:i" ?
Dev: send array of settings, check if datetime and datetimesaveformat is set in array

PR #2848, opened 2023-01-16
Fixed Issue #18370: Password suggestions sometimes doens't match rules

gabrieljenik:
Fixed issue
Added a few methods on test base class for enhancing the testing of plugins.
Added tests for this issue (generate random) as well as for checkValidity (issue 18542)


application/libraries/PluginManager/PluginEvent.php
```diff
@@ -173,6 +173,20 @@ public function set($key, $value)
         return $this;
     }
 
+    /**
+     * Set mass key/value pair to be used by plugins hanlding this event.
+     *
+     * @param array $values
+     * @return \PluginEvent Fluent interface
+     */
+    public function setFromArray($values)
```
Shnoulle, line 182: Only used for test ?

gabrieljenik: Right now, yes. New cases shall appear as it is available.

Shnoulle: We already have API issue with only a single function, but you add more ?

Have such function disable all core fix see https://github.com/LimeSurvey/LimeSurvey/pull/2753 or https://github.com/LimeSurvey/LimeSurvey/pull/2743 …

Shnoulle: I don't say there are conflict , clearly not **currently**. But in some month/year : there are potentially more API broke.

Why not, but i don't really find why adding this function is interesting. 

Maybe i'm wrong ?

gabrieljenik: Don't agree. Still, already rolledback :)

application/helpers/common_helper.php
```diff
@@ -5041,3 +5041,27 @@ function standardDeviation(array $numbers): float
...
     return sqrt($variance / $numberOfElements);
 }
+
+/**
+ * Returns a random number
+ * @param int $max The number maximun of string
+ * @param int $min The number minimum of string
+ * @return int
+ */
+function safeRandom($max, $min = 0)
```
Shnoulle, line 5051: This 2 function are only used for this plugin, right ?

Adding it as private static function inside plugin seems better in my opinion.

Shnoulle (changes requested):
I think it's better to have function for plugin inside plugin only.

Because usage of limesurvey global function are NOT in API ([common_helper](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/common_helper.php), [frontend_helper](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/frontend_helper.php) etc …)

We already broke global function without api update.

If you really think it must be in API : add it in API : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/libraries/PluginManager/LimesurveyApi.php

My opinion : leave it in plugin only.

gabrieljenik:
> I think it's better to have function for plugin inside plugin only.

I placed them in commons because they can be helpfull for others.
At the very beggining new functions will always be used on limited placed.
Still, thinking forward they can be usefull.

If all is private, nothing is reused, difficult to set standards, reuse code, and not reinvent the wheel.

> Adding it as private static function inside plugin seems better in my opinion.

I think code reviews are for reviewing stuff lines up with a global way of coding.
Is not for approving line by line.
 I understand you don't agree, but I believe we can live with some disagreements.

gabrieljenik:
> We already broke global function without api update.

What does this mean?

application/helpers/common_helper.php
olleharstedt: I don't care that much about private vs common but why do we need safe random at all? Can't we just use mt_rand? Or PHP version issue?

application/helpers/common_helper.php
```diff
@@ -5041,3 +5041,27 @@ function standardDeviation(array $numbers): float
...
+{
+    if (function_exists('random_int')) {
+        return random_int($min, $max);
+    }
+    return mt_rand($min, $max);
+}
+
+/**
+ * Returns the character of a string
```
olleharstedt, line 5060: "the" --> "a random"?

tests/TestBaseClass.php
```diff
@@ -238,4 +256,22 @@ protected static function addUserPermissions(int $userId, array $permissions)
...
+     * Expected to be used for unit testing purposes on edge cases.
+     *
+     * @param object $object
+     * @param string $methodName
+     * @param array $parameters
+     * @return mixed
+     */
+    public function invokePrivateMethod(&$object, $methodName, array $parameters = array())
+    {
+        $reflection = new \ReflectionClass(get_class($object));
```
olleharstedt, line 271: Ah this is pretty nice. We've started to use another mocking framework too to get around PHPUnit limitations.

application/helpers/common_helper.php
```diff
@@ -5041,3 +5041,27 @@ function standardDeviation(array $numbers): float
 
     return sqrt($variance / $numberOfElements);
 }
+
+/**
+ * Returns a random number
+ * @param int $max The number maximun of string
```
olleharstedt, line 5047: "string" is probably wrong here

Shnoulle:
> > We already broke global function without api update.
> 
> What does this mean?

Updating function in common_helper didn't update API, then : if plugin used a common_helper function : maybe it can break during a minor update.

And it's already happen

Shnoulle:
> > I think it's better to have function for plugin inside plugin only.
> 
> I placed them in commons because they can be helpfull for others. At the very beggining new functions will always be used on limited placed. Still, thinking forward they can be usefull.
> 
> If all is private, nothing is reused, difficult to set standards, reuse code, and not reinvent the wheel.

I know that : but it's here plugin, it must work during little fix, and use a function not in API is not something to do.


> > Adding it as private static function inside plugin seems better in my opinion.
> 
> I think code reviews are for reviewing stuff lines up with a global way of coding. Is not for approving line by line. I understand you don't agree, but I believe we can live with some disagreements.

But ? Did you think the same when looking my Pull Request ?

gabrieljenik:
> But ? Did you think the same when looking my Pull Request ?

Yes, usually I try to.

I think is not common for me to ask for change requests.
I try to make them as an exceptional thing.

> If you think it can be interesting for other plugin : add it in API

I think it can be interesting not only for API, but for other stuff as well.

Shnoulle:
> I think it can be interesting not only for API, but for other stuff as well.

Why not in global + in API.

But still : i close conversation. I write clearly my opinion about usage of core function out of API.

Shnoulle (approved):
Still think usage of core function must be avoided by API.

Except : no issue in code.

gabrieljenik:
I am moving the code as static method. Issue solved.

Shnoulle:
I already approved before

Shnoulle (approved):

tests/TestBaseClass.php
Shnoulle: Just a confirmation : this can not be used via **URL** hacking to call a private method and create security issue ?

gabrieljenik:
> I already approved before

Yes, seems we didi it at almostthe same time

tests/TestBaseClass.php
gabrieljenik: I can change it to protected probably.

Shnoulle: I really think it can not happen, only controller can be call directly. But why not protected, think it must be used only in test.

If needed elsewhere : move the function to public is better.

:+1:

application/helpers/common_helper.php
gabrieljenik: Updated

tests/TestBaseClass.php
```diff
@@ -256,22 +256,4 @@ protected static function addUserPermissions(int $userId, array $permissions)
...
-    /**
-     * Invokes a private method from an object
-     * Expected to be used for unit testing purposes on edge cases.
-     *
-     * @param object $object
-     * @param string $methodName
-     * @param array $parameters
-     * @return mixed
-     */
-    public function invokePrivateMethod(&$object, $methodName, array $parameters = array())
```
Shnoulle: Removed ?

tests/TestBaseClass.php
gabrieljenik: At last, I removed it. I think I initially used it and then just used the events.
When needed can be restablished

tests/TestBaseClass.php
gabrieljenik: https://github.com/LimeSurvey/LimeSurvey/pull/2848#discussion_r1073555751

application/helpers/common_helper.php
gabrieljenik: Updated

olleharstedt:
Password unit test fails, it seems?

gabrieljenik:
> Password unit test fails, it seems?

he he...
Thats because this test relies on a fix pushed on this PR:
https://github.com/LimeSurvey/LimeSurvey/pull/2847#top

Thing is I discovered the issue #18542 while doing the password auto tests.
It seemed to me it was OK to do it as a separate issue.

olleharstedt:
Still fails?

gabrieljenik:
> Still fails?

It was an old run. Need to re run it.

gabrieljenik:
> tter to have function for plugin inside pl

Now Pull is green :)

olleharstedt:
> > tter to have function for plugin inside pl
> 
> Now Pull is green :)

You sure?

gabrieljenik:
Thats because this test relies on a fix pushed on this PR:
https://github.com/LimeSurvey/LimeSurvey/pull/2847

Push is failing. Pull is good. Should I merge master as to have all green?

gabrieljenik:
Now is green green

PR #2849, opened 2023-01-16
Fixed issue #18592: When creating a survey Default group may not be the ones preselected

gabrieljenik:


Shnoulle (approved):
:+1:

Shnoulle:
We must review the way gsid is set when create survey , it's not in model default … but before https://github.com/LimeSurvey/LimeSurvey/blob/f7874d79e812c59dbee03b8a29565e207341beac/application/models/Survey.php#L185

PR #2852, opened 2023-01-19
install nestedSortable package directly from github

jack-han-115:
Dev:


Shnoulle:
We need to move to github to have version 2 ?

PR #2858, opened 2023-01-23
Fixed issue #18594: Slider: Min and max values not checked if nothing selected

gabrieljenik:
When initial position was set to zero from question settings, it was ignored as considered false.
https://github.com/LimeSurvey/LimeSurvey/commit/0b638791ec885d464d5e49300ef53012a52ba1f5#diff-225f3da7558dd8688295831028645da8573ba81e87e7c76b4f802d8e6ce3c721L21

So adding a new block for this specific situation.


Shnoulle (approved):
:+1: 

<s>Why it's related to this issue ? </s>
Value range allow missing is set 
The n even without any default value : user can not move next

Got it : value range allow missing is ON

gabrieljenik:
I know it is approbed. But as to understand the comment.
What do you mean by the following?

> Value range allow missing is set

Shnoulle:
With "Allow range allow missing"  to Yes : you don't need to set a value :)

«Is no answer (missing) allowed when either 'Equals sum value' or 'Minimum sum value' are set?»

Here : you can move next 

@jmantysalo say «If there is a slider array with "The sum must equal" -setting and default value zero, then the user can continue without setting any value.»

It's because allow range allow missing is Yes here.  I think jmantysalo must use OFF

gabrieljenik:
> It's because allow range allow missing is Yes here. I think jmantysalo must use OFF because it's not related only to settings 0. Issue is "user can continue without setting any value", but tgis part is the needed behaviour.

OK, so you say, if I understand correctly ...
- it was an error on the question configuration.
- but still, this fix is OK to push.

Right?

Shnoulle:
> OK, so you say, if I understand correctly ...
> 
>     * it was an error on the question configuration.
>     * but still, this fix is OK to push.
> 
> 
> Right?

yes, this fix is **needed** but the option of Question can be improved by user

gabrieljenik:
Quick question: If 
- "allow value missing" is ON.
- "equals sum value" = 100

Should the sum validation be enforced?

Shnoulle:
> Should the sum validation be enforced?

I think need all value missing (set to ""). Unsure on this point … 

https://manual.limesurvey.org/Question_type_-_Multiple_numerical_input/en#Value_range_allows_missing_.28value_range_allows_missing.29

gabrieljenik:
> > Should the sum validation be enforced?
> 
> I think need all value missing (set to ""). Unsure on this point …
> 
> https://manual.limesurvey.org/Question_type_-_Multiple_numerical_input/en#Value_range_allows_missing_.28value_range_allows_missing.29

I think equls_sum should be included in that doc as well.
If value missing is ON, no other validations should be enforced.
equlas_sum would be enforced then, if user actually enters a value.

Still, if there is initial_position, value is not missing anymore and then the equls_sum validation should be enforced.
We are good in that point.

In regard to the value_missing vs equal_sum, I will review it and create a ticket then if needed.

Shnoulle:
> Still, if there is initial_position, value is not missing anymore and then the equls_sum validation should be enforced.

If you check «Slider initial value set at start»
Capture d’écran du 2023-01-25 15-53-35


(and remind the default value too … 🤪 )

jmantysalo:
Not sure if I understood all. Hence, here is what the user asked about me:

A question asks something like "how many hours in a normal workday you use for these task", and have quite long list. Most participants have only few tasks they do. It is slightly irritating to click many times "I don't do this, so zero", "Neither this, again zero", "Nope, zero"... Hence the default value zero.

I think that the question is hard in any case -- but can't figure better way to suggest for the user. Also my work is more to implement what is asked than tell what to do.

Shnoulle:
> I think that the question is hard in any case -- but can't figure better way to suggest for the user. Also my work is more to implement what is asked than tell what to do.

You need to know Value range allow missing role …

gabrieljenik:
> > When taking the survey, if the question is not answered, the sum validation is not enforced. I understand that's OK.
> > Maybe we should update doc?
> 
> You mean https://manual.limesurvey.org/Question_type_-_Multiple_numerical_input/en#Value_range_allows_missing_.28value_range_allows_missing.29 ? It's clearly the **purpose** if this settings. Ad default is OFF, no ?
> 
> Or adding «(If value range allow missing is set : value range is not tested if user didn't answer on all sub question)» at https://manual.limesurvey.org/Question_type_-_Multiple_numerical_input/en#Equals_sum_value_.28equals_num_value.29

Disregard my comment. No amendment needed.

Shnoulle:
> Disregard my comment. No amendment needed.

Maybe we can add a sentence about value range allow missing on Equals sum value and Minimum sum value .

My English is really to dirty to have a simple sentence …

gabrieljenik:
> > Disregard my comment. No amendment needed.
> 
> Maybe we can add a sentence about value range allow missing on Equals sum value and Minimum sum value .
> 
> My English is really to dirty to have a simple sentence …

It is there, but the attribute names are the technical ones, not the human ones.

>  If enabled, a survey respondent can skip answering the multiple numerical question even if equals_num_value or min_num_value are set.

Shnoulle:
> It is there, but the attribute names are the technical ones, not the human ones.

Yes but i mean updating [QS:equals_num_value](https://manual.limesurvey.org/QS:Equals_num_value) and [QS:min_num_value](https://manual.limesurvey.org/QS:Min_num_value)

I put something, if you can review …

Shnoulle:
> I put something, if you can review …

https://manual.limesurvey.org/index.php?title=QS%3AMin_num_value&type=revision&diff=175646&oldid=149116

gabrieljenik:
> > I put something, if you can review …
> 
> https://manual.limesurvey.org/index.php?title=QS%3AMin_num_value&type=revision&diff=175646&oldid=149116
> 
> https://manual.limesurvey.org/index.php?title=QS%3AEquals_num_value&type=revision&diff=175647&oldid=149114

I would rather say: 
> If the setting ''Value range allows missing'' is set to ON, the question may be left unanswered, then the validation will not be enforced.

PR #2863, opened 2023-01-23
Fixed issue #18579: Question theme are not shown in Survey Logic file

gabrieljenik:


application/helpers/expressions/em_manager_helper.php
```diff
@@ -9485,7 +9492,7 @@ public static function ShowSurveyLogicFile($sid, $gid = null, $qid = null, $LEMd
                     . "onclick='window.open(\"$editlink\",\"_blank\")'>"
                     . $rootVarName . "</span>";
             }
-            $questionRow .= "</b><br />[<a target='_blank' href='$editlink'>QID $qid</a>]<br/>$typedesc [$type] $errText $sWarningsText</td>"
+            $questionRow .= "</b><br />[<a target='_blank' href='$editlink'>QID $qid</a>]<br/>$typedesc [$type] $themeDesc $errText $sWarningsText</td>"
```
Shnoulle, line 9498: Maybe time to use some `sprintf(QID $s)` and wrap each part on span or div … 

<div class="question-link">[<a target='_blank' href='$editlink'>sprintf(gT("QID $s",$qid</a>]</div><div class="type-description">$typedesc</div> etc …

gabrieljenik:
> The </b><br />[<a target='_blank' href='$editlink'>QID $qid</a>]<br/>$typedesc [$type] $themeDesc $errText $sWarningsText is really awfull …

I think that's out of the scope of the fix.
We should wait a bit on BS5 maybe?

Shnoulle:
> I think that's out of the scope of the fix. We should wait a bit on BS5 maybe?

The fix add `$themeDesc` without wrapper. Have a wrapper here seems needed in my opinion.

gabrieljenik:
> The fix add $themeDesc without wrapper. Have a wrapper here seems needed in my opinion.

All other items in that line don't have a wrapper. Would this be the only one which have a wrapper?
That's not the minimal needed fix.

I am not saying is the greatest. I am just keeping consistency and minimal fix needed.

> Have a wrapper here seems needed in my opinion.

I agree. But, it is part of this ticket to do it?

I will do it anyway. Seems the quicker solution.

Shnoulle (approved):
Thank you

olleharstedt:
Thanks!

PR #2866, opened 2023-01-24
Fixed issue #18582: RemoteControl invite_participants unexpectedly stops after first failure

gabrieljenik:


Shnoulle (approved):
This fix the issue

gabrieljenik:
Marking it as tested as it has the automatic test

PR #2867, opened 2023-01-24
Fixed issue #18596: Reintroduce lost unit tests from merge

gabrieljenik:


gabrieljenik:
Tests seem were failing. Strange is they are failing on places not updated by this PR.
The parent commit was green. Not sure what is happening.

olleharstedt:
Hmmm rerunning jobs to check.

olleharstedt:
Still failing.

olleharstedt:
$_SESSION problem maybe? Or database fixture not cleared?

gabrieljenik:
Will review.

olleharstedt:
For one test I had to do $_SESSION = [] to clear session from previous test.

gabrieljenik:
seems `surveyDynamic()::model($surveyId)-->getMaxId()` had some prior info and it was not getting refreshed.
We added a manual refresh.

Maybe we should:
- create clearMaxId() function (or an extra param)
- overload activeRecord->refreshMetadata() as to also refresh those
?

Also, didn't like the fact `static $maxId` is declared locally.
We can't reset it without recalculating.

olleharstedt:
Well that was a new one :) Yeah, it's a problem for tests to cleanup everything when done, if the model classes are stateful like that. :stuck_out_tongue:

gabrieljenik:
> seems `surveyDynamic()::model($surveyId)-->getMaxId()` had some prior info and it was not getting refreshed. We added a manual refresh.
> 
> Maybe we should:
> 
> * create clearMaxId() function (or an extra param)
> * overload activeRecord->refreshMetadata() as to also refresh those
>   ?
> 
> Also, didn't like the fact `static $maxId` is declared locally. We can't reset it without recalculating.

Does it worth putting this into a story?
Can this kind of situation happen in real life?

GetMaxId() should probably be refreshed on other AR operations, like Save(), Add(), Delete(), ...

PR #2878, opened 2023-01-30
Dev #18453: Add PHP 8.1 compatibility

gabrieljenik:


olleharstedt:
As we talked about, can you change PHP 8.0 to 8.1 in the Github Actions file?

gabrieljenik:
> @gabrieljenik Can you please update the patch as high priority so we can merge it now? Thank you!

Done. No conflicts

gabrieljenik:
> As we talked about, can you change PHP 8.0 to 8.1 in the Github Actions file too?

Done. Still, some stuff failed. Do they sound familiar?

application/controllers/admin/CheckIntegrity.php
```diff
@@ -1074,7 +1074,7 @@ protected function checkintegrity()
         $aFullOldTokenSIDs = array();
 
         foreach ($aTables as $sTable) {
-            list($sOldText, $SurveyText, $iSurveyID, $sDateTime) = explode('_', substr($sTable, strlen($sDBPrefix)));
+            list($sOldText, $SurveyText, $iSurveyID, $sDateTime) = explode('_', substr((string) $sTable, strlen((string) $sDBPrefix)));
```
olleharstedt, line 1077: Can $sTable be null? What about using `??` to give it a default value when null?

olleharstedt: Question is, do you want the empty string default in all places?

application/controllers/FailedEmailController.php
```diff
@@ -96,7 +96,7 @@ public function actionResend()
         $deleteAfterResend = App()->request->getParam('deleteAfterResend');
         $preserveResend = is_null($deleteAfterResend);
         $item = [App()->request->getParam('item')];
-        $items = json_decode(App()->request->getParam('sItems'));
+        $items = json_decode((string) App()->request->getParam('sItems'));
```
olleharstedt, line 99: A little bit sloppy? getParam() already has a second argument to give default value at null.

olleharstedt (changes requested):

application/controllers/admin/CheckIntegrity.php
gabrieljenik: Well, that was the suggested fix by rector.
https://github.com/rectorphp/rector

olleharstedt: Gah

application/controllers/ResponsesController.php
```diff
@@ -799,7 +798,7 @@ public function actionDeleteAttachments(int $surveyId, int $responseId = null):
             throw new CHttpException(405, gT("Invalid action"));
         }
 
-        $stringItems = json_decode($request->getPost('sItems'));
+        $stringItems = json_decode((string) $request->getPost('sItems'));
```
olleharstedt, line 801: Another place where you should use Yii functionality to assign default value instead of (string).

olleharstedt (changes requested):

application/controllers/UserManagementController.php
```diff
@@ -924,7 +924,7 @@ public function actionBatchSendAndResetLoginData()
             );
         }
 
-        $aUsers = json_decode(App()->request->getPost('sItems', "[]"));
+        $aUsers = json_decode((string) App()->request->getPost('sItems', "[]"));
```
olleharstedt, line 927: Not needed, since it already defaults to [].

application/core/LSGettextMoFile.php
```diff
@@ -0,0 +1,28 @@
...
+* to the GNU General Public License, and as distributed it includes or
+* is derivative of works licensed under the GNU General Public License or
+* other free or open source software licenses.
+* See COPYRIGHT.php for copyright notices and details.
+*/
+
+/**
+ * Extends CGettextMoFile to make it compatible with PHP 8.1
+ */
+class LSGettextMoFile extends CGettextMoFile
```
olleharstedt, line 17: Hm this wasn't fixed in Yii 1.1 update?

application/core/LSYii_FilterValidator.php
```diff
@@ -0,0 +1,41 @@
...
+ * other free or open source software licenses.
+ * See COPYRIGHT.php for copyright notices and details.
+ *
+ */
+
+/**
+ * Extends CFilterValidator class to add PHP 8.1 compatibility. Based on Yii 2's FilterValidator
+ * @inheritdoc
+ */
+class LSYii_FilterValidator extends CFilterValidator
```
olleharstedt, line 23: Same question here, it's not in any Yii 1.1 patch? Did you submit a patch to Yii 1.1?

gabrieljenik: If I am not wrong, I think we did this before Yii had any support for PHP8.1.
I could be wrong though.

Any way, We should check, agree

application/core/LSGettextMoFile.php
gabrieljenik: If I am not wrong, I think we did this before Yii had any support for PHP8.1.
I could be wrong though.

Any way, We should check, agree

tassoman:
Any chance to have this backported into 5.x branch?

PR #2880, opened 2023-01-31
18337: Placeholders VALIDFROM and VALIDUNTIL at email templates do not use localized dates

gabrieljenik:
Tests for localize_date were added

### DIFF
--- a/tests/unit/plugins/LocalizeDateTest.php
+++ b/tests/unit/plugins/LocalizeDateTest.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace ls\tests;
+
+/**
+ * Tests for the localize_date function.
+ */
+
+class LocalizeDateTest extends TestBaseClass
+{
+    public static function setupBeforeClass(): void
+    {
+        parent::setupBeforeClass();
+        \Yii::import('application.core.plugins.dateFunctions.EMFunctions', true);
+        self::importSurvey(\Yii::app()->getConfig('rootdir') . '/tests/unit/plugins/surveys/limesurvey_survey_854771.lss');
+    }
+
+    /**
+     * Testing for different date strings.
+     */
+    public function testDateStrings(): void
+    {
+        //Testing for an empty date string
+        $date = \dateFunctions\EMFunctions::localize_date('');
+        $this->assertEmpty($date, 'An empty string was passed as the date parameter');
+
+        //Testing for a date string in an incorrect format.
+        //Trying to localize January 30th 2023.
+        $date = \dateFunctions\EMFunctions::localize_date('30-01-2023');
+        $this->assertNotEquals($date, '01-30-2023', 'A date string with a wrong format was passed (30-01-2023)');
+
+        //Testing for day only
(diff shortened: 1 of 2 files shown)


tests/unit/plugins/LocalizeDateTest.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Tests for the localize_date function.
+ */
+
+class LocalizeDateTest extends TestBaseClass
+{
+    public static function setupBeforeClass(): void
+    {
+        parent::setupBeforeClass();
+        \Yii::import('application.core.plugins.dateFunctions.EMFunctions', true);
```
Shnoulle, line 14: Import ? Not activate https://github.com/LimeSurvey/LimeSurvey/blob/4e5e6d2f3e2c17ce4bfdae9d86e4c9ef5eacc72a/tests/TestBaseClass.php#L147 ?

tests/unit/plugins/LocalizeDateTest.php
```diff
@@ -0,0 +1,63 @@
...
+        self::importSurvey(\Yii::app()->getConfig('rootdir') . '/tests/unit/plugins/surveys/limesurvey_survey_854771.lss');
```
Shnoulle, line 15: Uwsage of `self::$surveysFolder`

Shnoulle (commented):
Seems you don't use lss file, if not : `\Yii::import('application.core.plugins.dateFunctions.EMFunctions', true);` is great just remove the lss file and import.

If yes : use included function and helper for activate plugin and import survey

tests/unit/plugins/LocalizeDateTest.php
Shnoulle: Keep it but remove lss and importSurvey seems a better option.

gabrieljenik: > Import ? Not activate

That's for calling the function inside the plugin directly.
Not related with LSS.

gabrieljenik:
> Seems you don't use lss file

I do use it.

> If yes : use included function and helper for activate plugin and import survey

I import LSS as needed to run the test
Call plugin function directly.

What's wrong with current way?

tests/unit/plugins/LocalizeDateTest.php
Shnoulle: Then : remove the lss file and the import no ?

Shnoulle:
> I import LSS as needed to run the test

I don't understand : you don't use lss file here. Best is to be clear : this test don't do WebDriver. The lss file is not needed.

Shnoulle:
> > Seems you don't use lss file
> 
> I do use it.
> 

Where ? I really think you don't use it. Maybe i made an error , but looking at code : you don't use it.

Shnoulle (approved):

gabrieljenik:
@olleharstedt Being this an automated test, should I markit as tested now that is green?

olleharstedt:
> @olleharstedt Being this an automated test, should I markit as tested now that is green?

Hmmm maybe we should add a new tag for automated test added?

PR #2881, opened 2023-01-31
Fixed issue #18587: Question type button not visible after having chosen the type

gabrieljenik:


Shnoulle (approved):
OK, but really time to remove this awful files … ;)

PR #2882, opened 2023-01-31
Fixed issue #17486: Current Page Values not saved if saving pressing Enter

gabrieljenik:
Adding the same button behaviour as for the submit button on the survey taking process.


Shnoulle (commented):
I really think we must not fix such issue with javascriptb only ? it mits work without JS.

It's due to the default button ? The move value is default ?

Shnoulle:
And more : save.twig is a lot replaced by template manager

gabrieljenik:
Closing this in favor of the #2889 solution

PR #2883, opened 2023-01-31
change template images with new design

jack-han-115:
Dev:


application/views/admin/survey/subview/accordion/_generaloptions_panel.php
```diff
@@ -316,7 +316,7 @@ function checkSelect2Languages(mylangs) {
         </div>
         <!-- Template -->
 
-        <div class="mb-3">
+        <div class="mb-3" style="height: 300px;">
```
Shnoulle, line 319: No way to have auto height ? I use different size for skelvanilla in 3 and 5 : it work good. 

I know i must redo whole skelvanilla :)

Capture d’écran du 2023-02-02 12-03-57

PR #2888, opened 2023-02-04

Shnoulle:
Dev: Move role=group to a div wrapper of the array with aria-labelledby
Dev: Put the role="columnheader" on the each header 
Dev:Give an id to each header ( answer-{SGQA}-{code})
Dev: Each line stay role="radiogroup"/"group" with good  aria-labelledby
Dev:Put aria-hidden to label used for labelling and shown on phone (muts be hidden since table is tabke (else double read)
Dev:Set aria-labelledby to each header for each input (aria-labelledby="answer-{SGQA}-{code}")

- [x] array
- [x] array / dropdown
- [x] 5 points
- [x] 10 points
- [x] column
- [x] dualscale
- [x] increasesamedecrease
- [x] multiflexi
- [x] texts
- [x] yesnouncertain


Shnoulle:
@ptelu 

1. Move role=group to a div wrapper of the array with aria-labelledby question->question
2. Put the role="columnheader" on the each header (answer text)
3. Give an id to each header ( answer-{SGQA}-{code})
4. Each line stay role="radiogroup" with aria-labelledby subquestion->question (canbe reol group for other array)
5. Put aria-hidden to label used for labelling and shown on phone
6. Set aria-labelledby to each header for each input (aria-labelledby="answer-{SGQA}-{code}")


Lot of work to do it on all other array:  **do I continue** or this was never merged ?
I add some data like oAnswer to twig views, not needed but can be interesting

Shnoulle:
- [ ] array
- [ ] array / dropdown
- [ ] 5 points
- [ ] 10 points
- [ ] column
- [ ] dualscale
- [ ] increasesamedecrease
- [ ] multiflexi
- [ ] texts
- [ ] yesnouncertain

Shnoulle:
@tpartner for 6.X version : i think a lot of workaround stay compatible with this solution.

Shnoulle:
About the bugs : 

1. lack of th : it's because th are in sRows or sHeaders
2. lack of caption : true, but since it's a table : we need caption about roles of colunheader and rowheader : sometghing like that for Array : _A table with each line a radio list, the first cell are the related question part, the answer option are the row header_ **but** with all options : it can easily break : for example with hidden answers (1st column width : 0)

Shnoulle:
Sample with skelvanilla on 3.X : https://www.shnoulle.net/test/arrayWCAG/tableau_role-table-2.html 
(hacked manually)

Vanilla have a lot of another a11y issue …

Shnoulle:
Fixing conflict …

@ptelu you can check ?

ptelu:
> Fixing conflict …
> 
> @ptelu you can check ?

checking it out now

ptelu:
looks fine to me, i merged once the dev into the pr because there was a bug introduced that got fixed a day later.
The only thing that confuses me is sometimes the meaning behind variables especially in the surveytaking controllers.

maybe we can make the variables a bit more clear:
basename = questioncode(databasecolumnname) without subquestion?
code = answeroption id?

But it can be done later if want it to merged faster, just some nagging from me.

Shnoulle:
I think use basename because it was used on other twig part 

code: is answer->title , show as code in admin GUI
Capture d’écran du 2023-03-31 16-11-53

But : yes : we need better system : currently var can be different for each twig files … 
related issues : https://bugs.limesurvey.org/view.php?id=18691 and https://bugs.limesurvey.org/view.php?id=14228

ptelu:
Alright yeah makes sense was just looking at an example where thats not the case.
image

Shnoulle:
> Alright yeah makes sense was just looking at an example where thats not the case. image

This one didn't have code , because it's fixed answer. 

But it's the answer «code» saved in database 

You have same for Y/N/U array : Y is saved in DB when you export response as code :)
Capture d’écran du 2023-03-31 16-24-04

In 5.X : you have ld, xc, etc … code seems more clear ? No ?

ptelu:
yeah all good :)

Shnoulle:
But right : maybe the best was

answer[
'code'=> 1,
"label" => 1, // or shown ? Like ExpressionManager ?
"value" => null, // in case assesment
]

PR #2889, opened 2023-02-06
Fixed issue #17486: Current Page Values not saved if saving pressing Enter

gabrieljenik:
Replacement for #2882 
This is a no JS solution. Better.


Shnoulle:
Really great ! And twig updated compatible !!! 


Maybe we can add saveall with value saveall ? Then the action done was save ?

Shnoulle (approved):

PR #2893, opened 2023-02-06
Fixed Issue #18438: Fixed Show popups, wrong text in inherit-button

gabrieljenik:


application/views/themeOptions/options_core.php
```diff
@@ -121,8 +121,13 @@
                                 $optionsValues = !empty($attribute['options']) ? explode('|', $attribute['options']) : array();
                                 $optionLabels = !empty($attribute['optionlabels']) ? explode('|', $attribute['optionlabels']) : array();
                                 $options = array_combine($optionsValues, $optionLabels);
-                                if ($bInherit && isset($sParentOption)){
-                                    $options['inherit'] = gT("Inherit") . ' [' . $sParentOption . ']';
+                                if ($bInherit && isset($sParentOption)) {
+                                    if(is_numeric($sParentOption) && array_key_exists($sParentOption, $options)) {
```
Shnoulle, line 125: Why is_numeric ?

Shnoulle (approved):
No code issue, but why only to is_numeric ? 'options' => `['Y' => 'Yes']` for example

application/views/themeOptions/options_core.php
gabrieljenik: That;s how we expect the `sParentOption` to come like.
If it comes like that, we map it. If not, we show it as it is coming.

gabrieljenik:
> No code issue, but why only to is_numeric ? 'options' => `['Y' => 'Yes']` for example

For this particular attribute? I believe they are being stored like numbers here.

application/views/themeOptions/options_core.php
Shnoulle: For Show popusp : OK https://github.com/LimeSurvey/LimeSurvey/blob/096476f68a8edf1f2eea9bce4ba5817ef6b09238/themes/survey/vanilla/config.xml#L63

But for all other : https://github.com/LimeSurvey/LimeSurvey/blob/096476f68a8edf1f2eea9bce4ba5817ef6b09238/themes/survey/vanilla/config.xml#L65 

We need to translate too …

Shnoulle:
> > No code issue, but why only to is_numeric ? 'options' => `['Y' => 'Yes']` for example
> 
> For this particular attribute? I believe they are being stored like numbers here.

Yes, this one , but you really don't fix on/off or Y/N etc … ?

Shnoulle:
I don't know what to say when it's totally obvious by reading the code...
Capture d’écran du 2023-02-10 09-06-02

It must be **Hérite (Non)**

gabrieljenik:
> I don't know what to say when it's totally obvious by reading the code...

Haven't seen that... Maybe it is not that obvious to everyone
Will do another ticket for that

gabrieljenik:
Created
18633: Theme Options: Inherited value doesn't match any attribute option
https://bugs.limesurvey.org/view.php?id=18633

Shnoulle:
> Created 18633: Theme Options: Inherited value doesn't match any attribute option https://bugs.limesurvey.org/view.php?id=18633

In my opinion it's clearly a duplicate …

gabrieljenik:
> > Created 18633: Theme Options: Inherited value doesn't match any attribute option https://bugs.limesurvey.org/view.php?id=18633
> 
> In my opinion it's clearly a duplicate …

Let it be , please? .... :)

application/views/themeOptions/options_core.php
olleharstedt: Another reason to not use hungarian notation, lol. `$sParentOption`, s = string, but is_numeric check? ^^

application/views/themeOptions/options_core.php
```diff
@@ -121,8 +121,13 @@
...
+                                        $sParentLabelOption = $options[$sParentOption];
+                                        $options['inherit'] = gT("Inherit") . ' [' . gT($sParentLabelOption) . ']';
+                                    } else {
+                                        $options['inherit'] = gT("Inherit") . ' [' . gT($sParentOption) . ']';
```
olleharstedt, line 129: Bug here if sParentOption is empty, it seems.

olleharstedt: Tim fixed it in latest release.

olleharstedt: How was this tested, @gabrieljenik ???

gabrieljenik: I am noticing it. I guess that the tested case was the bug one. Sorry

gabrieljenik: On the else part of the if statement, we were supposed to leave the original statement:
`$options['inherit'] = gT("Inherit") . ' [' . $sParentOption . ']';`

PR #2900, opened 2023-02-08
Fixed issue #18527: Can't validate phone number usign Regex if phone number starts with +

gabrieljenik:


gabrieljenik:
Should this be sent to DEV @olleharstedt ?
It is a bug fix, but updating a library.

Shnoulle (approved):
Updating a lobrary is always a good idea.

BUT : muts be tested a lot ;)

olleharstedt:
I think it depends on the version change of that library, if it's a bug fix update or breaking update (assuming they use semver).

If I read the code right, it's an update from decimal.js 6 to 10? So yeah, probably better in dev branch then.

```
   *  decimal.js v6.0.0
  /*
   *  decimal.js v10.4.3
   *
```

olleharstedt:
Also I'm not sure we have any automatic tests that include the behaviour of decimal.js.

gabrieljenik:
> Also I'm not sure we have any automatic tests that include the behaviour of decimal.js.

Yes, not the best situation. Let me know how you want to handle it.

Thanks

olleharstedt:
@Shnoulle Any idea which question types depend on the decimal lib?

Shnoulle:
> @Shnoulle Any idea which question types depend on the decimal lib?

Unsure (i don't like decimal.js) : numeric + multiple numeric + short text with setting "As number"

Maybe check array number ?

olleharstedt:
Thoughts and prayers.

PR #2905, opened 2023-02-09
Fixed Issue #18620: Column relevance disabling not working as expected

gabrieljenik:


Shnoulle (approved):
seems OK, simple

application/views/questionAdministration/subquestionRow.twig
```diff
@@ -129,21 +129,18 @@
         </div>
     </td>
 
-    {% set relevanceTooltip = supportsRelevance ? gT("Click to expand") : gT("Relevance is not currently supported for columns") %}
```
olleharstedt, line 132: Why remove this?

gabrieljenik: Because now the column is being hidden if relevance not supported.
If the column is gone, the tooltip is not needed

gabrieljenik: > I don't think it's the reason

What is the reason then? 

https://github.com/LimeSurvey/LimeSurvey/pull/2810/files

Shnoulle: > > I don't think it's the reason
> 
> What do you think is the reason then?
> 

I write my answer before seeing yours : I don't think the reason was mine «it's awfull and uneeded … it give no information»

:)

PR #2909, opened 2023-02-10
Fixed issue #18634: Export single selected response can export a empty line

Shnoulle:
Dev: id set in url, resonseId in POST (by session)
Dev: remove session usage
Dev: can move to window-location-href, but ready for POST


Shnoulle:
Surely same issue in 5.X

gabrieljenik (approved):
Code Looks good. Haven't tested it.
Do we need same for master?

Shnoulle:
> Do we need same for master?

Yes, reported separately : https://bugs.limesurvey.org/view.php?id=18635 , but wit for the OK here to do same fix.

PR #2915, opened 2023-02-13
Fixed issue #18639: Token browse can be unusable with lot of data in attribute

Shnoulle:
Dev: wrap value inside in cell (and encode)
Dev: PHP part, need a micro CSS art after


application/models/TokenDynamic.php
```diff
@@ -796,6 +797,19 @@ public function getAttributesForGrid()
         return array_merge($this->standardColsForGrid, $aCustomAttributesCols);
     }
 
+    public static function formatAttribute($data, $name)
```
gabrieljenik, line 800: Doc?

Shnoulle: > Maybe adding it in the LSActiveRecord as a non-static function?

:+1: for LSActiveRecord, unsure about non-static , I

>  At last, the length could be a paameters as well ?

Don't know, since it's more related to display, css part fix the length show too. But I can add a config for that.

Shnoulle: About name default is https://www.yiiframework.com/doc/api/1.1/CFormatter#formatText-detail then : formatLongText seems OK.

Shnoulle: Move to formatter https://www.yiiframework.com/doc/api/1.1/CGridView#formatter-detail

Shnoulle: >  At last, the length could be a paameters as well ?

Can be set in config, but don't found the way to set formatter params easily in [GridView](https://www.yiiframework.com/doc/api/1.1/CGridView) ?

application/models/TokenDynamic.php
```diff
@@ -787,12 +787,11 @@ public function getAttributesForGrid()
             $aCustomAttributesCols[] = array(
                 'header' => $desc, // $aAttributedescriptions->$sColName->description,
                 'name' => $sColName,
-                'value'=>'$data->'.$sColName,
+                'type' => 'longtext',
```
gabrieljenik, line 790: Nice!

gabrieljenik (approved):
I like the way long text formatting came out :)

Shnoulle:
> I like the way long text formatting came out :)

Was used before … 

But i don't find a way to set longTextMaxLength different for each attribute (for example) …

tiborpacalat:
Tested, looks good.
Screenshot 2023-04-27 at 15 06 44

PR #2920, opened 2023-02-15
Update getSurveyList() from common_helper.php to lazy load survey lan…

kevin-foster-uk:
Update getSurveyList() from common_helper.php to lazy load survey language settings

When the site has many surveys this method was consuming allot of memory in loading all of the survey language settings data into memory. This was found after a complaint from phsinterface.limequery.com.

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -181,7 +181,6 @@ function getSurveyList($bReturnArray = false)
     if (is_null($cached)) {
         $surveyidresult = Survey::model()
             ->permission(Yii::app()->user->getId())
-            ->with('languagesettings')
             ->findAll();
         foreach ($surveyidresult as $result) {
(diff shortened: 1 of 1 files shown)


Shnoulle:
On a plugin : i use `survey->LocalizedTitle` i need to compare time with current method (3 seconds for the survey list when copying …)

kevin-foster-uk:
@c-schmitz Just going through my open PR. Do we still want this back port, shall we close the PR? or leave it on hold?

Shnoulle:
Must check with restriction on size too : https://github.com/LimeSurvey/LimeSurvey/commit/da574452e23c0f1ee37c8e3175d82b69801a9d3e

Difference on speed between both optimization …

The master version get all information in one request, but only needed information

c-schmitz:
PR is obsolete

PR #2921, opened 2023-02-15
Fixed issue #18407: On-page conditions not working when pre-filling q…

Shnoulle:
Fixed issue #18407: On-page conditions not working when pre-filling question with equation on the same page
Dev: Check existence of variable before trying to set
Dev: use jquery to be sure to don't throw error.

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -1250,12 +1250,13 @@ class ExpressionManager
         $tokens = $this->RDP_tokens;
         /* @var string|null used for ASSIGN expression */
         $idToSet = null;
-        $debug = false;
+        /* @var string[] the final expression line by line (to be join at end) */
         $stringParts = array();
         $numTokens = count($tokens);
 
-        /* Static function management */
+        /* @var integer bracket count for static function management */
         $bracket = 0;
+        /* @var string static string to be parsed bedfore send to JS */
         $staticStringToParse = "";
         for ($i = 0; $i < $numTokens; ++$i) {
             $token = $tokens[$i]; // When do these need to be quoted?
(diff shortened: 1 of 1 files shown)


Shnoulle:
It's near the same fix than 3.X , this don't update equation without ASSIGN (`=`)

Shnoulle:
About code smell : «Remove this empty statement.» i can't … default adding something … i can set 

```
                            case '=':
                                $stringParts[] = ' '; /* ASSIGN : usage jquery: don't add anything (disable default) */;
                                break;
                            default:
                                $stringParts[] = ' ' . $token[0] . ' ';
```
Bu it think it's better with nothing

About lines … yes … but how ?

application/helpers/expressions/em_core_helper.php
```diff
@@ -1248,11 +1248,15 @@ public function GetJavaScriptEquivalentOfExpression()
             return '';
         }
         $tokens = $this->RDP_tokens;
+        /* @var string|null used for ASSIGN expression */
```
gabrieljenik, line 1251: Nice doc :)

gabrieljenik (approved):
Code structure looks OK. 
Haven't tested it. 
Not following to much the solution as not too much into EM. 
Still, code writting, looks OK.

As always with EM, hard to test there are no regressions

Shnoulle:
> As always with EM, hard to test there are no regressions

I try to limit regression ONLY if there are a ASSIGN inside the equation … 
And here : it's already limited to JS part.

I don't know what happen if you set (for relevance equation) something like `G02Q01 = 2` for G01Q02 question but i think it's not a available usage.

Another idea { G01Q01 = G01Q02 = G02Q01} in G01Q03, since G02Q01 is not in the page : no JS action. Previous one break … after previous one …

I think we must clearly support only ONE assign solution `{QUESTIONCODE = (the equation to set QUESTIONCODE)}` in Equation question. (you can set more with `{QUESTIONCODE = (the equation to set QUESTIONCODE)}{QUESTIONCODE1 = (the equation to set QUESTIONCODE1)}{QUESTIONCODE2 = (the equation to set QUESTIONCODE2)}` multiple bracket).

Maybe need to update manual ?

gabrieljenik:
> I try to limit regression ONLY if there are a ASSIGN inside the equation …
> And here : it's already limited to JS part.

Probably right :)

> I don't know what happen 

Honestly, not that into EM so it is hard for me follow this with fluency just by reading.
So can't give opinion

PR #2922, opened 2023-02-15
 Fixed issue #18639: Token browse can be unusable with lot of data

Shnoulle:
Fixed issue #18639: Token browse can be unusable with lot of data in attribute
Dev: wrap value inside in cell (and encode)
Dev: ellipsed bigger value
Dec: set a min-width and a max-width with a container for all value


Shnoulle:
@gabrieljenik there are some more fix related to usability set only with css : min-width+max-width for long text question

gabrieljenik (approved):

gabrieljenik:
> @gabrieljenik there are some more fix related to usability set only with css : min-width+max-width for long text question

Not sure what you mean. Sorry

Shnoulle:
> > @gabrieljenik there are some more fix related to usability set only with css : min-width+max-width for long text question
> 
> Not sure what you mean. Sorry

Difference with the previous version you review :)

Shnoulle:
Too complex to fix conflict redo the PR …

PR #2923, opened 2023-02-16
Fixed issue #18635: Export single response can goes to a white page

Shnoulle:
Dev: remove session usage for export response in massive action widget
Dev: send it via POST and clean up some $_GET value


gabrieljenik (approved):
Code structure is OK. Haven't tested

PR #2927, opened 2023-02-17
Fixed issue #18625: Cannot copy a survey when using a survey alias. T…   …ests added.

gabrieljenik:
Added unit tests.

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -1419,6 +1419,17 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         $surveyLanguageSetting = new SurveyLanguageSetting();
         $surveyLanguageSetting->setAttributes($insertdata, false);
         try {
+            // Clear alias if it was already in use
+            $surveyLanguageSetting->checkAliasUniqueness();
+            if ($surveyLanguageSetting->hasErrors('surveyls_alias')) {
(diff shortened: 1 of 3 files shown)


tests/unit/helpers/ImportTest.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace ls\tests;
+
+class ImportTest extends TestBaseClass
+{
+    /**
+     * Test importing a survey that has an alias.
+     */
+    public function testCopyASurveyWithAnAlias(): void
```
olleharstedt, line 10: +1 for using quick tests instead of slow browser scripting

application/helpers/admin/import_helper.php
```diff
@@ -1419,6 +1419,17 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         $surveyLanguageSetting = new SurveyLanguageSetting();
         $surveyLanguageSetting->setAttributes($insertdata, false);
         try {
+            // Clear alias if it was already in use
+            $surveyLanguageSetting->checkAliasUniqueness();
```
olleharstedt, line 1423: checkAliasUniqueness() - this sounds like it should return a boolean or something. But old function, or?

gabrieljenik: That's a validation rule for the model. Not sure if it can handle return values, so it can be called outside a model validation.

PR #2928, opened 2023-02-19
ucfirst(null) depreactated in PHP8.X

pitbulk:
Verify $model contains a value before calling ucfirst($model)

Dev:


gabrieljenik:
I think this was already taken care on master.
I will close the PR for now.
If you feel should be reopened, please do!

Thanks

PR #2929, opened 2023-02-20
replace table in the survey permission page

jack-han-115:
Dev:


Shnoulle:
It's only for the User list ?

1. SurveyGroup Permission must use same too then ? https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/SurveysGroupsPermission/subviews/currentUsersList.php
2. SureyGroups avec 2 permission model SurveysInGroup and SurveysGroups

If we can offer more global view (when adding Permission to Theme , to label Sets etc …)

ptelu:
merged

PR #2934, opened 2023-02-20
Fixed issue #18644: YII_DEBUG was not used inside KCFinder

gabrieljenik:


vendor/kcfinder/core/bootstrap.php
```diff
@@ -108,6 +108,42 @@ function checkLSSession()
...
+
+    if (!defined('YII_DEBUG')) {
+        if (file_exists(APPPATH.'config'.DIRECTORY_SEPARATOR.'config.php')) {
+            $aSettings= include(APPPATH.'config'.DIRECTORY_SEPARATOR.'config.php');
+        } else {
+            $aSettings=array();
+        }
+        
+        // Set debug : if not set : set to default from PHP 5.3
+        if (isset($aSettings['config']['debug'])) {
```
Shnoulle, line 120: It's a copy/paste from index ? Unsure if we need to move to a function ?

vendor/kcfinder/core/bootstrap.php
```diff
@@ -108,6 +108,42 @@ function checkLSSession()
     {
         define("BASEPATH", realpath($sLimesurveyFolder . "/../vendor/yiisoft/yii/framework") . "/");
     }
+
+    if (!defined('YII_DEBUG')) {
+        if (file_exists(APPPATH.'config'.DIRECTORY_SEPARATOR.'config.php')) {
```
Shnoulle, line 113: config is already read here : https://github.com/LimeSurvey/LimeSurvey/blob/47c5aa59c3163da201af9b7269e8a20920b25813/vendor/kcfinder/core/bootstrap.php#L114

(see [internal](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/config/internal.php#L13))

then action must be done after in my option.

vendor/kcfinder/core/bootstrap.php
Shnoulle: Or maybe include a file to set debug (same for index and kcfinder (and maybe command)) `include(application/config/setdebug.php)`

Shnoulle (commented):
1. config is already read in this function : take action after
2. rename the function form `checkLSSession() `to `setLSConfig()`
3. Maybe move the `debug/define/error_reporting` to an included file ?

vendor/kcfinder/core/bootstrap.php
gabrieljenik: The changes the you suggest would require to move `$config = require_once(APPPATH . 'config/internal.php'); ` and some other includes to the top. 

To me, the position of those includes, taken for the index.php file, is set on purpose.
Wouldn't change that now and do that code rorg right now, but leave it to some other code reorg. The code reorg to the config files I believe is indeed needed, but wouldn't tackle it now.

Yes, code is repeating itself form index.php. But we are not creating a new entry of duplicated code, just fixing old duplicates. Not the best, but I don't think this suggested code reorg is priority right now.

@olleharstedt Thoughts?

vendor/kcfinder/core/bootstrap.php
gabrieljenik: If doing this, we would need to load the config file independently for that function. Hence on the index.php the config file would be loaded twice.

Again, I believe we need some code reorg, but I believe it should be as a separate task.

vendor/kcfinder/core/bootstrap.php
Shnoulle: You answer on both comments her .

Why not

```
    $config = require_once(APPPATH . 'config/internal.php');
    /* Set debug system here */
    require_once APPPATH . '../vendor/autoload.php';
    require_once BASEPATH . 'yii.php';
    require_once APPPATH . 'core/LSYii_Application.php';
    Yii::$enableIncludePath = false;
```

Why moving debug system after loading break something ?

vendor/kcfinder/core/bootstrap.php
Shnoulle: > If doing this, we would need to load the config file independently for that function. Hence on the index.php the config file would be loaded twice.

We just include the system to set debug, not load config a second time …

Just include …

> Again, I believe we need some code reorg, but I believe it should be as a separate task.

Currently : debug system is only in index, you copy paste on this commit … then it be a reason of reorganization.

vendor/kcfinder/core/bootstrap.php
gabrieljenik: > Why moving debug system after loading break something ?

Just feel the order is like that and can't be updated that easily. Not sure if  want to mess with it right now for this bug fix.
It is like a pandora box.

For example: 
- config/internal.php doesn't rely on anything from autoload?
- I have just spotted config/internal.php relies on YII_DEBUG being set already.
- I have just spotted config/internal.php relies on CMap, so Yii should already be loaded.
...

So, I think experimenting with the order doesn't fit in this ticket.

vendor/kcfinder/core/bootstrap.php
gabrieljenik: > We just include the system to set debug, not load config a second time …

But the config.php would be included twice, right? once on the general bootstrapping and then again for the setdebug.php new file, right?

But maybe already it is happening now... I will give it a thought.

Shnoulle: > then again for the setdebug.php new file, right?

Maye just a finction in include, using $settings array ?

gabrieljenik: Check now please

vendor/kcfinder/core/bootstrap.php
```diff
@@ -108,6 +108,9 @@ function checkLSSession()
     {
         define("BASEPATH", realpath($sLimesurveyFolder . "/../vendor/yiisoft/yii/framework") . "/");
     }
+
+    include APPPATH . '../setdebug.php';
```
olleharstedt, line 112: Hacking into vendor/ folder???

gabrieljenik: Well, that file is already hacked :)
It is not KCFinder original file

Shnoulle (approved):

PR #2941, opened 2023-02-24
Fixed issue #18476: CPDB - Add attribute modal - delete language does not work

gabrieljenik:


application/views/admin/participants/modal_subviews/_editAttribute.php
```diff
@@ -219,7 +219,7 @@
             e.preventDefault();
             var self = this,
                 attribute_id = $('#ParticipantAttributeName_attribute_id').val(),
-                formGroup = $(this).closest('div.form-group.'),
```
Shnoulle, line 222: Argl …

Shnoulle (approved):

PR #2952, opened 2023-02-28
Fixed issue #18658: Upload file question type

gabrieljenik:
Removed the decodeComponent. Believe is not needed.


Shnoulle (approved):
Don't find security issue
Test must be done with + %20 etc …

c-schmitz (approved):

PR #2953, opened 2023-02-28
Fixed issue #18662: Update DB from 491 break on home page

Shnoulle:
Dev: check if exist before use it


Shnoulle:
I really think i can test it (i have the issue and i fix it …)

olleharstedt:
Danke schön :)

PR #2956, opened 2023-03-01
Fixed issue #18655: Data loss if field in response table it missing

gabrieljenik:
New method: setAllAttributes.

### DIFF
--- a/application/models/Response.php
+++ b/application/models/Response.php
@@ -259,7 +259,6 @@ abstract class Response extends Dynamic
     /**
      * Set all the specified attributes.
      * If any attribute doesn't exist in the DB, an Exception is thrown.
-     * @inhe
      */
     public function setAllAttributes($values)
     {
--- a/application/models/Response.php
+++ b/application/models/Response.php
@@ -255,4 +255,24 @@ abstract class Response extends Dynamic
         }
         return $aAttributes;
     }
+
+    /**
+     * Set all the specified attributes.
+     * If any attribute doesn't exist in the DB, an Exception is thrown.
+     * @inhe
+     */
+    public function setAllAttributes($values)
+    {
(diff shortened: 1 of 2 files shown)


Shnoulle:
this mean setAttributes have more attribute than setAllAttributes ?

Maybe name is not good ? Or fix can be different ?

The previous issue was about setAttributes with token attributes inside ?  setTableAttributes for the new name ?

gabrieljenik:
> this mean setAttributes have more attribute than setAllAttributes ?

setAllAttributes is a brand new method.
We can name it differently if prefer.

> The previous issue was about setAttributes with token attributes inside ? setTableAttributes for the new name ?

I prefer to fix that on another ticket. The original issue is a blocking one that needs to be solved.

Shnoulle (approved):
Unsure on the name , but don't found better idea :)

:+1:

application/models/Response.php
```diff
@@ -255,4 +255,24 @@ public static function getEncryptedAttributes($surveyid = 0)
         }
         return $aAttributes;
     }
+
+    /**
+     * Set all the specified attributes.
+     * If any attribute doesn't exist in the DB, an Exception is thrown.
+     * @inhe
```
olleharstedt, line 262: inhe? :)

olleharstedt (changes requested):

olleharstedt:
> @olleharstedt Does this requires changes still?

No idea, ping Carsten on Cliq?

gabrieljenik:
> > @olleharstedt Does this requires changes still?
> 
> No idea, ping Carsten on Cliq?

Ohh, I mention it as you were the one requesting changes.
I will ping Carsten still. Thanks!

PR #2957, opened 2023-03-02
Fixed issue #18630: Survey list showing wrong icon for surveys. Tests.   Unit tests added for getRunning function.

gabrieljenik:


c-schmitz:
See comments on https://bugs.limesurvey.org/view.php?id=18630

PR #2958, opened 2023-03-02
Propagate survey context to submenus

yent:
Fixes broken "Simple plugins" survey menu item link

On survey admin, open up the plugin menu in the sidebar, the link on the "Simple plugins" item is broken, it is missing value for the "surveyid" parameter, this is because the survey context was not passed from Surveymenu::createSurveymenuArray to Surveymenu::getSurveymenuSubmenus

Fixed issue # cannot log into mantis with account I just created


c-schmitz:
Hi, thank you for the patch. We cannot reproduce the issue with the 'Simple plugin' link being broken.
I think the issue might be related to an outdated menu entry.  Can you try to reset the menu entries in the menu entries dialog and check if it then works without your patch?

c-schmitz:
Closed due to missing feedback from author.

PR #2960, opened 2023-03-02
Fixed issue #18630: Survey list showing wrong icon for surveys.

gabrieljenik:
The icon was failing only when time adjust was zero.

Created tests.
Did a small change to fix the situation, and avoid using the "sometimes-wrongly-calculated" value.

These needs to be reviewed a bit more as there is a lot of duplicated code.
Draft as to continue later: https://github.com/LimeSurvey/LimeSurvey/pull/2957

So far, with this I think is a good starter.
- Fix
- No reorg
- Tests (which will be usefull for making sure future reorg is ok)

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -1154,7 +1154,7 @@ class Survey extends LSActiveRecord implements PermissionInterface
             // If it's active, then we check if not expired
             // Time adjust
             $sNow    = date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime(date("Y-m-d H:i:s"))));
-            $sStop   = ($this->expires != '') ? date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime($this->expires))) : null;
+            $sStop   = ($this->expires != '') ? date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime($this->expires))) : $sNow;
             $sStart  = ($this->startdate != '') ? date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime($this->startdate))) : $sNow;
 
             // Time comparaison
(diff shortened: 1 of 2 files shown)


application/models/Survey.php
```diff
@@ -1154,7 +1154,7 @@ public function getRunning()
             // If it's active, then we check if not expired
             // Time adjust
             $sNow    = date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime(date("Y-m-d H:i:s"))));
-            $sStop   = ($this->expires != '') ? date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime($this->expires))) : null;
+            $sStop   = ($this->expires != '') ? date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime($this->expires))) : $sNow;
```
Shnoulle, line 1157: ? If expires is not set : it must stay not set, why set as now ?

Shnoulle: Why not `$sStop = ''` in this case ?

application/models/Survey.php
```diff
@@ -1169,17 +1169,17 @@ public function getRunning()
...
-            $sIconFuture  = '<a href="' . App()->createUrl('/surveyAdministration/view/surveyid/' . $this->sid) . '" class="survey-state" data-toggle="tooltip" title="' . sprintf(gT('Start: %s'), $sStart) . '"><span class="fa  fa-clock-o text-warning"></span><span class="sr-only">' . sprintf(gT('Start: %s'), $sStart) . '</span></a>';
+            $sIconFuture  = '<a href="' . App()->createUrl('/surveyAdministration/view/surveyid/' . $this->sid) . '" class="survey-state" data-toggle="tooltip" title="' . sprintf(gT('Start: %s'), $sStart) . '"><span class="fa fa-clock-o text-warning"></span><span class="sr-only">' . sprintf(gT('Start: %s'), $sStart) . '</span></a>';
 
             // Icon parsing
             if ($bExpired || $bWillRun) {
                 // Expire prior to will start
                 $running = ($bExpired) ? $sIconExpired : $sIconFuture;
             } else {
-                if ($sStop == null) {
+                if ($this->expires == '') {
```
Shnoulle, line 1182: Then $sTop === ''

Shnoulle (commented):
I don't understand the update ?

I think it's near same code after and before …

application/models/Survey.php
```diff
@@ -1154,16 +1154,16 @@ public function getRunning()
...
+            $sStart  = ($this->startdate != '') ? date("Y-m-d H:i:s", strtotime(Yii::app()->getConfig('timeadjust'), strtotime($this->startdate))) : null;
 
             // Time comparaison
             $oNow   = new DateTime($sNow);
             $oStop  = new DateTime($sStop);
             $oStart = new DateTime($sStart);
 
-            $bExpired = ($oStop < $oNow);
-            $bWillRun = ($oStart > $oNow);
+            $bExpired = (!is_null($sStop) && $oStop < $oNow);
```
Shnoulle, line 1165: Oh yes, here : i better understand …  (but don't reproduce too … )

Shnoulle (approved):
More clear :+1:

PR #2975, opened 2023-03-08
Fixed #17845: Label set preview only shows last item, other items are missing

blaniel:
Create a new Label instead of trying to get one which does not exists since every label is deleted at the start of 'modlabelsetanswers'. Besides, Label search was using LabelSet id (lid) which led to labels corruption.


gabrieljenik:
Can I please ask you to review if this patch is still needed with the new v6? Thanks

c-schmitz:
Closed due to missing feedback from author.

PR #2976, opened 2023-03-08
Cr 892/statistics response disable

jack-han-115:
Dev:


application/models/Surveymenu.php
```diff
@@ -177,7 +177,15 @@ public function createSurveymenuArray($oSurveyMenuObjects, $collapsed = false, $
                 if ($oDataAttribute->isActive !== null && $oSurvey != null) {
                     if ($oDataAttribute->isActive == true && $oSurvey->active == 'N') {
                         $aEntry['disabled'] = true;
-                        $aEntry['disabled_tooltip'] = sprintf(gT("The '%s' section is not available yet. Please activate your survey to enable this section."), gT($aEntry['menu_title']));
+                        if($aEntry['name']==='responses') {
+
+                            $aEntry['disabled_tooltip'] = sprintf(gT("This survey is not active and has no %s."), gT($aEntry['name']));
```
olleharstedt, line 182: $aEntry['name'] is always "responses" here, no need to use the variable?

olleharstedt: Also gT($var) should be avoided in general, since the task to pull translations won't understand it.

PR #2977, opened 2023-03-08
Bug/18071 importing question through remote control may fail test ii

Shnoulle:
Fixed issue #18071: Importing Question through Remote Control may fail if the imported question already exists
Fixed issue #18070: Importing Question through Remote Control doesn't set text for all languages


gabrieljenik:
You can push the fix here whenever you want :)

Shnoulle:
> @Shnoulle You can push the fix here whenever you want :)

yes :) I know :)

PR #2978, opened 2023-03-09
show error alert if sending email failed

jack-han-115:
Dev:

### DIFF
--- a/application/models/UserGroup.php
+++ b/application/models/UserGroup.php
@@ -591,7 +591,7 @@ class UserGroup extends LSActiveRecord
                 $msgToUser .= gT('Sending successful') . "<br>";
             } else {
                 throw new Exception("Failed to send mail");
-                $msgToUser .= gT('Error: ') . $emaiLResult['msg'] . "<br>";
+                // $msgToUser .= gT('Error: ') . $emaiLResult['msg'] . "<br>";
             }
         }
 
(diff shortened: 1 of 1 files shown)


application/models/UserGroup.php
```diff
@@ -590,6 +590,7 @@ public function sendUserEmails($ugid, $subject, $body, $copy)
             if ($emaiLResult['success']) {
                 $msgToUser .= gT('Sending successful') . "<br>";
             } else {
+                throw new Exception("Failed to send mail");
                 $msgToUser .= gT('Error: ') . $emaiLResult['msg'] . "<br>";
```
olleharstedt, line 594: This line is dead code, can be removed.

jack-han-115: without this line, error is not catched

https://github.com/LimeSurvey/LimeSurvey/blob/91db5ff516f75b4f947c929aebbbf5220620d0a7/application/controllers/UserGroupController.php#L475

application/models/UserGroup.php
```diff
@@ -590,7 +590,8 @@ public function sendUserEmails($ugid, $subject, $body, $copy)
             if ($emaiLResult['success']) {
                 $msgToUser .= gT('Sending successful') . "<br>";
             } else {
-                $msgToUser .= gT('Error: ') . $emaiLResult['msg'] . "<br>";
+                throw new Exception("Failed to send mail");
+                // $msgToUser .= gT('Error: ') . $emaiLResult['msg'] . "<br>";
```
olleharstedt, line 594: No reason to leave it in the code, @jack-han-115 . :) If we need to find it, it's in the commit history.

PR #2983, opened 2023-03-10
Fixed issue #18674: Simple user can update javascript with XSS and di…

Shnoulle:
Fixed issue [security] #18674: Simple user can update javascript with XSS and disable script ON
Dev: model part : can not be updated by simple user
Dev: readonly mode on textarea


Shnoulle:
TODO : readonly mode (or disable) on ace system sisn't work currently

Shnoulle:
@c-schmitz https://github.com/LimeSurvey/LimeSurvey/blob/master/vendor/jquery-ace/jquery.ace.js#L4 

get in our own repo or goes to https://cheef.github.io/jquery-ace/

application/core/LSYii_NoUpdateValidator.php
```diff
@@ -26,10 +26,6 @@ class LSYii_NoUpdateValidator extends CValidator
      */
     public function validateAttribute($object, $attribute)
     {
-        if (Yii::app()->user->isScriptUpdateAllowed()) {
```
gabrieljenik, line 29: I am not sure about this. How this plays out?

gabrieljenik: This actually reverses this:
https://github.com/LimeSurvey/LimeSurvey/commit/3eacd351e644ffb46ec06c69d3f5112febffa8a6

Shnoulle: Yes : because is the way to do … 
LSYii_NoUpdateValidatordisable usage of LSYii_NoUpdateValidator for script …

@c-schmitz : why you do https://github.com/LimeSurvey/LimeSurvey/commit/3eacd351e644ffb46ec06c69d3f5112febffa8a6 this break the action on script

LSYii_NoUpdateValidator : don't update an attribute

Need to be done for script

Shnoulle: PS : without this fix : any LimeSurvey admin can update script …  on all condition.

Shnoulle:
Added an auto test (showing current system allow any admin to add script)

tests/functional/backend/XssAndScriptCheckTest.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * @return void
+     */
+    public function testXssOnAndScriptOn()
+    {
+        App()->setConfig('filterxsshtml', true);
+        App()->setConfig('disablescriptwithxss', true);
+        $questionl10n = $this->importSurveyAndgetQuestionI10n();
```
gabrieljenik, line 36: So you are testing the validator by triggering an import.
Can't you just do a question save? Isn't that a more direct test?

Also, maybe a unit test to isScriptUpdateAllowed method?

Shnoulle: >  Can't you just do a question save? Isn't that a more direct test?

It must be done too in my opinion, but import **must** work too (and in general is the 2st place where it's not done : only usage of model, in GUI : we use controller).

Like it must be done on group, survey content etc …

> Also, maybe a unit test to isScriptUpdateAllowed method could be usefull?

Script update is allowed if XSS is POFF. The 2 feature are really linked. Test both at same time seems better.

Shnoulle: > Also, maybe a unit test to isScriptUpdateAllowed method could be usefull?

better understand : `testXssOnAndScriptOff` do it

PR #2984, opened 2023-03-10
Fixed issue #18671: Potential loop with equation set question

Shnoulle:
Dev: be sure to do action for hidden equation item


Shnoulle:
For the issue : i don't found why bit action is done with a hidden input without name … maybe an input inside the HTML updated ?

With the sample survey : values are `/` (and after string for the real input)

gabrieljenik (approved):
Code structure is OK. Not tested

PR #2985, opened 2023-03-11
Audio recording question

AmauryVanEspen:
New feature #<Mantis issue number>: audio record question
Dev: @mhkuu


c-schmitz:
Unfortunately, the patch is really outdated as it was still for version 5.

PR #2997, opened 2023-03-16
Bug/18674 script allowed for restricted user test

Shnoulle:
To ass test on https://github.com/LimeSurvey/LimeSurvey/pull/2983 

BUT : broken actually


Shnoulle:
OK, goty it

PR #2998, opened 2023-03-17
Fixed issue #18655: Data loss if field in response table it missing.

gabrieljenik:
DEV: Remote control export responses test added.
DEV: Placing all export responses tests together.
DEV: Refresh maxId() on remote control handle. Avoid caching. Error-prone when multiple export call on the same request (unusual on real life, usual on auto tests)
DEV: Moving code sniffer to the bottom as to have failed tests quicker


Shnoulle:
? I don't understand : update file are remote_control ?

Shnoulle (commented):
OK, got it.

I think we must fix getMaxid

Move static variable to Model static variable https://github.com/LimeSurvey/LimeSurvey/blob/b69fe90f78d3327b4f1e62b830b9cd802a3467ef/application/models/LSActiveRecord.php#L122

Because the issue here is nusage of same static variable for different model

Shnoulle:
I can create a PR on this one if needed.

gabrieljenik:
We reviewed that with @olleharstedt .
The diema was this is only happening on these tests as the model is used with different surveyIds within the same HTTP request. 
This is not happening on real life.
So should we spend time on this?

@olleharstedt Has this matter.

My idea was to overload refreshMetadata() and reset the maxId calling it with the refresh parameter.

gabrieljenik:
> why you don't want to fix it ?

Never said I don't want to fix it.
Not sure if worth it.

> @olleharstedt Has this matter.
> 
> My idea was to overload refreshMetadata() and reset the maxId calling it with the refresh parameter.

 @olleharstedt What do you think about a PR to DEV?

Shnoulle:
Issue : static var maxId is global must be attached to Model

No need complex code here …

Shnoulle (approved):

Shnoulle:
OK, i made an error : usage of `public static $maxIds = [];` didn't fix the issue …

Survey::model() share the same maxIds.

Using dynamicId start to be complex …

gabrieljenik:
I would go with:

1. Add a new optional parameter to getMaxId
`public function getMaxId($field = null, $forceRefresh = false, $clear=false)`

2. Create LSActiveRecord::refreshMetadata() (overload of CActiveRecord).
    Make the refreshMatedata to call getMaxId using the clear parameter.

Shnoulle:
> I would go with:
> 
>     1. Add a new optional parameter to getMaxId
>        `public function getMaxId($field = null, $forceRefresh = false, $clear=false)`
> 
>     2. Create LSActiveRecord::refreshMetadata() (overload of CActiveRecord).
>        Make the refreshMatedata to call getMaxId using the clear parameter.

I don't understand clear ?

Else, in my opinion a lot must be automatic. If not : it's unusuable

For example, you can do 

```
$maxId1 = \Group::model()->getMaxId();
$maxId2 = \QuestionGroup::model()->getMaxId();
$maxId3 = \Token::model(123)->getMaxId();
$maxId4 = \SurveyDynamic::model(123)->getMaxId();
$maxId5 = \SurveyDynamic::model(321)->getMaxId();
$maxId6 = \Response::model(321)->getMaxId();
```

You must have the real `$maxId` at each time.

And more ! : 

```
$maxId2 = \QuestionGroup::model()->getMaxId();
$oQuestionGroup = new \QuestionGroup());
…
$oQuestionGroup->save();
$maxId3 = \QuestionGroup::model()->getMaxId();
```

must be OK too 

A question : did we really need this one ? Maybe just delete this statci var

I think it's best to optimize other request (and use static or cache)

- https://bugs.limesurvey.org/view.php?id=18473
- https://bugs.limesurvey.org/view.php?id=18472
- etc …

olleharstedt:
> A question : did we really need this one ? Maybe just delete this statci var

Yes, this was my conclusion too. It looks like premature optimization in this case.

Shnoulle:
> > A question : did we really need this one ? Maybe just delete this statci var
> 
> Yes, this was my conclusion too. It looks like premature optimization in this case.

Must do some requets and see how many times it used .

But clearly when we have 396 times `SELECT users where uid = 1` and 192 times select SurveyGroups …   i think we must improve this before all other action.

olleharstedt:
> Must do some requets and see how many times it used .

Just look at the code, man.

olleharstedt:
> But clearly when we have 396 times SELECT users where uid = 1 and 192 times select SurveyGroups … i think we must improve this before all other action.

Sure, but focus on the bottlenecks first, not just random stuff.

gabrieljenik:
Created a ticket. Please follow conversation there :)
https://bugs.limesurvey.org/view.php?id=18699

PR #3001, opened 2023-03-20
Fixed issue #18633: Inherited value doesn't match any attribute option

gabrieljenik:


application/views/themeOptions/options_core.php
```diff
@@ -126,7 +126,7 @@
                                         $sParentLabelOption = $options[$sParentOption];
                                         $options['inherit'] = gT("Inherit") . ' [' . gT($sParentLabelOption) . ']';
                                     } else {
-                                        $sParentOption = $sParentOption !== '' ? gT($sParentOption) : $sParentOption;
+                                        $sParentOption = isset($options[$sParentOption]) ? gT($options[$sParentOption]) : $sParentOption;
```
Shnoulle, line 129: usage of empty maybe ? here you seems you can go to gT("") ?

And previous system seems to send '' as value

Shnoulle (commented):
Usage of empty ? "" null and 0 must not be translated

Shnoulle (approved):
:+1:

application/views/themeOptions/options_core.php
```diff
@@ -126,7 +126,7 @@
                                         $sParentLabelOption = $options[$sParentOption];
                                         $options['inherit'] = gT("Inherit") . ' [' . gT($sParentLabelOption) . ']';
                                     } else {
-                                        $sParentOption = $sParentOption !== '' ? gT($sParentOption) : $sParentOption;
+                                        $sParentOption = ! empty($options[$sParentOption]) ? gT($options[$sParentOption]) : '';
```
olleharstedt, line 129: Can we use `??` instead please?

olleharstedt (changes requested):

olleharstedt:
Pretty big conflict, probably due to the LS6 release. Might be easier to redo the PR?

application/views/themeOptions/options_core.php
Shnoulle: ?? is only for null, here it's for `'' `

olleharstedt: No, the check is on `empty`

olleharstedt: `$options[$sParentOption] ?? ''`

Shnoulle: > `$options[$sParentOption] ?? ''` should mean the same.

No , if `$options[$sParentOption] === ''` then you return `gT('')` we already have it in 5.X for help when help was not set by plugin (then are '' (and not null).

olleharstedt: Oh. So

```php
gT($options[$sParentOption] ?? '')
```

?

olleharstedt: Ah not exactly the same I guess.

gabrieljenik: > No, the check is on `empty`

Empty returns false if string is empty.
That's why it is not the same.

PR #3003, opened 2023-03-22
Fixed issue #18610: Email attachments expose absolute path when exporting LSS

gabrieljenik:
- Turning paths from absulote to relative before exporting.
- Turning paths from relative to absulote before importing.
- Exporting attachment info in json format


Shnoulle:
Why not save it as relative to `rootdir` ?

And more i think its must be relative to `upload` dir

gabrieljenik:
Well, I was trying to solve the export issue first, in a quick and not too complex way.
The tackle how showuld we save it.

For saving, it is not only about relative, but also not allowing traversal, for example.
(Maybe not easy to do it now from current UI but yes messing with the HTML a bit)
And maybe there are other stuff as well.. So, I prefer to handle that on a diff ticket...

Shnoulle:
>  (Maybe not easy to do it now from current UI but yes messing with the HTML a bit)

😩 totally right we have really complex UI sometimes …

Shnoulle (approved):

gabrieljenik:
> why not save as json directly ?

On this ticket, only import export.
Why? 
- Minimize the impact: easier faster testing & deploy, less risk
- Prioritize "saving as json on db" as needed. Could be right away if required.

Shnoulle:
I assign @c-schmitz on mantis : see this message : https://bugs.limesurvey.org/view.php?id=18610#c74792

It's OK for me

c-schmitz (approved):

PR #3004, opened 2023-03-22

gabrieljenik:


Shnoulle:
OK for new one, 
But must fix in Vanilla theme too : filter value before using it.

gabrieljenik:
> But must fix in Vanilla theme too : filter value before using it.

Filter what? if it is already filtered.

Shnoulle (approved):

application/models/Survey.php
```diff
@@ -552,6 +552,11 @@ public function rules()
             array('startdate', 'date','format' => ['yyyy-M-d H:m:s.???','yyyy-M-d H:m:s','yyyy-M-d H:m'],'allowEmpty' => true),
             array('datecreated', 'date','format' => ['yyyy-M-d H:m:s.???','yyyy-M-d H:m:s','yyyy-M-d H:m'],'allowEmpty' => true),
             array('expires', 'checkExpireAfterStart'),
+            // The Google Analytics Tracking ID is inserted in a JS script. If the following rule is changed, make sure
+            // that it doesn't render it vulnerable to XSS attacks.
+            array('googleanalyticsapikey', 'match', 'pattern' => '/^[a-zA-Z\-\d]*$/',
+                'message' => gT('Google Analytics Tracking ID may only contain alphanumeric characters and hyphens.'),
+            ),
```
c-schmitz, line 559: Is 'allowEmpty' missing here?

c-schmitz (changes requested):

application/models/Survey.php
gabrieljenik: Do we need one if the pattern allows empty (there is a star near the end of the pattern)?

Shnoulle:
Needed for 3.X and 5.X right ?

Shnoulle:
Confirm : still XSS issue in 3.X and 5.X

tiborpacalat:
If it is still an issue then we need to port this fix to 3 and 5 as well.

Shnoulle:
> If it is still an issue then we need to port this fix to 3 and 5 as well.

Yes, i put a comment on mantis

PR #3005, opened 2023-03-22
Cr 1045 formextension rendering

kevin-foster-uk:
Refactor form extension to avoid duplication in rendering code.


olleharstedt:
Looks good, but are changes needed to the LimeSurvey professional plugin now? Because you changed namespace names.

kevin-foster-uk:
@olleharstedt This has been hanging around for a while. Maybe we should discuss?

kevin-foster-uk:
> Looks good, but are changes needed to the LimeSurvey professional plugin now? Because you changed namespace names.

I tend to avoid plurals in namespaces.

The plural indicates a collection but you have classes which are not of type "Input":  Inputs\DefaultBaseRenderer and Inputs\GlobalSettingsRenderer.

olleharstedt:
>      I tend to avoid plurals in namespaces.

Isn't there a PSR standard for that tho?

kevin-foster-uk:
@olleharstedt Not sure why this was never merged? Should be abandon this PR?

olleharstedt:
> @olleharstedt Not sure why this was never merged? Should be abandon this PR?

Hm not sure. It needs to be tested with the cloud plugin before merge, else it looks good, I think. :)

PR #3008, opened 2023-03-23
DEV #18099: Automatic testing for navigation buttons.

gabrieljenik:
Test for index navigation added.


tests/functional/frontend/SurveyIndexNavigationTest.php
```diff
@@ -0,0 +1,90 @@
...
+            );
+            $groupTitle = $groupTitleDiv->getText();
+            $this->assertEquals("My second question group", $groupTitle);
+
+            // Move from second group back to first group
+            //Click on index menu
+            $web->findElement(WebDriverBy::cssSelector('li.ls-index-menu'))->click();
+
+            //Click on menu option
+            $menuItems = $web->findElements(WebDriverBy::cssSelector('li.ls-index-menu ul.dropdown-menu li.index-item'));
```
Shnoulle, line 74: We can't use a selector like `data-limesurvey-submit='{ "move":1 }'` (for the top menu link) ?
And submit  `name="move" value ="1"` for the button ?

Because some class are related to theme here. the real action is done with this part, not other one.

You can directly add button like this in a question : it **must** work

Shnoulle (commented):
I really like to see real selector used not Theme dependant.

Shnoulle:
For the js hidden button : i think you can add it inside a question text for example.

(personally i like to create my own element for testing purpose with clean id). 

If you want i can update the test :)

gabrieljenik:
> For the js hidden button

I got lost. What hidden button?

> : i think you can add it inside a question text for example.

Why adding a new button?

Shnoulle:
> > For the js hidden button
> 
> I got lost. What hidden button?

https://github.com/LimeSurvey/LimeSurvey/blob/master/themes/survey/vanilla/views/subviews/navigation/question_index_buttons.twig
but : ls-js-hidden : https://github.com/LimeSurvey/LimeSurvey/blob/master/themes/survey/vanilla/views/subviews/navigation/navigator_complement.twig

> 
> > : i think you can add it inside a question text for example.
> 
> Why adding a new button?

Because it's a great feature for template : more clear than the menu : https://demo.sondages.pro/427384 for example.

gabrieljenik:
> Because it's a great feature for template : more clear than the menu

So, if I understand correctly, you are not talking about updating the test, but updaintg the theme?

The test currently works and tries to mimic what a user would do.
If that is fine, I would just accept it and then tackle updates to theme on another ticket.

tests/functional/frontend/SurveyIndexNavigationTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+            // Wait max 10 second to find the second group title
+            $groupTitleDiv = self::$webDriver->wait(10)->until(
+                WebDriverExpectedCondition::presenceOfElementLocated(
+                    WebDriverBy::cssSelector('#group-0 .group-title')
+                )
+            );
+            $groupTitle = $groupTitleDiv->getText();
+            $this->assertEquals("My first question group", $groupTitle);
+        } catch (\Exception $ex) {
```
olleharstedt, line 83: For future reference, you might want to use Throwable instead of Exception, since it can catch more issues.

PR #3018, opened 2023-03-28
Fixed issue #18693: Using checkbox question for yes/no loses data when data entry view

gabrieljenik:


application/controllers/admin/DataEntry.php
```diff
@@ -614,12 +614,25 @@ public function editdata($subaction, $id, $surveyid)
...
+                $aDataentryoutput .= "</td>\n";
+                // Second column (Unseen)
+                $aDataentryoutput .= "<td style='width: 120px;'>\n";
+                // Only show the "Unseen" checkbox for questions.
+                if (!empty($fname['title'])) {
+                    $aDataentryoutput .= "<div class='checkbox'>\t"
+                        . "<input type='checkbox' name='{$fname['fieldname']}_unseen' id='{$fname['fieldname']}_unseen'"
+                        . (is_null($idrow[$fname['fieldname']]) ? " checked" : "")
+                        . ">"
+                        . "<label for='{$fname['fieldname']}_unseen'>" . gT("Unseen") . "</label>"
```
Shnoulle, line 630: Potential issue with subquestion with title == 'unseen'

Maybe use non alphanumeric character ? Using `:` maybe ? or `-` ? (boith seems allowed in name and in id)

Shnoulle (commented):
Since unseen can be used as subquetiontitle, maybe use a different separator :unseen can work ?

See : https://www.w3.org/TR/html401/types.html#type-name

application/controllers/admin/DataEntry.php
```diff
@@ -614,12 +614,25 @@ public function editdata($subaction, $id, $surveyid)
...
-                . "\n";
+                $aDataentryoutput .= ">\n";
+                // First column (Question)
+                $aDataentryoutput .= "<td>";
                 $aDataentryoutput .= stripJavaScript($question);
-                $aDataentryoutput .= "</td>\n"
-                . "<td>\n";
+                $aDataentryoutput .= "</td>\n";
+                // Second column (Unseen)
+                $aDataentryoutput .= "<td style='width: 120px;'>\n";
```
Shnoulle, line 623: Use em (or rem) for width . Then if user have bigger font size : it follow.

application/controllers/admin/DataEntry.php
```diff
@@ -620,14 +620,14 @@ public function editdata($subaction, $id, $surveyid)
                 $aDataentryoutput .= stripJavaScript($question);
                 $aDataentryoutput .= "</td>\n";
                 // Second column (Unseen)
-                $aDataentryoutput .= "<td style='width: 120px;'>\n";
+                $aDataentryoutput .= "<td style='width: 8em;'>\n";
                 // Only show the "Unseen" checkbox for questions.
                 if (!empty($fname['title'])) {
                     $aDataentryoutput .= "<div class='checkbox'>\t"
-                        . "<input type='checkbox' name='{$fname['fieldname']}_unseen' id='{$fname['fieldname']}_unseen'"
+                        . "<input type='checkbox' name='unseen_{$fname['fieldname']}' id='unseen_{$fname['fieldname']}'"
```
Shnoulle, line 627: But why not {$fname['fieldname']}:unseen ?

Shnoulle (approved):

Shnoulle:
The question still exist «why not {$fname['fieldname']}:unseen » ?

application/controllers/admin/DataEntry.php
gabrieljenik: I believe the `:` can be error prone or problematic.
The current naming schema works. So, don't want to add an extra layer of "risk" (as a way to say it) :)

gabrieljenik:
> The question still exist «why not {$fname['fieldname']}:unseen » ?

I believe the `:` can be error prone or problematic.
The current naming schema works. So, don't want to add an extra layer of "risk" (as a way to say it) :)

Shnoulle:
> 



> I believe the `:` can be error prone or problematic.
> The current naming schema works. So, don't want to add an extra layer of "risk" (as a way to say it) :)

You think i write : without controlling it's OK ? 

> ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").

Source : https://www.w3.org/TR/html401/types.html#type-name

gabrieljenik:
> You think i write : without controlling it's OK ?

No, of course not. Why do you take it like that.
I am just saying it is not very usual to use `:` on names, so prefer to not use it.

We can disagree sometimes, right? If at the end it works....

Shnoulle:
PS : strange when i try to test : seems to be a 5.X

Shnoulle:
> No, of course not. Why do you take it like that.

Because you don't tell why you don't want it …

> I am just saying it is not very usual to use : on names, so prefer to not use it.

Its one of the reason to use it, you have less chance to have it in a name. And it's make clear that is a new name.

In LimeSurvey : we already use _ as separator between code and subquestion code, the when see _ : a new xdev can think it's a separator between code and subcode.

Another : there are discussion since years to use QuestionCode in place of sidXgidXqid, with unseen at start with _ : you disable usage of unseen.

> We can disagree sometimes, right? If at the end it works....

Yes, but maybe discuss on idea is a good solution to have something work for long term and more understandable

olleharstedt:
Right, both underscore and hash are used in question code generation? Or is that SGQA code only?

Where in the code is the current naming schema decided? Can't find it in the DataEntry class.

@Shnoulle You have an example that could break the current schema?

olleharstedt:
`name='unseen_{$questionInputField}'`

This one right? Pja, can it break if someone creates a question with code "unseen" and then subquestions?

Shnoulle:
>@Shnoulle You have an example that could break the current schema?

There none with unseen before, potential with unseen after.


> `name='unseen_{$questionInputField}'`
> 
> This one right? Pja, can it break if someone creates a question with code "unseen" and then subquestions?

Only if we update ColumnName to question->title in future (years and years 🤪)

> 
> Double underscore is another alternative, maybe. Or `name=unseen[$questionCode]`


`:` still available in all condition … but why not an array of unseen


But : here : it's not a blocker for merigng (for me) : see https://github.com/LimeSurvey/LimeSurvey/pull/3109


1. Single choice with comment  not answered : unseen state
2. Numeric : unseen state shown for not answered
3. Multi numeric : unseen state shown for not answered
4. Date : unseen state shown for not answered

But the worst : We must avoid to loose text (sometime big) by error Then
https://bugs.limesurvey.org/view.php?id=18693#c74857

Shnoulle:
Need master merge in it to test and review  : currently it's the Admin GUI of 5.X

gabrieljenik:
> PS : strange when i try to test : seems to be a 5.X

Yes, we started when master was v5
Should be v6 now.

gabrieljenik:
Will review the comments you tested. Thanks

Shnoulle:
> Will review the comments you tested. Thanks

My previous comment are just comment : no «Need update by author». 

But when do a quick test : i see 4 + 1 issues.

gabrieljenik:
Waiting on approavl of #3109 before updating this one.

gabrieljenik:
Picked up. Ready for re-reviewal

application/controllers/admin/DataEntry.php
```diff
@@ -1456,6 +1529,24 @@ public function update()
                 continue;
             }
             $thisvalue = Yii::app()->request->getPost($fieldname, '');
+            // For questions, if the "Unseen" checkbox is checked, we must set the field to null.
+            // There are some special cases we need to handle.
```
Shnoulle, line 1533: Argl … always special case :((((((

application/controllers/admin/DataEntry.php
```diff
@@ -1456,6 +1529,24 @@ public function update()
...
+            if ($irow['type'] == Question::QT_R_RANKING) {
+                $unseenFieldName = "unseen:" . $irow['sid'] . 'X' . $irow['gid'] . 'X' . $irow['qid'];
+            } elseif ($irow['type'] == Question::QT_P_MULTIPLE_CHOICE_WITH_COMMENTS) {
+                // Remove trailing "comment" from the fieldname, if present
+                $unseenFieldName = "unseen:" . preg_replace('/comment$/', '', $fieldname);
+            } else {
+                $unseenFieldName = "unseen:" . $fieldname;
+            }
+            if (!empty($irow['title']) && Yii::app()->request->getPost($unseenFieldName, false)) {
+                // Throw an error if "unseen" is checked but the field is not empty. This should never happen.
```
Shnoulle, line 1543: Send a flash error seems better in my opinion ?

1. If value + unseen is set : save the value
2. But send a alert danger message 

```
if (!empty($irow['title']) && Yii::app()->request->getPost($unseenFieldName, false)) {
    // Throw an error if "unseen" is checked but the field is not empty: javascript didn't work good here
    if ($thisvalue === '') {
        $oResponse->$fieldname = null;
    } else {
        App()->addFlash("Question {$irow['title']} was set to unseen but have value. Valuse was saved", 'danger');
    }
    continue;
}
```

Else a JS speed issue or anything other broken : user loose all data sent …

Shnoulle (commented):
Please : don't loose data sent

(i didn't review all currently)

gabrieljenik:
> Please : don't loose data sent

We set JS validation, just like the other branch...

Shnoulle:
> > Please : don't loose data sent
> 
> We set JS validation, just like the other branch...

Don't send a 400 error in case JS broken !

If JS broken : save data but send an flash alert.

Shnoulle:
> > Please : don't loose data sent
> 
> We set JS validation, just like the other branch...

About other branch : this one is the 1st done, i didn't see the 400 error before sorry.

PR #3025, opened 2023-03-29
Fixed issue #18679 API Access to Question Type Array Dual Scale merge scales if answer code is not unique.

gabrieljenik:
- Fixed: Now returning multi scale answer options 
- Fix language issue when using multiple scale answer option


Shnoulle (approved):
seems clean (not tested)

Shnoulle:
has auto-test and not a visual issue : i think it's OK to merge

PR #3038, opened 2023-04-05

Shnoulle:
Fixed issue : Multiple choice questions columns not in one (ul) tag.
Fixed issue : broken HTML for image multiple choice (li without ul)
Dev: theme style solution
Dev: work for limited column number


Shnoulle:
I still have an difference between multiple choice and single choice with other, 
Don't understand …

Capture d’écran du 2023-04-05 11-47-02
 
[limesurvey_survey_Other4Columns7choices.lss.txt](https://github.com/LimeSurvey/LimeSurvey/files/11157453/limesurvey_survey_Other4Columns7choices.lss.txt)

assets/packages/limesurvey/survey.css
```diff
@@ -177,3 +177,28 @@ box-shadow: 0 0 0 #fff;
...
+  data-column-count-9{column-count:9}
+  data-column-count-10{column-count:10}
+  data-column-count-11{column-count:11}
+  data-column-count-12{column-count:12}
+  data-column-count-13{column-count:13}
+  data-column-count-14{column-count:14}
+  data-column-count-15{column-count:15}
+  data-column-count-16{column-count:16}
+  data-column-count-17{column-count:17}
+}
```
ptelu, line 204: this doesnt seem to work, i added it manually to test it.
Also seems very random to put a finite amount of numbers here. I think this could be solved in a better way.
There is also the issue with small screen sizes where previously below 768px the columns option was ignored since it would grow way to small.
ul1

application/core/QuestionTypes/ListRadio/RenderListRadio.php
```diff
@@ -58,17 +54,12 @@ public function __construct($aFieldArray, $bRenderDirect = false)
...
 
         if ($this->iNbCols > 1) {
-            // Add a class on the wrapper
             $this->sCoreClass .= " multiple-list nbcol-{$this->iNbCols}";
-            // First we calculate the width of each column
-            // Max number of column is 12 http://getbootstrap.com/css/#grid
-            $this->iColumnWidth = round(12 / $this->iNbCols);
-            $this->iColumnWidth = ($this->iColumnWidth >= 1) ? $this->iColumnWidth : 1;
-            $this->iColumnWidth = ($this->iColumnWidth <= 12) ? $this->iColumnWidth : 12;
-            $this->iMaxRowsByColumn = ceil($this->getAnswerCount() / $this->iNbCols);
```
ptelu, line 68: the good thing about this behavior was that it didnt allow the user to enter a ridiculous amount of columns, the hard limit with this is 12. Also the column-count doesnt seem to utilize the full width available.
Example with 16:
ul2
ul3

assets/packages/limesurvey/survey.css
Shnoulle: > Also seems very random to put a finite amount of numbers here.

1. Before we use only bootstrap grid ( 1,2,34,6,12)
2. See https://github.com/LimeSurvey/LimeSurvey/pull/3038/commits/786b0536b43d6e00c334e4eea9c4b30eb088eb49#diff-bae3ead6fc4f735cd15964c14a48fc6eb812031b0050ea1fcc72fe26b42a0d77R309 `[data-column-count]{column-count:attr(data-column-count)}` didn't work
3. I really don't see how to get to any number, have another idea ?

>  There is also the issue with small screen sizes where previously below 768px the columns option was ignored since it would grow way to small.

It was limited to 768px : https://github.com/LimeSurvey/LimeSurvey/pull/3038/files#diff-02c3953154c744920a0bd8c8a9a1937be4a92d3c550a3391b9a3e09e7e4b2d9aR182 it don't work ?

application/core/QuestionTypes/ListRadio/RenderListRadio.php
Shnoulle: > the good thing about this behavior was that it didnt allow the user to enter a ridiculous amount of columns, the hard limit with this is 12.

What ? Why it's a good thing to disallow user to set 14 ? Why you want user dpon't do what they want, and if it don't work : they can review the system… If user want 14 columns : maybe there are a good reason.

> Also the column-count doesn't seem to utilize the full width available.

Columns take same height, and you set as 16 columns, i think it brother dependant and depend on the content of each choice . I don't see another way with
```
A C
B D
```
order than using column-count.
With BS : need extra element 

With 
```
A B
C D
```
we can use https://getbootstrap.com/docs/5.3/layout/grid/#row-columns
But i think it's must be a new option ([Feature request](https://bugs.limesurvey.org/view.php?id=7927))

assets/packages/limesurvey/survey.css
Shnoulle: > Also seems very random to put a finite amount of numbers here.

In next comment you write «it's better to force limit to 12» … seems inconsistent …

application/core/QuestionTypes/ListRadio/RenderListRadio.php
Shnoulle: >  Example with 16:

Before you can not set 16 … 

Even with 8 : it don't work
Capture d’écran du 2023-04-14 11-36-02

Shnoulle: > Also the column-count doesnt seem to utilize the full width available.

Same with 5.X (and with same width for all part)

Shnoulle:
Please compare with current situation :)

Where 

1. 16 is not allowed
2. Setting 8 didn't work
3. Full with not set too (see 4 : only 3 filed columns)

I add a situation where 16 work.

I update to put `[data-column-count]{column-count:attr(data-column-count)}`

application/core/QuestionTypes/ListRadio/RenderListRadio.php
ptelu: > > Example with 16:
> 
> Before you can not set 16 …
> 
> Even with 8 : it don't work Capture d’écran du 2023-04-14 11-36-02

off course it doesnt if the code code is transforming it into a equal 12 size column layout

ptelu: > 
Please stop twisting my sentences is said the "good thing" not that it is bad to allow 14. The good thing about the previous one is that it will prevent cutoff text if it limiting the size and preventing user error when playing around with css and html they dont understand.

assets/packages/limesurvey/survey.css
ptelu: > 

sure will discuss internally if there are better options for this specific usecase with many columns >12

Shnoulle:
ping @ptelu

assets/packages/limesurvey/survey.css
Shnoulle: > sure will discuss internally if there are better options for this specific usecase with many columns >12. And also how we can prevent user error reasonably without restricting to much.

It can break with big sentence inside one answer with 6 columns. But still visual issue only, no ?

application/core/QuestionTypes/ListRadio/RenderListRadio.php
```diff
@@ -218,20 +183,23 @@ public function render($sCoreClasses = '')
...
             'sRows'     => $this->getRows(),
             'name'      => $this->sSGQA,
             'basename'  => $this->sSGQA,
             'value'     => $this->mSessionValue,
             'coreClass' => $this->sCoreClass,
-            'othertext' => $this->sOthertext
+            'othertext' => $this->sOthertext,
+            'iNbCols' => $this->iNbCols,
+            'iCountAnswers' => $this->iCountAnswers,
+            'iMaxRowsByColumn' => $this->getAnswerCount() +1, // Leave it for old question theme compatibility, be sure to don't add columns
```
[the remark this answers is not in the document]
Shnoulle: Let me check, and i put comment when i fix conflict.

assets/packages/limesurvey/survey.css
Shnoulle: > im fine with leaving the suggested way with "data-column-count-17{column-count:17}", i would also limit the questions attribute to have a max value that will have the same limit as the highest column count used in the css. Also the dynamic call with attr() still doesnt work and is set as experimental with anything else then "content" https://developer.mozilla.org/en-US/docs/Web/CSS/attr?retiredLocale=de

I know it doesn't work : https://github.com/LimeSurvey/LimeSurvey/pull/3038/files#diff-02c3953154c744920a0bd8c8a9a1937be4a92d3c550a3391b9a3e09e7e4b2d9aR184

About limit in value , i'm really against it, 1 to 25 for example work.

If we allow to set 25 : user can just update custom.css on theme
If we disallow 25 : user must create a theme to allow it …

IMHO : allowing is always better if it don't broke security or accessibility. 

Maybe adding it in help ?

Shnoulle: > im fine with leaving the suggested way with "data-column-count-17{column-count:17}", i would also limit the questions attribute to have a max value that will have the same limit as the highest column count used in the css. Also the dynamic call with attr() still doesnt work and is set as experimental with anything else then "content" https://developer.mozilla.org/en-US/docs/Web/CSS/attr?retiredLocale=de

Current is limited to 12, but no max_value set. User set 13 and it don't work.
Updated the help

application/core/QuestionTypes/ListRadio/RenderListRadio.php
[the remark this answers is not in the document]
Shnoulle: `iMaxRowsByColumn` is not used anymore

in master

```
$ grep -r "iMaxRowsByColumn" *
core/QuestionTypes/ListRadio/RenderListRadio.php:    private $iMaxRowsByColumn;
core/QuestionTypes/ListRadio/RenderListRadio.php:            $this->iMaxRowsByColumn = ceil($this->getAnswerCount() / $this->iNbCols);
core/QuestionTypes/ListRadio/RenderListRadio.php:            $this->iMaxRowsByColumn = $this->getAnswerCount() + 3; // No max : anscount + no answer + other + 1 by security
core/QuestionTypes/ListRadio/RenderListRadio.php:            if ($this->iRowCount == $this->iMaxRowsByColumn) {
core/QuestionTypes/MultipleChoice/RenderMultipleChoice.php:    private $iMaxRowsByColumn;
core/QuestionTypes/MultipleChoice/RenderMultipleChoice.php:        $this->iMaxRowsByColumn = ceil($this->getQuestionCount() / $this->iNbCols);
core/QuestionTypes/MultipleChoice/RenderMultipleChoice.php:            'iMaxRowsByColumn' => $this->iMaxRowsByColumn,
helpers/qanda_helper.php:        $iMaxRowsByColumn = ceil($anscount / $iNbCols);
helpers/qanda_helper.php:        $iMaxRowsByColumn = $anscount + 3; // No max : anscount + no answer + other + 1 by security
helpers/qanda_helper.php:        if ($iRowCount == $iMaxRowsByColumn) {
helpers/qanda_helper.php:        if ($iRowCount == $iMaxRowsByColumn) {
helpers/qanda_helper.php:    $iMaxRowsByColumn = ceil($anscount / $iNbCols);
helpers/qanda_helper.php:        'iMaxRowsByColumn' => $iMaxRowsByColumn,
views/survey/questions/answer/multiplechoice/answer.twig:            {% if loop.index0 is divisible by(iMaxRowsByColumn) and loop.index0 != 0  %}
```

I can set it to $this->getAnswerCount() + 3; then be sure to don't apply to mutiple choce

papuche:
up :) still have the issue with image multiple choice and columns

Shnoulle:
> up :) still have the issue with image multiple choice and columns

Maybe if you can test the patch on your instance and tell us if it's better for you ?

ptelu:
@Shnoulle Hey sorry it took so long to get back to you on this. I tested this and as mentioned previously, the old and your new implementation have both a couple of drawbacks based on how the columns are visualized. I created small example PR based on this one https://github.com/LimeSurvey/LimeSurvey/pull/3894 
I had to merge in the master branch to be able to test it, the only commit i added was the last one.
It uses flex to dynamically render the columns.
Please let me know if you like it as well, then i will fully adjust it and remove unneeded parts and apply it to listradio as well. Currently, It's only on multiplechoice for demonstration.
Then i will merge my PR into yours.

Shnoulle:
Files changed 1,289 ???

I really can not check the difference here.

ptelu:
as in my previous comment the only commit that needs to be checked is the last one the rest is just from merging master

Shnoulle:
1. I don't like to add forced limit, pout the information in help.
2. We can do it with css class only : using a ls-column-count-1 to ls-column-count-16 and/or (boith is good too) data-column-count={{iNbCols}}

`<ul class="list-unstyled {{coreClass}} d-flex flex-wrap {% if iNbCols > 1 %} ls-column-count-{{iNbCols}}{% endif %}"{% if iNbCols > 1 %} data-column-count={{iNbCols}}{% endif %}>`

Shnoulle:
Reason to add only via css class, and survey.css file : more easy to update and use js or another solution in theme.

Shnoulle:
I update to allow different way : adding columlns style to allow 

A B
C D

or

A C
B D

Shnoulle:
I update to master, i create a for for fex test

Shnoulle:
Here the current test with columns css class

Screenshot 2024-06-24 at 09-14-13 List (Radio) questions columns not in one (ul) tag
Screenshot 2024-06-24 at 09-14-47 List (Radio) questions columns not in one (ul) tag
Screenshot 2024-06-24 at 09-19-40 List (Radio) questions columns not in one (ul) tag
[Uploading limesurvey_survey_multiColumn.lss.txt…]()

ptelu (approved):

Shnoulle:
Issues with image theme

themes/question/image_select-multiplechoice/survey/questions/answer/multiplechoice/answer.twig
```diff
@@ -13,7 +13,7 @@
 <!-- answer -->
 {# What is the usage of this input ? #}
 <input type="hidden" name="MULTI{{ name }}" value="{{ anscount }}" />
-<ul class="list-unstyled {{coreClass}}{% if iNbCols > 1 %} data-column-count-{{iNbCols}}{% endif %}" role="group" aria-labelledby="ls-question-text-{{basename}}"{% if iNbCols > 1 %} data-column-count={{iNbCols}}{% endif %}>
```
Shnoulle: Why o remove the group role ?

application/config/version.php
```diff
@@ -16,5 +16,5 @@
 $config['buildnumber'] = '';
 $config['updatable'] = true;
 $config['templateapiversion']  = 3;
-$config['assetsversionnumber'] = '30449';
+$config['assetsversionnumber'] = '30454';
```
Shnoulle: Maybe need one more for Git user tester ?

Shnoulle:
@tiborpacalat fixed issue about multple choice width/height :)

tiborpacalat:
Now the checkbox is not in the right position.

This branch:
Screenshot 2025-10-13 at 13 01 41

Master:
Screenshot 2025-10-13 at 13 04 16

Shnoulle:
@tiborpacalat can you test it again ?

PR #3040, opened 2023-04-07
Dev: remove the fix for MSSQL, included in YII 1.1.28

Shnoulle:
Dev: https://github.com/LimeSurvey/LimeSurvey/blob/f4233db80422ff9e40c8a6b7c79d96946f569d2a/vendor/yiisoft/yii/CHANGELOG#L8


Shnoulle:
Unsure we need to remove for 5.X , but can be.

PR #3042, opened 2023-04-11
Fixed issue #18737: Missing response action: View response as queXML PDF

adamzammit:
Fixed issue #18737: Missing response action: View response as queXML PDF


tiborpacalat:
Tested and verified the fix.

PR #3046, opened 2023-04-12
Fixed issue #18684: missing translation 

gabrieljenik:
- Added new command for making the translation bot to scan XML files.


gabrieljenik:
@olleharstedt I think a new command should be added to the build process, right?

PR #3049, opened 2023-04-12
Fixed issue #18679 API Access to question type array dual scale merge scales if answer code is not unique

gabrieljenik:
Extra testing routines for the API


Shnoulle (approved):

Shnoulle (approved):

gabrieljenik:
@olleharstedt New tests for the remote control get_question_properties method.

olleharstedt:
47% duplicated code? :thinking:

Shnoulle:
> > 47% duplicated code? thinking
> 
> There are tests for different question types which look a like

Yes, can be done using array (of settings and return string etc…), but i really think it's better like that too.

gabrieljenik:
> > > 47% duplicated code? thinking
> > 
> > 
> > There are tests for different question types which look a like
> 
> Yes, can be done using array (of settings and return string etc…), but i really think it's better like that too.

Yes, this way gives a bit more flexibility

Shnoulle:
> > > > 47% duplicated code? thinking
> > > 
> > > 
> > > There are tests for different question types which look a like
> > 
> > 
> > Yes, can be done using array (of settings and return string etc…), but i really think it's better like that too.
> 
> Yes, this way gives a bit more flexibility

And stay understandable …

```

                'AO01' => array(
                    'code' => 'AO01',
                    'answer' => 'Opción 1, escala 1',
                    'assessment_value' => '0',
                    'scale_id' => '0',
                    'order' => '0'
                    'checks' => array(
                        array(
                        'function' => 'assertEquals',
                        'waitValue' => '1',
                        'checkwith => 'type',
                        'labelError' => 'The question type is not correct.'
                        ),
                        array(
                            [etc]
                        )
                    )
                ),
```

😰

PR #3051, opened 2023-04-13
Dev: improve the response time of gh action

tiborpacalat:
Dev:
I rewrote the main.yml github action to run the tests as soon as possible and not wait on them at the end of the run. Now they are running in parallel as much as they can.

Note:
Currently some syntax checks are failing and I am not entirely sure if I missed something in this rewrite, so if I did please let me know.


olleharstedt:
Great, thank you!

PR #3056, opened 2023-04-18
Fixed issue #18191: Cannot use statCount in specific condition

Shnoulle:
Dev: add a surveyid checker : is set + survey exist + survey is active


Shnoulle:
I put Hard to test because : Hard to reproduce …

PR #3058, opened 2023-04-18
Fixed issue #18756: Unable to set non-integer values for initial slider value

Shnoulle:
Dev: parseInt to parseFloat


gabrieljenik (approved):
Code loks OK. Havent tested it

tiborpacalat:
Tested this in LS 5.x and 6.x and works as expected.

PR #3059, opened 2023-04-19
Task/cr 529 apply psalm to application all

olleharstedt:


application/controllers/admin/ParticipantsAction.php
```diff
@@ -2373,7 +2374,7 @@ public function rejectShareParticipant()
      * @param int $shareUid
      * @return void
      */
-    public function deleteSingleParticipantShare($participantId, $shareUid)
+    public function deleteSingleParticipantShare($participantId = null, $shareUid = null)
```
ptelu, line 2377: i think this might be a mistake, those 2 funtion params are needed. They are missing in the calling function, theres also a comment for it.

application/controllers/admin/Statistics.php
```diff
@@ -669,7 +669,7 @@ public function simpleStatistics($surveyid)
 
         if (!$oSurvey->isActive) {
             Yii::app()->setFlashMessage(gT("This survey is not active and has no responses."), 'error');
-            $this->getController()->redirect($this->getController()->createUrl("/surveyAdministration/view/surveyid/{$oSurveyid}"));
+            $this->getController()->redirect($this->getController()->createUrl("/surveyAdministration/view/surveyid/{$surveyid}"));
```
ptelu, line 672: this should probably use the sanitized variable iSurveyId

application/controllers/admin/SurveymenuController.php
```diff
@@ -190,7 +190,7 @@ public function massDelete()
                 $success[$menuid] = $model->delete();
             }
 
-            $debug = $userConfig['config']['debug'] ?? 0;
+            $debug = 0;
```
ptelu, line 193: should use the actual userconfig else some next parts of the code are pointless

application/controllers/admin/ParticipantsAction.php
olleharstedt: Maybe yeah. Question is, what's the proper fix? :scream:

application/controllers/admin/SurveymenuController.php
olleharstedt: Pja. `$debug = YII_DEBUG` might be enough?

application/controllers/admin/Statistics.php
olleharstedt: Agree

application/core/LS_Twig_Extension.php
```diff
@@ -617,11 +617,13 @@ public static function darkencss($cssColor, $grade = 10, $alpha = 1)
...
      * @param array $haystack The array.
      * @param bool $strict If the third parameter strict is set to TRUE then the in_array() function will also check the types of the needle in the haystack.
+     * @todo in_array_r is not defined - delete this method?
      */
     function in_multiarray($needle, $haystack, $strict = false)
     {
 
         foreach ($haystack as $item) {
+            /** @psalm-suppress UndefinedFunction */
             if (($strict ? $item === $needle : $item == $needle) || (is_array($item) && in_array_r($needle, $item, $strict))) {
```
ptelu, line 627: this is for sure meant to be in_array, would change it. The function also seeems completely unzused right now but some custom survey themes might use it

application/extensions/captchaExtended/CaptchaExtendedAction.php
```diff
@@ -477,6 +477,7 @@ protected function renderImage($code){
 		}
 
 		if($this->fontFile === null){
+            /** @psalm-suppress UndefinedFunction TODO: realname does not exist */
 			$this->fontFile = realname(Yii::app()->basePath."/../assets/fonts/font-src/lato-v11-latin-700.ttf");
```
ptelu, line 481: this probably needs to be realpath or dirname

application/helpers/admin/import_helper.php
```diff
@@ -2008,6 +2008,7 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
 
     // Import assessments --------------------------------------------------------
     if (isset($xml->assessments)) {
+        $aASIDReplacements = [];
```
ptelu, line 2011: i dont think this is a good idea it will always set the array to empty also the ocnditions are clearly expecting a value here

olleharstedt: Sure, but the important thing is that the array initialization happens _outside_ the loop, which id does. So should be ok?

application/controllers/admin/SurveymenuEntryController.php
```diff
@@ -280,7 +281,7 @@ public function massDelete()
                 $success[$menuEntryid] = $model->delete();
             }
 
-            $debug = $userConfig['config']['debug'] ?? 0;
+            $debug = 0;
```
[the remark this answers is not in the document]
olleharstedt: Sure, but it's not _worse_ than before. :sweat_smile:

application/controllers/admin/ParticipantsAction.php
olleharstedt: Wanna move this out to a separate bug?

olleharstedt:
I'll merge this and create a new task for the found issues. Most of them seem really old or even dead code.

application/controllers/admin/ParticipantsAction.php
olleharstedt: Added to https://bugs.limesurvey.org/view.php?id=18850

application/core/LS_Twig_Extension.php
olleharstedt: Added to https://bugs.limesurvey.org/view.php?id=18850

application/controllers/admin/ParticipantsAction.php
Shnoulle: > Maybe yeah. Question is, what's the proper fix? scream

Move participantId to POST value not param.

olleharstedt: Yeah, good point @Shnoulle

PR #3060, opened 2023-04-19
Fixed issue #18315 Uses left value goes to negative

gabrieljenik:
Label message for answer persistence option is set dynamically.

### DIFF
--- a/application/views/admin/survey/subview/accordion/_tokens_panel.php
+++ b/application/views/admin/survey/subview/accordion/_tokens_panel.php
@@ -45,6 +45,61 @@ App()->getClientScript()->registerScript("tokens-panel-variables",
     
 ",
     LSYii_ClientScript::POS_BEGIN);
+
+App()->getClientScript()->registerScript("edit-after-completion-message", "
+    (function(){
+        let showInherited = " . $bShowInherited . ";
+        let inheritedAnonymizedOption = '" . $oSurvey->oOptions->anonymized . "';
+        let inheritedPersistenceOption = '" . $oSurvey->oOptions->tokenanswerspersistence . "'
+        let multipleResponsesText = '" . gT('Allow multiple responses with the same access code') . "';
+        let updateResponsesText = '" . gT('Allow to update the responses using the access code') . "';
+
+        $(document).ready(function(){
+            let persistenceOption = $('input[name=\"tokenanswerspersistence\"]:checked').val();
+            let anonymizedOption = $('input[name=\"anonymized\"]:checked').val();
+            
+            changeAllowEditLabel(anonymizedOption, persistenceOption);
+            console.log('Selected on document ready.');
+            
+            $('input[name=\"anonymized\"]').change(function(){
+
+                let anonymizedOption = $(this).val();
+                let persistenceOption = $('input[name=\"tokenanswerspersistence\"]:checked').val();
+
+                changeAllowEditLabel(anonymizedOption, persistenceOption);
+                console.log('Selected on anonymized option changed.');
+    
+            });
+
+            $('input[name=\"tokenanswerspersistence\"]').change(function(){
+                let persistenceOption = $(this).val();
+                let anonymizedOption = $('input[name=\"anonymized\"]:checked').val();
+                
+                changeAllowEditLabel(anonymizedOption, persistenceOption);
+                console.log('Selected on token persistence option changed');
+        
+            });
+        });
+
+        function changeAllowEditLabel( anonymizedOption, persistenceOption )
+        {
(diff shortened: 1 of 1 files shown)


gabrieljenik:
@Shnoulle Can you review?

gabrieljenik:
Maybe instead of this label
`Allow multiple responses with the same access code` 
we should say `Allow unlimited responses with the same access code`
?

application/views/admin/survey/subview/accordion/_tokens_panel.php
```diff
@@ -45,6 +45,61 @@ function alertPrivacy()
     
 ",
     LSYii_ClientScript::POS_BEGIN);
+
+App()->getClientScript()->registerScript("edit-after-completion-message", "
```
Shnoulle, line 49: No way to add the sentence in HTML directly inside block and just shown/hide good blick (div or p) according to the current settings.

More easy to move to extrenal JS after.
Stgart by show a string for both and dynamicall hide/show 

```
            <div class="mb-3">
                <label class=" form-label" for='alloweditaftercompletion' title='<?php  eT("If participant-based response persistence is enabled a participant can update his response after completion, otherwise a participant can add new responses without restriction."); ?>'>
                    <?php  eT("Allow multiple responses or update responses with one access code:"); ?>
                </label>
                <div>
                    <?php $this->widget('ext.ButtonGroupWidget.ButtonGroupWidget', [
                        'name'          => 'alloweditaftercompletion',
                        'checkedOption' => $oSurvey->alloweditaftercompletion,
                        'selectOptions' => ($bShowInherited)
                            ? array_merge($optionsOnOff, ['I' => $oSurveyOptions->alloweditaftercompletion . " ᴵ"])
                            : $optionsOnOff
                    ]); ?>
                </div>
                <div id="alloweditaftercompletion-base" class="form-text"><?php eT('Allow multiple responses or update the responses using the access code'); ?></div>
                <div id="alloweditaftercompletion-update" class="form-text hidden"><?php eT('Allow to update the responses using the access code'); ?></div>
                <div id="alloweditaftercompletion-multiple" class="form-text hidden"><?php eT('Allow multiple responses with the same access code'); ?></div>
            </div>
```

application/views/admin/survey/subview/accordion/_tokens_panel.php
```diff
@@ -106,6 +161,7 @@ function alertPrivacy()
                             : $optionsOnOff
                     ]); ?>
                 </div>
+                <span id="multiResponseHint" class="hint" style="display:none;"><?php eT("Participants will be able to enter as many responses as they want, despite what Uses Left token attribute is set to."); ?></span>
```
Shnoulle, line 164: https://getbootstrap.com/docs/5.3/forms/overview/ 

`<div id="multiResponseHint" class="form-text hidden">We'll never share your email with anyone else.</div>`

Shnoulle (commented):
1. Add help sentence in content directly
2. use bs5 class name.

application/views/admin/survey/subview/accordion/_tokens_panel.php
gabrieljenik: > No way to add the sentence in HTML directly inside block and just shown/hide good blick (div or p) according to the current settings ?

If that's the case, you can just move the block code to extenral JS and leave the language definition here.

application/views/admin/survey/subview/accordion/_tokens_panel.php
gabrieljenik: If we do that, we wouldn't be able to use `toggle()` inside the code.
Also, that code is a snippet taken from Global Settings

application/views/admin/survey/subview/accordion/_tokens_panel.php
Shnoulle: ? In my proposed solution : JS do JS , no language …

In your proposed solution you create var for language string… you can not leave the language definition here …

application/views/admin/survey/subview/accordion/_tokens_panel.php
Shnoulle: You mean about hidden ? 

1. Did you test
2. OK for hidden but `<div class="form-text`

gabrieljenik: > OK for hidden but <div class="form-text

I took it from Global Settings screen

Shnoulle: And ? 

Global settings have dev issue : must be reported as dev issue

Here you ADD new element.

Shnoulle (approved):
Thank you

Shnoulle (approved):

PR #3061, opened 2023-04-19
Fixed issue #18765: Answers are not populated correct in array dual scale after clicking "Back"

gabrieljenik:


application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
```diff
@@ -294,7 +294,8 @@ public function parseSubquestionsNoDropdown(&$aData)
 
                 if ($aData['shownoanswer']) {
                     // No answer for accessibility and no javascript (but hide hide even with no js: need reworking)
-                    if (empty($this->getFromSurveySession($myfname0))) {
+                    $fname0value = $this->getFromSurveySession($myfname0);
+                    if (!isset($fname0value) || $fname0value == '') {
```
Shnoulle, line 298: `$fname0value === ''` (code can be 0)

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
```diff
@@ -317,15 +318,17 @@ public function parseSubquestionsNoDropdown(&$aData)
             $aData['answertextright'] = $answertextright;
             if ($aData['shownoanswer']) {
                 if (count($aData['labelans1']) > 0) {
-                    if (empty($this->getFromSurveySession($myfname1))) {
+                    $fname1value = $this->getFromSurveySession($myfname1);
+                    if (!isset($fname1value) || $fname1value == '') {
```
Shnoulle, line 322: `$fname1value === ''` (code can be 0)

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
```diff
@@ -317,15 +318,17 @@ public function parseSubquestionsNoDropdown(&$aData)
...
                         #$answer .= CHECKED;
                         $aData['aSubQuestions'][$i]['myfname1_notset'] = CHECKED;
                     } else {
                         $aData['aSubQuestions'][$i]['myfname1_notset'] = "";
                     }
                 } else {
-                    if (empty($this->getFromSurveySession($myfname0))) {
-                        $answer .= CHECKED;
+                    $fname0value = $this->getFromSurveySession($myfname0);
+                    if (!isset($fname0value) || $fname0value == '') {
```
Shnoulle, line 330: `$fname0value === ''` (code can be 0)

Shnoulle (changes requested):
`=== ''` 

0 == '' == false

Shnoulle:
Right 0 != '0'
But still risky :)

Shnoulle:
OK, loose compare string with string

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
Shnoulle: Tets is OK because $fname0value is a string (then with "0" it's OK).

But maybe best to use `===` more clear

Shnoulle: ping @olleharstedt ?

Microfix not needed to fix the issue, but i think it's more clear and clean.

gabrieljenik: We will review the fix and apply something a bit more clear.

olleharstedt: There's also `empty()` if you wanna be explicit about looking for all possible "falsy" values in PHP.

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
```diff
@@ -295,7 +295,7 @@ public function parseSubquestionsNoDropdown(&$aData)
                 if ($aData['shownoanswer']) {
                     // No answer for accessibility and no javascript (but hide hide even with no js: need reworking)
                     $fname0value = $this->getFromSurveySession($myfname0);
-                    if (!isset($fname0value) || $fname0value == '') {
+                    if ($fname0value !== '0' && empty($fname0value)) {
```
Shnoulle, line 298: I really think it's less clear … why not `(!isset($fname0value) || $fname0value === '')` or better `(is_null($fname0value) || $fname0value === '') {` 

There are only 2 value for no answer : NULL : don't see it or '' empty string : see but don't answer.

Shnoulle (commented):
Only 2 value for no answer null or ===''

Usage of is_empty make it complex to understand. And clearly : we need to check this 2 values.

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
```diff
@@ -317,15 +318,17 @@ public function parseSubquestionsNoDropdown(&$aData)
             $aData['answertextright'] = $answertextright;
             if ($aData['shownoanswer']) {
                 if (count($aData['labelans1']) > 0) {
-                    if (empty($this->getFromSurveySession($myfname1))) {
+                    $fname1value = $this->getFromSurveySession($myfname1);
+                    if ($fname1value !== '0' && empty($fname1value)) {
```
Shnoulle, line 322: The previous code was easier to understand …

gabrieljenik: Why? This means: not-set  = anaything falsy but different to "0" string

gabrieljenik: That was exactly the case which triggered the error.
Actually, I think this fix is better as changes less.

olleharstedt: Ehm, the previous line is easier to read,  yes, but it's not correct, is it? :sweat_smile:

gabrieljenik: > The previous code was easier to understand …

Denis means this code: `if (!isset($fname0value) || $fname0value == '') {`
https://github.com/LimeSurvey/LimeSurvey/pull/3061/commits/bd25e53309895520fa8a9b9c0bf94e1009ba0633#r1183785808

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
Shnoulle: OK, comment replace

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
Shnoulle: Yes,

Comment replace the code.

Shnoulle (approved):

application/core/QuestionTypes/ArrayMultiscale/RenderArrayMultiscale.php
olleharstedt: Ah ok. Well, no biggie, really, I don't think we need to waste time here. :)

Shnoulle: I already resolve :)

PR #3066, opened 2023-04-21
Fixed issue #18703: Upload files show error if not seen (using index)

Shnoulle:
Dev: same fix than master and 3.X


tiborpacalat:
Works as intended.

PR #3067, opened 2023-04-21
Fixed issue #18456: Interface language parmater use exotic dropdown list.

gabrieljenik:


application/views/admin/survey/subview/accordion/_generaloptions_panel.php
```diff
@@ -319,7 +319,7 @@ function checkSelect2Languages(mylangs) {
         <div class="mb-3" >
             <label class=" form-label" for='template'><?php eT("Theme:"); ?></label>
             <div class="">
-                <select id='template' class="form-select" name='template' data-updateurl='<?php echo App()->createUrl('themeOptions/getPreviewTag') ?>'
+                <select id='template' style="width:100%;" class="form-select activate-search" name='template' data-updateurl='<?php echo App()->createUrl('themeOptions/getPreviewTag') ?>'
```
Shnoulle, line 322: style is needed ?

Shnoulle (approved):
No issue in code

This fix the autofocus ?

gabrieljenik:
> This fix the autofocus ?

Select2 in more dropdowns

application/views/admin/survey/subview/accordion/_generaloptions_panel.php
gabrieljenik: Yes, as to no have the dropdown trimmed

Shnoulle: There are no form-XX class for this … :(

Shnoulle:
> > This fix the autofocus ?
> 
> Select2 in more dropdowns

Yes, but there are a autofocus issue in select2 currently see https://github.com/select2/select2/pull/6044

Shnoulle:
> > > > This fix the autofocus ?
> > > 
> > > 
> > > Select2 in more dropdowns
> > 
> > 
> > Yes, but there are a autofocus issue in select2 currently see [select2/select2#6044](https://github.com/select2/select2/pull/6044)
> 
> Ohh you are right, you mean we should update the select2 to this version? [select2/select2#6044](https://github.com/select2/select2/pull/6044)

Don't like to merge locally on vendor … :(

gabrieljenik:
> Don't like to merge locally on vendor … :(

What do you mean? Sorry, got lost

Shnoulle:
> > Don't like to merge locally on vendor … :(
> 
> What do you mean? Sorry, got lost

You give a link to fix on a vendor repository ;) i don't like apply fix in vendor directory …

Else : i still wait for the A11Y version of select2 … :( https://bugs.limesurvey.org/view.php?id=17813 
But don't want to move to a BETA version (and i have one issue with one of my plugins with this version).

gabrieljenik:
Not sure if we are on the same page.
I got the link from the 18456 ticket. That link was posted by you.
https://bugs.limesurvey.org/view.php?id=18456#c72570

Anyway, that is still on develop and not part their stable release, right?
So, maybe just better to go with this as Carsten suggested.
```
      $(document).on('select2:open', () => {
        document.querySelector('.select2-search__field').focus();
      });
```

From https://stackoverflow.com/a/68030102/1392917`

Thoughts?

Shnoulle:
> Not sure if we are on the same page. I got the link from the 18456 ticket. That link was posted by you. https://bugs.limesurvey.org/view.php?id=18456#c72570
> 
> Anyway, that is still on develop and not part their stable release, right? 

Yes, it's the issue :)

> So, maybe just better to go with this as Carsten suggested.
> 
> ```
>       $(document).on('select2:open', () => {
>         document.querySelector('.select2-search__field').focus();
>       });
> ```
> 
> From [https://stackoverflow.com/a/68030102/1392917`](https://stackoverflow.com/a/68030102/1392917%60)
> 
> Thoughts?

If it work great, why not :)

PR #3071, opened 2023-04-24
initial attempt to fix double ci runs

tiborpacalat:
Dev:
Fixing double github workflow runs (pull_request and pull)

Note: this is WIP

Before the fix:
Screenshot 2023-04-24 at 12 43 19


tiborpacalat:
If we want to have push and pull_request the runs will be double.

PR #3080, opened 2023-04-26
dev: printanswers should show decrypted responses (all types)

Trischi80:
Fixed issue : bug reported by customer. encrypted responses question should be shown decrypted when using printanswer after submitting a survey


c-schmitz (approved):

application/models/SurveyDynamic.php
```diff
@@ -1006,7 +1006,17 @@ public function getPrintAnswersArray($sSRID, $sLanguage, $bHonorConditions = fal
 
         $oSurvey = self::$survey;
         $aGroupArray = array();
+
         $oResponses = SurveyDynamic::model($oSurvey->sid)->findByAttributes(array('id' => $sSRID));
+
```
Shnoulle, line 1011: Why not 
```
if ($oResponses) {
     $oResponses->decrypt();
}
```

It decrypt whole and used  a lot :)

application/models/SurveyDynamic.php
```diff
@@ -1001,21 +1001,30 @@ public function getQuestionArray($oQuestion, $oResponses, $bHonorConditions, $su
...
+    {
+        //get response values which are encrypted
+        $encryptedAttr = Response::getEncryptedAttributes($this->getSurveyId());
+        $attributes = $this->attributes;
+        $sodium = Yii::app()->sodium;
+        foreach ($encryptedAttr as $key) {
+            $this->setAttribute($key, $sodium->decrypt($attributes[$key]));
+        }
+    }
+
```
Shnoulle, line 1019: https://github.com/LimeSurvey/LimeSurvey/blob/23d4342755a31b5729f3e9802987e3a647d1d58a/application/models/LSActiveRecord.php#L468

PR #3082, opened 2023-04-26
Fixed issue # Fixed issue #18639: Token browse can be unusable with lot of data in attribute

Shnoulle:
Dev: longtext formatter + scss
Dev: TODO css


Shnoulle:
@ptelu : how to construct the css via the scss ?

Or where ad this scss fix ?

Shnoulle:
Thank you,

Seems `grid-view-ls` was removed between 5.X to 6.X, i add it again.

OK or need another class ?

ptelu:
i think its fine yes, we removed it mostly because everything was based on the table class. But i agree that gridviews should have a distinct name. We didn't get to moving the remnants remnants out of adminbasics_temporary. They have their own component folder in the theme "assets/admin_themes/Sea_Green/tables". I would rename/readd the grid class to "ls-gridview" then it would fit with our custom naming as well as with the controller.

For this fix i think it is fine if you add the class again, we can add another task to rename/readd and move the code to the right folder.

tiborpacalat:
I have tested this and it works now as expected.
Screenshot 2023-04-27 at 18 00 22

Shnoulle:
> I have tested this and it works now as expected.

Did you see : seems BS5 have fixed the issue with tooltip too.

themes/admin/Sea_Green/css/sea_green.css
```diff
@@ -14543,6 +14543,11 @@ html[dir=rtl] th {
   overflow-x: visible;
 }
 
+/** Hide Relevance Equation column for subquestions when not supported */
```
Shnoulle, line 14546: ? Didn't added this in my commit … :(

PR #3088, opened 2023-04-27
Bug/spss export wrong wording

Trischi80:
Fixed issue : update the syntax to "RENAME VARIABLES" in the SPSS export functionality to align with the correct SPSS syntax


Trischi80:
something went wrong with creating this PR

PR #3093, opened 2023-04-28
New feature: Update the style of the admin file manager

jack-han-115:
Dev:


assets/packages/adminbasics/src/adminbasicsmain.js
```diff
@@ -175,4 +180,4 @@ const AdminCore = function(){
         onLoadRegister();
 }
 
-AdminCore();
+AdminCore();
```
olleharstedt, line 183: Not sure I understand why this change is needed for file manager?

assets/packages/adminbasics/src/components/file-manager.js
```diff
@@ -0,0 +1,201 @@
...
+    {
+        originIcon: "a[href='kcact:mvdir'] span",
+        newIcon: "ri-eraser-line",
+    },
+    {
+        originIcon: "a[href='kcact:rmdir'] span",
+        newIcon: "ri-delete-bin-line",
+    },
+
+    // folder regular
```
olleharstedt, line 59: Hm what does this mean? :)

assets/packages/adminbasics/src/components/file-manager.js
```diff
@@ -0,0 +1,201 @@
...
+
+            // Load sea_green css again after iframe is fully loaded
+            handleAppendCssLink({
+                header: fileManagerIframe.contentWindow.document.head,
+                linkUrl: "/themes/admin/Sea_Green/css/sea_green.css",
+            });
+            handleAppendCssLink({
+                header: fileManagerIframe.contentWindow.document.head,
+                linkUrl:
+                    "http://ls-ce/assets/fonts/font-src/remix/remixicon.css",
```
olleharstedt, line 197: Hardcoded link to what?

assets/packages/adminbasics/src/components/file-manager.js
```diff
@@ -0,0 +1,201 @@
...
+    },
+    {
+        originIcon: "a[href='kcact:rmdir'] span",
+        newIcon: "ri-delete-bin-line",
+    },
+
+    // folder regular
+];
+
+const handleReplaceIcons = ({ iframeSource, originIcon, newIcon }) => {
```
olleharstedt, line 62: All these functions are missing docblocks. Is that by design? What's our current policy?

assets/packages/adminbasics/src/adminbasicsmain.js
jack-han-115: we are loading fileManager in the iframe.
the custom styles are loaded first and the iframe is loaded after that.
so styles are not applied.
so, i added file-manager.js in the adminbasicsmain.js and load seagreen css and remix icons style after the iframe is loaded fully.

assets/packages/adminbasics/src/components/file-manager.js
olleharstedt: Resolved but still hardcoded? :) What's the intended behaviour?

assets/admin_themes/Sea_Green/fileManager/fileManager.scss
```diff
@@ -0,0 +1,308 @@
...
+      display: none;
+    }
+    .about div.head a {
+      background: none;
+    }
+  }
+
+  #menu {
+    .ui-menu {
+      box-shadow: none !important;
```
olleharstedt, line 70: Lots of `important` going on. Is that needed even when the selector is made more precise than the previous "default" CSS selectors? Dunno which is best.

PR #3094, opened 2023-04-28
Bug/ct 129 decrypt responses before output 5.x

Trischi80:
Fixed issue : implemented new additional function for decrypted output (response values)


Shnoulle:
Why  not use already included function ?
Used on all other controller and helpers .

function : https://github.com/LimeSurvey/LimeSurvey/blob/77a234a70fbca29ce65180e895cde85b85ebdbf8/application/models/LSActiveRecord.php#L356

Usage
https://github.com/LimeSurvey/LimeSurvey/blob/77a234a70fbca29ce65180e895cde85b85ebdbf8/application/helpers/expressions/em_manager_helper.php#L5165
https://github.com/LimeSurvey/LimeSurvey/blob/77a234a70fbca29ce65180e895cde85b85ebdbf8/application/helpers/admin/export/Writer.php#L313
https://github.com/LimeSurvey/LimeSurvey/blob/77a234a70fbca29ce65180e895cde85b85ebdbf8/application/helpers/export_helper.php#L133

etc …

Trischi80:
Hi Deni, i tried together with Patrick. There seems to be a bug when using this function you mentioned. We could give it try later, but now this solution (which is working) should do it, as a fix. Thanks for your hints.

Shnoulle:
Maybe issue with usage of `$LEM = LimeExpressionManager::singleton();` must be used only for ExpressionManager

Not for statitistics and export and other ?

Trischi80:
> Maybe issue with usage of `$LEM = LimeExpressionManager::singleton();` must be used only for ExpressionManager
> 
> Not for statitistics and export and other ?

Yes, i agree. I will create a ticket in our internal bug system. At some point someone should take a look at it. The decription function should be intependent from ExpressionManager.

PR #3117, opened 2023-05-08
Fixed issue #18765: Answers are not populated correct in array dual scale after clicking "Back"

gabrieljenik:


Shnoulle (approved):
Already reviewed for master

PR #3120, opened 2023-05-09
Fixed issue #18808: Mandatory soft only works once per survey if over multiple pages/groups

Shnoulle:
Dev: not used currently, to be used for feature 18332


gabrieljenik (approved):
Code looks OK. Seems it was added not long ago for mandSoft.

Shnoulle:
> Code looks OK. Seems it was added not long ago for mandSoft.

Yes … my fault in last commit … think of the other fix to do.

tiborpacalat:
Tested and verified the fix.

PR #3122, opened 2023-05-09
Fixed issue #18670: "Validation of quota" action button always shows 1st quota

gabrieljenik:


Shnoulle (approved):
:+1: 

OK : each is needed

PR #3129, opened 2023-05-10
Fixed issue #18811: Button that triggers HTML editor in pop-up window doesn't work

Shnoulle:
Dev: Review questionEditor script
Dev: placeholder without url encoded character

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -683,7 +683,7 @@ class QuestionAdministrationController extends LSBaseController
         }
         $html  = [];
         $first = true;
-        $qid   = 'new' . rand(0, 99999);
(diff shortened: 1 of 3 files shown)


Shnoulle:
OK … …

Tested on subquestin + answers

1. Direct
2. Add one
3. Add label set
4. Quick add
5. Replace

tiborpacalat:
There is just a small styling issue that I noted in Mantis.

Shnoulle:
@ptelu i update the view for button, the reason i ask a review from you.

Take https://getbootstrap.com/docs/5.3/forms/input-group/#button-addons 
commit


I try some extra class too
input-group-text
bg-info
bg-primary

About input-group-text : i think it not the best : it's seems like the other addon without action. There are no visual difference between action and no-action.

ptelu (approved):
looks good!

PR #3131, opened 2023-05-11
clear all old bootstrap files

jack-han-115:
Dev:


application/views/admin/dataentry/vvimport.php
```diff
@@ -158,18 +158,3 @@
         <?php } ?>
 
 </div></div></div>
-<?php
-App()->getClientScript()->registerScript('VVImportBSSwitcher', "
-LS.renderBootstrapSwitch();
-
-$('#noid').on('switchChange.bootstrapSwitch', function(event, state) {
```
olleharstedt, line 165: Is this correct, to remove logic like this without replacing it with something?
What's the proper test for this piece of code?

application/views/admin/token/invite.php
```diff
@@ -58,10 +58,10 @@
                             <div class='mb-3'>
                                 <?php echo CHtml::label(
                                     gT("Bypass date control before sending email:"),
-                                    'bypassdatecontrol',
```
olleharstedt, line 61: There are way too many clean-up changes like this, makes it harder to review the PR. Please in the future, do clean-up in separate tasks instead of mixing it up with other. Might also be an automatic setting in your editor you need to check.

application/views/admin/token/massive_actions/_update.php
```diff
@@ -422,14 +422,6 @@ class="YesNoDatePicker form-control action_datepickerUpdateHiddenField"
 
 <?php App()->getClientScript()->registerScript("Tokens:MassActionUpdateView_Scripts", "
 
-   var bindBSSwitch = function(formGroup){
-       console.log(\"bindBSSwitch run on:\",formGroup);
-       //Script to update the completed settings
-       formGroup.find('.YesNoSwitch').on('switchChange.bootstrapSwitch', function(e, state){
```
olleharstedt, line 428: Another piece of logic we need to test?

assets/scripts/admin/users.js
```diff
@@ -38,10 +38,6 @@ function bindButtons(){
     $('.action_usercontrol_button').on('click', function(){
         runAction(this);
     });
-    $('input[name="alltemplates"]').on('switchChange.bootstrapSwitch', function(event, state) {
```
olleharstedt, line 41: Need testing instructions, I think.

application/views/admin/dataentry/vvimport.php
jack-han-115: Added this functionality.

application/views/admin/token/invite.php
jack-han-115: ok, this was due to my VS code setting,  i will disable.
image

assets/scripts/admin/users.js
jack-han-115: deleted this as didn't find any files using this js file.
will revert the change.

application/views/admin/token/invite.php
jack-han-115: i reverted file format changes.

application/views/admin/token/massive_actions/_update.php
jack-han-115: my mistake.
This should not be removed. :(

olleharstedt: Alright, that's great, but I still might need testing instructions. :D For Tibor at least.

application/extensions/admin/grid/MassiveActionsWidget/assets/listActions.js
```diff
@@ -278,6 +278,7 @@ var onClickListAction =  function () {
  *
  */
 
+// i think below two functions are not used and can be deleted
```
olleharstedt, line 281: Maybe add a `TODO` annotation so it can be searched easier later?

application/views/admin/dataentry/vvimport.php
olleharstedt: Great, I will unresolve this so that Tibur can find it later.

assets/admin_themes/adminbasics_temporary/000-lime-admin-common.scss
```diff
@@ -2054,10 +2054,6 @@ table#responsedetail tr td {
     margin-bottom: 30px;
 }
 
-#insertmethod-container {
```
olleharstedt, line 2057: Hm is this a cleanup?

olleharstedt:
Hm looks like the resolve conflicts commit broke the CI?

olleharstedt:
Reverted, please have another look.

PR #3132, opened 2023-05-11
Dev: Generate simple statistics test.

gabrieljenik:
Test added.


olleharstedt:
Status here?

gabrieljenik:
> Status here?

This is for simple stats. This is ready to go. Needs reviewal.
Would need to review if more question types could be added to this tests

olleharstedt:
Well, something is better than nothing. :) Thanks!

PR #3138, opened 2023-05-15
Fixed issue #18262: QCode.gid is not working anymore

Shnoulle:
Dev: lose it when moving to static and dynamic var
Dev: adding jsName_on for JS solution ?


Shnoulle:
Oups … i don't close it before ?

This one make some reorganization  in order of fixed var.

PR #3139, opened 2023-05-15
Fixed issue #18670: "Validation of quota" action button always shows 1st quota

gabrieljenik:


Shnoulle (approved):
We don't already have another fix somewhere ?

tiborpacalat:
> We don't already have another fix somewhere ?

Yes it is here for 5.x already reviewed by you https://github.com/LimeSurvey/LimeSurvey/pull/3122/files

Shnoulle (approved):

PR #3141, opened 2023-05-15
Fixed issue #18814: Array question statistics fail for Remote Control.

gabrieljenik:
Dev: Remote Control export statistics tests.
Fixed issue #18814: Array question statistics fail for Remote Control.


application/helpers/common_helper.php
```diff
@@ -1215,7 +1215,7 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
 
                 //go through all the (multiple) answers
                 foreach ($result as $row) {
-                    $myfield2 = $myfield . reset($row);
+                    $myfield2 = $myfield . $row['title'];
```
Shnoulle, line 1218: Gloups ? 
Why it don't break for GUI statistics here ? Not used for GUI ?

gabrieljenik: That method is only used on RC.
Seems there is a bit of code duplication and the GUI part was working fine :)

Shnoulle:
OK, no problem in code (lot of auto test)

PR #3144, opened 2023-05-17
Fixed issue #18818: dump when using getQuestions API

gabrieljenik:
- The $language parameter is not removed to avoid breaking plugins that are already working on version 5.


Shnoulle:
Reason why i don't use API …

Before : `$this->getApi()->getQuestions($qid)->question`  return the text of quetsion,
After your fix `$this->getApi()->getQuestions($qid)->question` break and return error.

Need to create a Class ?

Shnoulle (commented):
This fix the issue, 
But if a plugin use api->getQuestions($qid, $langauge) maye it use `question` and `help`.

OK, API update then usage is broken, but removing the function is more clear.

gabrieljenik:
ok, I forgot the comment.
I am not intending to make the function backward compatible.
I am just trying to avoid the dump.

I didn't break anything.

Follow discussion on mantis

gabrieljenik:
> But if dev do

Yes, in that situation will dump. Object Model had a non-backward compatible update which propagated through the PluginAPI.

That requires:
- Update to the PluginAPI
- Update to the code that uses the PluginAPI

We tackled the first task here.

Right now, PluginAPI is dumping no matter what.
With the fix, PluginAPI is not dumping, still code that uses it should be adapted, which is some how un-avoidable.

> Another solution :

I guess that is removing features from the API.
Right now the API provides a way to return a question object.
What's the upside on removing it? 

> Here you still create a function that broke in 7 or 8 etc … no interest, is still don't use it.

Here we are just bugfixing. 
Design decisions can be split on a different ticket.

c-schmitz:
I think the function was barely used in LS5, otherwise this would have shown up alot earlier.
As the object does not exist in the same form as in LS3 we cannot make it backward compatible without creating a new backward-compatible class extending the current Question class, but that seems to be too much effort.

The proper way would be to deprecate the function, and create a new one with a similar name for LS5. 

However, nobody seems to care, so I'd say @gabrieljenik instead use the existing function but at least use the language to filter down the objoct querts to return the requested language only and  if language is null, return all languages, for each question entry.

gabrieljenik:
>  I'd say @gabrieljenik instead use the existing function but at least use the language to filter down the objoct querts to return the requested language only and if language is null, return all languages, for each question entry.

I am sorry I don't understand. How would I use the language as a filter for question objects if language is not part of question class anymore? Got lost there. 

Look forward for the clarification.

Thanks

Shnoulle:
> I think the function was barely used in LS5, otherwise this would have shown up alot earlier.

Exacly why i think we must remove or return null directly. Don't try to create a «this function are good to be used» and broke at next update

Maybe return as array (merging attributes) ? Then : it can be backward compatible in 7,8,9 etc …

olleharstedt:
Has unit-test now? So can be merged?

olleharstedt:
Oh right, exactly :joy:

PR #3148, opened 2023-05-17
Fixed issue #18823: STATA Export fails with certain question types

Oz-men:
Fixed issue #18823: Stata export failed with question type "radio"

### DIFF
--- a/application/core/plugins/ExportSTATAxml/STATAxmlWriter.php
+++ b/application/core/plugins/ExportSTATAxml/STATAxmlWriter.php
@@ -401,7 +401,7 @@ class STATAxmlWriter extends Writer
                             // For radio lists, user wants code, not label
                             // TODO: We could skip this loop if we had answer code
                             foreach ($this->customFieldmap['answers'][$iQID][$iScaleID] as $answer) {
-                                if (is_array($answer) && isset($answer['answer']) && $answer['answer'] == $response) {
+                                if (isset($answer['answer']) && $answer['answer'] == $response) {
                                     $response = $answer['code'];
                                     break;
                                 }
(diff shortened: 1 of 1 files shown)


application/core/plugins/ExportSTATAxml/STATAxmlWriter.php
```diff
@@ -401,7 +401,7 @@ protected function updateCustomresponsemap()
                             // For radio lists, user wants code, not label
                             // TODO: We could skip this loop if we had answer code
                             foreach ($this->customFieldmap['answers'][$iQID][$iScaleID] as $answer) {
-                                if ($answer['answer'] == $response) {
+                                if (is_array($answer) && isset($answer['answer']) && $answer['answer'] == $response) {
```
Shnoulle, line 404: `if (isset($answer['answer']) && $answer['answer'] == $response) {`

Check if it's an array not needed, just check if it exist.

Shnoulle (changes requested):
Check if it's an array not needed

Shnoulle:
PS: thank you :)

application/core/plugins/ExportSTATAxml/STATAxmlWriter.php
Oz-men: Thank you! As requested I removed the array check.

Shnoulle (approved):
:+1:

gabrieljenik:
I have tested this. A dump is not produced :)
Still, I can't evaluate the results of the file.
By the type of change here, I guess it is OK to merge?

Thgoughts @Shnoulle ? @olleharstedt ?

Shnoulle:
> Still, not sure if this is relevant: https://bugs.limesurvey.org/view.php?id=18823#c75167

I think you're right about workaround, but search original issue take more time a lot more time …

Someone report another issue for a lack of answers or something like this in STATA export, then maybe we can fix it.

Here : reporter have issue and create a PR : i really think we must merge !

PR #3153, opened 2023-05-18
Update timer.js

BertHankes:
https://bugs.limesurvey.org/view.php?id=18784
timer.js backword compatible with bootstrap 5 and lower classes: 'd-done hidden'

BTW. I think timer.min.js needs to be minfied from this timer.js. I think the timer.min.js is 2 years old already!

Fixed issue #18784:


ptelu:
closing this since everything will be solved in https://github.com/LimeSurvey/LimeSurvey/pull/3151

PR #3155, opened 2023-05-18
Fixed issue #18343: Problem with double slash in an URL(IIS)

Shnoulle:
Dev: use intval
Dev: most simple fix to avoid more commit (see mantis [18811](https://bugs.limesurvey.org/view.php?id=18811))


gabrieljenik (approved):

gabrieljenik:
Wathc out: some checks are not green

PR #3156, opened 2023-05-19
Fixed issue #18731: Edit a question "script" tab has disappeared

Shnoulle:
Dev: use default in question editor


gabrieljenik (approved):
Code seems OK. Haven't tested it.

tiborpacalat:
Tested and verified the fix.

PR #3164, opened 2023-05-23
Fixed issue: error and save pages were loading twig files of wrong su…

twilligls:
- Error page was initing Template model without any hint of actual survey the error was coming from
- "Resume later" page was unnecessarily initing Template + passing wrong templatename

### DIFF
--- a/application/controllers/SurveysController.php
+++ b/application/controllers/SurveysController.php
@@ -86,7 +86,7 @@ class SurveysController extends LSYii_Controller
             // TODO: Remove? It seems this can never happen because it's already caught by LSYii_Application::onException() (see commit c792c2e).
             $this->spitOutJsonError($error, $oException);
         } elseif ($error) {
-            $this->spitOutHtmlError($error, $oException);
+            $this->spitOutHtmlError($error, $request->getParam('sid'));
         } else {
             throw new CHttpException(404, 'Page not found.');
         }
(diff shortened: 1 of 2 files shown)


application/controllers/SurveysController.php
```diff
@@ -86,7 +86,7 @@ public function actionError()
             // TODO: Remove? It seems this can never happen because it's already caught by LSYii_Application::onException() (see commit c792c2e).
             $this->spitOutJsonError($error, $oException);
         } elseif ($error) {
-            $this->spitOutHtmlError($error, $oException);
+            $this->spitOutHtmlError($error, $request->getParam('sid'));
```
Shnoulle, line 89: :+1: GREAT ! 🎉

Shnoulle: `$request->getParam('sid', $request->getParam('surveyid') ) `just in case ?

application/controllers/SurveysController.php
```diff
@@ -106,10 +106,13 @@ public function actionError()
...
      */
-    public function spitOutHtmlError(array $error, $oException = null)
+    public function spitOutHtmlError(array $error, $surveyId)
     {
-        // TODO: getGlobalSetting is DEPRECATED.
-        $oTemplate = Template::model()->getInstance(getGlobalSetting('defaulttheme'));
+        if ($surveyId) {
+            $oTemplate = Template::model()->getInstance('', $surveyId);
+        } else {
+            $oTemplate = Template::getLastInstance();
```
Shnoulle, line 114: If template is  not already set : we have the default template ?

We don't need `$oTemplate = Template::model()->getInstance(App()->getConfig('defaulttheme'));` ?

Shnoulle: Great !
Then : a plugin can set to another instance before !!
This is really great !

PR #3175, opened 2023-05-26
Fixed issue #18468: User with rights to view other users can not see them at the survey permission screen

gabrieljenik:


application/models/services/SurveyPermissions.php
```diff
@@ -415,4 +447,46 @@ public function getUserIdsWithSurveyPermissions(): array
...
+        return $this->filteredUserIdList;
+    }
+
+    /**
+     * Returns true if the current user can manage survey permissions
+     * for the given user (ie. can add/remove permissions for the given user).
+     * @param int $userId
+     * @return bool
+     */
+    public function canManageSurveyPermissionsForUser($userId)
```
Shnoulle, line 465: This really need a clean manual page here :)

application/models/services/SurveyPermissions.php
```diff
@@ -328,6 +356,10 @@ public function saveUserGroupPermissions($userGroupId, $permissions)
      */
     public function deleteUserPermissions($userId)
     {
+        if (!$this->canManageSurveyPermissionsForUser($userId)) {
+            throw new \Exception(gT("No permission to delete survey permissions from user."));
```
Shnoulle, line 360: 403 error ?

application/models/services/SurveyPermissions.php
```diff
@@ -253,6 +277,10 @@ public function saveUserPermissions($userId, $permissions)
     {
         $isSaved = true;
 
+        if (!$this->canManageSurveyPermissionsForUser($userId)) {
+            throw new \Exception(gT("No permission"));
```
Shnoulle, line 281: 403 http error ?

Shnoulle (approved):
Seems OK,

We need same fix for group ? 
https://github.com/LimeSurvey/LimeSurvey/blob/01df49d78be9220154ef654a0247670b278cc087/application/models/services/SurveyPermissions.php#L148

Checking if user can see all groups or not ?

gabrieljenik:
> Seems OK,
> 
> We need same fix for group ?
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/01df49d78be9220154ef654a0247670b278cc087/application/models/services/SurveyPermissions.php#L148
> 
> Checking if user can see all groups or not ?

Created 18842: Assign Survey Permissions: What groups should someone see?

application/models/services/SurveyPermissions.php
```diff
@@ -183,12 +197,14 @@ public function getSurveyUserList()
...
 
         if ($this->userControlSameGroupPolicy) {
-            $authorizedUsersList = getUserList('onlyuidarray');
+            $authorizedUsersList = $this->getFilteredUserIdList();
         }
         $index = 0;
         foreach ($users as $user) {
             if (
-                !$this->userControlSameGroupPolicy || in_array($user->uid, $authorizedUsersList)
+                $hasUserReadPermission ||
```
Shnoulle, line 205: 204 | ERROR | [x] Boolean operators between conditions must be at the beginning or end of the line, but not both

;)

olleharstedt:
Please wait with code-review until CI is green (ignore SonarCloud tho).

olleharstedt:
> application/models/services/SurveyPermissions.php:10  ExcessiveClassComplexity  The class SurveyPermissions has an overall complexity of 56 which is very high. The configured complexity threshold is 50.

The method `getUserGroupNames` is not called by any other method in the class, and it also does not use `this`, and seems semantically kinda unrelated since it does not deal with permissions. Moving it out would reduce the complexity of the class slightly.

olleharstedt:
What about also moving out `getTooltipAllPermissions` to another class? It's kinda related to presentation, and the only `this` access is survey id, which can be given as an argument, or constructor in a new class.

gabrieljenik:
Created https://bugs.limesurvey.org/view.php?id=18875

application/models/services/SurveyPermissions.php
```diff
@@ -59,7 +66,9 @@ public function getUserPermissionCriteria()
         $criteria->condition = 't.entity_id =:entity_id';
         $criteria->addCondition('t.entity =:entity');
         $criteria->addNotInCondition('u.uid', [$currentUserId]);
-        $criteria->addInCondition('t.uid', $userList);
+        if (!$hasUserReadPermission && $this->userControlSameGroupPolicy) {
+            $criteria->addInCondition('t.uid', $userList);
```
olleharstedt, line 70: Psalm warning on this line?

application/models/services/SurveyPermissions.php
```diff
@@ -415,4 +454,50 @@ public function getUserIdsWithSurveyPermissions(): array
...
+    /**
+     * Returns the list of user ids that can be used in the user selector.
+     * @return int[]|null
+     */
+    private function getFilteredUserIdList()
+    {
+        if (!isset($this->filteredUserIdList)) {
+            $this->filteredUserIdList = getUserList('onlyuidarray');
+        }
+        return $this->filteredUserIdList;
```
olleharstedt, line 467: Maybe you want to return the empty list here instead of null? Using `??` for example.

Shnoulle: return https://github.com/LimeSurvey/LimeSurvey/blob/de071e01d299b41cc79cc4067f06760ef351da31/application/helpers/common_helper.php#L664 in all case.

Then array.

application/models/services/SurveyPermissions.php
Shnoulle: Came from https://github.com/LimeSurvey/LimeSurvey/blob/de071e01d299b41cc79cc4067f06760ef351da31/application/helpers/common_helper.php#L666 

No user entered value, and return only id.

What is the alert ?

application/models/services/SurveyPermissions.php
```diff
@@ -173,6 +182,14 @@ public function addUserGroupToSurveyPermissions($userGroupId)
      */
     public function getSurveyUserList()
     {
+        $hasUserReadPermission = \Permission::model()->hasGlobalPermission('users', 'read');
+
+        // If the user doesn't have User Read permission, and it's not allowed
+        // to see at least the users of the same group, then return an empty list
+        if (!$hasUserReadPermission && !$this->userControlSameGroupPolicy) {
```
Shnoulle, line 189: It's not the proposed solution on https://bugs.limesurvey.org/view.php?id=18468#c75188

It must show all users (like in current 3.X)

And see : https://bugs.limesurvey.org/view.php?id=18468#c75192

gabrieljenik: > It must show all users (like in current 3.X)

Actually, that's the part we said it had to be corrected.
Right now it can see al users. But if no permission, it should not see any user.
Row 6 and 8 of the table in the ticket
image

Shnoulle: 1. In 3. X : you can allow any user to give permission to any other user without access to User list (with email and login).
It's not possible after this fix.

2. In 3.X : you can allow an user to see only some user in User Management : it's not possible after this update.

gabrieljenik: Can we pick this up as to wrap this PR?
Can you please describe a bit more the situations 1?

As per number 2, I believe we are not updating anything from user managemnt screen.

olleharstedt: @Shnoulle Ping for reminder :)

Shnoulle: > @Shnoulle Ping for reminder :)

I report issue and feature : 
- https://bugs.limesurvey.org/view.php?id=18921
- https://bugs.limesurvey.org/view.php?id=18916

My global opinion :  we need more detail on needed behavior …

I check in 3.X (and 5.X): an user without Permission on User see ALL user in Survey permission (no group management used)

Shnoulle: We need a list with all option possible and for each what is the desired behavior and what is the current.

If current was updated : it's a major API version update : can only be in LimeSurvey 7 (in my opinion)

gabrieljenik: > We need a list with all option possible and for each what is the desired behavior and what is the current

We did this already. right?
https://bugs.limesurvey.org/view.php?id=18468#c75188

That's where we detected the bug.

Shnoulle: Yes, but there are 2 issues with this list

1. We don't separate Seeing user in user management and in List : did we need to do it or not ? Maybe have a settings in security, at same place than _Group member can only see own group:_  something like _User need user permission to see user in list_ (current situation : No) : apply to line with _Empty list_ in the table
2. Where you set Expected to _All users_ : currently this user didn't see anybody, after update they see all user: **It can be a big issue for some LimeSurvey user** , maybe you show a client to another client.
3. The Expected to _All users_ is totally with this commit : https://github.com/LimeSurvey/LimeSurvey/commit/4824bc93f08ac4e504a2464ecf0616fffe38a0f5 where action is done ONLY if user is parent_id or have superadmin permission


The point 3 is really important here. If you set Permission View all : you see ALL user, then if you set update ALL : it seems logical to have the right to modify ALL users

And sorry : at start of this issue : i don't know for the point 3 …

gabrieljenik: > We don't separate Seeing user in user management and in List

As far as I know the tables only concerns "giving permission to users within a survey".
I understand as well the impact of this PR only affects that process as well.

>  Maybe have a settings in security ...

I suggest to first agree on the scope of this PR.
As I understand, this is only about "giving permission to users within a survey".
If so, that setting still holds? needed? 

> Where you set Expected to All users : currently this user didn't see anybody, after update they see all user

That was part of the bug identified. If he has permission to see "all users", isn't it expected to see all users in the dropdown?

> The Expected to All users is totally at odds with this commit

That commit is about rules for delete user. This is for adding usres to a survey.
Not getting how they relate.

> The point 3 is really important here. If you set Permission View all : you see ALL user, then if you set update ALL : it seems logical to have the right to modify ALL users

Is the `update ALL` permission used here?

----

I would need someone else view here. Maybe I am missign something

Shnoulle: > As far as I know the tables only concerns "giving permission to users within a survey".

Yes : and then i currently have a lot of big system where _Group member can only see own group:_ is OFF to allow user to give permission to any user.

And at contrario i have client where we activate _Group member can only see own group:_ to disallow this. My client create group for each of this client.


> That was part of the bug identified. If he has permission to see "all users", isn't it expected to see all users in the dropdown?

Yes, but it's **not** the case today. I don't have this situation, but if a LimeSurvey user have this situation : you show something not shown before.
I don't know such situation, but it can exist.

> That commit is about rules for delete user. This is for adding usres to a survey.
> Not getting how they relate.

With current situation :

- Permission see user : see ALL user in all condition
- Permission delete user : delete SOME user
- Permission update user : update SOME user

I think what is expected must be

- Permission see user : see ALL user in all condition
- Permission delete user : delete ALL user
- Permission update user : update ALL user

OR 

- Permission see user : see SOME user in all condition
- Permission delete user : delete SOME user (same the  see)
- Permission update user : update SOME user (same the  see)

gabrieljenik: As per the first 2 items you mentioned, I will let someone else to jump in and comment.

As per the last item, I believe that would be a different ticket and when tackled, we should update this dropdown list.
It is more about incongruencies in access in between list, delete and update permissions.
Not specifically related to the scope of this ticket.

gabrieljenik:
Lot of tests added :)

application/models/services/SurveyPermissions.php
gabrieljenik: @tiborpacalat @olleharstedt @c-schmitz thoughts?

PR #3176, opened 2023-05-30
Fixed issue #18845: Question theme of type 'S' (extending short free text) does not display correct preview image

adamzammit:
Fixed issue #18845: Question theme of type 'S' (extending short free text) does not display correct preview image


tiborpacalat:
On current 5.x the behaviour is the same as on 6.x, which is correct imo.

PR #3177, opened 2023-05-30
Fixed issue #18845: Question theme of type 'S' (extending short free text) does not display correct preview image

adamzammit:
Fixed issue #18845: Question theme of type 'S' (extending short free text) does not display correct preview image

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -2973,14 +2973,6 @@ class QuestionAdministrationController extends LSBaseController
                     'questionGroupName' => $questionTheme->group
                 );
             }
-            $imageName = $questionTheme->question_type;
-            if ($imageName == ":") {
-                $imageName = "COLON";
-            } elseif ($imageName == "|") {
-                $imageName = "PIPE";
-            } elseif ($imageName == "*") {
-                $imageName = "EQUATION";
-            }
             $questionThemeData = [];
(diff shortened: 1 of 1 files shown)


application/controllers/QuestionAdministrationController.php
```diff
@@ -2973,14 +2973,6 @@ private function getQuestionTypeGroups(array $questionThemes)
                     'questionGroupName' => $questionTheme->group
                 );
             }
-            $imageName = $questionTheme->question_type;
```
Shnoulle, line 2976: Why remove this part ?

adamzammit: Only used in this function and only if == 'S' so seems redundant to assign a separate variable

Shnoulle:
Just comment/discussion

Maybe best is to remove the 2 images system, have exception are always a bad idea if we can avoid (proof by this issue).

[S2](https://github.com/LimeSurvey/LimeSurvey/blob/master/assets/images/screenshots/S2.png) are not really needed (it must be a theme)
We don't show Slider for [K ](https://github.com/LimeSurvey/LimeSurvey/blob/master/assets/images/screenshots/K.png) for example, or checkbox for [COLON](https://github.com/LimeSurvey/LimeSurvey/blob/master/assets/images/screenshots/COLON.png)

Alternate solution : merge S and S2 with a separator

c-schmitz:
Closing this patch since it is obsolete.

PR #3183, opened 2023-06-01
Fixed issue #18856 - buildXMLFromQuery fails to implement LIMIT and can end…

TonisOrmisson:
… up throwing an error

Fixed issue #18856: buildXMLFromQuery fails to implement LIMIT

### DIFF
--- a/application/helpers/export_helper.php
+++ b/application/helpers/export_helper.php
@@ -760,7 +760,15 @@ function buildXMLFromQuery($xmlwriter, $Query, $tagname = '', $excludes = array(
                 $result[] = $row->decrypt()->attributes;
             }
         } else {
-            $QueryResult = Yii::app()->db->createCommand($Query)->limit($iChunkSize, $iStart)->query();
+            /** @var CDbConnection $db */
+            $db = Yii::app()->db;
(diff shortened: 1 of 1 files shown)


application/helpers/export_helper.php
```diff
@@ -760,7 +760,15 @@ function buildXMLFromQuery($xmlwriter, $Query, $tagname = '', $excludes = array(
                 $result[] = $row->decrypt()->attributes;
             }
         } else {
-            $QueryResult = Yii::app()->db->createCommand($Query)->limit($iChunkSize, $iStart)->query();
+            /** @var CDbConnection $db */
+            $db = Yii::app()->db;
+            if(is_string($Query)) {
```
gabrieljenik, line 765: Isn't $Query expected to be a string every time?

TonisOrmisson: no, mostly seems to be an array, but is sting in some cases. The ->limit() method only works while query is an array, see docs here:
https://www.yiiframework.com/doc/api/1.1/CDbCommand#__construct-detail
image

if query is a string - calling ->limit() has no effect (hence the pr).

PR #3119, opened 2023-05-09
Fixed issue #18808: Mandatory soft only works once per survey if over multiple pages/groups

Shnoulle:

PR #3195, opened 2023-06-07
Fixed issue #18808: Mandatory soft only works once per survey if over multiple pages/groups

Shnoulle:
Fixed issue #18808: Mandatory soft only works once per survey if over multiple pages/groups
Dev: with auto test


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/3119

Unsure we still need test and review ?

olleharstedt:
> [](https://github.com/LimeSurvey/LimeSurvey/pull/3119)
> 
> Unsure we still need test and review ?

Nah it's probably fine.

PR #3210, opened 2023-06-09
 Fixed issue [security] #18866: Label sets can be replaced by any admin user

Shnoulle:
Fixed issue [security] #18866: Label sets can be replaced by any admin user
Fixed issue [security]#18868: No CRSF control for delete action of label set 
Dev: check global right
Dev: use getPost for lid and send error if CRSF is deactivated


Shnoulle:
Except 5.X : all gren

Shnoulle:
Code review https://github.com/LimeSurvey/LimeSurvey/pull/3211#issuecomment-1584357393

PR #3211, opened 2023-06-09
Fixed issue [security] #18866: Label sets can be created by any admin user

Shnoulle:
Fixed issue [security] #18868: No CRSF control for action of label set
Dev: add global permission check
Dev: get lid by POST (process action came from label set edition)


olleharstedt:
Code looks good.

PR #3212, opened 2023-06-09
Fixed issue [security] #18866: Label sets can be created by any admin user

Shnoulle:
Fixed issue [security] #18868: No CRSF control for action of label set
Dev: add global permission check
Dev: get lid by POST (process action came from label set edition)


Shnoulle:
For tester : remind to test different delete action and update action : must still work (if you have right).

I test some, but maybe not whole. 2 ways for deletion (i labelset, and in list) , 2 way for updating (in labelset and in Question)

Shnoulle:
Code review : https://github.com/LimeSurvey/LimeSurvey/pull/3211#issuecomment-1584357393

PR #3215, opened 2023-06-12
Fixed issue #18871: Can't add email attachment in email templates

gabrieljenik:


Shnoulle (approved):
Don't see issue, but not tested (i mean : didn't check if it can fix the issue looking at code)

PR #3216, opened 2023-06-12
Fixed issue #18368: export_statistics from API Yields error 

gabrieljenik:
Sanitization of the path was removing / changing characters in the filename and making the file to be not found.


application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -573,16 +573,20 @@ public function export_statistics($sSessionKey, $iSurveyID, $docType = 'pdf', $s
         $aSummary = createCompleteSGQA($iSurveyID, $oAllQuestions, $sLanguage);
 
         $helper = new statistics_helper();
+        $tempDir = Yii::app()->getConfig("tempdir");
         switch ($docType) {
             case 'pdf':
-                $sTempFile = sanitize_filename(
-                    $helper->generate_statistics($iSurveyID, $aSummary, $aSummary, $graph, $docType, 'F', $sLanguage)
+                $sTempFile = $tempDir . DIRECTORY_SEPARATOR . sanitize_filename(
```
Shnoulle, line 579: why sanitize_filename since [helper->generate_statistics ](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/admin/statistics_helper.php#L3796) already set sanitized filename. No reason ro sanitize a generated filename, no user entered.

If  needed : need to be done in helpe.

I think best is to remove sanitizer.

And then : no need extra var here
https://github.com/LimeSurvey/LimeSurvey/blob/de071e01d299b41cc79cc4067f06760ef351da31/application/helpers/admin/statistics_helper.php#L3884
https://github.com/LimeSurvey/LimeSurvey/blob/de071e01d299b41cc79cc4067f06760ef351da31/application/helpers/admin/statistics_helper.php#L4079

Shnoulle (commented):
No reason to sanitize, it's the original issue.

file name are generated, didn't come from user input.

application/helpers/remotecontrol/remotecontrol_handle.php
Shnoulle: Sonar_cloud make error. We really don'(t need to sanitize this var.

c-schmitz: Well, remove it and check it Sonarcloud thinks it is a security issue.

Shnoulle: We don't have way to add comments for sonar cloud ? Here :

1. it's clearly NOT a security issue
2. The fix apply for sonar_cloud create bug
3. The fix make the system more complex

gabrieljenik: Seems soundcloud is ok with this?
https://sonarcloud.io/project/security_hotspots?id=LimeSurvey_LimeSurvey&pullRequest=3216

Shnoulle: :+1:

tests/unit/helpers/remotecontrol/ExportStatisticsTest.php
```diff
@@ -202,6 +202,28 @@ public function testSurveyWithThreeQuestions()
         $this->assertSame($q1Data[6][2], '100.00%', 'The Percentage is incorrect for this option.');
     }
 
+    /**
+     * Testing that exports_statistics returns
+     * statistics in xls format.
+     */
+    public function testXlsStatistics()
```
Shnoulle, line 209: :+1:

olleharstedt:
> No reason to sanitize, it's the original issue.
> 
> file name are generated, didn't come from user input.

I wouldn't be that confident without first checking the tainted code path as reported by SonarQube.

PR #3217, opened 2023-06-13
Qe 250 survey update services

kevin-foster-uk:
QE-250 Refactor Survey update controller logic into service classes


application/controllers/admin/Database.php
```diff
@@ -277,571 +224,95 @@ private function actionUpdateDefaultValues($iSurveyID)
...
-
-    /**
-     * action to do when update survey seetings + survey language
+     * Action to run when update survey settings + survey language
+     *
+     * Refactored to use Services\SurveyUpdater 2023-05-30 (kfoster).
+     *
      * @param integer $iSurveyID
+     * @param ?array $input For dependency injection during testing
      * @return void (redirect)
```
olleharstedt, line 233: Could actually be the `never` return type here :)

olleharstedt: Ah wait you didn't write that line

application/controllers/admin/Database.php
```diff
@@ -277,571 +224,95 @@ private function actionUpdateDefaultValues($iSurveyID)
...
 
-        //@todo  here is something wrong ...
-        $oSurvey = Survey::model()->findByPk($iSurveyID);
-        $languagelist = $oSurvey->additionalLanguages;
-        $languagelist[] = $oSurvey->language;
-
-        Yii::app()->loadHelper('database');
-
-        $hasSurveyLanguageSettingError = false;
-        if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveylocale', 'update')) {
```
olleharstedt, line 581: Permission checks should still be in the controller, IMO, and not service classes.

kevin-foster-uk: Logic around who can and can not perform certain actions is business logic.

Controllers need to handle permission failure in its own specific way which is why the service throws a permission failure exception.

olleharstedt: Lemme think tho

olleharstedt: What if you wanna use that service class from a CLI command, where no permissions should be checked at all?

olleharstedt: Well you could inject a dummy permission checker, I guess.

application/controllers/admin/Database.php
```diff
@@ -277,571 +224,95 @@ private function actionUpdateDefaultValues($iSurveyID)
...
-                }
-                $aURLParam['sid'] = $iSurveyID;
 
-                $param = new SurveyURLParameter();
-                foreach ($aURLParam as $k => $v) {
-                                    $param->$k = $v;
-                }
-                $param->save();
-            }
+        $meta = [];
```
olleharstedt, line 289: What does `meta` mean? :)

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -0,0 +1,570 @@
...
+ * Dependencies are injected to enable mocking.
+ * @SuppressWarnings(PHPMD.ExcessiveClassComplexity)
+ */
+class GeneralSettings
+{
+    private Permission $modelPermission;
+    private Survey $modelSurvey;
+    private LSYii_Application $yiiApp;
+    private SessionData $sessionData;
+    private PluginManager $yiiPluginManager;
```
olleharstedt, line 32: I would not include `yii` here, since the plugin manager event system is actually _not_ based on the Yii events.

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -0,0 +1,570 @@
...
+                'surveysettings',
+                'update'
+            );
+        if (!$hasPermission) {
+            throw new ExceptionPermissionDenied(
+                'Permission denied'
+            );
+        }
+
+        $survey = $this->modelSurvey->findByPk(
```
olleharstedt, line 86: If this is the only usage of $modelSurvey we might as well inject the $survey at call site instead?

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -0,0 +1,570 @@
...
+                $survey,
+                $initAttributes['language']
+            );
+
+            $this->dispatchPluginEventBeforeSurveySettingsSave(
+                $survey
+            );
+
+            if (!$survey->save()) {
+                throw new ExceptionPersistError(
```
olleharstedt, line 148: Nice :+1:

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -0,0 +1,570 @@
...
+class GeneralSettings
+{
+    private Permission $modelPermission;
+    private Survey $modelSurvey;
+    private LSYii_Application $yiiApp;
+    private SessionData $sessionData;
+    private PluginManager $yiiPluginManager;
+    private LanguageConsistency $languageConsistency;
+
+    const FIELD_TYPE_YN = 'yersno';
```
olleharstedt, line 35: Is that correct, `yersno`? Yer?

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -0,0 +1,570 @@
...
+            $settingsEvent->set('survey', $surveyId);
+            $this->yiiPluginManager
+                ->dispatchEvent($settingsEvent, $plugin);
+        }
+    }
+
+    /**
+     * Dispatch plugin event before survey settings save
+     *
+     * @param int $surveyId
```
olleharstedt, line 434: params are wrong here, weird that Psalm didn't react. :/

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -0,0 +1,570 @@
...
+    const FIELD_TYPE_DATETIME = 'dateime';
+    const FIELD_TYPE_GAKEY = 'gakey';
+    const FIELD_TYPE_USE_CAPTCHA = 'use_captcha';
+
+    const GA_GLOBAL_KEY = '9999useGlobal9999';
+
+    public function __construct(
+        Permission $modelPermission,
+        Survey $modelSurvey,
+        LSYii_Application $yiiApp,
```
olleharstedt, line 45: Or just `$app`? Only used to load libraries, so maybe just load them before instead? Unless that's "for later", but then we might want to write them down somehow, either as "todo" or a task. Or both.

application/libraries/SessionData.php
```diff
@@ -0,0 +1,67 @@
...
+
+namespace LimeSurvey;
+
+use CHttpSession;
+use ArrayAccess;
+
+/**
+ * Session Data
+ *
+ * Simple array access wrapper of CHttpSession to allow
```
olleharstedt, line 11: Can you explain again why this is needed instead of just using CHttpSession? Purism reason or pragmatic? :))

olleharstedt:
Everything I did not comment on looks good :) No way to approve single files, it seems.

application/models/services/SurveyUpdater.php
```diff
@@ -0,0 +1,84 @@
...
+     * @return array
+     */
+    public function update($surveyId, $input)
+    {
+        $this->languageSettings->update(
+            $surveyId,
+            $input
+        );
+
+        $meta = $this->generalSettings->update(
```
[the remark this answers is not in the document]
kevin-foster-uk: If you look at the original methods from database controller "actionUpdateSurveyLocaleSettingsGeneralSettings" and "actionUpdateSurveyLocaleSettings" they both update some survey object settings and also language specific settings in a single request. I refactored this into a SurveyUpdated::update(). Only data that you pass in is actually updated. If you don't pass in language specific settings we don't attempt yo update anything language specific.

The focus is refactoring existing functionality for re-usability and at the same time making the code testable. This may not be the best way to organise the code but generally improving code organisation was not the focus of task. 

The more "issues" we try to fix in one go, the more code we change, the higher the chance of us breaking something. 

Screenshot 2023-06-27 at 12 01 45

application/controllers/admin/Database.php
kevin-foster-uk: Meta data. The method returns data about what it did. Specifically the fields that it updated.

application/models/services/SurveyUpdater/GeneralSettings.php
kevin-foster-uk: You are suggesting that the permission check and the loading of the survey should happen in the controller rather than in the service. In this case you could do that but I dont think its a good idea. We only then have to repeat this code each time we call this method. I dont see any gain from doing that. 

Having this logic in the service allows us to to define it once and test that it works correctly once. Both the permission check and the not found exception are already covered by unit tests.

application/models/services/SurveyUpdater/GeneralSettings.php
kevin-foster-uk: Yeah this is a mistake. I didnt use the yii prefix for plugin manager elsewhere.

application/models/services/SurveyUpdater/GeneralSettings.php
kevin-foster-uk: Should be 'yesorno'. Fixed.

application/models/services/SurveyUpdater/GeneralSettings.php
kevin-foster-uk: Load them before where?

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -0,0 +1,570 @@
...
+            : [];
+
+        $hasPermission = $this->modelPermission
+            ->hasSurveyPermission(
+                $surveyId,
+                'surveysettings',
+                'update'
+            );
+        if (!$hasPermission) {
+            throw new ExceptionPermissionDenied(
```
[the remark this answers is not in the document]
kevin-foster-uk: If we were to try and follow PHP code naming scheme we would be guaranteed a mess. PHP code naming is all over the place.

Using a prefix to group classes of the same type allows them to be more easily identified when they are open in your editor. 

You can see this pattern throughout the code I write:

application/models/services/SurveyUpdater/GeneralSettings.php
kevin-foster-uk: Fixed.

application/models/services/SurveyUpdater/LanguageConsistency.php
```diff
@@ -0,0 +1,137 @@
...
+     * Update
+     *
+     * @param Survey $survey
+     * @param string $initBaseLanguage
+     * @param array $aAvailableLanguages
+     * @return void
+     */
+    private function cleanup(Survey $survey, $initBaseLanguage, $aAvailableLanguages)
+    {
+        fixLanguageConsistency(
```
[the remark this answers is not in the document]
kevin-foster-uk: Bad how? Because its a global function?

Again this is copy and paste. This code is not currently triggered by any unit tests so I made no attempt to deal with it.

application/libraries/SessionData.php
kevin-foster-uk: I cant mock in the data in CHttpSession.

application/models/services/SurveyUpdater.php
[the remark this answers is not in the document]
olleharstedt: Alright all good :)

application/controllers/admin/Database.php
olleharstedt: OK perhaps write a comment about that in the code?

application/models/services/SurveyUpdater/GeneralSettings.php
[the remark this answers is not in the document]
olleharstedt: Both Yii and PHP put "Exception" at the end of the class name. Please follow existing naming schemas instead of making your own. ;)

application/models/services/SurveyUpdater/GeneralSettings.php
olleharstedt: I was thinking in terms of functional core, imperative shell, and that it's easier to unit test service classes the fewer things you have to mock. So instead of doing

    $generalSettings->update($surveyId, $input)

you'd do

    $generalSettings->update(Survey::model()->findByPk($surveyId), $input);

You'd get a nullable type tho, and would still have to check for empty inside update(), probably throw a InvalidArgumentException or whatever.

This is what _composability_ (of functions) is all about. :)

olleharstedt: https://en.wikipedia.org/wiki/Composability

https://en.wikipedia.org/wiki/Function_composition_(computer_science)

application/models/services/SurveyUpdater/GeneralSettings.php
olleharstedt: Load them before the GeneralSettings object is constructed?

Meh, lack of autoloading for functions in PHP is a PITA...

application/libraries/SessionData.php
olleharstedt: Why not? Would you just do

```php
$mockSession = $this
  ->createMock(CHttpSession::class)
  ->method('getX')
  ->willReturn($yourData);
```
Or similar?

application/models/services/SurveyUpdater/LanguageConsistency.php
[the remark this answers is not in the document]
olleharstedt: Yeah I know. Can you write a todo, that fixLanguageConsistency should be lifted into this class at some point? Then just resolve this comment.

application/models/services/SurveyUpdater/GeneralSettings.php
olleharstedt: Btw, I'm assuming controller actions will only be tested with functional (scripting the browser) testing, not unit tests. At least in our current architecture.

application/models/services/SurveyUpdater/GeneralSettings.php
olleharstedt: The permission model is used in two different places, so that one makes less sense to try to move out, I think.

application/models/services/SurveyUpdater/GeneralSettings.php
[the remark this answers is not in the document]
olleharstedt: Perhaps more generally, if Yii, PHP and PSR all use the same naming scheme, we should follow it. When they diverge, the coder can pick one (or we agree upon one for LS and write it down in the code guideline).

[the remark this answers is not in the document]
kevin-foster-uk: Do you want to define rules for how we arrange and name classes in project? That would help avoid confusion and disagreements going forward.

application/libraries/SessionData.php
kevin-foster-uk: If it was that simple I would have done that. I tried to figure out a way to mock it and it looked to be super messy so I went with a simpler approach.

application/models/services/SurveyUpdater/LanguageConsistency.php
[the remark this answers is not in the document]
kevin-foster-uk: There are a bunch on functions like this littered all over the code.

application/models/services/SurveyUpdater/GeneralSettings.php
olleharstedt: The gain is one less injection.

application/models/services/SurveyUpdater/GeneralSettings.php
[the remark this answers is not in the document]
olleharstedt: Yeah, we didn't add new files and classes for a long while.

Some small notes here: https://manual.limesurvey.org/Code_quality_guide#Class_name

application/models/services/SurveyUpdater/GeneralSettings.php
kevin-foster-uk: One less injection in the constructor in return for code duplication in the controller and the REST command.

application/libraries/SessionData.php
olleharstedt: Could you explain why you couldn't do it like that? I don't get it.

application/models/services/SurveyUpdater/GeneralSettings.php
olleharstedt: The gain is to have one less injection and to not pollute service classes with "administrative code". The cost is some code duplication, but that's only due to lack of autoloading of functions in PHP.

At least you could import the needed libs in the constructor, no need to assign app to a class property in this case.

application/models/services/SurveyUpdater/GeneralSettings.php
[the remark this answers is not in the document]
olleharstedt: @kevin-foster-uk Are you happy with the current formulation in the manual section linked above?

application/models/services/SurveyUpdater/LanguageConsistency.php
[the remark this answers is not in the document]
olleharstedt: Of course, but if we add TODOs in the service classes we can later grep for them for further cleanups. :)

application/models/services/SurveyUpdater/GeneralSettings.php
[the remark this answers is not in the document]
kevin-foster-uk: @olleharstedt Yes I am ok with the addition of Exception as a class name suffix.

application/models/services/SurveyUpdater/GeneralSettings.php
olleharstedt: Alright, I will resolve this then.

application/libraries/SessionData.php
olleharstedt: Cool cool, will resolve this.

PR #3218, opened 2023-06-13
Fixed issue #18793: Quick Translation crash with MS SQL Database Error

gabrieljenik:


Shnoulle (approved):
Without default scope : what the final order ?

Why adding t.code in order ?

Just question : the fix is clean

gabrieljenik:
> Without default scope : what the final order ?

It is on the ticket.
`ORDER BY group_order, question.question_order, t.scale_id, t.sortorder, sortorder, code`

You can see sortorder is repeated.

> Why adding t.code in order ?

That was part of the default scope, which was removed. 
So we bring it back.

Shnoulle:
> It is on the ticket.

Oh : only  answers defaults cope  https://github.com/LimeSurvey/LimeSurvey/blob/1337418e4a02e3a8be83f0740912e89a39dec0c5/application/models/Answer.php#LL104C24-L104C41

I just check : we don't have defaultScope for Question and QuestionGroup … strange … i think we added it before ?

PR #3221, opened 2023-06-14

gabrieljenik:


application/helpers/sanitize_helper.php
```diff
@@ -512,5 +512,5 @@ function check_ip_address($ip)
  */
 function sanitize_alphanumeric($value)
 {
-    return preg_replace("[^a-zA-Z0-9\-\_]", "", $value);
+    return preg_replace("/[^a-zA-Z0-9\-\_]/", "", $value);
```
Shnoulle, line 515: The previous one have an issue ? It was used elsewhere ?

Shnoulle (approved):

application/helpers/sanitize_helper.php
gabrieljenik: > The previous one have an issue ? 

Yes, it didn't sanitized. Just returned the same string as inputed.

> It was used elsewhere ?

Just in 2 other places from a quick search

PR #3222, opened 2023-06-14

gabrieljenik:


application/helpers/sanitize_helper.php
```diff
@@ -520,5 +520,5 @@ function check_ip_address($ip)
  */
 function sanitize_alphanumeric($value)
 {
-    return preg_replace("[^a-zA-Z0-9\-\_]", "", $value);
+    return preg_replace("/[^a-zA-Z0-9\-\_]/", "", $value);
```
Shnoulle, line 523: Same question then 3.X

PR #3223, opened 2023-06-14
Task #CR-1070: implement version validation checks for adminthemes/surveythemes and display errors to the user

ptelu:


application/config/version.php
```diff
@@ -12,7 +12,7 @@
  */
 
 $config['versionnumber'] = '6.1.2';
-$config['dbversionnumber'] = 606;
+$config['dbversionnumber'] = 607;
```
olleharstedt, line 15: Dev branch is already on version 608.

application/helpers/update/updates/Update_607.php
```diff
@@ -0,0 +1,76 @@
...
+     * @inheritDoc
+     * @throws \CException
+     */
+    public function up()
+    {
+        $surveyThemes = $this->db->createCommand()
+            ->select('*')
+            ->from('{{templates}}')
+            ->queryAll();
+        if (!empty($surveyThemes)) {
```
olleharstedt, line 23: If you want to avoid excessive indentation, you can do an early return here instead

```php
if empty $surveyThemes
  return;
```

application/helpers/update/updates/Update_607.php
```diff
@@ -0,0 +1,76 @@
...
+                        if ($rowsDeleted >= 1) {
+                            $this->db->createCommand()->delete(
+                                '{{template_configuration}}',
+                                'template_name=:templateName',
+                                [':templateName' => $surveyThemeName]
+                            );
+                        }
+                    }
+                } catch (Exception $e) {
+                    rollBackToTransactionBookmark();
```
olleharstedt, line 43: Will this work even if you have multiple bookmarks set? Or do we need to remove the bookmark on success?

application/helpers/update/updates/Update_607.php
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+    }
+
+    /**
+     * Check if a survey theme is a core theme
+     * @param $surveyThemeName
+     * @return bool
+     */
+    private function isStandardTemplate($surveyThemeName): bool
```
olleharstedt, line 54: Couldn't this be just a hard-coded list?

application/models/Template.php
```diff
@@ -268,20 +268,7 @@ public static function checkTemplateXML($templateName, $templateFolder)
...
-            // TODO: write recursiveuninstall for childthemes
-            $extendedTemplates = (new Template)->findAll('extends = :templateName', [':templatename' => $templateName]);
-            if (!empty($extendedTemplates)) {
-                foreach ($extendedTemplates as $extendedTemplate) {
-                    TemplateConfig::uninstall($extendedTemplate->name);
-                }
-            }
-
-            TemplateConfig::uninstall($templateName);
+        if (!TemplateConfig::validateTheme($templateName, $currentThemePath)) {
```
olleharstedt, line 271: :+1: for moving code to functions :)

application/models/TemplateConfig.php
```diff
@@ -300,9 +300,9 @@ public function throwConsoleError($sCustomMessage = null)
 
 
     /**
-     * @todo document me
-     * @todo missing return value (php warning)
-     * @return boolean|null
+     * Check if this template is a standard template and save it in current model $this->isStandard
```
olleharstedt, line 303: Using "and" in a function description is a tell that it should rather be two functions. But it's old, right?

application/models/TemplateConfig.php
```diff
@@ -300,9 +300,9 @@ public function throwConsoleError($sCustomMessage = null)
...
+     * @return void
+     * @throws CException
```
olleharstedt, line 305: Nice corrections

application/models/TemplateConfig.php
```diff
@@ -885,6 +885,74 @@ public static function uninstall($templatename)
         return false;
     }
 
+    /**
+     * Uninstalls all surveythemes that are being extended from the supplied surveytheme name
+     * @param $templateName
+     * @return void
+     * @throws CDbException
+     */
+    public static function uninstallThemesRecursive($templateName): void
```
olleharstedt, line 894: Should be in service class, or...? Only way we can add unit-tests in the future.

application/models/TemplateConfig.php
```diff
@@ -885,6 +885,74 @@ public static function uninstall($templatename)
...
+        return true;
+    }
+
+    /**
+     * Checks if theme is compatible with the current limesurvey version
+     * @param $themePath
+     * @param bool $redirect
+     * @return bool
+     */
+    public static function isCompatible($themePath): bool
```
olleharstedt, line 944: I feel like such a function should not be static? `$surveyTheme->isCompatible()` would be more natural or expected OOP design, no?

olleharstedt: Ah wait, that already exists.

olleharstedt: Or perhaps `$surveyTheme->config->isCompatible()`

application/models/TemplateConfig.php
```diff
@@ -1007,10 +1075,11 @@ public function getAllDbTemplateFolders()
     }
 
     /**
-     * Returns an array with uninstalled or incompatible survey themes
-     * @return array|null
+     * Returns an array with uninstalled and/or incompatible survey themes
+     * @return array
```
olleharstedt, line 1079: If you wanna make future Psalm analysis happy, you can be more precise in your annotations than `array`, like `TemplateConfig[]` or `array<string, TemplateConfig>`, whatever is the correct type (didn't read the function body).

application/models/TemplateManifest.php
```diff
@@ -613,24 +613,6 @@ public function getOtherFiles()
         return $otherfiles;
     }
 
-
-    /**
-     *
-     */
-    public function getTemplateURL()
```
olleharstedt, line 620: Duplication?

application/views/themeOptions/index.php
```diff
@@ -58,19 +60,50 @@
                             </thead>
 
                             <tbody>
-                            <?php foreach ($oSurveyTheme->templatesWithNoDb as $oTemplate) : ?>
```
olleharstedt, line 61: :+1: for avoiding duplication of database calls

application/views/themeOptions/index.php
```diff
@@ -58,19 +60,50 @@
...
                                 <tr class="odd">
                                     <td class="col-lg-1"><?php echo $oTemplate->getPreview(); ?></td>
                                     <td class="col-lg-2"><?php echo $oTemplate->sTemplateName; ?></td>
                                     <td class="col-lg-3"><?php echo $oTemplate->getDescription(); ?></td>
                                     <td class="col-lg-2"><?php eT('XML themes'); ?></td>
                                     <td class="col-lg-1"><?php echo $oTemplate->config->metadata->extends; ?></td>
-                                    <td class="col-lg-2"><?php echo $oTemplate->getButtons(); ?></td>
+                                    <?php if (TemplateConfig::isCompatible($oTemplate->path . 'config.xml')): ?>
+                                        <td class="col-lg-2"><?php echo $oTemplate->getButtons(); ?></td>
+                                    <?php else: ?>
```
olleharstedt, line 74: I think it's better to look at this HTML than to just read it. :) But Tibor/I would need instructions on how to do that.

themes/admin/Sea_Green/config.xml
```diff
@@ -19,6 +19,10 @@
         <lastUpdate>2017-27-11 17:00:00</lastUpdate>
     </metadata>
 
+    <compatibility>
+        <version>6.0</version>
```
olleharstedt, line 23: You don't need to add this to survey themes too?

application/config/version.php
ptelu: I wouldnt artificially change it right now but instead do it during a merge

application/helpers/update/updates/Update_607.php
ptelu: as far as i read up it always saves the newest savepoint when you generate one with the same name, so there shouldnt be any issues, correct me if im wrong. Changed the name to "beforeThemeDelete"

application/helpers/update/updates/Update_607.php
ptelu: Well i could just check against specific themenames that are standardthemes and  in the table.

application/models/TemplateConfig.php
ptelu: I would like to but this hole setup with 3 different classes is already to much if i now also at a serviceclass everything might get more confusing it should all be one class with service/helper classes.

application/models/TemplateConfig.php
ptelu: its a bit problematic since there are functionscalling this who are also static and dont have a tempalte class

application/models/TemplateManifest.php
ptelu: yes

application/views/themeOptions/index.php
ptelu: this is just displayinga button and modal instead of the normal buttons, should be tested in manual testing yeah

themes/admin/Sea_Green/config.xml
ptelu: yeah it was in the commit that is outside of this pull request as mentioned in the ticket

PR #3225, opened 2023-06-14
Fixed issue: #18823 STATA Export fails with certain question types.

gabrieljenik:
Tests added.


Shnoulle:
Oh, you add test on 5.X too ? I think we can keep it only on master ?

Shnoulle (approved):

PR #3229, opened 2023-06-20
Fixed issue 18821: Public url is not used for SURVEYURL

twilligls:
Dev:

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -2333,13 +2333,19 @@ class Survey extends LSActiveRecord implements PermissionInterface
                 }
 
                 // Create the URL according to the configured format
+                $sPublicUrl = Yii::app()->getConfig("publicurl");
+                $aPublicUrl = parse_url($sPublicUrl);
+                $baseUrl = Yii::app()->getBaseUrl(true);
+                if (isset($aPublicUrl['scheme']) && isset($aPublicUrl['host'])) {
+                    $baseUrl = $sPublicUrl;
+                }
                 $urlManager = Yii::app()->getUrlManager();
                 $urlFormat = $urlManager->getUrlFormat();
                 if ($urlFormat == CUrlManager::GET_FORMAT) {
-                    $url = Yii::app()->getBaseUrl(true);
+                    $url = $baseUrl;
                     $params = [$urlManager->routeVar => $alias] + $params;
                 } else {
-                    $url = Yii::app()->getBaseUrl(true) . '/' . $alias;
+                    $url = $baseUrl . '/' . $alias;
                 }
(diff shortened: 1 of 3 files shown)


Shnoulle:
Another solution can be add a LSYii_Application->getBaseUrl ?

Shnoulle:
Put this on dev mantis maybe, we need to review App->getController()->getAbsoluteUrl()

twilligls:
> 

Doesn't work, because you would override the baseUrl all the time with configured publicurl, even when you don't need it.

twilligls:
> Another solution can be add a LSYii_Application->getBaseUrl ?


Doesn't work, because you would override the baseUrl all the time with configured publicurl, even when you don't need it.

twilligls:
> Put this on dev mantis maybe, we need to review App->getController()->getAbsoluteUrl()

I don't understand

Shnoulle:
> > Put this on dev mantis maybe, we need to review App->getController()->getAbsoluteUrl()
> 
> I don't understand

Just create a Development mantis issue : https://bugs.limesurvey.org/bug_report_page.php

> > Another solution can be add a LSYii_Application->getBaseUrl ?
> 
> Doesn't work, because you would override the baseUrl all the time with configured publicurl, even when you don't need it.

 :+1: 

Maybe a new function then ?  I think we need it for API too.

gabrieljenik:
Some comments, maybe for the DEV branch:
- I agree we need a method on the app but only for the createPublicUrl
  + Make new method based on LSYiiController::createAbsoluteUrl()
  + Make LSYiiController::createAbsoluteUrl() to call the new app method.
  + Probably rename the LSYiiController::createAbsoluteUrl() to LSYiiController::createPublicUrl()

application/models/Survey.php
```diff
@@ -2351,7 +2357,8 @@ public function getSurveyUrl($language = null, $params = [], $preferShortUrl = t
 
         // If short url is not preferred or no alias is found, return a traditional URL
         $urlParams = array_merge($params, ['sid' => $this->sid, 'lang' => $language]);
-        $url = Yii::app()->createAbsoluteUrl('survey/index', $urlParams);
+        $url = App()->getController()->createAbsoluteUrl('survey/index', $urlParams);
```
gabrieljenik, line 2360: I don't like to call a controller from a model, but if no code reorg is to be done, I don't see any other way

gabrieljenik: Maybe you can just move the code to a helper?
Maybe now is a good time to create App::createPublicUrl() ?

twilligls: ok, you convinced me now :)

twilligls: Please check my latest commit

PR #3233, opened 2023-06-20
Fixed issue #18884: Stored XSS in Survey Groups Function

gabrieljenik:
Avoid the issue for users other than superadmin. 
It doesn't treat existing groups.

### DIFF
--- a/application/controllers/admin/SurveysGroupsController.php
+++ b/application/controllers/admin/SurveysGroupsController.php
@@ -163,7 +163,7 @@ class SurveysGroupsController extends SurveyCommonAction
         $aData = array(
             'model' => $model,
             'action' => App()->createUrl("admin/surveysgroups/sa/update", array('id' => $model->gsid, '#' => 'settingsForThisGroup')),
-            'pageTitle' => gT('Update survey group: ') . $model->title,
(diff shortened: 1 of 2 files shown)


Shnoulle:
No,
we need to encode it when shown.

1. XSS protection is here for survey, for participant only. Not for admin part
2. No admin part must allow any XSS, by superadmin or not
3. This description didn't have an HTML editor (and there are no real reason to hide HTML)
4. This is the lits of group : no reason to have HTML

The 5.X have the good solution. No need to filter or restrict, just is not HTML, then always encode when show (we show it only on list)

Shnoulle:
title must be encoded too …

Shnoulle:
> This patch applies the same principles that are applied on other parts of the system, so not sure about that statement.

No,

**It's false** … you can add script in group title : the group title are filtered when view in group list.

Same for question text i organizer or question list.

Please : check the 5.X version where we don't have this issue…

Shnoulle:
> show them as saved (LSYii_Validator).

We **NEVER**  show something as saved in admin part ! ONLY on public survey page.

gabrieljenik:
Ready for re-reviewal @Shnoulle

gabrieljenik:
Closed in favor of #3241

PR #3238, opened 2023-06-21
Fixed issue #17536: Create automatic test for XSS Purifier

gabrieljenik:
Added:
- Test: Config variable filterxsshtml vs user type
- Test: xssFilter with tags and html_entity_decode
- Test: isLanguageMulti
- Test: isLanguage

### DIFF
--- a/tests/unit/LSYiiValidatorsTest.php
+++ b/tests/unit/LSYiiValidatorsTest.php
@@ -35,11 +35,35 @@ class LSYiiValidatorsTest extends TestBaseClass
      */
     public function testXssFilterAttributeForSuperAdmin()
     {
-        //For test : we use user #1 as superadmin.
-        \Yii::app()->session['loginID'] = 1;
+        //Create super admin
+        $userName = \Yii::app()->securityManager->generateRandomString(8);
+
+        $userData = array(
+            'users_name' => $userName,
+            'email'      => $userName . '@example.com'
+        );
+
+        $permissions = array(
+            'superadmin' => array(
+                'create' => true,
+                'read'   => true,
+                'update' => true,
+                'delete' => true,
+                'import' => true,
+                'export' => true,
+            )
+        );
+
+        $user = self::createUserWithPermissions($userData, $permissions);
+
+        //Login as super admin.
+        \Yii::app()->session['loginID'] = $user->uid;
         $superAdminValidator = new \LSYii_Validators();
 
-        $this->assertTrue(\Yii::app()->getConfig('filterxsshtml'), 'filterxsshtml should be true by default.');
+        //Save config state in order to restore it later.
+        $filterXssTmp = \Yii::app()->getConfig('filterxsshtml');
+        \Yii::app()->setConfig('filterxsshtml', true);
+
         $this->assertFalse($superAdminValidator->xssfilter, 'The xssfilter attribute should be false for super admins.');
 
         //Changing filterxsshtml.
@@ -50,8 +74,11 @@ class LSYiiValidatorsTest extends TestBaseClass
         $this->assertFalse($newSuperAdminValidator->xssfilter, 'The xssfilter attribute should be false for super admins.');
 
         //Returning to original values.
-        \Yii::app()->setConfig('filterxsshtml', true);
+        \Yii::app()->setConfig('filterxsshtml', $filterXssTmp);
         \Yii::app()->session['loginID'] = null;
+
+        //Delete user.
+        $user->delete();
     }
 
     /**
@@ -69,7 +96,10 @@ class LSYiiValidatorsTest extends TestBaseClass
         \Yii::app()->session['loginID'] = $userId;
         $regularUserValidator = new \LSYii_Validators();
 
-        $this->assertTrue(\Yii::app()->getConfig('filterxsshtml'), 'filterxsshtml should be true by default.');
+        //Save config state in order to restore it later.
+        $filterXssTmp = \Yii::app()->getConfig('filterxsshtml');
+        \Yii::app()->setConfig('filterxsshtml', true);
+
         $this->assertTrue($regularUserValidator->xssfilter, 'The xssfilter attribute should be true for regular users.');
 
         //Changing filterxsshtml.
@@ -80,7 +110,7 @@ class LSYiiValidatorsTest extends TestBaseClass
         $this->assertFalse($newRegularUserValidator->xssfilter, 'The xssfilter attribute should be false for regular users with filterxsshtml set to false.');
 
         //Returning to original values.
-        \Yii::app()->setConfig('filterxsshtml', true);
+        \Yii::app()->setConfig('filterxsshtml', $filterXssTmp);
         \Yii::app()->session['loginID'] = null;
 
         //Delete user.
@@ -122,7 +152,7 @@ class LSYiiValidatorsTest extends TestBaseClass
             ),
             array(
                 'string'   => `{join("<","script",">",'alert("Test")',"<","/script",">")}`,
-                'expected' => ''
+                'expected' => '&lt;script&gt;alert("Test")&lt;/script&gt;'
             ),
             array(
                 'string'   => '{join("<s", "cript>alert("Test")</script>")}',
@@ -143,6 +173,7 @@ class LSYiiValidatorsTest extends TestBaseClass
         $validator = new \LSYii_Validators();
 
         $cases = array(
+            '<h1>Header</h1>',
             '<p>Paragraph</p>',
             '<strong>Text</strong>',
             '<span>Some text</span>',
--- a/tests/unit/LSYiiValidatorsTest.php
+++ b/tests/unit/LSYiiValidatorsTest.php
@@ -27,4 +27,57 @@ class LSYiiValidatorsTest extends TestBaseClass
             $this->assertEquals($expected, $actual);
         }
     }
+
+    /**
+     * Testing that the xssfilter attribute varies
+     * depending on the user.
+     */
+    public function testXssFilterAttribute()
+    {
+        $regularUserValidator = new \LSYii_Validators();
+
+        $isFiltered = $regularUserValidator->xssfilter;
+        $this->assertTrue($isFiltered, 'The xssfilter attribute should be true for regular users.');
+
+        \Yii::app()->session['loginID'] = 1;
+        $superAdminValidator = new \LSYii_Validators();
+
+        $isFiltered = $superAdminValidator->xssfilter;
+        $this->assertFalse($isFiltered, 'The xssfilter attribute should be false for super admins.');
+    }
+
+    /**
+     * Testing that any script or dangerous HTML is removed.
+     */
+    public function testXssFilterApplied()
+    {
+        $validator = new \LSYii_Validators();
+
+        $cases = array(
+            array(
+                'string'   => '<script>alert(`Test`)</script>',
+                'expected' => ''
+            ),
+            array(
+                'string'   => `{join('html_entity_decode("', '<script>alert("Test")</script>")')}`,
+                'expected' => ''
+            ),
+            array(
+                'string'   => '<title>html_entity_decode("<script>alert("Test")</script>")</title>',
+                'expected' => 'html_entity_decode("")'
+            ),
+            array(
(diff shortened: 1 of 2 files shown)


tests/unit/LSYiiValidatorsTest.php
```diff
@@ -27,4 +27,155 @@ public function testHtmlEntityDecodeFilter()
...
+
+    /**
+     * Testing that the xssfilter attribute will always be
+     * false for super admin users even if filterxsshtml is
+     * changed.
+     */
+    public function testXssFilterAttributeForSuperAdmin()
+    {
+        //Mocking super admin login.
+        \Yii::app()->session['loginID'] = 1;
```
Shnoulle, line 39: We must put a information on manual : For test : we use user #1 as superadmin. 

https://github.com/LimeSurvey/LimeSurvey/blob/a3605854a63d56b360660dea47e144ea766d382f/application/config/config-defaults.php#L673

tests/unit/LSYiiValidatorsTest.php
```diff
@@ -27,4 +27,155 @@ public function testHtmlEntityDecodeFilter()
...
+
+    /**
+     * Testing that the xssfilter attribute varies
+     * for regular users depending on filterxsshtml.
+     */
+    public function testXssFilterAttributeForRegularUsers()
+    {
+        //Create user.
+        $newPassword = createPassword();
+        $userId = \User::insertUser('test_user', $newPassword, 'John Doe', 1, 'jd@mail.com');
```
Shnoulle, line 65: I like a random namle for user just in case 

(i already have test, test_rc etc … in my dev instance) 

see https://github.com/LimeSurvey/LimeSurvey/blob/a3605854a63d56b360660dea47e144ea766d382f/tests/functional/backend/XssAndScriptCheckTest.php#L22

tests/unit/LSYiiValidatorsTest.php
```diff
@@ -27,4 +27,155 @@ public function testHtmlEntityDecodeFilter()
...
+
+        //Mocking regular user login.
+        \Yii::app()->session['loginID'] = $userId;
+        $regularUserValidator = new \LSYii_Validators();
+
+        $this->assertTrue(\Yii::app()->getConfig('filterxsshtml'), 'filterxsshtml should be true by default.');
```
Shnoulle, line 71: Then we can not use this test on a local dev instance :(

tests/unit/LSYiiValidatorsTest.php
```diff
@@ -27,4 +27,155 @@ public function testHtmlEntityDecodeFilter()
...
+            array(
+                'string'   => `{join('html_entity_decode("', '<script>alert("Test")</script>")')}`,
+                'expected' => ''
+            ),
+            array(
+                'string'   => '<title>html_entity_decode("<script>alert("Test")</script>")</title>',
+                'expected' => 'html_entity_decode("")'
+            ),
+            array(
+                'string'   => `{join('html_entity_decode("', '<s', 'cript>alert("Test")</script>")')}`,
```
Shnoulle, line 111: Must test too `{join("<s", "cript>alert("Test")</script>")}` see https://github.com/LimeSurvey/LimeSurvey/blob/a3605854a63d56b360660dea47e144ea766d382f/application/core/LSYii_Validators.php#L146 no ?

Alls DS_STRING and DQ_STRING are purified

tests/unit/LSYiiValidatorsTest.php
```diff
@@ -27,4 +27,155 @@ public function testHtmlEntityDecodeFilter()
...
+
+        foreach ($cases as $case) {
+            $this->assertSame($case['expected'], $validator->xssFilter($case['string']), 'Unexpected filtered dangerous string.');
+        }
+    }
+
+    /**
+     * Testing that safe HTML tags are not removed.
+     */
+    public function testSafeHtml()
```
Shnoulle, line 132: Adding some testUnsageHtml ?

`<strong>strong` `<style>` `

?

tests/unit/LSYiiValidatorsTest.php
```diff
@@ -27,4 +27,155 @@ public function testHtmlEntityDecodeFilter()
...
+    /**
+     * Testing the language filters
+     * through the Survey model.
+     */
+    public function testLanguageFilters()
+    {
+        // Testing languageFilter.
+        $survey = \Survey::model()->insertNewSurvey(array('language' => 'ko')); // Set language to Korean.
+
+        $this->assertSame('ko', $survey->language, 'The language filter did not return a correctly filtered language string.');
```
Shnoulle, line 156: must check some other specific one `de-informal`  `de-easy`  too.

tests/unit/LSYiiValidatorsTest.php
```diff
@@ -27,4 +27,155 @@ public function testHtmlEntityDecodeFilter()
...
+
+        $survey->language = 'enǵ';
+        $survey->save();
+
+        $this->assertSame('en', $survey->language, 'The language filter did not return a correctly filtered language string.');
+
+        // Testing multiLanguageFilter.
+        $survey->additional_languages = 'es';
```
Shnoulle, line 164: Why not `en es` here ? Must work

Shnoulle (commented):
This disallow usage on specific dev instance (about the test of the default one)

And some more test here ?

tests/unit/LSYiiValidatorsTest.php
gabrieljenik: Yes, will add it.
We just grab the ones from the ticket.

I guess even more can be added.
The idea is to add more as they appear

tests/unit/LSYiiValidatorsTest.php
Shnoulle: > Isn't that already tested on testXssFilterApplied?

No, testXssFilterApplied seems to test only script, not broken HTML.

But it's more new test to do. Can be added or not. Task is unclear here.

gabrieljenik: Created new ticket: 18905: Create automatic test for XSS Purifier: Broken HTML

tests/unit/LSYiiValidatorsTest.php
Shnoulle: Yes, but here : you check more than the previous pone.

if `{join("<s", "cript>alert("Test")</script>")}` is available, no need `html_entity_decode` ;)

Shnoulle: The expression for testing :` {join("<","script,">",'alert("Test")',"<","/script",">")}`

Must return `&lt;script&gt;alert("Test")&lt;/script&gt;`

tests/unit/LSYiiValidatorsTest.php
gabrieljenik: Done :)
https://github.com/LimeSurvey/LimeSurvey/commit/613bf98a4218f7d25f2f11bf0ee5e969e6a4ca6f#diff-9829ec9a114cf86c2fa07767e34fdb96ee9d28b0a571c1f5247c8408d194f048R193

tests/unit/LSYiiValidatorsTest.php
gabrieljenik: `en` is the main language.
There is now an assert for more than 2 languages.

tests/unit/LSYiiValidatorsTest.php
Shnoulle: @olleharstedt on some local instance : we can not be sure to have filterxsshtml to ON by default.

Did we test default options ?

Shnoulle: I'm unsure here …

I use an existing system to create test, but manual "Important.png  Warning : NEVER run tests on a production system. The tests WILL delete and modify the limesurvey database as well as the config.php file. Always do a backup of important data before running the tests."

https://manual.limesurvey.org/index.php?title=Regression_and_unit_tests

Shnoulle: Yes : but when construct test : i have my own test DB ;)

Not really important

tests/unit/LSYiiValidatorsTest.php
gabrieljenik: Added it below.

gabrieljenik:
@Shnoulle Please review now.
- not relying on user #1.
- setting back config settings after test is finished.
- More tests added

Shnoulle:
> not relying on user # 1.

How ! I don't ask this … it's really an information to provide on manual for other dev …

> setting back config settings after test is finished.

It's not the task of global tester ?

Shnoulle (approved):

Shnoulle:
test more than needed ! Great !

gabrieljenik:
@Shnoulle 

Still one of the new newly added test expressions is not coming as expected.
https://github.com/LimeSurvey/LimeSurvey/actions/runs/5348498962/jobs/9698445995#step:7:47
Are you sure about that expected value?

Shnoulle:
> @Shnoulle
> 
> Still one of the new newly added test expressions is not coming as expected. https://github.com/LimeSurvey/LimeSurvey/actions/runs/5348498962/jobs/9698445995#step:7:47 Are you sure about that expected value?

NO :)

`{join("","script","&gt;",'alert("Test")',"","/script","&gt;")}`

I think santizer set `<` to `&lt;` but no :(, need some space

Best : `{join(trim(" < "),"script",">",'alert("Test")',trim(" < "),"/script",">")}` must go to `{join(trim(" &lt; "),"script","&gt;",'alert("Test")',trim(" &lt; "),"/script","&gt;")}`

gabrieljenik:
> > @Shnoulle
> > Still one of the new newly added test expressions is not coming as expected. https://github.com/LimeSurvey/LimeSurvey/actions/runs/5348498962/jobs/9698445995#step:7:47 Are you sure about that expected value?
> 
> NO :)
> 
> `{join("","script","&gt;",'alert("Test")',"","/script","&gt;")}`
> 
> I think santizer set `<` to `&lt;` but no :(, need some space
> 
> Best : `{join(trim(" < "),"script",">",'alert("Test")',trim(" < "),"/script",">")}` must go to `{join(trim(" &lt; "),"script","&gt;",'alert("Test")',trim(" &lt; "),"/script","&gt;")}`

Done

olleharstedt:
Broken CI?

gabrieljenik:
I believe the new test is messing with existing tests.
- During the new test we create a new superadmin user. (I think it is user #2).
- The existing test doens't create a new user but relies on a user #2, which is not expected to be superadmin. 
- The permissions for superadmin are picked up during the new test, get cached and hence reused on the existing test, making it fail.

I believe this unveils a problem: The existing test maybe relying on a user id of a user which doesn't exist. So that test may not be as effective as expected.

Will fix and check

Shnoulle:
> I believe this unveils a problem: The existing test maybe relying on a user id of a user which doesn't exist. So that test may not be as effective as expected.

:+1: need an user : create it and get the userid :)

olleharstedt:
Sometimes it helps to set `$_SESSION = [];` in test setup phase. Perhaps always should be done actually.

tests/unit/models/PermissionTest.php
```diff
@@ -83,7 +83,10 @@ public function testOwnershipSuccess()
     public function testOwnershipFailure()
     {
         // NB: Not 1 (superadmin).
-        $userId = 2;
+        //Create user.
+        $newPassword = createPassword();
+        $userName = \Yii::app()->securityManager->generateRandomString(8);
+        $userId = \User::insertUser($userName, $newPassword, 'John Doe', 1, $userName . '@mail.com');
```
Shnoulle, line 89: `@example.org` : it's done for sample and test.

Do not use an existing website (except `limesurvey.(coo|org)`)

tests/unit/LSYiiValidatorsTest.php
```diff
@@ -40,7 +40,7 @@ public function testXssFilterAttributeForSuperAdmin()
 
         $userData = array(
             'users_name' => $userName,
-            'email'      => $userName . '@example.com'
+            'email'      => $userName . '@example.org'
```
Shnoulle, line 43: example.com is OK ;)

gabrieljenik:
All green!

PR #3239, opened 2023-06-21
Fixed issue #18885: CSRF Leading to reset Boxes

gabrieljenik:
This script had its own confirm. I could have put the POST in that confirm, but it didn't make sense. Took it out and used the confirmation modal that is used in other screens. As we did in 16792.


Shnoulle (approved):

Shnoulle:
using filter for post validation is really great !

I'll keep that in mind for future fix.

PR #3241, opened 2023-06-22
Bug/18884 fix stored xss in surveygroup

ptelu:
Dev:


twilligls (approved):

twilligls (approved):

twilligls (approved):

application/core/LSYii_Validators.php
```diff
@@ -46,21 +46,21 @@ class LSYii_Validators extends CValidator
...
         // If run from console there is no user
         $this->xssfilter = (
-            $this->xssfilter && // this
-            (
-                (defined('PHP_ENV') && PHP_ENV == 'test') || // phpunit test : don't check controller
-                (
-                    ($controller = Yii::app()->getController()) !== null && // no controller
-                    (get_class($controller) !== 'ConsoleApplication') // ConsoleApplication
+            $this->xssfilter
+            && ((defined('PHP_ENV') // phpunit test : don't check controller
```
olleharstedt, line 56: Uhm this logical chunk is equally unreadable as before. Use this instead? https://refactoring.guru/decompose-conditional

application/extensions/admin/grid/CLSGridView.php
```diff
@@ -72,14 +80,31 @@ protected function lsAfterAjaxUpdate(): void
         }
     }
 
+    protected function rowLink()
```
olleharstedt, line 83: Missing phpdoc

application/extensions/admin/survey/ListSurveysWidget/views/listSurveys.php
```diff
@@ -49,8 +50,7 @@
                 [
                     'header'            => gT('Survey ID'),
                     'name'              => 'survey_id',
-                    'type'              => 'raw',
-                    'value'             => 'CHtml::link($data->sid, Yii::app()->createUrl("surveyAdministration/view/",array("iSurveyID"=>$data->sid)))',
```
olleharstedt, line 53: Maybe I'm an idiot, but how is removing links related to XSS? o0

application/extensions/admin/grid/CLSGridView.php
olleharstedt: You might want to move up $options = [] to top and return $options at bottom, to always return an array instead of array|void.

application/core/LSYii_Validators.php
ptelu: sure we could do that in general, probably wont do it here right now

application/extensions/admin/survey/ListSurveysWidget/views/listSurveys.php
ptelu: type text is running the value through encoding, we tried to get a lot of information into that field by adding raw type. So instead we now have an option for the gridview that can add this functionality

PR #3249, opened 2023-06-23
Fixed issue #18917: [security] Stored XSS in the user group deletion confirmation popup

ptelu:
…confirmation popup

Dev:

### DIFF
--- a/application/models/UserGroup.php
+++ b/application/models/UserGroup.php
@@ -347,7 +347,6 @@ class UserGroup extends LSActiveRecord
             'url'              => Yii::app()->createUrl("userGroup/mailToAllUsersInGroup/ugid/$this->ugid"),
             'enabledCondition' => $permissionUsergroupsEdit
         ];
-
         $deletePostData = json_encode(['ugid' => $this->ugid]);
         $dropdownItems[] = [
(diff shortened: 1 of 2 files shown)


application/models/UserGroup.php
```diff
@@ -357,7 +356,7 @@ public function getButtons()
                 'data-bs-toggle' => "modal",
                 'data-post-url'  => App()->createUrl("userGroup/deleteGroup"),
                 'data-post-datas' => $deletePostData,
-                'data-message'   => sprintf(gt("Are you sure you want to delete user group '%s'?"), $this->name),
+                'data-message'   => sprintf(gt("Are you sure you want to delete user group '%s'?"), sanitize_paranoid_string($this->name)),
```
Shnoulle, line 359: encode didn't work ?

Maybe we must auto filter groupname when save. We don't really need HTML here.

Shnoulle: My comment are not a request change comment :)

More a opinion ;)

ptelu: you are correct, was a bit of complicated nested string, but after i fixed the action_dropdown code, encode is enough

PR #3252, opened 2023-06-23
 Fixed issue #18193: Question theme error: Missing metadata (Type: X)

Shnoulle:
Dev: fix filter to be applye
Dev: Fix invalid theme when save


application/models/Question.php
```diff
@@ -1725,16 +1723,30 @@ public function getScaledSubquestions()
...
+        if (empty($this->type)) {
+            return null;
+        }
+        /* not needed in child question */
+        if (!empty($this->parent_qid)) {
+            return null;
+        }
         if (!empty($this->question_theme_name) && $this->question_theme_name != 'core') {
-            return;
+            $criteria = new CDbCriteria();
```
gabrieljenik, line 1735: Not sure if we should put this into a new method, just as to give some strength to the liason in between the models, and a have a more notorious place to update if needed.

application/models/Question.php
```diff
@@ -1725,16 +1723,30 @@ public function getScaledSubquestions()
...
+            $criteria->addCondition('question_type = :question_type AND name = :name');
+            $criteria->params = [':question_type' => $this->type, ':name' => $this->question_theme_name];
+            $questionTheme = QuestionTheme::model()->query($criteria, false);
+            if ($questionTheme) {
```
gabrieljenik, line 1739: What should happen if expected question theme doesn't exist?

Shnoulle: Reset to the QuestionTheme from question type.

If Question type didn't have theme : no solution. You can not update question->type

application/models/Question.php
Shnoulle: Maybe , but before must understand why create a broken solution in QuestionTheme : https://github.com/LimeSurvey/LimeSurvey/blob/6e5a5767219ee97ce8f81cfbc85c20b4cd4331c8/application/models/QuestionTheme.php#L1084


I can update here : https://github.com/LimeSurvey/LimeSurvey/blob/6e5a5767219ee97ce8f81cfbc85c20b4cd4331c8/application/models/QuestionTheme.php#L629 
If needed, but i think it's out of fix issue. It's rewriting for dev.

gabrieljenik: > Maybe , but before must understand why create a broken solution in QuestionTheme :

That code is not recent, so I will try to recall.
I understand it was related to the situation when creating a question.
There is a point in time where it is not saved yet, but still need to use the same mechanisms as if it would be being edited. Then, the relationship with QuestionTheme is not materialized yet, then we instantiate a dummy question theme to be used temporarily.

application/models/Question.php
gabrieljenik: Would like to call @olleharstedt or someone else to review this.
Another set of eyes would be good.

Shnoulle: If question type are disabled : we really have no way to use it (if survey is activated).

application/models/Question.php
```diff
@@ -168,7 +168,7 @@ public function rules()
             return 'N';
         });
         /* Don't save empty or 'core' question theme name */
-        $aRules[] = ['question_theme_name', 'questionThemeNameValidator'];
+        $aRules[] = ['question_theme_name', 'filter', 'filter' =>  [$this, 'questionThemeNameValidator'] ];
```
olleharstedt, line 171: Is this rule applied only when question_them_name is modified btw? Is Yii smart enough to optimize that?

Shnoulle: No,

Another way to broke

1. Import a questiontheme
2. Create a question with this theme
3. Delete the theme

question theme was not updated but broke.

olleharstedt: > No,

:(

olleharstedt:
Are there testing instructions in the bug report?

application/models/Question.php
Shnoulle: > Ah but that's missing logic in question theme delete process, then.

Sorry : i mean : Delete the Theme directory.

Else: before it was in afterFind, moved to when save : action is done ONLY by admin. 

Why create an optimization here ?

olleharstedt: > > Ah but that's missing logic in question theme delete process, then.
> 
> Sorry : i mean : Delete the Theme directory.
> 
> Else: before it was in afterFind, moved to when save : action is done ONLY by admin.
> 
> Why create an optimization here ?

Just avoiding needless queries to the db.

Shnoulle:
> Are there testing instructions in the bug report?

I updated with a new file

1. Import limesurvey_survey_BadTheme.lsa
2. Check theme of question in theme editor : see the error
3. Edit : try to update theme

application/models/Question.php
Shnoulle: > > > Ah but that's missing logic in question theme delete process, then.
> > 
> > 
> > Sorry : i mean : Delete the Theme directory.
> > Else: before it was in afterFind, moved to when save : action is done ONLY by admin.
> > Why create an optimization here ?
> 
> Just avoiding needless queries to the db.

But here : only when save action by admin, it's already a save action.

If we can not fix here : when we fix it ?

application/models/Question.php
Shnoulle: > Still, I was thinking just about a new static method `QuestionTheme::ByNameAndType`

Maybe it's better to use existing method ?

application/models/Question.php
Shnoulle: > Just avoiding needless queries to the db.

Else : it's important when there a 100 queries in same page : https://bugs.limesurvey.org/view.php?id=18474 (and related)

But here : just a sample query …

Maybe we can improve it, but still need it …

Alternative solution : fix it when load page by admin ? In QuestionController ? 
But i think Load QuestionController happen more than save.

Shnoulle:
Merged

PR #3253, opened 2023-06-26
Fixed issue #18913: [security] incorrect permissions for useraction

ptelu:
Dev:

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -289,20 +289,76 @@ class UserManagementController extends LSBaseController
      */
     public function actionDeleteUser()
     {
-        if (!Permission::model()->hasGlobalPermission('users', 'delete')) {
-            return $this->renderPartial(
-                'partial/error',
-                ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
-            );
-        }
-        $userId = (int) Yii::app()->request->getPost('userid');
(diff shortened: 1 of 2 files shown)


application/controllers/UserManagementController.php
```diff
@@ -1150,21 +1213,57 @@ public function actionTakeOwnership()
...
+
+
+        // Check if user owns a survey
+        $aOwnedSurveys = Survey::model()->findAllByAttributes(['owner_id' => $userId]);
+        if (count($aOwnedSurveys)) {
             return false;
         }
 
-        $oUser = User::model()->findByPk($uid);
+        // Transfer any Participants owned by this user to site's admin
```
twilligls, line 1263: 👍

application/controllers/UserManagementController.php
```diff
@@ -1150,21 +1213,57 @@ public function actionTakeOwnership()
      */
     public function deleteUser(int $uid): bool
```
twilligls, line 1214: Why is the actionDeleteUser function not using this function?
It's doing basically the same isn't it?

ptelu: Because someone thought it would be funny to have 2 different systems :).
The real answer would be that the massive action renders a completely different modal which expects other values and formats.

twilligls: nice ;)

PR #3260, opened 2023-06-27
Optimization of loadFromFile for config.xml

jjborie:
We encountered some performance issue due to the reloading everytime for the config.xml files, as our files are stored on EFS in AWS.

Here is a small optimization that cache simplexml_load_file result to a static cache. This avoid to reload the file several times for the same pages.

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized extension configuration loading to improve application responsiveness and reduce system overhead during extension initialization.

### DIFF
--- a/application/models/ExtensionConfig.php
+++ b/application/models/ExtensionConfig.php
@@ -149,9 +149,11 @@ class ExtensionConfig
      */
     public static function loadFromFile($file)
     {
-        if (!file_exists($file)) {
-            return null;
-        } else {
+        static $_c = [];
+        if (!array_key_exists($file, $_c)) {
+            if (!file_exists($file)) {
+                return null;
+            }
             if (\PHP_VERSION_ID < 80000) {
                 libxml_disable_entity_loader(false);
             }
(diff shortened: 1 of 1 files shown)


application/models/ExtensionConfig.php
```diff
@@ -149,9 +149,11 @@ public function isCompatible()
      */
     public static function loadFromFile($file)
     {
-        if (!file_exists($file)) {
-            return null;
-        } else {
+        static $_c = [];
+        if (!array_key_exists($file, $_c)) {
```
gabrieljenik, line 153: Why not making
```
if (array_key_exists($file, $_c)) {
  return $_c[$file];
}
```

That would make code a bit more easier to read and analyze, as it would be flatter.

olleharstedt:
Local static variables like these can mess up the unit tests, as seems to have been the case? They can also make reading the code a bit harder, as you might be forced to clear static variables manually in some cases.

Is there an optimization that can be done on a higher level instead? What does your stack trace look like for your bottlenecks?

gabrieljenik:
> Local static variables like these can mess up the unit tests, as seems to have been the case?

Maybe make it a static attribute?

> They can also make reading the code a bit harder, as you might be forced to clear static variables manually in some cases, basically increasing the knowledge burden of the programmer.

Agree. Maybe there could be a init routine which, based on a config variable, will (optionally) enable the optimization.

> Is there an optimization that can be done on a higher level instead? What does your stack trace look like for your bottlenecks?

Nice

c-schmitz:
@coderabbitai review

PR #3267, opened 2023-07-03
Fixed issue #18798: Out-of-memory problem in statistics export

gabrieljenik:


olleharstedt:
Same code as already merged into 5.x.

PR #3270, opened 2023-07-04
Fixed issue #18928: Stored XSS on Survey "Notification and data function"

gabrieljenik:
- Fixed XSS running htmlspecialchars before showing
- Fixed wrong extra space in notification email fields


Shnoulle (approved):
I prefer [CHtml::encode](https://www.yiiframework.com/doc/api/1.1/CHtml#encode-detail) BUT we use [htmlspecialchars](https://github.com/LimeSurvey/LimeSurvey/blob/848a0e18d1e0f9cea054dc2685e42d307ecbf53e/application/views/admin/survey/subview/accordion/_notification_panel.php#L212-L213) in 5.X 

:+1:

PR #3271, opened 2023-07-04
Fixed issue #18934: Stored XSS in label set administration

gabrieljenik:


application/models/Label.php
```diff
@@ -65,6 +65,13 @@ public function rules()
                             'params' => array(':lid' => $this->lid)
                     ),
                     'message' => '{attribute} "{value}" is already in use.'),
+            // Only alphanumeric
```
Shnoulle, line 68: It's NOT in the rules ?

application/models/Label.php
```diff
@@ -65,6 +65,13 @@ public function rules()
...
+            array(
```
Shnoulle, line 69: No : pattern can be removed or updated by script.

Must encode always

Shnoulle (changes requested):
«If an attacker can control a script that is executed in the victim's browser, then they can typically fully compromise that user. The attacker can carry out any of the actions that are applicable to the impact of reflected XSS vulnerabilities.«


If an attacker can add script : he can remove any pattern and send the POST.

Must encode result too.

Shnoulle (approved):
OK, added in rules.

PR #3272, opened 2023-07-04
Fixed issue #18927: Export user roles without authorization

gabrieljenik:


application/controllers/UserRoleController.php
```diff
@@ -242,6 +242,10 @@ public function actionViewRole(int $ptid)
      */
     public function actionRunExport($ptid)
     {
+        if (!Permission::model()->hasGlobalPermission('superadmin', 'read')) {
```
Shnoulle, line 245: 403 ?

Shnoulle (approved):
OK, but maybe 403 ?

application/controllers/UserRoleController.php
gabrieljenik: Sorry, forgot to comment on it.
A 403 would have looked better to me as well, but the other stocks already had a flash message.

PR #3273, opened 2023-07-04
Fixed issue #18932: 404 error after importing responses from a old response table

gabrieljenik:
Apparently the problem is that the Responses index page has been removed. 
A redirect statement was missing to be updated


Shnoulle (approved):
:+1:

PR #3274, opened 2023-07-05
QE-297 refactor question group controller

twilligls:
Dev:


twilligls:
WIP

application/models/services/QuestionGroupService.php
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        $questionGroup = QuestionGroup::model()->findByPk($questionGroupID);
+        if (!$questionGroup) {
+            throw new Exception(
+                'Group not found'
+            );
+        }
+
+        $questionGroup = $this->editQuestionGroup($questionGroup, $input['questionGroup']);
```
olleharstedt, line 66: Maybe `updateQuestionGroup` is a better name?

application/models/services/QuestionGroupService.php
```diff
@@ -0,0 +1,265 @@
...
+        $survey = $this->getSurvey($surveyId);
+
+        $questionGroup = new QuestionGroup('search');
+
+        if (isset($_GET['QuestionGroup']['group_name'])) {
+            $questionGroup->group_name = $_GET['QuestionGroup']['group_name'];
+        }
+
+        if (isset($_GET['pageSize'])) {
+            App()->user->setState('pageSize', (int)$_GET['pageSize']);
```
olleharstedt, line 243: Inject app or user in constructor instead.

application/models/services/QuestionGroupService.php
```diff
@@ -0,0 +1,265 @@
...
+     * @return QuestionGroup
+     */
+    public function getGroupData($surveyId)
+    {
+        $survey = $this->getSurvey($surveyId);
+
+        $questionGroup = new QuestionGroup('search');
+
+        if (isset($_GET['QuestionGroup']['group_name'])) {
+            $questionGroup->group_name = $_GET['QuestionGroup']['group_name'];
```
olleharstedt, line 239: Globals such as `_GET` should not be accessed, better to inject the request object.

application/controllers/QuestionAdministrationController.php
```diff
@@ -336,7 +336,7 @@ public function actionListQuestions($surveyid, $landOnSideMenuTab = 'settings')
         $questionGroupService = $diContainer->get(
             LimeSurvey\Models\Services\QuestionGroupService::class
         );
-        $aData['groupModel'] = $questionGroupService->getGroupData($iSurveyID);
+        $aData['groupModel'] = $questionGroupService->getGroupData($iSurveyID, $aData['oSurvey']);
```
olleharstedt, line 339: Hm why would you need both survey id and the survey model? Survey model contains id.

twilligls: Ah, that's an unnecessary leftover...

application/models/services/QuestionGroupService.php
```diff
@@ -0,0 +1,265 @@
...
+     *                     ['questionGroupI10N']
+     *                         [en]
+     *                            [group_name]
+     *                            [description]
+     *                         [...]    //more languages
+     * @return QuestionGroup
+     */
+    public function updateGroup($surveyId, $questionGroupID, $input)
+    {
+        $survey = $this->getSurvey($surveyId);
```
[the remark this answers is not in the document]
kevin-foster-uk: @olleharstedt Often the service logic needs to load multiple objects from persistence in order to carry out its work. The objects to be loaded may even depend on some initial logic to work out which objects we are need to work on.

This logic does not belong in the controller. 

There may be an argument for organising the business logic in order to isolate this work but the refactoring we are doing right now is minimal work in order to achieve two things:

1. Reusability
2. Testability

Anything refactoring beyond that is out of scope.

application/models/services/QuestionGroupService.php
```diff
@@ -18,22 +19,31 @@ class QuestionGroupService
 {
     private Permission $modelPermission;
     private Survey $modelSurvey;
-    private CWebUser $modelUser;
-    private CHttpRequest $modelRequest;
+    private Question $modelQuestion;
+    private QuestionGroup $modelQuestionGroup;
+    private QuestionGroupL10n $modelQuestionGroupL10n;
+    private CWebUser $yiiUser;
```
olleharstedt: Do we need the prefix `yii`?  Same with `yiiRequest` below. What else would it be except yii?

application/models/services/QuestionGroupService.php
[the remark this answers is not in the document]
olleharstedt: In this case, the loading of the survey does not depend on any logic tho. So the controller code could easily be:

```php
$service->updateGroup(Survey::model()->findByPk($surveyId), $questionGroupId, $input)
```

Function composition, pretty basic. :) But I've made this argument before, and you didn't agree then either.

application/models/services/QuestionGroupService.php
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * Deletes a question group and all its dependencies.
+     *
+     * @param int $questionGroupId the question group id
+     * @param int $surveyId the survey id
+     * @return int|null number of deleted rows
+     * @throws CHttpException
+     */
+    public function deleteGroup(int $questionGroupId, int $surveyId)
```
olleharstedt, line 128: If you pass in QuestionGroup model directly instead of just the id, you might not need to inject the QuestionGroup modal in the constructor. Again, @kevin-foster-uk will probably not agree. Just saying, having so many dependencies in the constructor when they can be given as method arguments instead might be preferable at some point.

application/models/services/QuestionGroupService.php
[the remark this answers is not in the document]
kevin-foster-uk: Sure in this case its relatively simply in other cases its not. Regardless, in all cases, this logic for loading objects does not belong in the controller.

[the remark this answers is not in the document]
olleharstedt: I disagree wholeheartedly. :) In this case, it very much belongs to the _setup_ phase of the service, which is in the controller.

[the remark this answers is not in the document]
kevin-foster-uk: @olleharstedt 

The point of the application model (which are implementing as services) is that it abstracts away the business logic of the application. It is a context agonistic (http, cli, unit test) boundary point. It allows us to interact with the application via an API in the same way that we might consume any other API. 

When designing an API the main focus is the interface. How does the consumer interact with the API? The ideal implementation will keep the API as simple as possible. The less the consumer needs to know about the internal workings of the API the better. 

When you interact with the business model you should not need to be aware of any implementation details. You should not need to know about HOW data is persisted in order to interact with it. The implementation details of data persistence is handled by the model. 

As I have said, there is an argument for different approaches to separating persistence logic from pure business logic. Regardless of how you handle that separation, it should still be handled within model, since its not a concern of the consumer of the model. The consumer just wants to add a question or update a survey setting. They don't care HOW you do it.

Yes I agree is there code smell, but thats not an argument for putting persistence logic in the controller and duplicating it in unit tests and cli scripts and anywhere else we want to use the model.

application/models/services/QuestionGroupService.php
kevin-foster-uk: As I was discussing with Tim yesterday, some of these dependancies should not being injected at all. Dependancies with the "yii" prefix fall in to the category and the yii prefix helps us easily identify them as we move forward in future refactoring.

The application model to be portable and not dependant on any specific MVC framework code or any code relating to a specific environment in which it could be run. 

To factor out all reference to Yii code would massively increase the effort and time required to complete this refactoring round. 

In short the yii prefix highlights that the dependancies is MVC framework specific and should be factored out in future.

application/models/services/QuestionGroupService.php
kevin-foster-uk: I agree that are ways the code can be improved. That isn't the focus on this round of refactoring. 

As I have said before. This refactoring round is not intended to fix all code smell. We have two goals:

1. Re-usability
2. Test ability

The process for refactoring has been this:

- Copy existing code into a service class
- If the class is too big break it into multiple composed classes
- Replace Yii MVC specific code with model specific code where possible (e.g. Request data, Exceptions)
- Inject dependancies to allow for testing
- Try to pass code-sniffer and psalm rules

I have not spent much time looking at the implementation of the business logic or how it could be improved other than when it was necessary in order to meet goal 1. and 2. above. The work has mostly been copying and pasting and re-arranging.

Again, I agree that are ways the code can be improved. That isn't the focus on this round of refactoring.

application/models/services/QuestionGroupService.php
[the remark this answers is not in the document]
olleharstedt: @kevin-foster-uk The controller is the _correct_ place to have persistence logic, since it's the "imperative shell" of the application. :D Service classes should have as few side-effects as possible, in my opinion - that makes them way easier to test and reuse in different contexts.

Your other arguments seem to agree with the point of lifting loading data out of the service class when possible, since it makes the service class _more_ agnostic of how it's loaded.

Duplicating the loading in unit test is not needed - it won't be loaded at all, it will just be a mock. Instead of mocking the loading, you mock the survey object directly. No loading needed.

But yes, some duplication will appear if you use the service class from CLI or REST. That's totally fine, the same type of duplication happens already with the DI setup, right?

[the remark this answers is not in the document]
kevin-foster-uk: @olleharstedt We are not designing an application from scratch here. We have some existing code, we are trying to make it reusable and testable. We can argue as much as you want about how it should be organised right now its irrelevant. This is what we have.

If you want to take over directing the refactoring tasks and take it in a new direction we can discuss that.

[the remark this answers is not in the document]
olleharstedt: Just a simple suggestion to improve testability of the service class, Kevin. :) We can mark it as resolved if we can't agree.

application/models/services/QuestionGroupService.php
```diff
@@ -0,0 +1,434 @@
...
+     * @return QuestionGroup
+     * @throws NotFoundException
+     * @throws PermissionDeniedException
+     * @throws PersistErrorException
+     */
+    public function updateGroup(int $surveyId, int $questionGroupId, array $input)
+    {
+        $survey = $this->getSurvey($surveyId);
+
+        if (!$this->modelPermission->hasSurveyPermission($survey->sid, 'surveycontent', 'update')) {
```
kevin-foster-uk, line 82: > The PSR-12 standard doesn't impose a hard limit on line length, but it recommends a soft limit of 120 characters, and to keep lines under 80 characters where possible.

If you look at my service classes. I break code down to keep lines less than 80 characters. This does increase line length but I think it helps read ability. It doesn't make sense to make these changes when making a small change in a file but when creating a bran new class we should follow this.

I wont go through all of the code and make the same comment but this involves updating a lot of lines.

application/controllers/QuestionGroupsAdministrationController.php
```diff
@@ -100,47 +103,31 @@ public function actionView(int $surveyid, int $gid, $landOnSideMenuTab = 'struct
...
+        $aData = $this->setSurveyIdAndObject([], $surveyid);
+        $aData['gid'] = $gid;
+        $aData['condarray'] = getGroupDepsForConditions($surveyid, "all", $gid, "by-targgid");
         //check if group with the gid exists
-        $questionGroup = QuestionGroup::model()->findByPk($gid);
-        if ($questionGroup === null) { //group does not exists ...
+        try {
+            $questionGroupService = $this->getQuestionGroupServiceClass();
+            $aData['oQuestionGroup'] = $oQuestionGroup = $questionGroupService->getQuestionGroupObject($surveyid, $gid);
+        } catch (Exception $e) {
```
Shnoulle, line 116: If you catch an exception : youn can not be sure the error was "Question group does not exists"

Must use $e , no ?

application/controllers/QuestionGroupsAdministrationController.php
```diff
@@ -438,80 +272,33 @@ public function actionListquestiongroups($surveyid)
...
-        $iSurveyID = $surveyid = $aData['surveyid'] = (int) $_POST['sid'];
-        $survey = Survey::model()->findByPk($iSurveyID);
+        $action = App()->request->getPost('action', '');
+        $aData = $this->setSurveyIdAndObject([], (int) App()->request->getPost('sid', null));
 
-        if (!Permission::model()->hasSurveyPermission($surveyid, 'surveycontent', 'import')) {
+        if (!Permission::model()->hasSurveyPermission($aData['surveyid'], 'surveycontent', 'import')) {
             App()->user->setFlash('error', gT("Access denied"));
-            $this->redirect(array('questionGroupsAdministration/listquestiongroups/surveyid/' . $surveyid));
+            $this->redirect(array('questionAdministration/listQuestions/surveyid/' . $aData['surveyid']));
```
Shnoulle, line 280: We can not be redirected to good tab ? with # at end of URL ?

application/controllers/QuestionGroupsAdministrationController.php
```diff
@@ -532,31 +319,29 @@ public function actionImport()
...
-                . " (" . gT("ID") . ":" . $iSurveyID . ")";
+            $aData['title_bar']['title'] = $aData['oSurvey']->currentLanguageSettings->surveyls_title
+                . " (" . gT("ID") . ":" . $aData['surveyid'] . ")";
 
             $this->aData = $aData;
             $this->render('importGroup_view', $aData);
         } else {
             App()->user->setFlash('error', gT("Access denied"));
-            $this->redirect(array('questionGroupsAdministration/listquestiongroups/surveyid/' . $surveyid));
+            $this->redirect(array('questionAdministration/listQuestions/surveyid/' . $aData['surveyid']));
```
Shnoulle, line 344: To tab of group ?

application/controllers/QuestionGroupsAdministrationController.php
twilligls: The exception error message could be not so user friendly. Of course we can not be 100% sure what exception was thrown. But the most obvious one is the NotFoundException

Shnoulle: I don't look fiurther, but why not use something similar than model->getErrors ?

application/models/services/QuestionGroupService.php
```diff
@@ -251,7 +264,10 @@ public function importQuestionGroup(int $surveyId, string $tmpDir, string $trans
             || $_FILES['the_file']['error'] == 2
         ) {
             $fatalerror = sprintf(
-                gT("Sorry, this file is too large. Only files up to %01.2f MB are allowed."),
+                gT(
+                    'Sorry, this file is too large. '
+                    . 'Only files up to %01.2f MB are allowed.'
+                ),
```
twilligls, line 270: Doesn't that break our translation functionality?

kevin-foster-uk: No. The concatenated string is the same.

olleharstedt: Hm what about the script that scans the codebase looking for strings inside gT()?

PR #3276, opened 2023-07-07
Fixed issue #18942: Roles: Permissions Modal is too narrow

gabrieljenik:


Shnoulle:
Great .

If i don't make error : we add a system for model-lg ?

Why not modal-xl here ?
https://getbootstrap.com/docs/5.3/components/modal/#optional-sizes

Shnoulle:
Or better  : https://getbootstrap.com/docs/5.3/components/modal/#fullscreen-modal

`modal-fullscreen-xxl-down`

Must test live

Shnoulle:
I test with `modal-fullscreen`  i really like it.

gabrieljenik:
> I test with `modal-fullscreen` i really like it.

I would leave that to @jorynag3l He is more into UI decisions.
I was just trying to fix the bug

PR #3278, opened 2023-07-10
New feature: Export responses by multiple tokens

dcore94:
Make it possible to optionally export responses with several tokens as filter instead of only one.

Dev:


dcore94:
This helps us reduce by one order of magnitude the number of REST calls in some scenarios.

olleharstedt:
Looks good. I assume it doesn't break backwards compatibility, since string is still accepted as argument for token?

olleharstedt:
Would it be possible to write a unit test for this new feature? There are many tests already written for the RPC API, which you can look at for inspiration.

dcore94:
> Looks good. I assume it doesn't break backwards compatibility, since string is still accepted as argument for token?

We are temporarily using a version of LimeSurvey with this change built in and we do not experience any break of backward compatibility.

olleharstedt:
Thanks for adding the unit test and for you contribution to LimeSurvey! :)

olleharstedt:
This feature will be released in LimeSurvey 6.3.0, by the way.

PR #3282, opened 2023-07-12
Fixed issue #18937: Action menu stop work after any action on any user

gabrieljenik:
That method is from the GridActionsWidget, and from what I understand there could be tables using CLSGridView without having actions with that widget.


ptelu:
Hey Gabriel i updated the pull request with some changes, i already had this issue marked for me internally so i check your commit and the issue and updated the code accordingly.

twilligls (approved):
LGTM

PR #3283, opened 2023-07-12
Fixed issue #18876: Plugin expose full path when an error is reported in the plugin list

gabrieljenik:


Shnoulle:
Same fix then 5?

Shnoulle (approved):

gabrieljenik:
> Same fix then 5?

Yes.

PR #3285, opened 2023-07-13
Fixed issue #18832: Participants can't re-enter a submitted survey if setting is using "inherit" at participant settings

gabrieljenik:
- Adding oOptions to the conditions
- Placing some conditions as model methods (this is to start giving entity to some business rules)


Shnoulle (approved):
:+1:

Shnoulle:
**Not related to this issue** : Looking at code i think we must check Valid DFrom and Valid Until too

If someone want to check and report.

application/models/TokenDynamic.php
```diff
@@ -1105,4 +1105,30 @@ public function getSurveyId()
     {
         return self::$sid;
     }
+
+    /**
+     * Returns true if the token can be used
+     * @return bool
+     */
+    public function canBeUsed()
```
Shnoulle, line 1113: :+1:

application/models/TokenDynamic.php
```diff
@@ -1105,4 +1105,30 @@ public function getSurveyId()
...
+    {
+        return !empty($this->token)
+            && (
+                $this->completed == "N"
+                || empty($this->completed)
+                || $this->survey->oOptions->alloweditaftercompletion == "Y"
+            );
+    }
+
+    public function canBeEmailed()
```
Shnoulle, line 1123: :+1:

application/models/TokenDynamic.php
```diff
@@ -1105,4 +1105,30 @@ public function getSurveyId()
...
+    {
+        return !empty($this->token)
+            && $this->emailstatus == "OK"
+            && $this->email
+            && $this->completed == "N"
```
Shnoulle, line 1128: `( $this->completed == "N" || empty($this->completed)))`

Same than can be used

application/models/TokenDynamic.php
```diff
@@ -1105,4 +1105,30 @@ public function getSurveyId()
...
+    /**
+     * Returns true if the token can be used
+     * @return bool
+     */
+    public function canBeUsed()
+    {
+        return !empty($this->token)
+            && (
+                $this->completed == "N"
+                || empty($this->completed)
```
Shnoulle, line 1118: No `$this->usesleft > 0 `check ? (same than email)

Shnoulle (approved):
OK, but maybe some test improvment.

gabrieljenik:
> OK, but maybe some test improvment.

Yes, will add them later on another PR

application/models/TokenDynamic.php
gabrieljenik: That was ported from the gridButtons part.
I think that could be part of `canBeUsed` but have some doubts about scenarios, ex: editing a prior response.
Anyway, don't want to update anything now, jusr reog a bit the code to start cetralizing business rules.

application/models/TokenDynamic.php
gabrieljenik: Yes, there seems to be some relationship. So far, don't want to make further changes until we have clear scenarios and better tests in place

Shnoulle:
> > OK, but maybe some test improvment.
> 
> Yes, will add them later on another PR

My question are something between : «Need to report a DEV issue on mantis ?» and «Maybe we can fix it quickly now»

No strong opinion. approved the update ;)

Shnoulle (approved):
better !

PR #3286, opened 2023-07-13
Fixed issue #18936: User count in group is not OK after deleting a user

gabrieljenik:
- Moved code to a new function on a service class
- Added the "remove from groups" operation


Shnoulle (approved):
No problem in code, bu we really need all this code ?

gabrieljenik:
> bu we really need all this code ?

Well, we moved some stuff from controller to a service class.
We needed a way to communicate in between both.
Did something reusable for it.

Shnoulle:
> Did something reusable for it.

A lot of things reusable never reused in LS … :(

Maybe we need page for reusable part in manual ?

gabrieljenik:
> A lot of things reusable never reused in LS … :(

Keep the faith (and use it :) )

PR #3288, opened 2023-07-14
Fixed issue #18888: General issues with different questions

gabrieljenik:
- Fixed Ranking Advanced

### DIFF
--- a/application/core/QuestionTypes/RankingStyle/RenderRanking.php
+++ b/application/core/QuestionTypes/RankingStyle/RenderRanking.php
@@ -76,9 +76,10 @@ class RenderRanking extends QuestionBaseRenderer
                     'optiontext' => gT('Please choose...'),
                 );
             }
-    
+
+            $this->aDisplayAnswers = [];
             foreach ($this->aAnswerOptions[0] as $oAnswer) {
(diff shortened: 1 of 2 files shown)


application/core/QuestionTypes/RankingStyle/RenderRanking.php
```diff
@@ -76,9 +76,10 @@ public function getRows()
                     'optiontext' => gT('Please choose...'),
                 );
             }
-    
+
+            $this->aDisplayAnswers = [];
             foreach ($this->aAnswerOptions[0] as $oAnswer) {
-                $this->aDisplayAnswers[$oAnswer->aid] = array_merge($oAnswer->attributes, $oAnswer->answerl10ns[$this->sLanguage]->attributes);
+                $this->aDisplayAnswers[] = array_merge($oAnswer->attributes, $oAnswer->answerl10ns[$this->sLanguage]->attributes);
```
Shnoulle, line 82: We can not fix it in twig ?

Shnoulle (changes requested):
It's a Question theme issue : must be fixed in theme , not in core.

Why use key in theme ? Just use a counter …

PR #3289, opened 2023-07-14
Fixed issue #18573 Insert new box with overlapping position, makes the dashboard to show duplicate boxes

gabrieljenik:
Unique validator added to model rules.


Shnoulle (approved):
:+1:

PR #3290, opened 2023-07-19
Fixes issue #18958: Multiple numerical input text input box width not having an effect

adamzammit:
Fixed issue #18958: Multiple numerical input text input box width not having an effect
Dev: Will require a separate fix if you want to migrate from older LS / LSS files to ensure question setting text_input_width converted to text_input_columns as expected in LS6


tiborpacalat:
@olleharstedt Alternative solution was merged in here https://github.com/LimeSurvey/LimeSurvey/pull/3297. Should we close this one?

adamzammit:
Closed as alternative patch applied

PR #3292, opened 2023-07-19
Fixed issue [security] #18967: User can add malicious content in User Roles

Shnoulle:
Dev: encode html in view
Dev: encode html for message


Shnoulle:
Here encode, but maybe we can filter.

I don't think it broke semver API, but ????  It's always possible for some workaround

application/models/Permissiontemplates.php
```diff
@@ -214,7 +214,7 @@ public function getButtons(): string
                 'data-bs-toggle' => "modal",
                 'data-post-url'  => $deleteUrl,
                 'data-post-datas' => $deletePostData,
-                'data-message'   => sprintf(gt("Are you sure you want to delete user role '%s'?"), $this->name),
+                'data-message'   => CHtml::encode(sprintf(gt("Are you sure you want to delete user role '%s'?"), CHtml::encode($this->name))),
```
gabrieljenik, line 217: Isn't this double encoded? Would this create an error?

Shnoulle: There are an XSS vector attack in javascript part , js part remove the 1st encode.

See https://discord.com/channels/757925259885412413/760146382790197248/1131251472777093241

gabrieljenik: Not sure I follow.

My question is:
- If  `this->name` is `Gabriel "GJ"`.
- First Encode will make it `Gabriel &quot;GJ&quot;`
- Second Encode will make it `Gabriel &amp;quot;GJ&amp;quot;`
right?

Shnoulle: > Not sure I follow.
> 
> My question is:
> 
>     * If  `this->name` is `Gabriel "GJ"`.
> 
>     * First Encode will make it `Gabriel &quot;GJ&quot;`
> 
>     * Second Encode will make it `Gabriel &amp;quot;GJ&amp;quot;`
>       right?

Yes, but when show to user (via JS ) come to Gabriel "GJ" (HTML encoded).

Shnoulle: > Are you referring to this? https://github.com/LimeSurvey/LimeSurvey/blob/master/assets/packages/adminbasics/src/parts/confirmationModal.js#L115
> 
> Somehow I feel that if we need to double encode, something is off. Also, that could be spreaded across all site.
> 
> What if we encode on the confirmation modal? Or we could do .text() instead of .html() to inside the modal.

OK :)

I got it : we mlust fix this 

I update .

Not needed in 5.X

gabrieljenik: What's different in 5.x? How it was tackled?

Shnoulle: > What's different in 5.x? How it was tackled?

No usage of name : https://github.com/LimeSurvey/LimeSurvey/blob/73991364056de70405b7158bac1b70a11633b1ee/application/models/Permissiontemplates.php#L221

Adding name is better for UX (For once there is positive user experience on version 6...).

Shnoulle: > What if we encode on the confirmation modal? Or we could do .text() instead of .html() to inside the modal.

Done use text.
using text is same than encoding result and send it as html.

gabrieljenik (approved):

PR #3296, opened 2023-07-20
Fixed issue #18278 Quick translation: "Nothing to translate" appears wrongly.

gabrieljenik:


application/views/quickTranslation/translatetabs_view.php
```diff
@@ -65,8 +68,8 @@ class='ajax-loader' alt='<?php eT("Loading..."); ?>'/>
     </div>
     <?php
     if (isset($singleTabFieldsData)) {
-        if ($singleTabFieldsData[0]['all_fields_empty']) : ?>
-            <p><?php eT("Nothing to translate on this page"); ?></p><br/>
+        if ($allFieldsEmpty) : ?>
+            <p><?php eT("Nothing to translate on this page.."); ?></p><br/>
```
tiborpacalat, line 72: why was this copy changed? you added two dots at the end, which will invalidate every existing translation of this. Please change it back.

tiborpacalat (changes requested):

application/views/quickTranslation/translatetabs_view.php
gabrieljenik: Nice catch. Sorry.

tiborpacalat (approved):

PR #3299, opened 2023-07-21
Updated translation

tiborpacalat:
Updated translation strings to be conform to our style guide.

Dev:

### DIFF
--- a/themes/survey/fruity_twentythree/views/subviews/content/save.twig
+++ b/themes/survey/fruity_twentythree/views/subviews/content/save.twig
@@ -107,7 +107,7 @@
                                 'id'          : 'savename',
                                 'name'        : 'savename',
                                 'class'       : 'form-control ls-important-field',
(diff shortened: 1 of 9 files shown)


themes/survey/fruity_twentythree/views/subviews/content/save.twig
```diff
@@ -186,7 +186,7 @@
                                 'id'          : 'saveemail',
                                 'name'        : 'saveemail',
                                 'class'       : 'form-control ls-important-field',
-                                'placeholder' : gT('enter e-mail address'),
+                                'placeholder' : gT('Enter e-mail address'),
```
ptelu, line 189: shouldn't this and the 3 above have the same convention either have your or not first one has your

tiborpacalat: I don't want to change too much until we really have "style guide for copy" (copy = fancy wording for text)

PR #3303, opened 2023-07-24
Fixed issue [security] #18979: User can put survey in surveygroup without access permission

Shnoulle:
Dev: fix it in rules : must hack HTML


gabrieljenik (approved):
Not the best for performance if there are +1000 survye groups.
Still, fast way of implementing the rule. :)

PR #3304, opened 2023-07-24
Fixed issue [security] #18979: User can put survey in surveygroup without access permission

Shnoulle:
Dev: rules disable this (muts hack HTML)


Shnoulle:
Can not test currently :  no way to save general settings : https://bugs.limesurvey.org/view.php?id=18983

gabrieljenik (approved):
If the user has access to +1000 survey groups this could be non performant.
Not the best way to implement the rule, but works.

Shnoulle:
> If the user has access to +1000 survey groups this could be non performant. Not the best way to implement the rule, but works.


I start to do it in controller (checking incluedd in https://github.com/LimeSurvey/LimeSurvey/blob/751073b4f2744a7334bfd972163f614bbcb9ba2f/application/models/SurveysGroups.php#L417)

I can add a new SurveyGroups function and a new internal validator in Survey . Or add an option to get only sid in https://github.com/LimeSurvey/LimeSurvey/blob/751073b4f2744a7334bfd972163f614bbcb9ba2f/application/models/SurveysGroups.php#L347C27-L347C44

But Saving Survey happen not a lot of time currently.

gabrieljenik:
> I start to do it in controller (checking incluedd in

Actually, the controller may be a better place (or more usual on LS). Why not there?

Shnoulle:
> > I start to do it in controller (checking incluedd in
> 
> Actually, the controller may be a better place (or more usual on LS). Why not there?

Remote control, another controller (since there are a lot of new controller on a lot pof place now …)

Else : directly in rules seem interesting too (since we have an array).

gabrieljenik:
I would just try to keep consistency and do it in the controller.

application/controllers/admin/Database.php
```diff
@@ -364,12 +364,15 @@ protected function actionUpdateSurveyLocaleSettingsGeneralSettings($surveyId)
             'admin' => $request->getPost('admin'),
             'adminemail' => $request->getPost('adminemail'),
             'bounce_email' => $request->getPost('bounce_email'),
-            'gsid' => $request->getPost('gsid'),
             'format' => $request->getPost('format'),
             'owner_id' => $request->getPost('owner_id'),
             'template' => $request->getPost('template')
         ];
-
+        if (SurveysGroups::model()->permission(App()->user->getId())->findByPk(intval($request->getPost('gsid')))) {
```
gabrieljenik, line 371: You will probably don't like my comment, still, I feel need to.
On this line we are intending a business rule: Does the user has permission for the survey group?
Shouldn't we enclose that on a method somewhere so we can better reuse it in the future and enhance it if needed?

Shnoulle: >  On this line we are intending a "business" rule / "access" rule: Does the user has permission for the survey group?

It's in model, permission scope … done for this, like Survey->permission
Like `SurveyGroup::getSurveyGroupsList`

>  Shouldn't we enclose that on a method somewhere so we can better reuse it in the future and enhance it if needed?

It's in scope : it's **clearly** sufficient. 

We can just move it to PermissionTrait and Interface. **But** when i want to add it : after a long , too long discussion i remove it and close the PR. You can not ask now to add it again … 

And here !: it's a **FIX** , it's not a rewrite … you want i check some previous PR where need

Shnoulle:
> You will probably don't like my comment, still, I feel need to.

It's not i don't like, but clearly : what is the purpose ? I don't understand here … 

Agin : it's a fix, you are the 1st to say on your PR : i do a microfix, other is for dev …

application/controllers/admin/Database.php
Shnoulle: Ese : before rewrite something clean in model : need to move all admin/database.php elwewhere …

Shnoulle:
> I would just try to keep consistency and do it in the controller.

No consistency in controller : 

1. currently no control is done, just try/catch
2. rules already validate XSS user related fix.

I move it agin in  model->rules

Shnoulle:
Test and code review already doine

Shnoulle:
> > I start to do it in controller (checking incluedd in
> 
> Actually, the controller may be a better place (or more usual on LS). Why not there?

[18996: Move admin/Database controller to specific controller](https://bugs.limesurvey.org/view.php?id=18996)

gabrieljenik:
You arelady moved in a different direction, so this is just comments.

> It's not i don't like, but clearly : what is the purpose ? I don't understand here …

Purpose was to enclose a rule into a method that can be clearly identified, reused and get its own entity.
Later on, when having to reuse the rule, there is a clear place where it can be looked for.

> Agin : it's a fix, you are the 1st to say on your PR : i do a microfix, other is for dev …

I believe that adding a method is still not big deal for a bugfix and has a good level of effort/advanages balance.

Shnoulle:
> Purpose was to enclose a rule into a method that can be clearly identified, reused and get its own entity.

[scope](https://www.yiiframework.com/doc/guide/1.1/en/database.ar#named-scopes) (or rules) is clearly identified and is definitively the way to do.

gabrieljenik:
> > Purpose was to enclose a rule into a method that can be clearly identified, reused and get its own entity.
> 
> [scope](https://www.yiiframework.com/doc/guide/1.1/en/database.ar#named-scopes) (or rules) is clearly identified and is definitively the way to do.

scope() + findbyPk() doesn't sound like something with its own entity and which can be enhanced easily to turn the now simple rule into something more complex.

> is definitively the way to do

I guess we can have different views, right?

Shnoulle:
> I guess we can have different views, right?

Here : see yes.

About moving to controller : i create an issue.
Maybe we can link we some other Class/Service needed.

PR #3306, opened 2023-07-25
Fixed issue #18936: User count in group is not OK after deleting a user

gabrieljenik:
- Moved code to a new function on a service class
- Added the "remove from groups" operation


Shnoulle (approved):

application/models/services/UserManager.php
```diff
@@ -83,4 +85,57 @@ public function canEdit()
                 && $this->targetUser->parent_id == $this->managingUser->id
             );
     }
+
+    /**
+     * Deletes the user with the given id.
+     * @param int $userId
+     * @return OperationResult
+     */
+    public function deleteUser($userId)
```
olleharstedt, line 94: 1. If you already have target user in the constructor, why do you need $userId here?
2. Maybe target user is _not_ good to inject as a dependency in the constructor, but rather should be an argument?
3. I usually prefer to pass around models instead of ids to models:

```php
public function deleteUser(User $user)
```

PR #3308, opened 2023-07-26
Fixed issue [security] #18977: Improper Authorization in add role fun…

Shnoulle:
Fixed issue [security] #18977: Improper Authorization in add role function leads to privilege escalation
Dev: limited to superadmin (read)

Dev:

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -1159,7 +1159,8 @@ class UserManagementController extends LSBaseController
      */
     public function actionBatchApplyRoles()
     {
-        if (!Permission::model()->hasGlobalPermission('users', 'update')) {
+        /* Need super admin roles */
+        if (!Permission::model()->hasGlobalPermission('superadmin')) {
             return $this->renderPartial(
                 'partial/error',
                 ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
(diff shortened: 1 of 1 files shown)


application/controllers/UserManagementController.php
```diff
@@ -1174,8 +1175,8 @@ public function actionBatchApplyRoles()
             $model = $this->loadModel($sItem);
             $aResults[$sItem]['title'] = $model->users_name;
 
-            //check if user is admin otherwhise change the role
-            if (intval($sItem) == 1) {  //todo REFACTORING is admin id always 1?? is there another possibility to check for admin user?
+            if (Permission::isForcedSuperAdmin($sItem)) {
```
gabrieljenik, line 1178: Is this part of the reported issue?
Code looks OK. I am just rising my hand as to make sure it is tested.

gabrieljenik (approved):

application/controllers/UserManagementController.php
Shnoulle: No :part of `//todo REFACTORING is admin id always 1?? is there another possibility to check for admin user?`

isFoircedAdmin return https://github.com/LimeSurvey/LimeSurvey/blob/49eab4bf643943d3b02f6f81420a70af055c11d4/application/config/config-defaults.php#L673 (by default).

But superadmin 1 can be deleted or not be a superadmin in 3,5 or 6.

Shnoulle: there are a TODO, the function is here since years , the dev who add it don't know this function (we need a better place for such function)

gabrieljenik: Nevermind...

Shnoulle: PS : i think this part can be removed … you can remove the superadmin permission to forcedsuperadmin : it stay a forcedsuperadmin :)

There are no difference if you don't update config.php settings here.

I can remove the fix if you prefer. But since there are a TODO with an existing solution done exactly for this : unsue .

PR #3309, opened 2023-07-26
Fixed issue [security] #18985: Reflected XSS

Shnoulle:
Dev: set variable to real final format.

Dev:

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -722,9 +722,9 @@ class UserManagementController extends LSBaseController
                 ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
             );
         }
-       if( !in_array($importFormat,['csv','json'])) {
-               throw new LSUserException(400, gT("Invalid format"));
-       }
+        if (!in_array($importFormat, ['csv', 'json'])) {
+            throw new LSUserException(400, gT("Invalid format"));
+        }
 
         switch ($importFormat) {
             case "json":
@@ -758,7 +758,9 @@ class UserManagementController extends LSBaseController
                 ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
             );
         }
-
+        if (!in_array($importFormat, ['csv', 'json'])) {
+            throw new LSUserException(400, gT("Invalid format"));
+        }
         $overwriteUsers = boolval(App()->getRequest()->getPost('overwrite'));
 
         switch ($importFormat) {
@@ -768,7 +770,6 @@ class UserManagementController extends LSBaseController
             case "csv":
             default:
                 $aNewUsers = UserParser::getDataFromCSV($_FILES);
-                $importFormat = 'csv'; /* Fix usage of importFormat in view */
         }
         if (empty($aNewUsers)) {
             Yii::app()->setFlashMessage(gT("No user definition found in file."), 'error');
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -722,6 +722,9 @@ class UserManagementController extends LSBaseController
                 ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
             );
         }
+       if( !in_array($importFormat,['csv','json'])) {
+               throw new LSUserException(400, gT("Invalid format"));
+       }
 
         switch ($importFormat) {
             case "json":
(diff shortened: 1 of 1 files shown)


gabrieljenik (commented):
Code seems OK. 

Still, where is the security fix being done? 
I am missing it, but I only see formatting stuff.

Shnoulle:
> Code seems OK.
> 
> Still, where is the security fix being done? I am missing it, but I only see formatting stuff.

`$importFormat = 'csv';`

importFormat is used in view. then : if it's not json : force ti be csv Can be used in view without issue.

application/controllers/UserManagementController.php
```diff
@@ -762,7 +765,8 @@ public function actionImportUsers(string $importFormat = 'csv'): string
                 break;
             case "csv":
             default:
-                $aNewUsers = UserParser::getDataFromCSV($_FILES); //importFormat default is csv ...
+                $aNewUsers = UserParser::getDataFromCSV($_FILES);
+                $importFormat = 'csv'; /* Fix usage of importFormat in view */
```
gabrieljenik, line 769: Probably this works,
Still, I believe the inteded behaviour is initializing the importFormat variable if it is not json nor CSV. right?
If that's the case, I feel the current implementation could be missleading/confusing.
I would feel a more clasic approach would be better.
`if (empty($importFormat)) $importFormat = 'csv'`

Then, also if no import format matches in the swithc, we could throw an error.
Why default is csv?

Shnoulle: >  if (empty($importFormat)) $importFormat = 'csv'

Not : must be 

`if( !in_array($importFormat,['csv','json']) {` because here importFormat came from URL (look at the issue)

At start i think i do such fix
```
if( !in_array($importFormat,['csv','json']) {
	throw new CHttpException(400);
}
```
But since previous version allow anything but force file to be csv (same after upload) : i allow it too.

You think it's better ?

Shnoulle: > I would say, to initilize the variable to 'csv' if it comes empty (as to have clean inside the function) and that's it. Just to avoid future XSS in this function if someone updates it.

I don't understand : if user create the url with userManagement/renderUserImport?importFormat=HACKED


You want to shown «Select HACKED file [file input]» ?

Can you explain what do you think it's better exacltly with example …



- userManagement/renderUserImport?importFormat=HACKED : Throw a 403 Yes or No
- userManagement/renderUserImport?importFormat=HACKED : show «Please choose HACKED file» Yes or No

gabrieljenik: This is what I picture... makes sense to you?

Before line 760:
```
// Initialize $importFormat if not given. 
if (empty($importFormat)) {
   $importFormat = 'csv'
}
```

Then 
```
switch ($importFormat) {
    case "json":
        $aNewUsers = UserParser::getDataFromJSON($_FILES);
        break;
    case "csv":
        $aNewUsers = UserParser::getDataFromCSV($_FILES);
        break;
    default:
       // If no matching format, throw error
       throw new CHttpException(400);
}
```

About the error, maybe it is better to use LSUserxception?

Shnoulle: Initilizing is not needed : include in function,

Then i think a 

```
if( !in_array($importFormat,['csv','json']) {
	throw new LSUserxception(400);
}
```

is more clear.
I update only this.

gabrieljenik (approved):

Shnoulle:
> The following block,
> 
> ```
>        if( !in_array($importFormat,['csv','json'])) {
>                throw new LSUserException(400, gT("Invalid format"));
>        }
> ```
> 
> shouldn't be placed here as well? https://github.com/LimeSurvey/LimeSurvey/pull/3309/files?diff=unified&w=1#diff-6f1c5976e0d24b8aa7349798b9a06c59081576fef190ed5a0a7a7cf6a947caedR761

You can not ask  to fix only a specific issue on other PR and here add unrelated commit.

It's not related to this issue, it's done when import.

No reason to update

gabrieljenik:
> You can not ask to fix only a specific issue on other PR and here add unrelated commit.

I have just noticed it. If you feel it not worth it, don't add it.

Shnoulle:
> > You can not ask to fix only a specific issue on other PR and here add unrelated commit.
> 
> I have just noticed it. If you feel it not worth it, don't add it.

Please review the last version.

gabrieljenik (approved):

PR #3312, opened 2023-07-26
Fixed issue #18832: Participants can't re-enter a submitted survey if setting is using "inherit" at participant settings

gabrieljenik:
- Adding oOptions to the conditions
- Placing some conditions as model methods (this is to start giving entity to some business rules)


gabrieljenik:
@Shnoulle We are noticing the condition for showing buttons are different in between master and v5.
Ex: SurveyActive being checked. Which version is OK?

Also, this seems redundant, right?
https://github.com/LimeSurvey/LimeSurvey/pull/3312/files#diff-b3f19f674a1168bc407eb88ba80a9f01ab555fa5dee7b7b7051bfa6c54558c5fR839

application/models/TokenDynamic.php
```diff
@@ -1079,4 +1063,30 @@ public function getSurveyId()
...
+            );
+    }
+
+    public function canBeEmailed()
+    {
+        return !empty($this->token)
+            && $this->emailstatus == "OK"
+            && $this->email
+            && $this->completed == "N"
+            && (
```
Shnoulle, line 1087: replace by && $this->canBeused()

Shnoulle (approved):
Some simplifciatoion can be great : canbeMailed can use canBeUsed

application/models/TokenDynamic.php
gabrieljenik: As to avoid complexity, and release the fix, i will do that on another ticket.
Also, please bear in mind the usesLeft on mailed but not on canBeused

gabrieljenik: Created this: https://bugs.limesurvey.org/view.php?id=18993

Shnoulle: > Also, please bear in mind the usesLeft on mailed but not on canBeused

Right, but to be sended : it seems clar it must be used ;) : and if useLeft > 0 : can be used , no ?

Shnoulle (approved):

gabrieljenik:
Tests added

PR #3315, opened 2023-07-28
Fixed issue #CR-1265: fileupload question rtl

twilligls:
Dev:


assets/survey_themes/fruity_twentythree/questiontypes/file_upload/file_upload.scss
```diff
@@ -1,3 +1,7 @@
 .file-preview fieldset span {
     margin-left: 0;
+}
+
+.uploadedfiles tbody tr td {
+    text-align: left;
```
Shnoulle, line 6: Needed ? Inherited value are not OK ?

Shnoulle (commented):
The issue came from https://github.com/LimeSurvey/LimeSurvey/blob/3ee3a558a71a34da2be44eb4976b05b5bd924acd/assets/packages/questions/upload/styles/uploader.css#L91 

It must be fixed in this assets.
Not by theme 

(i think we wan remove a lot of text-align etc … since we have BS_5 now (and https://getbootstrap.com/docs/5.0/getting-started/rtl/ )

assets/survey_themes/fruity_twentythree/questiontypes/file_upload/file_upload.scss
twilligls: It would be center if inherited

Shnoulle: Sicne the text-aligne: left came from a specific package : must be fixed in this specific package.

Else : it was centered in 5.X : it's an issue ?

twilligls: 5x also has align left for file upload table.
I'd love to use bootstrap classes, but the way RTL is handled in the new theme vs the old themes, makes it impossible to use those classes.

Shnoulle: Yes, but not with skelvanilla.

Shnoulle: Still : since text-align is set in https://github.com/LimeSurvey/LimeSurvey/blob/3ee3a558a71a34da2be44eb4976b05b5bd924acd/assets/packages/questions/upload/styles/uploader.css#L91 rtl system must be set in https://github.com/LimeSurvey/LimeSurvey/blob/3ee3a558a71a34da2be44eb4976b05b5bd924acd/assets/packages/questions/upload/styles/uploader.css#L91

Shnoulle:
Thank you :)

PR #3316, opened 2023-07-28
Fixed issue #18906: Unable to fix or edit survey with UpperCase templ…

Shnoulle:
Fixed issue #18906: Unable to fix or edit survey with UpperCase template name
Dev: set to existing directory/folder

PR #3319, opened 2023-07-31
Fixed issue #18906: Unable to fix or edit survey with UpperCase templ…

Shnoulle:
Fixed issue #18906: Unable to fix or edit survey with UpperCase template name
Dev: set to existing directory/folder

Dev:

### DIFF
--- a/application/models/Template.php
+++ b/application/models/Template.php
@@ -147,8 +147,8 @@ class Template extends LSActiveRecord
         /* Validate if template is OK in user dir, DIRECTORY_SEPARATOR not needed "/" is OK */
         $oTemplate = self::model()->findByPk($sTemplateName);
 
-        if (is_object($oTemplate) && $oTemplate->checkTemplate() && self::checkTemplateXML($oTemplate->folder)) {
-            self::$aNamesFiltered[$sTemplateName] = $oTemplate->folder;
+        if (is_object($oTemplate) && $oTemplate->checkTemplate() && $oTemplate->name === $sTemplateName && self::checkTemplateXML($oTemplate->folder)) {
+            self::$aNamesFiltered[$sTemplateName] = $sTemplateName;
             return self::$aNamesFiltered[$sTemplateName];
         }
 
--- a/application/models/Template.php
+++ b/application/models/Template.php
@@ -147,8 +147,8 @@ class Template extends LSActiveRecord
         /* Validate if template is OK in user dir, DIRECTORY_SEPARATOR not needed "/" is OK */
         $oTemplate = self::model()->findByPk($sTemplateName);
 
-        if (is_object($oTemplate) && $oTemplate->checkTemplate() && (self::checkTemplateXML($oTemplate->folder))) {
-            self::$aNamesFiltered[$sTemplateName] = $sTemplateName;
+        if (is_object($oTemplate) && $oTemplate->checkTemplate() && self::checkTemplateXML($oTemplate->folder)) {
+            self::$aNamesFiltered[$sTemplateName] = $oTemplate->folder;
             return self::$aNamesFiltered[$sTemplateName];
         }
 


application/models/Template.php
```diff
@@ -147,8 +147,8 @@ public static function templateNameFilter($sTemplateName)
         /* Validate if template is OK in user dir, DIRECTORY_SEPARATOR not needed "/" is OK */
         $oTemplate = self::model()->findByPk($sTemplateName);
 
-        if (is_object($oTemplate) && $oTemplate->checkTemplate() && (self::checkTemplateXML($oTemplate->folder))) {
-            self::$aNamesFiltered[$sTemplateName] = $sTemplateName;
+        if (is_object($oTemplate) && $oTemplate->checkTemplate() && self::checkTemplateXML($oTemplate->folder)) {
+            self::$aNamesFiltered[$sTemplateName] = $oTemplate->folder;
```
gabrieljenik, line 151: I am not to sure about this.
I would like to have someone else review this as well.

By returning the foldername, somehow:
1 - the intention of the method seems to be changed.
2 - folder and templatename may match so far, but i have my doubts about exchanging them so freely.

Are we sure the source of the problem is here?

olleharstedt:
@Shnoulle Something to look at?

application/models/Template.php
Shnoulle: > Are we sure the source of the problem is here?

Yes : on linux MyTheme != mytheme, but on DB MyTheme == mytheme

The other way are are to check MyTheme != mytheme on DB too.

Shnoulle:
> @Shnoulle Something to look at?

What do you mean ?

Else i put a comment on https://github.com/LimeSurvey/LimeSurvey/pull/3316#issuecomment-1655269258

Shnoulle:
> @Shnoulle Something to look at?

You put **Need update** : but  what i have to update ? There are no decision here …

Shnoulle:
> > @Shnoulle Something to look at?
> 
> You put **Need update** : but what i have to update ? There are no decision here …

ping @olleharstedt

olleharstedt:
No idea.

gabrieljenik:
> No idea.

Maybe about this?
https://github.com/LimeSurvey/LimeSurvey/pull/3319#discussion_r1279689737

Shnoulle:
> > No idea.
> 
> Maybe about this? [#3319 (comment)](https://github.com/LimeSurvey/LimeSurvey/pull/3319#discussion_r1279689737)

Yes, but there are no decision done :)

1. Are  ThemeName and themename are a different theme : Yes/no
2. It can not be a different theme on windows server

Shnoulle:
I need to check : but i think https://github.com/LimeSurvey/LimeSurvey/pull/3738 fix the issue

application/models/Template.php
Shnoulle: I add `$oTemplate->name === $sTemplateName` in comparaiosn.

For information name and folder are same in object (we can remove folder and use name)

Shnoulle:
> I need to check : but i think #3738 fix the issue

Didn't fix it

gabrieljenik (approved):
Code structure looks good

application/models/Template.php
```diff
@@ -147,7 +147,7 @@ public static function templateNameFilter($sTemplateName)
         /* Validate if template is OK in user dir, DIRECTORY_SEPARATOR not needed "/" is OK */
         $oTemplate = self::model()->findByPk($sTemplateName);
 
-        if (is_object($oTemplate) && $oTemplate->checkTemplate() && (self::checkTemplateXML($oTemplate->folder))) {
+        if (is_object($oTemplate) && $oTemplate->checkTemplate() && $oTemplate->name === $sTemplateName && self::checkTemplateXML($oTemplate->folder)) {
```
gabrieljenik, line 150: This condition here ` $oTemplate->name === $sTemplateName` resembles a lot of a filter.
Have you thought about overriding findByPk on the template class?
In there, after finding we would apply this extra filter. Return null if doesn't full fills.

Maybe too wide of an update?

Shnoulle: > Have you thought about overriding findByPk on the template class?
No, it's a great idea !

> Maybe too wide of an update?

I don't think so, i create an update of this fix ? Your opinion ?

gabrieljenik: I don't understand the whole semantics of this change, but seems templatename case spelling is a central rule, so why not bringing it into the Model?

Initially in the findBYPk, but later on some validations...

> I don't think so, i create an update of this fix ? Your opinion ?

This is already merged, so maybe add it like a DEV PR?
I would say to go for it. This is one of those rules that can easily layer get lost in code but seems important...

Shnoulle: > I don't understand the whole semantics of this change, but seems templatename case spelling is a central rule, so why not bringing it into the Model?

I already answer it's a good idea :)

> This is already merged, so maybe add it like a DEV PR?

OK i create a DEV mantis issue

PR #3320, opened 2023-07-31
Fixed issue #18942: Roles: Permissions Modal is too narrow

gabrieljenik:
table-responsive class added.


Shnoulle:
We use a bigger modal in 6 ?
No ?

gabrieljenik:
> We use a bigger modal in 6 ? No ?

I think there are better things because of BS5.
I am just reusing what we already have in other screens as to keep consistency and proved recipes.

Shnoulle (approved):

PR #3321, opened 2023-08-01
QE-334 Task/op handler survey entity

Trischi80:
Dev: implementation for OpHandler entity survey, operation update; including testclass


application/libraries/Api/Command/V1/SurveyPatch/OpHandlerSurveyUpdate.php
```diff
@@ -0,0 +1,69 @@
...
+use LimeSurvey\Api\Transformer\TransformerInterface;
+use LimeSurvey\ObjectPatch\Op\OpInterface;
+use LimeSurvey\ObjectPatch\OpHandler\OpHandlerException;
+use LimeSurvey\ObjectPatch\OpHandler\OpHandlerInterface;
+use LimeSurvey\ObjectPatch\OpType\OpTypeUpdate;
+use Survey;
+
+class OpHandlerSurveyUpdate implements OpHandlerInterface
+{
+    protected ?TransformerInterface $transformer = null;
```
kevin-foster-uk, line 15: Transform should not be nullable since it is required.

application/libraries/Api/Command/V1/SurveyPatch/OpHandlerSurveyUpdate.php
```diff
@@ -0,0 +1,69 @@
...
+    protected string $entity;
+    protected CModel $model;
+
+    public function __construct(string $entity, CModel $model, TransformerInterface $transformer = null)
```
kevin-foster-uk, line 19: Again here transformer should not be nullable.

application/libraries/Api/Command/V1/SurveyPatch/OpHandlerSurveyUpdate.php
```diff
@@ -0,0 +1,69 @@
...
+        $props = $op ->getProps();
+        if ($props === null) {
+            throw new OpHandlerException(
+                printf(
+                    'No values to update for entity %s',
+                    $op->getEntityType()
+                )
+            );
+        }
+        $survey->setAttributes($props);
```
kevin-foster-uk, line 59: Need to use the transformer to transform the props before saving.

tests/unit/api/opHandlers/OpHandlerSurveyUpdateTest.php
kevin-foster-uk, line 1: When testing these new service classes we want to test only the new class we created in isolation. The test should not require a working install of the application or a database. We do this by creating an instance of the class we want to test and passing in mock objects for its dependancies.

Have a look at examples from mine or Tims code. We can have a call to discuss it more if that will help.

PR #3331, opened 2023-08-03
Fixed issue #CR-1253: There is no validation on admin email, bounce email in general survey settings

mohabmes:


application/controllers/admin/Database.php
```diff
@@ -369,6 +369,11 @@ protected function actionUpdateSurveyLocaleSettingsGeneralSettings($surveyId)
             'gsid' => $request->getPost('gsid'),
             'template' => $request->getPost('template')
         ];
+        if (empty($input['adminemail']) || empty($input['bounce_email'])) {
```
twilligls, line 372: Validation for form inputs is (most of the time :) ) declared inside of the rules function of the model.
In this case see: application/models/Survey.php 
public function rules()
There is already some validation for adminemail and bounce_email using the LSYii_FilterValidator.

As I understand it's still ok to have those fields empty, but when they are not they should be checked for valid email.
You could try with 
array('adminemail', 'LSYii_FilterValidator', 'filter' => 'email', 'skipOnEmpty' => true)

application/views/admin/survey/subview/accordion/_generaloptions_panel.php
```diff
@@ -179,7 +179,7 @@ function checkSelect2Languages(mylangs) {
                 <div class="col-12 col-lg-8 content-right">
                     <label class=" form-label" for='adminemail'><?php eT("Administrator email address:"); ?></label>
                     <input class="form-control inherit-edit <?php echo($bShowInherited && $adminemail === 'inherit' ? 'd-none' : 'd-block'); ?>" type='<?php echo ($bShowInherited) ? 'text' : 'email'; ?>'
-                           size='50' id='adminemail' name='adminemail' value="<?php echo htmlspecialchars((string) $adminemail); ?>" data-inherit-value="inherit" data-saved-value="<?php echo htmlspecialchars((string) $adminemail); ?>"/>
+                           size='50' id='adminemail' name='adminemail' value="<?php echo htmlspecialchars((string) $adminemail); ?>" data-inherit-value="inherit" data-saved-value="<?php echo htmlspecialchars((string) $adminemail); ?>" required="required"/>
```
twilligls, line 182: I don't think we need required fields here. See above comment

application/views/admin/survey/subview/accordion/_generaloptions_panel.php
```diff
@@ -206,7 +206,7 @@ function checkSelect2Languages(mylangs) {
                     <label class=" form-label" for='bounce_email'><?php eT("Bounce email address:"); ?></label>
                     <input class="form-control inherit-edit <?php echo($bShowInherited && $bounce_email === 'inherit' ? 'd-none' : 'd-block'); ?>"
                            type='<?php echo ($bShowInherited) ? 'text' : 'email'; ?>' size='50' id='bounce_email' name='bounce_email' value="<?php echo htmlspecialchars((string) $bounce_email); ?>"
-                           data-inherit-value="inherit" data-saved-value="<?php echo htmlspecialchars((string) $bounce_email); ?>"/>
+                           data-inherit-value="inherit" data-saved-value="<?php echo htmlspecialchars((string) $bounce_email); ?>" required="required"/>
```
twilligls, line 209: I don't think we need required fields here. See above comment

twilligls (changes requested):

application/models/Survey.php
```diff
@@ -496,8 +496,8 @@ public function rules()
             array('expires', 'default', 'value' => null),
             array('admin', 'LSYii_Validators'),
             array('admin', 'length', 'min' => 1, 'max' => 50),
-            array('adminemail', 'LSYii_FilterValidator', 'filter' => 'trim', 'skipOnEmpty' => true),
-            array('bounce_email', 'LSYii_FilterValidator', 'filter' => 'trim', 'skipOnEmpty' => true),
+            array('adminemail', 'LSYii_FilterValidator', 'filter' => 'trim', 'skipOnEmpty' => true, 'allowOnlyEmail' => true),
+            array('bounce_email', 'LSYii_FilterValidator', 'filter' => 'trim', 'skipOnEmpty' => true), 'allowOnlyEmail' => true,
```
twilligls, line 500: here is a syntax problem with ")"

application/core/LSYii_FilterValidator.php
```diff
@@ -36,6 +37,22 @@ protected function validateAttribute($object, $attribute)
...
+                $emailRegex = '/^([a-zA-Z0-9.!#$%&’*+\/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+))(,([a-zA-Z0-9.!#$%&’*+\/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)))*$/';
+                $isValid = preg_match($emailRegex, (string) $object->$attribute);
+                if (!$isValid)
+                {
+                    Yii::app()->setFlashMessage(gT("Invalid email address."), 'error');
+                    Yii::app()->end();
+                }
+            }
+        }
+
```
twilligls, line 55: You don't need to create that functionality, it is already there in yii.
Just use 'filter' => 'email' in rules()
Then with the help of yii magic the CEmailValidator class is used for the validation.  Looks like it even uses the same regex expression.

twilligls: 'filter' => 'email' is obviously not correct :)

twilligls (approved):

PR #3334, opened 2023-08-04
Fixed issue #19009: Cannot unset expiry or start date of survey

adamzammit:
Fixed issue #19009: Cannot unset expiry or start date of survey

### DIFF
--- a/application/models/services/SurveyUpdater/GeneralSettings.php
+++ b/application/models/services/SurveyUpdater/GeneralSettings.php
@@ -313,7 +313,7 @@ class GeneralSettings
             case static::FIELD_TYPE_DATETIME:
                 $value = !empty($value)
                     ? $this->formatDateTimeInput($value)
-                    : $default;
+                    : null;
                 break;
             case static::FIELD_TYPE_YN:
                 if (!in_array('' . $value, ['Y', 'N', 'I'])) {


application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -313,7 +313,7 @@ private function setField($field, $input, Survey $survey, $meta, $fieldOpts = nu
             case static::FIELD_TYPE_DATETIME:
                 $value = !empty($value)
                     ? $this->formatDateTimeInput($value)
-                    : $default;
+                    : null;
```
gabrieljenik, line 316: We are applying a general change here while the bug is for couple of fields.
Should we do so?

Have you tried reviewing `$fieldOpts['default']` ?
Maybe we just need to change the options passed to the method...?

Maybe the method doesn't allow a mechanism for "clearing" the values?

adamzammit: > We are applying a general change here while the bug is for couple of fields. Should we do so?
> 
> Have you tried reviewing `$fieldOpts['default']` ? Maybe we just need to change the options passed to the method...?
> 
> Maybe the method doesn't allow a mechanism for "clearing" the values?

The issue is here: https://github.com/LimeSurvey/LimeSurvey/blob/3eac536c0bc6f4ac55458d4be6b6af1ced787123/application/models/services/SurveyUpdater/GeneralSettings.php#L300 the check for empty against null means that the value is set to "initvalue". initvalue is the current value of the field so it is never overwritten by null.

adamzammit: Also it appears nothing else uses the field type datetime and it makes sense to write null when null is given.

kevin-foster-uk:
This is a bug but its not the correct fix. I will create another PR and provide you the link shortly.

kevin-foster-uk:
Here is the PR. 

https://github.com/LimeSurvey/LimeSurvey/pull/3355

Im going to ask someone else on the team to test it before a merge. In the mean time feel free to test it yourself.

PR #3335, opened 2023-08-04
Fixed issue #18439: Link set when create user are unusuable in text view

gabrieljenik:
Altbody was never set as phpMailer always set it based on the original body.
Hence, the custom "alt-body creation" was not applied. This last was fixed.

### DIFF
--- a/application/core/LimeMailer.php
+++ b/application/core/LimeMailer.php
@@ -585,11 +585,10 @@ class LimeMailer extends \PHPMailer\PHPMailer\PHPMailer
             if (strpos($this->Body, "<html>") === false) {
                 $this->Body = "<html>" . $this->Body . "</html>";
             }
-            $this->msgHTML($this->Body, App()->getConfig("publicdir")); // This allow embedded image if we remove the servername from image
-            if (empty($this->AltBody)) {
-                $html = new \Html2Text\Html2Text($this->Body);
-                $this->AltBody = $html->getText();
-            }
+            $this->msgHTML($this->Body, App()->getConfig("publicdir"), function($html) {
+                return (new \Html2Text\Html2Text($html))->getText();
+            }); // This allow embedded image if we remove the servername from image
+
         }
         return $this->Send();
     }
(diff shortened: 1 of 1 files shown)


application/core/LimeMailer.php
```diff
@@ -585,11 +585,10 @@ public function sendMessage()
...
             }
-            $this->msgHTML($this->Body, App()->getConfig("publicdir")); // This allow embedded image if we remove the servername from image
-            if (empty($this->AltBody)) {
-                $html = new \Html2Text\Html2Text($this->Body);
-                $this->AltBody = $html->getText();
-            }
+            $this->msgHTML($this->Body, App()->getConfig("publicdir"), function($html) {
+                return (new \Html2Text\Html2Text($html))->getText();
+            }); // This allow embedded image if we remove the servername from image
+            // TODO: Original AltBody is overwritten by msgHTML. Do we need to set it again if there was one?
```
Shnoulle, line 591: Before we leave https://github.com/LimeSurvey/LimeSurvey/blob/898dec3ad68a0186f815eed8c5f6dd2ca7f4e4eb/vendor/phpmailer/phpmailer/src/PHPMailer.php#L4387

And if it's empty we use `\Html2Text\Html2Text` 

Now we force `\Html2Text\Html2Text` ?

Maybe extend html2text function on LimeMailer directly is better ? and more understandable ?

Shnoulle (commented):
Extend directly https://github.com/LimeSurvey/LimeSurvey/blob/898dec3ad68a0186f815eed8c5f6dd2ca7f4e4eb/vendor/phpmailer/phpmailer/src/PHPMailer.php#L4387C21-L4387C30 ?

(good idea to check AltBody before updating : leave plugin to set AltBody)

application/core/LimeMailer.php
gabrieljenik: > Maybe extend html2text function on LimeMailer directly is better ? and more understandable ?

Not sure why it would be more understandable.
As it is now, we are using the intended way of overriding the standard converter and use a custom one.

> Now we force \Html2Text\Html2Text ?

Same way as we were trying to force it before.

Shnoulle: > Not sure why it would be more understandable. As it is now, we are using the intended way of overriding the standard converter and use a custom one.

Because html2text phpMailer function wads never used. Since we already extend it : more clear to replace the function. Really more clear

Shnoulle (approved):

PR #3341, opened 2023-08-08
Qe 296 question services

kevin-foster-uk:
Refactor question administration controller functionality in service classes.


tiborpacalat:
As I already shown to Kevin, the text of the question is not saved. Neither on create, not on edit. I can see that it is in the post payload, but it is not saved to the database.

PR #3345, opened 2023-08-09
MW-446: flow account and instance existing

lajosarpad:
To test, one needs to make sure that the

tmp/templates

folder exists and there are lss files inside. (you can export a survey into lss and copy it to the desired location if needed)

Then, encode the filename (without the path) into base64 and, while being logged out, load the

http://ls-ce/index.php?templatetoken=< encoded filename >&r=admin/authentication/sa/login

path. For example:

http://ls-ce/index.php?templatetoken=bGltZXN1cnZleV9zdXJ2ZXlfOTc5NTczLmxzcw==&r=admin/authentication/sa/login

where bGltZXN1cnZleV9zdXJ2ZXlfOTc5NTczLmxzcw== is the token of limesurvey_survey_979573.lss.

The page will redirect you to the login page where you could log in.

After logging in, we expect to be redirected to the survey page and a popup being shown asking whether you want to install the template.

If you leave this unanswered and reload the page, the confirm dialog should still be displayed.

If you click:

- outside of the popup content
- the X of the dialog
- No

then the popup should be dismissed and if you reload the page, the popup should no longer be displayed.

Otherwise, if you click on Yes, then a peview should be shown (in the context of this PR this is an empty popup, it's a known issue) and if you close it, then the grid should be refreshed with the imported template.


application/controllers/AdminController.php
```diff
@@ -120,6 +120,30 @@ public function error($message, $sURL = array())
...
     }
 
+    public function actionRemoveTemplateToken()
+    {
+        if (isset(Yii::app()->session['templatetoken'])) {
+            unset(Yii::app()->session['templatetoken']);
+        }
+    }
+
+    public function actionInstallTemplateByToken()
```
olleharstedt, line 130: Why is this needed if we already installed template via CLI?

application/controllers/admin/Authentication.php
```diff
@@ -40,6 +40,17 @@ public function index()
             Yii::app()->session['adminlang'] = Yii::app()->request->getParam('loginlang', 'default');
             Yii::app()->setLanguage(Yii::app()->session["adminlang"]);
         }
+        /* Checking whether we have any templates to import */
+        if (($templatetoken = Yii::app()->request->getParam('templatetoken', 'default')) !== 'default') {
```
olleharstedt, line 44: Again, should already have been installed via CLI, or?

application/controllers/AdminController.php
```diff
@@ -120,6 +120,30 @@ public function error($message, $sURL = array())
...
+    {
+        if (isset(Yii::app()->session['templatetoken'])) {
+            Yii::import('application.helpers.admin.token_helper', true);
+            $filename = decodeFilename(Yii::app()->session['templatetoken']);
+            $this->actionRemoveTemplateToken();
+            if (!preg_match('/^[a-zA-Z0-9_\.]*$/', $filename)) {
+                echo "badly formatted file";
+            } else {
+                exec("php application/commands/console.php importsurvey import-file {$filename}", $output);
```
olleharstedt, line 139: I'm not sure this will be needed, since we'll use our message system and cronjobs to distribute and install template files. Correct me if I'm wrong, maybe there's a use-case I'm not aware of?

application/controllers/AdminController.php
lajosarpad: This action is triggering the CLI command to install the template by token

application/commands/ImportSurveyCommand.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+Yii::import('application.helpers.replacements_helper', true);
+Yii::import('application.helpers.expressions.em_manager_helper', true);
+Yii::import('application.helpers.common_helper', true);
+Yii::import('application.helpers.admin.import_helper', true);
+
+class ImportSurveyCommand extends CConsoleCommand
```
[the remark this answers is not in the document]
lajosarpad: Applied the change

application/controllers/admin/Authentication.php
lajosarpad: Here we check for the existence of a token parameter in the URL and store it into the session, so later on the Survey site will know that a confirmation will needed to be displayed for the template.

image

Reference: https://limesurvey.atlassian.net/browse/MW-446

application/controllers/AdminController.php
olleharstedt: Also need a permission check here I think? Either super admin or extension access.

olleharstedt: If we wanna do template install via URL instead of message queue, I'd rather go via RPC API. Me and Carsten discussed it briefly, but decided to use the message queue system instead.

lajosarpad: Applied some changes to check for permissions

application/controllers/AdminController.php
olleharstedt: It's gonna be hard for me to approve a PR in which the functionality is spread out like this in different controllers, so starting work with the message queue system might be closer at hand for now.

lajosarpad: I have no knowledge about the message queue system yet, so I will need to get through a learning curve in order to become able to apply this idea

PR #3356, opened 2023-08-15
Fixed issue #18648: Unable to upload a file with theme animations enabled

gabrieljenik:


Shnoulle (approved):
Totally unsure here …

themes/survey/fruity/views/subviews/header/custom_header.twig
```diff
@@ -139,7 +139,7 @@
         .question-container {
             animation-name: {{ aSurveyInfo.options.questionanimation  }};
             animation-duration: {{ questionanimationduration }}ms;
-            animation-fill-mode: both;
+            animation-fill-mode: none;
```
Shnoulle, line 142: Since non is default : maybe remove ?

gabrieljenik: I would just leave it like that as to make sure this will not be inherited and also to reflect we want this to be default explicitely

Shnoulle: Maybe add the mantis id in twig comment ?

PR #3357, opened 2023-08-15
Fixed issue #18976 - default answers should persist with subquestions

kevin-foster-uk:
Fixed issue #18976: default answers should persist with subquestions
Dev: Minimal update as affected code is in the process of being refactored


olleharstedt:
Looks good, only question I have is if re-order still works? IIRC, that's why we delete all subquestions before save.

kevin-foster-uk:
> Looks good, only question I have is if re-order still works? IIRC, that's why we delete all subquestions before save.

The sort order is set for all subquestions (existing or new) based on the order of the input data array. So yes, re-order still works.

PR #3358, opened 2023-08-15
Fixed issue #19025: When there are question themes installed, can't upload survey themes

gabrieljenik:


olleharstedt (approved):

application/views/themeOptions/availablethemelist.php
```diff
@@ -12,29 +12,31 @@
...
+                            <td class="col-lg-2"><?php echo $oTheme->name; ?></td>
+                            <td class="col-lg-3"><?php echo $oTheme->description; ?></td>
+                            <td class="col-lg-2"><?php eT('XML themes'); ?></td>
+                            <td class="col-lg-2"><?php echo $oTheme->extends; ?></td>
+                            <td class="col-lg-1"><?php echo $oTheme->manifestbuttons; ?></td>
+                        </tr>
+                    <?php endforeach; ?>
+                    </tbody>
+                </table>
+            </div>
```
Shnoulle, line 38: This double div end can not move to the wrapper ?
At the same place where it start ? (in same file)

Shnoulle (commented):
I really think it's a bad solition to start a element and end it in another file.

Here seems we have 4 div started elsewhere ??? %Maybe time to move it in the file it start.

application/views/themeOptions/availablethemelist.php
gabrieljenik: These are clsing the ones opened in line 9
https://github.com/LimeSurvey/LimeSurvey/pull/3358/files#diff-b21a655d0b694f558c8e64c048f73126507817841d37584f482ea9a80271c38fR9

Right?

Shnoulle: Oh right … sorry …

Shnoulle (approved):

PR #3359, opened 2023-08-15
Fixed issue #19019: Stored XSS in description of theme

gabrieljenik:


application/models/TemplateConfiguration.php
```diff
@@ -545,6 +545,7 @@ public function getDescription()
           // Note: if no twig statement in the description, twig will just render it as usual
         try {
             $sDescription = App()->twigRenderer->convertTwigToHtml($this->template->description);
+            $sDescription = (new LSYii_Validators())->xssFilter($sDescription);
```
Shnoulle, line 548: My opinion : the description must be filtered when show.

For super admin too.

Shnoulle (commented):
I think kit must be filtered for all : here it was not filtered for superadmin.

It's not a validator, it's a display system.

application/models/TemplateConfiguration.php
gabrieljenik: > My opinion : the description must be filtered when show.

As the filter is on the getDescription() method, somehow the description is also filtered when shown ...
image

> For super admin too.

It is applued to all users.

Shnoulle (changes requested):
Use a dedicated function seems really better.

Validator is here to validate

application/models/TemplateConfiguration.php
Shnoulle: You're right, sorry. But : totally unclear usage of a validator …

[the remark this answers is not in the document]
Shnoulle: > Well, but aren't we using that sometimes for filtering as well? Ex: https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/LSActiveRecord.php#L235

It's in model in save function … 

> 
> https://github.com/LimeSurvey/LimeSurvey/blob/f083a6e7b73e36b2cb9bcf2f413b771f3fbfb73d/application/helpers/viewHelper.php#L265
> is different than the xssFilter function from the LSYii_Validator.

xssFilter need to allow expression manager AND to filter xss inside Expression. 
It' totally different : here you just show the HTML to administrator


> So maybe we have 2 different and separate problems:
> 
>     * The function may not be on the right place, whereas it should be maybe on another helper.
>     * Purifier xss is different than LSYii_Validator and may not be very clear when to use one or the other,

xssFilter must be used only when save : it allow save Expression with HTML inside but disallow script or bad HTML.
It's NOT related to view, when you need to show something with script : use something done for view.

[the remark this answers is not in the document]
Shnoulle: > is different than the xssFilter function from the LSYii_Validator.

Yes : and it's the reason why you need to use it here …

Shnoulle (approved):

PR #3361, opened 2023-08-16
Fixed bug #18799 Admin user with full CPDB permissions can no access attribute section.

gabrieljenik:


Shnoulle (approved):
Why blacklist only for admin user too ?

https://github.com/LimeSurvey/LimeSurvey/blob/775d0166f09850b2c5821518727e41513e7fc163/application/views/admin/participants/partial/topbarBtns/leftSideButtons.php#L63

application/controllers/admin/ParticipantsAction.php
```diff
@@ -1361,6 +1361,8 @@ public function changeblackliststatus()
      */
     public function attributeControl()
     {
+        $this->checkPermission('read');
```
Shnoulle, line 1364: 👍 👍 👍 👍

PR #2754, opened 2022-12-01
Bug/18530 surveys in group permission more

Shnoulle:
More https://github.com/LimeSurvey/LimeSurvey/pull/2752

PR #3368, opened 2023-08-18

gabrieljenik:
- Now: Users will only be able to set as owners to users which are under their vsibility level

### DIFF
--- a/application/controllers/admin/Database.php
+++ b/application/controllers/admin/Database.php
@@ -896,7 +896,13 @@ class Database extends SurveyCommonAction
             $oSurvey->owner_id == Yii::app()->session['loginID']
             || Permission::model()->hasGlobalPermission('superadmin', 'read')
         ) {
-            $oSurvey->owner_id = $request->getPost('owner_id');
+            $ownerId = $request->getPost('owner_id');
+            // Before setting the owner, check if the user exists and can be seen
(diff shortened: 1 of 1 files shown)


Shnoulle (commented):
OK

We really need a User model function for userlist …

gabrieljenik:
> We really need a User model function for userlist …

Will do some kind of refactoring in master

Shnoulle:
Code review done :)

Shnoulle:
> > We really need a User model function for userlist …
> 
> Will do some kind of refactoring in master. (not allowed in v5) Let's just pass this as it is now. What do you think?

https://github.com/LimeSurvey/LimeSurvey/pull/2754 for example … new scope 

https://github.com/LimeSurvey/LimeSurvey/pull/2754/files#diff-1071addb9a37bbecc23fa2136980caf42b5693fc2658f42d58d9136d3c1f1c43R60

Shnoulle:
Arg ¿ https://github.com/LimeSurvey/LimeSurvey/blob/5c4fb4e0137d8e564c14d30c3c9563ffd7ecaacd/application/helpers/remotecontrol/remotecontrol_handle.php#L409C31-L409C31

remote control …

tiborpacalat:
@Shnoulle is something wrong with this PR that was already merged?

PR #3370, opened 2023-08-18
Fixed issue #19018: Postgres: trying to update from 5.6.31 to master caused an error

gabrieljenik:


Shnoulle:
No way to check if an index exist ?

Must check if dropIndex can be used without valid index.

Shnoulle (approved):
I like to have a checkIndex solution , but not implemented in Yii1

PR #3371, opened 2023-08-18
Fixed issue 18821: Public url is not used for SURVEYURL

gabrieljenik:
- Test for cretaePublicUrl and getPublicBaseUrl
- Porting fix for LSYiiValidatorsTest::testXssFilterAttributeForSuperAdmin()
- Porting fix about urls

### DIFF
--- a/tests/unit/LSYiiApplicationTest.php
+++ b/tests/unit/LSYiiApplicationTest.php
@@ -1,0 +1,178 @@
+<?php
+
+namespace ls\tests;
+
+use Yii;
+
+/**
+ * Test LSYii_Application class.
+ */
+class LSYiiApplicationTest extends TestBaseClass
+{
+    /**
+     * Get public base url, previously set in publicurl config attribute.
+     */
+    public function testGetpublicBaseUrlFromConfig()
+    {
+        $tmpPublicUrl = Yii::app()->getConfig('publicurl');
+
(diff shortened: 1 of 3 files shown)


tests/unit/LSYiiApplicationTest.php
```diff
@@ -0,0 +1,178 @@
...
+     * Get public base url, previously set in baseUrl request attribute.
+     * A public url is also set in the baseUrl config attribute,
+     * but the one in the request attribute should be returned.
+     */
+    public function testGetpublicBaseUrlFromConfigNoHostInConfigPublicUrl()
+    {
+        $tmpConfigPublicUrl = Yii::app()->getConfig('publicurl');
+        $tmpRequestPublicUrl = Yii::app()->getRequest()->getBaseUrl();
+
+        Yii::app()->setConfig('publicurl', 'http://');
```
Shnoulle, line 108: My opinion : test must break here : publicurl must be always used at last option, even if it's a broken url

gabrieljenik:
This test is already merged on master.
v5 will be EOL soon, so, I wouldn't folllow up on this.

Shnoulle:
> This test is already merged on master. 

OK, i think its a new test here :)

Shnoulle (approved):

PR #3372, opened 2023-08-18
Fixed bug #18627 Missing SQL query errors in admin mail (SQL CODE THAT FAILED)

gabrieljenik:
- Moving out the try/catch 
  + from inside the mdoel method `insertRecords` 
  + to outside of the model, usually controller.
- Adding test
- Adding more info to the submitfailed

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -5062,7 +5062,9 @@ class LimeExpressionManager
             } catch (Exception $e) {
                 $srid = null;
                 $message .= $this->gT("Unable to insert record into survey table"); // TODO - add SQL error?
-                submitfailed($this->gT("Unable to insert record into survey table"), $e->getMessage());
+                $query = $e->getMessage();
+                $trace = $e->getTraceAsString();
+                submitfailed($this->gT("Unable to insert record into survey table"), $query . "\n\n" . $trace);
             }
 
             //Insert Row for Timings, if needed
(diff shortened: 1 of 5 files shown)


application/helpers/expressions/em_manager_helper.php
```diff
@@ -5055,15 +5055,21 @@ private function _UpdateValuesInDatabase($finished = false)
...
                 $srid = $iNewID;
                 $_SESSION[$this->sessid]['srid'] = $iNewID;
-            } else {
+            } catch (Exception $e) {
                 $srid = null;
                 $message .= $this->gT("Unable to insert record into survey table"); // TODO - add SQL error?
-                submitfailed($this->gT("Unable to insert record into survey table"));
+                $query = $e->getMessage();
+                $trace = $e->getTraceAsString();
+                submitfailed($this->gT("Unable to insert record into survey table"), $query . "\n\n" . $trace);
```
olleharstedt, line 5070: I'm having a silent fail here. Isn't `submitfailed` suppose to end survey taking?

olleharstedt: > La propietat "SurveyDynamic"."datestamp" no es troba definida.

olleharstedt: In other places in EM, it's used as

```php
$message = submitfailed($this->gT('The data could not be saved because the response does not exist in the database.'));
LimeExpressionManager::addFrontendFlashMessage('error', $message, $this->sid);
return;
```

olleharstedt: https://bugs.limesurvey.org/view.php?id=19335

gabrieljenik: > I'm having a silent fail here. Isn't `submitfailed` suppose to end survey taking?

Were you getting an email message at least?

olleharstedt: I didn't check the email. Customer didn't mention it. But yeah, they should have.

PR #3373, opened 2023-08-18
Fixed bug #18911 import of theme fails if it contains resource with uppercase extension

gabrieljenik:


application/helpers/admin/template_helper.php
```diff
@@ -134,7 +134,11 @@ function templateExtractFilter($p_event, &$p_header)
     $aAllowExtensions[] = 'twig';
     $info = pathinfo($p_header['filename']);
 
-    if ($p_header['folder'] || !isset($info['extension']) || in_array($info['extension'], $aAllowExtensions)) {
+    if (isset($info['extension'])) {
+        $p_header['filename'] = preg_replace('/\.\w+$/', '.' . strtolower($info['extension']), $p_header['filename']);
```
Shnoulle, line 138: You rename it here ? 

I don't think it's a good idea : if user add  in is twig file

Shnoulle (commented):
Dpon't rename it (if id odn't make error)

application/helpers/admin/template_helper.php
gabrieljenik: > I don't think it's a good idea : if user add  in is twig file

He should edit it in that case.
Lower case extensions will minimize errors I believe.

@olleharstedt thoughts?

Shnoulle: >  Lower case extensions will minimize errors I believe.

There are no reason to didn't allow picture.PNG

gabrieljenik: Done

olleharstedt:
Neat!

PR #3374, opened 2023-08-18
Fixed issue #18574: Fruity: If a variation is selected at survey level, custom.css is not loaded as expected

gabrieljenik:
When applying a variation to a theme which inherits all css files, this happened:
- When adding the variation, overwrote the "inherit" and was only leaving the variation css.
- Then when rendering, nothing was inherited and only the variation css was loaded.

Solution:
- When adding the varaition, and the css files is "inherit", it replaces it by the "contents of the parent" (what should be inherited).
- Then the variation is added.

### DIFF
--- a/application/views/themeOptions/update.php
+++ b/application/views/themeOptions/update.php
@@ -194,7 +194,7 @@ echo viewHelper::getViewTestTag('surveyTemplateOptionsUpdate');
                                 <div class="row">
                                     <div class="form-group">
                                         <?php echo $form->labelEx($model, 'files_css'); ?>
-                                        <?php echo $form->textArea($model, 'files_css', ['rows' => 6, 'cols' => 50]); ?>
+                                        <?php echo $form->textArea($model, 'files_css', ['rows' => 6, 'cols' => 50, 'data-inherit-source' => 'optionCssFiles']); ?>
                                         <?php echo $form->error($model, 'files_css'); ?>
                                     </div>
                                 </div>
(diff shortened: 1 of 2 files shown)


ptelu (changes requested):
when selecting inherit for the variations after previously selecting a different variation it does not load again the inherited value, same happens when selecting "inherit everything" through the switch.

Shnoulle:
Quick looking at code ,

If you have 

1. Group1 : Fruity with variation Orange
2. Survey1 in group1 : Fruit set as inherit 

survey1 cssoptions set as optionCssFiles ? No ?

This variation system didn't really work good … (i remove usage in skel)

gabrieljenik:
> when selecting inherit for the variations after previously selecting a different variation it does not load again the inherited value, same happens when selecting "inherit everything" through the switch.

Couldn't reproduce it. 
Peek_2023-11-08_16-08

Can you provide more detailed reproduction steps?

Thanks

gabrieljenik:
> Quick looking at code ,
> 
> If you have
> 
> 1. Group1 : Fruity with variation Orange
> 2. Survey1 in group1 : Fruit set as inherit
> 
> survey1 cssoptions set as optionCssFiles ? No ?
> 
> This variation system didn't really work good … (i remove usage in skel)

Sorry, don't follow.
Can you provide more detailed explanation of the problem you see?

Shnoulle (approved):

tiborpacalat:
QA test failed, 5x is not longer supported.

Shnoulle:
PR from 2023 … 

I'm sure the issue is still there, i report for master next week

tiborpacalat:
Denis, there is issue on master, Patrick is working on it.

PR #3376, opened 2023-08-18
CR-1117: We should restrict naming of themes because it causes errors in some cases

mohabmes:


application/models/Template.php
```diff
@@ -48,6 +48,8 @@ class Template extends LSActiveRecord
     /** @var Template - The instance of template object */
     private static $instance;
 
+    public static $sTemplateNameRegex = '/^[a-zA-Z0-9_ -]+$/';
```
olleharstedt, line 51: Not sure this will work in an international context. What if your keyboard is Arabic, Greek or Russian?

application/models/Template.php
```diff
@@ -77,6 +80,33 @@ public function rules()
...
+    /**
+     * Validate the template name.
+     *
+     * @param string $templateName The name of the template
+     */
+    public static function validateTemplateName($templateName)
+    {
+        if (strpbrk((string) $templateName, Template::$sTemplateNameIllegalChars)) {
+            Yii::app()->setFlashMessage(sprintf(gT("The name contains special characters.")), 'error');
+            Yii::app()->getController()->redirect(array('themeOptions/index'));
```
Shnoulle, line 101: It's not the way to do validation …

Shnoulle: Here : best seems to filter directly …

Shnoulle:
It's really a bad solution to redirect in model … 
Controller must redirect or show the error return by model when save

We MUST use `$this->addError($object, $attribute, gT('The string need to be returned.'));`

PR #3382, opened 2023-08-22
Fixed bug #18342: Default ranking help text is missing helpm icon

gabrieljenik:
Icon added.

### DIFF
--- a/assets/packages/questions/ranking/scripts/ranking.js
+++ b/assets/packages/questions/ranking/scripts/ranking.js
@@ -19,7 +19,7 @@ var RankingQuestion = function (options) {
         rankingID = "javatbd" + rankingName;
 
     //define HTML snippets
-    var screenReader = "<div class='sr-only'>" + $('#question' + questionId + ' .em_default').html() + "</div><div aria-hidden='true'>" + LSvar.lang.rankhelp + "</div>"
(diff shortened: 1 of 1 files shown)


Shnoulle:
The class is sr-only : no need icon ?

assets/packages/questions/ranking/scripts/ranking.js
```diff
@@ -19,7 +19,7 @@ var RankingQuestion = function (options) {
         rankingID = "javatbd" + rankingName;
 
     //define HTML snippets
-    var screenReader = "<div class='sr-only'>" + $('#question' + questionId + ' .em_default').html() + "</div><div aria-hidden='true'>" + LSvar.lang.rankhelp + "</div>"
+    var screenReader = "<div class='sr-only'>" + $('#question' + questionId + ' .em_default').html() + "</div><div aria-hidden='true'><span class='fa fa-exclamation-circle' aria-hidden='true'></span> " + LSvar.lang.rankhelp + "</div>"
```
Shnoulle, line 22: 1. It seems to be the screenreader help, not the help for other user.
2. Maybe best is to copy current help and replace the text : then it came from twig file (template can replace this twig file)

gabrieljenik: > 1. It seems to be the screenreader help, not the help for other user.

Well. Not sure about that. The icon appears after this.
image

> Maybe best is to copy current help and replace the text : then it came from twig file (template can replace this twig file)

The ticket is just about showing the icon. This is v5. Not sure it is worth here.

gabrieljenik:
> 

It is like that on the twig files.

Shnoulle:
The icon is here without JS

Shnoulle (approved):
Like to have it by twig, but don't find a good way.

PR #3392, opened 2023-08-28
Fixed issue #18610: Email attachments expose absolute path when exporting LSS

gabrieljenik:
Turning paths from absulote to relative before exporting.
Turning paths from relative to absulote before importing.
Exporting attachment info in json format

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -1366,6 +1366,9 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
         }
     }
 
+    // Single flag to indicate if the attachements format is wrong, to avoid showing the warning multiple times
+    $wrongAttachmentsFormat = false;
+
     // Import survey languagesettings table ===================================================================================
     foreach ($xml->surveys_languagesettings->rows->row as $row) {
         $insertdata = array();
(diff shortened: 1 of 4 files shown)


application/helpers/admin/import_helper.php
```diff
@@ -1410,6 +1413,28 @@ function XMLImportSurvey($sFullFilePath, $sXMLdata = null, $sNewSurveyName = nul
...
+                        if (!isAbsolutePath($attachment['url'])) {
+                            $attachment['url'] = $uploadDir . DIRECTORY_SEPARATOR . $attachment['url'];
+                        }
+                    }
+                }
+            } elseif (is_null($attachments)) {
+                // JSON decode failed. Most probably the attachments were in the PHP serialization format.
+                $wrongAttachmentsFormat = true;
+            }
+            $insertdata['attachments'] = serialize($attachments);
```
kevin-foster-uk, line 1435: This line should move into the if statement above. We don't want to serialise if we did not manage to parse anything from the file.

gabrieljenik: You meant to avoud serializing an empty string?
I believe the rest of the code expects serlized data there. Even if empty. This is working on master.

kevin-foster-uk: If we dont parse anything it should be an empty string. Right now if we dont parse anything we might store invalid data.

gabrieljenik: > If we dont parse anything it should be an empty string

Or a seralized empty string?

kevin-foster-uk: Serialised empty string is fine. Right now it's serialising anything coming from the input which means I could enter serialised PHP in the .lss file and it would serialise that. Maybe that wont break anything but do we want that?

kevin-foster-uk: Sure its a step in the right direction.

kevin-foster-uk:
Issues that still need to be dealt with, as I am aware:

- If import does not not contain valid JSON we should store the input data at all
- New file paths should have new survey id 
- Add these changes to LS6 / master?

gabrieljenik:
Thanks!

>> If import does not contain valid JSON we should store the input data at all

Replied on the thread.

>> New file paths should have new survey id

Tackled on 18557 / 18602

>> Add these changes to LS6 / master?

Done already

PR #3395, opened 2023-08-28
Fixed bug #19002: Unauthorized access to Survey menu entries 

gabrieljenik:


application/controllers/admin/SurveymenuController.php
```diff
@@ -367,6 +381,7 @@ protected function performAjaxValidation($model)
      **/
     public function index()
     {
+        // Permission check is done in the constructor.
```
Shnoulle, line 384: Why in the constructoi ?

It's the only one controller where we check permission in constrictor, no ?

I think it's better to do on function OR add it too in other controller constructor (minimal access : read)

Shnoulle (commented):
Why in constructor ?

application/controllers/admin/SurveymenuController.php
gabrieljenik: > It's the only one controller where we check permission in constrictor, no ?

Well, not the only place.
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/globalsettings.php#L32
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/admin/SurveymenuEntryController.php#L13

> I think it's better to do on function OR add it too in other controller constructor (minimal access : read)

I added the comment on the functions as to try to clarify

Shnoulle: > Well, not the only place.

OK,
then : let's go :)

Shnoulle (approved):

PR #3400, opened 2023-08-29
Fixed issue #18978: User is able to change survey group's code that is by default unchangeable

gabrieljenik:


Shnoulle (commented):
Must be in model too ?
No ?

Shnoulle (approved):

PR #3408, opened 2023-08-30
Dev: Minor textual enhancement

TillerBelotti:
Dev: Minor textual enhancement


Shnoulle:
It's an header, i think is better with only Created

TillerBelotti:
> It's an header, i think is better with only Created

Can be, but in some other languages it is a little weird. Translating correctly with "Created" only can be tricky.

Shnoulle:
What about _Date created_

Ping @c-schmitz

tiborpacalat:
I think this is a bad idea, since this change will invalidate all other translations of the original string.

c-schmitz:
Patch is obsolete and texts have been reworked in v7, anyway.

PR #3411, opened 2023-08-30
Fixed issue #18593: Float attributes are not supported anymore

gabrieljenik:


application/extensions/AdvancedSettingWidget/views/float.php
```diff
@@ -0,0 +1,23 @@
...
+    <input
+        type="number" 
+        class="form-control" 
+        name="<?= $inputBaseName ?>"
+        id="<?= CHtml::getIdByName($inputBaseName); ?>"
+        <?= ($this->setting['help']) ? 'aria-describedby="help-' . CHtml::getIdByName($inputBaseName) . '"' : "" ?>
+        value="<?= CHtml::encode($this->setting['value']); ?>"
+        min="<?= $this->setting['min'] ?? ''?>"
+        max="<?= $this->setting['max'] ?? ''?>"
+        pattern="^[-+]?[0-9]*\.[0-9]+$"
```
Shnoulle, line 16: Why ? Type number do the restriction.

Pattern only need for integer or specific number

Shnoulle (approved):
Why a pattern ?
Uneeded

application/extensions/AdvancedSettingWidget/views/float.php
```diff
@@ -0,0 +1,23 @@
...
+    <?php endif; ?>
+    <input
+        type="number" 
+        class="form-control" 
+        name="<?= $inputBaseName ?>"
+        id="<?= CHtml::getIdByName($inputBaseName); ?>"
+        <?= ($this->setting['help']) ? 'aria-describedby="help-' . CHtml::getIdByName($inputBaseName) . '"' : "" ?>
+        value="<?= CHtml::encode($this->setting['value']); ?>"
+        min="<?= $this->setting['min'] ?? ''?>"
+        max="<?= $this->setting['max'] ?? ''?>"
```
Shnoulle, line 15: If max is not set : must be blank, here it's max=''

Shnoulle (changes requested):
Remove invalid potential invalid html if no min or no max set.

> The [min](https://html.spec.whatwg.org/multipage/input.html#attr-input-min) attribute, if specified, must have a value that is a [valid floating-point number](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-floating-point-number). The [max](https://html.spec.whatwg.org/multipage/input.html#attr-input-max) attribute, if specified, must have a value that is a [valid floating-point number](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-floating-point-number).

Shnoulle (approved):

PR #3418, opened 2023-08-31
Fixed issue #18602: When surveys are copied without copying attachments, attachment info is copied but not updated

gabrieljenik:


kevin-foster-uk:
Its difficult to test this branch because email attachment import functionality is saving the data as serialised PHP but then the .lss export functionality is saving the data as JSON.

Do you need to merge the latest 5.x into this branch to get the recent changes from other PRs?

gabrieljenik:
> Do you need to merge the latest 5.x into this branch to get the recent changes from other PRs?

Strange.
Isn't it already?

gabrieljenik:
> Its difficult to test this branch because email attachment import functionality is saving the data as serialised PHP but then the .lss export functionality is saving the data as JSON.

Import saves inside LS as serialized, as it has always been. We changed how info is being stored, just the import and export format from serialized to json.

kevin-foster-uk:
OK thanks for clarifying. 

One thing I did notice is that if you attempt to import a .lss file which contains JSON attachments to a LimeSurvey version which expects serialised data, you get a PHP error.

PR #3423, opened 2023-09-01
Fixed issue #19038: Privilege escalation bug to creation survey-group with others group as parent

gabrieljenik:


application/models/services/SurveysGroupCreator.php
```diff
@@ -51,7 +55,24 @@ public function __construct(
...
     public function save()
     {
-        $this->surveysGroup->attributes = $this->request->getPost('SurveysGroups');
+        $attributes = $this->request->getPost('SurveysGroups');
+
+        // Check if the parent group is valid
+        if (!empty($attributes['parent_id'])) {
+            $parentId = $attributes['parent_id'] ;
+            /* Check permission */
+            $availableParents = $this->surveysGroup->getParentGroupOptions();
```
Shnoulle, line 64: The getParentGroupOptions have more code than https://github.com/LimeSurvey/LimeSurvey/blob/76bb242b4224d0ae58f11a50d6c1f6adac58d4fd/application/models/SurveysGroups.php#L347

(we still need a scope for permission … )

Shnoulle (commented):
Some comment

1. getSurveyGroupsList is already used for Survey 
2. Add it to rules (and same way than Surveys maybe ?)

Shnoulle:
_General comment for discussion_ : We don't have a permission about "Create child group in this survey group", maybe a feature here ?
Maybe it's more clear …
Here : allow to see => allow to use , maybe need «Permission to modify survey group security settings» ?

Capture d’écran du 2023-09-06 09-53-11

gabrieljenik:
Got lost in this. Can you re-review @Shnoulle  ? Thanks

Shnoulle (approved):

PR #3426, opened 2023-09-06
Fixed issue #19056: save button in create dummy participants page actually saves and closes

gabrieljenik:


application/controllers/admin/Tokens.php
```diff
@@ -1081,7 +1081,8 @@ public function addDummies(int $iSurveyId, string $subaction = '')
                 [
                     'showSaveAndCloseButton' => true,
```
Shnoulle, line 1082: No way to renane "Save and close" to "Create" ?

Shnoulle (approved):
If no way : need to create an UX issue. It muts be “Create participants“

application/controllers/admin/Tokens.php
gabrieljenik: Text is not customizable. But could it be... let's leave it for another ticket

Shnoulle: Yep

> If no way : need to create an UX issue. It muts be “Create participants“

PR #3433, opened 2023-09-11
New feature #18880: New imagefile option type for theme options

gabrieljenik:


ptelu (approved):
Looks fine, dont like that we are adding more inline code inside this already complicated and to long view file, but so be it.

PR #3434, opened 2023-09-11
Fixed issue #19015: User with superadmin role does not get superadmin privileges

Shnoulle:
Dev: check it inside static view


application/models/Permission.php
```diff
@@ -785,6 +792,23 @@ public static function getUserRole($iUserID)
...
     }
 
+        /**
+     * get permission by user roles
+     * @param integer[] $rolesIds array of roles id
+     * @param string $permission;
+     * @param string $crud
+     * @return boolean
+     */
+    public static function getPermissionByRoles($rolesIds, $permission, $crud = 'read')
```
Shnoulle, line 802: We can really use this function in https://github.com/LimeSurvey/LimeSurvey/blob/57a22a6b6b44d913159c683598bf9a76e2dad53e/application/models/Permission.php#L607-L614

adamzammit: I agree as functions so similar, but also would need to updated here: https://github.com/LimeSurvey/LimeSurvey/blob/0a9690936d48465ecc0eadf2b12933b81ae26cf3/application/models/Permissiontemplates.php#L370  and https://github.com/LimeSurvey/LimeSurvey/blob/0a9690936d48465ecc0eadf2b12933b81ae26cf3/application/views/userRole/partials/_permissions.php#L92

adamzammit: This patch resolves the issue. I think could be applied as is, and then update and remove references to reoleHasPermission later. Or could be done in one patch for completeness?

Shnoulle: > I agree as functions so similar, but also would need to updated here:

- roleHasPermission function : one role give has permission
- getPermissionByRoles function : list of roles

You think allow  `$iRoleId` to be an array in https://github.com/LimeSurvey/LimeSurvey/blob/0a9690936d48465ecc0eadf2b12933b81ae26cf3/application/models/Permission.php#L720 ?

Alternative solution : 
````
    /**
     * get permission by user roles
     * @param integer $userId array of roles id
     * @param string $permission;
     * @param string $crud
     * @return boolean|null : boolean if have permission set, null if not
     */
    public static function getPermissionByRoles($userID, $permission, $crud = 'read')
    {
        $aRolesList = CHtml::listData(self::getUserRole($userID), 'ptid', 'ptid');
        if (empty($aRolesList)) {
            return null;
        }
        $criteria = new CDbCriteria();
        $criteria->compare("entity", "role");
        $criteria->compare("permission", $permission);
        $criteria->addInCondition('entity_id', $rolesIds);
        $criteria->compare($crud . '_p', 1);
        return boolval(self::model()->count($criteria));
    }
````
 For using it 
````
    $byRole = self::getPermissionByRoles($userID, 'superadmin')
    if (!is_null($byRole)) {
        $permission = $byRole;
    }
````

But here : usage is really more complex … if user don't have role : permission came from Permission, if have role : came from role.

2nd alt solution

- After plugin event : check if user have role
- If yes : create a new function checking only role for ALL system.

Comment (Shnoulle) [reply application/models/Permission.php:802]: > This patch resolves the issue. I think could be applied as is, and then update and remove references to reoleHasPermission later. Or could be done in one patch for completeness?

If possible : since i add function : it must be OK to be used elsewhere by other dev.

File: application/models/Permission.php [hunk 2]
```diff
@@ -785,6 +792,23 @@ public static function getUserRole($iUserID)
         return UserInPermissionrole::model()->getRoleForUser($iUserID);
     }
 
+        /**
```
gabrieljenik, line 795: Typo.

gabrieljenik:
Code looks good!
I there any unit test that should be updated or added ?
As this being security, maybe it is worth doing it? Thoughts?

twilligls:
Would be nice if this would be done soon. Just had a request from a user regarding this issue.

Shnoulle:
Need an unit test then ?

twilligls:
> Need an unit test then ?

This would be very nice :)

Shnoulle:
cannot test locally 
```

$ composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Package fzaninotto/faker is abandoned, you should avoid using it. No replacement was suggested.
Package phpunit/php-token-stream is abandoned, you should avoid using it. No replacement was suggested.
Package webmozart/path-util is abandoned, you should avoid using it. Use symfony/filesystem instead.
Generating autoload files

In ClassMapGenerator.php line 129:
                                                                                                                                                                
  Could not scan for classes inside "/media/shnoulle/data/webdev/lstest/vendor/hamcrest/hamcrest-php/hamcrest" which does not appear to be a file nor a folder  

```

Hope it work

Shnoulle:
> Hope it work

It must break
```
        $this->assertTrue(\Permission::model()->hasGlobalPermission('superadmin', 'read', $userId));
        $this->assertFalse(\Permission::model()->hasGlobalPermission('superadmin', 'read', $userId));
```

But to be sure it break is a good things too !

Shnoulle:
> Need an unit test then ?

Done

tests/unit/models/UserInPermissionRoleTest.php
```diff
@@ -2,8 +2,82 @@
...
+        $permission->save();
+        $userInPermissionrole = new \UserInPermissionrole();
+        $userInPermissionrole->ptid = $newRoleId;
+        $userInPermissionrole->uid = $userIdSuperadminCreate;
+        $userInPermissionrole->save();
+        // User have create superadmin
+        $this->assertTrue(\Permission::model()->hasGlobalPermission('auth_db', 'read', $userIdSuperadminView));
+        $this->assertTrue(\Permission::model()->hasGlobalPermission('superadmin', 'read', $userIdSuperadminView));
+        $this->assertTrue(\Permission::model()->hasGlobalPermission('superadmin', 'create', $userIdSuperadminCreate));
+
```
Shnoulle, line 69: Maybe add forced superadmin here ?

gabrieljenik: Seems interesting...
What would be the expected value to assert?

PR #3447, opened 2023-09-14

gabrieljenik:


application/models/User.php
```diff
@@ -981,4 +981,48 @@ public function canEdit($loginId)
...
+        return $criteriaPerm;
+    }
+
+    /**
+     * Uses for scope in list and search
+     * This don't mean user are allowed to read all information or any other Permission
+     * @param integer $userid
+     * @return self
+     */
+    public function withListRight($userid = null)
```
Shnoulle, line 1023: Oh ?

Shnoulle (approved):

Shnoulle (changes requested):
Need to be in model too : https://github.com/LimeSurvey/LimeSurvey/blob/5c4fb4e0137d8e564c14d30c3c9563ffd7ecaacd/application/helpers/remotecontrol/remotecontrol_handle.php#L409C31-L409C31

tiborpacalat:
And some checks are failing. Please make it green :)

gabrieljenik:
> Need to be in model too : https://github.com/LimeSurvey/LimeSurvey/blob/5c4fb4e0137d8e564c14d30c3c9563ffd7ecaacd/application/helpers/remotecontrol/remotecontrol_handle.php#L409C31-L409C31

What do you mean in model? Check ownership permission in model?
I think usually those checks are outside of the model, right?
I would just add the check on the RC if that's what you mean

Shnoulle:
> > Need to be in model too : https://github.com/LimeSurvey/LimeSurvey/blob/5c4fb4e0137d8e564c14d30c3c9563ffd7ecaacd/application/helpers/remotecontrol/remotecontrol_handle.php#L409C31-L409C31
> 
> What do you mean in model? Check ownership permission in model? I think usually those checks are outside of the model, right? I would just add the check on the RC if that's what you mean

It's a question, not an affirmation. ping @olleharstedt 

In model : it fix for all controller.

Else here : must be reported for RC too.

Shnoulle (approved):

PR #3448, opened 2023-09-15
Fixed issue #18986: superadmin without create superadmin can create superadmin

Shnoulle:
Dev: specific test for Permissions['superadmin']

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -485,7 +485,7 @@ class UserManagementController extends LSBaseController
         // Check permissions
         $aBasePermissions = Permission::model()->getGlobalBasePermissions();
         // superadmin permission always need create
-        if(!Permission::model()->hasGlobalPermission('superadmin', 'create')) {
+        if (!Permission::model()->hasGlobalPermission('superadmin', 'create')) {
             unset($aBasePermissions['superadmin']);
         }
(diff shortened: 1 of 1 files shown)


Shnoulle:
Done inside model function but not in this controller.

See https://github.com/LimeSurvey/LimeSurvey/blob/5c4fb4e0137d8e564c14d30c3c9563ffd7ecaacd/application/models/Permission.php#L296

gabrieljenik (approved):
Code looks good from structure point of view. Haven't analyzed the solution nor tested.

application/controllers/UserManagementController.php
```diff
@@ -506,8 +506,10 @@ public function actionUserPermissions()
                 }
             }
             $aBasePermissions = $aFilteredPermissions;
+        } elseif(!Permission::model()->hasGlobalPermission('superadmin', 'create')) {
+            /* superadmin permission always need create */
```
gabrieljenik, line 510: Why using multiline comments?
`//` isn't a better fit?

Shnoulle: It's still an one line comment

Shnoulle: I think i make a simple version for master : just unset the superadmin in arrau of Permission. I do same here.

I really think it more clear

PR #3454, opened 2023-09-18
Fixed issue #18961: {SURVEYURL} didn't work with plugin using cli event

gabrieljenik:
Please consider the following when reviewing:
Had to add these to the config:
- Config section ('publicurl'=>'https://limesurvey.example.org:8000/limesurvey') 
- Components section ('request ' => ['baseUrl' => '/limesurvey'])

As this is only required when running ConsoleApplication, maybe there is the need of having specific config for console?


Shnoulle (commented):
No issue in code (seems cool) BUT : seems not related to the issue.

Shnoulle (approved):

Shnoulle:
> No issue in code (seems cool) BUT : seems not related to the issue.

OK, totally related to this issue. More clear now.

Shnoulle:
My own test

Checked with no update on sendMail Cron 
- OK with publicurl
- KO with only sendMailCron config : https://gitlab.com/SondagesPro/mailing/sendMailCron/-/blob/master/sendMailCron.php?ref_type=heads#L696-703


With https://github.com/LimeSurvey/LimeSurvey/commit/55d9eda4c9407db6982dfad0ae8355dfc3c0d572 (parent of https://github.com/LimeSurvey/LimeSurvey/commit/3c7045423a5918987840f2723d7ca2e7f4097823 ) :

- OK with only sendMailCron config

application/models/Traits/LSApplicationTrait.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Returns the relative URL for the application while
+     * considering if a "publicurl" config parameter is set to a valid url
+     * @param boolean $absolute whether to return an absolute URL. Defaults to false, meaning returning a relative one.
+     * @return string the relative or the configured public URL for the application
+     */
+    public function getPublicBaseUrl($absolute = false)
+    {
```
Shnoulle, line 35: I think it broke with current sendMailCron because there are no use of $baseUrl = CWebApplication::getBaseUrl($absolute); ? unsure ?

gabrieljenik:
> * KO with only sendMailCron config : https://gitlab.com/SondagesPro/mailing/sendMailCron/-/blob/master/sendMailCron.php?ref_type=heads#L696-703

Depends on what KO means or how it fails.
Still, if running from CLI, there is now way the app to know a full url, as the related variables on SERVER/REQUEST are not set. The app has no clue about the domain when called from CLI.

Shnoulle:
> Still, if running from CLI, there is now way the app to know a full url, as the related variables on SERVER/REQUEST are not set. The app has no clue about the domain when called from CLI.

senfMailCron plugin didn't work with the new version, need update. Before https://github.com/LimeSurvey/LimeSurvey/commit/3c7045423a5918987840f2723d7ca2e7f4097823 it work.

sendMailCron use Yii 
- `Yii::app()->request->hostInfo` https://www.yiiframework.com/doc/api/1.1/CHttpRequest#setHostInfo-detail 
- `Yii::app()->getUrlManager()->setBaseUrl($baseUrl);` https://www.yiiframework.com/doc/api/1.1/CHttpRequest#setBaseUrl-detail

Your solution disable this usage.  I really think it must allow it … (but i'm not the tester)

@maziminke have other plugins to send email via CLI.

gabrieljenik:
> Your solution disable this usage.

Please don't blame on me. Not fair to blame.

Also
- I just ported code or adapted prior code.
- https://github.com/LimeSurvey/LimeSurvey/commit/3c7045423a5918987840f2723d7ca2e7f4097823 is not my code.

Shnoulle:
> > Your solution disable this usage.
> 
> Please don't blame on me. Not fair to blame.

Yes, sorry … i didn't want to blame you.

I mean : the current fix didn't allow usage of Yii1 internal function. 

> I just ported code or adapted prior code.

Looking at the difference : seems the new updated version (current PR) didn't use `Yii::app()->getBaseUrl(true); ` but it's here … i don't understand …

gabrieljenik:
> I mean : the current fix didn't allow usage of Yii1 internal function.

Host Info could be something to enhance the url methods added on Application component/trait.

mbischof:
see https://www.yiiframework.com/doc/api/1.1/CUrlManager#baseUrl-detail
urlManager has an entry in application/config/config.php, so imho there is no need for an extra `publicurl` setting

Shnoulle:
> urlManager has an entry in application/config/config.php, so imho there is no need for an extra `publicurl` setting

Yes, BUT : some user need only publicurl, admin user access with limesurvey.intra/admin . with baseUrl : you set ALL url using the server, for example css link use absolute url.

It's not the same objective here. publicurl used ONLY for survey link or other publi link. Not for css, js etc …

Shnoulle (commented):
Place for the Trait

PR #3468, opened 2023-09-19
Fixed issue #18938: question_template_attribute show only default value

Shnoulle:
Dev: check strval of XML for i18, this allow empty string and 0 for false

Dev:

### DIFF
--- a/application/models/QuestionTemplate.php
+++ b/application/models/QuestionTemplate.php
@@ -285,7 +285,7 @@ class QuestionTemplate extends CFormModel
                     $this->aCustomAttributes = array();
                     foreach ($this->oConfig->attributes->attribute as $oCustomAttribute) {
                         $attribute_name = (string) $oCustomAttribute->name;
-                        if (isset($oCustomAttribute->i18n) && $oCustomAttribute->i18n) {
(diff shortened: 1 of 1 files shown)


gabrieljenik (approved):
Code structure look OK. No sure why strval does the trick, but if it works, great!

Shnoulle:
> Code structure look OK. No sure why strval does the trick (part of the analysis/solution I will not dive in), but if it works, great!

We get a SimpleXMLElement element not the value. Maybe there are some other issue ? I don't know why it work for core value ([included theme](https://github.com/LimeSurvey/LimeSurvey/tree/5.x/themes/question) don't have `<i18n><i18n>` xml element, core have it : maybe need to move some code here too ?)

`<i18n>0<i18n>` give SimpleXMLElement#1 ( [0] => '0' )' it's true
`<i18n><i18n>` give SimpleXMLElement#1 ( ) it's true

Shnoulle:
I think i git the difference : `$attributes = json_decode(json_encode((array)$xml_config->attributes), true);` at https://github.com/LimeSurvey/LimeSurvey/blob/c7ef106130924dd03b447a1da6f021ebd700dc1f/application/models/QuestionTheme.php#L1017

I put the link in comment on master

PR #3469, opened 2023-09-19
Fixed issue #19100: Store DOM XSS when creating a survey

Shnoulle:
Dev: Need superadmin or XSS off
Dev: fix aria issue (better no aria than bad aria …)

Dev:


gabrieljenik (approved):
Code structure looks OK

PR #3475, opened 2023-09-20
Fixed issue #18968: User with "View/read Users" permission needs to have "User details" action available for all users, not just itself

gabrieljenik:
User details option is now available for users with users/read permission.


Shnoulle (approved):
Yes !

application/models/User.php
```diff
@@ -480,7 +481,7 @@ public function getManagementButtons()
                 'data-href' => $detailUrl,
             ],
             'enabledCondition' =>
-                $permission_superadmin_read
+                $permission_superadmin_read || $permission_users_read
```
Shnoulle, line 484: `$permission_users_read` can be checked only : superadmin have user permission

PR #3481, opened 2023-09-21
Fixed issue #18938: question_template_attribute show only default value

Shnoulle:
Dev: check strval of XML for i18, this allow empty string and 0 for false


Shnoulle:
See https://github.com/LimeSurvey/LimeSurvey/pull/3468

gabrieljenik (approved):

PR #3485, opened 2023-09-22
Fixed issue [security] #18986: superadmin without create superadmin can create superadminu

Shnoulle:
Dev: unset Permissions['superadmin'] for all action (in controller)


gabrieljenik (approved):
Code structure looks good. Haven't tested nor analyzed the issue

PR #3487, opened 2023-09-22
 Fixed issue [security] #19106: Deleted account still has permissions

Shnoulle:
Dev: check user existence before Permission
 Dev: delete related Permission and settings  after user deletion

### DIFF
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -715,6 +715,10 @@ class LSYii_Application extends CWebApplication
         if (App() instanceof CConsoleApplication) {
             return null;
         }
+        if(empty(App()->session['loginID'])) {
+            // NULL for guest
+            return App()->session['loginID'];
+        }
         if (!is_null($this->currentUserId) && $this->currentUserId == App()->session['loginID']) {
             return $this->currentUserId;
         }
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -56,9 +56,6 @@ class LSYii_Application extends CWebApplication
      */
     protected $dbVersion;
 
-    /* @var integer| null the current userId for all action */
-    private $currentUserId;
-
     /**
      *
      * Initiates the application
@@ -703,26 +700,5 @@ class LSYii_Application extends CWebApplication
             $baseUrl = $sPublicUrl;
         }
         return $baseUrl;
-    }
-
-    /**
-     * get the current id of connected user,
-     * check if user exist before return for security
-     * @return int|null user id
-     */
-    public function getCurrentUserId()
-    {
-        if(empty(App()->session['loginID'])) {
-            // NULL for guest
-            return App()->session['loginID'];
-        }
-        if (!is_null($this->currentUserId) && $this->currentUserId == App()->session['loginID']) {
-            return $this->currentUserId;
-        }
-        $this->currentUserId = App()->session['loginID'];
-        if (!User::model()->findByPk($this->currentUserId)) {
-            $this->currentUserId = 0;
-        }
-        return $this->currentUserId;
     }
 }
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -18,6 +18,7 @@
  */
 
 require_once(dirname(dirname(__FILE__)) . '/helpers/globals.php');
+require_once __DIR__ . '/Traits/LSApplicationTrait.php';
 
 /**
 * Implements global config
@@ -36,6 +37,8 @@ require_once(dirname(dirname(__FILE__)) . '/helpers/globals.php');
 */
 class LSYii_Application extends CWebApplication
 {
+    use LSApplicationTrait;
+
     protected $config = array();
 
     /**
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -712,9 +712,6 @@ class LSYii_Application extends CWebApplication
      */
     public function getCurrentUserId()
     {
-        if (App() instanceof CConsoleApplication) {
-            return null;
-        }
         if(empty(App()->session['loginID'])) {
             // NULL for guest
             return App()->session['loginID'];
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -715,7 +715,7 @@ class LSYii_Application extends CWebApplication
         if (App() instanceof CConsoleApplication) {
             return null;
         }
-        if (!is_null($this->currentUserId)) {
+        if (!is_null($this->currentUserId) && $this->currentUserId == App()->session['loginID']) {
             return $this->currentUserId;
         }
         $this->currentUserId = App()->session['loginID'];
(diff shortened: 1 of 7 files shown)


application/models/Permission.php
```diff
@@ -720,15 +724,35 @@ private static function comparePermissionTitle($aApermission, $aBpermission)
...
         }
         return $iUserID;
     }
+
+    /**
+     * get the current id of connected user
+     * @return int|null user id
+     * @throws Exception
+     */
+    public static function getCurrentUserId()
```
Shnoulle, line 741: Maybe in https://github.com/LimeSurvey/LimeSurvey/blob/master/application/core/LSWebUser.php ?

Shnoulle:
Unsure on the good place for the new function ?
LSWebUser ?

Shnoulle:
OK, i need to fix the test too : can set App()->user ?

application/models/Permission.php
```diff
@@ -720,15 +724,35 @@ private static function comparePermissionTitle($aApermission, $aBpermission)
     public function getUserId($iUserID = null)
     {
         if (empty($iUserID)) {
-            if (Yii::app() instanceof CConsoleApplication) {
-                /* Alt : return 1st forcedAdmin ? */
+            if (App() instanceof CConsoleApplication) {
```
gabrieljenik, line 727: Wondering, what's the difference in between App() and Yii::app(). I guess there is a slight difference...

application/models/Permission.php
gabrieljenik: Why not in LsYiiApplication?

application/models/Permission.php
```diff
@@ -720,15 +724,35 @@ private static function comparePermissionTitle($aApermission, $aBpermission)
...
+
+    /**
+     * get the current id of connected user
+     * @return int|null user id
+     * @throws Exception
+     */
+    public static function getCurrentUserId()
+    {
+        if (App() instanceof CConsoleApplication) {
+            return null;
```
gabrieljenik, line 744: Wouldn't this be problemátic?

gabrieljenik, line 744: I wonder how this would impact Email Sending crons... isn't there any perission checking there?

application/models/Permission.php
gabrieljenik: Yes, I think this has to be moved.

application/models/User.php
```diff
@@ -148,6 +148,35 @@ public function attributeLabels()
...
     }
 
+    /**
+     * @inheritDoc
+     * Delete user in related model after deletion
+     * return void
+     **/
+    protected function afterDelete()
+    {
+        parent::afterDelete();
```
gabrieljenik, line 158: As this are dependant/child relations, should this happen after delete or before delete?
Shouldn't we take advantage of relations in the model?

gabrieljenik (changes requested):
Let's move `getCurrentUserId` to Application?

application/models/Permission.php
Shnoulle: No diiference App is a shortcut for Yii:app

application/models/User.php
Shnoulle: after delete clearly

For the other : it work like that.

Shnoulle:
> Let's move `getCurrentUserId` to Application?

And use session and not user …

application/models/Permission.php
Shnoulle: No, you don't have to check Permission in console

application/models/User.php
Shnoulle: 1. Doing this afterDelete is clear (plugin can disable deletion for example)
2. Unsure we can do $this-> since user don't exist (to check)
3. We don't use relation in https://github.com/LimeSurvey/LimeSurvey/blob/dc083bd9e3280fbe58da885f157c6e48707cae2f/application/models/Survey.php#L324

Shnoulle:
Need a reset for test …

application/core/LSYii_Application.php
```diff
@@ -56,6 +56,9 @@ class LSYii_Application extends CWebApplication
      */
     protected $dbVersion;
 
+    /* @var integer| null the current userId for all action */
+    private $currentUserId;
```
gabrieljenik, line 60: Here or on the trait (which maybe not here yet because of other story not merged yet)?

application/models/Permission.php
```diff
@@ -35,6 +35,9 @@ class Permission extends LSActiveRecord
     /* @var array[]|null The global base Permission LimeSurvey installation */
     protected static $aGlobalBasePermissions;
 
+/* @var interger| null the current userId for all permission check */
+    private static $currentUserId;
```
gabrieljenik, line 39: Is this necesarry? Wouldn't this create problems as being duplicaed from Application?

application/core/LSYii_Application.php
Shnoulle: I do,'t understand your question ? There are no trait for LSYii_Application

application/models/Permission.php
Shnoulle: I don't remove it ?

Sorry … must be done

application/core/LSYii_Application.php
Shnoulle: OK, got it … think it resolved now ?

gabrieljenik: This bheaviour may be needed in both console application and web application, so that's why I am saying to add it to a trait. We created a trait for 18961. Still that is not merged. That was what I was referring to.

gabrieljenik:
Don't you think we should create a trait for the new stuff that you added on application/core/LSYii_Application.php ?
That way we can apply the code console app as well as LsYii app.

Shnoulle:
> Don't you think we should create a trait for the new stuff that you added on application/core/LSYii_Application.php ? That way we can apply the code console app as well as LsYii app.

It's a security fix, i already rewrite more than needed , no ?

Because : we can add and add again fix, but here : it's a security bug fix.

Shnoulle:
> Don't you think we should create a trait for the new stuff that you added on application/core/LSYii_Application.php ? That way we can apply the code console app as well as LsYii app.

We don't have Trait , tell me where to put the function …

Shnoulle:
Oh sh… broken …

application/models/Permission.php
```diff
@@ -720,15 +721,15 @@ private static function comparePermissionTitle($aApermission, $aBpermission)
     public function getUserId($iUserID = null)
```
gabrieljenik, line 721: As you are editing this, I wonder, do you know what this is used for? Somehow I think we need to add a `@todo`  here for replacing its usage, in the future, for App->getCurrentUserId().

I think I have missed this comment on the master branch. Maybe the `@todo` is even more important to be added there. What do you think? Just a comment, not request.

gabrieljenik (approved):

application/models/Permission.php
Shnoulle: @deprecated ? And update usage directly ?

In master only then.

PR #3488, opened 2023-09-22
New feature #19109: Allow 'remove' in cpd_importParticipants

TijZwa:
New feature #19109
Dev:
I have a single source of truth, where all my participants are registered.
Using the function 'cpd_importParticipants' I can import and update participants. I always keep the ID's in sync.

I want the function 'cpd_importParticipants' to delete all participants from the CPD if their ID is not in the import list.

## Summary by CodeRabbit

* **New Features**
  * Added optional participant removal during imports. When enabled, participants not included in the import list are automatically removed, with the removal count reported in the import status.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -3888,7 +3888,6 @@ class remotecontrol_handle
         if ($remove) {
             $allParticipants = Participant::model()->getParticipantsWithoutLimit();
             foreach ($allParticipants as $dbParticipant) {
-
                 if (!array_search($dbParticipant["participant_id"], array_column($participants, "participant_id"))) {
                     Participant::model()->deleteParticipants($dbParticipant["participant_id"]);
                     $aResponse['RemoveCount']++;
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -3886,15 +3886,11 @@ class remotecontrol_handle
         // Remove param is set to True
         // Remove all participants whose ID does not appear in the import list.
         if ($remove) {
-            $allParticipants = Participant::model()->getParticipantsWithoutLimit();
-            foreach ($allParticipants as $dbParticipant) {
-                if (!array_search($dbParticipant["participant_id"], array_column($participants, "participant_id"))) {
-                    Participant::model()->deleteParticipants($dbParticipant["participant_id"]);
-                    $aResponse['RemoveCount']++;
-                }
-            }
+    		$participantsId = array_column($participants, 'participant_id');
+    		$criteria = new CDbCriteria();
+    		$criteria->addNotInCondition('participant_id', $participantsId);
+    		$aResponse['RemoveCount'] = Participant::model()->deleteAll($criteria);
         }
-
         return $aResponse;
     }
 
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -3780,7 +3780,6 @@ class remotecontrol_handle
     }
 
 
-
     /**
      * Import a participant into the LimeSurvey CPDB
      *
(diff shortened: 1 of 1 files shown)


application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -3881,6 +3883,18 @@ public function cpd_importParticipants($sSessionKey, $participants, $update = fa
             }
         }
 
+        // Remove param is set to True
+        // Remove all participants whose ID does not appear in the import list.
+        if ($remove) {
+            $allParticipants = Participant::model()->getParticipantsWithoutLimit();
```
olleharstedt, line 3889: This might not scale. If nr of participants is too high, RAM will explode. Fetch by batch query or use other strategy.

Shnoulle: It's not related to this commit here. It's an existing `Participant::model()->getParticipantsWithoutLimit();` issue. And this function exist.

Shnoulle: (But a deleteAll($criteria) with criteria IN array can be cool

c-schmitz: Still, then it is not the correct function to use.

Shnoulle: > Still, then it is not the correct function to use.

Yes, and maybe cpd_removeParticipants new function is better, no ?

TijZwa: How to move forward with this PR?

olleharstedt: Either write a custom query to apply the change without using the ActiveRecord system, or use the batch query system in Yii. Hm I'm only able to find batch query for Yii 1.1, not 2. Perhaps your google skills are better? Or ask on their Slack channel.

olleharstedt: Custom queries should use the query builder, btw, preferably not raw SQL (because of the different vendors we support).

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -3881,6 +3883,18 @@ public function cpd_importParticipants($sSessionKey, $participants, $update = fa
             }
         }
 
+        // Remove param is set to True
```
Shnoulle, line 3886: All of this can be replaced by

```
Can be replaced by 
$participantsId = array_column($participants, 'participant_id');
$criteria = new CDbCriteria();
$criteria->addNoInCondition('participant_id', $participantsId);
$aResponse['RemoveCount'] = Participant::model()->deleteAll($criteria);
```

Shnoulle (commented):
1. Never call ALL db values
2. Use deleteAll and not foreach

application/helpers/remotecontrol/remotecontrol_handle.php
Shnoulle: A deleteAlll can fix the issue … one call with a [NOT IN](https://www.yiiframework.com/doc/api/1.1/CDbCriteria#addNotInCondition-detail)

olleharstedt: Right, good one :)

application/helpers/remotecontrol/remotecontrol_handle.php
TijZwa: Thanks! 
addNoInCondition => addNo**t**InCondition

TijZwa:
@Shnoulle @olleharstedt Thanks for the directions. I've updated the PR.

application/helpers/remotecontrol/remotecontrol_handle.php
Shnoulle: Yes ;)

Shnoulle (approved):

c-schmitz:
@coderabbitai review

c-schmitz:
@TijZwa Please check out the comments by rabbitai - these issues need to be addressed first by you.
Would you care to fix these? Or should we rather close the PR after such a long time?

TijZwa:
Hi @c-schmitz,

I don’t use this software anymore since I’ve moved to another employer. I’ll reach out to one of my former colleagues to confirm whether they still use it.

PR #3491, opened 2023-09-26
Fixed issue #CR-1347: duplicate entries in template configuration and incorrect theme being saved in general settings

ptelu:


Shnoulle:
Plugin using afterFindSurvey event need a way to reset the static variable. Else the only way to have the updated value is to do it in beforeControllerAction.

https://gitlab.com/SondagesPro/coreAndTools/reloadAnyResponse/-/blob/master/reloadAnyResponse.php?ref_type=heads#L663

ptelu:
> Plugin using afterFindSurvey event need a way to reset the static variable. Else the only way to have the updated value is to do it in beforeControllerAction.
> 
> https://gitlab.com/SondagesPro/coreAndTools/reloadAnyResponse/-/blob/master/reloadAnyResponse.php?ref_type=heads#L663

Thanks for checking, will have a look it before finishing this

application/models/Survey.php
```diff
@@ -149,15 +149,7 @@ class Survey extends LSActiveRecord implements PermissionInterface
...
-    /**
-     * This is a static cache, it lasts only during the active request. If you ever need
-     * to clear it, like on activation of a survey when in the same request a row is read,
-     * saved and read again you can use resetCache() method.
-     *
-     * @var array $findByPkCache
-     */
-    protected $findByPkCache = array();
-
+    protected static $cacheSurvey = [];
```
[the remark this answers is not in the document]
ptelu: i kept it, it makes sense to have the reset

application/helpers/update/updates/Update_616.php
```diff
@@ -0,0 +1,93 @@
...
+     */
+    private function deleteDuplicateTemplateConfigurationEntries()
+    {
+        // count all entries
+        $iTemplateConfigurationCount = $this->db->createCommand()
+            ->select('count(*)')
+            ->from('{{template_configuration}}')
+            ->queryScalar();
+        $limit = 1000;
+        for ($i = 0; $i < $iTemplateConfigurationCount; $i += $limit) {
```
Shnoulle, line 32: Arg … since we can have 500 template_configuration : this can broke easily update on some installation.

No way to use join etc ? 

We **must** avoid foreach in update part … clearly …

application/helpers/update/updates/Update_616.php
```diff
@@ -0,0 +1,93 @@
...
+            $aTemplateConfigurations = $this->db->createCommand()
+                ->select('template_name, sid, gsid, uid')
+                ->from('{{template_configuration}}')
+                ->limit($limit)
+                ->queryAll();
+            foreach ($aTemplateConfigurations as $aTemplateConfiguration) {
```
Shnoulle, line 38: A foreach inside a for ?

application/models/SurveysGroupsettings.php
```diff
@@ -356,6 +356,24 @@ public static function getInstance($iSurveyGroupId = 0, $oSurvey = null, $instan
...
+                $instance->oOptions->template = 'inherit';
+            }
+        }
+
+        // check the global configuration for template inheritance if surveygroup is 0 (global survey) and template set to inherit
+        if (
+            $iSurveyGroupId === 0
+            && $instance->shouldInherit('template')
+        ) {
+            $instance->oOptions->template = App()->getConfig('defaulttheme');
```
Shnoulle, line 374: SurveyGroup->id = 0 ?

application/helpers/update/updates/Update_616.php
twilligls: I think deleting duplicate entries in a table in a pure sql manner is way more inefficient and time consuming (especially considering we have to use mySQL, pgsql and mssql syntax).
To be more save @ptelu could maybe think about bringing down the limit.

Shnoulle: > I think deleting duplicate entries in a table in a pure sql manner is way more inefficient and time consuming (especially considering we have to use mySQL, pgsql and mssql syntax).

unsure on this point, i clearly think we can have more than 5000 template configuration (there are a lot of way to have more than one theme by surveys, and have 3000 surveys is not rare) : the we star by a `for (1 to 5000)` with multiple SQL request inside …

Please : try this with a old instance with more than 1000 sureys and SQL server not on localhost …

For reminder : when update  DB to 400, it can take more than 30 minutes, do it on SQL only really fix it : https://github.com/LimeSurvey/LimeSurvey/pull/1282

ptelu: I will see if i can make this update more performant

Shnoulle: Thank you :)

PS :  i use CLI for updating each time for major release, and some for all release  (and my demo instance). But here i really think we have 90 percent chance of negative feedback on the forum.

ptelu: so just some numbers regarding the query, i ran it with 17717 entries from which are 445 correct.

With the original query it took 3.6436128616333 seconds
with my initial sql query with subselect it took 240.41818499565 seconds
`        $subquery = $this->db->createCommand()
            ->select("MAX(id)")
            ->from("(SELECT * FROM {{template_configuration}}) AS sub_template_configuration")
            ->group(['template_name', 'sid', 'gsid', 'uid'])
            ->getText();
        $criteria = $this->db->getCommandBuilder()->createCriteria();
        $criteria->select = 'id, template_name, sid, gsid, uid';
        $criteria->addCondition("id NOT IN ($subquery)");
        $this->db->getCommandBuilder()->createDeleteCommand('{{template_configuration}}', $criteria)->execute();`

with optimzed sql query it took 0.3056058883667 seconds
`        $aKeepIds = $this->db->createCommand()
            ->select("MAX(id) AS maxRecordId")
            ->from("{{template_configuration}}")
            ->group(['template_name', 'sid', 'gsid', 'uid'])
            ->queryAll();
        $aKeepIds = array_column($aKeepIds, 'maxRecordId');
        $criteria = $this->db->getCommandBuilder()->createCriteria();
        $criteria->select = 'id, template_name, sid, gsid, uid';
        $criteria->addNotInCondition('id', $aKeepIds);
        $this->db->getCommandBuilder()->createDeleteCommand('{{template_configuration}}', $criteria)->execute();`

Shnoulle: subselect is never a good idea …

Max seems great ! No foreach ?  😍

Shnoulle: Else

> so just some numbers regarding the query, i ran it with 17717 entries from which are 445 correct.

😩 What ?

Shnoulle:
💓

PR #3493, opened 2023-09-26
Fixed issue [security] #19106: Deleted account still has permissions

Shnoulle:
Dev: check user existence before Permission
Dev: delete related Permission and settings after user deletion
Dev: always check session value

Dev:

### DIFF
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -18,6 +18,7 @@
  */
 
 require_once(dirname(dirname(__FILE__)) . '/helpers/globals.php');
+require_once __DIR__ . '/Traits/LSApplicationTrait.php';
 
 /**
 * Implements global config
@@ -36,6 +37,8 @@ require_once(dirname(dirname(__FILE__)) . '/helpers/globals.php');
 */
 class LSYii_Application extends CWebApplication
 {
+    use LSApplicationTrait;
+
     protected $config = array();
 
     /**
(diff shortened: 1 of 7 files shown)


application/core/LSYii_Application.php
```diff
@@ -711,4 +714,28 @@ public function getPublicBaseUrl($absolute = false)
...
     }
+
+    /**
+     * get the current id of connected user,
+     * check if user exist before return for security
+     * @return int|null user id
+     */
+    public function getCurrentUserId()
+    {
+        if (App() instanceof CConsoleApplication) {
```
Shnoulle, line 725: Maybe remove it ? Just leave `if(empty(App()->session['loginID'])) {`

gabrieljenik:
Will review after we wrap the discussion on the other branch

application/core/LSYii_Application.php
gabrieljenik: Yes, seems better. 

Also, I am not sure CConsoleApplication will create LSYii_Application

Shnoulle:
Trait added

gabrieljenik (approved):
Code looks OK!

application/models/Permission.php
```diff
@@ -769,8 +769,7 @@ public function getUserId($iUserID = null)
                 /* Alt : return 1st forcedAdmin ? */
                 throw new Exception('Permission must not be tested with console application.');
             }
-            /* See TestBaseClass tearDownAfterClass */
```
gabrieljenik, line 772: Maybe we need to add `@todo` to have its usage be replaced by App->getCurrentUserId ?

Shnoulle: @deprecated on the function ? Good idea !

Shnoulle: Not this one, this one is linked to Permission : Permission must not test console application currently.

Or only on a 7.X version

Shnoulle:
We have same issue for App()->user->id : we must avoid usage too …

Shnoulle:
@gabrieljenik something like this ? https://github.com/LimeSurvey/LimeSurvey/pull/3493/commits/f505c290e340bf4beaf33cb0d38e8480a8308121

application/models/Permission.php
gabrieljenik: Not sure why ... but I will follow you.. :) Don't add comment then :)

gabrieljenik:
> @gabrieljenik something like this ? [f505c29](https://github.com/LimeSurvey/LimeSurvey/commit/f505c290e340bf4beaf33cb0d38e8480a8308121)

Thanks

Shnoulle:
I create a Dev issue when all of this User management issue are fixed …

PR #3494, opened 2023-09-27
Fixed issue #19115: Menu toggler appears when not needed

Shnoulle:
Fixed issue : invalid link without any text at start
Fixed issue : potential invalid HTML empty ul
Dev: add a var to check if menu is needed
Dev: remove invalid link


Shnoulle:
> Code looks Ok. Should this be in other themes as well? if so, should this be enclosed in some kind of function or central place?

It's in Vanilla : parent of boostrap and fruity. According to reporter : fruity_twentythree dodn't have the issue (but yes : code can be improved).

Shnoulle:
> > Code looks Ok. Should this be in other themes as well? if so, should this be enclosed in some kind of function or central place?
> 
> It's in Vanilla : parent of boostrap and fruity. According to reporter : fruity_twentythree dodn't have the issue (but yes : code can be improved).

Not the same code : https://github.com/LimeSurvey/LimeSurvey/blob/25f63d5758134ad5eca607453deb8a9e94b7f2bd/themes/survey/fruity_twentythree/views/subviews/header/nav_bar.twig#L27

Only one test, no var needed.

PR #810, opened 2017-09-20
New feature #12690: Allow super-admin set by other user's than 1

Shnoulle:
- this add [supermadmin][create] by DB used for managing (set supermadmin/read and supermadmin/create)
- add a App()->getConfig('superadmin_force');
- Strangely : any superadmin can update email of user->uid == 1 before, leave like this

PR #3495, opened 2023-09-27
Fixed issue [security] #19107: Superadmin can update any admin's information

Shnoulle:
Dev: move all test to UserManager (maybe need to be in User->hasPermission)
Dev: forcedsuperamdin can update all other user (include forcedsuperamdin)
Dev: forcedsuperamdin can be updated only by forcedsuperamdin
Dev: superadmin can be update only by superamdin with create permission (maybe need update êrmission)
Dev: other can be updated by superadmin or parent

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -1322,7 +1322,10 @@ class UserManagementController extends LSBaseController
         // Abort if logged in user has no access to this user.
         // Using same logic as User::getButtons().
         $userManager = new UserManager(Yii::app()->user, $oUser);
-        if (!$userManager->canEdit() || $aUser['uid'] == Yii::app()->user->id) {
+        if (
+            !$userManager->canEdit()
+            || $aUser['uid'] == Yii::app()->user->id // To update self : must use personal settings
+        ) {
             throw new CHttpException(403, gT("You do not have permission to access this page."));
         }
 
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -1310,21 +1310,18 @@ class UserManagementController extends LSBaseController
     /**
      * Update admin-user
      *
-     * REFACTORED (in UserManagementController)
-     *
      * @param array $aUser array with user details
      * @return object user - updated user object
      * @throws CException
      */
     public function updateAdminUser(array $aUser): User
     {
-        $oUser = User::model()->findByPk($aUser['uid']);
+        $oUser = $this->loadModel($aUser['uid']);
         // Abort if logged in user has no access to this user.
         // Using same logic as User::getButtons().
-        $userManager = new UserManager(Yii::app()->user, $oUser);
         if (
-            !$userManager->canEdit()
-            || $aUser['uid'] == Yii::app()->user->id // To update self : must use personal settings
+            !$oUser->canEdit()
+            || $aUser['uid'] == App()->user->id // To update self : must use personal settings
         ) {
             throw new CHttpException(403, gT("You do not have permission to access this page."));
         }
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -1319,20 +1319,10 @@ class UserManagementController extends LSBaseController
     public function updateAdminUser(array $aUser): User
     {
         $oUser = User::model()->findByPk($aUser['uid']);
-        //If the user id of the post is spoofed somehow it would be possible to edit superadmin users
-        //Therefore we need to make sure no non-superadmin can modify superadmin accounts
-        //Since this should NEVER be the case without hacking the software, this will silently just do nothing.
-        if (
-            !Permission::model()->hasGlobalPermission('superadmin', 'read', Yii::app()->user->id)
-            && Permission::model()->hasGlobalPermission('superadmin', 'read', $oUser->uid)
-        ) {
-            throw new CException("This action is not allowed, and should never happen", 500);
-        }
-
         // Abort if logged in user has no access to this user.
         // Using same logic as User::getButtons().
         $userManager = new UserManager(Yii::app()->user, $oUser);
-        if (!$userManager->canEdit()) {
+        if (!$userManager->canEdit() || $aUser['uid'] == Yii::app()->user->id) {
             throw new CHttpException(403, gT("You do not have permission to access this page."));
         }
 
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -540,13 +540,8 @@ class User extends LSActiveRecord
             'linkAttributes'   => [
                 'data-href' => $editUrl,
             ],
-            'enabledCondition' =>
-                ($permission_superadmin_read
-                    && $this->uid != 1
-                )
-                || (!$permission_superadmin_read
-                    && $this->canEdit(App()->session['loginID'])
-                )
+            'enabledCondition' => $this->canEdit()
+                                && $this->uid != App()->user->getId()
         ];
         $dropdownItems[] = [
             'title'            => gT('Template permissions'),
@@ -972,15 +967,12 @@ class User extends LSActiveRecord
     /**
      * Returns true if logged in user with id $loginId can edit this user
      *
-     * @param int $loginId
      * @return bool
      */
-    public function canEdit($loginId)
+    public function canEdit()
     {
-        return
-            Permission::model()->hasGlobalPermission('superadmin', 'read')
-            || $this->uid == $loginId
-            || (Permission::model()->hasGlobalPermission('users', 'update') && $this->parent_id == $loginId);
+        $userManager = new UserManager(App()->getUser(), $this);
+        return $userManager->canEdit();
     }
 
     /**
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -973,7 +973,7 @@ class User extends LSActiveRecord
     public function canEdit($managerId = null)
     {
         if (is_null($managerId)) {
-            Permission::model()->getUserId();
+            $managerId = Permission::model()->getUserId();
         }
         /* user can update himself */
         if ($managerId == $this->uid) {
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -965,14 +965,39 @@ class User extends LSActiveRecord
     }
 
     /**
-     * Returns true if logged in user with id $loginId can edit this user
+     * Return true if user with id $managerId can edit this user
+     * @param int|null $managerId default to current user
      *
      * @return bool
      */
-    public function canEdit()
+    public function canEdit($managerId = null)
     {
-        $userManager = new UserManager(App()->getUser(), $this);
-        return $userManager->canEdit();
+        if (is_null($managerId)) {
+            Permission::model()->getUserId();
+        }
+        /* user can update himself */
+        if ($managerId == $this->uid) {
+            return true;
+        }
+        /* forcedsuperamdin (user #1) can always update all */
+        if (Permission::isForcedSuperAdmin($managerId)) {
+            return true;
+        }
+        /* forcedsuperamdin can not be update (except by another forcedsuperamdin done before) */
+        if (Permission::isForcedSuperAdmin($this->uid)) {
+            return false;
+        }
+        /* If target user is superamdin : managingUser must be allowed to create superadmin */
+        if (Permission::model()->hasGlobalPermission('superadmin', 'read', $this->uid)) {
+            return Permission::model()->hasGlobalPermission('superadmin', 'create', $managerId);
+        }
+        /* superamin can update all othert user */
+        if (Permission::model()->hasGlobalPermission('superadmin', 'read', $managerId)) {
+            return true;
+        }
+        /* Finally : simple user can update only childs users */
+        return Permission::model()->hasGlobalPermission('users', 'update', $managerId)
+                && $this->parent_id == $managerId;
     }
 
     /**
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -541,7 +541,7 @@ class User extends LSActiveRecord
                 'data-href' => $editUrl,
             ],
             'enabledCondition' => $this->canEdit()
-                                && $this->uid != App()->user->getId()
+                                && $this->uid != App()->user->getId() // To update self : must use personal settings
         ];
         $dropdownItems[] = [
             'title'            => gT('Template permissions'),
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -987,11 +987,12 @@ class User extends LSActiveRecord
         if (Permission::isForcedSuperAdmin($this->uid)) {
             return false;
         }
-        /* If target user is superamdin : managingUser must be allowed to create superadmin */
+        /* If target user is superamdin : managingUser must be allowed to create superadmin and be parent */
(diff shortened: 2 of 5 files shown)


application/models/services/UserManager.php
```diff
@@ -68,22 +68,38 @@ public function canAssignRole()
...
-            );
+        /* user can update himself */
+        if ($this->managingUser == $this->targetUser) {
+            return true;
+        }
+        /* forcedsuperamdin (user #1) can always update all */
+        if (Permission::isForcedSuperAdmin($this->managingUser->id)) {
+            return true;
+        }
+        /* forcedsuperamdin can not be update (except by another forcedsuperamdin done before) */
```
Shnoulle, line 88: We can move this test before if needed

application/models/services/UserManager.php
```diff
@@ -68,22 +68,38 @@ public function canAssignRole()
...
+        if (Permission::isForcedSuperAdmin($this->targetUser->uid)) {
+            return false;
+        }
+        /* If target user is superamdin : managingUser must be allowed to create superadmin */
+        if (Permission::model()->hasGlobalPermission('superadmin', 'read', $this->targetUser->uid)) {
+            return Permission::model()->hasGlobalPermission('superadmin', 'create', $this->managingUser->id);
```
Shnoulle, line 94: Or check if is parent ?

application/controllers/UserManagementController.php
```diff
@@ -1319,20 +1319,10 @@ public function loadModel(int $id): User
...
-            && Permission::model()->hasGlobalPermission('superadmin', 'read', $oUser->uid)
-        ) {
-            throw new CException("This action is not allowed, and should never happen", 500);
-        }
-
         // Abort if logged in user has no access to this user.
         // Using same logic as User::getButtons().
         $userManager = new UserManager(Yii::app()->user, $oUser);
-        if (!$userManager->canEdit()) {
+        if (!$userManager->canEdit() || $aUser['uid'] == Yii::app()->user->id) {
```
Shnoulle, line 1325: This is not the place to edit self.

gabrieljenik: This rule should be inside the userManager->canEdit() ?

application/models/User.php
```diff
@@ -540,13 +540,8 @@ public function getManagementButtons()
...
             ],
-            'enabledCondition' =>
-                ($permission_superadmin_read
-                    && $this->uid != 1
-                )
-                || (!$permission_superadmin_read
-                    && $this->canEdit(App()->session['loginID'])
-                )
+            'enabledCondition' => $this->canEdit()
+                                && $this->uid != App()->user->getId()
```
gabrieljenik, line 544: This rule should be inside the userManager->canEdit() ?

application/models/User.php
```diff
@@ -972,15 +967,12 @@ public function getGroupMemberListButtons()
     /**
      * Returns true if logged in user with id $loginId can edit this user
      *
-     * @param int $loginId
      * @return bool
      */
-    public function canEdit($loginId)
+    public function canEdit()
```
gabrieljenik, line 972: Changing the signature of the public function?
If so, why not just remove it? Seems to be just an alias.

application/controllers/UserManagementController.php
gabrieljenik: Actually, I am seeing below, isn't there already? Then, is this necesarry?

application/models/User.php
gabrieljenik: Actually, I am seeing below, isn't there already? Then, is this necesarry?

Can we just call userManager->canEdit() ?

application/models/services/UserManager.php
gabrieljenik: Both. Need to be allowed to create superadmin as well as be its parent.

PS: permission superadmin-create, does it exist?

Shnoulle: Yes : superadmin read and create

Before no superadmin can create new superadmin (only superadmin #1) it's a new feature in 4.X (lost in the changelog …) https://github.com/LimeSurvey/LimeSurvey/pull/810

application/controllers/UserManagementController.php
Shnoulle: > This rule should be inside the userManager->canEdit() ?

Before userManager->canEdit()  allow $aUser['uid'] == Yii::app()->user->id


> Actually, I am seeing below, isn't there already? Then, is this necesarry?

No canEdit return true, but  _This is not the place to edit self._

application/models/User.php
Shnoulle: 1. There are 2 function to do the same things …
2. You can not get App()->getUser() with `$loginId`
3. You right, maybe remove 

Or create a 3rd one … with both are shortcut for this 3rd one …  and sometime someone look at other function by other dev and not create new functions again and again …


Then what do you want

1. Stay like that
2. remove this one
3. Create a third one and the 2 other call the 3rd one

application/controllers/UserManagementController.php
Shnoulle: > This rule should be inside the userManager->canEdit() ?

1. [userManager->canEdit()](https://github.com/LimeSurvey/LimeSurvey/blob/25f63d5758134ad5eca607453deb8a9e94b7f2bd/application/models/services/UserManager.php#L82) **allow** user to edit himself
2. UserManagement [must disallow user to edit himself](https://github.com/LimeSurvey/LimeSurvey/blob/25f63d5758134ad5eca607453deb8a9e94b7f2bd/application/models/User.php#L584)  it's the task of personalsettings not usermanagement.

We can allow it, but currently : we hide the button. Let @tiborpacalat choose maybe ? 
_Did user can edit himself in user management._

application/models/User.php
Shnoulle: No , see https://github.com/LimeSurvey/LimeSurvey/pull/3495#discussion_r1338619193

Currently button his hidden for self.

application/models/User.php
gabrieljenik: Let's resolve the other comments (https://github.com/LimeSurvey/LimeSurvey/pull/3495#discussion_r1338619193) first

Shnoulle: > Let's resolve the other comments ([#3495 (comment)](https://github.com/LimeSurvey/LimeSurvey/pull/3495#discussion_r1338619193)) first

In my opinion : there are nothing to resolve here … if we stay like 3,5 and 6 : we need to disallow user to update himself on this way.

And this don't touch this part about canEdit function …

application/controllers/UserManagementController.php
gabrieljenik: > [userManager->canEdit()](https://github.com/LimeSurvey/LimeSurvey/blob/25f63d5758134ad5eca607453deb8a9e94b7f2bd/application/models/services/UserManager.php#L82) allow user to edit himself
UserManagement [must disallow user to edit himself](https://github.com/LimeSurvey/LimeSurvey/blob/25f63d5758134ad5eca607453deb8a9e94b7f2bd/application/models/User.php#L584) it's the task of personalsettings not usermanagement.

Oh, I think I got it. Then I think good.

gabrieljenik: Why don't we add these exact comments in the code?

> [userManager->canEdit()](https://github.com/LimeSurvey/LimeSurvey/blob/25f63d5758134ad5eca607453deb8a9e94b7f2bd/application/models/services/UserManager.php#L82) allow user to edit himself
UserManagement [must disallow user to edit himself](https://github.com/LimeSurvey/LimeSurvey/blob/25f63d5758134ad5eca607453deb8a9e94b7f2bd/application/models/User.php#L584) it's the task of personalsettings not usermanagement.

application/models/User.php
Shnoulle: Adding comment is a good idea : let's go.

application/models/User.php
gabrieljenik: I would just leave this function after all

gabrieljenik (approved):
Code looks OK

application/models/User.php
Shnoulle: > I would just leave this function after all

After reflection : i think you're right here : set it s deprecated (both) and use a new one ?

User->hasPermission ? update and create are specific

Shnoulle: > I would just leave this function after all

Other solution : move function logic here, Service function are shortcut.

application/models/services/UserManager.php
Shnoulle: > Both. Need to be allowed to create superadmin as well as be its parent.
> 

```
        if (Permission::model()->hasGlobalPermission('superadmin', 'read', $this->uid)) {
            return Permission::model()->hasGlobalPermission('superadmin', 'create', $managerId)
                && $this->parent_id == $managerId;
        }

```

application/models/User.php
Shnoulle: > Changing the signature of the public function?

No change now :)

Shnoulle:
Oh no, sorry 

_   │ Exception: Test ls\tests\controllers\UserManagementTest::testUpdateAdminUserTamperproofed failed: _

Shnoulle:
> ls\tests\controllers\UserManagementTest::testUpdateAdminUserTamperproofed failed

Because App()->user->id != Yii::app()->session['loginID']

I check to create a test function to set both if it work. No local test, just a quick test before create function

Shnoulle:
Ouf … @gabrieljenik finally, test corrected ...

Can you review again ?

Doing for 5 after.

application/controllers/UserManagementController.php
```diff
@@ -1310,29 +1310,19 @@ public function loadModel(int $id): User
...
      * @param array $aUser array with user details
      * @return object user - updated user object
      * @throws CException
      */
     public function updateAdminUser(array $aUser): User
     {
-        $oUser = User::model()->findByPk($aUser['uid']);
-        //If the user id of the post is spoofed somehow it would be possible to edit superadmin users
-        //Therefore we need to make sure no non-superadmin can modify superadmin accounts
-        //Since this should NEVER be the case without hacking the software, this will silently just do nothing.
```
gabrieljenik, line 1324: I guess this situations is now talen care on the `userManager->canEdit()` method, right?

application/models/User.php
```diff
@@ -540,13 +540,8 @@ public function getManagementButtons()
             'linkAttributes'   => [
                 'data-href' => $editUrl,
             ],
-            'enabledCondition' =>
-                ($permission_superadmin_read
```
gabrieljenik, line 544: I guess this situations is now talen care on the `userManager->canEdit()` method, right?

application/models/User.php
```diff
@@ -970,17 +965,40 @@ public function getGroupMemberListButtons()
...
     /**
-     * Returns true if logged in user with id $loginId can edit this user
+     * Return true if user with id $managerId can edit this user
+     * @param int|null $managerId default to current user
      *
-     * @param int $loginId
      * @return bool
      */
-    public function canEdit($loginId)
+    public function canEdit($managerId = null)
```
gabrieljenik, line 973: So, this method is more like a `caneBeEditedBy` ?
Maybe we can rename this? (I guess is not widely used)

application/models/User.php
```diff
@@ -970,17 +965,40 @@ public function getGroupMemberListButtons()
...
     {
-        return
-            Permission::model()->hasGlobalPermission('superadmin', 'read')
-            || $this->uid == $loginId
-            || (Permission::model()->hasGlobalPermission('users', 'update') && $this->parent_id == $loginId);
+        if (is_null($managerId)) {
```
gabrieljenik, line 975: Is this code to be here or on the UserManager? @olleharstedt What do you say?

application/models/User.php
Shnoulle: It's the same name than before, renaming it broke API version …

I add again $managerId because you say it broke API versin (and you're right) … then  : **no, i don't rename it**  it's not a new function.

application/models/User.php
Shnoulle: yes

Shnoulle: > I guess this situations is now talen care on the `user->canEdit()` method, right?

It's ALL the point : to move to the same function for testing and show …

application/controllers/UserManagementController.php
Shnoulle: > I guess this situations is now talen care on the `userManager->canEdit()` method, right?

It's ALL the point : to move to the same function for testing and show … 

and one function use another one …

application/models/User.php
Shnoulle: > Is this code to be here or on the UserManager? @olleharstedt What do you say?

This code in the 1st commit was in UserManager but it broke API …

The only way to don't broke API is to do here or to create a new function …

application/models/User.php
Shnoulle: > So, this method is more like a `caneBeEditedBy` ? Maybe we can rename this? (I guess is not widely used)

https://github.com/LimeSurvey/LimeSurvey/pull/3495#discussion_r1338537273 please …

application/models/User.php
Shnoulle: If you user UserManager : you can not have `canEdit($loginId)` because i can not have a App()->user with only id.

Then    `canEdit($loginId)`  CAN NOT use UserManager https://github.com/LimeSurvey/LimeSurvey/blob/f5e3a0f252bae70ec89ba76e1ac938e12df729ac/application/models/services/UserManager.php#L16 

And the original issue are done because we have 3 functions for same usage …

gabrieljenik: > If you user UserManager : you can not have `canEdit($loginId)` because i can not have a App()->user with only id.

Not sure I follow, but I was thinking something like this:

On `getManagementButtons()` you could have:

```
$userManager = new UserManager($this, App()->user->getId())
$canEdit = $userManager->canEdit();
```

Still, not sure these back and forths wort too much. Let's just wait for @olleharstedt to let us know where he thinks the best palce for that logic is :)

Shnoulle: > > If you user UserManager : you can not have `canEdit($loginId)` because i can not have a App()->user with only id.
> 
> Not sure I follow, but I was thinking something like this:
> 
> On `getManagementButtons()` you could have:
> 
> ```
> $userManager = new UserManager(App()->user, $this)
> $canEdit = $userManager->canEdit();
> ```
> 
> Still, not sure these back and forths wort too much. Let's just wait for @olleharstedt to let us know where he thinks the best palce for that logic is :)

Yes : this make https://github.com/LimeSurvey/LimeSurvey/blob/73f26617a5d149a74d611697721fbfda8813c332/application/models/User.php#L978 unusable like this. Then it broke API

I start with this solution : look at https://github.com/LimeSurvey/LimeSurvey/pull/3495/commits/dee58d835c3975d1ebf8fe7141ea603556d27700 and i update after your advice about API version (updating a public function signature)

gabrieljenik: > Your sample is exactly the new function can edit : https://github.com/LimeSurvey/LimeSurvey/commit/dee58d835c3975d1ebf8fe7141ea603556d27700#diff-0a0e2b974470e717a216950e1374f76ba04f409a55f562a9b59f91cb79451ccdR974

Great. Not sure if you are expecting any action from me here or not, sorry

PR #3497, opened 2023-09-28
added xsd extension to the allowedpluginupload setting

mbischof:
Fixed issue #19098: Add XSD files to allowed files for plugin uploads


maziminke:
@c-schmitz was this also merged into 3.x and 5.x? We need it there as well.

PR #3498, opened 2023-09-28
Bug/19108 subquestion save

kevin-foster-uk:
Fix subquestion editing and code uniqueness validation.

This PR fixes several issues:

- Order of subquestions can cause unexpected uniqueness validation error because codes are validated one at a time, at the point of saving each subquestion rather than up front as a set. This was fixed by disabling the unique validation at the point of save and doing the validation of the set before saving.
- Using old code to reference question records can produce unexpected results when updating a subquestion code. Switched to referencing subquestions by qid.
- The subquestion table was not being reloaded after save. 

When testing this we should test these question types:

- Multiple choice (Type: M)
- Array (Texts) (Type: ;)

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -2964,13 +2964,11 @@ class QuestionAdministrationController extends LSBaseController
                     )
                 ) {
                     throw (
-                        new LSUserException(
+                    new LSUserException(
                         500,
                         gT('Could not save subquestion')
-                        )
-                    )->setDetailedErrors(
-                        ['Subquestion codes must be unique.']
-                    );
+                    )
+                    )->setDetailedErrors(['Subquestion codes must be unique.']);
                 }
                 $codes[$subquestionId][$scaleId][] = $data['code'];
             }
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -114,9 +114,9 @@ class QuestionAdministrationController extends LSBaseController
      * @return void
      * @throws CHttpException
      */
-    public function actionEdit(int $qid, string $tabOverviewEditor = null)
+    public function actionEdit(int $questionId, string $tabOverviewEditor = null)
     {
-        $questionId = (int) $qid;
+        $questionId = (int) $questionId;
         if (!in_array($tabOverviewEditor, ['overview', 'editor'], true)) {
             $tabOverviewEditor = null;
         }
@@ -195,6 +195,7 @@ class QuestionAdministrationController extends LSBaseController
         $jsVariablesHtml = $this->renderPartial(
             '/admin/survey/Question/_subQuestionsAndAnwsersJsVariables',
             [
+                'qid'               => $question->qid,
                 'anslangs'          => $question->survey->allLanguages,
                 // TODO
                 'assessmentvisible' => false,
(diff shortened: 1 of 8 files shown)


application/controllers/QuestionAdministrationController.php
```diff
@@ -2853,24 +2879,13 @@ function ($collector, $oDefaultValue) use ($oQuestion) {
      */
     private function storeSubquestions($question, $subquestionsArray)
     {
+        $this->validateSubquestionCodes($subquestionsArray);
```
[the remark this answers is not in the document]
kevin-foster-uk: If this validation fails it would return a duplicate code error modal. Not simply a 500.

assets/scripts/admin/questionEditor.js
```diff
@@ -1906,12 +1928,13 @@ $(document).on('ready pjax:scriptcomplete', function () {
 
             if (textStatus === 'success') {
               // Show confirm message.
-              LS.LsGlobalNotifier.createAlert(json.message, 'success', {showCloseButton: true});
+              LS.LsGlobalNotifier.createAlert(json.message + 'test', 'success', {showCloseButton: true});
```
twilligls, line 1931: remove the "test"

application/controllers/QuestionAdministrationController.php
```diff
@@ -263,6 +264,32 @@ public function renderFormAux(Question $question)
...
 
+    public function actionAjaxLoadExtraOptions($questionId)
+    {
+        $questionId = (int) $questionId;
+        $question = Question::model()->findByPk($questionId);
+        if (empty($question)) {
+            throw new CHttpException(404, gT('Invalid question id'));
+        }
+
+        if (!Permission::model()->hasSurveyPermission($question->sid, 'surveycontent', 'update')) {
```
Shnoulle, line 275: `'surveycontent', 'read'` ? We don't save here.
In 3.X : if i don't make error : an user with read can see all form, but not submit.

application/models/Question.php
```diff
@@ -189,7 +189,8 @@ public function rules()
                         ':scale_id' => $this->scale_id
                         )
                     ),
-                    'message' => gT('Subquestion codes must be unique.')
+                    'message' => gT('Subquestion codes must be unique.'),
+                    'except' => 'save-set'
```
Shnoulle, line 193: save-multiple or save-all ?

I mean : a better understandable name

application/models/Question.php
```diff
@@ -266,7 +267,7 @@ public function rules()
                 'except' => 'archiveimport'
             );
             $aRules[] = array(
-                'title', 'match', 'pattern' => '/^[[:alnum:]]*$/',
+                'title', 'match', 'pattern' => '/^[a-zA-z0-9]*$/',
```
Shnoulle, line 270: Argl …

tests/functional/backend/SaveQuestionAttributesTest.php
```diff
@@ -125,7 +125,7 @@ public function testSaveQuestionAttributes()
             $savebutton = $web->findElement(WebDriverBy::id('save-and-close-button-create-question'));
             $savebutton->click();
 
-            $alert = $this->waitForElementShim($web, '#notif-container .alert');
+            $alert = $this->waitForElementShim($web, '#notif-container .alert', 20);
```
Shnoulle, line 128: A timeout of 20 😩

PR #3499, opened 2023-09-28
Fixed issue #19136: Undefined variable in admin global settings gener…

mfavetti:
Fixed issue #19136: Undefined variable in admin global settings general subview
[https://bugs.limesurvey.org/view.php?id=19136](https://bugs.limesurvey.org/view.php?id=19136)

5.x not affected


Shnoulle (approved):
:+1:

PR #3503, opened 2023-10-02
Fixed issue #19112: Store XSS at Label sets list

Shnoulle:
Dev: escape value … always escape value in text input …

### DIFF
--- a/application/views/admin/labels/labelRow.twig
+++ b/application/views/admin/labels/labelRow.twig
@@ -31,7 +31,7 @@
     {% endif %}
 
     <td>
-        <input type='text' class=" form-control  " name='title_{{ language }}_{{ rowId }}' id='title_{{ language }}_{{ rowId }}' maxlength='3000' size='80' value="{{ title }}" />
+        <input type='text' class=" form-control  " name='title_{{ language }}_{{ rowId }}' id='title_{{ language }}_{{ rowId }}' maxlength='3000' size='80' value="{{ title|escape }}" />
     </td>
     <td>
         <div class="icon-btn-row">


gabrieljenik (approved):
Looks Ok. I guess HTML is not needed on `title` field.

Shnoulle:
> Looks Ok. I guess HTML is not needed on `title` field.

? What do you mean ? It's an input

gabrieljenik:
> > Looks Ok. I guess HTML is not needed on `title` field.
> 
> ? What do you mean ? It's an input

I mean this input is not supposed to allow html, right?

Shnoulle:
> > > Looks Ok. I guess HTML is not needed on `title` field.
> > 
> > 
> > ? What do you mean ? It's an input
> 
> I mean this input is not supposed to allow html, right?

Why ? See answers for example, or subquestions.

And here : label value can be used for answers or subquestions : then need to allow HTML …

gabrieljenik:
> > > > Looks Ok. I guess HTML is not needed on `title` field.
> > > 
> > > 
> > > ? What do you mean ? It's an input
> > 
> > 
> > I mean this input is not supposed to allow html, right?
> 
> Why ? it's answers for example, or subquestions.

If this are supposed to allow HTML, then:
1. Enter: `Me & You`
2. Will be saved as `Me &amp; You`
3. After refresh the input will have the value `Me &amp;amp; You`

Am I Right?

Shnoulle:
> Am I Right?

No …  it's principle of escaping … we always encode value … **always** … 

Peek 02-10-2023 16-14

gabrieljenik (approved):

PR #3505, opened 2023-10-03
Fixed issue [security] #19146: Store XSS when editing label set

Shnoulle:
Dev: encode in dropdown
Dev: no need HTML in label_name : filtering

### DIFF
--- a/application/models/LabelSet.php
+++ b/application/models/LabelSet.php
@@ -289,7 +289,7 @@ class LabelSet extends LSActiveRecord implements PermissionInterface
     }
 
     /**
-     * Sanitize string for any attribute
+     * Sanitize string for any attribute, XSS and XSS in javascript function too.
      * @todo create a Validator to be used for all such element : no need HTML, informative input used only for admin purpose.
      * @param string $attribute to sanitize
      * @return string sanitized attribute
--- a/application/models/LabelSet.php
+++ b/application/models/LabelSet.php
@@ -296,6 +296,6 @@ class LabelSet extends LSActiveRecord implements PermissionInterface
      */
     public static function sanitizeAttribute($attribute)
     {
-        return str_replace( ['<','>','&','\'','"'], "", $attribute);
+        return str_replace(['<','>','&','\'','"'], "", $attribute);
     }
 }
--- a/application/models/LabelSet.php
+++ b/application/models/LabelSet.php
@@ -290,6 +290,7 @@ class LabelSet extends LSActiveRecord implements PermissionInterface
 
     /**
      * Sanitize string for any attribute
+     * @todo create a Validator to be used for all such element : no need HTML, informative input used only for admin purpose.
      * @param string $attribute to sanitize
      * @return string sanitized attribute
      */
--- a/application/models/LabelSet.php
+++ b/application/models/LabelSet.php
@@ -295,6 +295,6 @@ class LabelSet extends LSActiveRecord implements PermissionInterface
      */
     public static function sanitizeAttribute($attribute)
     {
-        return filter_var($attribute, @FILTER_SANITIZE_STRING);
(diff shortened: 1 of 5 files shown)


application/models/LabelSet.php
```diff
@@ -286,4 +287,14 @@ public function hasPermission($sPermission, $sCRUD = 'read', $iUserID = null)
...
     }
+
+    /**
+     * Sanitize string for any attribute
+     * @param string $attribute to sanitize
+     * @return string sanitized attribute
+     */
+    public static function sanitizeAttribute($attribute)
+    {
+        return filter_var($attribute, @FILTER_SANITIZE_STRING);
```
gabrieljenik, line 298: What's the `@` for in front of the constant? Never used it like that... 

Also, do we have any other way of sanitizing on ther models? I am realziing, `FILTER_SANITIZE_STRING` is deprectaed. So if using something else somwehere, better use that. right?

Shnoulle: > Also, do we have any other way of sanitizing on ther models?

No, else i use it. Near copy/paste from Token here.

Shnoulle: > FILTER_SANITIZE_STRING` is deprectaed.

If i understand right : it's FILTER_FLAG_NO_ENCODE_QUOTES deprecated.

> Encoding quotes can be disabled by setting FILTER_FLAG_NO_ENCODE_QUOTES. (Deprecated as of PHP 8.1.0, use [htmlspecialchars()](https://www.php.net/manual/en/function.htmlspecialchars.php) instead.)

gabrieljenik (approved):
Code looks OK

application/models/LabelSet.php
gabrieljenik: Look at the changelog section here.
https://www.php.net/manual/en/filter.filters.sanitize.php
image

> If i understand right : it's FILTER_FLAG_NO_ENCODE_QUOTES deprecated.

>> Encoding quotes can be disabled by setting FILTER_FLAG_NO_ENCODE_QUOTES. (Deprecated as of PHP 8.1.0, use [htmlspecialchars()](https://www.php.net/manual/en/function.htmlspecialchars.php) instead.)

It is a bit confusing, but in changelog is clear.

gabrieljenik (changes requested):
Let's keep on reviewing about the sanitize... I am not too sure about inclding something which we know is already deprecated.

gabrieljenik:
> Code looks OK

Sorry, reviewed again.

application/models/LabelSet.php
Shnoulle: A good old preg_replace ?

Shnoulle: Or a `FILTER_SANITIZE_SPECIAL_CHARS` ?

Shnoulle: strip_tags ?

gabrieljenik: hmm whatever is already used in other models for similar situations... :) sanitize _helper?

gabrieljenik (approved):
Code looks OK

application/models/LabelSet.php
Shnoulle: > hmm whatever is already used in other models for similar situations... :) sanitize _helper?

There no global santizer validator, only xss, but here : i jyust don't want (care) of HTML, and needed for all user (admin included)

Found only 2 place

- https://github.com/LimeSurvey/LimeSurvey/blob/d01845f9895f4ada9abe31e0b2a5b772468babc5/application/models/Token.php#L322
- https://github.com/LimeSurvey/LimeSurvey/blob/d01845f9895f4ada9abe31e0b2a5b772468babc5/application/models/Token.php#L311

Shnoulle:
All green

PR #3508, opened 2023-10-04
Fixed issue [security] #19146: Store XSS when editing label set

Shnoulle:
Dev: sanitize when save


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/3505

gabrieljenik (approved):

PR #3509, opened 2023-10-04
 Fixed issue [security] #19134: User with "Users" permission only can add arbitrary users to the not owned user group

Shnoulle:
Dev: protect the action in same way than UserGroupController
 Dev: filter usergroups dropdown in same way

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -1114,18 +1114,29 @@ class UserManagementController extends LSBaseController
 
         if ($iUserGroupId) {
             $oUserGroup = UserGroup::model()->findByPk($iUserGroupId);
-            $aResults = [];
-
-            foreach ($aItems as $sItem) {
-                $aResults[$sItem]['title'] = '';
-                $model = $this->loadModel($sItem);
-                $aResults[$sItem]['title'] = $model->users_name;
-                if (!$oUserGroup->hasUser($sItem)) {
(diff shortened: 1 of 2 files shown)


Shnoulle:
Minimal fix : all of this need more detail …

Here : for example : UserGroup/create : no way to create group

gabrieljenik (approved):
Code looks OK

application/controllers/UserManagementController.php
```diff
@@ -1114,18 +1114,29 @@ public function actionBatchAddGroup()
...
-                $aResults[$sItem]['title'] = $model->users_name;
-                if (!$oUserGroup->hasUser($sItem)) {
-                    $aResults[$sItem]['result'] = $oUserGroup->addUser($sItem);
-                } else {
-                    $aResults[$sItem]['result'] = false;
-                    $aResults[$sItem]['error'] = gT('User is already a member of the group.');
+            /* check if have permission */
+            if (
+                Permission::model()->hasGlobalPermission('usergroups', 'update') /* Global update permission @see UserGroupController->actionEdit */
+                || $oUserGroup->requestEditGroup($oUserGroup->ugid, Yii::app()->session['loginID'])  /* This user group permission */
```
gabrieljenik, line 1120: Can we use getCurrentUserId()?

Shnoulle: Now, yes, not before.

PR #3512, opened 2023-10-05

gabrieljenik:


application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -90,6 +94,19 @@ public function update($surveyId, $input)
             throw new NotFoundException();
         }
 
+        // Before setting the owner, check if the user exists and can be seen
+        // by the current user (in case the request was forged)
+        // NOTE: Internally, the getUserList function will use objects (like the Yii App and Permission model) that
+        //       currently may differ from the ones injected in this service.
+        if (!empty($input['owner_id'])) {
+            $owner = $this->modelUser->findByPk($input['owner_id']);
+            if (!isset($owner) || !in_array($input['owner_id'], getUserList('onlyuidarray'))) {
```
Shnoulle, line 103: On Service ? Not in controller or model ?

gabrieljenik: Is a new pattern being implemented for placing all LS logic outside of framework, on own files.
This is the new "controller"-service on master. That's why we are adding it here.

Shnoulle: OK :) 

Some other old function need to be move in Service then (DEV issue not needed)

application/models/services/SurveyUpdater/GeneralSettings.php
```diff
@@ -90,6 +94,19 @@ public function update($surveyId, $input)
...
+                throw new PermissionDeniedException(
+                    'Permission denied'
```
Shnoulle, line 105: Desired behavior : throw error, filter or send a model error ?

Shnoulle (commented):
What is the desired behavior ?

The only way to set a invalid userId is to hack HTML or $_POST value ?

Shnoulle:
Desired behavior : user try to hack server. 

Let's sysadmin check error log for such action. 

A 403 is the best !

application/models/services/SurveyUpdater/GeneralSettings.php
gabrieljenik: I guess that refactoring will slowly be happening with time. :)

gabrieljenik:
Label "code review done" was added but approval is pending... :)

Shnoulle (approved):

PR #3520, opened 2023-10-09
Fixed issue #19094: Input Validation Vulnerability Leading to 500 on responses page

Shnoulle:
Dev: sanitize when upload, check when view
Dev: check when get it for export

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -1104,8 +1104,13 @@ function getExtendedAnswer($iSurveyID, $sFieldCode, $sValue, $sLanguage)
                             if (!isset($file['comment'])) {
                                 $file['comment'] = '';
                             }
+                            $size = "";
+                            if($file['size'] && strval(floatval($file['size'])) == strval($file['size'])) {
+                                // avoid to throw PHP error if size is invalid
+                                $size = sprintf('%s KB', round($file['size']));
+                            }
                             $sValue .= rawurldecode((string) $file['name']) .
-                            ' (' . round($file['size']) . 'KB) ' .
+                            ' (' . $size .' ) ' .
                             strip_tags((string) $file['title']);
                             if (trim(strip_tags((string) $file['comment'])) != "") {
                                 $sValue .= ' - ' . strip_tags((string) $file['comment']);
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -8512,6 +8512,8 @@ report~numKids > 0~message~{name}, you said you are {age} and that you have {num
                                             }
                                             $aFiles[$i]->filename = $sDestinationFileName;
                                         }
+                                        /* Sanitize size */
+                                        $aFiles[$i]->size = floatval($aFiles[$i]->size);
                                     }
                                     $value = ls_json_encode($aFiles);  // so that EM doesn't try to parse it.
                                 }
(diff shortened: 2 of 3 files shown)


gabrieljenik (approved):
Code looks OK

application/helpers/expressions/em_manager_helper.php
```diff
@@ -8512,6 +8512,8 @@ public static function ProcessCurrentResponses()
                                             }
                                             $aFiles[$i]->filename = $sDestinationFileName;
                                         }
+                                        /* Sanitize size */
+                                        $aFiles[$i]->size = floatval($aFiles[$i]->size);
```
gabrieljenik, line 8516: Use `sanitize_float` from `sanitize_helper`?

application/helpers/common_helper.php
```diff
@@ -1104,8 +1104,13 @@ function getExtendedAnswer($iSurveyID, $sFieldCode, $sValue, $sLanguage)
                             if (!isset($file['comment'])) {
                                 $file['comment'] = '';
                             }
+                            $size = "";
+                            if($file['size'] && strval(floatval($file['size'])) == strval($file['size'])) {
```
gabrieljenik, line 1108: What is the logic behind this?
` strval(floatval($file['size'])) == strval($file['size'])`

Shnoulle: It's not a float : it's a string currently, then : just to be sure to have a string «like» a float.

Avoit AAA , but avoid 12.1a too : it's still invalid.

application/helpers/expressions/em_manager_helper.php
Shnoulle: 1. I really don't want to allow 1,2 : it must be 1.2, real float.
2. It must be float for round, not GMP number 

https://github.com/LimeSurvey/LimeSurvey/blob/b95bf50c0dd76788d906978de4f589fd86c75ef2/application/helpers/sanitize_helper.php#L341

application/helpers/common_helper.php
gabrieljenik: I understand. Still, disagree a bit.
On line 1110 you do `round($file['size'])`, so after all, you are expecting a float.
Why not just converting to float and then sanitizing?

This is just a feedback comment. It is not worth going back and forth.

gabrieljenik (approved):
Code looks OK

application/helpers/common_helper.php
Shnoulle: >  Why not just converting to float and then sanitizing?

1. Previous survey can still have invalid value. 
2. VV import can have invalid value
3. LSA import can have invalid value

New value have `strval(floatval($file['size'])) == strval($file['size'])` OK,  Except for 0 (when floatval("A") for example.

application/helpers/expressions/em_manager_helper.php
Shnoulle, line 8516: @gabrieljenik : maybe we can check the size of the fiilenamle (fu_XXX) at this point and don't use user submitted value ?

Advice ?

(only in master)

Shnoulle (commented):
A advice :)

application/helpers/expressions/em_manager_helper.php
gabrieljenik: That would be good too. Always better to have clean data on the DB. Still, not sure if the change is worth to be done now.

Shnoulle: > Also, another possibility: clean the data when submitted?

I don't understand ? What other data can be cleaned ?

> Still, not sure if the change is worth to be done now.

Maybe a dev/feature mantis

gabrieljenik: > I don't understand ? What other data can be cleaned ?

This size data. But nevermind. I don't thin it worths it.

Shnoulle: We are in ProcessCurrentResponses : it's **THE** place to clean data before all other process.

The other place it's clean data before save to DB https://github.com/LimeSurvey/LimeSurvey/blob/90197188821c602a5d5dfc2772bbf552f3e7aa01/application/helpers/expressions/em_manager_helper.php#L9739

With checkValidityAnswer : you can add error to the respondant.

PR #3522, opened 2023-10-09
Fixed issue #18776: Question ordering not correct

mfavetti:
Fixed issue #18776: Question ordering not correct

[https://bugs.limesurvey.org/view.php?id=18776](https://bugs.limesurvey.org/view.php?id=18776)


Shnoulle (approved):
👍

PR #3524, opened 2023-10-10
Fixed issue #19117: Account past their expiration date can be still active

Shnoulle:
Dev: add active scope to User
Dev: use App->getCurrentuserId in App()->user->id (used in authorisation)
Dev: logout invalid user

### DIFF
--- a/application/core/Traits/LSApplicationTrait.php
+++ b/application/core/Traits/LSApplicationTrait.php
@@ -8,7 +8,6 @@
 
 trait LSApplicationTrait
 {
-
     /* @var integer| null the current userId for all action */
     private $currentUserId;
     /**
@@ -18,7 +17,7 @@ trait LSApplicationTrait
      */
     public function getCurrentUserId()
     {
-        if(empty(App()->session['loginID'])) {
+        if (empty(App()->session['loginID'])) {
             /**
              * NULL for guest,
              * null by default for CConsoleapplication, but Permission always return true for console
@@ -31,7 +30,7 @@ trait LSApplicationTrait
         }
         /* use App()->session and not App()->user fot easiest unit test */
         $this->currentUserId = App()->session['loginID'];
-        if ($this->currentUserId && !User::model()->findByPk($this->currentUserId)) {
+        if ($this->currentUserId && !User::model()->active()->findByPk($this->currentUserId)) {
             $this->currentUserId = 0;
         }
         return $this->currentUserId;
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -937,6 +937,19 @@ class User extends LSActiveRecord
         ));
     }
 
+    /** @inheritdoc */
+    public function scopes()
+    {
+        return array(
+            'active' => array(
+                'condition' => "expires > :now OR expires IS NULL",
+                'params' => array(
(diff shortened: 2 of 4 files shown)


application/core/Traits/LSApplicationTrait.php
```diff
@@ -31,7 +30,7 @@ public function getCurrentUserId()
         }
         /* use App()->session and not App()->user fot easiest unit test */
         $this->currentUserId = App()->session['loginID'];
-        if ($this->currentUserId && !User::model()->findByPk($this->currentUserId)) {
+        if ($this->currentUserId && !User::model()->active()->findByPk($this->currentUserId)) {
```
gabrieljenik, line 33: Cool

gabrieljenik: `$this->currentUserId && ` --> Wouldn't be good to move this to an independent if?
Something like 
```
if (!$this->currentUserId || empty($this->currentUserId)) {
  return 0
}
```

That would make code a bit less complex

gabrieljenik (approved):

application/core/Traits/LSApplicationTrait.php
Shnoulle: Unsure : more clear ?

```
if (empty($this->currentUserId)) {
  return $this->currentUserId;
}
```

before ?

Shnoulle: MASTER only ;)

gabrieljenik: I think would be better, yes

Shnoulle: Sorry didn't see your answer before merge …

application/models/User.php
```diff
@@ -937,6 +937,19 @@ public function search()
         ));
     }
 
+    /** @inheritdoc */
+    public function scopes()
+    {
+        return array(
+            'active' => array(
+                'condition' => "expires > :now OR expires IS NULL",
```
olleharstedt, line 945: Yo @Shnoulle , this break upgrade from LS 3 to 5, since the `expires` column does not exists when the upgrade starts.

olleharstedt: The upgrade procedure must not use any ActiveRecord models. They cache their columns etc, it's not reliable.

Shnoulle: OK, then broke before upgrade happen.

DB version dependant scope ? OK for this fix ?

olleharstedt: @ptelu can discuss

Shnoulle: I don't see how do it except a version dependant scope … i create a PR for master … OK ?

Shnoulle: > Maybe you can add the scope dependant on the version?

?

Shnoulle: @olleharstedt  https://github.com/LimeSurvey/LimeSurvey/pull/3607

I test an upgrade connected and not connected from 3.X : seems OK.

Shnoulle: > Maybe you can add the scope dependant on the version?

Oh, got it :) Yes an alternative.  I think scope dependant allow to use it on other code in future :)

PR #3525, opened 2023-10-10
Fixed issue #19117: Account past their expiration date can be still active

Shnoulle:
Dev: add active scope to User
Dev: use App->getCurrentuserId in App()->user->id (used in authorisation)
Dev: logout invalid user

### DIFF
--- a/application/core/LSWebUser.php
+++ b/application/core/LSWebUser.php
@@ -23,8 +23,8 @@ class LSWebUser extends CWebUser
             return parent::getId();
         }
         $id = App()->getCurrentUserId();
-        if (empty($id)) {
-            /* If still connected but invalid : logout */
+        if ($id === 0) {
+            /* User is still connected but invalid : logout */
             $this->logout();
         }
         return $id;
--- a/application/core/LSWebUser.php
+++ b/application/core/LSWebUser.php
@@ -14,6 +14,39 @@ class LSWebUser extends CWebUser
     }
 
     /**
+     * @inheritDoc
+     * Replace auto getter to check if current user is valid or not
+     */
+    public function getId()
+    {
+        if (empty(parent::getId())) {
+            return parent::getId();
+        }
+        $id = App()->getCurrentUserId();
+        if (empty($id)) {
+            /* If still connected but invalid : logout */
+            $this->logout();
+        }
+        return $id;
+    }
+
+    /**
+     * @inheritDoc
+     * Add the specific plugin event and regenerate CSRF
+     */
+    public function logout($destroySession = true)
+    {
+        /* Adding beforeLogout event */
+        $beforeLogout = new PluginEvent('beforeLogout');
+        App()->getPluginManager()->dispatchEvent($beforeLogout);
+        regenerateCSRFToken();
+        parent::logout($destroySession);
+        /* Adding afterLogout event */
+        $event = new PluginEvent('afterLogout');
+        App()->getPluginManager()->dispatchEvent($event);
+    }
+
+    /**
      * @inheritdoc
      */
     public function checkAccess($operation, $params = array(), $allowCaching = true)
--- a/application/core/LSWebUser.php
+++ b/application/core/LSWebUser.php
@@ -32,6 +32,16 @@ class LSWebUser extends CWebUser
 
     /**
      * @inheritDoc
+     * Set id in session too
+     */
+    public function setId($id)
+    {
+        parent::setId($id);
+        \Yii::app()->session['loginID'] = $id;
+    }
+
+    /**
+     * @inheritDoc
      * Add the specific plugin event and regenerate CSRF
      */
     public function logout($destroySession = true)
--- a/application/core/Traits/LSApplicationTrait.php
+++ b/application/core/Traits/LSApplicationTrait.php
@@ -8,7 +8,6 @@
 
 trait LSApplicationTrait
 {
-
     /* @var integer| null the current userId for all action */
     private $currentUserId;
     /**
@@ -18,7 +17,7 @@ trait LSApplicationTrait
      */
     public function getCurrentUserId()
     {
-        if(empty(App()->session['loginID'])) {
+        if (empty(App()->session['loginID'])) {
             /**
              * NULL for guest,
              * null by default for CConsoleapplication, but Permission always return true for console
@@ -31,7 +30,7 @@ trait LSApplicationTrait
         }
         /* use App()->session and not App()->user fot easiest unit test */
         $this->currentUserId = App()->session['loginID'];
-        if ($this->currentUserId && !User::model()->findByPk($this->currentUserId)) {
+        if ($this->currentUserId && !User::model()->active()->findByPk($this->currentUserId)) {
             $this->currentUserId = 0;
         }
         return $this->currentUserId;
--- a/application/core/Traits/LSApplicationTrait.php
+++ b/application/core/Traits/LSApplicationTrait.php
@@ -13,7 +13,7 @@ trait LSApplicationTrait
     /**
      * get the current id of connected user,
      * check if user exist before return for security
-     * @return int|null user id
+     * @return int|null user id, 0 mean invalid user
      */
     public function getCurrentUserId()
     {
--- a/application/models/User.php
+++ b/application/models/User.php
@@ -127,6 +127,19 @@ class User extends LSActiveRecord
         );
     }
 
+    /** @inheritdoc */
+    public function scopes()
+    {
+        return array(
+            'active' => array(
+                'condition' => "expires > :now OR expires IS NULL",
+                'params' => array(
+                    'now' => dateShift(date("Y-m-d H:i:s"), "Y-m-d H:i:s", Yii::app()->getConfig("timeadjust")),
+                )
+            )
+        );
+    }
+
     public function attributeLabels()
     {
         return [
--- a/application/controllers/admin/Authentication.php
+++ b/application/controllers/admin/Authentication.php
@@ -270,17 +270,8 @@ class Authentication extends SurveyCommonAction
      */
     public function logout()
     {
-        /* Adding beforeLogout event */
-        $beforeLogout = new PluginEvent('beforeLogout');
-        App()->getPluginManager()->dispatchEvent($beforeLogout);
-        regenerateCSRFToken();
         App()->user->logout();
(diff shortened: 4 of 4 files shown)


Shnoulle:
WTF ?

```
shnoulle@Belgarion:/media/shnoulle/data/webdev/lstest$ git branch
* (HEAD détachée sur origin/bug/19117_master_invalidUser_date)
  bug/19107_superadminupdate_review
  master
shnoulle@Belgarion:/media/shnoulle/data/webdev/lstest$ DOMAIN=lstest.local ./vendor/bin/phpunit ./tests/unit/controllers/UserManagementTest.php
Error Reporting and Debug: No
Set $forceDebug=true in tests/bootstrap.php to enable more logging.

PHPUnit 8.5.34 by Sebastian Bergmann and contributors.

Error:         This version of PHPUnit does not support code coverage on PHP 8

...                                                                 3 / 3 (100%)

Time: 148 ms, Memory: 18.00 MB

OK (3 tests, 3 assertions)
```

application/core/LSWebUser.php
```diff
@@ -13,6 +13,49 @@ public function __construct()
...
+     * Replace auto getter to check if current user is valid or not
+     */
+    public function getId()
+    {
+        if (empty(parent::getId())) {
+            return parent::getId();
+        }
+        $id = App()->getCurrentUserId();
+        if (empty($id)) {
+            /* If still connected but invalid : logout */
```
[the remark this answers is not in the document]
gabrieljenik: Not sure if I am right. Sorry I didn't notice it on the other branch

application/models/User.php
```diff
@@ -127,6 +127,19 @@ public function rules()
...
     }
 
+    /** @inheritdoc */
+    public function scopes()
+    {
+        return array(
+            'active' => array(
+                'condition' => "expires > :now OR expires IS NULL",
+                'params' => array(
+                    'now' => dateShift(date("Y-m-d H:i:s"), "Y-m-d H:i:s", Yii::app()->getConfig("timeadjust")),
```
gabrieljenik, line 137: Isn't there a common function to do this? If not, maybe we could add it and that wuld be the start of using it.
Disregard if you believe that is out of the scope.

Shnoulle: `dateShift` is the common function, no ?

application/core/LSWebUser.php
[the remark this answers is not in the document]
Shnoulle: And here : `App()->getCurrentUserId()` can return 0 without user loggued out or session time. All session value still there.

application/models/User.php
gabrieljenik: I meant something like `timeAdjustedDate()`.

application/core/LSWebUser.php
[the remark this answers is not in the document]
gabrieljenik: Yes, I meant, when logged out, woudl this code be executed? If logged out, `parent::getId()` wouldn't it be empty?
But if you already consider it, just disregard it

application/models/User.php
Shnoulle:
```
function configTimeAdjustedDate($date = null, $format = "Y-m-d H:i:s") {
    if (empty($date)) {
        $date = date("Y-m-d H:i:s");
    }
    return dateShift($date, $format, Yii::app()->getConfig("timeadjust")),
}
```

Really ?

Where ?

application/core/LSWebUser.php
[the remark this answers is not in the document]
Shnoulle: If already logedout : you can logout a second time without issue, but yes : check if can be avoided …

application/models/User.php
gabrieljenik: I was saying to create that. Then, in other time, we can spread it

application/core/LSWebUser.php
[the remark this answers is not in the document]
Shnoulle: Maybe replace https://github.com/LimeSurvey/LimeSurvey/blob/03658b26a7b87e4e51cbb432ceeea3517081e74b/vendor/yiisoft/yii/framework/web/auth/CWebUser.php#L288 too ?

application/models/User.php
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/3525#discussion_r1358315823

Shnoulle: My opinion  it's not needed

gabrieljenik: > My opinion it's not needed

What do you mean?
If you feel creating that function is too much, just don't. It was just a suggestion

Shnoulle: I think dateShift do the job.

application/core/LSWebUser.php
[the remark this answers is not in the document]
Shnoulle: >  Can you remind me the whole purpose of this method?

User with date still be connected. We must release session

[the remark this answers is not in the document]
Shnoulle: expire user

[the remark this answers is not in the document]
Shnoulle: Really better like this : thanks

gabrieljenik (approved):
Code structure looks OK.
If there is a plugin that use login / logout event could be good to test it with that as well.

Shnoulle:
>  If there is a plugin that use login / logout event could be good to test it with that as well.

I don't know if exist.

But since i move logout to somewhere out of controller : it seems logic to move it to User.

application/core/Traits/LSApplicationTrait.php
```diff
@@ -8,17 +8,16 @@
...
     /**
      * get the current id of connected user,
      * check if user exist before return for security
-     * @return int|null user id
+     * @return int|null user id, 0 mean invalid user
      */
     public function getCurrentUserId()
     {
-        if(empty(App()->session['loginID'])) {
+        if (empty(App()->session['loginID'])) {
```
Shnoulle, line 20: @gabrieljenik empty or is_null here ?

gabrieljenik:
> > If there is a plugin that use login / logout event could be good to test it with that as well.
> 
> I don't know if exist.
> 
> But since i move logout to somewhere out of controller : it seems logic to move it to User.

Yes. That is good. 
Also, in case you think usefull, we could think about using native Yii events (https://www.yiiframework.com/doc/api/2.0/yii-web-user#beforeLogin()-detail)

application/core/LSWebUser.php
```diff
@@ -13,6 +13,49 @@ public function __construct()
...
+     * @inheritDoc
+     * Replace auto getter to check if current user is valid or not
+     */
+    public function getId()
+    {
+        if (empty(parent::getId())) {
+            return parent::getId();
+        }
+        $id = App()->getCurrentUserId();
+        if ($id === 0) {
```
gabrieljenik, line 26: Why not `if (empty)` here ?

application/core/Traits/LSApplicationTrait.php
gabrieljenik: I would say empty better. But why the doubt? Which edge case are you thinking about?

application/core/LSWebUser.php
Shnoulle: It's the point : function https://github.com/LimeSurvey/LimeSurvey/pull/3525/files#diff-ec864d1f2c65a2e53bd6cc84323c092139025e28b43b107af90ea46f2a5e3ed5R18 check session.

If sessio is set : check user : if user is invalid : return 0. User was still connected but are disable or deleted.

application/core/Traits/LSApplicationTrait.php
Shnoulle: I really don't know who wan use `App()->session['loginID'] = 0` or `App()->session['loginID'] = ''` ?

We currently use loginID in all test and set it when login. But it's old version, when we don't use Yii …

Shnoulle:
> Also, in case you think usefull, we could think about using native Yii events (https://www.yiiframework.com/doc/api/2.0/yii-web-user#beforeLogin()-detail)

Yea, but i stay near current version currently …

gabrieljenik:
> > Also, in case you think usefull, we could think about using native Yii events (https://www.yiiframework.com/doc/api/2.0/yii-web-user#beforeLogin()-detail)
> 
> Yea, but i stay near current version currently …

Sorry, I missed the version on the link. Anyway...

application/core/Traits/LSApplicationTrait.php
gabrieljenik: I like empty more.

application/core/LSWebUser.php
gabrieljenik: What if `App()->getCurrentUserId();` returns `null`?

application/core/Traits/LSApplicationTrait.php
Shnoulle: > I like empty more. Catches more situtations.

Yes, 
In all current code code : empty App()->session['loginID'] is always null. Then no issue. I just think of plugin or hacked version using 0 or ''
Here : potential issue about loop.

application/core/LSWebUser.php
Shnoulle: Logically : user is already log out.

See https://github.com/LimeSurvey/LimeSurvey/pull/3525/files#diff-ec864d1f2c65a2e53bd6cc84323c092139025e28b43b107af90ea46f2a5e3ed5R16

> @return int|null user id, **0 mean invalid user**

```
        if ($this->currentUserId && !User::model()->active()->findByPk($this->currentUserId)) {
            $this->currentUserId = 0;
            $this->currentUserId = 0;
        }
        }
```

Maybe we can compare App()->session['loginID'] with [parent::getId]($this->getState('__id');) 

LimeSurvey core use sometime App()->session['loginID'] and sometime App()->user->id …

Maybe i need to extend https://github.com/LimeSurvey/LimeSurvey/blob/03658b26a7b87e4e51cbb432ceeea3517081e74b/vendor/yiisoft/yii/framework/web/auth/CWebUser.php#L288 too ?

gabrieljenik: > What if App()->getCurrentUserId(); returns null?

I mean that maybe we should use empty as it will allow to catch more situations and flexibility.

> Maybe i need to extend ...

Hmm, not sure. Got a bit lost there.
If you need me to join you on the analysis, let's just make a call or something...

PR #3532, opened 2023-10-11
'Fixed issue #19160 - subquestion save, scale id code

kevin-foster-uk:
Partial fix of subquestion uniqueness handling. This fixes the issue of using the same sub-id code within on a different scale. 

This PR does not fix the issue of re-naming a subquestion code to one that already existed. I think this is an edge case and I don't think its worth the effort of fixing this in 5.x.

### DIFF
--- a/application/controllers/QuestionAdministrationController.php
+++ b/application/controllers/QuestionAdministrationController.php
@@ -2772,6 +2772,7 @@ class QuestionAdministrationController extends LSBaseController
                     $subquestion = Question::model()->findByAttributes([
                         'sid' => $question->sid,
                         'parent_qid' => $question->qid,
+                        'scale_id' => $scaleId,
                         'title' => $data['oldcode']
                     ]);
                 }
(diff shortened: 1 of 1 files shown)


application/controllers/QuestionAdministrationController.php
```diff
@@ -2772,6 +2772,7 @@ private function storeSubquestions($question, $subquestionsArray)
                     $subquestion = Question::model()->findByAttributes([
                         'sid' => $question->sid,
                         'parent_qid' => $question->qid,
+                        'scale_id' => $scaleId,
```
gabrieljenik, line 2775: Shouldn't you add scale id also around here: https://github.com/LimeSurvey/LimeSurvey/pull/3532/files#diff-107f8c81519e24601f4b3670bb92ccb4daa2c444134d1c14b24cce40b1e4d722R2786 ?

kevin-foster-uk: Its already there just further down (line #2797).

gabrieljenik: Sorry. My bad.

gabrieljenik (approved):
Code looks OK!

PR #3533, opened 2023-10-11
Dev: Add test

Shnoulle:


tests/unit/models/QuestionTest.php
```diff
@@ -44,4 +44,94 @@ public function testGetQuestionThemeOnSavedQuestion()
...
+    /**
+     * Test getting a default theme name from an invalid type.
+     */
+    public function testGetDefaultQuestionThemeNameFromInvalidType()
+    {
+        $question = new Question();
+        $question->type = 'Test';
+
+        $themeName = $question->questionThemeNameValidator();
+        $this->assertNull($themeName, 'An invalid validation result was expected since "Z" question type does not exist.');
```
Shnoulle, line 94: 'An invalid validation result was expected since "Test" question type does not exist.'

tests/unit/models/QuestionTest.php
```diff
@@ -44,4 +44,94 @@ public function testGetQuestionThemeOnSavedQuestion()
...
+     * Test getting the core theme name.
+     */
+    public function testGetCoreQuestionThemeName()
+    {
+        $question = new Question();
+        $question->type = "M";
+        $question->question_theme_name = 'core';
+
+        $themeName = $question->questionThemeNameValidator();
+        $this->assertSame('multiplechoice', $themeName, 'Incorrect core question theme name.');
```
Shnoulle, line 107: It's not core now ? I don't rememeber  …

Shnoulle:
@gabrieljenik still have ` Failed asserting that null is identical to 'longfreetext'.` unsure on the needed value currently … maybe it was updated since i do the PR

tests/unit/models/QuestionTest.php
gabrieljenik: I think so. I mean, it is not failing. Let'wait and see. I don't recall either.

Shnoulle: There are update on QuestionTheme system since i create the PR …

@ptelu you work on the new QuetsionTheme system ?

Shnoulle: >  Maybe related to some refactoring that Kevin, Tim and Patricia are doing

Some month ago : core theme are set to `core` in DB. Unsure when it was updated too. But you wrok on it 5 years before … 
https://github.com/LimeSurvey/LimeSurvey/commit/7d977047c6b6773cf2e8e531c3299f79851a9a1d 😭

It's done via https://github.com/LimeSurvey/LimeSurvey/blob/b1146f3f31437c3b4891a74418041dc1441b76d8/application/core/LsDefaultDataSets.php#L1860

gabrieljenik: > shouldnt this be done by finding the core theme based on "core_theme" and "question_type" column in question_themes table?

Why? Don't we already know we expect "multiplechoice" for that combination of `type` and `uestion_theme_name`?

Shnoulle:
Original was merged into master
I give you the Pr

tests/unit/models/QuestionTest.php
Shnoulle: Yep :)

Some more DEV issue to report, unsure on desired behaviour

- disable core question theme and allow only user question theme
- disable core question theme without user question theme (i have it for upload question type)

gabrieljenik:
What about now @Shnoulle ?

tests/unit/models/QuestionTest.php
```diff
@@ -44,4 +44,84 @@ public function testGetQuestionThemeOnSavedQuestion()
...
+     * Expecting to use default theme name from type
+     */
+    public function testValidateQuestionThemeNameMismatch()
+    {
+        $question = new Question();
+        $question->type = "M";
+        $question->question_theme_name = 'longfreetext';
+
+        $themeName = $question->questionThemeNameValidator();
+        $this->assertSame('multiplechoice', $themeName, 'Question theme was not auto corrected and derived from the type\'s default theme.');
```
Shnoulle, line 125: 👍

Shnoulle:
> What about now @Shnoulle ?

Seems OK, if `core` work on import, survey etc … it's OK. 

We can update if we update the desired behavior.

All other seems really great (in my opinion).

Shnoulle (commented):
I can not Approve ;)

Shnoulle:
I wait for green an merge

PR #3534, opened 2023-10-11
Fixed issue #18776: Question ordering not correct

mfavetti:
Fixed issue #18776: Question ordering not correct

[https://bugs.limesurvey.org/view.php?id=18776](https://bugs.limesurvey.org/view.php?id=18776)

Same as #3522, but for 5.x branch


Shnoulle:
Thank you !

PR #3539, opened 2023-10-16
Qe 356 rest api json error handlers

kevin-foster-uk:
Yii is configured to display a standard error page in response to unhandled exceptions and fatal errors. When making REST API requests the client expects an JSON response so the standard HTML error page need to be replaced with a JSON response.


application/core/LSYii_Application.php
```diff
@@ -717,4 +717,93 @@ public function getPublicBaseUrl($absolute = false)
...
+
+        return $acceptsJson || $isRestUrl;
+    }
+
+	/**
+	 * Handles uncaught PHP exceptions.
+	 *
+	 * @param Exception $exception exception that is not caught
+	 */
+    public function handleException($exception)
```
olleharstedt, line 744: Hm isn't this configured in config.php? https://www.yiiframework.com/doc/guide/1.1/en/topics.error
Or you can't branch on that point, only on runtime?

application/core/LSYii_Application.php
```diff
@@ -717,4 +717,93 @@ public function getPublicBaseUrl($absolute = false)
         }
         return $baseUrl;
     }
+
+	/**
+	 * Is REST request
+	 *
+	 * @return boolean
+	 */
+    public function isRestRequest()
```
olleharstedt, line 726: Should this rather be a method of the request class? CHttpRequest? Dunno if it's subclassed already.

application/core/LSYii_Application.php
kevin-foster-uk: I noticed we have custom error handling code in the onException() method. I have moved the error handling code in to a separate class and added the extra functionality.

application/libraries/Yii/Application/ErrorHandler.php
```diff
@@ -0,0 +1,223 @@
...
+
+use Yii;
+use CWebApplication;
+use CHttpException;
+use LSUserException;
+
+/**
+ * Trait for Application error handling
+ */
+class ErrorHandler
```
olleharstedt, line 13: We should probably note somewhere that this class is NOT related to the CErrorHandler component built in in Yii, https://www.yiiframework.com/doc/api/1.1/CErrorHandler

kevin-foster-uk: I renamed the class to AppErrorHandler. I think this and the namespace makes it clear that this is not Yii code but LimeSurvey code relating to Yii.

olleharstedt (approved):

olleharstedt:
Approved, but unsure how to test?

PR #3540, opened 2023-10-16
Fixed issue #19173: encrypted file upload question - delete file results in crash

mfavetti:
Fixed issue #19173: encrypted file upload question - delete file results in crash

[https://bugs.limesurvey.org/view.php?id=19173](https://bugs.limesurvey.org/view.php?id=19173)

### DIFF
--- a/application/models/Response.php
+++ b/application/models/Response.php
@@ -109,14 +109,18 @@ abstract class Response extends Dynamic
             ->findAllByAttributes($aConditions);
         $files = array();
         foreach ($aQuestions as $question) {
+            $encrypted = $question->encrypted === 'Y';
             $field = $question->sid . 'X' . $question->gid . 'X' . $question->qid;
             $fieldDataJson = $this->getAttribute($field);
-            if ($question->encrypted === 'Y') {
+            if ($encrypted) {
                 $fieldDataJson = self::decryptSingle($fieldDataJson);
             }
             $fieldData = json_decode(stripslashes((string) $fieldDataJson), true);
             if (is_array($fieldData)) {
-                $files[$field] = $fieldData;
+                $files[$field] = array(
+                    'files' => $fieldData,
+                    'encrypted' => $encrypted,
+                );
             }
         }
         return $files;
@@ -179,8 +183,15 @@ abstract class Response extends Dynamic
         $success = 0;
         $uploaddir = Yii::app()->getConfig('uploaddir') . "/surveys/{$this->dynamicId}/files/";
         $filesData = $this->getFilesAndSqga();
-        foreach ($filesData as $sgqa => $fileInfos) {
-            foreach ($fileInfos as $i => $fileInfo) {
+
+        foreach ($filesData as $sgqa => $aQuestion) {
+
+            [
+                'files' => $files,
+                'encrypted' => $encrypted,
+            ] = $aQuestion;
+
+            foreach ($files as $i => $fileInfo) {
                 $basename = basename((string) $fileInfo['filename']);
                 $fullFilename = $uploaddir . $basename;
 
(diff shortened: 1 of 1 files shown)


application/models/Response.php
```diff
@@ -109,14 +109,20 @@ public function getFilesAndSqga($sQID = 0)
             ->findAllByAttributes($aConditions);
         $files = array();
         foreach ($aQuestions as $question) {
+
+            $aQuestion = array(
```
Shnoulle, line 113: Maybe more clear with `$encrypted = $question->encrypted === 'Y',`

application/models/Response.php
```diff
@@ -109,14 +109,20 @@ public function getFilesAndSqga($sQID = 0)
...
             $field = $question->sid . 'X' . $question->gid . 'X' . $question->qid;
             $fieldDataJson = $this->getAttribute($field);
-            if ($question->encrypted === 'Y') {
+            if ($aQuestion['encrypted']) {
                 $fieldDataJson = self::decryptSingle($fieldDataJson);
             }
             $fieldData = json_decode(stripslashes((string) $fieldDataJson), true);
             if (is_array($fieldData)) {
-                $files[$field] = $fieldData;
+                $aQuestion['files'] = $fieldData;
```
Shnoulle, line 124: And here : 
```

$files[$field] = [
    'file'= $fieldData,
    'encrypted' => $encrypted
];
```

Shnoulle (commented):
Code is OK,

but whole system seems really complex for objectives 

```
./application$ grep -r "getFilesAndSqga" *
models/Response.php:    public function getFilesAndSqga($sQID = 0)
models/Response.php:        $filesData = $this->getFilesAndSqga();
./application$ grep -r "deleteFilesAndFilename" *
controllers/ResponsesController.php:                [$success, $errors] = $response->deleteFilesAndFilename();
models/Response.php:    public function deleteFilesAndFilename()

```

mfavetti:
> Code is OK,
> 
> but whole system seems really complex for objectives
> 
> ```
> ./application$ grep -r "getFilesAndSqga" *
> models/Response.php:    public function getFilesAndSqga($sQID = 0)
> models/Response.php:        $filesData = $this->getFilesAndSqga();
> ./application$ grep -r "deleteFilesAndFilename" *
> controllers/ResponsesController.php:                [$success, $errors] = $response->deleteFilesAndFilename();
> models/Response.php:    public function deleteFilesAndFilename()
> ```

Can't really speak to existing implementation...

Do you mean that each function is only used in one place, so no need to abstract them? Like it's better to just use the function's code directly where called?

Shnoulle:
> > Code is OK,
> > but whole system seems really complex for objectives
> > ```
> > ./application$ grep -r "getFilesAndSqga" *
> > models/Response.php:    public function getFilesAndSqga($sQID = 0)
> > models/Response.php:        $filesData = $this->getFilesAndSqga();
> > ./application$ grep -r "deleteFilesAndFilename" *
> > controllers/ResponsesController.php:                [$success, $errors] = $response->deleteFilesAndFilename();
> > models/Response.php:    public function deleteFilesAndFilename()
> > ```
> 
> Can't really speak to existing implementation...
> 
> Do you mean that each function is only used in one place, so no need to abstract them? Like it's better to just use the function's code directly where called?

Really unsure, but when reviewing : i start to make some comment for code optimization but after reviewing more : i start to understand why you di it like this.

Here : i really don't understand the point to use `getFilesAndSqga`

Just doing a 

```
foreach ($aQuestions as $question) {
    $field = $question->sid . 'X' . $question->gid . 'X' . $question->qid;
    $fieldDataJson = $this->getAttribute($field);
    if ($question->encrypted === 'Y') {
        $fieldDataJson = self::decryptSingle($fieldDataJson);
    }
    $fieldData = json_decode(urldecode((string) $fieldDataJson), true);
    if (is_array($fieldData)) {
        // Do the delete action
    }
}
```


Seems more clear and clean.

Shnoulle:
PS : remind to activate phpcs on your code.

kevin-foster-uk:
Fixed in: https://github.com/LimeSurvey/LimeSurvey/pull/3564

mfavetti:
> PS : remind to activate phpcs on your code.

i will look into this, i didn't know about it

Shnoulle:
> > PS : remind to activate phpcs on your code.
> 
> i will look into this, i didn't know about it

What is your editor ?

mfavetti:
> > > PS : remind to activate phpcs on your code.
> > 
> > 
> > i will look into this, i didn't know about it
> 
> What is your editor ?

sublime

Shnoulle:
> > > > PS : remind to activate phpcs on your code.
> > > 
> > > 
> > > i will look into this, i didn't know about it
> > 
> > 
> > What is your editor ?
> 
> sublime

Seems there are plugin : https://github.com/benmatselby/sublime-phpcs

Currently i use geany for simple plugin, but use vscodium automatic fix before push (when i remind …)

PR #3547, opened 2023-10-17
Fixed issue [security] #19169: Privilege escalation to edit survey

Shnoulle:
Dev: Permission check must be done before updating survey

### DIFF
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -813,18 +813,18 @@ class SurveyAdministrationController extends LSBaseController
 
         foreach ($aSIDs as $iSurveyID) {
             $oSurvey = Survey::model()->findByPk((int)$iSurveyID);
-            $oSurvey->gsid = $iSurveyGroupId;
             $aResults[$iSurveyID]['title'] = $oSurvey->correct_relation_defaultlanguage->surveyls_title;
             if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveysettings', 'update')) {
                 $aResults[$iSurveyID]['result'] = false;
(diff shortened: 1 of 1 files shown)


application/controllers/SurveyAdministrationController.php
```diff
@@ -813,18 +813,18 @@ public function actionChangeMultipleSurveyGroup()
 
         foreach ($aSIDs as $iSurveyID) {
             $oSurvey = Survey::model()->findByPk((int)$iSurveyID);
-            $oSurvey->gsid = $iSurveyGroupId;
```
gabrieljenik, line 816: Why moving this line? Intended?

Shnoulle: It's **the** issue if you set the gid before testing is user have Permission : it was tested with new gid. If user have Survey permission onnew gid : return true.

It must be tested with current gid

Shnoulle: See : https://github.com/LimeSurvey/LimeSurvey/blob/d3fb278e57e5561a62b1e05c619db5b1c5fbef73/application/models/Survey.php#L2279

return with new gid here : https://github.com/LimeSurvey/LimeSurvey/blob/d3fb278e57e5561a62b1e05c619db5b1c5fbef73/application/models/Survey.php#L2294

Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/3547/commits/173c2d0755b278cace63511ce9e9bad03810a270

gabrieljenik: > Why after ? We already check if user have permission to view surveygroup in Survey model.

Ok, didn't recall it.

Shnoulle:
Code review done ?

PR #3550, opened 2023-10-18
New feature #19201: Add column "active" to users, to be able to activate or deactivate users

mohabmes:


Shnoulle:
My opinion : seems to be create event for only one plugin. Can be great if we can use for other purpose

application/controllers/UserManagementController.php
```diff
@@ -420,6 +424,253 @@ public function actionDeleteUser()
...
+
+    /**
+     * Stores the status settings run via MassEdit
+     *
+     * @throws CException
+     */
+    public function actionBatchStatus()
+    {
+        if (!Permission::model()->hasGlobalPermission('users', 'update')) {
+            return $this->renderPartial(
```
olleharstedt, line 523: I think in other places we throw a 403 exception html code.

Shnoulle: 403 is always better for security. Allow to be tracked by other tool.

Shnoulle:
OK, i think it's a PR to add a core plugin for Cloud version.

But it's updating CE version for cloud version … 
`joomlaVerifyPassword` included in CE.

Seems strange for me.

application/controllers/UserManagementController.php
```diff
@@ -67,10 +67,14 @@ public function actionIndex()
             false
         );
 
+        $oEvent = new PluginEvent('beforeAdminUserActivation');
+        $oEvent->set('request', App()->request);
+        App()->getPluginManager()->dispatchEvent($oEvent);
+        $isUserLimitReached = $oEvent->get('is_user_limit_reached', null);
```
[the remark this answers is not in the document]
olleharstedt: Please ignore this PR, Denis, it's not ready for outside review.

olleharstedt:
> OK, i think it's a PR to add a core plugin for Cloud version.
> 
> But it's updating CE version for cloud version … `joomlaVerifyPassword` included in CE.
> 
> Seems strange for me.

Yea, please wait, we'll clean this up. ^^

olleharstedt:
Modified the title of the PR.

olleharstedt:
Next step would be to remove everything LS Cloud related from this PR and only keep functionality related to the new feature activate/deactivate user. :)

application/core/plugins/Authdb/Authdb.php
```diff
@@ -62,11 +62,12 @@ public function createNewUser()
             $new_email = flattenText($preCollectedUserArray['email']);
             $new_full_name = flattenText($preCollectedUserArray['full_name']);
             $presetPassword = flattenText($preCollectedUserArray['password']);
+            $status = $preCollectedUserArray['status'];
```
olleharstedt, line 65: It's not enough to change only the Authdb plugin, because there are many other auth plugins, e.g. SAML. The limitation to block non-activated users from logging in must happen somewhere else.

application/helpers/LayoutHelper.php
```diff
@@ -151,6 +151,9 @@ public function showadminmenu($aData): ?string
 
             // Fetch extra menus from plugins, e.g. last visited surveys
             $aData['extraMenus'] = $this->fetchExtraMenus($aData);
+            if (\Permission::model()->hasGlobalPermission("superadmin")) {
+                $aData['extraMod'] = $this->fetchExtraLimeServiceMod();
```
olleharstedt, line 155: Not part of this PR

application/helpers/LayoutHelper.php
```diff
@@ -164,6 +167,27 @@ public function showadminmenu($aData): ?string
         return null;
     }
 
+    /**
+     * Get extra menus from plugins that are using event beforeAdminMenuRender
+     *
+     * @param array $aData
+     * @return array<ExtraMenu>
+     */
+    protected function fetchExtraLimeServiceMod(): ?string
```
olleharstedt, line 176: Not part of this PR

application/helpers/LayoutHelper.php
```diff
@@ -338,6 +362,11 @@ public function getAdminFooter(string $url, bool $return = false, bool $question
 
         $aData['imageurl'] = Yii::app()->getConfig("imageurl");
         $aData['url'] = $url;
+
+        $oEvent = new PluginEvent('beforeFooterRender');
```
olleharstedt, line 366: Not part of this PR

application/models/User.php
```diff
@@ -43,6 +43,7 @@
  * @property string $last_login
  * @property Permissiontemplates[] $roles
  * @property UserGroup[] $groups
+ * @property bool $status
```
olleharstedt, line 46: Please document the valid values for $status and what they mean.

application/models/User.php
```diff
@@ -300,6 +276,13 @@ public function checkPassword($sPassword)
             $this->setPassword($sPassword, true);
             return true;
         }
+
+        // If password comes from the Joomla user, use this to check.
+        // TODO: Method copied from Authdb plugin.
+        if ($this->joomlaVerifyPassword($sPassword, $this->password)) {
+            return true;
```
olleharstedt, line 283: Not part of this PR

application/models/User.php
```diff
@@ -425,6 +408,57 @@ public function getPasswordHelpText()
         return($txt);
     }
 
+    /**
+     * Copied from Authdb.php.
+     * @param   string   $password  The plaintext password to check.
+     * @param   string   $hash      The hash to verify against.
+     * @return  boolean  True if the password and hash match, false otherwise
+     */
+    protected function joomlaVerifyPassword($password, $hash)
```
olleharstedt, line 417: Not part of this PR

application/views/admin/super/footer.php
```diff
@@ -246,6 +246,7 @@ class='btn btn-warning btn-sm'
 
 <?php
 $this->renderPartial('/admin/htmleditor/modal_editor_partial');
+$this->renderPartial('/userManagement/partial/planupgrade', ['currentPlan' => $currentPlan]);
```
olleharstedt, line 249: Not part of this PR

application/views/userManagement/index.php
```diff
@@ -31,6 +31,8 @@
             'lsAfterAjaxUpdate' => [
                 'bindListItemclick();',
                 'LS.UserManagement.bindButtons();',
+                'showDeactivatedUserTooltip();',
+                'LS.UserManagement.updateUserLimit();',
```
olleharstedt, line 35: Not part of this PR

application/core/plugins/Authdb/Authdb.php
Shnoulle: new login with low priority (happen after all) at https://manual.limesurvey.org/AfterSuccessfulLogin  ?

I think it can work.

application/helpers/LayoutHelper.php
Shnoulle: No :the event can be in this commit, but the content added/returned/updated in the plugin : no ?

application/views/userManagement/partial/confirmuseractivation.php
```diff
@@ -0,0 +1,22 @@
+<?php
+Yii::app()->getController()->renderPartial(
+    '/layouts/partial_modals/modal_header',
+    ['modalTitle' => gT(ucfirst($action) .' user')]
+);
+?>
+
+<div class="modal-body">
+    <p class='modal-body-text'><?php eT("Do you want to $action this user?"); ?></p>
```
olleharstedt, line 9: Not a valid localized string. :) Please see https://manual.limesurvey.org/Code_quality_guide#Localization

application/views/userManagement/index.php
Shnoulle: Need an js event where plugin can add own js action ?

something like $("#grid").trigger("gridUpdated:end") ([see why adding :](https://learn.jquery.com/events/introduction-to-custom-events/#naming-custom-events))

application/views/userManagement/partial/confirmuseractivation.php
```diff
@@ -0,0 +1,22 @@
...
+</div>
+<div class="modal-footer">
+    <?=TbHtml::formTb(null, App()->createUrl('userManagement/userActivateDeactivate'), 'post', ["id"=>"UserManagement--modalform"])?>
+    <input type="hidden" name="userid" value="<?= $userId ?>" />
+    <input type="hidden" name="action" value="<?= $action ?>" />
+    <button type="button"  class="btn btn-cancel" data-bs-dismiss="modal">&nbsp;<?php eT("Cancel"); ?></button>
+    <?php if (isset($showUpgradeModal) && $showUpgradeModal): ?>
+        <button class="btn btn-primary" id="modalPlanUpgradeOpenAfterSave" data-bs-dismiss="modal" data-stackmodal="#modalPlanUpgrade" onclick="modalPlanUpgradeOpen()">&nbsp;<?php eT("Save"); ?></button>
```
olleharstedt, line 17: Not in this PR

application/views/userManagement/partial/planupgrade.php
```diff
@@ -0,0 +1,140 @@
+<div id="modalPlanUpgrade" class="modal fade border-10" role="dialog" style="margin: auto; background-color:rgba(0, 0, 0, 0.5);">
```
olleharstedt, line 1: Not in this PR

application/views/userManagement/partial/planupgradebutton.php
```diff
@@ -0,0 +1,6 @@
+<div>
+    <p>
+        <?php echo gT("Please contact your administrator or upgrade your plan to activate more users."); ?>
```
olleharstedt, line 3: Not in this PR

application/views/userManagement/partial/topbarBtns/leftSideButtons.php
```diff
@@ -1,33 +1,43 @@
...
 
 if (Permission::model()->hasGlobalPermission('users', 'create')) {
+    $userHtmlOptions = [
+        'class' => 'btn btn-primary UserManagement--action--openmodal',
+        'data-bs-toggle' => 'modal',
+        'data-href' => $this->createUrl("userManagement/addEditUser")
+    ];
+
+    if ($showUpgradeModal) {
+        $userHtmlOptions['data-stackmodal'] = "#modalPlanUpgrade";
```
olleharstedt, line 11: Not in this PR

application/controllers/AdminController.php
```diff
@@ -498,6 +498,10 @@ public function getAdminFooter($url, $explanation, $return = false)
             $aData['versiontitle'] = gT('Version');
         }
 
+        $oEvent = new PluginEvent('beforeFooterRender');
+        App()->getPluginManager()->dispatchEvent($oEvent);
+        $aData['currentPlan'] = $oEvent->get('current_plan', 'free');
```
olleharstedt, line 503: Not in this PR

application/controllers/UserManagementController.php
```diff
@@ -67,10 +67,14 @@ public function actionIndex()
             false
         );
 
+        $oEvent = new PluginEvent('beforeAdminUserActivation');
+        $oEvent->set('request', App()->request);
+        App()->getPluginManager()->dispatchEvent($oEvent);
+        $isUserLimitReached = $oEvent->get('disable', null);
```
olleharstedt, line 73: User limit checks must be moved to the LimeSurvey Professional plugin on LS Cloud repo.

application/controllers/UserManagementController.php
```diff
@@ -420,6 +424,253 @@ public function actionDeleteUser()
...
+    /**
+     * Show user activation confirmation
+     *
+     * @return void|string
+     * @throws CException
+     */
+    public function actionActivationConfirm()
+    {
+        if (!Permission::model()->hasGlobalPermission('users', 'update')) {
+            return $this->renderPartial(
```
olleharstedt, line 436: Throw 403 exception instead (see other parts of code).

application/controllers/UserManagementController.php
```diff
@@ -420,6 +424,253 @@ public function actionDeleteUser()
...
+    /**
+     * Stores the status settings
+     *
+     * @return void|string
+     * @throws CException
+     */
+    public function actionUserActivateDeactivate()
+    {
+        if (!Permission::model()->hasGlobalPermission('users', 'delete')) {
+            return $this->renderPartial(
```
olleharstedt, line 471: 403 Exception

application/controllers/UserManagementController.php
```diff
@@ -420,6 +424,253 @@ public function actionDeleteUser()
...
+                'partial/error',
+                ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
+            );
+        }
+
+        $userIds = json_decode(Yii::app()->request->getPost('sItems', "[]"));
+        $status = Yii::app()->request->getPost('status_selector', '');
+
+        $oEvent = new PluginEvent('beforeAdminUserActivation');
```
olleharstedt, line 532: Same here, this check must be moved to LS Cloud plugin.

application/helpers/LayoutHelper.php
olleharstedt: Yea but only if the event is needed. Not sure yet.

Shnoulle: > Yea but only if the event is needed. Not sure yet.

it can be just great to have a GUI event here

olleharstedt: beforeCloseHtml is only for survey rendering then, not admin view?

Shnoulle: > beforeCloseHtml is only for survey rendering then, not admin view?

[Yes](https://github.com/LimeSurvey/LimeSurvey/blob/8b2d1c7ac40b40caea6a90df2df581fe6bb11661/application/core/LSYii_ClientScript.php#L620), and not allowed to add content at any place, only at end.

I only use it to add script sometimes.

application/core/plugins/Authdb/Authdb.php
mohabmes: AfterSuccessfulLogin will not be useful in this context.

This functionality is intended to create a user with a predetermined status based on your chosen plan. If the maximum number of active users is reached, any additional accounts will be automatically deactivated; otherwise, they will be activated.

mohabmes: Moving it to the LS Cloud repository, it will always remain active for the CE.

Shnoulle: > AfterSuccessfulLogin will not be useful in this context.

We don't know if it's a new user here ? 

Maybe need to add an event inside Auth(DB|LDAP|…) plugin ?

mohabmes: We will use createNewUser but will move it to LS cloud instead.

olleharstedt:
@mohabmes Looks like someone else also defined a database update for 616, which means you'll have to move yours to 617. And then pray no one adds 617 before yours is merged. :)

assets/admin_themes/Sea_Green/admin_customizations.scss
```diff
@@ -192,3 +192,9 @@ blockquote {
 #surveySettingsForThisGroup {
   column-gap: 20px;
 }
+
+#usermanagement--identity-gridPanel tr.disabled {
+  color            : $g-400;
+  pointer-events   : auto;
+  cursor           : not-allowed;
+}
```
ptelu, line 200: did you check if this needs to also be adjusted in other themes? same with the next usermanagement.js file

Shnoulle:
> Next step would be to remove everything LS Cloud related from this PR and only keep functionality related to the new feature activate/deactivate user. :)

For information : for such purpose, i already use BeforePermssion event : permission to authenticat : return now and add an alert.

It work :)

application/models/User.php
```diff
@@ -527,6 +501,37 @@ public function getManagementButtons()
                     )
                 )
         ];
+
+        $permission = ( $permission_superadmin_read && !(Permission::isForcedSuperAdmin($this->uid) || $this->uid == App()->user->getId()))
+            || (!$permission_superadmin_read && ($this->uid != App()->session['loginID'] //Can't change your own permissions
+                    && ( $permission_users_update && $ownedOrCreated)
```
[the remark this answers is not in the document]
Shnoulle: Not currently : see https://bugs.limesurvey.org/view.php?id=19127

application/models/User.php
```diff
@@ -231,6 +206,7 @@ public static function insertUser($new_user, $new_pass, $new_full_name, $parent_
         $oUser->created = date('Y-m-d H:i:s');
         $oUser->modified = date('Y-m-d H:i:s');
         $oUser->expires = $expires;
+        $oUser->status = $status;
```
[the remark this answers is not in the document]
Shnoulle: > Why?

I write : **unsure** : https://github.com/LimeSurvey/LimeSurvey/blob/c5680fd935db6de2d3178fa81dafc70345f1f1ee/vendor/yiisoft/yii/framework/db/schema/mssql/CMssqlSchema.php#L41

If i don't make error : we use 1/0 on other action for boolean
But seems it's OK : https://learn.microsoft.com/en-us/sql/t-sql/data-types/bit-transact-sql?view=sql-server-ver16


mysql use tinyint(1) : https://github.com/LimeSurvey/LimeSurvey/blob/c5680fd935db6de2d3178fa81dafc70345f1f1ee/vendor/yiisoft/yii/framework/db/schema/mysql/CMysqlSchema.php#L38
But seems OK too : https://dba.stackexchange.com/a/233133

No issue.

olleharstedt:
Merged develop into this branch again.

application/controllers/UserManagementController.php
```diff
@@ -420,6 +420,204 @@ public function actionDeleteUser()
...
+     * @throws CException
+     */
+    public function actionUserActivateDeactivate()
+    {
+        if (!Permission::model()->hasGlobalPermission('users', 'delete')) {
+            throw new CHttpException(403, gT("You do not have permission to access this page."));
+        }
+        $userId = sanitize_int(Yii::app()->request->getParam('userid'));
+        $action = Yii::app()->request->getParam('action');
+        $oUser = User::model()->findByPk($userId);
```
olleharstedt, line 459: You wanna throw a chttp exception 404 if user is not found?

application/controllers/UserManagementController.php
```diff
@@ -420,6 +420,204 @@ public function actionDeleteUser()
...
+     * Handle user activation with regard to users limit
+     *
+     * @return array
+     */
+    public function limitedUserActivation($userIds, $operation, $limit)
+    {
+        $results = [];
+        foreach ($userIds as $iUserId) {
+            $oUser = User::model()->findByPk($iUserId);
+            $results[$iUserId]['title'] = $oUser->users_name;
```
olleharstedt, line 579: Same here, check for null :)

olleharstedt: Or wait, maybe this function should be moved to cloud plugin instead? Because core LS does not care about limits.

olleharstedt: Which also means the "unlimited" above should be renamed. No point have "unlimited" if there's no "limited". :)

application/controllers/UserManagementController.php
```diff
@@ -420,6 +420,204 @@ public function actionDeleteUser()
...
+    public function actionUserLimit()
+    {
+        if (!Permission::model()->hasGlobalPermission('users', 'update')) {
+            return $this->renderPartial(
+                'partial/error',
+                ['errors' => [gT("You do not have permission to access this page.")], 'noButton' => true]
+            );
+        }
+
+        $oEvent = new PluginEvent('beforeAdminUserActivation');
```
olleharstedt, line 529: This event sounds like it should be triggered before a user status is updated?

olleharstedt: We can also use a more general model event for the same purpose: https://manual.limesurvey.org/BeforeModelSave

application/controllers/UserManagementController.php
```diff
@@ -420,6 +420,204 @@ public function actionDeleteUser()
...
+        Yii::app()->getController()->renderPartial('ext.admin.survey.ListSurveysWidget.views.massive_actions._action_results', $data);
+    }
+
+    /**
+     * Get the users limit & active users count
+     *
+     * @return string | JSON
+     * @throws CException
+     */
+    public function actionUserLimit()
```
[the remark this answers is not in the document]
mohabmes: Removed from CE

application/controllers/AdminController.php
```diff
@@ -498,6 +498,14 @@ public function getAdminFooter($url, $explanation, $return = false)
             $aData['versiontitle'] = gT('Version');
         }
 
+        $oEvent = new PluginEvent('beforeFooterRender');
+        App()->getPluginManager()->dispatchEvent($oEvent);
+        $data = $oEvent->get('data');
+
+        foreach ($data as $key => $item) {
```
olleharstedt, line 505: Isn't this just `array_merge`?

Shnoulle: Usage of `$event->append(` for plugins 

See https://manual.limesurvey.org/BeforeToolsMenuRender 

https://github.com/LimeSurvey/LimeSurvey/blob/26cd37a09c7250b75ef3e1a8f212ba8fb4e8be31/application/core/SurveyCommonAction.php#L774

**Warning** : broke if data is null with php8.1. (remind to activate debug for fix and new feature).

application/controllers/AdminController.php
```diff
@@ -498,6 +498,14 @@ public function getAdminFooter($url, $explanation, $return = false)
             $aData['versiontitle'] = gT('Version');
         }
 
+        $oEvent = new PluginEvent('beforeFooterRender');
+        App()->getPluginManager()->dispatchEvent($oEvent);
+        $data = $oEvent->get('data');
```
olleharstedt, line 503: If multiple plugins are using the same event, you might need to append the data rather than just getting it. Or we don't care about that.

application/models/User.php
```diff
@@ -59,6 +60,9 @@ class User extends LSActiveRecord
      * @var string $lang Default value for user language
      */
     public $lang = 'auto';
+
+    public $status;
```
olleharstedt, line 64: This is already implied by the automagic LSActiveRecord, no?

installer/create-database.php
```diff
@@ -1097,7 +1097,8 @@ function populateDatabase($oDB)
             'validation_key' => 'string(38)',
             'validation_key_expiration' => 'datetime',
             'last_forgot_email_password' => 'datetime',
-            'expires' => 'datetime'
+            'expires' => 'datetime',
+            'status' => 'BOOLEAN DEFAULT TRUE'
```
olleharstedt, line 1101: Sadly, this is a keyword in MySQL. Rename to active_status or such?

application/controllers/AdminController.php
olleharstedt: To include custom data from database to javascript, one can use asset manager, register script, and the script can contain a global JS object with the custom data

```
App()->registerScript('name_of_script', <<<JAVASCRIPT
var installationData = {
  current_plan: $currentPlan  // Etc more data here
};
```

application/controllers/AdminController.php
```diff
@@ -498,6 +498,14 @@ public function getAdminFooter($url, $explanation, $return = false)
             $aData['versiontitle'] = gT('Version');
         }
 
+        $oEvent = new PluginEvent('beforeFooterRender');
```
olleharstedt, line 501: This is event is perhaps not needed if you use asset manager to register a script with current_plan instead?

https://github.com/LimeSurvey/cloud_client_limesurvey/pull/24/files#r1415473939

installer/create-database.php
mohabmes: Renamed to user_status

application/controllers/AdminController.php
olleharstedt: @mohabmes Can you comment on why you marked this as resolved?

mohabmes: Registering a script with the `current_plan` at this point serves no purpose; we only wanted that event to transmit the current plan for the partial rendering of the appropriate plan upgrade modal, displaying the accurate next bigger plan. (no JS involved).

https://github.com/LimeSurvey/cloud_client_limesurvey/blob/cf8134b0814d93a98659e8f5ad9c54364a070ac4/dbversion-600/application/views/admin/super/footer.php#L239

olleharstedt: Ah ok. In that case, it's probably better if the beforeFooterRender spits out HTML instead of a data array, so that other plugins can insert whatever HTML they want without changing any LS core view files.

That also means that the view file `planupgrade.php` can be moved to the LS Pro plugin folder instead of the core view folder.

olleharstedt: Compare with this event: https://manual.limesurvey.org/BeforeCloseHtml (only available in survey taking view, afaik; @Shnoulle you can confirm?).

So you can add a similar event to admin view called `beforeCloseAdminHtml`.

Shnoulle: Menu class : yes, it's interesting.

With public : we can rplace or add .twig view, but need netter too … with a way to really use twig block and extend

olleharstedt: Seems overkill for just adding some footer HTML, or?

Shnoulle: > Seems overkill for just adding some footer HTML, or?

I speak in general here :)

application/core/plugins/Authdb/Authdb.php
```diff
@@ -171,7 +178,7 @@ public function newUserSession()
             $this->setAuthFailure(self::ERROR_USERNAME_INVALID);
             return;
         }
-        if ($user->isExpired()) {
+        if ($user->isExpired() || (isset($user->user_status) && !$user->user_status)) {
```
olleharstedt, line 181: I think this check has to be moved to

application/core/LSUserIdentity.php
 71                 $authEvent = new PluginEvent('newUserSession', $this); // TODO: rename the plugin function authenticate()                                                                  

Basically, never fire up the event "newUserSession" if username belongs to a non-active user.

olleharstedt: Ah yeah, I did think about how to achieve that.

olleharstedt: But then we get the same issue with updates, or?

Shnoulle: > But then we get the same issue with updates, or?

Yes, but need to be fixed on the same way :)

Shnoulle:
Someone need to report the issue with user keeping session when set as inactive !

PR #3551, opened 2023-10-19
Fixed issue #19180: SMTP settings are not enabled when SMTP is selected in global settings

mfavetti:
Fixed issue #19180: SMTP settings are not enabled when SMTP is selected in global settings

[https://bugs.limesurvey.org/view.php?id=19180](https://bugs.limesurvey.org/view.php?id=19180)

### DIFF
--- a/assets/scripts/admin/globalsettings.js
+++ b/assets/scripts/admin/globalsettings.js
@@ -112,19 +112,14 @@ function Emailchange(ui,evt)
 {
     const selectedMethod = $('#emailmethod input:radio:checked').val();
 
-    const smtp_enabled = (selectedMethod == 'smtp');
-    if (smtp_enabled) {
-        $('#emailsmtpssl label').removeClass('disabled');
-        $('#emailsmtpdebug label').removeClass('disabled');
-    } else {
-        $('#emailsmtpdebug label').addClass('disabled');
-        $('#emailsmtpssl label').addClass('disabled');
-    }
-    $("#emailsmtphost").prop('disabled', true);
-    $("#emailsmtpuser").prop('disabled', true);
-    $("#emailsmtppassword").prop('disabled', true);
+    const smtp_enabled = selectedMethod === 'smtp';
+    $('#emailsmtpssl label').toggleClass('disabled', !smtp_enabled);
+    $('#emailsmtpdebug label').toggleClass('disabled', !smtp_enabled);
+    $("#emailsmtphost").prop('disabled', !smtp_enabled);
+    $("#emailsmtpuser").prop('disabled', !smtp_enabled);
+    $("#emailsmtppassword").prop('disabled', !smtp_enabled);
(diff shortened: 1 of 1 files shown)


assets/scripts/admin/globalsettings.js
```diff
@@ -112,19 +112,14 @@ function Emailchange(ui,evt)
...
-        $('#emailsmtpdebug label').removeClass('disabled');
-    } else {
-        $('#emailsmtpdebug label').addClass('disabled');
-        $('#emailsmtpssl label').addClass('disabled');
-    }
-    $("#emailsmtphost").prop('disabled', true);
-    $("#emailsmtpuser").prop('disabled', true);
-    $("#emailsmtppassword").prop('disabled', true);
+    const smtp_enabled = selectedMethod === 'smtp';
+    $('#emailsmtpssl label').toggleClass('disabled', !smtp_enabled);
```
Shnoulle, line 116: toggleClass ? https://api.jquery.com/toggleClass/

No parameters in toggleClass

Here you remove if it's set and add if not. If you select 2 times smtp : you add disabled again.

I think the issue it ws not set at loading ? Maybe ?

Shnoulle (changes requested):
We can not use toggleClass with extra parameter

assets/scripts/admin/globalsettings.js
```diff
@@ -112,19 +112,14 @@ function Emailchange(ui,evt)
...
+    $('#emailsmtpdebug label').toggleClass('disabled', !smtp_enabled);
+    $("#emailsmtphost").prop('disabled', !smtp_enabled);
```
Shnoulle, line 118: 👍 this part is OK :)

Only issue with label

assets/scripts/admin/globalsettings.js
mfavetti: ????

the function is overloaded and has multiple signatures. Its the second function signature on the page you just linked: https://api.jquery.com/toggleClass/#toggleClass-className-state

".toggleClass( className, state )
className
Type: String
One or more classes (separated by spaces) to be toggled for each element in the matched set.
state
Type: Boolean
A boolean (not just truthy/falsy) value to determine whether the class should be added or removed."

toggleClass is literally purpose built to make those if/else statements on addClass/removeClass unnecessary

been in jquery since 1.3

mfavetti: if the method is smtp, smtp_enabled is true

the disabled class will get removed from the two button groups
and the disabled property will be false for the three inputs

if the method is not smtp, smtp_enabled is false

the disabled class will get added to the two button groups
and the disabled property will be true for the three inputs

i've already tested it and it works

Shnoulle:
> i don't agree

Yes, you're right … sorry

Shnoulle (approved):

PR #3555, opened 2023-10-20
Fixed issue #CT-256: Problems with ini parsing for GTranslate.php

twilligls:
Fixed issue #CT-256: Problems with ini parsing for GTranslate.php

### DIFF
--- a/vendor/gtranslate-api/GTranslate.php
+++ b/vendor/gtranslate-api/GTranslate.php
@@ -143,15 +143,19 @@ class GTranslate
 		return false;
 	}
 
-	/**
-	* Parse available language from language file
-	* @access private
-	*/	
-	private function parseLanguageFile()
-	{
-		$this->available_languages = parse_ini_file($this->available_languages_file);
-	}	
-	
+    /**
+     * Parse available language from language file
+     * @access private
+     */
+    private function parseLanguageFile()
+    {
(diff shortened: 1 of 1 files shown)


twilligls:
Some server had problems parsing the ini file, this fix works there and everywhere

olleharstedt:
This should rather be a PR to the third party lib. Did you notify them about our issues?

twilligls:
> This should rather be a PR to the third party lib. Did you notify them about our issues?

The library was last updated in 2011. I don't think anyone is working on it.

olleharstedt:
> The library was last updated in 2011. I don't think anyone is working on it.

Are there other libraries with same functionality, that are still updated?

twilligls:
> > The library was last updated in 2011. I don't think anyone is working on it.
> 
> Are there other libraries with same functionality, that are still updated?

Looks like this seems to be the official google one:
https://github.com/googleapis/google-cloud-php-translate

olleharstedt:
> > > The library was last updated in 2011. I don't think anyone is working on it.
> > 
> > 
> > Are there other libraries with same functionality, that are still updated?
> 
> Looks like this seems to be the official google one: https://github.com/googleapis/google-cloud-php-translate

Nice. Guess I can merge this fix and then add the new lib to backlog...

PR #3556, opened 2023-10-20
Fixed issue #19161: invalid adminemail/bounce_email address in lss results in white screen on import

gabrieljenik:
Reverted CR-1253
- 5652f139c62323979da22e98c3ede44a4e4416b9
- 1e39019bd54b1c531a94fb95d885d6dcead00c25 

Changes:
- It prevents the import from “dying” with a blank screen.
- Remove the email validation from the filter validator, because it does not belong there. That is a pollyfil for php 8.1
- Moved it to LSYii_EmailIDNAValidator.php, adding `allowInherit` there.


gabrieljenik:
@mohabmes Would be good if you can take a look as well as to get your comments. Thanks

Shnoulle (approved):

application/core/LSYii_EmailIDNAValidator.php
```diff
@@ -21,23 +21,30 @@ class LSYii_EmailIDNAValidator extends CValidator
 
     public $allowEmpty = false;
     public $allowMultiple = false;
-
+    public $allowInherit = false;
 
     public function validateAttribute($object, $attribute)
     {
-
+        // If the attribute is empty and empty values are allowed, it's valid.
```
Shnoulle, line 28: Here : it's not the task to this rules to control empty :)

It's another rule

PR #3567, opened 2023-10-24
Fixed issue #19057: incorrect handling of wrong input in create dummy participant form

mfavetti:
Fixed issue #19057: incorrect handling of wrong input in create dummy participant form

[https://bugs.limesurvey.org/view.php?id=19057](https://bugs.limesurvey.org/view.php?id=19057)

master


Shnoulle (commented):
Avoid usage of elseif is not needed and more comments if possible

(same than 5.X : https://github.com/LimeSurvey/LimeSurvey/pull/3568#pullrequestreview-1694223332)

Shnoulle (approved):

PR #3568, opened 2023-10-24
Fixed issue #19057: incorrect handling of wrong input in create dummy participant form

mfavetti:
Fixed issue #19057: incorrect handling of wrong input in create dummy participant form

[https://bugs.limesurvey.org/view.php?id=19057](https://bugs.limesurvey.org/view.php?id=19057)

5.x

### DIFF
--- a/application/controllers/admin/Tokens.php
+++ b/application/controllers/admin/Tokens.php
@@ -910,7 +910,11 @@ class Tokens extends SurveyCommonAction
             }
             $aData['thissurvey'] = getSurveyInfo($iSurveyId);
             $aData['surveyid'] = $iSurveyId;
-            if ($cntAttributeErrors > 0) { // attribute validation errors
+            if ($newDummyToken === 0) {
+                $aData['success'] = false;
+                Yii::app()->session['flashmessage'] = gT("No dummy participants were added.");
+                $this->getController()->redirect(array("/admin/tokens/sa/browse/surveyid/{$iSurveyId}"));
+            } elseif ($cntAttributeErrors > 0) { // attribute validation errors
                 $aData['dateformatdetails'] = getDateFormatData(Yii::app()->session['dateformat'], App()->language);
                 $aData['aAttributeFields'] = getParticipantAttributes($iSurveyId);
 
--- a/application/views/admin/token/dummytokenform.php
+++ b/application/views/admin/token/dummytokenform.php
@@ -23,7 +23,7 @@
                 <div class="form-group">
                     <label  class=" control-label" for='amount'><?php eT("Number of participants:"); ?></label>
                     <div class="">
-                        <input class='form-control' type='text' size='20' id='amount' name='amount' value="<?php echo $amount; ?>" />
+                        <input class='form-control' type='number' min='1' size='20' id='amount' name='amount' value="<?php echo $amount; ?>" />
                     </div>
                 </div>
 


application/controllers/admin/Tokens.php
```diff
@@ -910,7 +910,11 @@ public function addDummies(int $iSurveyId, string $subaction = '')
             }
             $aData['thissurvey'] = getSurveyInfo($iSurveyId);
             $aData['surveyid'] = $iSurveyId;
-            if ($cntAttributeErrors > 0) { // attribute validation errors
+            if ($newDummyToken === 0) {
+                $aData['success'] = false;
+                Yii::app()->session['flashmessage'] = gT("No dummy participants were added.");
+                $this->getController()->redirect(array("/admin/tokens/sa/browse/surveyid/{$iSurveyId}"));
+            } elseif ($cntAttributeErrors > 0) { // attribute validation errors
```
Shnoulle, line 917: Avoid usage of elseif is not needed and more comments if possible

```
if ($newDummyToken === 0) {
    // No token created : show flash and redirect
    $aData['success'] = false;
    Yii::app()->session['flashmessage'] = gT("No dummy participants were added.");
    $this->getController()->redirect(array("/admin/tokens/sa/browse/surveyid/{$iSurveyId}"));
    // redirected : code end here
}
if ($cntAttributeErrors > 0) {
    // etc
```

application/views/admin/token/dummytokenform.php
```diff
@@ -23,7 +23,7 @@
                 <div class="form-group">
                     <label  class=" control-label" for='amount'><?php eT("Number of participants:"); ?></label>
                     <div class="">
-                        <input class='form-control' type='text' size='20' id='amount' name='amount' value="<?php echo $amount; ?>" />
+                        <input class='form-control' type='number' min='1' size='20' id='amount' name='amount' value="<?php echo $amount; ?>" />
```
Shnoulle, line 26: 👍

Shnoulle (commented):
Avoid usage of elseif is not needed and more comments if possible

application/controllers/admin/Tokens.php
mfavetti: i have to work within constraints of existing code... ideally i would cover this case and simply return to stop execution of the rest of the function, but none of the other cases do this already (no precedent to follow) and the zero case needs to be handled before the other cases, and the existing code uses if elseif else structure already

otherwise i would need to rewrite the entire function...

mfavetti: okay i can add comments, but does redirect call stop execution of the rest of the function?

Shnoulle: > okay i can add comments, but does redirect call stop execution of the rest of the function?

Yes : redirect is redirect


> existing code use if elseif, which implies that it does not

You add code : better to add it clean.

You're right : the other else are not needed, do it if you want.

Shnoulle: > no precedent to follow

On a lot of other file

mfavetti: can you give example to demonstrate? i'm new to the code guidelines for limesurvey, so not sure when I read code what is already good and what needs to be updated

Shnoulle: Just add 

```
if ($newDummyToken === 0) {
    // No token created : show flash and redirect
    $aData['success'] = false;
    Yii::app()->session['flashmessage'] = gT("No dummy participants were added.");
    $this->getController()->redirect(array("/admin/tokens/sa/browse/surveyid/{$iSurveyId}"));
    // redirected : code end here
}
```

before the current if , no ?

Shnoulle: Trying to create a PR on your branch : but can not :(

mfavetti: > > okay i can add comments, but does redirect call stop execution of the rest of the function?
> 
> Yes : redirect is redirect
> 
> > existing code use if elseif, which implies that it does not
> 
> You add code : better to add it clean.
> 
> You're right : the other else are not needed, do it if you want.

redirect can be tricky across frameworks... I looked up for Yii, 1.1 there is a terminate arg which defaults to true, which controls termination of current application, but yii 2 requires return of redirect to accomplish the same

mfavetti: > Trying to create a PR on your branch : but can not :(

whoops, i will look how to allow that

Shnoulle: > redirect can be tricky across frameworks... I looked up for Yii, 1.1 there is a terminate arg which defaults to true, which controls termination of current application, but yii 2 requires return of redirect to accomplish the same

We can add a App()->end()

mfavetti: > Trying to create a PR on your branch : but can not :(

is it a limitation with forks? I can't find where to enable/disable PRs for a repo

Shnoulle:
I add code review done : it's not needed :)

application/controllers/admin/Tokens.php
mfavetti: idk how to do it, i tried to invite you as collaborator, maybe that will work?

Shnoulle: I create a fork on my repo :) 

Since it's not a security issue : can take time ;)

Shnoulle (approved):

PR #3569, opened 2023-10-24
Fixed issue #19188: css problems in survey permissions after reload

mfavetti:
Fixed issue #19188: css problems in survey permissions after reload

[https://bugs.limesurvey.org/view.php?id=19188](mantis #19188)

master


Shnoulle (commented):
This apply to all select2, is this needed for all : potential issue ?

See [plugin manager ](https://github.com/LimeSurvey/LimeSurvey/blob/58effe2996d1fc4a07b02c95b90230d8db5cf8ba/application/extensions/SettingsWidget/SettingsWidget.php#L494)  you re sure 

Maybe better use it ?

If the select2 didn't have wrapper : it can take whole screen …


Alternative solution data-select2-width on the select element ?

mfavetti:
> This apply to all select2, is this needed for all : potential issue ?
> 
> See [plugin manager ](https://github.com/LimeSurvey/LimeSurvey/blob/58effe2996d1fc4a07b02c95b90230d8db5cf8ba/application/extensions/SettingsWidget/SettingsWidget.php#L494)
> 
> Maybe better use it ?
> 
> If the select2 didn't have wrapper : it can take whole screen …
> 
> Alternative solution data-select2-width on the select element ?

okay i need to investigate more... tomorrow

mfavetti:
okay, just fixed for this particular form

mfavetti:
whoops, pushed the wrong commit to that branch yesterday. pushed the correct one now.

Shnoulle:
> okay, just fixed for this particular form

Strange form-select already have a 100% width via CSS

Capture d’écran du 2023-10-25 08-29-35

mfavetti:
yes, but select2 check the style of the element as part of deciding width to use, if not specified. in all but one of the cases in the codebase, the width is unspecified, so uses default behavior. instead of changing all instances, i explicitly set the width on affected select elements. https://select2.org/appearance#container-width

Shnoulle:
OK, select2 test

1. style is set in option : use it
2. style is set on `select` : use it
3. check current width of the `select` and use it

the css style apply to the `select` are 1px : right ? Then why it not set to 1px after … :( 

Really strange issue here, in my opinion : F5 must work better thna click (see plugins settings and dropdown)

Shnoulle (approved):
👍

mfavetti:
> OK, select2 test
> 
> 1. style is set in option : use it
> 2. style is set on `select` : use it
> 3. check current width of the `select` and use it
> 
> the css style apply to the `select` are 1px : right ? Then why it not set to 1px after … :(
> 
> Really strange issue here, in my opinion : F5 must work better thna click (see plugins settings and dropdown)

yes it is strange, I just found another one with style="width:100%;" and copy that and it works *shrug*

PR #3573, opened 2023-10-25
Fixed issue #19110: with MSSQL Lime is unable to activate survey

mfavetti:
Fixed issue #19110: with MSSQL Lime is unable to activate survey

https://bugs.limesurvey.org/view.php?id=19110

master

### DIFF
--- a/application/models/SurveyActivator.php
+++ b/application/models/SurveyActivator.php
@@ -382,6 +382,7 @@ class SurveyActivator
         //if there is an autonumber_start field, start auto numbering here
         if ($iAutoNumberStart !== false && $iAutoNumberStart > 0) {
             if (Yii::app()->db->driverName == 'mssql' || Yii::app()->db->driverName == 'sqlsrv' || Yii::app()->db->driverName == 'dblib') {
+                Yii::app()->loadHelper("admin/activate");
                 mssql_drop_coulmn_with_constraints($this->survey->responsesTableName, 'id');
                 $sQuery = "ALTER TABLE {$this->survey->responsesTableName} ADD [id] int identity({$iAutoNumberStart},1)";
                 Yii::app()->db->createCommand($sQuery)->execute();


mfavetti:
> mssql_drop_coulmn_with_constraints

I can change the function name. It's only used in one place in the entire codebase. Why deprecate and not just change? Is it likely users are directly interfacing with functions in activate_helper.php rather than using the remote control api functions?

Shnoulle:
> I can change the function name. It's only used in one place in the entire codebase. Why deprecate and not just change? Is it likely users are directly interfacing with functions in activate_helper.php rather than using the remote control api functions?

Plugin dev use core function sometimes.

But you're totally right : usage of internal function are not in API (Limesurvey API for plugin is near empty ;) )

Shnoulle (approved):

Shnoulle:
Really not needed : leave it for another PR

tiborpacalat:
Resolved here https://github.com/LimeSurvey/LimeSurvey/pull/3696/files

Shnoulle:
> Resolved here https://github.com/LimeSurvey/LimeSurvey/pull/3696/files

Unsure : here we load the helper, the other one : we fix the helper name 

Ping @mfavetti do you reproduce ?

mfavetti:
didn't try to reproduce, but this line in #3696 

https://github.com/LimeSurvey/LimeSurvey/pull/3696/files#diff-a7360b0303ae1b44aeacb5170c8b4450121650b5f48547fed9f7e81bb40b9ee3R385

seems equivalent to what i've done

Shnoulle:
> didn't try to reproduce, but this line in #3696
> 
> https://github.com/LimeSurvey/LimeSurvey/pull/3696/files#diff-a7360b0303ae1b44aeacb5170c8b4450121650b5f48547fed9f7e81bb40b9ee3R385
> 
> seems equivalent to what i've done

I think I need an ophthalmic appointment.

PR #3577, opened 2023-10-25
Fixed bug #19165: CSRF edit Blacklist settings (YES to NO)

gabrieljenik:
Permission checks added.

### DIFF
--- a/application/controllers/admin/ParticipantsAction.php
+++ b/application/controllers/admin/ParticipantsAction.php
@@ -1313,6 +1313,13 @@ class ParticipantsAction extends SurveyCommonAction
      */
     public function storeBlacklistValues()
     {
+        $this->requirePostRequest();
+
+        if (!Permission::model()->hasGlobalPermission('settings', 'update')) {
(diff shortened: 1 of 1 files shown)


application/controllers/admin/ParticipantsAction.php
```diff
@@ -1313,6 +1313,13 @@ public function blacklistControl()
      */
     public function storeBlacklistValues()
     {
+        $this->requirePostRequest();
+
+        if (!Permission::model()->hasGlobalPermission('settings', 'update')) {
```
Shnoulle, line 1318: Not related to https://github.com/LimeSurvey/LimeSurvey/blob/590c1ac76e2da298d01c34ac4e9fe7f9138e343d/application/models/Permission.php#L869 `participantpanel` permission ?

Shnoulle (approved):
Great to add Permission check !

application/controllers/admin/ParticipantsAction.php
Shnoulle: OK : settings !

PR #3578, opened 2023-10-26
Fixed issue [security] #19169: Privilege escalation to edit survey

Shnoulle:
Dev: Permission check must be done before updating survey


Shnoulle:
Exactly same fix than https://github.com/LimeSurvey/LimeSurvey/pull/3547

PR #3581, opened 2023-10-27
Fixed issue #19203: alphabetical order didn't respect survey language

Shnoulle:
Dev: use and send QuestionBaseRenderer->sLanguage
Dev: follow it until needed in Question

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -781,7 +781,7 @@ class Question extends LSActiveRecord
 
         // Sort alphabetically if applicable
         if ($this->shouldOrderAnswersAlphabetically()) {
-            if (is_null($language)) {
+            if (empty($language) || !in_array($language, $this->survey->allLanguages)) {
                 $language = $this->survey->language;
             }
             foreach ($answerOptions as $scaleId => $scaleArray) {
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -726,7 +726,13 @@ class Question extends LSActiveRecord
         return $buttons;
     }
 
-    public function getOrderedAnswers($scale_id = null)
+    /**
+     * get the ordered answers
+     * @param null|integer scale
+     * @param null|string $language
+     * @return array
+     */
+    public function getOrderedAnswers($scale_id = null, $language = null)
     {
         //reset answers set prior to this call
         $aAnswerOptions = [
@@ -745,7 +751,7 @@ class Question extends LSActiveRecord
             return $aAnswerOptions[$scale_id];
         }
 
-        $aAnswerOptions = $this->sortAnswerOptions($aAnswerOptions);
+        $aAnswerOptions = $this->sortAnswerOptions($aAnswerOptions, $language);
         return $aAnswerOptions;
     }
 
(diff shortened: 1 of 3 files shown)


application/models/Question.php
```diff
@@ -774,15 +781,17 @@ private function sortAnswerOptions($answerOptions)
 
         // Sort alphabetically if applicable
         if ($this->shouldOrderAnswersAlphabetically()) {
+            if (is_null($language)) {
```
Shnoulle, line 784: did we need to check if language exist in survey->getAllLanguade ? Think it's already done in QuestionBaseRenderer ?

application/models/Question.php
```diff
@@ -774,15 +781,17 @@ private function sortAnswerOptions($answerOptions)
...
+                $language = $this->survey->language;
+            }
             foreach ($answerOptions as $scaleId => $scaleArray) {
                 $sorted = array();
 
-                // We create an aray sorted that will use the answer in the current language as key, and that will store its old index as value
+                // We create an array sorted that will use the answer in the current language as key, and that will store its old index as value
                 foreach ($scaleArray as $key => $answer) {
-                    $sorted[$answer->answerl10ns[$this->survey->language]->answer] = $key;
+                    $sorted[$answer->answerl10ns[$language]->answer] = $key;
```
Shnoulle, line 792: Check if isset ?

I can happen after import TSV or have specific issue …

application/models/Question.php
gabrieljenik: Maybe double checking here wouldn't be the worst as to make things more atomic.
Just checking if the first answer is set would be enough, right?

application/models/Question.php
gabrieljenik: I think this s covered on the comments above.

gabrieljenik (approved):
Code structure looks OK.

application/models/Question.php
Shnoulle: Ok

application/models/Question.php
Shnoulle: > I think this s covered on the comments above.

Not : _I can happen after import TSV or have specific issue …_ 

If you can not do the database repair : it happen a lot … but no issue with debug = 0

gabrieljenik: Sure. Let's check also then. It doesn't hurt.

Shnoulle: I check : seems OK with broken language to

Shnoulle: It break before

application/models/Question.php
```diff
@@ -781,7 +781,7 @@ private function sortAnswerOptions($answerOptions, $language = null)
 
         // Sort alphabetically if applicable
         if ($this->shouldOrderAnswersAlphabetically()) {
-            if (is_null($language)) {
+            if (empty($language) || !in_array($language, $this->survey->allLanguages)) {
```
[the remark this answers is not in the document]
Shnoulle: $language used ONLY for Alphabetically order.

The i'm sure it muts be tested only of needed

PR #3584, opened 2023-10-30
Fixed issue #19162: File Upload causes XSS in Central participant management/Import file CSV

gabrieljenik:


application/views/admin/participants/uploadSummary_view.php
```diff
@@ -55,15 +55,15 @@
             $warningSummary .= "<li style='width: 400px'>" . sprintf(gT("%s records with invalid email address removed"), count($invalidemaillist));
             $warningSummary .= "<div class='badtokenlist' id='invalidemaillist'><ul>";
             foreach ($invalidemaillist as $data) {
-                $warningSummary .= "<li>" . $data . "</li>";
+                $warningSummary .= "<li>" . CHtml::encode($data) . "</li>";
```
Shnoulle, line 58: No way to fix the data sent ?

Inside the controller ?

Shnoulle: I mean `sprintf(gT("sentence %s"), CHtml::encode($data))` for example.

Then allow `sprintf(gT("sentence '%s'"), "<em>" . CHtml::encode($data)) ."</em>")` for example.

gabrieljenik: > No way to fix the data sent ?
> Inside the controller ?

When fields are not filtered by rules, we always fix this things on the view, right?

> I mean `sprintf(gT("sentence %s"), CHtml::encode($data))` for example.
> 
> Then allow `sprintf(gT("sentence '%s'"), "\<em\>" . CHtml::encode($data)) ."\</em\>")` for example.

Sorry, waht do you mean?

Shnoulle: > > No way to fix the data sent ?
> > Inside the controller ?
> 
> When fields are not filtered by rules, we always fix this things on the view, right?

It's better but here `$data` didn't come directly from user value, came from controller.

> Sorry, waht do you mean?

must be encoded here : https://github.com/LimeSurvey/LimeSurvey/blob/9334cd43bc6bab0063e6e16d31e5a8d535af6dcf/application/controllers/admin/ParticipantsAction.php#L1074

**it's already a view part.**

OR 
`
$invalidemaillist[] = $line;
`

and do the view job in view …

Shnoulle (approved):

Shnoulle:
I think i report another issue about all this part : sometimes we have email, but email was not show :)

Shnoulle (approved):

PR #3587, opened 2023-10-31
Fixed issue [security] #19133: User can send email to group members, while not having permissions

Shnoulle:
Dev: only $userGroup->owner_id (and superadmin) can send email Usergroup

### DIFF
--- a/application/controllers/UserGroupController.php
+++ b/application/controllers/UserGroupController.php
@@ -101,10 +101,6 @@ class UserGroupController extends LSBaseController
      */
     public function actionViewGroup($ugid, bool $header = false)
     {
-        if (!Permission::model()->hasGlobalPermission('usergroups', 'read')) {
-            throw new CHttpException(403);
-        }
-
         $ugid = (int)$ugid;
         if (empty($ugid)) {
             throw new CHttpException(400, gT('GroupId missing'));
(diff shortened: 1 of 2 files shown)


application/controllers/UserGroupController.php
```diff
@@ -467,37 +460,36 @@ public function actionMailToAllUsersInGroup(int $ugid)
...
-                    $this->redirect(array('userGroup/viewGroup/ugid/' . $ugid));
-                }
-            } else {
-                Yii::app()->user->setFlash('error', gT("You do not have permission to send emails to all users. "));
-                $this->redirect(array('userGroup/viewGroup/ugid/' . $ugid));
-            }
-        } else {
-            $aData['ugid'] = $ugid;
+            $this->redirect($redirectUrl);
+            App()->end(); // redirect end : add it here for clarity
```
gabrieljenik, line 492: `App()->end()` or `return` ?

application/controllers/UserGroupController.php
```diff
@@ -467,37 +460,36 @@ public function actionMailToAllUsersInGroup(int $ugid)
     {
         $ugid = sanitize_int($ugid);
         $action = Yii::app()->request->getPost("action");
+        $currentUserId = App()->getCurrentUserId();
+        $userGroup = UserGroup::model()->findByPk($ugid);
+        if (empty($userGroup)) {
+            throw new CHttpException(404, gT("User group not found."));
+        }
+        if (
+            !Permission::model()->hasGlobalPermission('superadmin', 'read') // User is superadmin
```
gabrieljenik, line 469: I would change the comments: user is not superadmin and is not owner.

application/controllers/UserGroupController.php
```diff
@@ -158,8 +150,6 @@ public function actionViewGroup($ugid, bool $header = false)
             $aNewUserListData = CHtml::listData($aUsers, 'uid', function ($user) {
                 return \CHtml::encode($user->users_name) . " (" . \CHtml::encode($user->full_name) . ')';
             });
-            // Remove group owner because an owner is automatically member of a group
```
gabrieljenik, line 161: The comment is being removed. Doesnt apply anymore?
If so, the unset below, should til be there?

gabrieljenik (approved):
Code structure looks OK. A few comments to review but could be nothing

application/controllers/UserGroupController.php
Shnoulle: Redirect already App->end, advice take from a comment of @mfavetti

application/controllers/UserGroupController.php
Shnoulle: You're right … not related to this commit.

OIn 6X : we can remove the owner form user in group, not in 5X
Unsure we need to remove it

Shnoulle:
5X version https://github.com/LimeSurvey/LimeSurvey/pull/3586

PR #3588, opened 2023-10-31
Fixed issue #19139: CSRF in Reset Survey menus

gabrieljenik:
Enhanced the noCSRFValidation regex
Added automatic test


Shnoulle (approved):

Shnoulle:
The CSRF is due to bad tet here ?
Woh !

gabrieljenik:
Yes, we foudn some situations which were suspicious. 
Dig deeper as detectives and got to this :)

PR #3594, opened 2023-11-02
QE-513: tempId for newly created objects

twilligls:
Dev:


application/libraries/Api/Command/V1/SurveyPatch/OpHandlerQuestionCreate.php
```diff
@@ -81,29 +81,30 @@ public function canHandle(OpInterface $op): bool
      *                     "mandatory": false,
      *                     "relevance": "1",
      *                     "encrypted": false,
-     *                     "save_as_default": false
+     *                     "save_as_default": false,
```
kevin-foster-uk, line 84: This should be camel case.

application/libraries/Api/Command/V1/SurveyPatch/OpHandlerQuestionCreate.php
```diff
@@ -188,25 +193,59 @@ public function canHandle(OpInterface $op): bool
...
-        );
+        $transformedProps = $this->prepareData($op);
+        if (
+            is_array($transformedProps) &&
+            array_key_exists(
+                'question',
+                $transformedProps
+            )
+        ) {
+            $tempId = $this->extractTempId($transformedProps['question']);
```
kevin-foster-uk, line 214: You can avoid this nesting by checking for exception status first:

```
if ($something) {
    throw new Exception('Bla');
}

// rest of code here - no "else" statement needed
```

application/libraries/Api/Command/V1/SurveyPatch/OpHandlerQuestionTrait.php
```diff
@@ -288,4 +293,45 @@ private function getScaleIdFromData(array $questionData)
...
+                ) {
+                    $tempIds[$subQueData[$title]] = $subQueData['tempId'];
+                }
+            }
+        }
+        if (count($tempIds) > 0) {
+            $question->refresh();
+            foreach ($question->$object as $subquestion) {
+                if (array_key_exists($subquestion->$title, $tempIds)) {
+                    $mapping[$object . 'Map'][] = [
```
kevin-foster-uk, line 328: This value in this array could be an object. Maybe create a class with public properties. An objects public properties are represented in the JSON output:

```new TempIdMapItem('someTempId123', 'newRealId', 'fieldName')```

The benefit of this is that the next developer that needs to add a temp id mapping doesn't need to know the data structure to return. They just create one of these objects. Also its less likely that a make will be made in the data structure.

``` 
class TempIdMapItem
{
    public $tempId = null;
    public $id = null;
    public $field = 'id';
    
    public function __construct($tempId , $id, $field = 'id')
    {
        $this->tempId = $tempId;
        $this->id = $id;
        $this->field = $field;
    }
}
```

File: application/libraries/ObjectPatch/Patcher.php [hunk 4]
```diff
@@ -80,5 +84,65 @@ private function handleOp(OpInterface $op): void
...
+    }
+
+    /**
+     * Moves mappings of tempIds and their real ids of same object types
+     * together for a more structured return to the client.
+     * @param array $returnedData
+     * @param int $operationsApplied
+     * @return array
+     */
+    private function reorganizeReturnedData(
```
kevin-foster-uk, line 97: This method refers to concepts which are specific to our use of the patcher in PatcherSurvey (question, subquestion, answer) but this Patcher class is meant to be generic. Either this functionality belongs in PatcherSurvey or it should be modified so that the specific use case are passed in as an argument. For example you can pass in `$knownMaps` as a property.

PR #3595, opened 2023-11-03
Fixed issue #18350: activate_survey does not correctly set inherited …

mfavetti:
…properties for Notification and data management

Fixed issue #18350: activate_survey does not correctly set inherited properties for Notification and data management

master

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -474,45 +474,60 @@ class remotecontrol_handle
      * @access public
      * @param string $sSessionKey Auth credentials
      * @param int $iSurveyID ID of the Survey to be activated
+     * @param array $activationSettings (optional) survey activation settings to change prior to activation, format is array of settingName => settingValue pairs (default: [])
      * @return array in case of success result of the activation
      */
-    public function activate_survey($sSessionKey, $iSurveyID)
+    public function activate_survey($sSessionKey, $iSurveyID, $userActivationSettings = [])
     {
-        if ($this->_checkSessionKey($sSessionKey)) {
-            $iSurveyID = (int) $iSurveyID;
-            $oSurvey = Survey::model()->findByPk($iSurveyID);
-            if (is_null($oSurvey)) {
-                return array('status' => 'Error: Invalid survey ID');
-            }
-            if ($oSurvey->isActive) {
-                return array('status' => 'Error: Survey already active');
-            }
-            // Check consistency for groups and questions
-            Yii::app()->loadHelper('admin/activate');
-            $checkHasGroup = checkHasGroup($iSurveyID);
-            $checkGroup = checkGroup($iSurveyID);
-
-            if ($checkHasGroup !== false || $checkGroup !== false) {
-                return array('status' => 'Error: Survey does not pass consistency check');
-            }
-
-            $surveyActivator = new SurveyActivator($oSurvey);
-
-
-            if (Permission::model()->hasSurveyPermission($iSurveyID, 'surveyactivation', 'update')) {
-                $aActivateResults = $surveyActivator->activate();
-
-                if (isset($aActivateResults['error'])) {
-                    return array('status' => 'Error: ' . $aActivateResults['error']);
-                } else {
-                    return $aActivateResults;
-                }
-            } else {
-                            return array('status' => 'No permission');
-            }
-        } else {
-                    return array('status' => self::INVALID_SESSION_KEY);
+        // check provided session key and setup user session
+        if (!$this->_checkSessionKey($sSessionKey)) {
+            return array('status' => self::INVALID_SESSION_KEY);
         }
+        $iSurveyID = (int) $iSurveyID;
+        // check valid survey id and current user has permission
+        if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveyactivation', 'update')) {
+            return array('status' => 'No permission');
+        }
+        $oSurvey = Survey::model()->findByPk($iSurveyID);
+
+        // check if survey is already activated
+        if ($oSurvey->isActive) {
+            return array('status' => 'Error: Survey already active');
+        }
+        // Check consistency for groups and questions
+        Yii::app()->loadHelper('admin/activate');
+        $checkHasGroup = checkHasGroup($iSurveyID);
+        $checkGroup = checkGroup($iSurveyID);
+        if ($checkHasGroup !== false || $checkGroup !== false) {
+            return array('status' => 'Error: Survey does not pass consistency check');
+        }
+        $surveyActivator = new SurveyActivator($oSurvey);
+        // list of activation related survey settings and allowed values
+        $activationSettingNames = [
+            'anonymized',
+            'datestamp',
+            'ipaddr',
+            'ipanonymize',
+            'refurl',
+            'savetimings',
+        ];
+        // update survey activation settings
+        foreach ($activationSettingNames as $activationSettingName) {
+            // use inherited value
+            $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName};
+            // override inherited value with user input
+            if (isset($userActivationSettings[$activationSettingName]) && in_array($userActivationSettings[$activationSettingName], ['Y', 'N'])) {
+                $oSurvey->{$activationSettingName} = $userActivationSettings[$activationSettingName];
+            }
+        }
+        // persist changes
+        $oSurvey->save();
+        // attempt to activate the survey
+        $aActivateResults = $surveyActivator->activate();
+        if (isset($aActivateResults['error'])) {
+            return array('status' => 'Error: ' . $aActivateResults['error']);
+        }
+        return $aActivateResults;
     }
 
     /**


Shnoulle (commented):
My opinion, it's the bad way to do.

In SurveyActivator : https://github.com/LimeSurvey/LimeSurvey/blob/ea1c4ba3b12dd817bc820c6239b72e4d476ccd7d/application/models/SurveyActivator.php#L212 if IpAddress is OK,

I don't see where we get the timestamp.

Here : you update the model value, but it must stay inherit and use the option.

mfavetti:
> My opinion, it's the bad way to do.
> 
> In SurveyActivator :
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/ea1c4ba3b12dd817bc820c6239b72e4d476ccd7d/application/models/SurveyActivator.php#L212
> 
> if IpAddress is OK,
> I don't see where we get the timestamp.
> 
> Here : you update the model value, but it must stay inherit and use the option.

The problem is the survey taking part of the application seems to expect a value of "Y" or "N" for the activation survey settings. A setting of "I" does not result in the inherited value being used. If a survey is activated with one of these settings set to "I", the respondent can take the survey and submit it without error. However their responses are never actually recorded at all, resulting in data loss. The activation button and corresponding dialog in the admin interface change the survey settings to be either the result of merging the survey settings/survey group settings/global survey settings, or the user input. But they are no longer "I".

My changes just make the corresponding RPC call function the same way. First use the merged settings, with the possibility to override with user input. Update the survey settings, then activate. This also results in "Y" or "N" instead of "I".

Shnoulle:
> This also results in "Y" or "N" instead of "I".

Oh yes i got it : LSA can not accept I here … 

Why not a more simple function then ?
```
        /* Remove I from response table definition */
        $activationSettingNames = [
            'anonymized',
            'datestamp',
            'ipaddr',
            'ipanonymize',
            'refurl',
            'savetimings',
        ];
        foreach ($activationSettingNames as $activationSettingName) {
            $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName}
        }
        $oSurvey->save();
        /* OK continue */
```

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -474,45 +474,70 @@ public function set_survey_properties($sSessionKey, $iSurveyID, $aSurveyData)
...
+                $activationSettingValue = $activationSettings[$activationSettingName];
+                if (in_array($activationSettingValue, $allowedSettingValues)) {
+                    $oSurvey->{$activationSettingName} = $activationSettingValue;
+                }
+            }
+        }
+        // persist changes
+        $oSurvey->save();
+        // Make sure the saved values will be picked up
+        Survey::model()->resetCache();
```
Shnoulle, line 533: Not needed : $oSurvey->anonymized is already set in current state here. And we don't put result in cache : we only put table definition.

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -474,45 +474,70 @@ public function set_survey_properties($sSessionKey, $iSurveyID, $aSurveyData)
...
+        $oSurvey->setOptions();
```
Shnoulle, line 534: To be move before

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -474,45 +474,70 @@ public function set_survey_properties($sSessionKey, $iSurveyID, $aSurveyData)
...
-            }
-        } else {
-                    return array('status' => self::INVALID_SESSION_KEY);
+        // check provided session key and setup user session
+        if (!$this->_checkSessionKey($sSessionKey)) {
+            return array('status' => self::INVALID_SESSION_KEY);
+        }
+        $iSurveyID = (int) $iSurveyID;
+        // check valid survey id and current user has permission
+        if (!Permission::model()->hasSurveyPermission($iSurveyID, 'surveyactivation', 'update')) {
```
Shnoulle, line 488: 👍

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -474,45 +474,70 @@ public function set_survey_properties($sSessionKey, $iSurveyID, $aSurveyData)
...
+        // Check consistency for groups and questions
+        Yii::app()->loadHelper('admin/activate');
+        $checkHasGroup = checkHasGroup($iSurveyID);
+        $checkGroup = checkGroup($iSurveyID);
+        if ($checkHasGroup !== false || $checkGroup !== false) {
+            return array('status' => 'Error: Survey does not pass consistency check');
+        }
+        $surveyActivator = new SurveyActivator($oSurvey);
+        // grab computed survey settings from common helper
+        $surveysettings = getSurveyInfo($iSurveyID);
```
Shnoulle, line 505: My opinion : we must remove usage of getSurveyInfo if we can avoid : it create a big data with a lot of unused value.

mfavetti:
> > This also results in "Y" or "N" instead of "I".
> 
> Oh yes i got it : LSA can not accept I here …
> 
> Why not a more simple function then ?
> 
> ```
>         /* Remove I from response table definition */
>         $activationSettingNames = [
>             'anonymized',
>             'datestamp',
>             'ipaddr',
>             'ipanonymize',
>             'refurl',
>             'savetimings',
>         ];
>         foreach ($activationSettingNames as $activationSettingName) {
>             $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName}
>         }
>         $oSurvey->save();
>         /* OK continue */
> ```

Yes, I can use oOptions attribute instead of 


> > This also results in "Y" or "N" instead of "I".
> 
> Oh yes i got it : LSA can not accept I here …
> 
> Why not a more simple function then ?
> 
> ```
>         /* Remove I from response table definition */
>         $activationSettingNames = [
>             'anonymized',
>             'datestamp',
>             'ipaddr',
>             'ipanonymize',
>             'refurl',
>             'savetimings',
>         ];
>         foreach ($activationSettingNames as $activationSettingName) {
>             $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName}
>         }
>         $oSurvey->save();
>         /* OK continue */
> ```

Yes, I can updated to use oOptions instead of the getSurveyInfo from common_helper

> > This also results in "Y" or "N" instead of "I".
> 
> Oh yes i got it : LSA can not accept I here …
> 
> Why not a more simple function then ?
> 
> ```
>         /* Remove I from response table definition */
>         $activationSettingNames = [
>             'anonymized',
>             'datestamp',
>             'ipaddr',
>             'ipanonymize',
>             'refurl',
>             'savetimings',
>         ];
>         foreach ($activationSettingNames as $activationSettingName) {
>             $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName}
>         }
>         $oSurvey->save();
>         /* OK continue */
> ```

Yes, I can updated to use oOptions instead of the getSurveyInfo from common_helper.

But we need to keep the second foreach to handle the user input being able to override the merged settings.

mfavetti:
> > This also results in "Y" or "N" instead of "I".
> 
> Oh yes i got it : LSA can not accept I here …
> 
> Why not a more simple function then ?
> 
> ```
>         /* Remove I from response table definition */
>         $activationSettingNames = [
>             'anonymized',
>             'datestamp',
>             'ipaddr',
>             'ipanonymize',
>             'refurl',
>             'savetimings',
>         ];
>         foreach ($activationSettingNames as $activationSettingName) {
>             $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName}
>         }
>         $oSurvey->save();
>         /* OK continue */
> ```

Yes, I can updated to use oOptions instead of the getSurveyInfo from common_helper.

But we need to keep the second foreach to handle the user input being able to override the merged settings.

Shnoulle:
In my opinion: it's cleaner to use Survey::model directly than the old getSurveyInfo array. More understandable.

And add purpose of update in comments.

application/helpers/remotecontrol/remotecontrol_handle.php
mfavetti: > Not needed : $oSurvey->anonymized is already set in current state here. And we don't put result in cache : we only put table definition.

This is from https://github.com/LimeSurvey/LimeSurvey/blob/ea1c4ba3b12dd817bc820c6239b72e4d476ccd7d/application/controllers/SurveyAdministrationController.php#L1870

I worked from the existing functionality from the admin activate button/dialog

application/helpers/remotecontrol/remotecontrol_handle.php
mfavetti: Yes, I can refactor to remove this

Shnoulle:
> But we need to keep the second foreach to handle the user input being able to override the merged settings.

Right !
But still in one foreach then :)

Something like this

```
        /* Fix surveyb settings used in response table definition */
        $activationSettingNames = [
            'anonymized',
            'datestamp',
            'ipaddr',
            'ipanonymize',
            'refurl',
            'savetimings',
        ];
        foreach ($activationSettingNames as $activationSettingName) {
            /* set to inherit */
            $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName};
            /* get it from parameters */
            if (isset($activationSettings[$activationSettingName]) && in_array($activationSettings[$activationSettingName], ['Y','N'])) {
                $oSurvey->{$activationSettingName} = $activationSettings[$activationSettingName];
            }
        }
        $oSurvey->save();
        /* OK continue */
```

application/helpers/remotecontrol/remotecontrol_handle.php
mfavetti: seems to work without it though, I can remove

mfavetti:
> In my opinion: it's cleaner to use Survey::model directly than the old getSurveyInfo array. More understandable.
> 
> And add purpose of update in comments.



> In my opinion: it's cleaner to use Survey::model directly than the old getSurveyInfo array. More understandable.
> 
> And add purpose of update in comments.

haha, I don't know which one is newer/older from my own experience, but the oOptions seems simpler and avoid using extra helper

mfavetti:
okay, updated based on your review

ty!

application/helpers/remotecontrol/remotecontrol_handle.php
mfavetti: I remove 532-534

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -474,45 +474,63 @@ public function set_survey_properties($sSessionKey, $iSurveyID, $aSurveyData)
...
+            'ipanonymize',
+            'refurl',
+            'savetimings',
+        ];
+        // update survey activation settings
+        foreach ($activationSettingNames as $activationSettingName) {
+            // use inherited value
+            $oSurvey->{$activationSettingName} = $oSurvey->oOptions->{$activationSettingName};
+            // override inherited value with user input
+            if (array_key_exists($activationSettingName, $userActivationSettings)) {
```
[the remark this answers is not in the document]
mfavetti: ummmmm

array_key_exists -> true if the key exists in array
isset -> true if the key exists in array and value is not null

not sure about speed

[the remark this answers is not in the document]
Shnoulle: Oh have answer : https://stackoverflow.com/a/700257

https://stackoverflow.com/a/21759158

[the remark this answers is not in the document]
Shnoulle: > ummmmm
> 
> array_key_exists -> true if the key exists in array isset -> true if the key exists in array and value is not null
> 
> not sure about speed

Here : we don't want NULL value (or empty value) and speed is really better with isset.

Globally : we must avoid usage of `array_key_exists` if we don't want NULL value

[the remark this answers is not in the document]
mfavetti: 2/100's of a second per call :-/

ok i can update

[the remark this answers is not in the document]
Shnoulle: > 2/100's of a second per call :-/
> 
> ok i can update

`0.049405097961426` VS `0.027907848358154` for 1000000 iteration on **PHP 7.1** It's 1 VS 2

PR #3599, opened 2023-11-03
Fix CSRF validation for remotecontrol API route

edgarrmondragon:
Fixes an error introduced in #3588. The changes made the remote control route, `admin/remotecontrol`, no longer skip CSRF validation.

I can report the bug in https://bugs.limesurvey.org, but this error is present only in the `master` branch and I thought it quicker to send a PR.

Fixed issue [#19220](https://bugs.limesurvey.org/view.php?id=19220)

### DIFF
--- a/application/config/internal.php
+++ b/application/config/internal.php
@@ -156,7 +156,7 @@ $internalConfig = array(
             'noCsrfValidationParams' => array(),
             'noCsrfValidationRoutes' => array(
                 'rest',
-                'remotecontrol',
+                'admin/remotecontrol',
                 'plugins/unsecure',
             ),
             'csrfCookie' => array(
--- a/application/core/LSHttpRequest.php
+++ b/application/core/LSHttpRequest.php
@@ -264,6 +264,6 @@ class LSHttpRequest extends CHttpRequest
         // For example the routes "rest" (in the case of "index.php/rest?...") or "rest/..." (in the case of
         // "index.php/rest/...") should be matched by the rule "rest", but the route "admin/menus/sa/restore"
         // should not.
-        return preg_match('#^' . $rule . '$|^' . $rule . '/#', (string) $route);
+        return preg_match('#^/?' . $rule . '$|^/?' . $rule . '/#', (string) $route);
     }
 }
--- a/tests/unit/CsrfHttpRequestTest.php
+++ b/tests/unit/CsrfHttpRequestTest.php
@@ -41,8 +41,8 @@ class CsrfHttpRequestTest extends TestBaseClass
     public function testRemoteControlRoutesSkipCsrfValidation()
     {
         $routes = array(
-            'remotecontrol/actionOnItemById/15',
-            'remotecontrol/action',
+            'admin/remotecontrol/actionOnItemById/15',
+            'admin/remotecontrol/action',
         );
 
         foreach ($routes as $route) {
--- a/tests/unit/CsrfHttpRequestTest.php
+++ b/tests/unit/CsrfHttpRequestTest.php
@@ -43,6 +43,7 @@ class CsrfHttpRequestTest extends TestBaseClass
         $routes = array(
             'admin/remotecontrol/actionOnItemById/15',
             'admin/remotecontrol/action',
+            '/admin/remotecontrol',
         );
 
         foreach ($routes as $route) {


application/config/internal.php
```diff
@@ -156,7 +156,7 @@
             'noCsrfValidationParams' => array(),
             'noCsrfValidationRoutes' => array(
                 'rest',
-                'remotecontrol',
```
edgarrmondragon, line 159: It feels like a breaking change to expect users to update this if they have overridden this in their config, so I wonder if the maintainers feel strongly about relaxing the regex in `application/core/LSHttpRequest.php` instead.

mfavetti:
i don't think this is sufficient to fix the regression, please see my note: https://bugs.limesurvey.org/view.php?id=19220

application/config/internal.php
mfavetti: surely if a user override this, they extend the array, not replace it

Shnoulle:
> i don't think this is sufficient to fix the regression, please see my note: https://bugs.limesurvey.org/view.php?id=19220

I think it's OK here, why not sufficient ?

mfavetti:
only fix get, not path routing

application/config/internal.php
Shnoulle: > It feels like a breaking change to expect users to update this if they have overridden this in their config, so I wonder if the maintainers feel strongly about relaxing the regex in `application/core/LSHttpRequest.php` instead.

The commit was done to be sure to allow only needed route :). Relaxing : less CRSF control. I think we **must** keep current regexp.

edgarrmondragon:
> only fix get, not path routing

@mfavetti Let me know if d4b0490 addresses that 🙂

mfavetti:
> > only fix get, not path routing
> 
> @mfavetti Let me know if [d4b0490](https://github.com/LimeSurvey/LimeSurvey/commit/d4b04905b4a3b974515c1e79e7629eb568f1e055) addresses that 🙂

okay remote control works now using get and path routing
i think rest didn't work before on get anyways, path worked on gabrieljenik's original code
not sure how to test plugin/unsecure, didn't work much with plugins yet (basically can someone answer if this is the actual path or is it expanded to something like admin/plugin/unsecure?

weird to note that /admin/remotecontrol -> /admin/remotecontrol but /rest/v1/session -> rest/v1/session
why is the handling of leading slash different??

if it was the same then you would just have needed to fix remotecontrol -> admin/remotecontrol and no need to worry about the leading slash

application/config/internal.php
mfavetti: yeah I agree, it needs to stay as the more restrictive regexp

the issue before is that routes that match a portion of the routes exempted from csrf checking would also be exempt

ie 
1. rest/v1/session
2. admin/menus/sa/restore

both are exempted because they contain "rest" anywhere

this is definitely not desirable for security

Shnoulle:
> only fix get, not path routing



> only fix get, not path routing

It must be done on route, no dependant on path or get , then fix this part : https://github.com/LimeSurvey/LimeSurvey/blob/70b09896e023d919791b0b5bb8bc345eee1b98ac/application/core/LSHttpRequest.php#L158 if needed ?

Shnoulle:
> > only fix get, not path routing
> 
> @mfavetti Let me know if [d4b0490](https://github.com/LimeSurvey/LimeSurvey/commit/d4b04905b4a3b974515c1e79e7629eb568f1e055) addresses that 🙂

Strange ? Seems [parseUrl](https://www.yiiframework.com/doc/api/1.1/CUrlManager#parseUrl-detail) function is broken ?

In my opinon: route for 
`
example.org/controller/action` or `example.org/?r=controller/action` but be same (controller/action)

And route `example.org//controller/action` or `example.org//?r=controller/action`  : broken (/controller)

It's not the case ?

gabrieljenik (approved):
I think it is OK like this.
I would feel better if there would be a full RC test end to end that connects to the server, which is missing. There are functional tests that connect to webdirver, but not tests that connect to RC endpoint

gabrieljenik:
The offedning PR was reverted.
This PR, right now, needs conflict resolution.
Maybe this should be closed and incorporated in a new PR, related to the original fix.

gabrieljenik:
I will be working on this. Probbly on a new PR with a proper name

edgarrmondragon:
@gabrieljenik gotcha. Feel free to close this.

PR #3607, opened 2023-11-08
Fixed issue : unable to update from pre 495 version

Shnoulle:


application/models/User.php
```diff
@@ -130,6 +130,12 @@ public function rules()
     /** @inheritdoc */
     public function scopes()
     {
+        if (App()->getConfig("DBVersion") < 495) {
+            /* No expires column before 495 */
+            return array(
+                'active' => []
+            );
+        }
```
ptelu, line 138: you think this will be enough? what about future updates changing the model/database structure

Shnoulle: I really don't understand ?

`active` scopes use `expires` column. Not any other column … Why care of future column. If we need to add a scope in future we update this part

Else : i upgrade from 3 connected and not connected

olleharstedt: It's also a question of the Yii database schema cache and the attributes that are fetched automatically from the database. I dunno honestly.

Shnoulle: Can you test ?

1. At start of the update : test is done with empty active array: OK, user exist
2. Update done
3. database schema cache reset
4. test is done with real active array: OK, user exist

Potential issue if DB broke during upgrade ?

Shnoulle: > Potential issue if DB broke during upgrade ?

Seems not an issue :  https://github.com/LimeSurvey/LimeSurvey/blob/92806d663b4fdad285aa14488e58918091feda51/application/helpers/update/updatedb_helper.php#L97

PR #3610, opened 2023-11-09
Fixed issue #19214: JSON RPC response content type should be applicat…

mfavetti:
…ion/json

Fixed issue #19214: JSON RPC response content type should be application/json

master

### DIFF
--- a/application/libraries/LSjsonRPCServer.php
+++ b/application/libraries/LSjsonRPCServer.php
@@ -60,7 +60,7 @@ class LSjsonRPCServer extends jsonRPCServer
         // output the response
         if (is_null($request) || !empty($request['id'])) {
 // notifications don't want response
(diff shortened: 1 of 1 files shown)


Shnoulle:
We don't have an updated library ?

Shnoulle (approved):

mfavetti:
library we use is defunct now... beyond unmaintained

we could switch to a newer php json rpc library (with the benefit of adding support for v2) but that would be new feature maybe, not bugfix?

Shnoulle:
> library we use is defunct now... beyond unmaintained
> 
> we could switch to a newer php json rpc library (with the benefit of adding support for v2) but that would be new feature maybe, not bugfix?

Yes, if there are a good one : for [develop](https://github.com/LimeSurvey/LimeSurvey/tree/develop) branch

PR #3612, opened 2023-11-10
Fixed issue #19139: CSRF in Reset Survey menus (Tests)

gabrieljenik:
Tests for remote control: Real connection through RC client

### DIFF
--- a/tests/functional/backend/RemoteControlTest.php
+++ b/tests/functional/backend/RemoteControlTest.php
@@ -38,7 +38,7 @@ class RemoteControlTest extends TestBaseClassWeb
             self::$client = new \Zend_XmlRpc_Client($serverUrl);
         } elseif ($RPCType == 'json') {
             Yii::app()->loadLibrary('jsonRPCClient');
-            self::$client = new \jsonRPCClient($serverUrl, true);
+            self::$client = new \jsonRPCClient($serverUrl);
         }
     }
 
@@ -54,8 +54,9 @@ class RemoteControlTest extends TestBaseClassWeb
 
     public function testGetSessionKey()
     {
-        $sSessionKey = self::$client->call('get_session_key', array('admin', 'password'));
+        $sessionKey = self::$client->call('get_session_key', ['admin', 'password']);
+        $this->assertIsNotArray($sessionKey);
 
-        $this->assertTrue(true);
+        self::$client->call('release_session_key', [$sessionKey]);
     }
 }
--- a/tests/functional/backend/RemoteControlTest.php
+++ b/tests/functional/backend/RemoteControlTest.php
@@ -59,4 +59,13 @@ class RemoteControlTest extends TestBaseClassWeb
 
         self::$client->call('release_session_key', [$sessionKey]);
     }
+
+    public function testCredentialsError()
+    {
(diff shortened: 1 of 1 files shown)


tests/functional/backend/RemoteControlTest.php
```diff
@@ -0,0 +1,62 @@
...
+        $urlMan = Yii::app()->urlManager;
+        $urlMan->setBaseUrl(self::$tmpBaseUrl);
+
+        \SettingGlobal::setSetting('RPCInterface', self::$tmpRPCType);
+    }
+
+    public function testGetSessionKey()
+    {
+        $sessionKey = self::$client->call('get_session_key', ['admin', 'password']);
+        $this->assertIsNotArray($sessionKey);
```
Shnoulle, line 58: [assertIsString()](https://docs.phpunit.de/en/9.6/assertions.html#assertisstring) ?

A 500 error is not an array , right ?

Shnoulle (commented):
Why not https://docs.phpunit.de/en/9.6/assertions.html#assertisstring ?

tests/functional/backend/RemoteControlTest.php
gabrieljenik: I believe a 500 would be shown as an exception. Not sure.
Should try maybe...

gabrieljenik: Very edge case. Does it worth spending on it?

Shnoulle: > Very edge case. Does it worth spending on it?

The good result **MUST** be a string : then just check if it's a string.

gabrieljenik: This is what `get_session_key` returns.
```
        if (is_string($loginResult)) {
            return array('status' => $loginResult);
        }
        return array('status' => 'Invalid user name or password');
```

gabrieljenik: Also, jsonRpcClient, if something fails, throws Exception.
I would just leave it like this.

Shnoulle: > This is what `get_session_key` returns.
> 
> ```
>         if (is_string($loginResult)) {
>             return array('status' => $loginResult);
>         }
>         return array('status' => 'Invalid user name or password');
> ```

In case of error only 

In case of success: 
https://github.com/LimeSurvey/LimeSurvey/blob/5d97f49a06b610b32db927d434de76e88e1746a1/application/helpers/remotecontrol/remotecontrol_handle.php#L60


Else : if there are a CRSF issue : it's an Exception.

gabrieljenik: You are right

Shnoulle (approved):
PS : maybe check if it's an array with a bad password before ?

gabrieljenik:
> PS : maybe check if it's an array with a bad password before ?

Added

Shnoulle:
> > PS : maybe check if it's an array with a bad password before ?
> 
> Added

How, it's just an idea like this :)

Not mandatory

Shnoulle (approved):

PR #3614, opened 2023-11-10
Fixed issue #19232: Missing cssframework default values for survey th…

olleharstedt:


Shnoulle:
In 3.X : the default was boostrap.css, empty mean boostrap.css, then variation “replace” boostrap.css

olleharstedt:
> In 3.X : the default was boostrap.css, empty mean boostrap.css, then variation “replace” boostrap.css

Problem is, default is not "empty" now. If you reset the fruity theme, you get a JSON object saved instead of a simple string, and survey export breaks... :no_mouth:

Shnoulle:
Related to (array) json_decode ?

olleharstedt:
Not completely sure :stuck_out_tongue_closed_eyes:

Shnoulle:
Else: in 5 : saved in cssframework_css `{"replace":[["css/bootstrap.css","css/variations/slate.min.css"]]}` but set as flatly.min.css whene reset (don't know how …)

Shnoulle (approved):

PR #3617, opened 2023-11-14
Fixed issue #19228: Setting Bruteforce timeout values to empty string causes the administrator to be locked out

Shnoulle:
Dev: update HTML and put some information and restriction
Dev: check final value as integre or float


tiborpacalat (approved):

Shnoulle:
@tiborpacalat sorry for 1 issue … fixed now …

PR #3618, opened 2023-11-14
Fixed issue #19228: Setting Bruteforce timeout values to empty string causes the administrator to be locked out

Shnoulle:
Dev: update HTML and put some information and restriction
Dev: check final value as integer or float

### DIFF
--- a/application/models/FailedLoginAttempt.php
+++ b/application/models/FailedLoginAttempt.php
@@ -91,17 +91,20 @@ class FailedLoginAttempt extends LSActiveRecord
 
         switch ($attemptType) {
             case FailedLoginAttempt::TYPE_LOGIN:
-                $timeOut = Yii::app()->getConfig('timeOutTime');
-                $maxLoginAttempt = Yii::app()->getConfig('maxLoginAttempt');
+                $timeOut = floatval(App()->getConfig('timeOutTime'));
+                $maxLoginAttempt = intval(App()->getConfig('maxLoginAttempt'));
                 break;
             case FailedLoginAttempt::TYPE_TOKEN:
-                $timeOut = Yii::app()->getConfig('timeOutParticipants');
-                $maxLoginAttempt = Yii::app()->getConfig('maxLoginAttemptParticipants');
+                $timeOut = floatval(App()->getConfig('timeOutParticipants'));
+                $maxLoginAttempt = intval(App()->getConfig('maxLoginAttemptParticipants'));
                 break;
             default:
                 throw new InvalidArgumentException(sprintf("Invalid attempt type: %s", $attemptType));
         }
-
+        // Return false if disable my maxLoginAttempt
+        if ($maxLoginAttempt <= 0) {
+            return false;
+        }
         if (Yii::app()->getConfig('DBVersion') <= 480) {
             $criteria = new CDbCriteria();
             $criteria->condition = 'number_attempts >= :attempts AND ip = :ip';
@@ -124,6 +127,7 @@ class FailedLoginAttempt extends LSActiveRecord
         if ($row != null) {
             $lastattempt = strtotime((string) $row->last_attempt);
             if (time() > $lastattempt + $timeOut) {
+                // always true if $timeOut <= 0
                 $this->deleteAttempts($attemptType);
             } else {
                 $isLockedOut = true;
--- a/application/models/FailedLoginAttempt.php
+++ b/application/models/FailedLoginAttempt.php
@@ -91,11 +91,11 @@ class FailedLoginAttempt extends LSActiveRecord
 
         switch ($attemptType) {
             case FailedLoginAttempt::TYPE_LOGIN:
-                $timeOut = floatval(App()->getConfig('timeOutTime'));
+                $timeOut = intval(App()->getConfig('timeOutTime'));
                 $maxLoginAttempt = intval(App()->getConfig('maxLoginAttempt'));
                 break;
             case FailedLoginAttempt::TYPE_TOKEN:
-                $timeOut = floatval(App()->getConfig('timeOutParticipants'));
+                $timeOut = intval(App()->getConfig('timeOutParticipants'));
                 $maxLoginAttempt = intval(App()->getConfig('maxLoginAttemptParticipants'));
                 break;
             default:
--- a/application/views/admin/globalsettings/_security.php
+++ b/application/views/admin/globalsettings/_security.php
@@ -155,7 +155,7 @@
                         <?php eT("Maximum number of attempts:"); ?>
                     </label>
                     <div class="">
-                        <input class="form-control" type="number" min="1" pattern="^\d*$" name="maxLoginAttempt" placeholder="<?= gT("Disabled") ?>"
+                        <input class="form-control" type="number" min="1" step="1" pattern="^\d*$" name="maxLoginAttempt" placeholder="<?= gT("Disabled") ?>"
                             value="<?= App()->getConfig('maxLoginAttempt') !== "" ?? intval(App()->getConfig('maxLoginAttempt')) ?>"
                         />
                         <div class="form-text"><?= gT("Set an empty value disable brute force protection. Number of attemps are deleted each time.") ?></div>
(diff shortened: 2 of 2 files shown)


application/models/FailedLoginAttempt.php
```diff
@@ -91,17 +91,20 @@ public function isLockedOut(string $attemptType): bool
 
         switch ($attemptType) {
             case FailedLoginAttempt::TYPE_LOGIN:
-                $timeOut = Yii::app()->getConfig('timeOutTime');
-                $maxLoginAttempt = Yii::app()->getConfig('maxLoginAttempt');
+                $timeOut = floatval(App()->getConfig('timeOutTime'));
```
gabrieljenik, line 94: Float? Why not int?

Shnoulle: Why int ? why not 500ms ;) (but right : not really work since we use time)

Shnoulle: PS : i can do it work : 0.5 or 2.650 if needed. Not complex.

gabrieljenik: The amount already has a measuring unit, like seconds or minutes. Usually that is expressed with an int.
I don't see a reason for adding more complexity to this.

Shnoulle: OK, i update

application/models/FailedLoginAttempt.php
```diff
@@ -91,17 +91,20 @@ public function isLockedOut(string $attemptType): bool
...
-                $maxLoginAttempt = Yii::app()->getConfig('maxLoginAttemptParticipants');
+                $timeOut = intval(App()->getConfig('timeOutParticipants'));
+                $maxLoginAttempt = intval(App()->getConfig('maxLoginAttemptParticipants'));
                 break;
             default:
                 throw new InvalidArgumentException(sprintf("Invalid attempt type: %s", $attemptType));
         }
-
+        // Return false if disable my maxLoginAttempt
+        if ($maxLoginAttempt <= 0) {
```
gabrieljenik, line 105: Are you sure about this? This seems like new behaviour.
Can this, by accident, make this feature to be not active?
Maybe set a default fallback value of 5? (Yes, arbitrary)

application/views/admin/globalsettings/_security.php
```diff
@@ -147,23 +147,29 @@
...
                 <div class="mb-3">
                     <label class="form-label" for='timeOutTime'>
                         <?php eT("Lockout time in seconds (after maximum number of attempts):"); ?>
                     </label>
                     <div class="">
-                        <input class="form-control" type="number" min="0" name="timeOutTime" value="<?= Yii::app()->getConfig('timeOutTime') ?>" />
+                        <input class="form-control" type="number" min="0" step="1" pattern="^\d*$" name="timeOutTime" placeholder="<?= gT("Disabled") ?>"
+                            value="<?= App()->getConfig('timeOutTime') !== "" ?? intval(App()->getConfig('timeOutTime')) ?>"
+                        />
+                        <div class="form-text"><?= gT("Set an empty value or 0 disable brute force protection. Number of attemps are deleted each time.") ?></div>
```
gabrieljenik, line 172: I am not sure about this. If someone wants to disable it, they can set 999.
I would say it could lead to accidents

gabrieljenik: @olleharstedt What do you think? @tiborpacalat ?

application/views/admin/globalsettings/_security.php
```diff
@@ -147,23 +147,29 @@
...
                 <div class="mb-3">
                     <label class="form-label" for='maxLoginAttempt'>
                         <?php eT("Maximum number of attempts:"); ?>
                     </label>
                     <div class="">
-                        <input class="form-control" type="number" min="0" name="maxLoginAttempt" value="<?= Yii::app()->getConfig('maxLoginAttempt') ?>" />
+                        <input class="form-control" type="number" min="1" step="1" pattern="^\d*$" name="maxLoginAttempt" placeholder="<?= gT("Disabled") ?>"
+                            value="<?= App()->getConfig('maxLoginAttempt') !== "" ?? intval(App()->getConfig('maxLoginAttempt')) ?>"
+                        />
+                        <div class="form-text"><?= gT("Set an empty value disable brute force protection. Number of attemps are deleted each time.") ?></div>
```
tiborpacalat, line 161: Please write this sentence like this:
"Set an empty value to disable brute force protection. Number of attempts are deleted each time."

application/views/admin/globalsettings/_security.php
tiborpacalat, line 172: Please write this sentence like this:
"Set an empty value or 0 to disable brute force protection. Number of attempts are deleted each time."

tiborpacalat: I think empty or zero are logical choices here.

application/views/admin/globalsettings/_security.php
```diff
@@ -188,15 +194,21 @@
                     <?php eT("Maximum number of attempts:"); ?>
                 </label>
                 <div class="">
-                    <input class="form-control" min="0" type="number" name="maxLoginAttemptParticipants" value="<?= Yii::app()->getConfig('maxLoginAttemptParticipants') ?>" />
+                    <input class="form-control" type="number" min="1" step="1" pattern="^\d*$" name="maxLoginAttemptParticipants" placeholder="<?= gT("Disabled") ?>"
+                        value="<?= App()->getConfig('maxLoginAttemptParticipants') !== "" ??  intval(App()->getConfig('maxLoginAttemptParticipants')) ?>"
+                    />
+                    <div class="form-text"><?= gT("Set an empty value disable brute force protection. Number of attemps are never checked.") ?></div>
```
tiborpacalat, line 200: Please write this sentence like this:
"Set an empty value to disable brute force protection. Number of attempts are never checked."

application/views/admin/globalsettings/_security.php
```diff
@@ -188,15 +194,21 @@
...
             <div class="mb-3">
                 <label class="form-label" for='timeOutParticipants'>
                     <?php eT("Lockout time in seconds (after maximum number of attempts):"); ?>
                 </label>
                 <div class="">
-                    <input class="form-control" type="number" min="0" name="timeOutParticipants" value="<?= Yii::app()->getConfig('timeOutParticipants') ?>" />
+                    <input class="form-control" type="number" min="0" step="1" pattern="^\d*$" name="timeOutParticipants" placeholder="<?= gT("Disabled") ?>"
+                        value="<?= App()->getConfig('timeOutParticipants') !== "" ?? intval(App()->getConfig('timeOutParticipants')) ?>"
+                    />
+                    <div class="form-text"><?= gT("Set an empty value or 0 disable brute force protection. Number of attemps are deleted each time.") ?></div>
```
tiborpacalat, line 211: Please write this sentence like this:
"Set an empty value or 0 to disable brute force protection. Number of attempts are deleted each time."

tiborpacalat (changes requested):

gabrieljenik (approved):
Code structure looks OK.
We are introducing the possibilty of disabling the feature.

Shnoulle:
> We are introducing the possibilty of disabling the feature.

It exist before : you can set 0 and it work. 

Not a new feature :). Just more clear.

application/views/admin/globalsettings/_security.php
Shnoulle: > I think empty or zero are logical choices here.

It's tghe case here : allowing empty and 0 : «Number of attemps are deleted each time.»

application/models/FailedLoginAttempt.php
Shnoulle: >  Can this, by accident, make this feature to be not active?

If someone delete before the value here : he clearly want to disable it.

>  Maybe set a default fallback value of 5? (Yes, arbitrary)

No, never : 
1. I really don't want to disable 1 : it's the **best** solution for token (with 1 second)
2. allow user to disable BUT register the save the number of attempts can be really interesting.

application/views/admin/globalsettings/_security.php
Shnoulle: Oups … there are an issue in this sentence see https://github.com/LimeSurvey/LimeSurvey/pull/3617/files#diff-9277e62f270b5ccca6372e2d60aaaf7b1e48aded7cacd6536477f62e985186e1R141

It must be _"Set an empty value disable brute force protection. Number of attempts are never checked."_  both disable the system but not on the same way.

Shnoulle: _Set an empty value disable brute force protection. Number of attempts are never checked._

application/views/admin/globalsettings/_security.php
Shnoulle: _Set an empty value disable brute force protection. Number of attempts are never checked._

Shnoulle: Fixed by _Set an empty value to disable brute force protection.  Number of attempts are never checked._

application/views/admin/globalsettings/_security.php
Shnoulle: Fixed by _Set an empty value to disable brute force protection.  Number of attempts are never checked._

tiborpacalat (approved):

Shnoulle:
@tiborpacalat sorry for 1 issue … fixed now …

PR #3619, opened 2023-11-14
Fixed issue #19206: alphabetical order can be bad with utf8 caracter

Shnoulle:
Fixed issue #19208: alphabetical order delete answers
Dev: use Collator for ordering, check if exist
Dev: use collator_asort or asort to keep keys

### DIFF
--- a/application/helpers/SortHelper.php
+++ b/application/helpers/SortHelper.php
@@ -1,0 +1,111 @@
+<?php
+
+/**
+* This file is part of LimeSurvey
+* Offer some helper for sorting
+*/
+
+namespace LimeSurvey\Helpers;
+
+use Yii;
+
+// Can not use Collator class, must work without
+
+class SortHelper
+{
+    /** @var integer flag for regular sort */
+    const SORT_REGULAR = 0;
+    /** @var integer flag for numeric sort */
+    const SORT_NUMERIC = 1;
+    /** @var integer flag for string sort */
+    const SORT_STRING = 2;
+
+    /**
+     * @var self Singleton
+     */
+    private static $instance = null;
+
+    /**
+     * @var string current language
+     */
+    private static $language = null;
+    /**
+     * @var \Collator if possible
+     */
+    private static $collator = null;
+
+    /**
+     * Set the language when construct
+     */
+    private function __construct($language)
+    {
+        self::$language = $language;
+        if (class_exists('Collator')) {
+            $locale = $language;
+            $languageData = getLanguageData();
+            if (isset($languageData[$language]['cldr'])) {
+                $locale = $languageData[$language]['cldr'];
+            }
+            self::$collator = new \Collator($locale);
+        } else {
+            Yii::log(
+                "For better alphabeticall ordering : you need php-intl.",
+                'warning',
+                'application.helpers.sorthelper.construct'
+            );
+        }
+    }
+
+    public static function getInstance($language)
+    {
+        if (empty(self::$instance) || $language === self::$language) {
+            self::$instance = new self($language);
+        }
+
+        return self::$instance;
+    }
+
+    /**
+     * Sort array maintaining index association
+     * @see asort and Collator::asort
+     * @param string[] $array to sort
+     * @param integer $flags in self::SORT_REGULAR (default or invalid), self::SORT_NUMERIC, self::SORT_STRING
+     * @return boolean, see Collator::asort
+     */
+    public function asort(array &$array, int $flags = self::SORT_REGULAR): bool
+    {
+        if (is_null(self::$collator)) {
+            return asort($array, self::getFlag($flags));
+        }
+        return self::$collator->asort($array, self::getFlag($flags));
+    }
+
+    /**
+     * Return flag tupe depend on functoion sed
+     * @param integer
+     * @return integer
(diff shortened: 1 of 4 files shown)


Shnoulle:
- About collator : unsure need to fix more local (es-mx to es_MX for example)
- Seems php-intl is not mandatory currently when install. I add it on warning if it's OK.
- Why not using [setlocale](https://www.php.net/manual/en/function.setlocale) : «The locale information is maintained per process, not per thread.»

Shnoulle:
Without intl: do a DB request ?

gabrieljenik:
> 1. We really must avoid usage of [setlocale](https://www.php.net/manual/en/function.setlocale.php)

Naive question: Why we should avoid it? Could you expand?

application/models/Question.php
```diff
@@ -778,23 +778,37 @@ private function sortAnswerOptions($answerOptions, $language = null)
...
             if (empty($language) || !in_array($language, $this->survey->allLanguages)) {
                 $language = $this->survey->language;
             }
+            $locale = $language;
+            $aLanguageData = getLanguageData();
+            if (isset($aLanguageData[$language]['cldr'])) {
+                $locale = $aLanguageData[$language]['cldr'];
+            }
+            // need php_intl
+            if (function_exists('collator_create')) {
```
gabrieljenik, line 792: Can we create a wrapper for the collator? or maybe for the sorting procedures?
Then we wouldn't be doing function_exists

Shnoulle:
> > 1. We really must avoid usage of [setlocale](https://www.php.net/manual/en/function.setlocale.php)
> 
> Naive question: Why we should avoid it? Could you expand? 

>  The locale information is maintained per process, not per thread. 

Then : if you have 2 different language on 2 different thread : it can broke.

And if you have another php app : you can broke it.

application/models/Question.php
Shnoulle: > Can we create a wrapper for the collator? or maybe for the sorting procedures? Then we wouldn't be repeating function_exists so much

You mean 2 different functions with 80% same code ? Unsure on the best option here.

gabrieljenik: I meant a function like this:
```
collator_sort($sorted, $collator_sort_type, $fallback_sort_type) {
            // need php_intl
            if (function_exists('collator_create')) {
                $collator = collator_create($locale);
                collator_asort($collator, $sorted, collator_sort_type);
            } else {
                Yii::log("For better alphabeticall ordering : you need php-intl.", 'warning');
                asort($sorted, $fallback_sort_type);
            }            
}
```

Shnoulle: Inside a foreach : creating a new Collator for each foreach ?

Shnoulle: If i move too just use an uasort inside the foreach.

gabrieljenik: > Inside a foreach : creating a new Collator for each foreach ?

Yep, you are right. There other ways, like creating a class.
`class CollatorHelper` with a sort method.
But maybe is not needed. Leave it up to you.

gabrieljenik (approved):
Code structure looks OK
I think a helper would be usefull for the future and to spread out good sorting practices.
Still, could be left out of the scope of the fix.

application/models/Question.php
Shnoulle: It's a foreach for scale : maybe we can do it ? Since we have maximum 2 scale ?

Shnoulle:
> I think a helper would be usefull for the future and to spread out good sorting practices.

New class in https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/utils ?

Sorting.php ?

Still need the control when install.

gabrieljenik:
> > I think a helper would be usefull for the future and to spread out good sorting practices.
> 
> New class in https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/utils ?
> 
> Sorting.php ?

Maybe in application/helpers/ ?
sortingHelper.php

Shnoulle:
> Maybe in application/helpers/ ?
> sortingHelper.php

In application/helpers/ it's more for list of function. 
I check for inclusion to

Shnoulle:
- [ ] Move to an helper Class
- [ ] Add the warning in installer

Shnoulle:
`LimeSurvey\Helpers\SortHelper::getInstance($language)->asort($sorted, LimeSurvey\Helpers\SortHelper::SORT_STRING);`

application/helpers/SortHelper.php
```diff
@@ -0,0 +1,111 @@
...
+    private function __construct($language)
+    {
+        self::$language = $language;
+        if (class_exists('Collator')) {
+            $locale = $language;
+            $languageData = getLanguageData();
+            if (isset($languageData[$language]['cldr'])) {
+                $locale = $languageData[$language]['cldr'];
+            }
+            self::$collator = new \Collator($locale);
```
gabrieljenik, line 49: Does it wort caching the collator as in the cases this is called inside a for-loop? Not sure if itis a heavy load thing

gabrieljenik (approved):
Code looks good.
Would be nice to add a php unit.

Shnoulle:
Must add a master before

PR #3621, opened 2023-11-16
New feature #18880: New imagefile option type for theme options

gabrieljenik:


application/models/TemplateConfiguration.php
```diff
@@ -982,7 +982,7 @@ public function getOptionPageAttributes()
             foreach ($fileList as $file) {
                 $imageInfo = $this->getImageInfo($basePath . $file['name'], $pathPrefix);
                 if ($imageInfo) {
-                    $aData['imageFileList'][] = array_merge(
+                    $aData['imageFileList'][$imageInfo['filepath']] = array_merge(
```
Shnoulle, line 985: This don't disable usage of 2 filename in different directory ?
For example 

1. Logo.png at global upload directry
2. Logo.png at template directory
3. Logo.png at survey level directory

application/views/themeOptions/options_core.php
```diff
@@ -6,6 +6,42 @@
     "inherited value:"
 ) . ' ' . (isset($oParentOptions['font']) ? $oParentOptions['font'] : '') . ']</option>' : '');
 
+/** @var string The html for image file dropdown options */
+$imageOptions = '';
+$optgroup = '';
+foreach ($aTemplateConfiguration['imageFileList'] as $image) {
+    // If group is different than the previous one, close the previous optgroup and open a new one
+    if ($image['group'] != $optgroup) {
```
Shnoulle, line 14: We have 3 times near this function in one file ? Right ?

- https://github.com/LimeSurvey/LimeSurvey/blob/dc71be6dc23f871d9cd5f47b7f6dcd326afc5796/application/views/themeOptions/options_core.php#L12
- https://github.com/LimeSurvey/LimeSurvey/blob/dc71be6dc23f871d9cd5f47b7f6dcd326afc5796/application/views/themeOptions/options_core.php#L49
- https://github.com/LimeSurvey/LimeSurvey/blob/dc71be6dc23f871d9cd5f47b7f6dcd326afc5796/application/views/themeOptions/options_core.php#L74

Why not create a variable  or an function ?

application/models/TemplateConfiguration.php
gabrieljenik: Shouldn't. The key of the array is the path, not the filename.

application/views/themeOptions/options_core.php
gabrieljenik: Regarding code duplication, I believe it is not worth improving it now: What would have to be done is convert the logo and the background to the new type. Now it has been tripled because there is one for the new type, one for the logo and another for the background.

Shnoulle: Oh right : then when move logo and background to `$attribute['type'] == 'imagefile'` it fix the issue ?

Then why not do it now ?

gabrieljenik: I would say out fo the scope. It would require migration of already set options... a mees on its own.

Shnoulle: > I would say out fo the scope. It would require migration of already set options... a mees on its own.

Ah ! Right !
Need API version update,
Maybe for develop ?

It **must** be done at a time !

Shnoulle:
@gabrieljenik : can i create the Dev issue for fixing options xml and assign it to you ?

Shnoulle (approved):

gabrieljenik:
> @gabrieljenik : can i create the Dev issue for fixing options xml and assign it to you ?

Sure

Shnoulle:
I create it, but i think it's best if deprecate it now in current core theme

I mean here : 

- https://github.com/LimeSurvey/LimeSurvey/blob/74c5e99811fc97fac811403cc05d5bba89968f7f/themes/survey/vanilla/config.xml#L73
- https://github.com/LimeSurvey/LimeSurvey/blob/74c5e99811fc97fac811403cc05d5bba89968f7f/themes/survey/fruity/config.xml#L111
- https://github.com/LimeSurvey/LimeSurvey/blob/74c5e99811fc97fac811403cc05d5bba89968f7f/themes/survey/fruity/config.xml#L113
- etc …

then future copy of core plugin are OK for 7.X

Shnoulle:
> > @gabrieljenik : can i create the Dev issue for fixing options xml and assign it to you ?
> 
> Sure

https://bugs.limesurvey.org/view.php?id=19269

gabrieljenik:
> I create it, but i think it's best if deprecate it now in current core theme

Let;s follow up on the ticket.
https://bugs.limesurvey.org/view.php?id=19269

PR #3633, opened 2023-11-22
Fixed issue [security] #19252: Bad permission check when edit quota

Shnoulle:
Dev: always control if quota->sid are related to surveyid param

### DIFF
--- a/application/controllers/QuotasController.php
+++ b/application/controllers/QuotasController.php
@@ -9,7 +9,7 @@ class QuotasController extends LSBaseController
     {
         return array(
             'accessControl', // perform access control for CRUD operations
-            'postOnly + deleteQuota', // we only allow deletion via POST request
+            'postOnly + deleteAnswer, deleteQuota, insertQuotaAnswer', // we only allow deletion via POST request
         );
     }
 
@@ -35,10 +35,7 @@ class QuotasController extends LSBaseController
     public function actionIndex($surveyid)
     {
         $surveyid = sanitize_int($surveyid);
-        if (!(Permission::model()->hasSurveyPermission($surveyid, 'quotas'))) {
-            Yii::app()->user->setFlash('error', gT("Access denied."));
-            $this->redirect(Yii::app()->request->urlReferrer);
-        }
+        $this->checkPermissions($surveyid);
         $oSurvey = Survey::model()->findByPk($surveyid);
         $aData['surveyid'] = $oSurvey->sid;
         // Set number of page
@@ -105,10 +102,7 @@ class QuotasController extends LSBaseController
     public function actionQuickCSVReport($surveyid)
     {
         $surveyid = sanitize_int($surveyid);
-        if (!(Permission::model()->hasSurveyPermission($surveyid, 'quotas'))) {
-            Yii::app()->user->setFlash('error', gT("Access denied."));
-            $this->redirect(Yii::app()->request->urlReferrer);
-        }
+        $this->checkPermissions($surveyid);
         $oSurvey = Survey::model()->findByPk($surveyid);
 
         /* Export a quickly done csv file */
@@ -134,10 +128,7 @@ class QuotasController extends LSBaseController
     public function actionAddNewQuota($surveyid)
     {
         $surveyid = sanitize_int($surveyid);
-        if (!(Permission::model()->hasSurveyPermission($surveyid, 'quotas', 'create'))) {
-            Yii::app()->user->setFlash('error', gT("Access denied."));
-            $this->redirect(Yii::app()->request->urlReferrer);
-        }
+        $this->checkPermissions($surveyid, 'create');
 
         Yii::app()->loadHelper('admin.htmleditor');
 
@@ -170,8 +161,8 @@ class QuotasController extends LSBaseController
         $oQuota = new Quota();
         $oQuota->sid = $oSurvey->primaryKey;
         $quotaService = new \LimeSurvey\Models\Services\Quotas($oSurvey);
-        if (isset($_POST['Quota'])) {
-            $oQuota = $quotaService->saveNewQuota($_POST['Quota']);
+        if (App()->getRequest()->getPost('Quota')) {
+            $oQuota = $quotaService->saveNewQuota(App()->getRequest()->getPost('Quota'));
             if (!$oQuota->getErrors()) {
                 Yii::app()->user->setFlash('success', gT("New quota saved"));
                 $this->redirect($this->createUrl("quotas/index/surveyid/$surveyid"));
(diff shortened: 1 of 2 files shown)


application/controllers/QuotasController.php
```diff
@@ -198,17 +189,13 @@ public function actionEditQuota($surveyid)
     {
         $surveyid = sanitize_int($surveyid);
         $oSurvey = Survey::model()->findByPk($surveyid);
-        if (!(Permission::model()->hasSurveyPermission($surveyid, 'quotas', 'update'))) {
-            Yii::app()->user->setFlash('error', gT("Access denied."));
-            $this->redirect(Yii::app()->request->urlReferrer);
-        }
-
         $quotaId = sanitize_int(Yii::app()->request->getQuery('quota_id'));
+        $this->checkPermissions($surveyid, 'update', $quotaId);
```
gabrieljenik, line 193: If I am not wrong, usually the check permission is done using the surveyId from the quota model.

application/controllers/QuotasController.php
```diff
@@ -258,12 +245,9 @@ public function actionEditQuota($surveyid)
     public function actionDeleteQuota()
     {
         $surveyid = sanitize_int(Yii::app()->request->getPost('surveyid'));
-        if (!(Permission::model()->hasSurveyPermission($surveyid, 'quotas', 'delete'))) {
-            Yii::app()->user->setFlash('error', gT("Access denied."));
-            $this->redirect(Yii::app()->request->urlReferrer);
-        }
-
         $quotaId = Yii::app()->request->getPost('quota_id');
+        $this->checkPermissions($surveyid, 'delete', $quotaId);
```
gabrieljenik, line 249: Same applies as before.

application/controllers/QuotasController.php
```diff
@@ -284,18 +268,11 @@ public function actionNewAnswer(int $surveyid)
         $surveyid = sanitize_int($surveyid);
         $oSurvey = Survey::model()->findByPk($surveyid);
         $aData['surveyid'] = $surveyid;
+        $quotaId = Yii::app()->request->getParam('quota_id');
+        $this->checkPermissions($surveyid, 'update', $quotaId);
```
gabrieljenik, line 272: Same applies as before.

application/controllers/QuotasController.php
```diff
@@ -427,19 +402,45 @@ public function actionMassiveAction($action, $surveyid)
...
+     * @param integer|null $iQuotaid
+     * throw Exception
+     * @return void
+     */
+    private function checkPermissions($surveyId, $sPermission = 'read', $quotaId = null)
+    {
+        if (!Permission::model()->hasSurveyPermission($surveyId, 'quotas', $sPermission)) {
+            throw new CHttpException(403, gT("You do not have permission on this survey."));
+        }
+        if ($quotaId) {
```
gabrieljenik, line 436: Is this part of the checking permission? Or this is part of a parent (survey) - child (quota) relationship checking?
Usually these kind of checks are not done on checkPermission routines.
I would say this needs to be moved out of this method.

application/controllers/QuotasController.php
Shnoulle: It's a private function used here …

I can just rename it to checkPermissionAndValidity …

application/controllers/QuotasController.php
Shnoulle: >  I would say this needs to be moved out of this method.

You don't like Permission ? I can move to checkValidity …

gabrieljenik: > You don't like Permission ? I can move to checkValidity …

I don;t like checking the quota belongs to a survey inside a checkPermission.
I believe they should be 2 different functions, and 2 different steps, as done everywhere else, right?.

Shnoulle: > as done everywhere else, right?.

? Where ?

gabrieljenik: Ex:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/QuestionAdministrationController.php#L125

- The question is fetched and checked it exists.
- Then the permission is checked against the surveyid stored in the model.

This way of doing it:
- is not consistent with the rest of the code 
- I don't like the idea of checking permission with the surveyId got from a post when the quota model is available

Shnoulle: OK for master : 

```
$quota = $this->getQuotaWithPermission($quotaId, 'update');
$sureyid = $quota->sid;
```

But not for 5.X …

gabrieljenik: > But not for 5.X …

sounds reasonable.

Shnoulle:
```
    /**
     * Get a quote aftre check exist and permission on a specific quota using permission on survey
     * @param integer $quotaId
     * @param string $sPermission to check (on survey quotas)
     * throw Exception
     * @return \Quota
     */
    private function getQuotaWithPermission($quotaId, $sPermission = 'read')
    {
        $oQuota = Quota::model()->findByPk($quotaId);
        if (empty($oQuota)) {
            throw new CHttpException(404, gT("Quota not found."));
        }
        if (!Permission::model()->hasSurveyPermission($oQuota->sid, 'quotas', $sPermission)) {
            throw new CHttpException(403, gT("You do not have permission on this quota."));
        }
        return $oQuota;
    }
```

(i update phpdoc)

gabrieljenik (approved):
Code structure looks good

application/controllers/QuotasController.php
```diff
@@ -257,45 +249,34 @@ public function actionEditQuota($surveyid)
...
+        $surveyid = $oQuota->sid;
+
         Quota::model()->deleteByPk($quotaId);
         QuotaLanguageSetting::model()->deleteAllByAttributes(array('quotals_quota_id' => $quotaId));
         QuotaMember::model()->deleteAllByAttributes(array('quota_id' => $quotaId));
 
         Yii::app()->user->setFlash('success', sprintf(gT("Quota with ID %s was deleted"), $quotaId));
 
-        $this->redirect($this->createUrl("quotas/index/surveyid/$surveyid"));
+        $this->redirect($this->createUrl(["quotas/index/",[ 'surveyid' => $surveyid]]));
```
olleharstedt, line 262: I don't know why, but this change forwards the app to public survey list index.php instead of quota admin. One customer get a white-age, dunno why yet.

olleharstedt: Customer get a 400 invalid request, actually.

olleharstedt: Yeah, it works when I restore this line. Funny.

olleharstedt: Maybe I'll just revert all those redirect changes...

gabrieljenik: By comparing with other redirects, starting slash is missing here...

By looking here, there is some specific module behaviour if there is no starting slash
https://www.yiiframework.com/doc/api/1.1/CController#createUrl-detail

Shnoulle: :( sorry for mistake. Try ming remove laser / van eux issue.

Iis server ?

olleharstedt: @gabrieljenik nice, will have a look, thanks!

olleharstedt: Adding a slash did not help. Will just revert this line to its previous content.

Shnoulle: > Adding a slash did not help. Will just revert this line to its previous content.

It's not adding, it's remove … `$this->redirect($this->createUrl(["quotas/index",[ 'surveyid' => $surveyid]]));` 
no ?

gabrieljenik: By looking at it again, I just feel the function is being called wrong.
Signature: `createUrl(string $route, array $params=array ( ), string $ampersand='&')`

We are only using the routes param if I am not mistaken, right?

Shnoulle: Right : must be `$this->redirect($this->createUrl("quotas/index",[ 'surveyid' => $surveyid]));`

Shnoulle: Olle fix it with `$this->redirect($this->createUrl("quotas/index/surveyid/{$surveyid}")`

(why it work when i test ????

gabrieljenik: > (why it work when i test ????

You are sending the full route there, with the params embedded. Still, may not work for all route formats

PR #3634, opened 2023-11-22
Fixed issue [security] #19252: Bad permission check when edit quota (Thanks to gblanco )

Shnoulle:
Dev: always control if quota->sid are related to surveyid param
Dev: add some POST action control

### DIFF
--- a/application/controllers/admin/Quotas.php
+++ b/application/controllers/admin/Quotas.php
@@ -58,28 +58,36 @@ class Quotas extends SurveyCommonAction
     }
 
     /**
-     * Check permission cfor a survey about quota
+     * Check permission for a survey about quota, redirect if no permission
      * @param integer $iSurveyId
      * @param string $sPermission
-     * @param integer|null $iQuotaid
-     * throw Exception
      * @return void
      */
-    private function checkPermissions($iSurveyId, $sPermission, $iQuotaid = null)
+    private function checkPermissions($iSurveyId, $sPermission)
     {
         if (!empty($sPermission) && !(Permission::model()->hasSurveyPermission($iSurveyId, 'quotas', $sPermission))) {
             Yii::app()->session['flashmessage'] = gT('Access denied!');
             $this->redirectToIndex($iSurveyId);
         }
-        if ($iQuotaid) {
-            $model = Quota::model()->find(
-                "id = :quotaid AND sid = :surveyid",
-                [':quotaid' => $iQuotaid, ':surveyid' => $iSurveyId]
-            );
-            if (empty($model)) {
-                throw new CHttpException(403, gT("Bad quota id with this survey id."));
-            }
+    }
+
+    /**
+     * Check validity of a quotaid against a survey id
+     * @param integer $iSurveyId
+     * @param integer $quotaid
+     * throw Exception
+     * @return void
+     */
+    private function checkValidQuotaId($surveyId, $quotaId)
+    {
+        $model = Quota::model()->find(
+            "id = :quotaid AND sid = :surveyid",
+            [':quotaid' => $quotaId, ':surveyid' => $surveyId]
+        );
+        if (empty($model)) {
+            throw new CHttpException(403, gT("Bad quota id with this survey id."));
         }
+
     }
 
     private function redirectToIndex($iSurveyId)
@@ -302,7 +310,8 @@ class Quotas extends SurveyCommonAction
         $iSurveyId = sanitize_int($iSurveyId);
         $qid = Yii::app()->request->getPost('quota_qid');
         $quota_id = Yii::app()->request->getPost('quota_id');
-        $this->checkPermissions($iSurveyId, 'update', $quota_id);
+        $this->checkPermissions($iSurveyId, 'update');
+        $this->checkValidQuotaId($surveyId, $quota_id);
 
         $oQuotaMembers = new QuotaMember('create'); // Trigger the 'create' rules
         $oQuotaMembers->sid = $iSurveyId;
@@ -344,7 +353,8 @@ class Quotas extends SurveyCommonAction
         if (empty($QuotaMember)) {
             throw new CHttpException(404, gT("Quota memeber not found."));
         }
-        $this->checkPermissions($iSurveyId, 'update', $QuotaMember->quota_id);
+        $this->checkPermissions($iSurveyId, 'update');
+        $this->checkValidQuotaId($iSurveyId, $QuotaMember->quota_id);
         $QuotaMember->delete();
         self::redirectToIndex($iSurveyId);
     }
@@ -359,7 +369,8 @@ class Quotas extends SurveyCommonAction
 
         $iSurveyId = sanitize_int($iSurveyId);
         $quotaId = Yii::app()->request->getQuery('quota_id');
-        $this->checkPermissions($iSurveyId, 'delete', $quotaId);
+        $this->checkPermissions($iSurveyId, 'delete');
+        $this->checkValidQuotaId($iSurveyId, $quotaId);
 
         Quota::model()->deleteByPk($quotaId);
         QuotaLanguageSetting::model()->deleteAllByAttributes(array('quotals_quota_id' => $quotaId));
@@ -381,7 +392,8 @@ class Quotas extends SurveyCommonAction
         $oSurvey = Survey::model()->findByPk($iSurveyId);
         $quotaId = Yii::app()->request->getQuery('quota_id');
 
-        $this->checkPermissions($iSurveyId, 'update', $quotaId);
+        $this->checkPermissions($iSurveyId, 'update');
+        $this->checkValidQuotaId($iSurveyId, $quotaId);
         $aData = $this->getData($iSurveyId);
         $aViewUrls = array();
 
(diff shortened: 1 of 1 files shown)


Shnoulle:
Lot of difference from master https://github.com/LimeSurvey/LimeSurvey/pull/3633

application/controllers/admin/Quotas.php
```diff
@@ -58,14 +58,28 @@ private function getData($iSurveyId)
...
+     * @return void
      */
-    private function checkPermissions($iSurveyId, $sPermission)
+    private function checkPermissions($iSurveyId, $sPermission, $iQuotaid = null)
     {
         if (!empty($sPermission) && !(Permission::model()->hasSurveyPermission($iSurveyId, 'quotas', $sPermission))) {
             Yii::app()->session['flashmessage'] = gT('Access denied!');
             $this->redirectToIndex($iSurveyId);
         }
+        if ($iQuotaid) {
```
gabrieljenik, line 74: Is this part of the checking permission? Or this is part of a parent (survey) - child (quota) relationship checking?
Usually these kind of checks are not done on checkPermission routines.
I would say this needs to be moved out of this method.

application/controllers/admin/Quotas.php
```diff
@@ -271,13 +297,17 @@ public function index(int $iSurveyId, bool $quickreport = false)
      */
     public function insertquotaanswer(int $iSurveyId)
     {
+        $this->requirePostRequest();
+
         $iSurveyId = sanitize_int($iSurveyId);
-        $this->checkPermissions($iSurveyId, 'update');
+        $qid = Yii::app()->request->getPost('quota_qid');
+        $quota_id = Yii::app()->request->getPost('quota_id');
+        $this->checkPermissions($iSurveyId, 'update', $quota_id);
```
gabrieljenik, line 305: If I am not wrong, usually the check permission is done using the surveyId from the quota model.

Even more, if we have the quota model, why do we need surveyId parameter?
Maybe out of the scope of this fix, but at least we could stop using it and use the one from the quota model, right?

application/controllers/admin/Quotas.php
```diff
@@ -303,15 +333,19 @@ public function insertquotaanswer(int $iSurveyId)
...
+        $iSurveyId = sanitize_int($iSurveyId);
+        $id = App()->request->getPost('quota_member_id');
+        $qid = App()->request->getPost('quota_qid');
+        $code = App()->request->getPost('quota_anscode');
+        /* find related quota by quotamember->id */
+        $QuotaMember = QuotaMember::model()->findByPk($id);
+        if (empty($QuotaMember)) {
+            throw new CHttpException(404, gT("Quota memeber not found."));
+        }
+        $this->checkPermissions($iSurveyId, 'update', $QuotaMember->quota_id);
```
gabrieljenik, line 347: Same applies as before.

application/controllers/admin/Quotas.php
gabrieljenik: If we need a function for checking a quota belongs to a survey, we could do it on a new function.

application/controllers/admin/Quotas.php
```diff
@@ -324,9 +358,8 @@ public function delquota(int $iSurveyId)
         $this->requirePostRequest();
 
         $iSurveyId = sanitize_int($iSurveyId);
-        $this->checkPermissions($iSurveyId, 'delete');
-
         $quotaId = Yii::app()->request->getQuery('quota_id');
+        $this->checkPermissions($iSurveyId, 'delete', $quotaId);
```
gabrieljenik, line 362: Same applies as before.

application/controllers/admin/Quotas.php
Shnoulle: It check permission on quota here. I can rename it to checkValidity , but it still a private function.

application/controllers/admin/Quotas.php
Shnoulle: >  Maybe out of the scope of this fix, but at least we could stop using it and use the one from the quota model, right?

Clearly out of scope : valid for Question, Answer, Questioni10n, Answeri10n etc …

Then : if we construct something must be in trait

Shnoulle: > If we need a function for checking a quota belongs to a survey, we could do it on a new function.

Same for ALL other model related to Survey : out of scope.

gabrieljenik: You are checking the quota belongs to a survey on a check permission function. I am just saying, move that part to a new function.

application/controllers/admin/Quotas.php
gabrieljenik: The permission is checked for the survey. The quota id is only used for checking it belongs to it. right?

application/controllers/admin/Quotas.php
Shnoulle: > You are checking the quota belongs to a survey on a check permission function. I am just saying, move that part to a new function.

the checkPermission function is a private function … i just rename it then.

application/controllers/admin/Quotas.php
Shnoulle: Yes , and ?

gabrieljenik: Let's unify conversations: https://github.com/LimeSurvey/LimeSurvey/pull/3633#discussion_r1402080740

application/controllers/admin/Quotas.php
Shnoulle: Added

```
    /**
     * Check validity of a quotaid against a survey id
     * @param integer $iSurveyId
     * @param integer $quotaid
     * throw Exception
     * @return void
     */
    private function checkValidQuotaId($surveyId, $quotaId)
    {
        $model = Quota::model()->find(
            "id = :quotaid AND sid = :surveyid",
            [':quotaid' => $quotaId, ':surveyid' => $surveyId]
        );
        if (empty($model)) {
            throw new CHttpException(403, gT("Bad quota id with this survey id."));
        }

    }
```

application/controllers/admin/Quotas.php
```diff
@@ -344,7 +353,8 @@ public function delans(int $iSurveyId)
         if (empty($QuotaMember)) {
             throw new CHttpException(404, gT("Quota memeber not found."));
         }
-        $this->checkPermissions($iSurveyId, 'update', $QuotaMember->quota_id);
+        $this->checkPermissions($iSurveyId, 'update');
```
gabrieljenik, line 356: This checkPermissions shouldn't be called using QuotaMemeber->sid ?

gabrieljenik (approved):

application/controllers/admin/Quotas.php
Shnoulle: > This checkPermissions shouldn't be called using QuotaMemeber->sid ?

Look at the new code :  i create a second function, and use it here.

```
$this->checkPermissions($iSurveyId, 'update');
$this->checkValidQuotaId($iSurveyId, $QuotaMember->quota_id);
```

Shnoulle:
red … CI

application/controllers/admin/Quotas.php
gabrieljenik: I meant the iSurveyId that is used for CheckPermission comes from parameter, not from the model.
I guess it is could be a bit safer... not sure how, but feels a bit more protected...

gabrieljenik: Just disregard if you disagree.

gabrieljenik (approved):

application/controllers/admin/Quotas.php
Shnoulle: > I meant the iSurveyId that is used for CheckPermission comes from parameter, not from the model.

We are in 5.X : i don't disable usage of surveyId by url … i do it in master : remove usage of surveyId (in controller) when i can.

>  I guess it is could be a bit safer... not sure how, but feels a bit more protected...

in checkValidQuotaId : i control if $iSurveyId is the same on thequota with $QuotaMember->quota_id

Here : https://github.com/LimeSurvey/LimeSurvey/pull/3634/files#diff-6c267486bf17814384c0ac50ac57c00ce39fa9a3affd8dd2527f9275933f9e68R85

gabrieljenik: ok

Shnoulle: :)

application/controllers/admin/Quotas.php
```diff
@@ -271,13 +304,18 @@ public function index(int $iSurveyId, bool $quickreport = false)
      */
     public function insertquotaanswer(int $iSurveyId)
     {
+        $this->requirePostRequest();
+
         $iSurveyId = sanitize_int($iSurveyId);
+        $qid = Yii::app()->request->getPost('quota_qid');
+        $quota_id = Yii::app()->request->getPost('quota_id');
         $this->checkPermissions($iSurveyId, 'update');
+        $this->checkValidQuotaId($surveyId, $quota_id);
```
ptelu, line 313: @Shnoulle please be a bit more careful when changing variable names, $surveyId should be $iSurveyId here. Code assistance in IDE marks this directly as red error, since this variable is not set. There was another issue in ls 6 with the same problem.

PR #3637, opened 2023-11-22
Fixed issue #19206: alphabetical order can be bad with utf8 caracter 

Shnoulle:
Fixed issue #19208: alphabetical order delete answers
Dev: Create and use an helper
Dev: use Collator if exist
Dev: Check if exist in installer (optionnal)


Shnoulle:
Need test

Shnoulle:
Same than 5.X but with test (selenium)

gabrieljenik (approved):
Nice. To be honest I was expecting a unit test, but I guess it works the same way :)

PR #3638, opened 2023-11-23
Fixed issue #19276: When a user exit on quota, the theme is not applied correctly if "inherits everything" is set to off

gabrieljenik:


Shnoulle:
Ok partial fix : still need aSurveyInfo (need to be reported) i create a sample theme for testing

Shnoulle (approved):

PR #3639, opened 2023-11-24
Fixed issue #19276: When a user exit on quota, the theme is not applied correctly if "inherits everything" is set to off

gabrieljenik:


application/helpers/frontend_helper.php
```diff
@@ -1944,6 +1944,7 @@ function checkCompletedQuota($surveyid, $return = false)
 
 
     // Datas for twig view
+    $thissurvey['sid'] = $surveyid;
```
Shnoulle, line 1947: \o/

Shnoulle (commented):
There are a lack of  `$thissurvey = $aSurveyInfo;`

before this line

```
$thissurvey = $aSurveyInfo;
$thissurvey['sid'] = $surveyid;
……
```


After it sent as `'aSurveyInfo' => $thissurvey`

Seems better to add it in same commit and not wait a issue reported.

gabrieljenik:
> Seems better to add it in same commit and not wait a issue reported.

What would be the way to make it fail and to test it ?
If we add it without that it may not be tested.

Shnoulle:
> > Seems better to add it in same commit and not wait a issue reported.
> 
> What would be the way to make it fail and to test it ? If we add it without that it may not be tested.

`{{ processString(aSurveyInfo.description,1) }}`  inside layout_global.twig for example.


aSurveyInfo **must** be aSurveyInfo at every page …

gabrieljenik:
> > > Seems better to add it in same commit and not wait a issue reported.
> > 
> > 
> > What would be the way to make it fail and to test it ? If we add it without that it may not be tested.
> 
> `{{ processString(aSurveyInfo.description,1) }}` inside layout_global.twig for example.
> 
> aSurveyInfo **must** be aSurveyInfo at every page …

Tried that and generates some unexpected behaviour. Ex: The progress bars starts to appear.
I would just tackle it on another ticket.

Shnoulle:
> Tried that and generates some unexpected behaviour. Ex: The progress bars starts to appear.

Argl,

Just tested on 5.X with skelvanilla https://demo.sondages.pro/376568?newtest=Y&lang=en

Capture d’écran du 2023-11-27 08-46-55

Shnoulle (approved):

application/helpers/frontend_helper.php
Shnoulle: It's not better with `Template::model()->getInstance(null, $surveyid);` too ?

Shnoulle:
> > > > Seems better to add it in same commit and not wait a issue reported.
> > > 
> > > 
> > > What would be the way to make it fail and to test it ? If we add it without that it may not be tested.
> > 
> > 
> > `{{ processString(aSurveyInfo.description,1) }}` inside layout_global.twig for example.
> > aSurveyInfo **must** be aSurveyInfo at every page …
> 
> Tried that and generates some unexpected behaviour. Ex: The progress bars starts to appear. I would just tackle it on another ticket.

I test with 

```
    $aDataReplacement['thissurvey'] = array_merge($aDataReplacement['thissurvey'], $thissurvey);
    Yii::app()->twigRenderer->renderTemplateFromFile("layout_global.twig", $aDataReplacement, false);
```

in 3.X :  Seems really better https://demo.sondages.me/854333

aDataReplacement is not used, clearly done for replacement.

PR #3640, opened 2023-11-27
dev: rename services/SurveyUpdater to services/SurveyAggregateService

kevin-foster-uk:
Renamed SurveyUpdated to SurveyAggregateService. Updated all references and tests. Also ran also related unit tests and cypress tests to ensure no issues.

### DIFF
--- a/application/controllers/admin/Database.php
+++ b/application/controllers/admin/Database.php
@@ -226,7 +226,7 @@ class Database extends SurveyCommonAction
     /**
      * Action to run when update survey settings + survey language
      *
-     * Refactored to use Services\SurveyUpdater 2023-05-30 (kfoster).
+     * Refactored to use Services\SurveyAggregateService 2023-05-30 (kfoster).
      *
      * @param integer $iSurveyID
      * @param ?array $input For dependency injection during testing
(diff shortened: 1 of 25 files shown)


kevin-foster-uk:
@Shnoulle 
To bring consistency with other new service classes.

Shnoulle:
Only [QuestionAggregateService](https://github.com/LimeSurvey/LimeSurvey/tree/183b9a6ada2a1176f763faa0d1bdd7363ff22b87/application/models/services/QuestionAggregateService) no ?

What the difference between ModelAggregateService and other's ?

When we need to create a ModelAggregateService ? etc ?

(maybe need a Manual Dev part here ?)

kevin-foster-uk:
> What the difference between ModelAggregateService and other's ?

QuestionAggregateService was created after SurveyUpdater, and follows a better naming convention. Both services act on aggregate data. They manipulate not only the Survey but also the entities which make up the Survey (Questions, Question Groups, Language Settings etc). This is distinct from a SurveyService or a QuestionService which would act on a single entity only.

> When we need to create a ModelAggregateService ? etc ?

A service is an aggregate if it operates on a hierarchy of data entities rather than a single entity.

> (maybe need a Manual Dev part here ?)

This naming convention is not fixed in stone. It could change later. Renaming these services in future should not be considered a breaking change. Plugins should not be directly calling these services. 

Do you have code which is already calling SurveyUpdater?

Shnoulle:
Thanks ! Clear now !

> This naming convention is not fixed in stone. It could change later. 

OK, but if we can have some written convention  still not fixed in stone, i think it can be great. We have a lot of code at different place for the same objective.

> Renaming these services in future should not be considered a breaking change. Plugins should not be directly calling these services.

It's not related to this : not in LimeSurvey->API for plugin : no need API update.

For Survey theme : it's different … more complex

PR #3653, opened 2023-12-06
Add theme, format data in the rest api

jack-han-115:
Dev:


application/libraries/Api/Command/V1/Transformer/Output/TransformerOutputSurvey.php
```diff
@@ -76,6 +76,8 @@ public function __construct(
             "googleanalyticsstyle" => ['key' =>  'googleAnalyticsStyle', 'type' => 'int'],
             "googleanalyticsapikey" => 'googleAnalyticsApiKey',
             "showsurveypolicynotice" => ['key' =>  'showSurveyPolicyNotice', 'type' => 'int'],
+            'template' => ['key' => 'template'],
```
kevin-foster-uk, line 79: If the key is the same on both sides you can just set boolean true as the value.

PR #3659, opened 2023-12-07
New feature #19308: Preview survey in modal (MW-728)

lajosarpad:
How to test:

Load a page like http://ls-ce/index.php?r=survey/index&sid=917414&newtest=Y&lang=en&popuppreview=true

where the value of sid is an actual survey id

Expectation: it will hite buttons, warnings.

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -1729,6 +1729,15 @@ class SurveyRuntimeHelper
         extract($args);
 
         $this->aSurveyInfo                 = getSurveyInfo($this->iSurveyid, App()->getLanguage());
+        if (isset($args['popuppreview']) && $args['popuppreview']) {
+            $this->aSurveyInfo['showxquestions'] = 'N';
+            $this->aSurveyInfo['shownoanswer'] = 'N';
+            $this->aSurveyInfo['showwelcome'] = 'N';
+            $this->aSurveyInfo['showprogress'] = 'N';
+            $this->aSurveyInfo['format'] = 'A';
+            $this->aSurveyInfo['listpublic'] = 'N';
+            $this->aSurveyInfo['popupPreview'] = true;
+        }
         $this->aSurveyInfo['surveyUrl']    = App()->createUrl("/survey/index", array("sid" => $this->iSurveyid));
 
         // TODO: check this:
--- a/application/controllers/survey/SurveyIndex.php
+++ b/application/controllers/survey/SurveyIndex.php
@@ -92,6 +92,7 @@ class SurveyIndex extends CAction
 
         // collect all data in this method to pass on later
         $redata = compact(array_keys(get_defined_vars()));
+        $redata['popuppreview'] = Yii::app()->request->getParam('popuppreview', false);
 
 
         $previewmode = false;
@@ -623,8 +624,10 @@ class SurveyIndex extends CAction
             buildsurveysession($surveyid, true);
             randomizationGroupsAndQuestions($surveyid, true);
             initFieldArray($surveyid, $_SESSION['survey_' . $surveyid]['fieldmap']);
+
         }
 
+        $popuppreview = (Yii::app()->request->getParam("popuppreview", false) == "true");
         // Reset the question timers in preview
         if (!$isSurveyActive || $previewmode) {
             resetQuestionTimers($surveyid);
--- a/application/controllers/survey/SurveyIndex.php
+++ b/application/controllers/survey/SurveyIndex.php
@@ -624,7 +624,6 @@ class SurveyIndex extends CAction
             buildsurveysession($surveyid, true);
             randomizationGroupsAndQuestions($surveyid, true);
             initFieldArray($surveyid, $_SESSION['survey_' . $surveyid]['fieldmap']);
-
         }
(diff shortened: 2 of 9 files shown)


olleharstedt:
For traceability, I prefer to have a connected new-feature issue reported in Mantis. Did one here: https://bugs.limesurvey.org/view.php?id=19308

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -1729,6 +1729,15 @@ private function setSurveySettings($surveyid, $args)
         extract($args);
 
         $this->aSurveyInfo                 = getSurveyInfo($this->iSurveyid, App()->getLanguage());
+        if (isset($args['popuppreview']) && $args['popuppreview']) {
+            $this->aSurveyInfo['showxquestions'] = 'N';
+            $this->aSurveyInfo['shownoanswer'] = 'N';
```
Shnoulle, line 1734: Why ?

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -1729,6 +1729,15 @@ private function setSurveySettings($surveyid, $args)
...
+            $this->aSurveyInfo['showwelcome'] = 'N';
```
Shnoulle, line 1735: Why hide it ?

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -1729,6 +1729,15 @@ private function setSurveySettings($surveyid, $args)
...
+            $this->aSurveyInfo['showprogress'] = 'N';
```
Shnoulle, line 1736: Not needed in All in one

application/helpers/SurveyRuntimeHelper.php
lajosarpad: We had a task for the cloud version to make sure that we have a readonly preview for surveys for new users shown in a popup and these changes were implemented for that purpose. Here, in this repo we included the support for this feature due to a recent decision according to which we want to simplify code maintenance for the future.

application/helpers/SurveyRuntimeHelper.php
lajosarpad: See https://github.com/LimeSurvey/LimeSurvey/pull/3659#discussion_r1447718138

application/helpers/SurveyRuntimeHelper.php
lajosarpad: Can you elaborate?

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -1729,6 +1729,15 @@ private function setSurveySettings($surveyid, $args)
...
+            $this->aSurveyInfo['format'] = 'A';
+            $this->aSurveyInfo['listpublic'] = 'N';
```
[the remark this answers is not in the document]
lajosarpad: Can you elaborate?

application/helpers/SurveyRuntimeHelper.php
Shnoulle: If survey is All in one page : then showprogress is set to false (but right : only if done in https://manual.limesurvey.org/AfterFindSurvey )

application/helpers/SurveyRuntimeHelper.php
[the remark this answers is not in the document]
Shnoulle: listpublic is used on index page,  not elsewhere

application/helpers/SurveyRuntimeHelper.php
Shnoulle: Why for shownoanswer to false ?

The preview are not the same than reality in this case. More for a printopdf system ?

application/helpers/SurveyRuntimeHelper.php
Shnoulle: Still don't understand why hide it : if user want to show it : 

- He want to have it when previewing
- He want to have it doing a print HTML page for reviewing.

application/helpers/SurveyRuntimeHelper.php
lajosarpad: "The preview are not the same than reality in this case."

Yes, the intention was to have this difference. If you have some ideas for improvement, then please present them to Nathalie or Jory, both of them checked this popup preview and approved it.

application/helpers/SurveyRuntimeHelper.php
lajosarpad: Because that was the intention. It was a decision made by the management, The user chooses a survey in a very specific flow and I was explicitly asked to hide the Welcome. If you dislike this idea, please discuss it with Jory.

application/helpers/SurveyRuntimeHelper.php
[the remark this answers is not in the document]
lajosarpad: If you have ideas for improvement in mind, then can you implement the changes you propose? We could try it out.

application/helpers/SurveyRuntimeHelper.php
Shnoulle: Again don't have access to MW-728 project part.

PR #3660, opened 2023-12-11
fixed url when directLogin true

mbischof:
Fixed issue #19302: directLogin = true with wrong redirect


Shnoulle (approved):
:+1:

Shnoulle:
@tiborpacalat i really think we don't need test here. The patch is clearly needed. Ping @olleharstedt

PR #3661, opened 2023-12-12
Fixed issue #19245: Search surveys status options do not work

gabrieljenik:


application/models/Survey.php
```diff
@@ -1655,15 +1655,18 @@ public function search()
...
                     $criteria->compare("t.active", 'Y');
-                    $subCriteria1 = new CDbCriteria();
-                    $subCriteria2 = new CDbCriteria();
-                    $subCriteria1->addCondition("'{$sNow}' > t.startdate", 'AND');
-                    $subCriteria1->addCondition("t.startdate IS NULL", 'OR');
-                    $subCriteria2->addCondition("t.expires IS NULL", 'AND');
-                    $subCriteria2->addCondition("'{$sNow}' < t.expires", 'OR');
-                    $criteria->mergeWith($subCriteria1);
+                    $startedCriteria = new CDbCriteria();
+                    $startedCriteria->addCondition("'{$sNow}' > t.startdate", 'OR');
```
Shnoulle, line 1663: I don't understand : since `$startedCriteria` is empty : what the differerence between addCondition(…, OR) and addCondition(…, AND)  ?

Seems more clear (if it' same) to do
```
                    $startedCriteria = new CDbCriteria();
                    $startedCriteria->addCondition("'{$sNow}' > t.startdate");
                    $startedCriteria->addCondition('t.startdate IS NULL', "OR");
```

On the 1st action we are here : https://www.yiiframework.com/doc/api/1.1/CDbCriteria#addCondition-detail
```
    if($this->condition==='')
        $this->condition=$condition;
```

(PS : seems we can set both at same time : but add it one by one is clear).

```
                    $startedCriteria->addCondition(
                        [   
                            "'{$sNow}' > t.startdate",
                            't.startdate IS NULL'
                        ],
                        "OR"
                    );
```

Shnoulle (approved):

PR #3665, opened 2023-12-12
Fixed issue [security] #19117: Account past their expiration date can be still active

Shnoulle:
Dev: add constionnal scope according to DB version
Dev: use it when search for current user
Dev: App()->user->getId use session too


Shnoulle:
master version : 
https://github.com/LimeSurvey/LimeSurvey/pull/3525
+
https://github.com/LimeSurvey/LimeSurvey/pull/3607

olleharstedt (approved):

PR #3666, opened 2023-12-15
Fixed #19217: PHP8 Error Type NULL 

BertHankes:
Fixed issue #19217:

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -665,7 +665,10 @@ class Survey extends LSActiveRecord implements PermissionInterface
     public function getTokenAttributes()
     {
         $attdescriptiondata = decodeTokenAttributes($this->attributedescriptions ?? '');
-
+        if(!is_array(reset($attdescriptiondata)))
+        {
+            $attdescriptiondata=NULL;
+        }
         // Catches malformed data
         if ($attdescriptiondata && strpos((string) key(reset($attdescriptiondata)), 'attribute_') === false) {
             // don't know why yet but this breaks normal tokenAttributes functionning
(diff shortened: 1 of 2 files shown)


olleharstedt:
Check failed tests in github actions.

application/models/Survey.php
```diff
@@ -665,7 +665,9 @@ public function getAllLanguages()
     public function getTokenAttributes()
     {
         $attdescriptiondata = decodeTokenAttributes($this->attributedescriptions ?? '');
-
+        if (!is_array(reset($attdescriptiondata))) {
```
kevin-foster-uk, line 668: Why do you call reset() here?

BertHankes: Because the reset() function is used by the native code on line 670:
if ($attdescriptiondata && strpos((string) key(reset($attdescriptiondata)), 'attribute_') === false)

And if reset() is not returning an array (in this case a NULL).. key() gives a 500
In trying to debug it is/was very hard to get to this state... strpos(key(reset($attdescriptiondata))... has 3 functions in one line.

The reset(array) is returning null because there are no tokens!

From PHP help:
reset() rewinds array's internal pointer to the first element and returns the value of the first array element.
The value returned by reset(), when not having tokens, is null!

And in PHP<8 it was all good.. but now it is strict! Key(NULL) is not allowed anymore and results in 500

BertHankes: key is not allowed to have a string or null as paramater.
In this case the paramater is a empty string!

kevin-foster-uk (approved):

PR #3667, opened 2023-12-17
[Qe-648] publication access date update issue

jack-han-115:
Dev:


kevin-foster-uk (approved):
I was under the impression that we store date time values on the server side in server timezone (Timestamp) but in this case its stored as a timezone-less datetime so no time conversion is required. 

Ideally we would still use the datetime formatter but there would be an option to disable time zone conversion. For now this solution is fine.

PR #3674, opened 2023-12-22
Fixed issue #19323: Error When updating running Update_619.php on MSSQL

ConnorPrettyman:
Updated to hopefully fix bug #19323.

The change in code for MSSQL was used on a production instance at work, and allows the upgrade to continue.
I have not tested the switch-case, but I used the one present in Update_616.php as a basis for what I did here.


tiborpacalat:
This has been finally resolved by https://github.com/LimeSurvey/LimeSurvey/pull/3713

PR #3676, opened 2023-12-23
Fixed issue #19248: When deleting a token through the model, the AuditLog beforeDeleteToken event fails

gabrieljenik:


application/core/plugins/AuditLog/AuditLog.php
```diff
@@ -382,8 +383,21 @@ public function beforeTokenDelete()
...
+        // The code below accepts both kinds of interface.
+        $sTokenIds = $event->get('sTokenIds');
+        if (!empty($sTokenIds)) {
+            $aTokenIds = explode(',', (string) $sTokenIds);
+        } else {
+            // If sTokenIds is empty, assume we're dealing with a dynamic model event.
+            // In this case, the dynamicId parameter contains the token ID.
+            $aTokenIds = [$event->get('dynamicId')];
+        }
+        if (empty($aTokenIds)) {
```
Shnoulle, line 398: :+1:

application/core/plugins/AuditLog/AuditLog.php
```diff
@@ -395,14 +409,52 @@ public function beforeTokenDelete()
                 $oAutoLog->uid = $oCurrentUser->uid;
                 $oAutoLog->entity = 'token';
                 $oAutoLog->action = 'delete';
-                $oAutoLog->entityid = $aValues['participant_id'];
+                $oAutoLog->entityid = $aValues['participant_id'];   // TODO: Should this be the token ID?
```
Shnoulle, line 412: ? You right : participant_id is from https://github.com/LimeSurvey/LimeSurvey/blob/8c332cc638306875f965ccd347ef5e5573f27ed0/application/models/Participant.php#L22 not in token model.

application/core/plugins/AuditLog/AuditLog.php
```diff
@@ -104,6 +104,7 @@ public function init()
         $this->subscribe('beforeDataEntryImport');
         $this->subscribe('beforeTokenSave');
         $this->subscribe('beforeTokenDelete');
+        $this->subscribe('beforeTokenDeleteMany');
```
Shnoulle, line 107: OK 
https://github.com/LimeSurvey/LimeSurvey/blob/8c332cc638306875f965ccd347ef5e5573f27ed0/application/models/LSActiveRecord.php#L213

Shnoulle (approved):
Seems OK

application/core/plugins/AuditLog/AuditLog.php
Shnoulle: It's another issue : not related here.

gabrieljenik: Created new issue: https://bugs.limesurvey.org/view.php?id=19325

PR #3677, opened 2023-12-23
Fixed issue #19253: CSV injection in export quota

gabrieljenik:


Shnoulle:
My opinion : only escape formula when export

Reason: we need it on a lot of other CSV export

application/helpers/common_helper.php
```diff
@@ -5105,3 +5105,23 @@ function isAbsolutePath($path)
         return false;
     }
 }
+
+/**
+ * Escapes a string for use in a CSV file
+ * @param string|null $string
+ * @return string
+ */
+function csvEscape($string)
```
Shnoulle, line 5114: No need `preg_replace(array('~\R~u'), array("\n"), $value);` and `'"' . str_replace('"', '""', $sString) . '"';`

?

See https://github.com/LimeSurvey/LimeSurvey/blob/0cc10f52a6127da9f8267e658728af4c72a27cac/application/helpers/admin/export/CsvWriter.php#L102

Shnoulle (commented):
Need more escape ?

Shnoulle (approved):

PR #3679, opened 2023-12-26
[QE-666] add owner for survey settings

jack-han-115:
Dev:

### DIFF
--- a/application/libraries/Api/Command/V1/Transformer/Output/TransformerOutputSurveyDetail.php
+++ b/application/libraries/Api/Command/V1/Transformer/Output/TransformerOutputSurveyDetail.php
@@ -61,6 +61,9 @@ class TransformerOutputSurveyDetail extends TransformerOutputActiveRecord
         $survey['templateInherited'] = $data->oOptions->template;
         $survey['formatInherited'] = $data->oOptions->format;
         $survey['languages'] = $data->allLanguages;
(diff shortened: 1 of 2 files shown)


application/libraries/Api/Command/V1/Transformer/Output/TransformerOutputSurveyDetail.php
```diff
@@ -61,6 +61,8 @@ public function transform($data)
         $survey['templateInherited'] = $data->oOptions->template;
         $survey['formatInherited'] = $data->oOptions->format;
         $survey['languages'] = $data->allLanguages;
+        $survey['owner'] = $data->owner;
```
kevin-foster-uk, line 64: The owner data contains values which we do not want to output via the api such as 'password', 'one_time_pw' and 'validation_key'.

To handle this you can create a new Transfomer class maybe TransformerOutputSurveyOwner.

An output-transformer class is used to convert the internal representation of data (server side) to the external representation of the data (to be consumed by the client). We also have transformer input-transformer classes which convert the data coming into to the API from the external representation back to the internal representation. 

I can show you how to implement the transformer and/or ask Shixiong for help with it.

zhao-create-xyz:
Closing this PR because https://github.com/LimeSurvey/LimeSurvey/pull/3689 will fix this issue.

PR #3686, opened 2024-01-02
Fixed issue #19331: Try to add a subquestion to a matrix type question throws an SQL error

gabrieljenik:


Shnoulle (approved):

Shnoulle:
Test done (on pgsql) by reporter : https://bugs.limesurvey.org/view.php?id=19332

Shnoulle:
@gabrieljenik with php8.3

`DOMAIN=lstest.local  phpunit ./tests/unit/services/QuestionAggregateService/SubquestionsServiceTest.php`
```
…
	<p class="message">
		Calling get_class() without arguments is deprecated	</p>

	<div class="source">
		<p class="file">/media/shnoulle/data/webdev/lstest/application/helpers/ClassFactory.php(39)</p>

```

Shnoulle:
Move the error from 8.2 to 7.4 ??????  WTF ?

application/models/services/QuestionAggregateService/SubQuestionsService.php
```diff
@@ -162,6 +162,9 @@ private function storeSubquestion(
                 $subquestion = DI::getContainer()->make(Question::class);
             }
         }
+        if (empty($subquestion)) {
```
Shnoulle, line 164: I think we must avoid this before ?

No ?

Throw an exception is a really good idea :)

gabrieljenik: > I think we must avoid this before 

Not sure there are lot of places before where to move this... do you find a better spot?

Shnoulle: > > I think we must avoid this before
> 
> Not sure there are lot of places before where to move this... do you find a better spot?

Sorry, I didn't make myself clear :)

I mean : call SubQuestionsService with invalid values for sub qids must be checked before :) and have an error management before.

Here : seems a good place inside service, but my opinion : this is more a error before (in controller) : value send to service must be clean.

PR #3688, opened 2024-01-03
Fixed issue #19335: Message from submitfailed is never shown to user in one case

olleharstedt:
Dev:

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -5071,7 +5071,7 @@ class LimeExpressionManager
                 $trace = $e->getTraceAsString();
                 $message = submitfailed($this->gT("Unable to insert record into survey table"), $query . "\n\n" . $trace);
                 LimeExpressionManager::addFrontendFlashMessage('error', $message, $this->sid);
-                return;
+                return $message;
             }
 
             //Insert Row for Timings, if needed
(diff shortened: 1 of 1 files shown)


application/helpers/expressions/em_manager_helper.php
```diff
@@ -5070,7 +5070,9 @@ private function _UpdateValuesInDatabase($finished = false)
                 $message .= $this->gT("Unable to insert record into survey table"); // TODO - add SQL error?
```
ptelu, line 5070: this gets overwritten so it can be removed

application/helpers/expressions/em_manager_helper.php
```diff
@@ -5070,7 +5070,9 @@ private function _UpdateValuesInDatabase($finished = false)
...
                 $query = $e->getMessage();
                 $trace = $e->getTraceAsString();
-                submitfailed($this->gT("Unable to insert record into survey table"), $query . "\n\n" . $trace);
+                $message = submitfailed($this->gT("Unable to insert record into survey table"), $query . "\n\n" . $trace);
+                LimeExpressionManager::addFrontendFlashMessage('error', $message, $this->sid);
+                return;
```
ptelu, line 5075: i look at what happens with the return message and it is supposedly displaying debug information when enabled, so this should return the message

ptelu (changes requested):

application/helpers/expressions/em_manager_helper.php
olleharstedt: Fixed.

application/helpers/expressions/em_manager_helper.php
olleharstedt: Hm maybe both places should be

```php
return $message;
```

then

ptelu: yeah theres a debug setting that when it is turned on will display the message

olleharstedt: Note that it should probably not fall through, since 'srid' is not set, so this line at bottom of function throws a warning:

     5228             'code'      => $_SESSION[$this->sessid]['srid'],

ptelu (approved):

PR #3690, opened 2024-01-08
LA-9: User session is not killed when user is marked as non-active

mohabmes:


olleharstedt (approved):

Shnoulle (approved):
There are a little update on scope usage (API semver) but since active scope is new and it's not really in API : no issue.

edgarrmondragon:
I think this change introduced an error, at least for the RC2 API. The `get_session_key` method is failing with:

```
CDbCommand failed to execute the SQL statement: SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: boolean = integer
LINE 1: ...13 17:01:11' OR expires IS NULL) AND (user_status = 1)) AND ...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.. The SQL statement executed was: SELECT * FROM "lime_users" "t" WHERE ((expires > :now OR expires IS NULL) AND (user_status = :active)) AND ("t"."uid"=1) LIMIT 1. Bound with :now='2024-01-13 17:01:11', :active=1
```

This is using Postgres 14 and the latest `master` (6.4.1). Happy to report a bug too but thought it'd be good to first reach out to the folks involved directly.

olleharstedt:
Isn't user_status supposed to be an integer?

Shnoulle:
> Isn't user_status supposed to be an integer?
> 
> Also, please include which database vendor you're using.

BOOLEAN : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/update/updates/Update_619.php

Same when install.

Unsure it work on MSSQL here https://github.com/LimeSurvey/LimeSurvey/blob/b71596bcf0e4fa18cc50c8c889dd8469af6eea69/vendor/yiisoft/yii/framework/db/schema/mssql/CMssqlSchema.php#L41

Surely broke  with pgsql

Shnoulle:
> Isn't user_status supposed to be an integer?

only for mysql/mariadb : https://github.com/LimeSurvey/LimeSurvey/blob/b71596bcf0e4fa18cc50c8c889dd8469af6eea69/vendor/yiisoft/yii/framework/db/schema/mysql/CMysqlSchema.php#L38

tiborpacalat:
There was an issue that I discovered last Friday. The issue is in sql script for MSSQL and Postgres. It is already being worked on.

olleharstedt:
Ugh, should better be int or tinyint. We might want more user_status than 0/1 in the future...

olleharstedt:
Or enum? ;)

Shnoulle:
> Or enum? ;)

Unsure we have an easy way to have enum on my+ms+pg with Yii ;)

olleharstedt:
No, that was a joke, hehe. I wouldn't go there.

Shnoulle:
> No, that was a joke, hehe. I wouldn't go there.

I know :)

PR #3694, opened 2024-01-10
add cloud related rest api

jack-han-115:
Dev:


kevin-foster-uk (commented):
I think we need to discuss this more with Nathalie before merging this. Few issues:
- Cookie consent? Where and when is this value used? 
- Footer branding - we currently have a global config option for this which allows branding to be turned off depending on the customers package. This is related to that so I think this is a cloud only value.

I am still trying to work out how to handle cloud specific data correctly.

c-schmitz:
Obsolete

PR #3695, opened 2024-01-10
Fixed issue #19313: No warning list when import same question code

gabrieljenik:


application/controllers/QuestionAdministrationController.php
```diff
@@ -1189,6 +1189,13 @@ public function actionImport()
             return;
         }
 
+        // If there are warnings, we don't jump to the question.
+        // We need to show the warnings to the user, and they may be too important
+        // and/or too many to be shown in a flash message.
+        if (!empty($aImportResults['importwarnings'])) {
+            $jumptoquestion = false;
```
Shnoulle, line 1196: Code is OK, but my opinion : $jumptoquestion must be **always** false …

Shnoulle (approved):
Partial fix

application/controllers/QuestionAdministrationController.php
Shnoulle: > Well, but jumpToQuestion is an input... So, maybe that should be another story?

Oups, yes right …

I don't see it then…

Else: i don't find the import question in current version ????? (have a lot of JS issue in last version)

PR #3696, opened 2024-01-11
Fixed issue #19344: Survey activation error using MSSQL

c-schmitz:
Dev Also added a new integrity check function  to archive tables that are left from previous activation attempts


olleharstedt (approved):

application/models/SurveyActivator.php
```diff
@@ -324,6 +324,7 @@ protected function createResponseTable()
             // Refresh schema cache just in case the table existed in the past
             Yii::app()->db->schema->getTable($sTableName, true);
         } catch (Exception $e) {
+                $this->error = 'surveytablecreation';
             if (App()->getConfig('debug')) {
```
Shnoulle, line 328: Why not always add `$e->getMessage();` ?

PR #3700, opened 2024-01-15
Fixed issue #CR-1392: SQL not compatible with MSSQL and Postgres

mohabmes:


olleharstedt (approved):

mohabmes:
I believe there's no need to convert true/false to 1/0 since the code already utilizes 1/0 for the user_status. A BOOLEAN data type in MySQL, typically implemented with the TINYINT data type, ensures proper conversion of values.

mohabmes:
I believe there's no need to convert true/false to 1/0 since the code already utilizes 1/0 for the user_status. A BOOLEAN data type in MySQL, typically implemented with the TINYINT data type, ensures proper conversion of values.

PR #3703, opened 2024-01-18
Fixed issue #18895: No way to add textarea with core config.xml

Shnoulle:
Dev: type="textarea" rows="8"


gabrieljenik (approved):
Code structure looks OK

PR #3705, opened 2024-01-22
Fixed issue [security] #19368: Simple admin can add XSS on adminemail

Shnoulle:
Dev: minimal fix, filter XSS only


themes/survey/vanilla/views/layout_errors.twig
```diff
@@ -70,7 +70,7 @@
                         {{gT("For further information please contact %s:")|format (aSurveyInfo.admin)}}
                         {% if aSurveyInfo.adminemail %}
                             <br>
-                            <a href='mailto:{{ aSurveyInfo.adminemail }}'>{{ aSurveyInfo.adminemail }}</a>
+                            <a href='mailto:{{ aSurveyInfo.adminemail|url_encode }}'>{{ aSurveyInfo.adminemail }}</a>
```
gabrieljenik, line 73: Should we add an html_encode when displaying the  aSurveyInfo.adminemail insidethe a tag contents?

gabrieljenik: Asking... wondering... not sure.

Shnoulle: > Should we add an html_encode when displaying the aSurveyInfo.adminemail insidethe a tag contents?

I don't know … in master : adminemail was restricted to be a valid email. But in 5 and before : we can add HTML.

I think we need only url encoding in url. But since we don't encode adminname or anything else from view , why admin email ?

gabrieljenik: ok, fair enough... bounce email doens't need to be encoded? not displayed anywhere?

Shnoulle: > ok, fair enough... bounce email doens't need to be encoded? not displayed anywhere?

No display :)

gabrieljenik (approved):

PR #3709, opened 2024-01-22
Fixed issue #14661: Plugin development need a "Don't disable plugin"

Shnoulle:
Dev: debug>=2 : no disable plugin by PHP issue


application/libraries/PluginManager/PluginManager.php
```diff
@@ -322,6 +322,9 @@ public function scanPlugins($includeInstalledPlugins = false)
                                     ];
                                     $saveResult = Plugin::setPluginLoadError($plugin, $pluginName, $error);
                                     if (!$saveResult) {
+                                        if(App()->getConfig('debug') >= 2) {
+                                            throw $ex;
+                                        }
                                         // This only happens if database save fails.
```
gabrieljenik, line 328: I would add a comment here saying that this will happen only if debug is not set to 2, right?

application/models/Plugin.php
```diff
@@ -71,6 +71,9 @@ public function tableName()
      */
     public function setLoadError(array $error)
     {
+        if (App()->getConfig('debug') >= 2) {
```
gabrieljenik, line 74: I am not sure about this mechanism.
This method, which is supposed to only save is somehow now deciding when the message should be thrown or not.

Why not making  method called "handleLoadError".
That method will decide if saving or throwing.
It will return something only when saving.

It may be too much, but I would move the disabling of the plugin inside that method as well.

gabrieljenik: > I don't want to rename a public function.

I meant to create a new method, not rename a current one.
`handleLoadError` would call, in some cases, to `setPluginLoadError`

Shnoulle: Yes ! Great !

application/libraries/PluginManager/PluginManager.php
Shnoulle: 2 comments

```
                                        // If handlePluginLoadError return 0 because debug is set
                                        if (App()->getConfig('debug') >= 2) {
                                            throw $ex;
                                        }
                                        // If handlePluginLoadError fail without debug : have a DB related issue
```

Shnoulle:
Seems i have an issue … no link to edit settings with this commit.

Shnoulle:
OK, all done. 
Did i do same for master ?

gabrieljenik (approved):
Code structure looks good!

gabrieljenik (approved):
Seems same as in master
Code structure looks good

Shnoulle:
> I would set the same for master

Yep, cherry-picking only :)

O do it. Unsure for who can test

PR #3711, opened 2024-01-23
Fixed issue #19347: No generation of new user within first LDAP login

Shnoulle:
Dev: use existing event, no need to create an dummy one
Dev: Use event error set


Shnoulle:
One of reporter

> as ldap user I tried wrong password - failed
> 
> I deleted the ldap user
> 
> try to login as ldap user with wrong credentials - failed
> try to login as ldap user with right credetials - success

https://bugs.limesurvey.org/view.php?id=19347#c79299

If someon find a way to have a valid user at https://github.com/LimeSurvey/LimeSurvey/blob/880b292494005b59f96074dddb9a6868f1724b12/application/core/plugins/AuthLDAP/AuthLDAP.php#L557 but invalid at https://github.com/LimeSurvey/LimeSurvey/blob/880b292494005b59f96074dddb9a6868f1724b12/application/core/plugins/AuthLDAP/AuthLDAP.php#L270

c-schmitz (approved):

PR #3713, opened 2024-01-24
Fixed issue #19351: Limesurvey broken after 5.6.50 to 6.4.2 upgrade

mohabmes:


edgarrmondragon:
@mohabmes I think this was fixed already in https://github.com/LimeSurvey/LimeSurvey/pull/3700 (see also See https://github.com/LimeSurvey/LimeSurvey/pull/3690#issuecomment-1890596073).

Still, version 6.4.2 seems to be broken but I don't know if there's a procedure for "yanking" versions.

mohabmes:
@edgarrmondragon Apparently, https://github.com/LimeSurvey/LimeSurvey/pull/3700 wasn't sufficient, the issue here is that PG couldn't cast a boolean automatically to an integer. [#19351](https://bugs.limesurvey.org/view.php?id=19351#c79268)

Shnoulle:
I don't undertand why https://github.com/LimeSurvey/LimeSurvey/blob/fad12234212fcb7ac2e55227833c2ce2305c5b1c/application/helpers/update/updates/Update_621.php is not sufficient …

PS : maybe we can update https://github.com/LimeSurvey/LimeSurvey/blob/fad12234212fcb7ac2e55227833c2ce2305c5b1c/application/helpers/update/updates/Update_619.php on this commit.

mohabmes:
@Shnoulle Should be fixed in the last commit, right? 
Adopted @edgarrmondragon solution here to ensure the app runs with no exceptions then executed Update_623 to convert/cast bool to int only for pgsql

Shnoulle (approved):

Shnoulle:
@tiborpacalat i don't know if you want to do a second test (maybe start from 5.X directly ?)

PR #3718, opened 2024-01-30
Fixed issue #19383: When export multiple choice filtered : loose N/A value

Shnoulle:
Dev: return null value if it's a null value


gabrieljenik (approved):
Code structure looks good. Haven't tested it.
I would say we create a ticket for creating unit tests for these classes. We can take those unit tests.

Shnoulle:
> I would say we create a ticket for creating unit tests for these classes.

We don't have other test for export ?

PR #3725, opened 2024-02-06
CT-626: Implemented support for the Hausa language

lajosarpad:
Dev:

### DIFF
--- a/application/extensions/admin/survey/ListSurveysWidget/views/listSurveys.php
+++ b/application/extensions/admin/survey/ListSurveysWidget/views/listSurveys.php
@@ -68,7 +68,7 @@
                 [
                     'header'            => gT('Title'),
                     'name'              => 'title',
-                    'value'             => '$data->defaultlanguage->surveyls_title',
+                    'value'             => '$data->defaultlanguage->surveyls_title ?? null',
                     'htmlOptions'       => ['class' => 'has-link'],
                     'headerHtmlOptions' => ['class' => 'text-nowrap'],
                 ],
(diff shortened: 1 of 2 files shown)


application/extensions/admin/survey/ListSurveysWidget/views/listSurveys.php
```diff
@@ -68,7 +68,7 @@
                 [
                     'header'            => gT('Title'),
                     'name'              => 'title',
-                    'value'             => '$data->defaultlanguage->surveyls_title',
```
c-schmitz, line 71: Huh? don't think that belongs in here

c-schmitz (changes requested):

application/extensions/admin/survey/ListSurveysWidget/views/listSurveys.php
lajosarpad: @c-schmitz if I create a survey with Hausa as its base language and then listing the surveys, then without the fix above I have:

image

whereas, if the fix above is in place, then:

image

If you still prefer this adjustment not to exist, then I will remove it.

c-schmitz: That is because you have faulty surveys in your database where the entries in surveys_languagesettings are missing.  Sure, this will prevent the notice but strictly speaking still does not belong  to 'Implement support for Hausa language'

lajosarpad: Thanks for the insight

PR #3730, opened 2024-02-09
Fixed issue #19359: js error in console about ckeditor

gabrieljenik:
- Updated CKEditor to 4.22.1 (latest open source version of CKEditor 4)
- Updated with second link from build-config.js without additions


Shnoulle (approved):
Always approve an update.

ptelu:
Could you also update the plugins loaded in config.js to the matching version of ckeditor, to prevent any issues or if you know they are not used remove them.

PR #3731, opened 2024-02-12
Dev: Fix PHP-DI CActiveRecord factory

kevin-foster-uk:


application/libraries/DI.php
```diff
@@ -41,34 +42,49 @@ public static function getContainer()
...
-        });
-
-        $container->set(CDbConnection::class, function () {
-            return App()->db;
-        });
+        return $builder->build();
+    }
 
-        return $container;
+    private static function getActiveRecordDefinitions()
```
twilligls, line 65: never used?

twilligls:
Tested and reviewed ok, except for the unused getActiveRecordDefinitions() function

application/libraries/DI.php
kevin-foster-uk: Turns out that the CActiveRecord DI doesn't work. The problem is that container->make() returns MyModel::model(), which is not what we want. We only want that for container->get().

Regardless, models are working correctly even without DI working as expected. Possibly at the cost of some performance due to multiple model instances being used for querying. I don't have a solution for this.

Lets consider this PR a refactor of the DI container. It doesn't solve CActiveRecord issue but its an improvement to the DI config.

PR #3732, opened 2024-02-13
Fixed issue #19399: Survey do not show translated when register

Shnoulle:
Dev: aSurveyInfo with valid language


gabrieljenik (approved):
Code structure looks OK

PR #3734, opened 2024-02-13
MW-946: Aggregated common operations on group starts

lajosarpad:
This is a performance improvement for MW-946, the idea is to improve the speed of the `reset` method by aggregating the survey-level operations that were previously done for each group and thus performing the repetitive operations only once.

Measurements:

Before:

Language Settings took 0.010984897613525 seconds
General Settings took 0.0005490779876709 seconds
URL Params took 9.5367431640625E-7 seconds
Proxy Reset took 12.65425491333 seconds
Template Config took 0.0011730194091797 seconds
Database::actionUpdateSurveyLocaleSettings took 12.67779302597

After:

Language Settings took 0.012439012527466 seconds
General Settings took 0.00066709518432617 seconds
URL Params took 9.5367431640625E-7 seconds
Proxy Reset took 3.8038971424103 seconds
Template Config took 0.001168966293335 seconds
Database::actionUpdateSurveyLocaleSettings took 3.828782081604 seconds

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -6758,9 +6758,10 @@ class LimeExpressionManager
      * @param boolean|null $anonymized - whether anonymized
      * @param int|null $surveyid - the surveyId
      * @param boolean|null $forceRefresh - whether to force refresh of setting variable and token mappings (should be done rarely)
+     * @param boolean|null $postponeAggregation - whether to postpone aggregation
      * @return void
      */
-    public static function StartProcessingGroup($gseq = null, $anonymized = false, $surveyid = null, $forceRefresh = false)
+    public static function StartProcessingGroup($gseq = null, $anonymized = false, $surveyid = null, $forceRefresh = false, $postponeAggregation = false)
     {
         $LEM =& LimeExpressionManager::singleton();
         $LEM->em->StartProcessingGroup(
@@ -6772,18 +6773,37 @@ class LimeExpressionManager
         if (!is_null($gseq)) {
             $LEM->currentGroupSeq = $gseq;
             if (!is_null($surveyid)) {
-                $LEM->setVariableAndTokenMappingsForExpressionManager($surveyid, $forceRefresh, $anonymized);
+                if (!$postponeAggregation) {
+                    $LEM->setVariableAndTokenMappingsForExpressionManager($surveyid, $forceRefresh, $anonymized);
+                }
                 if ($gseq > $LEM->maxGroupSeq) {
                     $LEM->maxGroupSeq = $gseq;
                 }
 
                 if (!$LEM->allOnOnePage || ($LEM->allOnOnePage && !$LEM->processedRelevance)) {
-                    $LEM->ProcessAllNeededRelevance();  // TODO - what if this is called using Survey or Data Entry format?
-                    $LEM->_CreateSubQLevelRelevanceAndValidationEqns();
+                    if (!$postponeAggregation) {
+                        $LEM->ProcessAllNeededRelevance();  // TODO - what if this is called using Survey or Data Entry format?
+                        $LEM->_CreateSubQLevelRelevanceAndValidationEqns();
+                    }
                     $LEM->processedRelevance = true;
                 }
             }
         }
+    }
+
+    /**
+     * Does the group start's common logic in an aggregated manner
+     * @param boolean|null $anonymized - whether anonymized
+     * @param int|null $surveyid - the surveyId
+     * @param boolean|null $forceRefresh - whether to force refresh of setting variable and token mappings (should be done rarely)
+     * @return void
+     */
+    public static function AggregateGroupStart($anonymized = false, $surveyid = null, $forceRefresh = false)
+    {
+        $LEM =& LimeExpressionManager::singleton();
+        $LEM->setVariableAndTokenMappingsForExpressionManager($surveyid, $forceRefresh, $anonymized);
+        $LEM->ProcessAllNeededRelevance();  // TODO - what if this is called using Survey or Data Entry format?
+        $LEM->_CreateSubQLevelRelevanceAndValidationEqns();
     }
 
     /**
--- a/application/models/services/Proxy/ProxyExpressionManager.php
+++ b/application/models/services/Proxy/ProxyExpressionManager.php
@@ -64,14 +64,12 @@ class ProxyExpressionManager
                 $aGroup['gid'],
                 $survey->anonymized != 'Y',
                 $surveyId,
-                true,
                 true
             );
         }
         LimeExpressionManager::AggregateGroupStart(
             $survey->anonymized != 'Y',
-            $surveyId,
-            true
+            $surveyId
         );
         foreach ($aGrouplist as $aGroup) {
             LimeExpressionManager::FinishProcessingGroup();
--- a/application/models/services/Proxy/ProxyExpressionManager.php
+++ b/application/models/services/Proxy/ProxyExpressionManager.php
@@ -64,13 +64,14 @@ class ProxyExpressionManager
                 $aGroup['gid'],
                 $survey->anonymized != 'Y',
                 $surveyId,
-                false,
+                true,
                 true
             );
         }
         LimeExpressionManager::AggregateGroupStart(
             $survey->anonymized != 'Y',
-            $surveyId
+            $surveyId,
+            true
         );
         foreach ($aGrouplist as $aGroup) {
             LimeExpressionManager::FinishProcessingGroup();
--- a/application/models/services/Proxy/ProxyExpressionManager.php
+++ b/application/models/services/Proxy/ProxyExpressionManager.php
@@ -64,6 +64,7 @@ class ProxyExpressionManager
                 $aGroup['gid'],
                 $survey->anonymized != 'Y',
                 $surveyId,
+                false,
                 true
             );
         }
--- a/application/models/services/Proxy/ProxyExpressionManager.php
+++ b/application/models/services/Proxy/ProxyExpressionManager.php
@@ -63,8 +63,16 @@ class ProxyExpressionManager
             LimeExpressionManager::StartProcessingGroup(
                 $aGroup['gid'],
                 $survey->anonymized != 'Y',
-                $surveyId
+                $surveyId,
+                false,
+                true
             );
+        }
+        LimeExpressionManager::AggregateGroupStart(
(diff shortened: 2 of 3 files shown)


c-schmitz:
@ptelu & @Shnoulle  Can you please review this?

Shnoulle:
ProxyExpressionManager->reset seems to replace/reset the ExpressionManager session when something is updated in Survey.

We need to be sure it was done when

- Admin add/remove/rename a question in LimeSurvey
- Admin user click on preview or survey logic file
- Somple user start with newtest=Y (or empty session)

OK with this ?

If yes 

1. why not `LimeExpressionManager::AggregateGroupStart` before [`foreach ($aGrouplist as $aGroup) {` (line](https://github.com/LimeSurvey/LimeSurvey/blob/38093981a2c05e9ec2cfe5f7e305330870eadd41/application/models/services/Proxy/ProxyExpressionManager.php#L62)  ?
2. Use `AggregateGroupStart($survey->anonymized != 'Y', $surveyId, true)` (force refresh) 


I don't find where ProxyExpressionManager->reset is call : you know here ?

lajosarpad:
> ProxyExpressionManager->reset seems to replace/reset the ExpressionManager session when something is updated in Survey.
> 
> We need to be sure it was done when
> 
> * Admin add/remove/rename a question in LimeSurvey
> * Admin user click on preview or survey logic file
> * Simple user start with newtest=Y (or empty session)
> 
> OK with this ?
> 
> If yes
> 
> 1. why not `LimeExpressionManager::AggregateGroupStart` before [`foreach ($aGrouplist as $aGroup) {` (line](https://github.com/LimeSurvey/LimeSurvey/blob/38093981a2c05e9ec2cfe5f7e305330870eadd41/application/models/services/Proxy/ProxyExpressionManager.php#L62)  ?
> 2. Use `AggregateGroupStart($survey->anonymized != 'Y', $surveyId, true)` (force refresh)
> 
> I don't find where ProxyExpressionManager->reset is call : you know here ?
> 
> (i do some test when found some times)

@Shnoulle `ProxyExpressionManager->reset` is being called at `SurveyAggregateService::update` at the section of:

```
        $this->proxyExpressionManager
            ->reset($surveyId);
```

This is being triggered for example from `Database::actionUpdateSurveyLocalSettings`.

"We need to be sure it was done when

Admin add/remove/rename a question in LimeSurvey
Admin user click on preview or survey logic file
Simple user start with newtest=Y (or empty session)
OK with this ?"

I do not have enough context to answer this question. But if we take a look at the code before the patch, we will see that:

- we have a foreach inside application/models/services/Proxy/ProxyExpressionManager.php
- which calls `LimeExpressionManager::StartProcessingGroup` for each group
- and `StartProcecssingGroup` calls the methods that are being called in `AggregateGroupStart`
- these methods in the earlier version are being called after the main group handling of `StartProcessingGroup` has been done

So, these are common costs of the groups and the idea was to avoid repetitively doing these costly operations for each groups, and, instead, to do it once. The reason why we have it after the first foreach is that it seemed to me, given the position of these methods calls inside `StartProcessingGroup` that it is more appropriate to call them after the groups have started. But I'm not sure about this and if it is more appropriate to do it before the groups are initialized, then we can move the call for `AggregateGroupStart` prior the foreach. My thinking was that maybe $fieldMap changes during the group initialization which makes it necessary to call `setVariableAndTokenMappingsForExpressionManager` but I could be wrong about this.

The reason as of why force refresh was left false was that it was left force before the patch and I presumed it's okay that way:

image

What the patch does in short:

- it makes the calls for `setVariableAndTokenMappingsForExpressionManager`, `ProcessAllNeededRelevance` and `_CreateSubQLevelRelevanceAndValidationEqns` only if we want to (this is why the new parameter of `$postponeAggregation` was created)
- we do not want to call these methods for all groups in `ProxyExpressionManager::reset`, we postpone them to make sure we call them exactly once
- and this is why `AggregateGroupStart` was implemented

The tests were successful (I tested by changing the survey settings) and a process that was taking 13-17s prior the patch is now taking ~ 4s for the same survey. So, I think the idea is good, but we need your and ptelu's take on this to make sure all the details are correct. And, if some details need improvement, we will make sure they are being handled as instructed.

Thanks for the review and the questions.

Shnoulle:
> So, these are common costs of the groups and the idea was to avoid repetitively doing these costly operations for each groups, and, instead, to do it once.

Yes : i totally understand this , and clearly : doing it for each group is an issue :)


> given the position of these methods calls inside StartProcessingGroup that it is more appropriate to call them after the groups have started. But I'm not sure about this and if it is more appropriate to do it before the groups are initialized, then we can move the call for AggregateGroupStart prior the foreach.

OK, i test the 2 position, it's seemslogical for me to «update the _VariableAndTokenMappingsForExpressionManager_ for all survey, use new _VariableAndTokenMappingsForExpressionManager_ for all group» :) 

I try to understand better what happen adding some print_r and log.

> The reason as of why force refresh was left false was that it was left false before the patch and I presumed it's okay that way:

OK, surely resetted before (set to empty session)

lajosarpad:
> OK, i test the 2 position, it's seemslogical for me to «update the VariableAndTokenMappingsForExpressionManager for all survey, use new VariableAndTokenMappingsForExpressionManager for all group» :)

Yes, the case you are making is quite plausible. If you can confirm that it is more appropriate to put the call for it before the loop, then I will make the change accordingly. As about force refresh, I have implemented the change you have suggested.

Shnoulle:
> As about force refresh, I have implemented the change you have suggested.

No ? Maybe it's not needed always ???

lajosarpad:
> > As about force refresh, I have implemented the change you have suggested.
> 
> No ? Maybe it's not needed always ???

Should I set it back to false, like it was earlier, before https://github.com/LimeSurvey/LimeSurvey/pull/3734/commits/d3c5ed50907858b4d936f2d626cc07faa2b6eab4 ?

The code before my change used false as force refresh and I preserved the same behavior. I think I misunderstood your point when I changed it to true. I can change it back to false.

Shnoulle:
> The code before my change used false as force refresh and I preserved the same behavior. I think I misunderstood your point when I changed it to true. I can change it back to false.

It's more i totally unsure on this point. Maybe it's already resetted before, if we reset a second time : we do whole StatrSurvey a second time …

Shnoulle:
Some test :  hard to find when this was used … and since when testing : it resetted by session : i think you can remove the reset in `LimeExpressionManager::StartProcessingGroup`

It must be done only one time (why not with `LimeExpressionManager::AggregateGroupStart` ;))

Else: i try to find where it's used … i try updating question, preview, logic file …
 
**I feel incompetent to give an informed opinion here.**

Another issue find in logic file : we reset the setVariableAndTokenMappingsForExpressionManager for each group : here we really need to do it once (but sureyLogicFile not used a lot)

Shnoulle (approved):
No issue when tested, but still unsure on the optimisation

lajosarpad:
> Some test : hard to find when this was used … and since when testing : it resetted by session : i think you can remove the reset in LimeExpressionManager::StartProcessingGroup

It's made optional for the time being because it is used at multiple places and I was worried that I would miss something. If we can apply the same idea, then I will remove it accordingly.

> It must be done only one time (why not with LimeExpressionManager::AggregateGroupStart ;))

Yes, that's the exact idea, so that call is being made after the loop (we can move it before the loop)

> Another issue find in logic file : we reset the setVariableAndTokenMappingsForExpressionManager for each group : here we really need to do it once (but sureyLogicFile not used a lot)

The change in this PR does it once between two loops. So we start and finish n groups, but we call this method once:

```
        foreach ($aGrouplist as $aGroup) {
            LimeExpressionManager::StartProcessingGroup(
                $aGroup['gid'],
                $survey->anonymized != 'Y',
                $surveyId,
                true
            );
        }
        LimeExpressionManager::AggregateGroupStart(
            $survey->anonymized != 'Y',
            $surveyId
        );
        foreach ($aGrouplist as $aGroup) {
            LimeExpressionManager::FinishProcessingGroup();
        }
```

So, we:

- loop the groups and start them
- execute the aggregation once
- loop the groups and finish them

Without the change:

image

and with the change:

image

Basically we do not call some methods for each group and instead we call them in an aggregate manner outside the loop. Sorry for not explaining this clearly.

You can test this by using a large-ish survey and making a change on the text elements tab

Shnoulle:
> You can test this by using a large-ish survey and making a change on the text elements tab

I test with
```
                'vardumpFile' => array(
                    'class' => 'CFileLogRoute',
                    'categories' => 'vardump', // tracevar function
                    'logFile' => 'vardump.log',
                    'enabled' => true,
                ),
```

And adding `tracevar("AggregateGroupStart")` here : https://github.com/LimeSurvey/LimeSurvey/pull/3734/files#diff-fa8d35264aa59c85d37ca2d1756472996499f06fb346865594378102f5f7abc0R6803

Don't see anything … :( 

Else , **in my opinion** best seems to 

1. execute the aggregation once (needed for all group)
2.  loop the groups and start and finish theme

I don't understand why aggregation must be between both …

But yes, now i understand why we don't need to force reset on _loop the groups and start theme_

Shnoulle:
> I tested this with large surveys which had many groups. If you test with smaller surveys, or surveys with single groups, then you might not see significant improvements.

I don't search for improvement, just try to check the reset option ;)

let me check again

Shnoulle:
> let me check again

What i have on tmp/runtime/vardump.log with https://github.com/Shnoulle/LimeSurvey-Shnoulle/tree/bug/MW-946-performance

```
2024/02/16 12:52:11 [trace] [vardump] /media/shnoulle/data/webdev/master/application/helpers/expressions/em_manager_helper.php (3284):
'Force refresh by parameter'
2024/02/16 12:52:12 [trace] [vardump] /media/shnoulle/data/webdev/master/application/helpers/expressions/em_manager_helper.php (3281):
'no update'
```

I update to master and add [tracevar](https://github.com/Shnoulle/LimeSurvey-Shnoulle/commit/351b624f26bcfe800866539efe8359e9abe5466e).

Shnoulle:
Sorry for bad PUSH … 

Else : https://github.com/Shnoulle/LimeSurvey-Shnoulle/commit/16bc58554c7225f4b6915f57bb9a625e6141ac3a and no exeption …

Shnoulle:
`   │ CDbException: CDbCommand failed to execute the SQL statement: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'limesurvey.group_l10ns' doesn't exist`

Must update to master to have green CI

Shnoulle:
Else : avoid one reset by group

PR #3736, opened 2024-02-14
Fixed issue [security] #19087: Account takeover through Import User function

Shnoulle:
Dev: check if user can update user when import

### DIFF
--- a/application/controllers/UserManagementController.php
+++ b/application/controllers/UserManagementController.php
@@ -782,12 +782,12 @@ class UserManagementController extends LSBaseController
             if ($oUser  !== null) {
                 if ($overwriteUsers) {
                     /* Check permission to edit this user */
-                    if (!$oUser->canEdit() || $aUser['uid'] == App()->user->id) {
+                    if (!$oUser->canEdit()) {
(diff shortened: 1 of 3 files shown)


application/controllers/UserManagementController.php
```diff
@@ -775,12 +775,21 @@ public function actionImportUsers(string $importFormat = 'csv'): string
...
 
             if ($oUser  !== null) {
                 if ($overwriteUsers) {
+                    /* Check permission to edit this user */
+                    if (!$oUser->canEdit() || $aUser['uid'] == App()->user->id) {
+                        Yii::app()->setFlashMessage(sprintf(gT("You don't have permission to edit user %s."), $aNewUser['users_name']), 'warning');
+                        continue;
+                    }
+                    /* Check permission to edit self */
+                    if ($aUser['uid'] == App()->user->id) {
```
gabrieljenik, line 789: Will this lines in this if statement ever be executed?
If `$aUser['uid'] == App()->user->id` the script will enter in the line above, right?

Shnoulle: Oh yes : i just want to send a different message burt don't remove the or part …

gabrieljenik (approved):
Code structure looks good

PR #3738, opened 2024-02-14
Fixed issue #18473: Survey listing : multiple call of template

Shnoulle:
Dev: avoid multiple call of TemplateConfiguration::model()->search()

### DIFF
--- a/application/models/Template.php
+++ b/application/models/Template.php
@@ -312,12 +312,8 @@ class Template extends LSActiveRecord
      */
     public static function checkIfTemplateExists($sTemplateName)
     {
-        $atemplateList = self::getTemplateList();
-        // isset is faster, and we need a value
-        if (isset($atemplateList[$sTemplateName])) {
-            return true;
-        }
-        return false;
+        // isset is faster, and we need a value, no need var here
+        return isset(self::getTemplateList()[$sTemplateName]);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Shnoulle:
Clearly need a better solution here : 

1. no need to call _search_ to have only the list of templet
2. need only the template name from DB

Shnoulle:
Ok, better fix now !

application/models/Template.php
```diff
@@ -470,23 +467,32 @@ public static function getTemplatesFileFolder($sTemplateName = "")
...
+            return $aTemplateList;
+        }
         $aTemplateList = array();
-
-        $oTemplateList = TemplateConfiguration::model()->search();
-        $oTemplateList->setPagination(false);
-
-        foreach ($oTemplateList->getData() as $oTemplate) {
-            $aTemplateList[$oTemplate->template_name] =  (self::isStandardTemplate($oTemplate->template_name)) ?  Yii::app()->getConfig("standardthemerootdir") . DIRECTORY_SEPARATOR . $oTemplate->template->folder : Yii::app()->getConfig("userthemerootdir") . DIRECTORY_SEPARATOR . $oTemplate->template->folder;
+        $aTemplateList = [];
```
lajosarpad, line 481: Why do you create this array twice?

lajosarpad (changes requested):
Overall this looks good, we have a duplicated variable initialization though and we will need to get rid of that

application/models/Template.php
Shnoulle: Error :)

lajosarpad (approved):

c-schmitz (approved):

PR #3739, opened 2024-02-14
Fixed issue: Slow performance when saving a survey with more than one  language

c-schmitz:
Dev Instead of one query per item (questions, answers, assessments, etc) per language it is now only one query per language, saving alot of query time 
Dev Also removed a question language consistency check (fixInvalidQuestions) which should not be called here

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -3706,25 +3706,25 @@ function fixLanguageConsistency($sid, $availlangs = '', $baselang = '')
     }
     if (trim($availlangs) != '') {
         $availlangs = sanitize_languagecodeS($availlangs);
-        $langs = explode(" ", (string) $availlangs);
-        if ($langs[count($langs) - 1] == "") {
-            array_pop($langs);
+        $languagesToCheck = explode(" ", (string) $availlangs);
+        if ($languagesToCheck[count($languagesToCheck) - 1] == "") {
+            array_pop($languagesToCheck);
         }
         // If base language is in the list, remove it
-        if (($key = array_search($baselang, $langs)) !== false) {
-            unset($langs[$key]);
+        if (($key = array_search($baselang, $languagesToCheck)) !== false) {
+            unset($languagesToCheck[$key]);
         }
     } else {
-        $langs = Survey::model()->findByPk($sid)->additionalLanguages;
+        $languagesToCheck = Survey::model()->findByPk($sid)->additionalLanguages;
     }
-    if (count($langs) == 0) {
+    if (count($languagesToCheck) == 0) {
         return true; // Survey only has one language
     }
     $quotedGroups = Yii::app()->db->quoteTableName('{{groups}}');
-    $query = "SELECT * FROM $quotedGroups g JOIN {{group_l10ns}} ls ON ls.gid=g.gid WHERE sid='{$sid}' AND language='{$baselang}'  ";
+    $query = "SELECT g.gid, ls.group_name, ls.description FROM $quotedGroups g JOIN {{group_l10ns}} ls ON ls.gid=g.gid WHERE sid={$sid} AND language='{$baselang}'  ";
     $result = Yii::app()->db->createCommand($query)->query();
-    foreach ($langs as $lang) {
-        $query = "SELECT gid FROM {{group_l10ns}} WHERE language='{$lang}'";
+    foreach ($languagesToCheck as $lang) {
+        $query = "SELECT g.gid FROM $quotedGroups g JOIN {{group_l10ns}} ls ON ls.gid=g.gid WHERE sid={$sid} AND language='{$lang}'  ";
         $gresult = Yii::app()->db->createCommand($query)->queryColumn();
         foreach ($result->readAll() as $group) {
             if (!in_array($group['gid'], $gresult)) {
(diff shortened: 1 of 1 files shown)


application/helpers/common_helper.php
```diff
@@ -3706,25 +3706,25 @@ function fixLanguageConsistency($sid, $availlangs = '', $baselang = '')
...
         return true; // Survey only has one language
     }
     $quotedGroups = Yii::app()->db->quoteTableName('{{groups}}');
-    $query = "SELECT * FROM $quotedGroups g JOIN {{group_l10ns}} ls ON ls.gid=g.gid WHERE sid='{$sid}' AND language='{$baselang}'  ";
+    $query = "SELECT g.gid, ls.group_name, ls.description FROM $quotedGroups g JOIN {{group_l10ns}} ls ON ls.gid=g.gid WHERE sid={$sid} AND language='{$baselang}'  ";
     $result = Yii::app()->db->createCommand($query)->query();
-    foreach ($langs as $lang) {
-        $query = "SELECT gid FROM {{group_l10ns}} WHERE language='{$lang}'";
+    foreach ($languagesToCheck as $lang) {
+        $query = "SELECT g.gid FROM $quotedGroups g JOIN {{group_l10ns}} ls ON ls.gid=g.gid WHERE sid={$sid} AND language='{$lang}'  ";
```
Shnoulle, line 3727: :+1:

application/helpers/common_helper.php
```diff
@@ -3737,18 +3737,16 @@ function fixLanguageConsistency($sid, $availlangs = '', $baselang = '')
...
-        foreach ($result as $question) {
-            foreach ($langs as $lang) {
-                $query = "SELECT count(qid) FROM {{question_l10ns}} WHERE qid='{$question['qid']}' AND language='{$lang}'";
-                $gresult = Yii::app()->db->createCommand($query)->queryScalar();
-                if ($gresult < 1) {
+        foreach ($languagesToCheck as $lang) {
+            $query = "SELECT q.qid FROM {{questions}} q JOIN {{question_l10ns}} ls ON ls.qid=q.qid WHERE sid={$sid} AND language='{$lang}'";
+            $gresult = Yii::app()->db->createCommand($query)->queryColumn();
+            foreach ($result->readAll() as $question) {
+                if (!in_array($question['qid'], $gresult)) {
```
Shnoulle, line 3749: OK ! Move to one query !

No need to select `sid` and `language` on MSSQL ? I have some issue where value in condition are not selected ?

Shnoulle (approved):
OK !

Great improvment !

One question : Is it not necessary to select the columns used in JOIN or in conditions MSSQL database ? 

You still have a MSQL database for testing ?

Shnoulle:
Can we have it for 5.X too ?

application/helpers/common_helper.php
```diff
@@ -3737,18 +3737,16 @@ function fixLanguageConsistency($sid, $availlangs = '', $baselang = '')
...
+    $query = "SELECT q.qid, ls.question, ls.help FROM {{questions}} q JOIN {{question_l10ns}} ls ON ls.qid=q.qid WHERE sid={$sid} AND language='{$baselang}'";
+    $result = Yii::app()->db->createCommand($query)->query();
     if (count($result) > 0) {
-        foreach ($result as $question) {
-            foreach ($langs as $lang) {
-                $query = "SELECT count(qid) FROM {{question_l10ns}} WHERE qid='{$question['qid']}' AND language='{$lang}'";
-                $gresult = Yii::app()->db->createCommand($query)->queryScalar();
-                if ($gresult < 1) {
+        foreach ($languagesToCheck as $lang) {
+            $query = "SELECT q.qid FROM {{questions}} q JOIN {{question_l10ns}} ls ON ls.qid=q.qid WHERE sid={$sid} AND language='{$lang}'";
```
gabrieljenik, line 3746: If you would like to minimize queries even more, maybe the where clause could be something like `language in (...)`

application/helpers/common_helper.php
```diff
@@ -3780,19 +3780,16 @@ function fixLanguageConsistency($sid, $availlangs = '', $baselang = '')
...
-    $query = "SELECT * FROM {{assessments}} WHERE sid='{$sid}' AND language='{$baselang}'";
+    $query = "SELECT id, sid, scope, gid, name, minimum, maximum, message FROM {{assessments}} WHERE sid='{$sid}' AND language='{$baselang}'";
     $result = Yii::app()->db->createCommand($query)->query();
-    foreach ($result->readAll() as $assessment) {
-        foreach ($langs as $lang) {
-            $query = "SELECT count(id) FROM {{assessments}} WHERE sid='{$sid}' AND id='{$assessment['id']}' AND language='{$lang}'";
-            $gresult = Yii::app()->db->createCommand($query)->queryScalar();
-            if ($gresult < 1) {
+    foreach ($languagesToCheck as $lang) {
+        $query = "SELECT id FROM {{assessments}} WHERE sid='{$sid}' AND language='{$lang}'";
```
gabrieljenik, line 3789: Same applies, maybe we can query all languages at once.

application/helpers/common_helper.php
Shnoulle: Unsure on this point : if you have 100 questions with 20 language with Chinese and Arabic and long sentence : it can take a bunch of memory and broken a lot of instance.

I think it(rs best to separate each request by language

application/helpers/common_helper.php
Shnoulle: Same comment about size of final data ;) the more memory issue

application/helpers/common_helper.php
c-schmitz: Yeah, I thought about this too, but I think the number of queries reduced to the number of languages is fine here.
You will rarely see a survey with 50 languages where this would really start to matter.
For that case we could maybe introduce an arbitrary limit on number of languages per survey? 
I could vet our customer base and see what is the current maximum.

application/helpers/common_helper.php
c-schmitz: See my comment above.

c-schmitz:
> OK !
> 
> Great improvment !
> 
> One question : Is it not necessary to select the columns used in JOIN or in conditions MSSQL database ?
> 
> You still have a MSQL database for testing ?

I did select the columns names instead of asterisk to minimize memory usage - I will check in MSSQL before merging.

application/helpers/common_helper.php
Shnoulle: I see maximum of 23 language. But i see a lot of surveys with more than 300 questions ;)

Shnoulle: > You will rarely see a survey with 50 languages where this would really start to matter.

yes but increasing time is cheaper than adding memory (i already increase timeout before each update process to allow check integrity before update)

Shnoulle:
> I did select the columns names instead of asterisk to minimize memory usage - I will check in MSSQL before merging.

Oh yes ! Less data is always good, but i remind an issue with such request and MSSQL some month ago.

application/helpers/common_helper.php
gabrieljenik: I wouldn't limit the number of languages, at least not in the Community Version.
On the cloud version could be something limited by a plan :)

I have seen customers using many (maybe +50) languages.
If you limit the number of languages (at any number of languages), how will they work their surveys?

> Unsure on this point : if you have 100 questions with 20 language with Chinese and Arabic and long sentence : it can take a bunch of memory and broken a lot of instance.

What do you mean by "long sentences"? Only qid is selected.
100 questions x 20 languages = 2000 rows of a single integer. Doens't sound like a lot.
Am I missing something?

Shnoulle: > What do you mean by "long sentences"? Only qid is selected.
> 100 questions x 20 languages = 2000 rows of a single integer. Doens't sound like a lot.

You're right !

TonisOrmisson:
oh, thanks, this fixes issue #19066, I will close that

PR #3740, opened 2024-02-15
Fixed issue #CT-652: system slowness when saving survey settings

ptelu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected group processing logic in data entry, question administration, and PDF export operations to properly compute group sequence values.
  
* **Performance**
  * Improved question theme lookup performance through enhanced caching mechanism.


application/helpers/expressions/em_manager_helper.php
```diff
@@ -8233,11 +8233,18 @@ public static function UpgradeQuestionAttributes($changeDB = false, $iSurveyID =
...
      * @param string|null $lang
+     *
      * @return array
+     * @throws CException
+     * @throws EmCacheException
      */
     private function getQuestionAttributesForEM($surveyid = 0, $qid = 0, $lang = '')
     {
-        $cacheKey = 'getQuestionAttributesForEM_' . $surveyid . '_' . $qid . '_' . $lang;
+        $gid = 0;
```
Shnoulle, line 8243: $gid is in parameters ?

application/models/QuestionTheme.php
```diff
@@ -837,9 +837,11 @@ public static function getQuestionXMLPathForBaseType($type)
     {
         static $questionThemeCache = [];
```
Shnoulle, line 838: I dislike to use Cache for Static var … Static is  not caching ;)

Shnoulle (approved):
Maybe check  more on

`if (!empty($this->currentGroupSeq) && $this->currentGroupSeq > 0) {`

Something like
`if($surveyId && LimeExpressionManager::getLEMsurveyId() == $surveyId && (!empty($this->currentGroupSeq) && $this->currentGroupSeq > 0 ) {`

It was call without qid on setVariableAndTokenMappingsForExpressionManager

And this function is used in Logic file.

Shnoulle (commented):
Maybe check  more on

`if (!empty($this->currentGroupSeq) && $this->currentGroupSeq > 0) {`

Something like
`if($surveyId && LimeExpressionManager::getLEMsurveyId() == $surveyId && (!empty($this->currentGroupSeq) && $this->currentGroupSeq > 0 ) {`

It was call without qid on setVariableAndTokenMappingsForExpressionManager

And this function is used in Logic file.

(updated to comment : think it can show not remlated Group question in Admin part.

I public part : it's always OK, and can really improve.

application/models/QuestionTheme.php
ptelu: yeah makes sense, cant use it everywhere php will run out of memory eventually when we use it to much. Will look at yii's caching

Shnoulle: Currently : it's a static var :). I don't ask to put it in cache (maybe in future), just more clear name :)

lajosarpad: I also propose the creation of a `protected static $questionThemeCache = [];` data member to make the file more readable and the cache more accessible.

lajosarpad (approved):
This looks good, but I would avoid using `static` variables for caching.

application/models/QuestionTheme.php
Shnoulle: `protected static $questionThemeStatic = []` ;)

Shnoulle (approved):

c-schmitz:
@coderabbitai review

application/helpers/expressions/em_manager_helper.php
```diff
@@ -8377,6 +8252,15 @@ private function getQuestionAttributesForEM($surveyid = 0, $qid = 0, $lang = '')
         if (is_null($lang)) {
             $lang = '';
         }
+        if (is_null($gid) && $gid < 0) {
```
Shnoulle, line 8255: `if (is_null($gid) || $gid < 0) {` else always false

PR #3743, opened 2024-02-15
Fixed #19217: PHP8 Error Type NULL

gabrieljenik:
* Fixed: 19217 Update Survey.php

$attdescriptiondata = decodeTokenAttributes($this->attributedescriptions ?? ''); change to
$attdescriptiondata = decodeTokenAttributes($this->attributedescriptions ?? []); In PHP 7 it is a warning
in PHP 8 it's a 500! parameter key() must be an array!!

* Fixed 10217: Update common_helper.php

$aSavedExtraTokenFields = Survey::model()->findByPk($surveyid)->tokenAttributes; change to:
$aSavedExtraTokenFields = Survey::model()->findByPk($surveyid)->tokenAttributes ?? []; In PHP 7 it is a warning
in PHP 8 it's a 500! $aSavedExtraTokenFields is not allowed to be NULL anymore! in function array_intersect_key()!

* Update common_helper.php

Corrected code line 2946

* #19217: Update Survey.php

revert back to from ?? [] to ?? '' on line 667

* #19217: Update Survey.php

Extra check on first element of array
If it is not an array set array to NULL

* #19217: Update Survey.php

Final edit after line 667:
Check on array after reset(array)

* #19217: Update Survey.php

And some missing spaces...
this one should pass!!!

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -2927,7 +2927,7 @@ function getTokenFieldsAndNames($surveyid, $bOnlyAttributes = false)
     );
 
     $aExtraTokenFields = getAttributeFieldNames($surveyid);
-    $aSavedExtraTokenFields = Survey::model()->findByPk($surveyid)->tokenAttributes;
+    $aSavedExtraTokenFields = Survey::model()->findByPk($surveyid)->tokenAttributes ?? [];
 
     // Drop all fields that are in the saved field description but not in the table definition
     $aSavedExtraTokenFields = array_intersect_key($aSavedExtraTokenFields, array_flip($aExtraTokenFields));
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -671,8 +671,10 @@ class Survey extends LSActiveRecord implements PermissionInterface
      */
     public function getTokenAttributes()
     {
-        $attdescriptiondata = decodeTokenAttributes($this->attributedescriptions);
-
+        $attdescriptiondata = decodeTokenAttributes($this->attributedescriptions ?? '');
+        if (!is_array(reset($attdescriptiondata))) {
+            $attdescriptiondata = null;
+        }
         // Catches malformed data
         if ($attdescriptiondata && strpos(key(reset($attdescriptiondata)), 'attribute_') === false) {
             // don't know why yet but this breaks normal tokenAttributes functionning


application/models/Survey.php
```diff
@@ -671,8 +671,10 @@ public function getAllLanguages()
      */
     public function getTokenAttributes()
     {
-        $attdescriptiondata = decodeTokenAttributes($this->attributedescriptions);
-
+        $attdescriptiondata = decodeTokenAttributes($this->attributedescriptions ?? '');
+        if (!is_array(reset($attdescriptiondata))) {
```
Shnoulle, line 675: Why not check it in https://github.com/LimeSurvey/LimeSurvey/blob/a089c468838c1ff90603b78c44a5296817c5e16b/application/helpers/common_helper.php#L4764 directly ?

Check if the array return is valid ? If not return an empty array ?

Shnoulle: Ok, 

No and more : maybe must move https://github.com/LimeSurvey/LimeSurvey/blob/a089c468838c1ff90603b78c44a5296817c5e16b/application/helpers/common_helper.php#L4764 in this function or create a private function here. 

And deprecate common_helper function

application/helpers/common_helper.php
```diff
@@ -2927,7 +2927,7 @@ function getTokenFieldsAndNames($surveyid, $bOnlyAttributes = false)
     );
 
     $aExtraTokenFields = getAttributeFieldNames($surveyid);
-    $aSavedExtraTokenFields = Survey::model()->findByPk($surveyid)->tokenAttributes;
+    $aSavedExtraTokenFields = Survey::model()->findByPk($surveyid)->tokenAttributes ?? [];
```
Shnoulle, line 2930: https://github.com/LimeSurvey/LimeSurvey/blob/1a6503c0b46d83538afeb84768e5ea927dbc680f/application/models/Survey.php#L672 must return an array, if it's not the case : it's an issue. No ?

And i don't see in function how  it can return something other than an array https://github.com/LimeSurvey/LimeSurvey/blob/1a6503c0b46d83538afeb84768e5ea927dbc680f/application/models/Survey.php#L718-L734

Shnoulle (commented):
Survey::model()->findByPk($surveyid)->tokenAttributes **must** return a valid array, always !

gabrieljenik:
> Survey::model()->findByPk($surveyid)->tokenAttributes **must** return a valid array, always !

I was just asked to cherry pick something is already aproved and merged on master.

Shnoulle:
> I was just asked to cherry pick something is already aproved and merged on master. I haven't went through the solution. I understand is already approved and working.

https://github.com/LimeSurvey/LimeSurvey/pull/3666
 
> Could get into it if needed.

Yep, master have an issue here : https://github.com/LimeSurvey/LimeSurvey/blob/41a496f6ca6c27b583548ef26256edbecb3d2885/application/models/Survey.php#L670


I don't know here ? Fix both ?

PR #3746, opened 2024-02-19
MW-1027: Improved upon display_first_page to not force refresh, so we only refresh if needed

lajosarpad:
We can test this with large surveys

Huge 15:48s - 7.61s
Big 1.52s - 0.764s


Shnoulle:
Really unsure here : if we display firstpage : we need a updated data … the expressoion session muts be up to date.

Shnoulle:
My opinion : better to move to a more whole system : using some static var somewhere to know if we already reset the Expression Manager during process.

Shnoulle:
> My opinion : better to move to a more whole system : using some static var somewhere to know if we already reset the Expression Manager during process.

PS : we must check too here : 

1. Display first page :(EM resetted : OK)
2. Move next : no need EM reset (?)
3. Move previous : EM resetted : i think it's not OK, and must be avoided … 


And this commit are the solution :( . Like i write : **unsure** …

PR #3747, opened 2024-02-19
MW-1028: Improved performance of survey logic file

lajosarpad:
Huge survey
survey logic file 3.5m - 21.86s
survey logic printable 3.4m - 28s
Big Survey
survey logic file 12.63s - 4.99s
survey logic printable 13s - 6s

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -3276,7 +3276,7 @@ class LimeExpressionManager
         if (isset($_SESSION['LEMforceRefresh'])) {
             unset($_SESSION['LEMforceRefresh']);
             $forceRefresh = true;
-        } elseif ($forceRefresh === false && !empty($this->knownVars) && ((!$this->sPreviewMode) || ($this->sPreviewMode === 'database'))) {
+        } elseif ($forceRefresh === false && !empty($this->knownVars) && ((!$this->sPreviewMode) || ($this->sPreviewMode === 'database') || ($this->sPreviewMode === 'logic'))) {
             return false;   // means that those variables have been cached and no changes needed
         }
         $now = microtime(true);


Shnoulle:
The survey logic file is THE place where we need an updated expression manager.

I think it's already done by `$_SESSION['LEMforceRefresh']` actually. But is it a good solution to disable evaluation when previewmode are logic, same with database? And the best solution seems to be this one:

- All place where LimeExpressionManager->sPreviewMode = 'logic', replace by a LimeExpressionManager::SetDirtyFlag()
- All place where LimeExpressionManager->sPreviewMode = 'database', replace by a LimeExpressionManager::SetDirtyFlag()

We can do the same for question and questiongroup and deprecate sPreviewMode after. using only LimeExpressionManager::SetDirtyFlag() (or another function)

AmeniChakroun:
Please How to run the code (command i need)?

AmeniChakroun:
the code is paid or free (license) ?

lajosarpad:
> The survey logic file is THE place where we need an updated expression manager.
> 
> I think it's already done by `$_SESSION['LEMforceRefresh']` actually. But is it a good solution to disable evaluation when previewmode are logic, same with database? And the best solution seems to be this one:
> 
> * All place where LimeExpressionManager->sPreviewMode = 'logic', replace by a LimeExpressionManager::SetDirtyFlag()
> * All place where LimeExpressionManager->sPreviewMode = 'database', replace by a LimeExpressionManager::SetDirtyFlag()
> 
> We can do the same for question and questiongroup and deprecate sPreviewMode after. using only LimeExpressionManager::SetDirtyFlag() (or another function)

The idea is good. Currently we need to fix the performance issues so that Limesurvey will be quicker for the customers. We will try to return to code quality issues as well, as per the suggestion. We will need to be careful while doing so, because parts of the code rely on these preview values, such as

```
            if ($this->sPreviewMode == 'logic') {
                $this->subQrelInfo[$questionNum][$rowdivid]['aWarnings'] = $aWarnings;
            }
```

so in order to implement this improvement, we will also need to do some refactoring and further tests. I initiated discussions about this in the team.

lajosarpad:
> Please How to run the code (command i need)?

@AmeniChakroun in order to try this improvement, you can

1. Visit <yourroot>/index.php?r=surveyAdministration/listsurveys  (the survey list page)
2. Click on a survey
3. Click on Tools
4. Click on Survey logic file
5. Wait for the page to load and measure its loading length
image

The second test is to simply click on the Open printable view button on the survey logic file page: 
image

lajosarpad:
> the code is paid or free (license) ?

This code is free, you can find the branch of this fix at https://github.com/LimeSurvey/LimeSurvey/tree/bug/MW-1028-survey-logic-file-performance even if you are not logged in to GitHub and you can clone it. You can test this by running

```
git pull
git checkout bug/MW-1028-survey-logic-file-performance
```

in your local repository and then

```
git checkout master
```

if you want to test the current performance in the way I described in my previous comment and, you can repeat the same test with the better performing new code if you

```
git checkout bug/MW-1028-survey-logic-file-performance
```

c-schmitz (approved):

PR #3751, opened 2024-02-21
MW-1031: getQuestionQttributesForEMPerformance

lajosarpad:
This is a performance optimization that affects survey settings tab loading.

Measurements:

Text elements	6.98s	0.49s
Privacy policy	 	6.93s	0.65s
Presentation	 	6.8s	0.45s
Participant settings	 	6.95s	0.45s
Notifications & data	 	6.87s	0.43s
Publication & access	 	6.79s	0.46s
Panel integration	 	8.23s	1.12s
Resources	 	6.89s	0.44s
Simple plugins	 	10.67s	0.46s

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -8273,29 +8273,29 @@ report~numKids > 0~message~{name}, you said you are {age} and that you have {num
 #                return $aStaticQuestionAttributesForEM[$surveyid][0][$lang][$qid];
 #            }
         if ($qid) {
-            $oQids = Question::model()->findAll(
+            $oQuestions = Question::model()->findAll(
                 [
                     'condition' => "qid=:qid and parent_qid=0",
                     'params'    => [':qid' => $qid]
                 ]
             );
         } elseif ($surveyid) {
-            $oQids = Question::model()->findAll(
+            $oQuestions = Question::model()->findAll(
                 [
                     'condition' => "sid=:sid and parent_qid=0",
                     'params'    => [':sid' => $surveyid]
                 ]
             );
         } else {
-            $oQids = Question::model()->findAll(
+            $oQuestions = Question::model()->findAll(
                 [
                     'condition' => "parent_qid=0",
                 ]
             );
         }
         $aQuestionAttributesForEM = [];
-        foreach ($oQids as $oQid) {
-            $aAttributesValues = QuestionAttribute::model()->getQuestionAttributes($oQid, $lang);
+        foreach ($oQuestions as $oQuestion) {
+            $aAttributesValues = QuestionAttribute::model()->getQuestionAttributes($oQuestion, $lang);
             // Change array lang to value
             foreach ($aAttributesValues as &$aAttributeValue) {
                 if (is_array($aAttributeValue)) {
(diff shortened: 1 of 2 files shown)


c-schmitz (changes requested):
Code looks fine, but something broke in the tests. Please check.

Shnoulle (approved):
Avoid call of oQuestion if already loaded :+1:

Shnoulle:
About QuestionAttribute : a big difference between 3 and 4 are here too.

In 3 QuestionAttributedefinition was static, it's not the case (even for core part) in 4.

See [18475: Multiple event newQuestionAttributes happen](https://bugs.limesurvey.org/view.php?id=18475)

I think a good improvement can create core + plugin static function/var + update it with theme after ?

application/models/QuestionAttribute.php
```diff
@@ -219,16 +219,21 @@ public function setMultiple($iSid, $aQids, $aAttributesToUpdate, $aValidQuestion
...
      * @access public
-     * @param int $iQuestionID
+     * @param int|Question $q
      * @param string $sLanguage restrict to this language (if null $oQuestion->survey->allLanguages will be used)
      * @return array|false
      *
      * @throws CException throws exception if questiontype is null
      */
-    public function getQuestionAttributes($iQuestionID, $sLanguage = null)
+    public function getQuestionAttributes($q, $sLanguage = null)
```
olleharstedt, line 228: It wasn't possible to change all call locations to this method to always only pass object as input?

lajosarpad: We are currently working on performance issues. We could refactor the further references, but that would require a separate ticket and an approval

image

We can only change the parameter to accept question objects if we change and test all references. That would take a while and in the meantime the users would not get the benefits of our results. If we can create a ticket to refactor all references and then the parameter of this method, that would be great, I agree with you. But even then we should break it up to smaller PRs with the last one being the change of the parameter to not allow integers.

olleharstedt: Wow that's a lot of call sites.. Got it. :sweat_smile:

lajosarpad:
> Code looks fine, but something broke in the tests. Please check.

All green now

application/helpers/expressions/em_manager_helper.php
```diff
@@ -8275,36 +8275,27 @@ private function getQuestionAttributesForEM($surveyid = 0, $qid = 0, $lang = '')
         if ($qid) {
             $oQids = Question::model()->findAll(
                 [
-                    'select'    => 'qid',
-                    'group'     => 'qid',
-                    'distinct'  => true,
                     'condition' => "qid=:qid and parent_qid=0",
```
Shnoulle, line 8278: We better understand why the name it's $oQids and not $oQuestion : i think it's a good idea to rename this variable here.

Shnoulle (commented):
Maybe rename variable oQids to oQuestion ?

c-schmitz (approved):

application/helpers/expressions/em_manager_helper.php
Shnoulle: Thank you

lajosarpad:
@Dennis1993 @c-schmitz the change was implemented accordingly

olleharstedt:
> Maybe rename variable oQids to oQuestion ?

Or get rid of Hungarian notation and just use `$question` :sweat_smile:

Shnoulle:
> > Maybe rename variable oQids to oQuestion ?
> 
> Or get rid of Hungarian notation and just use `$questions` 😅

One or another, but no oQid ;)

PR #3753, opened 2024-02-22
Fixed issue #19435: Input validation does not work correctly with input numeric when reload response

Shnoulle:
Fixed issue #19435: Input validation does not work correctly with input numeric when reload response
Dev: fix numeric value when reload from DB directly

### DIFF
--- a/application/helpers/frontend_helper.php
+++ b/application/helpers/frontend_helper.php
@@ -81,7 +81,7 @@ function loadanswers()
     $submitdate = $oResponses->submitdate;
     $aRow = $oResponses->attributes;
     foreach ($aRow as $column => $value) {
-        tracevar($column);
+        if ($column === "token") {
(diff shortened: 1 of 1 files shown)


application/helpers/frontend_helper.php
```diff
@@ -97,22 +97,27 @@ function loadanswers()
...
+                        if($value[0] === ".") {
+                            $value = "0".$value;
+                        }
+                        if (strpos($value, ".") !== false) {
+                            $value = rtrim(rtrim($value, "0"), ".");
+                        }
+                    }
                 }
+                $_SESSION['survey_' . $surveyid][$column] = $value;
                 if (isset($token) && !empty($token)) {
```
Shnoulle, line 121: I don't update unrelated part here, but I think this 3 lines mus be removed

PR #3755, opened 2024-02-22
Fixed issue #16287: Contentious wording in the application texts

c-schmitz:
Dev Replaced Whitelist with Allowlist & replaced Blacklist with Blocklist 
Dev No function names or variables were renamed. This can be done on a major new version, only


Shnoulle:
> Dev No function names or variables were renamed. This can be done on a major new version, only

Must create a todo for 7

Shnoulle (approved):

PR #3759, opened 2024-02-26
Bug/mw 1036 survey permission improvements

lajosarpad:
Caching privileges that we have already checked for to prevent having to do it again

Survey1: 1.87s - 1.68s
Survey2: 11.76s - 10.66s
Survey3: 5.29s - 5.07

Test steps:

- open a survey in edit mode
- click on Overview questions & groups

### DIFF
--- a/application/models/Permission.php
+++ b/application/models/Permission.php
@@ -35,6 +35,9 @@ class Permission extends LSActiveRecord
     /* @var array[]|null The global base Permission LimeSurvey installation */
     protected static $aGlobalBasePermissions;
 
+    /* @var array[] The already loaded survey permissions */
+    protected static $aCachedSurveyPermissions = [];
+
     /** @inheritdoc */
     public function tableName()
     {
@@ -703,11 +706,20 @@ class Permission extends LSActiveRecord
      */
     public function hasSurveyPermission($iSurveyID, $sPermission, $sCRUD = 'read', $iUserID = null)
     {
-        $oSurvey = Survey::Model()->findByPk($iSurveyID);
-        if (!$oSurvey) {
-            return false;
+        if (isset(self::$aCachedSurveyPermissions[$iSurveyID][$sPermission][$sCRUD][$iUserID])) {
+            return self::$aCachedSurveyPermissions[$iSurveyID][$sPermission][$sCRUD][$iUserID];
         }
-        return $oSurvey->hasPermission($sPermission, $sCRUD, $iUserID);
+        if (!isset(self::$aCachedSurveyPermissions[$iSurveyID])) {
+            self::$aCachedSurveyPermissions[$iSurveyID] = [];
+        }
+        if (!isset(self::$aCachedSurveyPermissions[$iSurveyID][$sPermission])) {
+            self::$aCachedSurveyPermissions[$iSurveyID][$sPermission] = [];
+        }
+        if (!isset(self::$aCachedSurveyPermissions[$iSurveyID][$sPermission][$sCRUD])) {
+            self::$aCachedSurveyPermissions[$iSurveyID][$sPermission][$iUserID] = [];
+        }
+        $oSurvey = Survey::Model()->findByPk($iSurveyID);
+        return self::$aCachedSurveyPermissions[$iSurveyID][$sPermission][$sCRUD][$iUserID] = ($oSurvey ? $oSurvey->hasPermission($sPermission, $sCRUD, $iUserID) : false);
     }
 
     /**
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -2011,19 +2011,20 @@ class ExpressionManager
             } else {
                 ++$this->substitutionNum;
                 $expr = $this->ExpandThisVar(substr((string) $stringPart[0], 1, -1));
+                $prettyPrinted = $this->GetPrettyPrintString();
                 if ($this->RDP_Evaluate($expr, false, $this->resetErrorsAndWarningsOnEachPart)) {
                     $resolvedPart = $this->GetResult();
                 } else {
                     // show original and errors in-line only if user have the right to update survey content
                     if ($this->sid && Permission::model()->hasSurveyPermission($this->sid, 'surveycontent', 'update')) {
-                        $resolvedPart = $this->GetPrettyPrintString();
+                        $resolvedPart = $prettyPrinted;
                     } else {
                         $resolvedPart = '';
                     }
                 }
                 $onpageJsVarsUsed = $this->GetOnPageJsVarsUsed();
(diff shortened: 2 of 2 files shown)


application/models/Permission.php
```diff
@@ -703,11 +706,20 @@ public function getButtons(): string
      */
     public function hasSurveyPermission($iSurveyID, $sPermission, $sCRUD = 'read', $iUserID = null)
     {
-        $oSurvey = Survey::Model()->findByPk($iSurveyID);
```
Shnoulle, line 706: It's already a static variables

application/models/Permission.php
```diff
@@ -703,11 +706,20 @@ public function getButtons(): string
...
-        if (!$oSurvey) {
-            return false;
+        if (isset(self::$aCachedSurveyPermissions[$iSurveyID][$sPermission][$sCRUD][$iUserID])) {
+            return self::$aCachedSurveyPermissions[$iSurveyID][$sPermission][$sCRUD][$iUserID];
+        }
+        if (!isset(self::$aCachedSurveyPermissions[$iSurveyID])) {
+            self::$aCachedSurveyPermissions[$iSurveyID] = [];
         }
-        return $oSurvey->hasPermission($sPermission, $sCRUD, $iUserID);
```
Shnoulle, line 710: Already a partial static variables

https://github.com/LimeSurvey/LimeSurvey/blob/c10bac29fff26a55257a5b3d8997c18672e40eee/application/models/Permission.php#L619

c-schmitz (approved):

application/models/Permission.php
Shnoulle: Maybe we can move add

```
if (isset($aPermissionStatic[$iEntityID][$sEntityName][$iUserID][$sPermission][$sCRUD])) {
return $aPermissionStatic[$iEntityID][$sEntityName][$iUserID][$sPermission][$sCRUD];
}
```

on top of https://github.com/LimeSurvey/LimeSurvey/blob/c10bac29fff26a55257a5b3d8997c18672e40eee/application/models/Permission.php#L548

(and add some set elsewhere https://github.com/LimeSurvey/LimeSurvey/blob/c10bac29fff26a55257a5b3d8997c18672e40eee/application/models/Permission.php#L558 https://github.com/LimeSurvey/LimeSurvey/blob/c10bac29fff26a55257a5b3d8997c18672e40eee/application/models/Permission.php#L566 https://github.com/LimeSurvey/LimeSurvey/blob/c10bac29fff26a55257a5b3d8997c18672e40eee/application/models/Permission.php#L571 https://github.com/LimeSurvey/LimeSurvey/blob/c10bac29fff26a55257a5b3d8997c18672e40eee/application/models/Permission.php#L595

Shnoulle:
It's an API broker : beforeHasPermission happen each time a permission is done before.

I don't have plugin using this, but it can happen.

application/models/Permission.php
lajosarpad: @Shnoulle this seems to be a good idea, worth checking, to apply the idea to permission checking in general. But, if costly things happen between survey permission checking and the generic permission checking, then this change might make sense as well. It's worth investigating.

lajosarpad:
> It's an API broker : beforeHasPermission happen each time a permission is done before.
> 
> I don't have plugin using this, but it can happen.

If I understand you correctly, then you mean that some plugins may rely on the same permission being checked multiple times and performing some action before that occurs. And your worry is that if we check the permission of surveycontent for a survey the first time we need to do an action, then we might miss executing it.

But isn't this sounding like bad design? Do we need to execute permission check for permissions that we already checked? Do we know of examples of plugins relying on the execution of beforeHasPermission?

Shnoulle:
> But isn't this sounding like bad design? Do we need to execute permission check for permissions that we already checked? Do we know of examples of plugins relying on the execution of beforeHasPermission?

I don't know, but still an API broker. I really don't find a reason to change tis during page process but ?

application/models/Permission.php
Shnoulle: My opinion : currently : aPermissionStatic is badly used : we really need to use it before.

And it improve whol epermission system :)

application/models/Permission.php
Shnoulle: > we get:
> 
> image

Yes, but you pit the echo at the place you want …

> This was the reason behind the change

Yes : i understand this, but since both call use static var, i think there are maybe a better solution to use the static var at a better place in call.

Not in oSurvey, but maybe (surely) in hasPermission directly !

lajosarpad: @Shnoulle We may need both. I have measured the total time spent in

[1] `Permission::hasPermission` when being triggered from `Permission::hasSurveyPermission` either directly or indirectly.
[2] `Permission::hasSurveyPermission`

I did it cumulatively, so I added each measurement of [1] into one value and [2] into another one. If my caching mechanism was not in place, then the result was

image

so 39% of the time spent for survey performance detection happened outside calls for `hasPermission`, ~ 38ms.

With the caching I implemented the former takes next  to nothing and the latter takes ~25ms.

image

So, the question is whether the problem you mentioned exists. If so, then we will need to remove this improvement. If not, then this improvement is beneficial.

However, you are right about fixing the cache inside `hasPermission`, which is a good idea. Yet, besides of that we win further time by having a caching mechanism in oSurvey as well, as long as we do not cause harm by it. I will ask around about this.

application/models/Permission.php
lajosarpad: @Shnoulle I agree with your opinion. Yet, my point is that there are not one, but two issues to tackle with. First, we have the issue I tackled with here, regarding survey permissions, which has some benefit as long as it's not harmful (I will need to find it out, so if there is any harm, I will remove the caching system from oSurvey permission check) and the other topic is your suggestion for fixing the cache at the generic permission check.

I fully agree with you on the second and I will try to find out whether the first causes harm. If not, then it's a good change. If yes, then we will need to remove it. I will nevertheless look into prospects of generic permission checking.

c-schmitz:
@Shnoulle Do you know of any plugin that would rely on this 'bad' behaviour? I have problems to imagine a plugin like this.

Shnoulle:
> @Shnoulle Do you know of any plugin that would rely on this 'bad' behaviour? I have problems to imagine a plugin like this.

No, really no :)

I have one doing this for question update , but i search if we are inside question or not via request.

I don't think it can exist

lajosarpad:
> > But it did not do much of an improvement:
> 
> The plugin event take lot of time here ?
> 
> A reason more to fix [18475: Multiple event newQuestionAttributes happen](https://bugs.limesurvey.org/view.php?id=18475)

That's a very interesting problem. You have mentioned two others in Discord. I will look into these ideas. Thank you very much for all this information.

PR #3760, opened 2024-02-27
Fixed issue #19442: Generated analytics script does not properly escape question group name

mfavetti:
Fixed issue #19442: Generated analytics script does not properly escape question group name

https://bugs.limesurvey.org/view.php?id=19442

5.x


Shnoulle:
Same than master : need to be in theme only.

PR #3767, opened 2024-02-28
Fixed issue #19319: potential error fter Updating To LS 6.4.0

Shnoulle:
Dev: use scopes

### DIFF
--- a/application/core/LSUserIdentity.php
+++ b/application/core/LSUserIdentity.php
@@ -70,17 +70,18 @@ class LSUserIdentity extends CUserIdentity
                 // Never allow login for non-active or expired users.
                 $user = User::model()->notexpired()->active()->findByAttributes(array('users_name' => $this->username));
                 if (is_null($user)) {
-                    throw new CHttpException(403, gT("You do not have permission to access this page."));
-                }
-                // Delegate actual authentication to plugin
-                $authEvent = new PluginEvent('newUserSession', $this); // TODO: rename the plugin function authenticate()
-                $authEvent->set('identity', $this);
-                App()->getPluginManager()->dispatchEvent($authEvent);
-                $pluginResult = $authEvent->get('result');
-                if ($pluginResult instanceof LSAuthResult) {
-                    $result = $pluginResult;
+                    $result->setError(self::ERROR_USERNAME_INVALID);
                 } else {
-                    $result->setError(self::ERROR_UNKNOWN_IDENTITY);
+                    // Delegate actual authentication to plugin
+                    $authEvent = new PluginEvent('newUserSession', $this); // TODO: rename the plugin function authenticate()
+                    $authEvent->set('identity', $this);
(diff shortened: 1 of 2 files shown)


Shnoulle:
getSessionKey broken ?

Shnoulle:
Still need update :)

application/core/LSUserIdentity.php
```diff
@@ -67,20 +66,23 @@ public function authenticate()
...
-                    throw new CHttpException(403, gT("You do not have permission to access this page."));
-                }
-                // Delegate actual authentication to plugin
-                $authEvent = new PluginEvent('newUserSession', $this); // TODO: rename the plugin function authenticate()
-                $authEvent->set('identity', $this);
-                App()->getPluginManager()->dispatchEvent($authEvent);
-                $pluginResult = $authEvent->get('result');
-                if ($pluginResult instanceof LSAuthResult) {
-                    $result = $pluginResult;
+                if (!is_null($user) && (!$user->isActive() || $user->isExpired())) {
```
Shnoulle, line 72: An idea can be move this part after plugin event newUserSession : then plugin can update user before test.

For example : 

1. For an LDAP user : check LDAP status `limesureyaccess`
2. For a Saml account : check an attribute 

It's near a new feature, but this fix both issues : didn't allow non-active or expires users + mantis issue 19319

Shnoulle:
OK : tested with plugin (with auto create) + update expire date

application/core/LSUserIdentity.php
```diff
@@ -85,6 +79,16 @@ public function authenticate()
             }
         }
 
+        /* Check user exist, valid and not expired after plugin actions */
+        if ($result->isValid()) {
+            /** @var \User|null */
+            $user = User::model()->notexpired()->active()->findByAttributes(array('users_name' => $this->username));
```
gabrieljenik, line 85: The intention here is to check if the user is ok to login.
Instead of chainning the rules thrugh scopes, wouldn't it be better to:
- Look for the user
- Create a method canLogin()
- Call that method.

That way the method can easily be enriched with more rules.

application/models/User.php
```diff
@@ -996,6 +996,16 @@ public function isExpired()
         return $expired;
     }
 
+    /**
+     * Returns true if user is active
+     * @return boolean
+     */
+    public function isActive()
```
gabrieljenik, line 1003: Is this related to this PR?
I believe it is a good method, but asking if related to the solution as it is not called.

Shnoulle: I add it during devlopement but remove it after.
https://github.com/LimeSurvey/LimeSurvey/pull/3767/commits/1c0cf421b72bad9e7a66fc77c101ea36b0ae3f2f

You're right :)

application/core/LSUserIdentity.php
Shnoulle: :+1:

gabrieljenik (approved):

PR #3770, opened 2024-02-29
Fixed issue #14661: Plugin development need a "Don't disable plugin"

Shnoulle:
Dev: debug>=2 : disable all load_error action and check


Shnoulle:
Same code than : https://github.com/LimeSurvey/LimeSurvey/pull/3709

gabrieljenik (approved):
Seems the same as the one merged on 5.x
Code structure looks OK. 
Haven't tested it.

PR #3771, opened 2024-02-29
MW-1059: showing loader upon

lajosarpad:
This is all about disabling the save button and showing a loading upon it in the admin panel


Shnoulle:
689 files updated ? ;)

lajosarpad:
> 689 files updated ? ;)

Actually the only change I have implemented was in assets/packages/adminbasics/build/adminbasics.js, where I changed the `displayLoadingState` to

```
	    displayLoadingState = el => {
	      if ((el.id === "save-button") || (el.id === "save-form-button") || ($(el).data('form-id') == 'addnewsurvey')) {
	        const loadingSpinner = '<i class="ri-settings-5-fill remix-spin lsLoadingStateIndicator"></i>';
	        $(el).prop('disabled', true).append(loadingSpinner);
	      }
	    },
```

Yes, there are lots of changes in this PR, but they happened due to the fact that I used `master` as the point of departure, branched off `master` and after implementing the change, created a PR into `develop`. I was told that there are large changes in `develop` and those will be deployed. After that this PR should be much smaller.

Running the command of

```
git diff master feature/MW-1059-disable-save-when-loading
```

we will see the actual changes:

image

whereas, running:

```
git diff master develop
```

we see:

image

and endless further differences. So the reason why we have so many apparently changed files in the PR is the differences between `master` and `develop`.

PR #3772, opened 2024-02-29
Fixed issue #19239: Plugin with PHP error can not be resetted

Shnoulle:


gabrieljenik (approved):
Code structure looks good

PR #3773, opened 2024-03-01
Dev: Port question editor unique subquestion code fix from LS6

kevin-foster-uk:
Port question editor unique subquestion code fix from LS6


kevin-foster-uk:
Sonar cloud is failing because of the 80% code coverage rule. Can we remove this rule. It is not useful.

twilligls (approved):

PR #3775, opened 2024-03-04
Task/lb 132 encoding problems with the body of notification emails

marticos:
Bug fix for "Save Survey Details" email.

This adds the "unescaped" parameter to the translation function to avoid escaping special characters for the body of the mail that is not supposed to be in HTML format.

This affects languages that uses apostrophes like Catalan, Portuguese and so on.

Test instructions:
Please create a survey, for example, in Catalan or use a test survey and change the main language to Catalan.
Start the survey and use the functionality "Continue later" in the survey options in the upper right corner.
Field the required forms and check your email for the translated text.

Actual result:

> Gràcies per desar l&#039;enquesta en curs...

Expected result:

> Gràcies per desar l'enquesta en curs...

Thanks.

### DIFF
--- a/application/libraries/Save.php
+++ b/application/libraries/Save.php
@@ -211,11 +211,11 @@ class Save
                 $mailer->emailType = 'savesurveydetails';
                 $mailer->isHTML(false);
                 $mailer->Subject = gT("Saved Survey Details") . " - " . $thissurvey['name'];
-                $message  = gT("Thank you for saving your survey in progress.  The following details can be used to return to this survey and continue where you left off.  Please make sure to remember your password - we cannot retrieve it for you.");
+                $message  = gT("Thank you for saving your survey in progress.  The following details can be used to return to this survey and continue where you left off.  Please make sure to remember your password - we cannot retrieve it for you.", "unescaped");
                 $message .= "\n\n" . $thissurvey['name'] . "\n\n";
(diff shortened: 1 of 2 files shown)


application/libraries/Save.php
```diff
@@ -211,11 +211,11 @@ function saveSurvey()
                 $mailer->emailType = 'savesurveydetails';
                 $mailer->isHTML(false);
                 $mailer->Subject = gT("Saved Survey Details") . " - " . $thissurvey['name'];
-                $message  = gT("Thank you for saving your survey in progress.  The following details can be used to return to this survey and continue where you left off.  Please make sure to remember your password - we cannot retrieve it for you.");
+                $message  = gT("Thank you for saving your survey in progress.  The following details can be used to return to this survey and continue where you left off.  Please make sure to remember your password - we cannot retrieve it for you.", "unescaped");
                 $message .= "\n\n" . $thissurvey['name'] . "\n\n";
                 $message .= gT("Name") . ": " . Yii::app()->getRequest()->getPost('savename') . "\n";
```
c-schmitz, line 216: Please apply this fix also for the other translated textst in here (Name,Password, etc)-

c-schmitz (changes requested):

c-schmitz (approved):

PR #3777, opened 2024-03-05
Fixed issue #CT-681: Answer options order (answer_order) is missing as an attribute definition and causes the value to not persist through tsv exports and imports

ptelu:
…s an attribute definition and causes the value to not persist through tsv exports and imports

Dev:


Shnoulle:
The new TSV system are really more hard to edit for simple user …

Order was set by order of answer inside the file in the previous TSV : it's really easy to reorder it…

Shnoulle:
Sorry : bad look here … not the same order … really sorry

PR #3780, opened 2024-03-05
Fixed issue [security] #19459: Reflected Cross-Site Scripting (XSS)

Shnoulle:
Dev: encode in view text part
Dev: sanitize other parts


Shnoulle:
OK not toi be encode … then need sanitozer

Shnoulle:
Move it to controller

PR #3782, opened 2024-03-06
Dev: #19452 : getQuestionAttributes for UploaderController

Shnoulle:


application/controllers/UploaderController.php
```diff
@@ -143,11 +143,16 @@ public function run($actionID)
...
             $preview = Yii::app()->session['preview'];
-            $aFieldMap = createFieldMap($oSurvey, 'short', false, false, $sLanguage);
-            if (!isset($aFieldMap[$sFieldName])) {
-                throw new CHttpException(400); // See for debug > 1
+            /* Find the question by sFieldName : must be a upload question type, and id is end of sFieldName in $surveyid*/
+            $aFieldName = explode("X", $sFieldName);
+            if (empty($aFieldName[2]) || !ctype_digit($aFieldName[2])) {
+                throw new CHttpException(400);
+            }
+            $oQuestion = Question::model()->find("sid = :sid and qid = :qid and type = :type", [":sid" => $surveyid, ":qid" => $aFieldName[2], ":type" =>  Question::QT_VERTICAL_FILE_UPLOAD]);
```
lajosarpad, line 151: Since we do this twice (even though not in the same use-case), it would be feasible to have a method for this purpose and call it from both locations.

application/controllers/UploaderController.php
```diff
@@ -347,7 +352,11 @@ public function run($actionID)
 
         $fn = $sFieldName;
         $qid = (int) Yii::app()->request->getParam('qid');
-        $qidattributes = QuestionAttribute::model()->getQuestionAttributes($qid);
+        $oQuestion = Question::model()->find("sid = :sid and qid = :qid and type = :type", [":sid" => $surveyid, ":qid" => $qid, ":type" => Question::QT_VERTICAL_FILE_UPLOAD]);
```
lajosarpad, line 355: Here we have the same question load as we have seen earlier. While this is correct, we could create a method for finding the question and calling getQuestionAttributes for slightly shortening the code and making it easier to maintain.

lajosarpad (approved):
This looks good, I approve it, but I recommend the creation of a method to have the duplicated code separated. Also, we have validations like finding the question and verifying that it's part of the survey we expect it to be part of and having the type we expect and we throw a not found exception if these validations fail.

Maybe we could load the question by qid only and do the validation explicitly and differentiating between the different errors (question does not exist, or question exists, but it's not part of the survey we expected it to be part of, or question exists, it's part of the survey but it has the wrong type).

If you prefer, these changes can be implemented, but they are optional.

application/controllers/UploaderController.php
Shnoulle: > Here we have the same question load as we have seen earlier. While this is correct, we could create a method for finding the question and calling getQuestionAttributes for slightly shortening the code and making it easier to maintain.

We already get the question ? I don't see it.

application/controllers/UploaderController.php
Shnoulle: Yep, but in my opinion UploaderControler need more rewrite ;) if elseif elseif elseif …

private method inside UploaderController or a Quetion method ?

Shnoulle:
> This looks good, I approve it, but I recommend the creation of a method to have the duplicated code separated. Also, we have validations like finding the question and verifying that it's part of the survey we expect it to be part of and having the type we expect and we throw a not found exception if these validations fail.

This must be done for all other [400 error,](https://github.com/LimeSurvey/LimeSurvey/blob/e9211cf2a896a4983365992c9179aad6e6d27b14/application/controllers/UploaderController.php#L57) , clearly goest out of mantis issue


> Maybe we could load the question by qid only and do the validation explicitly and differentiating between the different errors (question does not exist, or question exists, but it's not part of the survey we expected it to be part of, or question exists, it's part of the survey but it has the wrong type).

On a private function why not. But why you want check if wrong type, wrong survey etc ? For login attacker ?

> If you prefer, these changes can be implemented, but they are optional.

In my opinion : we must 

- Move to switch in place of if then else
- Create function for each function

before other improvement.

application/controllers/UploaderController.php
lajosarpad: We agree that it needs more rewrite, but we cannot do a full rewrite as part of this work, it was not approved. If we manage to get approval for a full rewrite, we will need to create a separate ticket for that.

> private method inside UploaderController or a Quetion method ?

I'm good with either way

application/controllers/UploaderController.php
lajosarpad: Sorry for the confusion. I've just meant that we have two places where we load the question:

https://github.com/LimeSurvey/LimeSurvey/pull/3782/files#diff-da3ae6cdc4a22c11c24a0ed743eb4bc1a190312eda624342eff57cc1e27883eeR151

and

https://github.com/LimeSurvey/LimeSurvey/pull/3782/files#diff-da3ae6cdc4a22c11c24a0ed743eb4bc1a190312eda624342eff57cc1e27883eeR355

we do not load it twice in the same run, but we do have both in the code and separating this feature into a method and then calling the method from both places will make the code more readable and easier to maintain. Not a huge trouble really, but it's a small thing we can improve upon.

lajosarpad:
> > This looks good, I approve it, but I recommend the creation of a method to have the duplicated code separated. Also, we have validations like finding the question and verifying that it's part of the survey we expect it to be part of and having the type we expect and we throw a not found exception if these validations fail.
> 
> This must be done for all other [400 error,](https://github.com/LimeSurvey/LimeSurvey/blob/e9211cf2a896a4983365992c9179aad6e6d27b14/application/controllers/UploaderController.php#L57) , clearly goest out of mantis issue
> 
> > Maybe we could load the question by qid only and do the validation explicitly and differentiating between the different errors (question does not exist, or question exists, but it's not part of the survey we expected it to be part of, or question exists, it's part of the survey but it has the wrong type).
> 
> On a private function why not. But why you want check if wrong type, wrong survey etc ? For login attacker ?
> 
> > If you prefer, these changes can be implemented, but they are optional.
> 
> In my opinion : we must
> 
>     * Move to switch in place of if then else
> 
>     * Create function for each function
> 
> 
> before other improvement.

This is only a suggestion and only for the items we are touching. It's optional, we can decide to leave it as it is, but we do have the opportunity to improve it at two places without doing a large refactor that was not approved.

"On a private function why not. But why you want check if wrong type, wrong survey etc ? For login attacker ?"

For any error. Some person implements a buggy plugin that fails due to the question being not part of the survey and then from the error we would instantly see what the problem is. It's a small benefit, true, but not a great effort. But it's optional. The PR is already approved, we are merely discussing further possible improvements.

"
In my opinion : we must

    Move to switch in place of if then else
    Create function for each function

before other improvement.
"

This will prevent us from doing small improvements until a full refactor is approved, which may or may not happen. I'm more on the opinion that we can improve the code we touched, we do not need special privileges to improve upon what we change anyway. But we cannot do a full refactor on this without approval. So, if you decide not to go for these optional improvements, then we might end up never improving upon these places.

Shnoulle:
> For any error. Some person implements a buggy plugin that fails due to the question being not part of the survey and then from the error we would instantly see what the problem is. It's a small benefit, true, but not a great effort. But it's optional. The PR is already approved, we are merely discussing further possible improvements.

OK, log it + show more information on 400 error if are admin user ?

lajosarpad:
> > For any error. Some person implements a buggy plugin that fails due to the question being not part of the survey and then from the error we would instantly see what the problem is. It's a small benefit, true, but not a great effort. But it's optional. The PR is already approved, we are merely discussing further possible improvements.
> 
> OK, log it + show more information on 400 error if are admin user ?

I think if we log it, that's already great. If you can easily make it to be shown for an admin user and you would like to do it, that's even better. But all this is optional, the change is already good.

Shnoulle:
Done

```
     private static function getQuestion($surveyid, $qid)
     {
        $oQuestion = Question::model()->find("sid = :sid and qid = :qid and type = :type", [":sid" => $surveyid, ":qid" => $qid, ":type" =>  Question::QT_VERTICAL_FILE_UPLOAD]);
        if ($oQuestion) {
            return $oQuestion;
        }
        /* Log as warning */
        \Yii::log(sprintf("Invalid upload question %s in survey %s", $qid, $surveyid), 'warning', 'application.UploaderController');
        /* Show information if user have surveycontent/update (can set question type) */
        if (Permission::model()->hasSurveyPermission($surveyid, 'surveycontent', 'update')) {
            throw new CHttpException(400, sprintf(gT("Invalid upload question %s in survey %s"), $qid, $surveyid));
        }
        throw new CHttpException(400);
     }
```

for information seems json return broke currently

lajosarpad (approved):

kevin-foster-uk (approved):

PR #3783, opened 2024-03-06
Dev #19452: Improve usage of getQuestionAttributes : ConditionsAction

Shnoulle:


application/controllers/admin/ConditionsAction.php
```diff
@@ -1484,9 +1484,10 @@ protected function getCAnswersAndCQuestions(array $theserows)
                             'questionl10ns' => array(
                                 'condition' => 'questionl10ns.language = :lang',
                                 'params' => array(':lang' => $this->language)
-                            )))->findAllByAttributes(array('parent_qid' => $rows['qid']), array('order' => 'question_order desc'));
+                            )))->findAllByAttributes(array('parent_qid' => $rows['qid']), array('order' => 'question_order ASC, scale_id ASC'));
                 foreach ($aresult as $arows) {
-                    $attr = QuestionAttribute::model()->getQuestionAttributes($rows['qid']);
+                    /* Used to get dualscale_headerA and dualscale_headerB */
+                    $attr = QuestionAttribute::model()->getQuestionAttributes(Question::model()->findByPk($rows['qid']));
```
lajosarpad, line 1490: Instead of loading a question object individually for all rows that have subquestions we should pass the questions as part of `$theserows` to the method, loaded from the database with a single query, so if we end up having many questions, we send a single query to the database instead of sending individually a query for each row (we do not send multiple individual queries for each row even if they have multiple matches because there is a caching mechanism after each individual query)

lajosarpad (changes requested):
The

```
$attr = QuestionAttribute::model()->getQuestionAttributes(Question::model()->findByPk($rows['qid']));
```

is formally correct, that is, it contributes to the method definition change that is our goal, but if there are multiple rows, then we might end up sending multiple individual queries, as long as their type is '1' and they have at least a match.

application/controllers/admin/ConditionsAction.php
Shnoulle: Argl … ! Thank you !

Shnoulle:
> The
> 
> ```
> $attr = QuestionAttribute::model()->getQuestionAttributes(Question::model()->findByPk($rows['qid']));
> ```
> 
> is formally correct, that is, it contributes to the method definition change that is our goal, but if there are multiple rows, then we might end up sending multiple individual queries, as long as their type is '1' and they have at least a match.

questionAttributes are not related to type only :)

application/controllers/admin/ConditionsAction.php
```diff
@@ -260,8 +260,22 @@ public function index($subaction, $iSurveyID = null, $gid = null, $qid = null)
             }
         }
 
+
         // Previous question parsing ==> building cquestions[] and canswers[]
         if ($questionscount > 0) {
+            $qids = [];
```
Shnoulle, line 266: Clearly goes out of the mantis issue here.

lajosarpad:
> > The
> > ```
> > $attr = QuestionAttribute::model()->getQuestionAttributes(Question::model()->findByPk($rows['qid']));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is formally correct, that is, it contributes to the method definition change that is our goal, but if there are multiple rows, then we might end up sending multiple individual queries, as long as their type is '1' and they have at least a match.
> 
> questionAttributes are not related to type only :)

I did not claim that questionAttributes is about type only, so I'm not sure what you are responding to.

Our scenario is triggered when type is 1. If you reread the code you will see that there is a conditional that checks for type being 1. I'm of course speaking about the part of

```
            } elseif ($rows['type'] == "1") {
                /* Used to get dualscale_headerA and dualscale_headerB */
                $attr = QuestionAttribute::model()->getQuestionAttributes($rows['qObject']);
```

where you can see that `getQuestionAttributes` will be called **if and only if** type is "1".

application/controllers/admin/ConditionsAction.php
lajosarpad: @Shnoulle 

The goals of our work here are:

- changing `getQuestionAttributes` usages so that a question object is passed rather than a number to avoid querying the questions individually
- changing `getQuestionAttributes` to no longer allow the passing of question ids and to expect question object instead
- wherever possible change the usages of `getQuestionAttributes` to load the question objects via a single query

Your approach here was to load the question objects individually and only changing the format of how we are calling `getQuestionAttributes`. But that's missing the point.

We intended to reduce the number of queries in the first place and your PR only improved upon the method format, but not upon the usage of the method.

As a result, I have implemented the improvement myself, that is, I'm collecting the question ids into an array, that's the `$qids` variable for, **load the question with a single query rather than querying them separately** via

```
$rawQuestions = Question::model()->findAllByPk($qids);
```

and adding the resulting question objects to `$theserows` records, like we agreed last week. Consequently, we pass these question objects to `getQuestionAttributes` at the line of

```
                $attr = QuestionAttribute::model()->getQuestionAttributes($rows['qObject']);
```

instead of the `$attr = QuestionAttribute::model()->getQuestionAttributes(Question::model()->findByPk($rows['qid']));` that you had.

Shnoulle: > wherever possible change the usages of `getQuestionAttributes` to load the question objects via a single query

here restricted to type, it's really seems complex code (in my opinion) for this task.

Shnoulle: I add the _wherever possible change the usages of getQuestionAttributes to load the question objects via a single query_ in the mantis issue : it was not clear for me.

Else : here we can avoid the first foreach : something like that : 
```
// In case we need more type
$neededTypes = [QuestionType::QT_1_ARRAY_DUAL];
$criteria = new CdbCriteria();
$criteria->condition = "parent_qid = 0"
$criteria->compare("sid", $sureyid);
$criteria->addInCondition("type", $neededTypes);
$aQuestionOfNeededTypes = Question::model()->findAll($criteria);
```

And maybe after : just set the keys for $aQuestionOfNeededTypes  to qid.

lajosarpad (approved):

kevin-foster-uk (approved):

PR #3787, opened 2024-03-07
Fixed issue #18473: Survey listing : multiple call of SurveyGroup

Shnoulle:
Dev: use static var in SurveysGroupsettings::getInstance

### DIFF
--- a/application/models/SurveysGroupsettings.php
+++ b/application/models/SurveysGroupsettings.php
@@ -74,12 +74,12 @@ class SurveysGroupsettings extends LSActiveRecord
     public $active;
     public $additional_languages;
 
+    /* self[] used in self::getInstance() */
+    private static $aSurveysGroupSettings = [];
 
     /**
      * @return string the associated database table name
      */
-
-
     public function tableName()
     {
         return '{{surveys_groupsettings}}';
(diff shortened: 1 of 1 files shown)


Shnoulle:
Same than https://github.com/LimeSurvey/LimeSurvey/pull/3737

Shnoulle:
after this one, on local : user find came first 
Capture d’écran du 2024-03-07 12-22-23

application/models/SurveysGroupsettings.php
```diff
@@ -292,21 +301,24 @@ public function getIsAnonymized()
...
      */
     public static function getInstance($iSurveyGroupId = 0, $oSurvey = null, $instance = null, $iStep = 1, $bRealValues = false)
     {
 
-        if ($iSurveyGroupId > 0) {
-            $model = SurveysGroupsettings::model()->with('SurveysGroups')->findByPk($iSurveyGroupId);
-        } else {
-            //this is the default group setting with gsid=0 !!!
-            $model = SurveysGroupsettings::model()->findByPk($iSurveyGroupId);
+        if (!array_key_exists($iSurveyGroupId, self::$aSurveysGroupSettings)) {
```
gabrieljenik, line 313: issset() wou;dn't be faster than array_key_exists?

gabrieljenik (approved):
COde structure looks good.

application/models/SurveysGroupsettings.php
Shnoulle: > issset() wou;dn't be faster than array_key_exists?

It can be null here : https://github.com/LimeSurvey/LimeSurvey/blob/c16d1df55a68853c3c9c4f15d9640d6799b050b0/application/models/SurveysGroupsettings.php#L325

Yes isset is better if data is not null :)

gabrieljenik: Right.

PR #3789, opened 2024-03-12
Dev #19452: Improve usage of getQuestionAttributes PrintableSurvey

Shnoulle:
Dev: simple for question
Dev: add specific array for conditions (again array dual scale)
 Dev: avoid multiple call of same question attributes

### DIFF
--- a/application/controllers/admin/PrintableSurvey.php
+++ b/application/controllers/admin/PrintableSurvey.php
@@ -184,7 +184,6 @@ class PrintableSurvey extends SurveyCommonAction
                 foreach ($arQuestions as $arQuestion) {
                     // - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
                     // START doing questions
-
                     $qidattributes = QuestionAttribute::model()->getQuestionAttributes($arQuestion);
                     if ($qidattributes['hidden'] == 1 && $arQuestion['type'] != Question::QT_ASTERISK_EQUATION) {
                         continue;
@@ -248,7 +247,7 @@ class PrintableSurvey extends SurveyCommonAction
                                 // cqid == 0  ==> token attribute match
                                 $tokenData = getTokenFieldsAndNames($surveyid);
                                 preg_match('/^{TOKEN:([^}]*)}$/', (string) $distinctrow['cfieldname'], $extractedTokenAttr);
-                                $sExplanation .= "Your " . $tokenData[strtolower($extractedTokenAttr[1])]['description'] . " ";
+                                $sExplanation .= "Your " . ($tokenData[strtolower($extractedTokenAttr[1])]['description'] ?? "") . " ";
                                 if ($distinctrow['method'] == '==') {
                                     $sExplanation .= gT("is") . " ";
                                 } elseif ($distinctrow['method'] == '!=') {
@@ -266,7 +265,7 @@ class PrintableSurvey extends SurveyCommonAction
                                 } else {
                                     $sExplanation .= gT("is") . " ";
                                 }
-                                $answer_section = ' ' . $distinctrow['value'] . ' ';
+                                $answer_section = ' ' . ($distinctrow['value'] ?? "") . ' ';
                             }
 
                             $conresult = Condition::model()->getConditionsQuestions($distinctrow['cqid'], $arQuestion['qid'], $scenariorow['scenario'], $sLanguageCode);
@@ -451,7 +450,7 @@ class PrintableSurvey extends SurveyCommonAction
                             if ($distinctrow['cqid']) {
                                 $sExplanation .= " <span class='scenario-at-separator'>" . gT("at question") . "</span> '" . " [" . $subresult['title'] . "]' (" . strip_tags((string) $subresult->questionl10ns[$sLanguageCode]->question) . "$answer_section)";
                             } else {
-                                $sExplanation .= " " . $distinctrow['value'];
+                                $sExplanation .= " " . ($distinctrow['value'] ?? "");
                             }
                             //$distinctrow
                             $x++;
@@ -459,7 +458,13 @@ class PrintableSurvey extends SurveyCommonAction
                         $s++;
                     }
 
-                    $qinfo = LimeExpressionManager::GetQuestionStatus($arQuestion['qid']);
+                    $qinfo = LimeExpressionManager::GetQuestionStatus($arQuestion['qid']) ?? [
+                        "info" => [
+                            "relevance" => ""
+                        ],
+                        "relEqn" => "",
+                        "validTip" => ""
+                    ];
                     $relevance = trim((string) $qinfo['info']['relevance']);
                     $sEquation = $qinfo['relEqn'];
 
@@ -640,7 +645,7 @@ class PrintableSurvey extends SurveyCommonAction
                                 }
                             }
 
-                            if ($arQuestion['other'] == 'Y') {
+                            if (($arQuestion['other'] == 'Y') && isset($qidattributes["printable_help"])) {
                                 /*echo '<pre>';
                                 print_r($qidattributes);
                                 echo '</pre>';
@@ -718,7 +723,7 @@ class PrintableSurvey extends SurveyCommonAction
                                     ++$colcounter;
                                 }
                             }
-                            if ($arQuestion['other'] == "Y") {
+                            if (($arQuestion['other'] == "Y") && (isset($qidatteibutes["printable_help"]))) {
                                 if (trim((string) $qidattributes["printable_help"][$sLanguageCode]) == '') {
                                     $qidattributes["printable_help"][$sLanguageCode] = "Other";
                                 }
--- a/application/controllers/admin/PrintableSurvey.php
+++ b/application/controllers/admin/PrintableSurvey.php
@@ -137,6 +137,20 @@ class PrintableSurvey extends SurveyCommonAction
 
             $fieldmap = createFieldMap($oSurvey, 'full', false, false, $sLanguageCode);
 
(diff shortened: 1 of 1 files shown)


lajosarpad (approved):
Looks good. I have added some bug fixes for issues I encountered while testing. Those issues already existed before your changes, but since they broke the test, I fixed them, but for some reason they do not appear in the changes of this PR.

These are the changes I propose:
image
image

application/controllers/admin/PrintableSurvey.php
```diff
@@ -238,7 +247,7 @@ public function index($surveyid, $lang = null, $bReturn = false)
                                 // cqid == 0  ==> token attribute match
                                 $tokenData = getTokenFieldsAndNames($surveyid);
                                 preg_match('/^{TOKEN:([^}]*)}$/', (string) $distinctrow['cfieldname'], $extractedTokenAttr);
-                                $sExplanation .= "Your " . $tokenData[strtolower($extractedTokenAttr[1])]['description'] . " ";
+                                $sExplanation .= "Your " . ($tokenData[strtolower($extractedTokenAttr[1])]['description'] ?? "") . " ";
```
Shnoulle, line 250: If this is null : it's an issue with your Survey data …

application/controllers/admin/PrintableSurvey.php
```diff
@@ -631,7 +646,7 @@ public function index($surveyid, $lang = null, $bReturn = false)
                                 }
                             }
 
-                            if ($arQuestion['other'] == 'Y') {
+                            if (($arQuestion['other'] == 'Y') && isset($qidattributes["printable_help"])) {
```
Shnoulle, line 649: qidattributes["printable_help"] not set ? What question type ? It muts be always set to default : else it's another issue.

Shnoulle:
> Looks good. I have added some bug fixes for issues I encountered while testing. Those issues already existed before your changes, but since they broke the test, I fixed them, but for some reason they do not appear in the changes of this PR.
>
> I do not understand why the changes are not appearing here even though I pushed them

You mean bugfix on master ? Becvause it's appear on tyhe current commit.

I use debug 2 on PHP8.1 with Sample survey for testing: i don't have any issue . Maybe you must check your survey or if there are another issue

I really think we must not fix such issue for example existing QuestionAttribute must always be set to default (printablesurvey) : if not it's an issue in getQuestionAttribute

Shnoulle:
Maybe you need to https://manual.limesurvey.org/Check_data_integrity/en for testing ?

application/controllers/admin/PrintableSurvey.php
```diff
@@ -442,15 +450,22 @@ public function index($surveyid, $lang = null, $bReturn = false)
...
                             }
                             //$distinctrow
                             $x++;
                         }
                         $s++;
                     }
 
-                    $qinfo = LimeExpressionManager::GetQuestionStatus($arQuestion['qid']);
+                    //Defaulting to dummy array to avoid crashes when qinfo is not found for this question
+                    $qinfo = LimeExpressionManager::GetQuestionStatus($arQuestion['qid']) ?? [
```
Shnoulle, line 462: If qinfo retuien null : there are an issue oin ExpressionManager not in this commit !

application/controllers/admin/PrintableSurvey.php
lajosarpad: @Shnoulle you may be correct, but the page should nevertheless not crash even then.

application/controllers/admin/PrintableSurvey.php
lajosarpad: @Shnoulle if it's not set, for whatever reason, then we will have an error. We should not assume it is set unless we are absolutely sure. If we rely on it being set and it is not set, then - issue or not - we have a page crash. Instead, we should fail only for the question if such is the situation.

Shnoulle: > @Shnoulle if it's not set, for whatever reason, then we will have an error. We should not assume it is set unless we are absolutely sure. If we rely on it being set and it is not set, then - issue or not - we have a page crash. Instead, we should fail only for the question if such is the situation.

But you fix another issue in a dev commit. I can not reproduce your issue, maybe there are another issue : we must find and track the original issue !

lajosarpad:
> > Looks good. I have added some bug fixes for issues I encountered while testing. Those issues already existed before your changes, but since they broke the test, I fixed them, but for some reason they do not appear in the changes of this PR.
> > I do not understand why the changes are not appearing here even though I pushed them
> 
> You mean bugfix on master ? Because it's appear on the current commit and not in master
> 
> I use debug 2 on PHP8.1 with Sample survey for testing: i don't have any issue . Maybe you must check your survey or if there are another issue
> 
> I really think we must not fix such issue for example existing QuestionAttribute must always be set to default (printablesurvey) : if not it's an issue in getQuestionAttribute

It did not appear here yesterday, but for some reason it appears now.

I will check for Survey errors, but the code should never crash due to this reason. So if you would like to move these fixes someplace else instead, then I'm fine with that, as long as we have the issue fixed and no new issues are created in the process.

application/controllers/admin/PrintableSurvey.php
Shnoulle: > @Shnoulle We should not assume it is set unless we are absolutely sure.

If it's not set : **it's clearly an issue** : code is done to set ALL attributes default value. Set in the database or not.

It's not an issue in this code, but in getQuestionAttribute : if  getQuestionAttribute don't set default value : it's a BIG issue !

Shnoulle:
> we have the issue fixed and no new issues are created in the process.

Did you try with master : if issues exist in master : it must be fixed in master NOT in another commit …

application/controllers/admin/PrintableSurvey.php
lajosarpad: @Shnoulle I did not claim that your commit introduced the errors. My claim is that I have seen errors testing with a Survey that was exported from an actual customer's instance. We need to make sure that our code is bullet-proof.

lajosarpad:
> Maybe you need to https://manual.limesurvey.org/Check_data_integrity/en for testing ?

Did it. Here's the result:

# git hash: 05a9d4e9f0aebb43a4b6ea8417e1f0c94a839bd0

559767: crashes with old version
581394: crashes with old version
362779: works
764586: crashes with old version
793828: crashes with old version
775547: works

# git hash: e2c4a339e94390fbb9d6df3aa70eb05b65b65295

559767: works
581394: works
362779: works
764586: works
793828: works
775547: works

# Explanation:

Those ids are survey ids in my local database, basically results of the query of

```
select sid, count(*) from lime_conditions join lime_questions on lime_conditions.cqid = lime_questions.qid group by sid order by count(*) desc limit 0, 10;
```

How did I test:

```
        $obj = new PrintableSurvey();
        $obj->index(559767);
        exit;
```

(and I changed the id to be passed as a parameter)

application/controllers/admin/PrintableSurvey.php
Shnoulle: > @Shnoulle I did not claim that your commit introduced the errors. My claim is that I have seen errors testing with a Survey that was exported from an actual customer's instance. We need to make sure that our code is bullet-proof.

Did you try with master : if issues exist in master : it must be fixed in master NOT in another commit …

Shnoulle:
> Did it. Here's the result:

The question is with master, not with your fix.

lajosarpad:
> > Did it. Here's the result:
> 
> The question is with master, not with your fix.

The issue was not created by your work. It already existed. But we cannot accept a PR that crashes the page.

I understand that you want to create a separate ticket for this purpose, but I'm not sure it is a good idea to do so, because then we create prerequisites for this and thusly we postpone the completion of speed improvement, which, as a result will postpone the switch from LS5 to LS6.

These fixes are making the code less volatile and more reliable overall.

I have no issues if you create a new ticket in the future with the purpose of fixing these issues, but you will need to detect the problems during your testing then. In this case, they were detected while I tested with your work with some surveys and they were crashing. the page.

It would be quite the sacrifice in terms of time needed to create a new ticket, wait for approval, create a new PR, have it accepted, merge it into master, merge master back to this one, fix the merge conflicts that will not exist if we fix the bugs when we see a broken branch. All these prerequisites that we create this way postpones the completion of the task

Shnoulle:
> …

It's not about time issue , there are 2 issues on this method (PR with multiple issue)

1. here there are 2 problems which are, in my opinion, specific and which require an understanding of why there is a problem. Otherwise, the problem will reappear somewhere else.
2. We can not follow when something was update and why

If you don't want to create mantis issue : create a PR for each and merge it after @tiborpacalat test it.
I can make a review, but here : i didn't accept 

1. https://github.com/LimeSurvey/LimeSurvey/pull/3789/files#r1522656350 : must fix the way we get token attributes
2. https://github.com/LimeSurvey/LimeSurvey/pull/3789/files#r1522662524 : LimeExpressionManager::GetQuestionStatus muts NOT return an NULL value if the question exist in the survey : https://github.com/LimeSurvey/LimeSurvey/blob/4f7ecb93c29cffe8a62d14ca41bb01f898ee3a0e/application/helpers/expressions/em_manager_helper.php#L6679
3. https://github.com/LimeSurvey/LimeSurvey/pull/3789/files#r1522657115 here : maybe it's related to quetiontype ? But if an attribute exist he must be always set. If printable_help didn't exist for a question : it must be added (even Equation need printable_help)

Shnoulle:
> > …
>     2. We can not follow when something was update and why

This is the worst part : when you try to find what this line was added, if we can improve and have a cleaner code : know the reason of each update are really great !
And GIT history and blame really help for this !

PR #3793, opened 2024-03-19
MW-1040: statistics

lajosarpad:
Refactored the usage of `getQuestionAttributes` for Statistics. There was no opportunity for performance optimisation in this case, but there was an improvement upon the usage of the method.


application/controllers/admin/Statistics.php
```diff
@@ -685,13 +685,24 @@ public function simpleStatistics($surveyid)
         $rows = Question::model()->primary()->getQuestionList($surveyid);
         ;
 
+        $rawQuestions = Question::model()->findAll("sid = :sid and `type` = :type", [
```
twilligls, line 688: as those are only questions of type COLON_ARRAY_NUMBERS, I would include that in the variable names

application/config/version.php
```diff
@@ -11,10 +11,10 @@
  * See COPYRIGHT.php for copyright notices and details.
  */
 
-$config['versionnumber'] = '6.5.0-dev';
+$config['versionnumber'] = '6.5.0';
```
twilligls, line 14: the -dev should not be removed

twilligls (changes requested):

PR #3795, opened 2024-03-20
MW-1040: Applied a fix on LS_Twig_Extension

lajosarpad:
There was no optimization here, I just improved upon the usage


c-schmitz:
There is no benefit in this change, because it is probably outdated/obsolete, so i am closing this.

PR #3799, opened 2024-03-25
Qe 852 formatter format config optional and transient

kevin-foster-uk:


application/libraries/Api/Transformer/Formatter/FormatterYnToBool.php
```diff
@@ -105,15 +105,16 @@ protected function revert($value)
      */
     public function setClassBasedOnConfig($config)
```
kevin-foster-uk, line 106: @twilligls We need to remove this method. Instead of changes the state of the formatter to set the config options the config options should be passed down and evaluated transiently. 

For example:
```
    public function format($value, $config = [], $options = [])
    {
        $revert = isset($config['formatter']['revert'])
            ? $config['formatter']['revert'] : $this->revert;
        return $revert
            ? $this->revert($value, $config)
            : $this->apply($value, $config);
    }
```

Also it would be nice if the transformer only passed in the relevant config so it would look like this:
```
    public function format($value, $config = [], $options = [])
    {
        $revert = isset($config['revert'])
            ? $config['revert'] : $this->revert;
        return $revert
            ? $this->revert($value, $config)
            : $this->apply($value, $config);
    }
```

PR #3800, opened 2024-03-26
QE-831: Fixed API support for language setting change

lajosarpad:
Since this is about language settings, we receive props such as

```
[
    {
        id: 918864,
        op: "update",
        entity: "languageSetting",
        props: {
            en: {
                title: "kdsjdskl"
           }
        }
    }
]
```

whereas other end points do not receive languages, as they are not language-dependant.

As a result, in the `handle` method of `OpHandlerLanguageSettingsUpdate`we can simply loop the props, the keys being the languages and the values being the corresponding values. Previously, `'entityId' => $op->getEntityId()` was passing the survey id as the language :)

### DIFF
--- a/application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
+++ b/application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
@@ -108,23 +108,8 @@ class OpHandlerLanguageSettingsUpdate implements OpHandlerInterface
     public function validateOperation(OpInterface $op): array
     {
         $validationData = [];
-        $checkDataEntityId = $this->getSurveyIdFromContext($op);
-        $checkDataCollection = $this->validateCollection($op, []);
-        if (empty($checkDataEntityId) && empty($checkDataCollection)) {
-            if (!$this->getSurveyIdFromContext($op)) {
-                // operation data has an entity id and props came as collection
-                $validationData = $this->addErrorToValidationData(
-                    'props can not come as collection if id is set',
-                    $validationData
-                );
-            }
-        } elseif (!empty($checkDataEntityId) && !empty($checkDataCollection)) {
-            // operation data has no entity id and props came not as collection
-            $validationData = $checkDataCollection;
-        } elseif (!empty($checkDataEntityId)) {
-            // operation data has no entity id so collection indexes are validated
-            $validationData = $this->validateCollectionIndex($op, $validationData);
-        }
+        
+        $validationData = $this->validateCollectionIndex($op, $validationData);
 
         if (empty($validationData)) {
             $validationData = $this->transformer->validateAll(
--- a/application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
+++ b/application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
@@ -121,11 +121,14 @@ class OpHandlerLanguageSettingsUpdate implements OpHandlerInterface
     public function validateOperation(OpInterface $op): array
     {
         $validationData = [];
-        $checkDataEntityId = $this->validateEntityId($op, []);
+        $checkDataEntityId = $this->getSurveyIdFromContext($op);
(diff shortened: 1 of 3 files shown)


application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
```diff
@@ -96,21 +96,23 @@ public function handle(OpInterface $op): void
...
-            [
-                'operation' => $op->getType()->getId(),
-                'entityId'  => $op->getEntityId(),
-                'sid'       => $this->getSurveyIdFromContext($op)
-            ]
-        );
-        if (empty($data)) {
-            $this->throwNoValuesException($op);
+        foreach ($op->getProps() as $languageCode => $props) {
+            $data = $this->transformer->transformAll(
```
twilligls, line 100: transformAll() is meant to always take collections of entities as a whole and deal with it. Opposed to the function transform() which is mean to handle single entities
Looping over the props, splitting the collection up, would be the job of the transformAll function itself.

twilligls (changes requested):

application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
twilligls: I guess the whole issue is just an issue because the frontend always uses approach 2 (as seen in the docblock of \LimeSurvey\Api\Command\V1\SurveyPatch\OpHandlerLanguageSettingsUpdate::handle)?

twilligls: The id can be anything, it's not broken but also not very clean to pass language there.
The survey id is always passed in the "context" anyway.

If the props come in as 
```
{
            en: {
                title: "kdsjdskl"
           }
        }
```
they are considered a collection, even if the collection only consists of one element.
so transformAll should handle it.

application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
```diff
@@ -121,14 +108,16 @@ public function handle(OpInterface $op): void
     public function validateOperation(OpInterface $op): array
     {
         $validationData = [];
-        $checkDataEntityId = $this->validateEntityId($op, []);
+        $checkDataEntityId = $this->getSurveyIdFromContext($op);
         $checkDataCollection = $this->validateCollection($op, []);
         if (empty($checkDataEntityId) && empty($checkDataCollection)) {
```
twilligls, line 113: This block still needs to be changed. We actually don't care about the entityId anymore. And the data should come in as collection all the time, so 'props can not come as collection if id is set' doesn't make sense anymore.

I would just remove this whole if-block but keep the $validationData = $this->validateCollectionIndex($op, $validationData) 

validateCollectionIndex checks if it's a collection and in this case if the indexes are strings. We don't need anything else

application/libraries/Api/Command/V1/SurveyPatch/OpHandlerLanguageSettingsUpdate.php
```diff
@@ -121,14 +108,16 @@ public function handle(OpInterface $op): void
     public function validateOperation(OpInterface $op): array
     {
         $validationData = [];
-        $checkDataEntityId = $this->validateEntityId($op, []);
+        $checkDataEntityId = $this->getSurveyIdFromContext($op);
```
twilligls, line 111: this line can also be removed. see next comment

twilligls (changes requested):

tests/unit/api/opHandlers/OpHandlerLanguageSettingsTest.php
```diff
@@ -47,20 +47,6 @@ public function testLanguageSettingsUpdateMultipleValidation()
         $this->assertNotEmpty($validation);
     }
 
-    /**
```
twilligls, line 50: :) as you mentioned. It is absolutely correct to remove this test

twilligls (approved):

PR #3801, opened 2024-03-27
Fixed isssue #19500: As a non super admin, I can still access phpinfo page

gabrieljenik:


Shnoulle (approved):
It's OK, but i like to have a 403 error in all condition here.

Reason : any attack or attempted attack must be logged
1. 403 are tracked if you track error with Yii log
2. You can use some auto ban IP with a lot of 403 with server system

Shnoulle (approved):

Shnoulle:
Oh right PSR12 !
Still OK

PR #3803, opened 2024-04-02
Fixed issue #19501: Upgrade broke at 430 with ComfortUpdateChecker

Shnoulle:
Dev: check if already here

### DIFF
--- a/application/helpers/update/updates/Update_430.php
+++ b/application/helpers/update/updates/Update_430.php
@@ -6,6 +6,8 @@ class Update_430 extends DatabaseUpdateBase
 {
     public function up()
     {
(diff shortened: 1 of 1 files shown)


c-schmitz (approved):

Shnoulle:
@c-schmitz the issue are not reproductible like this ;)

There are multiple ComfortUpdateChecker when you setup by default (i add index on the test server)

Shnoulle:
PS : i think if you disable ComfortUpdateChecker before DB version 429 and update : it was added and then activated (or not ? unsure).

tiborpacalat:
Denis, is there something else to be done here?

Shnoulle:
> Denis, is there something else to be done here?

No i need to report the issue differently for master :)

PR #3804, opened 2024-04-02
Fixed issue #19501: Upgrade broke at 430 with ComfortUpdateChecker

Shnoulle:
Dev: check if already here


c-schmitz (approved):

Shnoulle:
fixed in https://github.com/LimeSurvey/LimeSurvey/pull/4339/files

PR #3807, opened 2024-04-08
CT-747: Applied a fix

lajosarpad:
As per https://help.limesurvey.org/agent/limesurvey/support/tickets/details/42705000034180009, applied a fix that avoids saving when the user starts a survey before doing anything.


c-schmitz:
Hm, did you figure out why this happens with this particular survey?

c-schmitz (changes requested):
As agreed, please investigate first the exact reason for this issue with that particular survey.

lajosarpad:
It turned out to be a non-issue.

PR #3819, opened 2024-04-18
dev: remove duplicate survey-find-by-pk calls

TonisOrmisson:
went through the code looking specifically uses of un-necessary duplicate survey findByPk() calls. 
There is a plugin event related to Survey afterFind() so any such call will also run the event and any plugin attached to that event. relates to #3816

### DIFF
--- a/application/controllers/ResponsesController.php
+++ b/application/controllers/ResponsesController.php
@@ -1081,11 +1081,11 @@ class ResponsesController extends LSBaseController
         $aData['imageurl'] = App()->getConfig('imageurl');
         $aData['action'] = App()->request->getParam('action');
         $aData['all'] = App()->request->getParam('all');
-
+        $oSurvey = Survey::model()->findByPk($surveyId);
         //OK. IF WE GOT THIS FAR, THEN THE SURVEY EXISTS AND IT IS ACTIVE, SO LETS GET TO WORK.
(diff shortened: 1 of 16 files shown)


Shnoulle:
> There is a plugin event related to Survey afterFind() so any such call will also run the event and any plugin attached to that event. relates to #3816

This event happens (in 3.X) only at 1st call due to the static usage, it was updated in 5 or 6 ?

Shnoulle:
The issue are not related to multiple call of `Survey::model()->findByPk($surveyid) `

But to Question::model()->survey usage.

Detail

1. 1st call of `Survey::model()->findByPk($surveyid) ` : event happen https://github.com/LimeSurvey/LimeSurvey/blob/e746adcda154109c9cefb5494d4e803f29ba5f5a/application/models/Survey.php#L1003
2. 2n call : https://github.com/LimeSurvey/LimeSurvey/blob/e746adcda154109c9cefb5494d4e803f29ba5f5a/application/models/Survey.php#L1001 no event

Shnoulle (approved):
More clear and clean. But no big improvement on speed, just remove some test : 

```
        if (empty($condition) && empty($params)) {
            if (array_key_exists($pk, self::$findByPkCache)) {
                return self::$findByPkCache[$pk];
```

PR #3823, opened 2024-04-18
Feature: Allow to assign a default role to LDAP users

tograss:
https://bugs.limesurvey.org/view.php?id=17227
New feature #17227: Allows to select a role, wich is assigned to all LDAP users. The default is to do nothing so existing setups still work. 

To keep the plugin decoupled and facilitate plugin development I also somewhat expanded the plugin api to work with roles.
I'm a bit unsure if how I set the timestamps in addRole is correct and weather it should be called role or permissiontemplate.


TonisOrmisson:
you probably meant to target develop, not master branch?

tograss:
Yes sorry. It was abit late yesterday. Can you correct it or should I open a new requests.

Am 19.04.2024 08:12 schrieb Tõnis Ormisson ***@***.***>:

you probably meant to target develop, not master branch?

Shnoulle:
When i need specific system by plugins : i create plugin event in plugin.

I don't know if this feature must be in plugin or in a new plugin.

If it's a needed feature : maybe we need a[ clear API function](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/libraries/PluginManager/LimesurveyApi.php) for other plugin too ?

tograss:
Hi @Shnoulle 
I'm not quite sure if you asking me and if you asking me what exactly the question / task is :-) 
I think it makes sense to have this setting in the current plugin for the following reasons:
1.) When using LDAP you more likely than not have a lot of users so you want to use roles any way. Needing to find a new plugin for this makes it more complicate.
2.) Something similar already exists with the "assign new users the create permission option" . However this can not be used if you want to assign more than just the create permission by default. 
3.) I / someone might extend the plugin further to allow assigning roles based on LDAP groups.

Shnoulle:
> I'm not quite sure if you asking me and if you asking me what exactly the question / task is :-)

No decision here. Just start a discussion about this feature :) . I am not a team leader here. Only an external developer.

>  1.) When using LDAP you more likely than not have a lot of users so you want to use roles any way. Needing to find a new plugin for this makes it more complicate.

I use SAML a lot, and I just use Permission (automatically set).

If I use Roles : I **surely** set roles in LDAP/SAML attribute ! New setting : _get role from :_  I already work on a little plugin who create/set user group according to LDAP group. If we can have from LDAP : i think we must allow it.

>  3.) I / someone might extend the plugin further to allow assigning roles based on LDAP groups.

It's the reason I like more to have : 

1. A new plugin event in LDAP OR globally : NewUserCreated
2. Create a little plugin to set Roles in NewUserCreated event

> If it's a needed feature : maybe we need a[ clear API function](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/libraries/PluginManager/LimesurveyApi.php) for other plugin too ?

Here : if final decision is to get in LDAP plugin : i like to have a function in LimesurveyApi to assign role.

tograss:
> No decision here. Just start a discussion about this feature :) . I am not a team leader here. Only an external developer.
Ahh... I did not know that. You are so active here and in the Bugtracking forum I mistook you for an maintainer :-)
> I use SAML a lot, and I just use Permission (automatically set). 
I you not mind sharing: What SAML plugin are you using? I might be interested as well.
> Here : if final decision is to get in LDAP plugin : i like to have a function in LimesurveyApi to assign role.
This is part of this pull request. I named the function addUserInRole. But naming things is hard :-)

Shnoulle:
> This is part of this pull request. I named the function addUserInRole. But naming things is hard :-)

Oh yes ! Lost in the partial update from master to develop too :)

Great :+1:

application/core/plugins/AuthLDAP/AuthLDAP.php
```diff
@@ -117,6 +117,14 @@ class AuthLDAP extends LimeSurvey\PluginManager\AuthPluginBase
         'allowInitialUser' => array(
             'type' => 'checkbox',
             'label' => 'Allow initial user to login via LDAP',
+        ),
+        // Need to be defined here otherwise saving does not work
+        'defaultRole' => array(
+            'type' => 'select',
+            'options' => array('none'=>'None'),
```
Shnoulle, line 124: Empty option with

```
'htmlOptions' => [
    'empty' => 'None'
]
```

default is empty.

Think it's better

application/core/plugins/AuthLDAP/AuthLDAP.php
```diff
@@ -390,13 +398,23 @@ public function newLoginForm()
...
      *
      * @param boolean $getValues
      * @return array
      */
     public function getPluginSettings($getValues = true)
     {
         $aPluginSettings = parent::getPluginSettings($getValues);
+        // Dynamically append available roles as options
+        $roles = $this->api->getRoles();
+        $roleOptions = array("none"=>'None');
```
Shnoulle, line 411: See previous comment about [htmlOptions](https://www.yiiframework.com/doc/api/1.1/CHtml#dropDownList-detail)

application/core/plugins/AuthLDAP/AuthLDAP.php
```diff
@@ -589,6 +608,20 @@ public function newUserSession()
             }
         }
         // If we made it here, authentication was a success and we do have a valid user
+        
+        // If requested assign role to user
+        if ($defaultrole !== 'none'){
```
Shnoulle, line 613: if (!empty($defaultrole)) {

application/core/plugins/AuthLDAP/AuthLDAP.php
tograss: I agree and updatet the pull request

application/core/plugins/AuthLDAP/AuthLDAP.php
tograss: Dito

application/core/plugins/AuthLDAP/AuthLDAP.php
tograss: Done

tograss:
Ok I open a new pull request based on develop. Sorry for the trouble.

Shnoulle:
> Ok I open a new pull request based on develop. Sorry for the trouble.

No need to be sorry here :)

PS : before , please : 

1. Create a Feature request on https://bugs.limesurvey.org/main_page.php (see https://community.limesurvey.org/feature-request/)
2. Use `New feature #XXXX : ` for the 1st commit an dthe title

Others advise on https://manual.limesurvey.org/How_to_contribute_new_features

PR #3825, opened 2024-04-22
Fixed issue #19178: The name of the survey disappear from ALL page

Shnoulle:
Dev: start process page by page …

### DIFF
--- a/application/controllers/ThemeOptionsController.php
+++ b/application/controllers/ThemeOptionsController.php
@@ -730,7 +730,16 @@ class ThemeOptionsController extends LSBaseController
             null,
             true
         );
-
+        /* @var \Survey|null used by view for UX */
+        $oSurvey = null;
+        /* @var \SurveysGroups|null used by view for UX */
+        $oSurveysgroup = null;
+        if ($sid) {
+            $oSurvey = Survey::model()->findByPk($sid);
+        }
+        if ($gsid) {
+            $oSurveysgroup = SurveysGroups::model()->findByPk($gsid);
+        }
         $oSimpleInheritance->options = 'inherit';
         $oSimpleInheritanceTemplate = $oSimpleInheritance->prepareTemplateRendering(
             $oModelWithInheritReplacement->sTemplateName
@@ -748,7 +757,9 @@ class ThemeOptionsController extends LSBaseController
             'oParentOptions'  => $oParentOptions,
             'sPackagesToLoad' => $oModelWithInheritReplacement->packages_to_load,
             'sid' => $sid,
-            'gsid' => $gsid
+            'oSurvey' => $oSurvey,
+            'gsid' => $gsid,
+            'oSurveysgroup' => $oSurveysgroup,
         );
 
         if ($sid !== null) {
--- a/application/extensions/LimeTopbarWidget/views/topbar.php
+++ b/application/extensions/LimeTopbarWidget/views/topbar.php
@@ -10,14 +10,14 @@
         <div class="row">
             <!-- Title or breadcrumb -->
             <div class="ls-breadcrumb col-xl-4 col-xxl-3">
-                <h1 class="align-items-center d-flex">
+                <div class="align-items-center d-flex">
                 <?php
                     if ($titleBackLink !== null) {
                         echo '<a class="h1" href="'.$titleBackLink.'">
                         <i class="ri-arrow-left-s-line"></i></a>' ;
                     }
                 ?>
-                <?= $leftSide ?></h1>
+                <?= $leftSide ?></div>
             </div>
 
             <!-- middle part with buttons -->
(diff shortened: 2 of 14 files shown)


Shnoulle:
Need advice / opinion

1. Use always small for Survey name
2. Use always d-block for Survey name
3. Use h2 class 

Basic
image

small on survey
image

small + dblock
image


h2 + small + d-block
image

Shnoulle:
Fixing heading map in same PR
image

Shnoulle:
Ready for code review 

1. Add title at near all Survey page (some use a specific view)
2. Review some heading map
3. Add survey name and id at all page

c-schmitz:
Oh wow. 
Wouldn't it have much more sense to update the breadcrumb to show the name instead of 3 dots? That was what I was always missing.
I don't need the full name of the survey on every page.
But I would also need it in the page title when navigating with several tabs.

c-schmitz:
Breadcrumb gets too  long? Ellipsize it in the middle of each segment , or make space that i can span across the buttons on top.

Shnoulle:
It's really better to have it on the title of the page.

- On my report : I compare with 5 : we have the survey name on Survey settings
- ricles report : near same https://bugs.limesurvey.org/view.php?id=19421
- [Joan99](https://bugs.limesurvey.org/view_user_page.php?id=394211) ask to [have survey name in all page](https://bugs.limesurvey.org/view.php?id=19516) 

Know the current survey is a **major** information inside admin GUI. Not a navigation information.

Breadcrumb are [breadcrumb](https://developer.mozilla.org/en-US/docs/Glossary/Breadcrumb) : help on navigation only.
Not “What is the usage of this page”

PS : we can easily [put only SID in the new title : it's a widget](https://github.com/LimeSurvey/LimeSurvey/pull/3825/files#diff-e973eb542dbf662352c1bdb3e768bd6d2c0f2fb371a43a71f77d0aafef73e610).

And more : it's restricted now 
image

I think

Shnoulle:
> Breadcrumb gets too long? Ellipsize it in the middle of each segment , or make space that it can span across the buttons on top.

My opinion : it's another issue to report : have real breadcrumb (and i think the current one are badly placed)

c-schmitz:
The breadcrumb is not for navgiation only, but also information on where exactly you are right now in the application.
It should reflect the survey name -> groupname -> questionname at all times up to a reasonable length.
It is the THE place where this information needs to be. 
I feel it is cluttering the interface by putting it in various locations inside the main content.

I am really sorry. but I wish this would have been discussed before you put in all the coding time.

Shnoulle:
[New issue : 19552: Breadcrumb are badly placed cut and don't show survey name, group name etc …](https://bugs.limesurvey.org/view.php?id=19552)

Else: https://bugs.limesurvey.org/view.php?id=18970#c76212 

> There is a tooltip if you hover over the 3 dots (...) in the breadcrumbs.

> Btw. this is not a bug, if you would like something different, it should be feature request.

ping @tiborpacalat

tiborpacalat:
Denis, what action is required here from me?

Shnoulle:
> Denis, what action is required here from me?

More a decision about [19552: Breadcrumb are badly placed cut and don't show survey name, group name etc …](https://bugs.limesurvey.org/view.php?id=19552) mantis issue.

And assignment

Shnoulle:
> Denis, what action is required here from me?

And maybe test survey management UX with or without this title

Shnoulle:
> Breadcrumb gets too long? Ellipsize it in the middle of each segment , or make space that it can span across the buttons on top.

Can we have both ? New title (easy to remove the Survey name, since it was an extension + real and complete breadcrumb

Shnoulle:
With this commit we have

1. The name of the survey (we can remove it when breadcrumb is not needed anymore)
2. The current action : needed for UX
3. The current group or question : needed for UX

Can someone test it and say «It's really not as good as before»

tiborpacalat:
Why are there still 87 files with changes?

Screenshot 2024-06-17 at 16 01 30

I don't think we need "Q9" in 'Question “Q9” summary (ID 16)'. There are 3 more places where you can see the question code. Same goes for group, I would leave "2" out of 'Group “2” summary' for example.

There are many conflicting files as well ...

Shnoulle:
> Why are there still 87 files with changes?

Because: I need to add title to all page view…
It was removed between 5 and 6

> I don't think we need "Q9" in 'Question “Q9” summary (ID 16)'. There are 3 more places where you can see the question code. Same goes for group, I would leave "2" out of 'Group “2” summary' for example.

For question : My opinion, I think it's a really needed information, in 3.X we have 3 times too. Have a place, same place in each page to know where you are with a quick look. Maybe some other advice by user @maziminke ?

For group : there are no real group code, i use the group step number here. I think it's really interesting to know with a quick look.

I ask some reporter opinion.

> There are many conflicting files as well ...

Since April … yes sure … and since I'm not sure whether it was merged or not. I clearly wait for fixing

Shnoulle:
I just report an issue about loosing information on current page : survey (with ID) + action + Group + Question …

Do not delte the PR

PR #3826, opened 2024-04-22
Fixed issue #19528: Bad usability when open in a new tab for participant edit

Shnoulle:
Dev: use Yii::app()->request->isAjaxRequest


Shnoulle:
Need advice / opinion

1. Use always small for Survey name
2. Use always d-block for Survey name
3. Use h2 class 

Basic
image

small on survey
image

small + dblock
image


h2 + small + d-block
image

application/controllers/admin/Tokens.php
```diff
@@ -737,17 +737,16 @@ public function addnew(int $iSurveyId)
      * Edit Tokens
      * @param int $iSurveyId
      * @param int $iTokenId
-     * @param boolean $ajax
      * @return false|null
-     * @todo When is this function used without Ajax?
      */
-    public function edit($iSurveyId, $iTokenId, $ajax = false)
+    public function edit($iSurveyId, $iTokenId)
```
gabrieljenik, line 742: Shouldn't we look for the places where this is called to remove a third parameter if appears?

Also, not sure about changing the signature of a public function.
We tried not to do so in the past. But I believe that could be evaluated case by case. As this is a controller, perfectly fine.

Shnoulle: Here the only usage is to return the partial view or the complete view.

You return the partial view if it's an ajax request, complete if not. Get is ajax request by parameter is an issue.

Maybe deprecate the unused part ?

Shnoulle: (i can search and remove the place where the ajax is added in url)

Shnoulle: > (i can search and remove the place where the ajax is added in url)

Done

Shnoulle: > Also, not sure about changing the signature of a public function. We tried not to do so in the past. But I believe that could be evaluated case by case. As this is a controller, perfectly fine.

Done too, but really deprecate ;) and put version

application/controllers/admin/Tokens.php
```diff
@@ -839,7 +838,14 @@ public function edit($iSurveyId, $iTokenId, $ajax = false)
                 foreach ($aAdditionalAttributeFields as $attr_name => $desc) {
                     $value = $request->getPost($attr_name, '');
                     if ($desc['mandatory'] == 'Y' && trim($value) == '') {
-                        $sOutput .= sprintf(gT("Notice: Field '%s' was left empty, even though it is a mandatory attribute."), $desc['description']) . '<br>';
+                        /* All this part is disable via JS, no way to submit : issue #19548 https://bugs.limesurvey.org/view.php?id=19548*/
```
gabrieljenik, line 841: What do you mean by disable?

application/controllers/admin/Tokens.php
```diff
@@ -853,22 +859,42 @@ public function edit($iSurveyId, $iTokenId, $ajax = false)
...
+                    App()->setFlashMessage(gT('The survey participant was successfully updated.'), 'success');
+                    if ($redirect) {
+                        $redirectUrl = Yii::app()->createUrl(
+                            "admin/tokens",
+                            [
+                                "sa" => 'index',
+                                "surveyid" => $iSurveyId,
+                            ]
+                        );
+                        $this->getController()->redirect($redirectUrl);
```
gabrieljenik, line 878: Does this make sense if we are running on ajax mode?

application/controllers/admin/Tokens.php
```diff
@@ -853,22 +859,42 @@ public function edit($iSurveyId, $iTokenId, $ajax = false)
...
-                    $errors = $token->getErrors();
-                    $firstError = reset($errors);
-                    \ls\ajax\AjaxHelper::outputError($firstError[0]);
+                    if ($ajax) {
+                        $errors = $token->getErrors();
+                        $firstError = reset($errors);
+                        \ls\ajax\AjaxHelper::outputError($firstError[0]);
+                        // App->end in AjaxHelper
+                    }
+                    App()->setFlashMessage(CHtml::errorSummary($token, '<div><strong>' . gT('The survey participant was not updated.') . '</strong></div>'), 'danger');
```
gabrieljenik, line 887: This would run if on ajax mode?

application/controllers/admin/Tokens.php
Shnoulle: > What do you mean by disable?

I put the number of mantis issue : https://bugs.limesurvey.org/view.php?id=19548

Here it's the PHP part : allow to save empty value in required attribute : but a JS function disable this.

application/controllers/admin/Tokens.php
Shnoulle: We never get here , see https://github.com/LimeSurvey/LimeSurvey/pull/3826/files#diff-af8e690b7162e715acc4d95c2dd74fb97479066cf40a03d33de4c393df0250e3R867

application/controllers/admin/Tokens.php
Shnoulle: Just the test and the line before please, with the comment https://github.com/LimeSurvey/LimeSurvey/pull/3826/files#diff-af8e690b7162e715acc4d95c2dd74fb97479066cf40a03d33de4c393df0250e3R885

PR #3827, opened 2024-04-22
Fixed issue #19526: debug=2 PHP8.1 : unable to send email without expiry

Shnoulle:
Dev: cast as string

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -22,8 +22,8 @@ use LimeSurvey\PluginManager\PluginEvent;
  * @property integer $gsid survey group id, from which this survey belongs to and inherits values from when set to 'I'
  * @property string $admin Survey Admin's full name
  * @property string $active Whether survey is acive or not (Y/N)
- * @property string $expires Expiry date (YYYY-MM-DD hh:mm:ss)
- * @property string $startdate Survey Start date (YYYY-MM-DD hh:mm:ss)
+ * @property string|null $expires Expiry date as SQL datetime (YYYY-MM-DD hh:mm:ss)
+ * @property string|null $startdate Survey Start date as SQL datetime (YYYY-MM-DD hh:mm:ss)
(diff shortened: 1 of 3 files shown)


gabrieljenik:
cast as string or strval ? 
I think we usually use cast as string.
Probably they have the same results on most cases.
Wonder why you chose strval()

gabrieljenik (approved):
Code structure looks good

gabrieljenik:
> cast as string or strval ? I think we usually use cast as string. Probably they have the same results on most cases. Wonder why you chose strval()

From AI:

In PHP, both casting a value to a string using (string) and using the strval() function serve the purpose of converting a value to a string. However, there are subtle differences between the two approaches:

Functionality: Both methods allow you to convert scalar values (like integers, floats, booleans, and objects that implement the __toString method) to strings. However, strval() cannot be used on arrays or objects that do not implement the __toString method, whereas (string) casting can be used in a broader range of contexts, including with arrays and objects that do not implement __toString.

Usage Context: strval() is a function call, making it explicit in the code that a conversion is taking place. On the other hand, (string) casting is an internal type casting method, which might be considered more concise and integrated within the language syntax.

Performance: In most practical scenarios, the performance difference between using strval() and (string) casting is negligible. However, in very specific performance-critical applications, there might be slight differences, though these are unlikely to be significant enough to impact the overall performance of your application.

Readability and Preference: Some developers prefer using (string) casting as it aligns with modern PHP practices and is seen as more readable and concise. Others might prefer strval() for its explicitness, especially in complex expressions where the type of conversion might not be immediately clear.

In summary, both strval() and (string) casting can be used interchangeably for converting values to strings in PHP. The choice between them often comes down to personal preference, readability considerations, and the specific requirements of your codebase.

Shnoulle:
I always use strval, boolval etc … function, i really dislike cast system. I prefer function

PR #3831, opened 2024-04-24
QE-903: create service for activate and deactivate survey

lajosarpad:
As per https://limesurvey.atlassian.net/browse/QE-903, this is the server-side of that task


application/models/services/SurveyActivate.php
```diff
@@ -0,0 +1,71 @@
...
+                $survey->{$field} = App()->request->getPost($field, $params[$field] ?? null);
+            }
+            $survey->save();
+
+            // Make sure the saved values will be picked up
+            Survey::model()->resetCache();
+            $survey->setOptions();
+        }
+
+        $surveyActivator = new SurveyActivator($survey);
```
kevin-foster-uk, line 67: `SurveyActivator` should be a constructor dependency.

application/models/services/SurveyActivate.php
```diff
@@ -0,0 +1,71 @@
...
+            $fields = [
+                'anonymized',
+                'datestamp',
+                'ipaddr',
+                'ipanonymize',
+                'refurl',
+                'savetimings'
+            ];
+            foreach ($fields as $field) {
+                $survey->{$field} = App()->request->getPost($field, $params[$field] ?? null);
```
kevin-foster-uk, line 58: App() can be injected using LSYii_Application

application/models/services/SurveyActivate.php
```diff
@@ -0,0 +1,71 @@
...
+            }
+            $survey->save();
+
+            // Make sure the saved values will be picked up
+            Survey::model()->resetCache();
```
kevin-foster-uk, line 63: Ideally we should inject `Survey` also but I am not sure if calling `resetCache()` on the injected instance will have the desired effect. In that case its fine to leave this as is.

application/models/services/SurveyActivate.php
lajosarpad: @kevin-foster-uk applied the change accordingly

application/models/SurveyActivator.php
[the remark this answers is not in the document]
lajosarpad: @kevin-foster-uk it was not my choice, see https://github.com/LimeSurvey/LimeSurvey/blame/a99ecb718e3fe053eb8293940394183853365d76/application/models/SurveyActivator.php#L7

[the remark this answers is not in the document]
kevin-foster-uk: Ah I see those models already existed. That explains it.

kevin-foster-uk (approved):

application/models/SurveyDeactivator.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /** @var array */
+    protected $timingsTableDefinition = [];
+
+    /** @var string */
+    protected $error;
+
+    public function __construct($survey = null)
+    {
+        $this->$survey = $survey;
```
TonisOrmisson, line 24: a bug here, should be $this->s.., not $this->$s... Error 500 "Creation of dynamic property SurveyDeactivator::$ is deprecated"

lajosarpad: Thanks, @TonisOrmisson for pointing this out, created a PR to fix it

PR #3833, opened 2024-04-25
Fixed issue #19527: Potential 500 error in survey list

Shnoulle:
Dev: getTemplateListWithPreviews use getTemplateList listing
Dev: deprecate unused getTemplateListWithPreviews


Shnoulle:
Near same than https://github.com/LimeSurvey/LimeSurvey/pull/3829

```
/application$ grep -ri 'getTemplateListWithPreviews' *
models/Template.php:    public static function getTemplateListWithPreviews()
```

PR #3834, opened 2024-04-26
Fixed issue #19526: debug=2 PHP8.1 : unable to send email without expiry

Shnoulle:
Dev: cast as string for replace


Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/3827 ????

PR #3835, opened 2024-04-30
QE-591: show button on more topbars, js for triggering the link for b…

Trischi80:
Dev: show editorbutton on more topbars, js for triggering the link for button


lajosarpad (approved):
Code-wise it looks good, I didn't test it though

PR #3837, opened 2024-05-01
Fixed issue #19543: KCFinder needs to be updated when using with PHP 8.1

gabrieljenik:


Shnoulle (approved):
Code is OK, but it's not a vendor anymore 

@c-schmitz maybe create a repository for kcfinder here ?

PR #3839, opened 2024-05-03
 Fixed issue #19550: Calling get_class() without arguments is deprecated PHP8.3

TonisOrmisson:
LimeSurvey manual says v6 php requirement is LS 6.x from PHP 7.4.x to 8.x

LS master should not use mehods deprecated by the latest php 8 (8.3) also the tests should run the 8.3 instead of 8.2.

This fix replaces all get_class() without arguments  that will throw an error in php 8.3 with get_class() WITH relevant argument $this or `new static` in static context to avoid errors in php 8.3

also I changed the php 8.2=>8.3 in ci tests matrix that without this fix would be red, but are green with this fix.


TonisOrmisson:
note: there is still relevant get_class() without arguments in vendor/pear/pear-core-minimal

gabrieljenik (approved):
Code looks OK.
Still a few remarks: 
- Should the PR go to DEV or master?
- @olleharstedt I added you as reviewer as there are changes to the automatic tests.

gabrieljenik:
I wouldn't merge it until getting another revision

application/core/plugins/TwoFactorAdminLogin/vendor/robthree/twofactorauth/demo/loader.php
```diff
@@ -14,7 +14,7 @@ protected static function initialize()
         if (static::$initialized) return;
         static::$initialized = true;
         static::$parentPath = __FILE__;
-        for ($i=substr_count(get_class(), (string) static::$nsChar);$i>=0;$i--) {
+        for ($i=substr_count(get_class(new static), (string) static::$nsChar);$i>=0;$i--) {
```
Shnoulle, line 17: Arg !
https://github.com/RobThree/TwoFactorAuth/commit/225216a61606297a0c9ca3d6820849c26dd5d729

:)

Shnoulle (approved):

Shnoulle:
> I wouldn't merge it until getting another revision

Send it in testing ?

We **need** PHP8.3 in some month. Some forum post already send with PHP8.3 issues

olleharstedt (approved):

olleharstedt:
Nice :)

.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         # Specify what versions of php you want to test
-        php-versions: ['7.4', '8.2']
+        php-versions: ['7.4', '8.3']
```
Shnoulle, line 20: Maybe need 7.4 + 8.1 + 8.3 ?

TonisOrmisson: It seemed to me that the idea here is to have the minimum & maximum version runs, so I just replaced the 8.2 with 8.3 as  minimal change here.

Shnoulle: I don't know … unsure.

But never mind : if it works on both : it surely works on middle.

tiborpacalat:
This has already been implemented. Sorry about it, but we did a global sweep for upgrade to php8.3. We appreciate the effort put into this.

PR #3840, opened 2024-05-07
Fixed issue #19555: afterFindSurvey gets triggered excessively

TonisOrmisson:
There is some additional discussion regarding this fix in #3816

The [afterFindSurvey event](https://manual.limesurvey.org/AfterFindSurvey) gets triggered each time survey is found found via `Survey::findByPk()`, but as well always when survey is found via relations eg `$question->survey`. 

With a survey page of multiple questions and even sub-questions the event `afterFindSurvey` might get triggered tens of times on one survey page load. 

There is a discussion whether this is a  bug or a feature. 

1) It's a feature view
`afterFindSurvey` event is tightly related to instantianting any survey object, also meaning it must be triggered after initializing a survey via relations or any time a survey object is populated. In this case it is not a bug and we should not fix it, this PR should be closed.

2) It's a bug view
The event manual states that the event is for:
> to be able to modify/override some display parameters for current display session

This means the whole display session (survey page) - meaning it does not need to be triggered tens of times during a a page load. 
It is not related to initializing _any_ Survey object it is meant specifically to trigger after `Survey::findByPk()`. In this case It's a bug and this PR will move the triggering essentially from populating any survey record to specifically only if `Survey::findByPk()` returns a new object (not stored also via `Survey::$findByPkCache`)

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -1008,7 +1008,9 @@ class Survey extends LSActiveRecord implements PermissionInterface
             }
         }
         $model = parent::findByPk($pk, $condition, $params);
-        $model ?? $model->afterFindSurvey();
+        if (!is_null($model)) {
+            $model->afterFindSurvey();
+        }
         return $model;
     }
 
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -1008,7 +1008,7 @@ class Survey extends LSActiveRecord implements PermissionInterface
             }
         }
         $model = parent::findByPk($pk, $condition, $params);
-            $model ?? $model->afterFindSurvey();
+        $model ?? $model->afterFindSurvey();
         return $model;
     }
 
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -185,7 +185,6 @@ class Survey extends LSActiveRecord implements PermissionInterface
         if ($this->isNewRecord) {
             $this->setAttributeDefaults();
         }
-        $this->attachEventHandler("onAfterFind", array($this, 'afterFindSurvey'));
     }
 
     private function setAttributeDefaults()
@@ -1003,11 +1002,13 @@ class Survey extends LSActiveRecord implements PermissionInterface
                 $model = parent::findByPk($pk, $condition, $params);
                 if (!is_null($model)) {
                     self::$findByPkCache[$pk] = $model;
+                    $model->afterFindSurvey();
                 }
                 return $model;
             }
         }
         $model = parent::findByPk($pk, $condition, $params);
+            $model ?? $model->afterFindSurvey();
         return $model;
     }
 


Shnoulle:
The function afterSurveyFind do some other things

For example https://github.com/LimeSurvey/LimeSurvey/blob/c2b4fd9f03a041278a8d56e644992c22688b0fcb/application/models/Survey.php#L588

I think it's better too to have it only one time but 

- When we get it via $question->survey : options are set
    - currently
    - after this commit 

Need some unit test too to check if 

1. plugin event can update something
2. template is fixed
3. options are set

Shnoulle (approved):

gabrieljenik:
The afterFind LS plugin event is tightly related to the onAfterFind Yii event.
https://www.yiiframework.com/doc/api/2.0/yii-db-activerecord#afterFind()-detail

What is being suggested here looks more to me like:
- Using the same afteFind event name.
- For a closely related, but different, purpose. If I understand correctly the afterFind Yii event will be triggered for many find situations. The new will only be triggered on the afterFindByPk.
- This may break compatibility backwards facing.

I would just:
- Create a new PluginEvent afterInit.
- Eventually, with time, review if the afterFind plugin event should be deprecated

Also, of course, clarify the situation with the documentation.

TonisOrmisson:
https://github.com/LimeSurvey/LimeSurvey/pull/3816#issuecomment-2120364100
I will comment the documentation comment here

It seems to me the issue should be solved starting from the stated purpose of the event. **Why** this event exists, what is the purpose of the event? The description of the event in the documentation could be old version/date-wise but there might not have anything significant changed code-wise for that specific event after documenting it. 

I'm not sure which way is better to go, it depends on how LS states the purpose of the event. Currently I see the fact is that the stated purpose and actual behavior do not match. People using the event for the stated functionality end up calling the event excessively possibly without realizing it (I know I was). 

'afterInit' - you mean afterInit Survey model? This seems very similar to the current afterFindSurvey essentially. I personally do not see use for that and have no comment on possible afterInitSurvey.

gabrieljenik:
> People using the event for the stated functionality end up calling the event excessively possibly without realizing it (I know I was).

Agree. So I would create a new event so people can start moving to use that event that better fits their situation.

> 'afterInit' - you mean afterInit Survey model? This seems very similar to the current afterFindSurvey essentially. I personally do not see use for that and have no comment on possible afterInitSurvey.

Similar, but would only be triggered once.
Not sure about the name. Probably the name will depend on when is technically being triggered.

TonisOrmisson:
> Similar, but would only be triggered once.

How once? Init gets triggered even more. I attached a log to Survey init() and afterFindSurvey and loaded one survey page. The result looks like that :

image
That for sure will not solve the excessive triggering issue

Shnoulle:
The issue is this one : With 3.X and before : LimeSurvey don't use $question->survey (and no pjax/ajax call (not a lot))

Then when using survey->findByPk : we call it one time when we work on a survey.

**Seems** the new versions : 5 and master have

- Call with $question->survey
- Called by ajax for one page

My opinion : 

1. afterSurveyFind event can update any attribute of Survey object
2. afterSurveyFind event musts be called one time for a page call : return same Survey object (static)
3. the survey object call by question->survey must return the same Survey object

Maybe we need to update the way Yii return $question->survey ? If we have a  $question->getSurvey : it happen when we use $question->survey ?

gabrieljenik:
> How once? Init gets triggered even more. 

I mean the new event should be triggered much less than the current afterFind. Expectedly, once.

> Not sure about the name. Probably the name will depend on when is technically being triggered.

I if ti shall be used for survey taking process, we could name it beforeSurveyLoadSurvey
Then, yes, I expect the new event ot be called once

Shnoulle:
> I wouldn't update how this works. I would create a new event for it. 
> Then developers should move their plugins as they seem appropriate

It work like that when it was created and updated : you have to reset the static variable in beforeSureyPage to allow update (i have a plugin with this).

The 5.X system work not like 3.X system then.

Shnoulle:
I setup a vanilla LimeSurvey master to check (without plugin) but on last 5.X when survey start with the sample survey included in doc :
I have 114 load or `SELECT survey.` 

If confirmed without plugin , this need to be fixed in master 

Capture d’écran du 2024-06-06 10-21-02

Shnoulle:
> I setup a vanilla LimeSurvey master to check (without plugin) but on last 5.X when survey start with the sample survey included in doc : I have 114 load or `SELECT survey.`
> 
> If confirmed without plugin , this need to be fixed in master
> 
> Capture d’écran du 2024-06-06 10-21-02
> 
> _Related issue, but not totally with the event_

https://bugs.limesurvey.org/view.php?id=19589

lajosarpad:
The suggestion if I understood it well is to change a Limesurvey event from occurring on every survey finds to occur the first time a survey is found so one could implement a template change there which would then happen once.

This may be an idea that appears to be working well according to all our tests, but unfortunately the event was working like it is working for many years and it is quite possible that some people have built some plugins that relied on the way the plugin was working. If we do such drastic changes, then such plugins may be broken as a result.

So the safer approach would be to either implement an optional config for this event to be singularly or plurally executed and, if this config is set and it is set to singular, then it would behave like in this suggestion, but if it is not set, or is set to plural, then the event would be triggered like it was up to now.

Alternatively, we could apply Gabriel's approach and create another event instead of doing drastical changes to this one.

Note that `findByPk` already caches the survey, as it has an override for the Survey class in order to cache:

```
    /**
     * @inheritdoc . But use a static var because can be used a lot of time.
     */
    public function findByPk($pk, $condition = '', $params = array())
    {
        /** @var self $model */
        if (empty($condition) && empty($params)) {
            if (array_key_exists($pk, self::$findByPkCache)) {
                return self::$findByPkCache[$pk];
            } else {
                $model = parent::findByPk($pk, $condition, $params);
                if (!is_null($model)) {
                    self::$findByPkCache[$pk] = $model;
                }
                return $model;
            }
        }
        $model = parent::findByPk($pk, $condition, $params);
        return $model;
    }

```

Shnoulle:
> This may be an idea that appears to be working well according to all our tests, but unfortunately the event was working like it is working for many years and it is quite possible that some people have built some plugins that relied on the way the plugin was working. If we do such drastic changes, then such plugins may be broken as a result.
> 


The event happen ONE time  (maybe 2 or 3) for many years. Please compare with 3.X … 

In 5.X : we move from 2/3 times to more than 100 for public ONE PAGE !

PR #3842, opened 2024-05-13
QE-920: Implemented a first version of the type transformer service

lajosarpad:
The purpose of this service is to have data transformers between types for the preview feature and to have general fallback logic whenever such a transformer is either not yet implemented, or not supported.


c-schmitz:
@coderabbitai review

PR #3844, opened 2024-05-14
QE-727: Question code IO transformation

mohabmes:


kevin-foster-uk:
We decided not to use this and store the value in app state instead.

PR #3845, opened 2024-05-14
Fixed issue #19404: unable to export survey with PHP 8.0

Shnoulle:
Dev: don't include @attributes array in xml exported
Dev: loose attributes, but are unused in case of survey export


application/helpers/export_helper.php
```diff
@@ -3144,6 +3144,11 @@ function writeXmlFromArray(XMLWriter $xml, $aData, $sParentKey = '')
 {
     $bCloseElement = false;
     foreach ($aData as $key => $value) {
+        if ($key == "@attributes") {
```
gabrieljenik, line 3147: Is this the right place for this filter?
This function is a writer. Should also have the logic for filtering the sources?

Maybe here better fit?
https://github.com/LimeSurvey/LimeSurvey/pull/3845/files#diff-1aecd900cde469d4c2ba64ab989062a075ed3ff9378180051d812b127a929dfeR3205

Shnoulle: > Is this the right place for this filter? 

I don't know :)

You're right about the place it was added

gabrieljenik (changes requested):
I think we should move the filter out of the writer.

PR #3847, opened 2024-05-15
MW-1149: template api

lajosarpad:
This is an API that serves the purpose of getting templates for the marketing site so they can be previewed


application/libraries/Api/Command/V1/SurveyTemplate.php
```diff
@@ -0,0 +1,114 @@
...
+            ->find('surveyls_survey_id = :sid and surveyls_language = :language', [
+            ':sid' => $surveyId,
+            ':language' => $language
+        ]);
+        $response = [];
+        if ($languageSettings) {
+            $response['title'] = $languageSettings->surveyls_title;
+            $response['subtitle'] = $languageSettings->surveyls_description;
+        }
+        $ch = curl_init();
```
olleharstedt, line 95: This will be hard to unit-test if you don't inject a Curl class ;)

PR #3851, opened 2024-05-15
Fixed issue #CT-778: wrong placeholder fields in html editor

twilligls:
When selecting placeholder fields for previous questions in the html editor via the { } button,
the generated qcodes where wrong for some occasions.

It was QUESTIONCODE.shown all the time.
for simple question types like textthat's ok, but subquestions/answeroptions, other option, scaleids need other formats.

What I adjusted here:

- For subquestions it is now QUESTIONCODE_SUBQCODE.shown
- For other options of those it is now  QUESTIONCODE_other (.shown is not working in that case)
- If the question has scale_ids then it is now QUESTIONCODE_SUBQCODE_SCALEID.shown

I can provide a test lss, if needed,  it is attached to task CT-778.

### DIFF
--- a/application/controllers/LimereplacementfieldsController.php
+++ b/application/controllers/LimereplacementfieldsController.php
@@ -212,8 +212,26 @@ class LimeReplacementFieldsController extends LSBaseController
                 $question = "[{$row['subquestion2']}] " . $question;
             }
 
+            $replacementCode = $row['title'];
+            if (array_key_exists('aid', $row) && $row['aid'] !== '') {
+                $replacementCode = $row['title'] . '_' . $row['aid'];
+                if (array_key_exists('scale_id', $row)) {
+                    $replacementCode = $replacementCode . '_' . $row['scale_id'];
+                }
+                if (strpos($replacementCode, '_other') === false) {
+                    $replacementCode = $replacementCode . '.shown';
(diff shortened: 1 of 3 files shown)


application/controllers/LimereplacementfieldsController.php
```diff
@@ -212,8 +212,26 @@ private function getChildQuestions(array $questions)
                 $question = "[{$row['subquestion2']}] " . $question;
             }
 
+            $replacementCode = $row['title'];
```
c-schmitz, line 215: This whole block looks very similar to collectQuestionReplacements()

c-schmitz (changes requested):

application/controllers/LimereplacementfieldsController.php
twilligls: I'll check soon

twilligls: This function would not fully deliver what we need. I could probably use it for parts of the placeholder code creation,
but it would produce another unneccessary db call.
We already have all previous questions data via createFieldMap().

I did update all the docblocks in this controller and factored the new functionality into a separate function for better maintainability

kevin-foster-uk (approved):

PR #3856, opened 2024-05-22
fixed issue #19570: javascript exception on survey taking on browsers…

TonisOrmisson:
Fixed issue #19570: javascript exception on survey taking on browsers not supporting localstorage

Since I am not sure which calls are for admin or only for survey, I changed usages that directly fixed the issue I was facing (checking timers) and some that seemed possible. 

Should have no change on browsers having localStorage


TonisOrmisson:
Also affects previous versions since ~sept 2020, so current V5/V3

assets/packages/limesurvey/survey.js
```diff
@@ -433,6 +433,9 @@ function activateSoftMandatory(){
 }
 
 function resetQuestionTimers(sid) {
+    if(!window.localStorage) {
```
Shnoulle, line 436: Then need to create a new mantis issue : Question timer didn't work without Local Storage :)

Shnoulle (approved):

assets/packages/limesurvey/survey.js
TonisOrmisson: I think we should just accept that timers will not work for browsers without localStorage? No localStorage seems to be a very low probability event, no? 

Bottom line here is I would rather have questionTimers malfunctioning quietly rather than whole javascript crashing on the page.

Shnoulle: > Bottom line here is I would rather have questionTimers malfunctioning quietly rather than whole javascript crashing on the page.

I accept the update for this reason :). This commit **must be merged** before creating another manitis issue :)

> I think we should just accept that timers will not work for browsers without localStorage? No localStorage seems to be a very low probability event, no?

Maybe, but then : must update manual and help of settings. Need a mantis issue for discussion.

assets/packages/limesurvey/survey.js
```diff
@@ -433,6 +433,9 @@ function activateSoftMandatory(){
...
+        return;
```
gabrieljenik, line 437: By applying this fix, what the participant experience would be like?
They would be able to take the survey without the timed questions?

If a survye amdin expected timedquestions, would it be ok to allow participants not to use it?
What if we raise an alert if this is the case?

TonisOrmisson: To be honest, I have not tested the timers, the aim for me was to get past javascript crashing.

If the participant does not have localStorage, the questionTimers could not have worked in any case since it uses the localStorage based on what I can read from the code. So logically assuming such users would go from a javascript exception (and a lot of things not potentially workining) to silently question timers not working.

Shnoulle: > By applying this fix, what the participant experience would be like? They would be able to take the survey without the timed questions?

Before : all javascript are broken 

> If a survye amdin expected timedquestions, would it be ok to allow participants not to use it? What if we raise an alert if this is the case?


Before it don't work too, then it's a different (and new issue) see my comment here : https://github.com/LimeSurvey/LimeSurvey/pull/3856#discussion_r1611140732

Shnoulle: PS : i always think timer script must move to a core external plugin.

TonisOrmisson: I see that some of the changes need the webpack files to be rebuilt, I will check those also and do a commit

TonisOrmisson: ok, I changed return to `null` instead of `0` in `_getTimerFromLocalStorage`. With 0 it ended up going instantly forward. Also I built the webpack files. 

if one has a timer question with no localStorage, the question with Timer now looks like this: 
image
and it will not trigger the timer.

without this fix, the page javascript throes exception:
image
the filter on page do not work, timer is not working, empty:
image

Shnoulle: > if one has a timer question with no localStorage, the question with Timer now looks like this: 

_now_ : you mean : current master ?

TonisOrmisson: I mean with THIS PR by "now"

Shnoulle: > I mean with THIS PR by "now"

OK then,

Without this PR
1. Timer didn't work
2. conditions didn't work

With this PR
1. Timer didn't work
2. conditions work

TonisOrmisson:
@Shnoulle im not sure what the standard is here. Can the "needs testing" label be removed. You added the label after I said I had not done any testing. After that I did actual testing and some code adjustments based on that. What is needed to be done to move on with merging this?

Shnoulle:
> @Shnoulle im not sure what the standard is here. Can the "needs testing" label be removed. You added the label after I said I had not done any testing. After that I did actual testing and some code adjustments based on that. What is needed to be done to move on with merging this?

@tiborpacalat need to test before it was merged. It's the current way to do 

1. Another dev do some review
2. Tibor with help of user or other dev do a real test

PR #3863, opened 2024-05-31
Dev simplify DataEntry

kw-pr:
Dev: Because usesleft is reduced by 1 in any case, this block can be simplified.

### DIFF
--- a/application/controllers/admin/DataEntry.php
+++ b/application/controllers/admin/DataEntry.php
@@ -1884,20 +1884,11 @@ class DataEntry extends SurveyCommonAction
                     }
                     // query for updating tokens uses left
                     $aToken = Token::model($surveyid)->findByAttributes(['token' => $_POST['token']]);
+                    $aToken->usesleft = ((int) $aToken->usesleft) - 1;
                     if (isTokenCompletedDatestamped($thissurvey)) {
-                        if ($aToken->usesleft <= 1) {
-                            $aToken->usesleft = ((int) $aToken->usesleft) - 1;
-                            $aToken->completed = $submitdate;
-                        } else {
-                            $aToken->usesleft = ((int) $aToken->usesleft) - 1;
-                        }
+                        $aToken->completed = $submitdate;
                     } else {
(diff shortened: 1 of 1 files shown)


application/controllers/admin/DataEntry.php
```diff
@@ -1884,20 +1884,11 @@ public function insert()
                     }
                     // query for updating tokens uses left
                     $aToken = Token::model($surveyid)->findByAttributes(['token' => $_POST['token']]);
+                    $aToken->usesleft = ((int) $aToken->usesleft) - 1;
                     if (isTokenCompletedDatestamped($thissurvey)) {
-                        if ($aToken->usesleft <= 1) {
```
ptelu, line 1888: keep this condition,  add "$aToken->completed = $submitdate;" in it

same for the other condition but with "$aToken->completed = 'Y';"
then it will be fine

ptelu: just had another look and need to make another adjustment, it can only be substracted after the condition was checked, so after the condition you need to add "$aToken->usesleft = ((int) $aToken->usesleft) - 1;" again

application/controllers/admin/DataEntry.php
```diff
@@ -1884,20 +1884,11 @@ public function insert()
...
+                        $aToken->completed = $submitdate;
                     } else {
-                        if ($aToken->usesleft <= 1) {
-                            $aToken->usesleft = ((int) $aToken->usesleft) - 1;
-                            $aToken->completed = 'Y';
-                        } else {
-                            $aToken->usesleft = ((int) $aToken->usesleft) - 1;
-                        }
+                        $aToken->completed = 'Y';
                     }
```
ptelu, line 1892: just had another look and you need to make another adjustment, it can only be substracted after the condition was checked, so you need to add the substraction behind the enitre condition block here "$aToken->usesleft = ((int) $aToken->usesleft) - 1;"

kw-pr:
You are right. Sorry about that. I make a new PR.

PR #3864, opened 2024-06-04
New feature: Survey preview by Rest API (MW-1149)

lajosarpad:
PR based on the accepted and merged https://github.com/LimeSurvey/LimeSurvey/pull/3847/files

This patch adds an API endpoint so that a requestor may send a request and receive a full preview structure in return. For example, one can send a request like

```
curl -X POST https://ls-ce/rest/v1/survey-template/571271 --insecure \
   -H 'Content-Type: application/json' \
   -d '{"language":"en"}'
```

and receive a JSON response like

```
{"title":"The Title","subtitle":"The subtitle","template":"<some HTML template>"}
```


c-schmitz:
@lajosarpad What does the patch exactly add? This is a public patch and it should have a proper describing commit message.

c-schmitz:
What about permissions? I cannot see that any permissions checks in the API?

lajosarpad:
> What about permissions? I cannot see that any permissions checks in the API?

This is a publicly available end point @c-schmitz

c-schmitz:
General question: If a survey admin makes a survey available by setting 'Link survey on public index page", would the admin expect that it becomes available on this API to everyone. I mean it is a quick way to grab/download someone elses survey.

lajosarpad:
> General question: If a survey admin makes a survey available by setting 'Link survey on public index page", would the admin expect that it becomes available on this API to everyone. I mean it is a quick way to grab/download someone elses survey. I think it is a security issue and this should not be public without authorization.

@c-schmitz I understand the concern, but it's already possible with or without this PR.

If you authenticate to a survey site as an admin and navigate to a survey page, then you have a PREVIEW button. By clicking on that button the survey will open up in a new tab.

But, if you copy the URL of that button and load the survey at a different browser where you are not logged in as an admin, then it will appear there too.

Example:

image

this is an inactive test survey. This is how its preview looks alike:

image

This is its preview in an incognito window where I'm not logged in

image

All this is possible independently of this PR.

If we have a security concern, then we will need a separate ticket to restrict the preview from showing the survey to unauthenticated users.

This current PR gets a user request and sends a cURL request on the user's behalf to the survey. So if the survey is restricted from accessing the survey, then the preview will not work and then the end point will not serve the survey either.

So your concern is a more general concern about survey preview access privileges and given the fact that here we restrict for survey previews, we merely load the page for the user which he could load anyways already.

c-schmitz:
yeah, that's a security issue right there.

c-schmitz:
Security  issue fixed with #3865

PR #3869, opened 2024-06-06
Dev simplify DataEntry

kw-pr:
Dev: Because usesleft is reduced by 1 in any case, this block can be simplified.


c-schmitz:
@coderabbitai review

c-schmitz:
@kw-pr I understand that this has been submitted a long time ago. So sorry for the lack of feedback. 
Would be available for fixing the issues in this PR or should we rather just close it?

kw-pr:
@c-schmitz 
Not sure about the issue the AI found but I think that has nothing to do with my change.

I simply found a this one line that is repeated 4 times in all 4 branches of the if/else construct. So I moved it out having just one line instead of 4. This left 2 of the 4 branches empty and I removed those, making this small code block simpler.

Close this if you like. I don't mind.

PR #3871, opened 2024-06-06
Fixed issue #19589: Multiple call of Survey->find for public survey

Shnoulle:
Dev: really low call on public, must check on admin
Dev: result is the same

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -110,11 +110,12 @@ class Question extends LSActiveRecord
      * @inheritdoc
      * replace under condition $oQuestion->survey to use Survey::$findByPkCache
      */
-    public function getRelated($name, $refresh=false, $params=array()) {
+    public function getRelated($name, $refresh = false, $params = array())
+    {
         if ($name == 'survey' && !$refresh && empty($params)) {
             return Survey::model()->findByPk($this->sid);
         }
-        return parent::getRelated($name, $refresh,$params);
+        return parent::getRelated($name, $refresh, $params);
     }
 
     /** @inheritdoc */
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -106,18 +106,6 @@ class Question extends LSActiveRecord
         return $model;
     }
 
-    /**
-     * @inheritdoc
-     * replace under condition $oQuestion->survey to use Survey::$findByPkCache
-     */
-    public function getRelated($name, $refresh = false, $params = array())
-    {
-        if ($name == 'survey' && !$refresh && empty($params)) {
-            return Survey::model()->findByPk($this->sid);
-        }
-        return parent::getRelated($name, $refresh, $params);
-    }
-
     /** @inheritdoc */
     public function tableName()
     {
@@ -152,6 +140,18 @@ class Question extends LSActiveRecord
             // when using question editor on create mode. Better use getQuestionTheme()
             'question_theme' => [self::HAS_ONE, 'QuestionTheme', ['question_type' => 'type', 'name' => 'question_theme_name']],
         );
+    }
+
+    /**
+     * @inheritdoc
+     * replace under condition $oQuestion->survey to use Survey::$findByPkCache
+     */
+    public function getRelated($name, $refresh = false, $params = array())
+    {
+        if ($name == 'survey' && !$refresh && empty($params)) {
+            return Survey::model()->findByPk($this->sid);
+        }
+        return parent::getRelated($name, $refresh, $params);
     }
 
     /**
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -106,6 +106,17 @@ class Question extends LSActiveRecord
         return $model;
     }
 
+    /**
+     * @inheritdoc
+     * replace under condition $oQuestion->survey to use Survey::$findByPkCache
+     */
+    public function getRelated($name, $refresh=false, $params=array()) {
+        if ($name == 'survey' && !$refresh && empty($params)) {
+            return Survey::model()->findByPk($this->sid);
+        }
+        return parent::getRelated($name, $refresh,$params);
+    }
+
     /** @inheritdoc */
     public function tableName()
     {
@@ -1544,6 +1555,7 @@ class Question extends LSActiveRecord
 
         return !$isAlreadySorted;
     }
+
 
     /**
      * Returns the highest question_order value that exists for a questiongroup inside the related questions.
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -187,7 +187,6 @@ class Survey extends LSActiveRecord implements PermissionInterface
         }
         $this->attachEventHandler("onAfterFind", array($this, 'afterFindSurvey'));
         $this->attachEventHandler("onAfterSave", array($this, 'unsetFromStaticPkCache'));
-
     }
 
     /**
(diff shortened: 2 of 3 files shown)


Shnoulle:
I check `Group Helper (ls\tests\GroupHelper)` after

Shnoulle:
This remove too on some situation  `system.db.CDbCommand.query(SELECT * FROM lime_templates t WHERE t.name='skelvanilla' LIMIT 1)` 317 times

TonisOrmisson:
It probably fixes the queries etc, but I personally tend to have a more general reluctance on this matter. 

Namely, I have a reluctance of using and depending on static class variables. In this case usage survey::$findByPkCache is essentially what solves the matter. 

Using static variables is bad practice, its not OOP, discussion on this subject generally say it's a quick fix but in the end will create more problems than it solves. 

I personally think we should rather be moving away from using non OOP code such as static variables etc. This pr will make us depend more on this, we are digging the hole deeper, not moving towards better code & best practices. 

I would ask some review from the core team on this

Shnoulle:
> I personally think we should rather be moving away from using non OOP code such as static variables etc. This pr will make us depend more on this, we are digging the hole deeper, not moving towards better code & best practices.

Maybe, but current situation : call 131 times this request for a public page make LimeSurvey 5 unusable in some server.
Moving to cache are surely the best solution, but current cache are really not tested …

TonisOrmisson:
by passing Survey object or languages to QuestionAttribute->getQuestionAttributes() as a parameter instead of querying it from inside of the method, it will end up 5-6 of the 130+ survey finds on your example case, I have it up and I have tested this on your referred test survey. 

Sure, its 66 calls to check, but I bet the Survey object is available all of the cases you call that from, so you simply have to pass the dependency on to the method and this specific problem is fixed without a) overriding framework and b) not having mode dependency on a bad practice. It's a number of calls, but in essence passing the dependency is a much simpler solution.  

I would vote all my hands in favor of passing the dependency here. And possibly look for similar issues elsewhere until we do not need the survey::$findByPkCache that much any more and it can be wiped entirely.

TonisOrmisson:
a buch of the calls come from here
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/QuestionTemplate.php#L143
the called attributes there are UNUSED ! :)

Shnoulle:
> a bunch of the calls come from here https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/QuestionTemplate.php#L143 the called attributes there are UNUSED ! :)

Good catch here, since i have issue too with this question theme request

Shnoulle:
> by passing Survey object or languages to QuestionAttribute->getQuestionAttributes() as a parameter instead of querying it from inside of the method, it will end up 5-6 of the 130+ survey finds on your example case, I have it up and I have tested this on your referred test survey.
> 
> I would vote all my hands in favor of passing the dependency here. And possibly look for similar issues elsewhere until we do not need the survey::$findByPkCache that much any more and it can be wiped entirely.

I like to have a really lesser call solution. Not dependent of number of questions in page or number of groups in survey.
Same for getQuestionAttribute ([compatible with plugin too : maybe just send to event](https://manual.limesurvey.org/BeforeQuestionRender)) and surely a lot of other.

It's really a **blocking point** when you move from 1 second to produce page to 5/6 seconds on similar servers (done for this)

TonisOrmisson:
remove this
[RenderLanguageSelector](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/QuestionTemplate.php#L143)

pass the relevant language in: 
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/SurveyRuntimeHelper.php#L272
and
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/SurveyRuntimeHelper.php#L1906
and 
[$aQuestionAttributes = QuestionAttribute::model()->getQuestionAttributes($ia[0]);](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/helpers/qanda_helper.php#L132)

and I get this survey calls down to 13 from 130+

Shnoulle:
@TonisOrmisson use cache here can really be easy and really improve for big system (with cache system).
Even for 1 request by page, if you have 1000 access by minutes, use cache can be really great, no ?

TonisOrmisson:
> @TonisOrmisson use cache here can really be easy and really improve for big system (with cache system).
> Even for 1 request by page, if you have 1000 access by minutes, use cache can be really great, no ?

Cache, sure. But the static variable survey::$findByPkCache is not cache.

I remember a few years ago we had a short discussion with someone from core team (maybe it was @olleharstedt ?) about whether we should start implement query cache. At the time I remember the MySql Query cache was "working" and I had the opinion that I would let the database engine worry about query caching. But Since V8.0 MySql retired the query cache entirely so I think yes, having a query cache enabled for some specific queries is reasonable. But not the static vars :) 

And also, I think we should take some focused time (like the case here) to identify cases where duplicate queries get huge hits and analyze if all the hits really are necessary. Looking at this case, It seems in many cases you can reduce the query by 80-90% by actually fixing the code (eg do not fetch all languages if you only need one etc).

olleharstedt:
> You can't : we need to separate ordering part and prefill part (and maybe token).

For randomized ordering and token attribute prefill? Yes, that makes sense.

> And about caching : remind to reset it

For sure :)

Shnoulle:
> > You can't : we need to separate ordering part and prefill part (and maybe token).
> 
> For randomized ordering and token attribute prefill? Yes, that makes sense.
> 
> Still, there are big chunks in that function that only needs to be run once for all participants, once the survey is activated (and does not change).

**TOTALLY** !! It's a must have !

Others one : 

https://github.com/LimeSurvey/LimeSurvey/blob/b5b022465765c1fbcec6d696e04a591d0168611b/application/helpers/expressions/em_manager_helper.php#L4005

https://github.com/LimeSurvey/LimeSurvey/blob/b5b022465765c1fbcec6d696e04a591d0168611b/application/models/services/QuestionAttributeFetcher.php#L8 unsure where, but question attributes settings are fixed during week/month until install. We don't need read again and again XML …

> 
> > And about caching : remind to reset it
> 
> For sure :)

Shnoulle:
@TonisOrmisson maybe create a singleton object ? For public survey usage principally ?

TonisOrmisson:
> @TonisOrmisson maybe create a singleton object ? For public survey usage principally ?

@Shnoulle I was wondering about this exactly yesterday :)

It seems logical that in public survey-filling context there is only one instance of the Survey object. But  at the same time, making a change like this in terms of changes in the code is similar to passing the survey object along as a dependency. Making it a Singleton seems to me, increases the complexity of the system, passing as dependency will have similar result but seems more straight forward, and also moving mode towards using DI is described in the [LS code quality guide](https://manual.limesurvey.org/Code_quality_guide)

Shnoulle:
> > @TonisOrmisson maybe create a singleton object ? For public survey usage principally ?
> 
> @Shnoulle I was wondering about this exactly yesterday :)
> 

Currently, I think of plugins too, plugins need a way to get survey object easily (and question object and question attribute for Question plugin) and quickly. Some event where it can be needed : afterGenerateToken, afterSurveyComplete, afterSurveyQuota, beforeQuestionRender, beforeSurveyEmail, beforeSurveyPage, saveSurveyForm…

Sometimes i use it in BeforeTwigRenderTemplate , but here : i can create my own static var in beforeSureyPage (and register at beforeTwigRenderTemplate if needed)

c-schmitz (approved):

Shnoulle:
> Need to update a function (with reset)

Not really related but potential issue

PR #3878, opened 2024-06-13
LA-74: Update Table Design and Widget for Column Filter

mohabmes:


Shnoulle:
Can we have a screenshot of the new version ?

PR #3885, opened 2024-06-14
Fixed issue [security] #19610: Security: queXML PDF discloses version specific information

adamzammit:
Fixed issue #19610: Security: queXML PDF discloses version specific information

Dev: Removes author email from queXMLPDF output
Dev: Ensures PDF output not cached

### DIFF
--- a/application/libraries/admin/pdf.php
+++ b/application/libraries/admin/pdf.php
@@ -230,6 +230,208 @@ class pdf extends TCPDF
     }
 
 
+	/**
+     * Override _putXMP function to ensure version doesn't appear in metadata
+     * Put XMP data object and return ID.
+	 * @return int The object ID.
+	 * @since 5.9.121 (2011-09-28)
+	 * @protected
+	 */
+	protected function _putXMP() {
+		$oid = $this->_newobj();
+		// store current isunicode value
+		$prev_isunicode = $this->isunicode;
+		$this->isunicode = true;
+		$prev_encrypted = $this->encrypted;
+		$this->encrypted = false;
+		// set XMP data
+		$xmp = '<?xpacket begin="'.TCPDF_FONTS::unichr(0xfeff, $this->isunicode).'" id="W5M0MpCehiHzreSzNTczkc9d"?>'."\n";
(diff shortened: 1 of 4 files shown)


adamzammit:
updated to remove TCPDF version

application/libraries/admin/pdf.php
```diff
@@ -230,6 +230,208 @@ public function setConfig($tcpdf)
...
 
 
+	/**
+     * Override _putXMP function to ensure version doesn't appear in metadata
+     * Put XMP data object and return ID.
+	 * @return int The object ID.
+	 * @since 5.9.121 (2011-09-28)
+	 * @protected
+	 */
+	protected function _putXMP() {
```
Shnoulle, line 240: Just remove https://github.com/tecnickcom/TCPDF/blob/951eabf0338ec2522bd0d5d9c79b08a3a3d36b36/tcpdf.php#L9691 ?

application/libraries/admin/pdf.php
```diff
@@ -230,6 +230,208 @@ public function setConfig($tcpdf)
...
+	}
+
+	/**
+     * Override _putinfo function to ensure version doesn't appear in PDF metadata
+     * Adds some Metadata information (Document Information Dictionary)
+	 * (see Chapter 14.3.3 Document Information Dictionary of PDF32000_2008.pdf Reference)
+	 * @return int object id
+	 * @protected
+	 */
+	protected function _putinfo() {
```
Shnoulle, line 391: Maybe leave the link but set is as comment with comment ?

```
		// default producer remove for LimeSurvey mantis issue #XXXX
		// $out .= ' /Producer '.$this->_textstring(TCPDF_STATIC::getTCPDFProducer(), $oid);
```

Shnoulle (approved):
Maybe add more comments ? about difference with TCPDF

Just in case

application/libraries/admin/pdf.php
adamzammit: It would be easier just to edit the TCPDF code but as it is vendor supplied and could be updated I've used an override

PR #3887, opened 2024-06-14
Fixed issue [security] #19608: XSS under user management (thanks to OIZ Stadt Zürich and Protec7)

Shnoulle:
Dev: encode needed name
Dev: check if ajax to renderPartial


gabrieljenik (approved):
I think there is a little bit more here than the secuirity fix. Should be tested as well

Shnoulle:
> I think there is a little bit more here than the secuirity fix. Should be tested as well

Not really since there are no GET url shown in GUI

PR #3892, opened 2024-06-18
Fixed issue #19309: Can not click item to get into Group Information if the name is empty

gabrieljenik:


Shnoulle:
No way to fix via a bootstrap class ?
A https://getbootstrap.com/docs/5.3/utilities/spacing/ utilitiy ?

gabrieljenik:
> No way to fix via a bootstrap class ? A https://getbootstrap.com/docs/5.3/utilities/spacing/ utilitiy ?

I think this is solution is ok.
If not, we would need to find every tag with question_text_ellipsize class and add new utility classes

Shnoulle:
> > No way to fix via a bootstrap class ? A https://getbootstrap.com/docs/5.3/utilities/spacing/ utilitiy ?
> 
> I think this is solution is ok. If not, we would need to find every tag with question_text_ellipsize class and add new utility classes

Since we use boostrap : I really think we must use boostrap system and utilities.

1. This allow Theme developer to create new template more easily, 
2. No reason to add BS if we don't use it.

Else : here the issue is not about all `question_text_ellipsize` but for empty group : let fix this part only, no ?

Shnoulle:
ping @ptelu or another reviewer :)

gabrieljenik:
> 2\. No reason to add BS if we don't use it.

Using BS doesn;t mean we can't use other stuff as well.
Also, we implemented something similar to your initial suggestion :)

Maybe another set of eyes could be good, either for reviewing or finding a better fix

Shnoulle:
> > 2. No reason to add BS if we don't use it.
> 
> Using BS doesn;t mean we can't use other stuff as well. 

Not of BS have already a solution.

> Maybe another set of eyes could be good, either for reviewing or finding a better fix

Ask to @ptelu

c-schmitz:
@coderabbitai review

Shnoulle:
<s>Must check if still valid. If i remind : we add &nsp; on the menu to avoid such situation. Then we have always a content inside the element.</s>

Not the case, still valid.

gabrieljenik:
Closed in favor of https://github.com/LimeSurvey/LimeSurvey/pull/4874.

PR #3895, opened 2024-06-21
Fixed issue [security] #19604: The reset password link is reusable after password change

Shnoulle:
Dev: delete validation_key afer success

### DIFF
--- a/application/controllers/admin/Authentication.php
+++ b/application/controllers/admin/Authentication.php
@@ -246,6 +246,9 @@ class Authentication extends SurveyCommonAction
             if (($password !== null && $passwordRepeat !== null) && ($password === $passwordRepeat) && $oPasswordTestEvent->get('passwordOk')) {
                 //now everything is ok, save password
                 $user->setPassword($password, true);
+                // And remove validation_key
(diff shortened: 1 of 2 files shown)


application/controllers/admin/Authentication.php
```diff
@@ -258,7 +261,7 @@ public function newPassword()
             'errorExists' => $errorExists,
             'errorMsg' => $errorMsg,
             'randomPassword' => $randomPassword,
-            'validationKey' => $user->validation_key
+            'validationKey' => $validation_key
```
gabrieljenik, line 264: Probably not part of this bug, but seems $validationKey is not sanitized.

https://github.com/LimeSurvey/LimeSurvey/pull/3895/files#diff-70d800faf45c8b44b4c4ae10435793c94be163ca4332e1701ab9c95812df4e7bR207

Or the standard anitization done by getParam('param') is enough?
Usually that's enough, but as we are then sending tha variable to the view, maybe we need something else as well?

Shnoulle: > No sanitization here, but you totally right : if it's invalid : it must be encoded in view. BUT : view didn't show it.

No real security issue : shown in input only of valid, and generated by LimeSurvey.

PR #3896, opened 2024-06-21
Fixed issue [security] #19603: All admin can be deactivated by user with only users delete permission

Shnoulle:
Dev: check permission with user->canEdit


gabrieljenik (approved):
Code structure looks good

PR #3897, opened 2024-06-24

Shnoulle:
Fixed issue : Multiple choice questions columns not in one (ul) tag.


Shnoulle:
@ptelu : there are a difference in view : if you have a choice with A1, A2, A3, A1 , with 2 coumns

With this solution

A1 A2
A3 A4

with columns and current solution

A1 A3
A2 A4


I can create a option to choose type : allowing columns or flex solution by a new question settings

Shnoulle:
columns is better at 16 than flex.
Strange issue about the label with flex.

Shnoulle:
columns are better at 16 than flex, strange issue with label

columns
columns-12
flex
flex-16

Shnoulle:
Close this one ,
Create it as new feature after, or can be done by theme only

https://github.com/LimeSurvey/LimeSurvey/pull/3038

PR #3901, opened 2024-07-01
QE-1042 - Core fixes for duplicate field names in response table

c-schmitz:
- QuestionID and groupID removed from the field name
- Response tables are now named response_<surveyID> (instead of survey_<surveyId>)
- Timings tables are now named timings_<surveyID> (instead of survey_<surveyId>_timings)

This is purely experimental and only meant as a base of discussion


Shnoulle:
What happens for `{123X23X12}` ? All expression must be rewritten … and rechecked… and tested

I think in this case : I prefer to move to `Q00` and `Q00_SQ01` etc … for responses_XXXX columns, really more interesting. The Expression Manager code directly

c-schmitz:
SGQA codes have been marked as obsolete for years, now.

PR #3904, opened 2024-07-01
Fixed issue #19607: Outdated libraries used with known vulnerabilities

gabrieljenik:
- Update TCPDF and other composer libraries


Shnoulle (approved):
Must be done … but right : must do a complete test !

PR #3912, opened 2024-07-10
Fixed issue #19643: Date question with restrictions does not show up like in versions 3 and 5

gabrieljenik:


application/extensions/DateTimePickerWidget/DateTimePicker.php
```diff
@@ -449,4 +453,33 @@ private function getShowComponent($component)
...
+    private function getViewDate()
+    {
+        if (!empty($this->value)) {
+            return "'" . $this->value . "'";
+        }
+
+        $minDate = $this->getValue('data-minDate', $this->htmlOptions, null);
+        if (isset($minDate)) {
+            // If min date is in the future, we set the view date to the min date
+            if (strtotime($minDate) > time()) {
```
Shnoulle, line 470: According to https://github.com/LimeSurvey/LimeSurvey/blob/9dc2b8e49f2c32c360e0a37dcb80194a1a7db90a/application/core/QuestionTypes/Date/RenderDate.php#L79

minDate can be 

1. A date like "YYYY-mm-dd" : OK, it work
2. An expression  : https://github.com/LimeSurvey/LimeSurvey/blob/9dc2b8e49f2c32c360e0a37dcb80194a1a7db90a/application/core/QuestionTypes/Date/RenderDate.php#L94 

Seems 2 are not covered here.

And there are 2 expression typoe : fixed (variable out of this group) , dynamic (variable in this page)

Shnoulle (commented):
Seems date with expression manager are not covered here.

Shnoulle (approved):

Shnoulle (approved):

PR #3918, opened 2024-07-15
LA-73: Update dashboard

mohabmes:


application/extensions/SideBarWidget/SideBarWidget.php
```diff
@@ -0,0 +1,79 @@
...
+            $boxData['ico'] = $box->getIconName();
+            $boxData['description'] = $box->desc;
+
+            // default permission if usergroup is not within expected values
+            $canSeeBox = false;
+            if (Permission::model()->hasGlobalPermission('superadmin', 'read')) {
+                $canSeeBox = true;
+            }
+            // everyone can see the box
+            if ($box->usergroup === -1) {
```
twilligls, line 46: checking usergroup value  for int, but value is stored as string in db.

This made my restricted user having no sidebar at all.

twilligls (changes requested):
see comments in code

application/extensions/SideBarWidget/SideBarWidget.php
mohabmes: The usergroup is of type int(11), not string, so it shouldn't be a problem, right?

application/controllers/SurveyAdministrationController.php
```diff
@@ -3390,4 +3390,55 @@ public function actionDeleteUrlParam()
...
+     * @throws CException
+     */
+    public function actionBoxList()
+    {
+        $limit = (int)App()->request->getQuery('limit');
+        $page = (int)App()->request->getQuery('page');
+        $offset = ($page ) * $limit;
+
+        if ($state = App()->request->getQuery('active')) {
+            $surveys = Survey::model()->findAll(
```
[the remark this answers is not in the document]
mohabmes: fixed

tests/functional/backend/CreateSurveyTest.php
```diff
@@ -85,7 +85,7 @@ public function testCreateSurvey()
             // Click on big "Create survey" button.
             $link = self::$webDriver->wait(10)->until(
                 WebDriverExpectedCondition::elementToBeClickable(
-                    WebDriverBy::cssSelector('#card-1[data-url]')
+                    WebDriverBy::linkText('Create survey')
```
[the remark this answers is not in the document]
mohabmes: fixed

application/extensions/SideBarWidget/SideBarWidget.php
twilligls: $box->usergroup is string.
In my case, I test with a permission restricted user, but I have boxes with usergroup: '-1'
Those should be shown to the user, but are not because the condition ist strict on type int

twilligls: Works fine now.
It's a yii thing, although the db col is integer add the model also declares the property as integer, 
it come as a string in the ActiveRecord obj.

application/extensions/admin/BoxesWidget/views/box.php
```diff
@@ -0,0 +1,122 @@
...
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        <?php elseif ($item['type'] == 2) : ?>
+            <div class="box-widget-card card-link m-2 align-middle d-inline-block"
+                 data-url="<?php echo $item['link'] ?>"
+                <?php if ($item['color']) : ?>
+                    style="color:<?php echo $item['color'] ?>;border: solid 2px <?= $item['color'] ?>"
```
ptelu, line 103: no inline styles, use css selectors

application/extensions/admin/BoxesWidget/assets/boxes-widget.js
```diff
@@ -0,0 +1,44 @@
+let cardHeaderRedirection = function () {
+    $(".box-widget-card-header, .box-widget-card-text").click(function (event) {
+        console.log()
```
ptelu, line 3: when hovering over box-widget-card-text not the entire area is turning the cursor into a pointer, either adjust what should be clickable or display a cursor for it depending oin the design

application/extensions/admin/BoxesWidget/views/box.php
```diff
@@ -0,0 +1,122 @@
...
+                                    </a>
+                                <?php endif; ?>
+
+                                <?php echo $item['survey']->getButtons(); ?>
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        <?php elseif ($item['type'] == 2) : ?>
```
ptelu, line 99: BoxesWidget::TYPE_LINK and true equal operators

ptelu (changes requested):

PR #3920, opened 2024-07-16
Fixed issue [security] : 19639: Host Header Injection in password reset function (thanks to paoloelia)

Shnoulle:
Dev: allow adding allowedHosts in config.php file

### DIFF
--- a/application/core/LSHttpRequest.php
+++ b/application/core/LSHttpRequest.php
@@ -298,7 +298,7 @@ class LSHttpRequest extends CHttpRequest
     }
 
     /**
-     * Check if an url are in allowed hosb (igf exist)
+     * Check if an url are in allowed host (if exist)
      * @var string $hostInfo
      * @throw Exception
      * @return void
@@ -308,7 +308,7 @@ class LSHttpRequest extends CHttpRequest
         $allowedHosts = App()->getConfig('allowedHosts');
         if (!empty($allowedHosts) && is_array($allowedHosts)) {
             $host = parse_url($hostInfo, PHP_URL_HOST);
-            if (!in_array($host, $allowedHosts)) {
+            if ($host && !in_array($host, $allowedHosts)) {
                  throw new CHttpException(400, gT("The request hostname is invalid.", 'unescaped'));
             }
         }
--- a/application/core/LSHttpRequest.php
+++ b/application/core/LSHttpRequest.php
@@ -293,6 +293,18 @@ class LSHttpRequest extends CHttpRequest
     public function getHostInfo($schema = '')
     {
         $hostInfo = parent::getHostInfo($schema);
+        self::checkIsAllowedHost($hostInfo);
+        return $hostInfo;
+    }
+
+    /**
+     * Check if an url are in allowed hosb (igf exist)
+     * @var string $hostInfo
+     * @throw Exception
+     * @return void
+     */
+    public static function checkIsAllowedHost($hostInfo)
+    {
         $allowedHosts = App()->getConfig('allowedHosts');
         if (!empty($allowedHosts) && is_array($allowedHosts)) {
             $host = parse_url($hostInfo, PHP_URL_HOST);
@@ -300,6 +312,5 @@ class LSHttpRequest extends CHttpRequest
                  throw new CHttpException(400, gT("The request hostname is invalid.", 'unescaped'));
             }
         }
-        return $hostInfo;
     }
 }
--- a/application/core/LSHttpRequest.php
+++ b/application/core/LSHttpRequest.php
@@ -293,8 +293,8 @@ class LSHttpRequest extends CHttpRequest
     public function getHostInfo($schema = '')
     {
         $hostInfo = parent::getHostInfo($schema);
-        if (!empty(App()->getConfig('allowedHosts')) && is_array(App()->getConfig('allowedHosts'))) {
-            $allowedHosts = App()->getConfig('allowedHosts');
+        $allowedHosts = App()->getConfig('allowedHosts');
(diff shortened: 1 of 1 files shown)


application/core/LSHttpRequest.php
```diff
@@ -286,4 +285,21 @@ public function isRestRequest()
...
+
+    /**
+     * @inheritdoc
+     * Check host with config['allowedHost'] if it set
+     */
+    public function getHostInfo($schema = '')
+    {
+        $hostInfo = parent::getHostInfo($schema);
+        if (!empty(App()->getConfig('allowedHosts')) && is_array(App()->getConfig('allowedHosts'))) {
+            $allowedHosts = App()->getConfig('allowedHosts');
```
gabrieljenik, line 297: Small Comment:

Shouldn't we pick up the variale before the if ?
That would aboid calling the same method 3 times.

application/core/LSHttpRequest.php
```diff
@@ -286,4 +285,21 @@ public function isRestRequest()
...
     }
+
+    /**
+     * @inheritdoc
+     * Check host with config['allowedHost'] if it set
+     */
+    public function getHostInfo($schema = '')
+    {
+        $hostInfo = parent::getHostInfo($schema);
+        if (!empty(App()->getConfig('allowedHosts')) && is_array(App()->getConfig('allowedHosts'))) {
```
gabrieljenik, line 296: From here and until the end of the function, wouldn't it be better to make it a function of its own?
CheckAllowedHost() or IsAllowedHost($host)

Then, the function will be easier to automatically test and enrich in the future.

gabrieljenik:
I know the validation will be done on lot of places, as the method is called by Yii directly in lot of places.
Still, is there any other place we could add this to make it more explicit?

gabrieljenik:
Sorry about all the comments :)

application/core/LSHttpRequest.php
Shnoulle: :+1:

Shnoulle:
> I know the validation will be done on lot of places, as the method is called by Yii directly in lot of places. Still, is there any other place we could add this to make it more explicit?

I don't understand : Question: where to test host ? Answer : in getHostInfo function !

It can not be more explicit

application/core/LSHttpRequest.php
Shnoulle: If you set a bad host : you get a 400 error on all admin page : easy to test

Here it's a 5 lines of code : really need a specific function ?

gabrieljenik:
> > I know the validation will be done on lot of places, as the method is called by Yii directly in lot of places. Still, is there any other place we could add this to make it more explicit?
> 
> I don't understand : Question: where to test host ? Answer : in getHostInfo function !
> 
> It can not be more explicit

Don't agree. That function is to get information about the request. 
Although the approach in the PR works, may not be the best place to add a validation.

application/core/LSHttpRequest.php
gabrieljenik: > If you set a bad host : you get a 400 error on all admin page : easy to test

I mean automatic testing.

> Here it's a 5 lines of code : really need a specific function ?

Well, it wouldn't be the worst having a 5 lines function either.
If the rules get more complex, and it is not on a function of its own, automatic testing would get harder.

gabrieljenik:
I am pretty wide open to hear someone else review. Maybe that can ease things

gabrieljenik (approved):
Code structure looks good

application/core/LSHttpRequest.php
Shnoulle: > I mean automatic testing.

Yes, it's easy to add it :) : just call admin page and check if its a 400 header.

> Well, it wouldn't be the worst having a 5 lines function either.

Separated

application/core/LSHttpRequest.php
```diff
@@ -286,4 +285,32 @@ public function isRestRequest()
...
+     * Check if an url are in allowed hosb (igf exist)
+     * @var string $hostInfo
+     * @throw Exception
+     * @return void
+     */
+    public static function checkIsAllowedHost($hostInfo)
+    {
+        $allowedHosts = App()->getConfig('allowedHosts');
+        if (!empty($allowedHosts) && is_array($allowedHosts)) {
+            $host = parse_url($hostInfo, PHP_URL_HOST);
```
Shnoulle, line 310: @gabrieljenik maybe lust check $host != false ?

application/core/LSHttpRequest.php
```diff
@@ -286,4 +285,32 @@ public function isRestRequest()
...
+            if ($host && !in_array($host, $allowedHosts)) {
```
gabrieljenik, line 311: If no host, the validation is not done?

Shnoulle: No, because `'/resetpassword.php'` is always a valid host

It can not happen for 19639 issue , because `parent::getHostInfo($schema);` always return an host.

Current usage : there are always an host.

gabrieljenik: if there is always a host, why checking it (on the if statement)?
What about throwing an exception if no host is found? Just in case

Shnoulle: PS : see the comment : “Check if an url are in allowed host **(if exist)**”

gabrieljenik: I don't completely agree, but we don't need to 100% of times :) 
I have already set the code review as OK last week I think

Shnoulle: I think if there are no hosts here : there are no security issue. Host not set : no security issue in URL :)

We need to check if it's a valid host for Host Injection. 

No Host _then_ no injection _then_ no issue

PR #3923, opened 2024-07-17
Accessible Limesurvey Version 6.5.3

dlsckathonrw:
New feature: Accessible Limesurvey Version 6.5.3 with Accessible WCAG Vanilla Front Theme
Dev: DLSC Katho NRW // Anblik.com


application/views/admin/emailtemplates/emailtemplates_view.php
```diff
@@ -23,14 +23,19 @@
     LSYii_ClientScript::POS_BEGIN);
 
 ?>
+<<<<<<< HEAD
+<div class="side-body <?php echo getSideBodyClass(false); ?>">
+    <h3 role="heading" aria-level="2"><?php eT("Edit email templates"); ?></h3>
+=======
```
c-schmitz, line 29: Hi, these <<<<<<< HEAD lines are unresolved conflicts.

application/views/admin/expressions/test/survey_logic_file.php
```diff
@@ -10,8 +10,13 @@
 echo viewHelper::getViewTestTag('surveyLogicFile');
 ?>
 
+<<<<<<< HEAD
```
c-schmitz, line 13: Here....

application/views/admin/participants/summary_view.php
```diff
@@ -17,48 +17,54 @@
...
-                        <th>
+                        <th scope="row">
                             <?php eT("Participants you have shared"); ?>
                         </th>
                         <td>
                             <?php echo $shared . ' / ' . $totalrecords; ?>
                         </td>
                     </tr>
                     <tr>
+<<<<<<< HEAD
```
c-schmitz, line 54: and here....

c-schmitz (changes requested):

c-schmitz:
Could you please also update the description and give a detailed description of what this patch changes. Thank you!

application/views/admin/participants/summary_view.php
dlsckathonrw: Yes, I see that now. On my local system it shows I have merged without any conflicts, but it seems like it is not done correctly. I will delete this pull request and do it again. Sorry for the inconvenience

application/views/admin/emailtemplates/emailtemplates_view.php
anblikauditteam: This is done!

dlsckathonrw:
Updated the description and fixed conflicts

themes/survey/wcag-vanilla-theme/scripts/custom.js
```diff
@@ -0,0 +1,219 @@
...
+        });
+    }
+    $('.answer-container').each(function() {
+        const that = this;
+        $(this).find('ul.ls-answers.answers-list.radio-list > li').each(function() {
+            $(this).attr("role", "none");
+            let label = $(this).find('label');
+            let input = $(this).find('input');
+            label.attr('aria-hidden', true);
+            input.attr("aria-label", label.text().trim());
```
Shnoulle, line 54: No !!! Label can be updated by Exprssin manager : in this cas : break whole.

Shnoulle:
My opinion : Here there are patch for admin part and patch for public part : complex, too complex. Best is to separate issues.

> Improving focus management to provide a better navigation experience for keyboard users.

In public part ? I don't see it. See https://bugs.limesurvey.org/view.php?id=18713 (and a core solution)

> Enhancing color contrast to meet accessibility standards.

For public part !: only the new specific theme ? Right ?

> Ensuring that status messages and notifications are announced appropriately by screen readers.

For public : all this part must come from core, not from custom.js  . I do it via twig files, but https://gitlab.com/SondagesPro/SurveyThemes/skelvanilla_ls6/-/commit/82fc6563786364cad5a54e3558af910c8705f91e but muts be on core.  see https://bugs.limesurvey.org/view.php?id=19648 


> Addressing issues with headings, table captions, and other structural elements to provide a more accessible and logical reading order.

Public part ?  It's about the hidden label in td ? I think we muts inverse system : by default show table as block  , and show it as tabvke only for @media screen up to XXXpx

AnblikAudit:
Would it be good if you send fixes one by one, may I know if someone has complete local dev setup instructions?

lajosarpad:
This PR is too big. I believe it needs to be broken down into several, smaller PRs instead of doing it at once.

anblikauditteam:
Please provide the local setup and build instructions so that we can build in your way and also split this pull into small pieces.

c-schmitz:
Please merge the master branch to your current development branch to resolve the conflicts.

c-schmitz:
Closing due to missing feedback from author.

PR #3925, opened 2024-07-19
Fixed issue #18207: User entered value can be parsed by expression script in valid answer string

Shnoulle:
Dev: use dedicated self::htmlSpecialCharsUserValue
Dev: not inside html attribute, no need encode quote.


gabrieljenik (approved):
Code structure looks ok

PR #3926, opened 2024-07-22
Make Limesurvey working on Apache 2.4 without access_compat module

weberhofer:
19662: Limesurvey does not work correctly with apache 2.4 without access_compat module


weberhofer:
Sorry, that was a false commit

PR #3927, opened 2024-07-22
Improve .htaccess files to work with apache 2.4 w/o mod_access_compat

weberhofer:
Manttis issue 19662: Limesurvey does not work correctly with apache 2.4 without access_compat module


application/.htaccess
```diff
@@ -1 +1,13 @@
-Deny from all
\ No newline at end of file
+<IfVersion < 2.4>
+    Order allow,deny
+    Deny from all
+</IfVersion>
```
Shnoulle, line 4: Why not

```
<IfVersion < 2.4>
    Order allow,deny
    Deny from all
</IfVersion>
<IfVersion >= 2.4>
        Require all denied
</IfVersion>
```

application/models/User.php
```diff
@@ -245,8 +245,11 @@ public function getDateFormat()
      */
     public function getFormattedDateCreated()
     {
-        $dateCreated = $this->created;
-        $date = new DateTime($dateCreated);
+	$dateCreated = $this->created;
+	if (empty($dateCreated)) {
```
Shnoulle, line 249: Not related ;)

Shnoulle (commented):
Why check `mod_access_compat` ?

Just check version

application/.htaccess
weberhofer: I think, you are partially right. When users are on 2.4+, they can in every case use the `Require` setting.
But if they have additionally enabled `access_compat` they could have set something additional. It's not fully clear which setting is preferred (Apaches documentation says, one should not use require and Allow/Deny together) the compat settings should be overwritten, too.

I will update the pull-request.

Shnoulle (approved):
Seems OK

weberhofer:
I'll prepare another pull request for the other issue with the non-set creation date of users tomorrow.

Shnoulle:
Tested : :+1: have a 500 error before.

Thank you !

Shnoulle:
@c-schmitz : can we merge it ?

PR #3928, opened 2024-07-22
Fixed issue #19492: Different minimum password requirements inside and outside of LimeSurvey

gabrieljenik:


Shnoulle (approved):
No issue in code (but I don't understand why it fixes the issue)

For testing :)

Shnoulle:
I test it and previous next are OK now.

When update the minimal date from another question : previous was set to enable

Good job :+1:

PR #3930, opened 2024-07-24
Fixed issue CT-907: Wrong File upload question texts in response view

mohabmes:


application/controllers/ResponsesController.php
```diff
@@ -246,6 +246,8 @@ public function actionView(int $surveyId, int $id, string $browseLang = ''): voi
                 }
                 $qidattributes = QuestionAttribute::model()->getQuestionAttributes($questions[$field['qid']]);
 
+                $question = viewHelper::getFieldText($field);
```
twilligls, line 249: I think line 220 can be removed then?

twilligls (approved):

PR #3933, opened 2024-07-31
fixed issue #19675 : error Non-static method CActiveRecord::findAll() cannot be called…

loiczam:
If we set a value for iFileUploadTotalSpaceMB > 0 Limesurvey crash with this error :

`Non-static method CActiveRecord::findAll() cannot be called statically`

Using model instead static call fix it


loiczam:
fixed issue #19675

Shnoulle (approved):
:+1: 

Thank you

c-schmitz (approved):

PR #3935, opened 2024-08-01
Fixed a typo

lajosarpad:
This is a typo that we missed and we should remove as soon as possible


olleharstedt:
Hm Psalm didn't pick this up? In the CI pipeline?

PR #3939, opened 2024-08-07
Fixed issue #19658: Condition designer - first option in Multiple short text (Type: Q) question is always assigned predefined (No Answer) value

gabrieljenik:
Only consider the `canswers` field if the selected tab is the one related to the field. That solves an initialization problem which was causing to save unexpected values.


Shnoulle (approved):
No issue in this update. 

But it showed that all code are complex 🤪

PR #3940, opened 2024-08-19
Fixed issue #19587: The theme option 'Background image' has a space in the name, so the image can't be loaded

gabrieljenik:


Shnoulle (approved):
Did we need to fix XSS on such part ?

PR #3942, opened 2024-08-20
Fixed issue #19713: Botstrap Buttons Multiple - When unchecking an exclusive option, the other options don't get re-enabled

gabrieljenik:


assets/packages/limesurvey/survey.js
```diff
@@ -66,7 +66,7 @@ function triggerEmRelevanceSubQuestion(){
                     $(item).prop("disabled", false );
                 });
             } else {
-                $(event.target).find('.answer-item input').each(function(itrt, item ){
+                $(event.target).find('.answer-item input, input.button-item').each(function(itrt, item ){
```
Shnoulle, line 69: My opinion : twig file must be updated, it's an issue in theme.

Else : next time we add `input.my_item, .mybroken-theme-item .tag`

Can you try adding answer-item class at https://github.com/LimeSurvey/LimeSurvey/blob/59d73eeb8067ef5f916341274c670eda33aaa5fd/themes/question/bootstrap_buttons_multi/survey/questions/answer/multiplechoice/rows/answer_row.twig#L24 ?

Shnoulle (commented):
Need to fix theme, to avoid adding more and more class.

Shnoulle:
answer_item class **must** be at https://github.com/LimeSurvey/LimeSurvey/blob/59d73eeb8067ef5f916341274c670eda33aaa5fd/themes/question/bootstrap_buttons_multi/survey/questions/answer/multiplechoice/rows/answer_row.twig#L23 

Like the original one https://github.com/LimeSurvey/LimeSurvey/blob/59d73eeb8067ef5f916341274c670eda33aaa5fd/application/views/survey/questions/answer/multiplechoice/rows/answer_row.twig#L22

gabrieljenik:
> answer_item class must be at

What about the question themes that were already created based on this? Wouldn't that be a backward incompratibility / or kind of like a break in the interface?

Shnoulle:
> > answer_item class must be at
> 
> What about the question themes that were already created based on this? Wouldn't that be a backward incompratibility / or kind of like a break in the interface?

It's an issue in question theme, i don't understand why we need to update core for this.

Again : if question theme developper create any class name without reading https://manual.limesurvey.org/Theme_editor#Global_classes_for_question_part we MUST NOT fix it in core.

next time : `input.my_item, .mybroken-theme-item .tag` ?

gabrieljenik:
I agree a change in theme is needed. I believe there's too much risk of doing it in master (lot of unexpected) on a minor release. I would solve it like this now and then in dev make it straight.

tiborpacalat:
This has already been resolved with another ticket, sorry :(

Shnoulle:
> This has already been resolved with another ticket, sorry :(

Can you show the fix ?

PR #3944, opened 2024-08-23
Fixed issue #19659: Conditions based on {LANG} variable fail

Shnoulle:
Dev: move LANG to knownVars
Dev: add comment in replacements_helper

### DIFF
--- a/application/helpers/replacements_helper.php
+++ b/application/helpers/replacements_helper.php
@@ -18,6 +18,9 @@ if (!defined('BASEPATH')) {
 
 /**
 * This function replaces keywords in a text and is mainly intended for templates
+* Replacement done on this function can not be used in Expression for condition or equation
+* If you want keywords available on condition, use LimeExpressionManager::setValueToKnowVar
+* Or add it in LimeExpressionManager->setVariableAndTokenMappingsForExpressionManager
 * If you use this functions put your replacement strings into the $replacements variable
 * instead of using global variables
 * NOTE - Don't do any embedded replacements in this function.  Create the array of replacement values and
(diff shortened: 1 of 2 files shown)


application/helpers/replacements_helper.php
```diff
@@ -18,6 +18,9 @@
 
 /**
 * This function replaces keywords in a text and is mainly intended for templates
+* Replacement done on this function can not be used in Expression for condition or equation
+* If you want keywords available on condition, use LimeExpressionManager::setValueToKnowVar
+* Or add it in LimeExpressionManager->setVariableAndTokenMappingsForExpressionManager
```
gabrieljenik, line 23: So, if they are added as known vars, the variables will be available for both replacement and EM ?

gabrieljenik (approved):

application/helpers/replacements_helper.php
Shnoulle: Yes, with [LimeExpressionManager::setValueToKnowVar](https://github.com/LimeSurvey/LimeSurvey/blob/59d73eeb8067ef5f916341274c670eda33aaa5fd/application/helpers/expressions/em_manager_helper.php#L7769) or in [LimeExpressionManager->setVariableAndTokenMappingsForExpressionManager](https://github.com/LimeSurvey/LimeSurvey/blob/59d73eeb8067ef5f916341274c670eda33aaa5fd/application/helpers/expressions/em_manager_helper.php#L3312)

For plugins : https://github.com/LimeSurvey/LimeSurvey/blob/59d73eeb8067ef5f916341274c670eda33aaa5fd/application/helpers/expressions/em_manager_helper.php#L4005 this event are done for this.

gabrieljenik: Then I would amend a bit the comment as to make it more clear:
"If you want keywords available on condition..." --> "If you want keywords available on both replacement and condition..."

Shnoulle: Done

PR #3945, opened 2024-08-23
Fixed issue #19711: Cannot access simple statistics

Shnoulle:
Dev: Remove MYSQL specific character


gabrieljenik (approved):

c-schmitz:
I stumbled on this accidentally and fixed it myself, yesterday. Your fix won't work on Postgres and MSSQL, btw, since type is a keyword.

PR #3948, opened 2024-08-28
New feature #19722: Unable to use ExpressionScript code when export by remote control

Shnoulle:
Dev: allow any FormattingOptions attributes
Dev: Specific one are set after (format, output, then no need to filter)


Shnoulle:
Since it's really between an issue and a feature, I think it can go to master.

gabrieljenik (approved):
Code structure looks good

PR #3949, opened 2024-08-28
Fixed issue #19617: No wau to get question text in get_question_properties

Shnoulle:
Dev: add question, help and script as single attribute
Dev: add questionl10ns attributes too


Shnoulle:
Move to switch is more rewrite

Shnoulle:
> Not sure if automatic tests should be ammended

You're right : if we have one it can be easy and quick.

gabrieljenik:
> > Not sure if automatic tests should be ammended
> 
> You're right : if we have one it can be easy and quick.

You cna look at this: tests/unit/helpers/remotecontrol/RemoteControlQuestionPropertiesTest.php

gabrieljenik (approved):

PR #3950, opened 2024-08-29
fixed issue #19534: Survey cannot be submitted if "soft force" options are used and "full index" is active

Shnoulle:
Dev: add a state in questionSeq2relevance for sofForced
Fixed issue #18443: Optional questions not working without "popup"
Fixed issue : Optional questions not working without javascript
Dev: #18332: Mandatory soft must be related to Question not Page


Shnoulle:
OK, at 1st validation of 2nd group : mandSoft are in POST for group 1 …

Shnoulle:
Ok, I found a way to get it for old template too 

Need to update fruity_twentythree theme for now

Shnoulle:
Capture d’écran du 2024-08-29 19-17-51

Shnoulle:
`   │ Facebook\WebDriver\Exception\NoSuchElementException: Unable to locate element: #ls\2D button\2D submit`

Sh**t ! it work with vanilla (fixed) and copy of vanilla (all files, not fixed) Ans same with fruity_twentythree

Shnoulle:
> ` │ Facebook\WebDriver\Exception\NoSuchElementException: Unable to locate element: #ls\2D button\2D submit`
> 
> Sh**t ! it work with vanilla (fixed) and copy of vanilla (all files, not fixed) Ans same with fruity_twentythree

Sily boy ! Send the twig file when you need it !

application/helpers/expressions/em_manager_helper.php
```diff
@@ -353,7 +353,8 @@ class LimeExpressionManager
      * 'hidden' => 0    // 1 if it should be always_hidden
      * 'gid' => "34"    // group id
      * 'mandatory' => 'N'   // 'Y' if mandatory, 'S' if soft mandatory
-     * 'eqn' => ""  // TODO ??
+     * 'mandSoftForced' => false // boolean True if already forced
```
gabrieljenik, line 356: Given the sensitibity of the change, i would take the opportunity to expand on the comments. What does this 'mandSoftForced' means or is used for?

gabrieljenik (approved):
Code Structure looks good. Not an expert in EM. I would ask for a second review given the subject

application/helpers/expressions/em_manager_helper.php
Shnoulle: Right ! :+1:

Shnoulle: > boolean value to keep Mandatroy soft question status. False if not seen, answered, not a soft mandatory or not checked one time. Check is done in self::_validateQuestion using $_POST['mandSoft']

application/helpers/expressions/em_manager_helper.php
```diff
@@ -353,7 +353,8 @@ class LimeExpressionManager
      * 'hidden' => 0    // 1 if it should be always_hidden
      * 'gid' => "34"    // group id
      * 'mandatory' => 'N'   // 'Y' if mandatory, 'S' if soft mandatory
-     * 'eqn' => ""  // TODO ??
+     * 'mandSoftForced' => false // boolean value to keep Mandatroy soft question status. False if not seen, answered, not a soft mandatory or not checked one time. Check is done in self::_validateQuestion using $_POST['mandSoft']
```
gabrieljenik, line 356: Thanks! This is more clear now. I would just add something like "If true, validation will fail and user will be asked to confirm the question is not to be answered. After being answered, will be set to false."

application/helpers/expressions/em_manager_helper.php
```diff
@@ -6274,16 +6285,34 @@ function ($sqRankAnwsers) {
                     break;
             }
         }
-        /* Set qmandViolation to false if mandSoft and POST is set */
```
gabrieljenik, line 6277: I would add something like
```
// Mandatory Soft Validation.
// - All questions are initilized with 'mandSoftForced' = false. (previously).
// Now, if there is a mandatory validation violation (question not answered and required or soft required) and mandatory soft feature is on, they will be set to true and the mandatory violation will be cleared. 
// If no mandatory violation, means answered or no mandatory feature is on, so they will be set to false again.
```

application/helpers/expressions/em_manager_helper.php
Shnoulle: Seems to be more for the manual ?  https://manual.limesurvey.org/Question_type_-_Array/en#Mandatory

This one need to be updated for "No popup" action.

application/helpers/expressions/em_manager_helper.php
Shnoulle: More for manual ?

application/helpers/expressions/em_manager_helper.php
Shnoulle: > "Controls when to show a mandatory soft alert to user. 

You can show it in Survey theme where you want, the checkbox is shown currently when try to submit a mandatory question.

> If true, validation will fail and user will be asked to confirm the question is not to be answered.

No : if true user already check the checkbox (or confirm on dialog box)

gabrieljenik: > Seems to be more for the manual ?

For me is documentation for the coder, that's why suggested it there.
But it just a suggestion.
Already approved the code review.

Shnoulle: >  Already approved the code review.

Yes, i know, but if comment are more clear : it's great :)

PR #3951, opened 2024-09-03
Fixed issue #19729 [security] : XSS on time_limit

Shnoulle:
Dev : cast as integer
Dev: move to twig and escape


gabrieljenik (approved):
Code structure looks OK

PR #3953, opened 2024-09-05
Fixed issue #19700: Invalid DB value can be entered with start URL

Shnoulle:
Fixed issue : invalid DB value can be entered in multiple choice question type
Dev: use checkValidityAnswer for startingValues
Dev: fix logic of 'M' question type


gabrieljenik (approved):
Code structure looks OK

PR #3954, opened 2024-09-05
Fixed issue #19733: XSS by simple user with end url

Shnoulle:
Dev: minimal url encoding : only "


Shnoulle:
It's really the minimal fix we can do I think.

c-schmitz (approved):

PR #3959, opened 2024-09-11
Dev #19691: php8.3 compatibility - pchart

gabrieljenik:


Shnoulle:
Where pchart is used ?

If we can : i think it's better to move to pchart2

Shnoulle (approved):

gabrieljenik:
Will be replaced by a master one

PR #3960, opened 2024-09-11
Dev #19685: PHP 8.3 compatibility - LS Core 

gabrieljenik:


application/core/plugins/TwoFactorAdminLogin/helper/phpqrcode.php
```diff
@@ -157,7 +157,9 @@ public static function set(&$srctab, $x, $y, $repl, $replLen = false) {
  */
 
     class QRtools {
-    
+
+        public $frames;
```
Shnoulle, line 161: Need to separate this one ?

See https://bugs.limesurvey.org/view.php?id=19686

Shnoulle (approved):
The code on core seems OK.
More mantis issue updated than just core issue

application/core/plugins/TwoFactorAdminLogin/helper/phpqrcode.php
gabrieljenik: Yep, should have been on that ticket. But let's leave it as it is

gabrieljenik:
It was decided that this needs to be merged into master not develop.
So closing the PR in favor of https://github.com/LimeSurvey/LimeSurvey/pull/3962

PR #3962, opened 2024-09-13
Dev #19685: PHP 8.3 compatibility - LS Core

gabrieljenik:


application/libraries/admin/pdf.php
```diff
@@ -843,7 +843,7 @@ public function write_out($name)
 
     public function delete_html($text)
     {
-        $text = html_entity_decode((string) $text, null, 'UTF-8');
+        $text = html_entity_decode((string) $text, ENT_QUOTES|ENT_SUBSTITUTE, 'UTF-8');
```
Shnoulle, line 846: Moving from `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401` to `ENT_QUOTES | ENT_SUBSTITUTE` no ?

Seems it's OK, but it update something.

application/models/TemplateConfiguration.php
```diff
@@ -1468,7 +1468,17 @@ public function __get($name)
             if ($sAttribute === 'inherit') {
                 // NOTE: this is object recursive (if parent configuration field is set to inherit,
                 // then it will lead to this method again.)
-                $sAttribute = $this->getParentConfiguration()->$name;
+                $oParentConfiguration = $this->getParentConfiguration();
+                /**
+                 * We check if $oParentConfiguration is the same as $this because if it is, $oParentConfiguration->$name will
+                 * try to directly access the property instead of calling the magic method, and it will fail for dynamic properties.
+                 * @todo: Review the behavior of getParentConfiguration(). Returning the same object seems to be a bug.
```
Shnoulle, line 1475: :+1:

application/helpers/userstatistics_helper.php
```diff
@@ -2621,7 +2621,7 @@ public function generate_statistics($surveyid, $allfields, $q2show = 'all', $use
             }
 
             // Creating the first worksheet
-            $this->sheet = $this->workbook->addWorksheet(utf8_decode('results-survey' . $surveyid));
+            $this->sheet = $this->workbook->addWorksheet(mb_convert_encoding('results-survey' . $surveyid, 'ISO-8859-1'));
```
Shnoulle, line 2624: https://www.php.net/manual/en/function.mb-convert-encoding.php 

Then must be `mb_convert_encoding('results-survey' . $surveyid, 'ISO-8859-1', 'UTF-8')` no ?

application/helpers/common_helper.php
```diff
@@ -260,7 +260,7 @@ function convertGETtoPOST($url)
     $stack = explode('?', $url);
     $calledscript = array_shift($stack);
     $query = array_shift($stack);
-    $aqueryitems = explode('&', $query);
+    $aqueryitems = explode('&', (string) $query);
```
Shnoulle, line 263: I know it's more a rewrite, but i thni it's better to check `?` here https://github.com/LimeSurvey/LimeSurvey/blob/f7ab872d576308c2cb65d213f9d01633b70f0a3d/application/helpers/common_helper.php#L259

And return directly the url with no POST value.

application/helpers/admin/statistics_helper.php
```diff
@@ -3856,7 +3856,7 @@ public function generate_statistics($surveyid, $allfields, $q2show = 'all', $use
             }
 
             // Creating the first worksheet
-            $this->sheet = $this->workbook->addWorksheet(utf8_decode('results-survey' . $surveyid));
+            $this->sheet = $this->workbook->addWorksheet(mb_convert_encoding('results-survey' . $surveyid, 'ISO-8859-1'));
```
Shnoulle, line 3859: https://www.php.net/manual/en/function.mb-convert-encoding.php 

Then must be `mb_convert_encoding('results-survey' . $surveyid, 'ISO-8859-1', 'UTF-8')` no ?

application/helpers/admin/statistics_helper.php
```diff
@@ -153,7 +153,7 @@ function createChart($iQuestionID, $iSurveyID, $type, $lbl, $gdata, $grawdata, $
 
             $counter = 0;
             foreach ($lblout as $sLabelName) {
-                $DataSet->SetSerieName(html_entity_decode($sLabelName, null, 'UTF-8'), "Serie" . $counter);
+                $DataSet->SetSerieName(html_entity_decode($sLabelName, ENT_QUOTES|ENT_SUBSTITUTE, 'UTF-8'), "Serie" . $counter);
```
Shnoulle, line 156: Moving from `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401` to `ENT_QUOTES | ENT_SUBSTITUTE` no ?

Seems it's OK, but it update something.

Shnoulle (commented):
Some comments, waiting for non draft version

application/helpers/common_helper.php
gabrieljenik: Not sure what you mean. Can you expand a bit more?
Still, I would try to stick with just the needed updates. Other enhancements could be done on different tickets

Shnoulle: Something like this

```
function convertGETtoPOST($url)
{
    $url = preg_replace('/&amp;/i', '&', (string) $url);
    if (strpos('?', $url) === false {
        return "window.LS.sendPost(\"" . $url ."\");";
    }
    $stack = explode('?', $url);
```

gabrieljenik: I would do afer hacing an automated test for it. Not now. For me, it doesn't worth the risk of updating it now.

application/helpers/admin/statistics_helper.php
```diff
@@ -153,7 +153,7 @@ function createChart($iQuestionID, $iSurveyID, $type, $lbl, $gdata, $grawdata, $
 
             $counter = 0;
             foreach ($lblout as $sLabelName) {
-                $DataSet->SetSerieName(html_entity_decode($sLabelName, null, 'UTF-8'), "Serie" . $counter);
+                $DataSet->SetSerieName(html_entity_decode($sLabelName, ENT_COMPAT, 'UTF-8'), "Serie" . $counter);
```
Shnoulle, line 156: null mean `ENT_COMPAT` ?

Why not 0 ?

Shnoulle (commented):
I think null can be replaced by 0 : no update.

application/helpers/admin/statistics_helper.php
gabrieljenik: Not sure what null means here... I incline to 0, and 0 = ENT_HTML401.
Still, after going deeper, I feel we should use `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401` which is the default for the function and will fit probably the need on this line. Thoughts?

Shnoulle: > Not sure what null means here... I incline to 0, and 0 = ENT_HTML401.

Then if we need no update : `ENT_HTML401`

>  Still, after going deeper, I feel we should use ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 which is the default for the function and will fit probably the need on this line. Thoughts?

It was my first idea. But null : mean default or 0 ?

Maybe test quickly on 7.4 for the result ?

gabrieljenik: > It was my first idea. But null : mean default or 0 ?

Don't know for sure... to me, it is 0, which matches ENT_COMPAT and ENT_HTML401 as well.

> Maybe test quickly on 7.4 for the result ?

We have been doing some tests about it and feel ` ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`

Shnoulle (approved):
Still draft ?

Maybe update to last master ?
For code check

PR #3970, opened 2024-09-20
Dev: #19496: plugins in plugin directory shown in git diff

Shnoulle:
Dev: #19753: Theme in theme directory shown in git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,35 @@
 !/tmp/runtime/index.html
 !/tmp/upload/index.html
 
+# ignore plugins directory by default except LS plugin
+# if you want to add a new distributed plugin : use /application/core/plugins
+/plugins/*
+!/plugins/index.html
+!/plugins/Demo/
+
+# ignore survey theme directory by default except LS survey theme
+# if you want to add a new distributed survey theme : add the directory here
+/themes/survey/*
+!/themes/survey/index.html
+!/themes/survey/bootswatch
+!/themes/survey/fruity
+!/themes/survey/fruity_twentythree
+!/themes/survey/vanilla
+
+# ignore question theme directory by default except LS question theme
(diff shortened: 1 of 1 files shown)


Shnoulle:
I have situation where plugins upload was disable, and sometimes where theme upload too (disable write by www-data in directory)
Then adding it in `.gitignore` in 3.X : we have plugins dir only. But theme is great too.

gabrieljenik (approved):
Code structure looks good

gabrieljenik:
Maybe smeone from Core team could also  review it as they may have other point of view?

Shnoulle:
Yes, i put decision needed and not need code review :)

Just tired to have bunch of files on my dev instance ;)

ptelu:
the /themes folder is only used for for core extensions (git repo) or /upload/themes for user based question and survey themes
the /plugins folder should never be used, it is an old relic, either use /application/core/plugins for core or /upload/plugins for user based plugins

so i would only add the plugins part to the gitignore
/plugins folder seems to already have correct gitignore definitions

Shnoulle:
> the /themes folder 

It's really great when you manage your own version to have a fixed theme not updatable via GUI inside theme/surveys/ folder.
I always have one or two theme inside this directory to manage the "base' theme.
A lot of other instances I know already use it for "base" theme.
And you have it on https://github.com/LimeSurvey/LimeSurvey/blob/5560c5f40a2c3b56153a0c3b44d5731f7611e463/application/config/config-defaults.php#L715 (i have instance where i update it to disallow vanilla or fuity, allow only one fixed theme).

> /upload/themes for user based question and survey themes

There are instances where we disable this : uploading question theme or create survey theme are disabled.

c-schmitz (dismissed):

ptelu (approved):

PR #3971, opened 2024-09-24
Fixed issue #19763: Unable to answer multiple choice question with other

Shnoulle:
Dev: my error here : https://github.com/LimeSurvey/LimeSurvey/commit/89a3800356a2e94cfb459f0a595cb382f4adb3e8

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -4664,14 +4664,11 @@ class LimeExpressionManager
                         $value = null;  // can't upload a file via GET
                         break;
                 }
-                /* Validate validity of startingValues : do not show error */
-                if (self::checkValidityAnswer($knownVar['type'], $value, $knownVar['sgqa'], $LEM->questionSeq2relevance[$knownVar['qseq']], false)) {
-                    $_SESSION[$LEM->sessid][$knownVar['sgqa']] = $value;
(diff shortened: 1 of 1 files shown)


Oliver3012:
Hey @Shnoulle, thanks for taking care of the bug! What do you think: how fast will the fix be live for our surveys? 

We have to make a decision now: stopping the survey, deactivating the "Other"-option and start the survey again (which includes losing all cached answers) OR waiting for your fix.

itsmevictor:
Thank you @Shnoulle for the quick fix! Obviously, the sooner the fix is pushed the better, but thank you already for reacting so rapidly. 

@Oliver3012 Depending on your survey, one option is to temporarily hide the multiple choice question by selecting "Yes" to "Always hide the question" in the question's settings. It's not optimal but at least you don't have to deactivate the survey at the risk of losing cached answers etc.

itsmevictor:
Unless I am mistaken, the problem seems to be resolved, although I don't really understand how that can be since no commits were pushed since this morning.

Oliver3012:
Our account has apparently been rolled back to 6.6.3:
image

itsmevictor:
I have no idea where to find the version we're running on but that was probably it!

ptelu:
The affected commit will be reverted until further testing has been done, and a new version will be released shortly.

PR #3979, opened 2024-09-26
Feature/19773 make reordering tab title translatable

aisreh:
n General survey settings > Overview questions & groups > Reorder (Tab)
The title (Reordering) in the blue alert info is not translatable


Shnoulle:
1. It's an issue 
2. Can you fix commit history ?

PR #3981, opened 2024-09-27
Fix #19290: Wrong Row Total in Array (Texts)

gabrieljenik:
- Fix minified script


Shnoulle (approved):
Yes clearly :)

Shnoulle:
Only one file updated in https://github.com/LimeSurvey/LimeSurvey/pull/3656/files

PR #3982, opened 2024-09-27
Fixed issue #19746: Auto correction of invalid input in Multiple numeric input type question could be improved

gabrieljenik:


assets/packages/expressions/em_javascript.js
```diff
@@ -139,9 +139,6 @@ function fixnum_checkconditions(value, name, type, evt_type, intonly)
         }
 
         var addition = "";
-        if(cleansedValue && cleansedValue.split("").pop().match(/(,)|(\.)/)){
```
gabrieljenik, line 142: Not sure why this was here on the first place.

Shnoulle: :+1:

Shnoulle:
Introducted here : https://github.com/SondagesPro/LimeSurvey-SondagesPro/commit/49aff574e363cf659e79101b16f59d839427755f

When we don't use `LEMradix === ','` to do action with , only …

My opinion : it's an issue …

Shnoulle:
> > My opinion : it's an issue …
> 
> Can you expando more on the issue?


If you set your separator to `. ` LimeSurvey update `0,` or `0.` : no reason.
And  separator to `.,` LimeSurvey update `0.` or `0,` : no reason.

But more : it does on ALL question type. Not rstrictyed to number

I already report this issue : where, in a text question we have to check a code with `.` inisde …

Shnoulle (approved):

PR #3983, opened 2024-09-30
Fixed issue #19779 : Question Type not translated

loiczam:
Fixed issue #19779


Shnoulle (approved):
Not perfect but exist (i like to have a translation from Question theme too)

gabrieljenik (approved):
Code Structure looks good

PR #3975, opened 2024-09-25
Dev: add test for Rempote control Xmlrpc

Shnoulle:
Dev: fix test for direct call

PR #3984, opened 2024-10-04
Fixed issue #19643: Date question with restrictions does not show up like in versions 3 and 5

gabrieljenik:


Shnoulle (approved):

Shnoulle:
1. Way to move to vendor ?
2. Add a test https://github.com/LimeSurvey/LimeSurvey/pull/3975 (needed IMHO)

PR #3991, opened 2024-10-10
New feature #19758: New plugin event beforeMovePublicFileUploaded

Shnoulle:
Dev: add plugin with :
Dev: way to disable upload
Dev: way to replace file content and filename

### DIFF
--- a/application/controllers/UploaderController.php
+++ b/application/controllers/UploaderController.php
@@ -169,8 +169,7 @@ class UploaderController extends SurveyController
                 Yii::app()->end();
             }
             /* Upload error due file size */
-            /* and check too $aAttributes['max_filesize'] */
-            if ($size > $maxfilesize || $_FILES['uploadfile']['error'] == 1 || $_FILES['uploadfile']['error'] == 2) {
+            if ($_FILES['uploadfile']['error'] == 1 || $_FILES['uploadfile']['error'] == 2) {
                 $return = array(
                     "success" => false,
                     "msg" => sprintf(gT("Sorry, this file is too large. Only files upto %s KB are allowed."), $maxfilesize)
@@ -180,17 +179,83 @@ class UploaderController extends SurveyController
                 Yii::app()->end();
             }
 
+            /* Some information for plugin */
             $valid_extensions_array = explode(",", (string) $aAttributes['allowed_filetypes']);
             $valid_extensions_array = array_map('trim', $valid_extensions_array);
             $pathinfo = pathinfo((string) $_FILES['uploadfile']['name']);
             $ext = strtolower($pathinfo['extension']);
-            $cleanExt = CHtml::encode($ext);
-            $randfilename = 'futmp_' . randomChars(15) . '_' . $pathinfo['extension'];
+            $randfilename = 'futmp_' . randomChars(15) . '_' . $ext;
             $randfileloc = $sTempUploadDir . $randfilename;
+            // event to allow updating some part
+            $event = new PluginEvent('beforeProcessFileUpload');
+            /* Current state */
+            $event->set('surveyId', $surveyid);
+            $event->set('responseId', Yii::app()->session['survey_' . $surveyid]['srid']); // NULL if not exist
+            $event->set('qid', $oQuestion->qid);
+            $event->set('preview', $preview);
+            $event->set('fieldname', $sFieldName);
+            $event->set('maxfilesize', $maxfilesize);
+            $event->set('valid_extensions_array', $valid_extensions_array);
+            $event->set('success', true);
+
+            /* The file */
+            $event->set('filename', $filename);
+            $event->set('size', $size);
+            $event->set('tmp_name', $_FILES['uploadfile']['tmp_name']);
+            $event->set('ext', $ext);
+            $event->set('randfilename', $randfilename);
+            $event->set('randfileloc', $randfileloc);
+
+            App()->getPluginManager()->dispatchEvent($event);
+            /* New state */
+            $success = $event->get('success', true);
+            $disableCheck = $event->get('disableCheck', false);
+            $moveFile = $event->get('movefile', true);
+            /* New file */
+            $filename = $event->get('filename');
+            $ext = $event->get('ext');
+            $size = $event->get('size');
+            $uploadfile_tmp_name = $event->get('tmp_name');
+            $randfilename = $event->get('randfilename');
+            $randfileloc = $event->get('randfileloc');
+
+            /* @var string used to show error on extension */
+            $cleanExt = CHtml::encode($ext);
+
+            if ($uploadfile_tmp_name !== $_FILES['uploadfile']['tmp_name']) {
+                @unlink($_FILES['uploadfile']['tmp_name']);
+            }
+            if (!$success) {
+                $return = array(
+                    "success" => false,
+                    "msg" => $event->get('msg', gT("An unknown error occurred"))
+                );
+                echo ls_json_encode($return);
+                Yii::app()->end();
+            }
+            if (!is_file($uploadfile_tmp_name)) {
+                $return = array(
+                    "success" => false,
+                    "msg" => $event->get('msg', gT("An unknown error occurred"))
+                );
+                echo ls_json_encode($return);
+                Yii::app()->end();
+            }
+
+            /* Check size against $aAttributes['max_filesize'] */
+            if (!$disableCheck && $size > $maxfilesize) {
+                $return = array(
+                    "success" => false,
+                    "msg" => sprintf(gT("Sorry, this file is too large. Only files upto %s KB are allowed."), $maxfilesize)
+                );
+                //header('Content-Type: application/json');
+                echo ls_json_encode($return);
+                Yii::app()->end();
+            }
 
             // check to see that this file type is allowed
             // it is also  checked at the client side, but jst double checking
-            if (!in_array($ext, $valid_extensions_array)) {
+            if (!$disableCheck && !in_array($ext, $valid_extensions_array)) {
                 $return = array(
                     "success" => false,
                     "msg" => sprintf(gT("Sorry, this file extension (%s) is not allowed!"), $cleanExt)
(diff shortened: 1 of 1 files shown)


application/controllers/UploaderController.php
```diff
@@ -237,11 +237,43 @@ public function run($actionID)
...
+            $event->set('tmp_name', $_FILES['uploadfile']['tmp_name']);
+            App()->getPluginManager()->dispatchEvent($event);
+            $success = $event->get('success', true);
+            $filename = $event->get('filename');
+            $uploadfile_tmp_name = $event->get('tmp_name');
+            if ($uploadfile_tmp_name !== $_FILES['uploadfile']['tmp_name']) {
+                @unlink($_FILES['uploadfile']['tmp_name']);
+            }
+            if (!$success) {
+                @unlink($uploadfile_tmp_name);
```
Shnoulle, line 257: I KNOW it allow deletion of any file : did we need to control ?

application/controllers/UploaderController.php
```diff
@@ -237,11 +237,43 @@ public function run($actionID)
...
+                $return = array(
+                    "success" => false,
+                    "msg" => $event->get('msg', gT("An unknown error occurred"))
+                );
+                echo ls_json_encode($return);
+                Yii::app()->end();
+            }
+            if (!is_file($uploadfile_tmp_name)) {
```
Shnoulle, line 265: Plugin error ? Leave it ?

application/controllers/UploaderController.php
```diff
@@ -237,11 +237,43 @@ public function run($actionID)
                 echo ls_json_encode($return);
                 Yii::app()->end();
             }
-
+            // event to allow updating some part
+            $event = new PluginEvent('beforeMovePublicFileUploaded');
+            $event->set('surveyId', $surveyid);
```
gabrieljenik, line 242: Maybe we can include $preview as a parameter as well?

application/controllers/UploaderController.php
```diff
@@ -237,11 +237,43 @@ public function run($actionID)
...
+            $event->set('responseId', Yii::app()->session['survey_' . $surveyid]['srid']); // NULL if not exist
+            $event->set('qid', $oQuestion->qid);
+            $event->set('success', true);
+            $event->set('filename', $filename);
+            $event->set('size', $size);
+            $event->set('tmp_name', $_FILES['uploadfile']['tmp_name']);
+            App()->getPluginManager()->dispatchEvent($event);
+            $success = $event->get('success', true);
```
gabrieljenik, line 250: Message in case of success is not later set in the output

application/controllers/UploaderController.php
gabrieljenik: Also  $realMimeType ?

gabrieljenik:
A small unit test could be added for testing how the app code reacts to different outputs of the plugin event.

gabrieljenik:
Idea: Maybe we could add a beforeFetchFile event and then open up the possibility to store files on S3, google drive, ...

application/controllers/UploaderController.php
Shnoulle: There are no messages in case of success.

I don't understand

application/controllers/UploaderController.php
Shnoulle: Not updatable for both : right ?

realMimeType can be found by plugin :)

Shnoulle:
> Idea: Maybe we could add a beforeFetchFile event and then open up the possibility to store files on S3, google drive, ...

Yes ! Can be great !

But need $randfileloc in plugin too ?

application/controllers/UploaderController.php
gabrieljenik: https://github.com/LimeSurvey/LimeSurvey/pull/3991/files/f97eba89064b5abc4ab5fcc7e05cced6c4cebdfb#diff-da3ae6cdc4a22c11c24a0ed743eb4bc1a190312eda624342eff57cc1e27883eeR313

Here a success message is shown, but the message can´t be set from the plugin.
That was what I was referring.

Shnoulle:
> > Idea: Maybe we could add a beforeFetchFile event and then open up the possibility to store files on S3, google drive, ...
> 
> Yes ! Can be great !
> 
> But need $randfileloc in plugin too ?

Or something to directy return succes/done , but then : need event for 

- UploaderController/filegetcontents
- UploaderController/delete
-  Response->deleteFile
- etc …

application/controllers/UploaderController.php
gabrieljenik: In case of error, my thought:
- I believe the unlinking of $uploadfile_tmp_name should happen inside the plugin.
- Outside the plugin, we delete $_FILES['uploadfile']['tmp_name'] if still there.
What do you think?

application/controllers/UploaderController.php
gabrieljenik: Never hurts!

application/controllers/UploaderController.php
gabrieljenik: > Not updatable for both : right ?

Agree

> realMimeType can be found by plugin :)

Size and filename as well, but we are passing them as they are available.

The following could be interesting and are not being passed either:
- $aFieldName
- $aAttributes
- $maxfilesize
- $valid_extensions_array

gabrieljenik: Maybe we could add as well an event beforeValidateFileUpload ?
That way we can perform custom validations or customize output messages.

Or said in another way, is that possible with the current event position?

application/controllers/UploaderController.php
Shnoulle: Thank you !

application/controllers/UploaderController.php
Shnoulle: > Maybe we could add as well an event beforeValidateFileUpload ? That way we can perform custom validations or customize output messages.
> 
> Or said in another way, is that possible with the current event position?

Oh, i don't want to push 2 events : maybe move current one to start of process and add a system to don't do other test (with condition).

gabrieljenik: I thought about moving the event a bit up in the process, could work. If so, maybe change the eventName to beforeProcessFileUpload?

gabrieljenik: I would say the event should be a little bit up, before the size check.
https://github.com/LimeSurvey/LimeSurvey/blob/f97eba89064b5abc4ab5fcc7e05cced6c4cebdfb/application/controllers/UploaderController.php#L171

> But moving test with $size > $aAttributes['max_filesize'] adter the event.

Maybe we are allgined. :)

> $disableValidation = $event->get('disableValidation', false);

Then you start the roll on what validations are disbaled and which are not: Extension, size, MimeType, ...

Shnoulle: > I would say the event should be a little bit up, before the size check.
> 

No : because it muts be AFTER` $_FILES['uploadfile']['error']` check.

Then we need to test file size 2 times , the order

1. check `$_FILES['uploadfile']['error']` retuirn json if error
2. event
3. check `$aAttributes['max_filesize']`


:)

application/controllers/UploaderController.php
Shnoulle: > I believe the unlinking of $uploadfile_tmp_name should happen inside the plugin.

Yes, then deletion of invalid files are done in plugin : :+1:

Shnoulle:
> A small unit test could be added for testing how the app code reacts to different outputs of the plugin event.

Can you help on this ?

c-schmitz:
This looks like a great feature, but yeah, a unit test is needed first.

Shnoulle:
> This looks like a great feature, but yeah, a unit test is needed first.

I ask to @tiborpacalat on disccord if i don't make error before start work on it out of dev version : https://bugs.limesurvey.org/view.php?id=19758#c81079

application/controllers/UploaderController.php
```diff
@@ -250,7 +327,7 @@ public function run($actionID)
                                 "name"          => rawurlencode(basename((string) $filename)),
                                 "ext"           => $cleanExt,
                                 "filename"      => $randfilename,
-                                "msg"           => gT("The file has been successfully uploaded.")
+                                "msg"           => $event->get('msg', gT("The file has been successfully uploaded."))
```
gabrieljenik, line 330: I wouldn´t use the $event variable here. I would have a local variable to handle the message output from the event, consistent with all others output variables from the event.

Shnoulle: I use it like that because the second parameters is the default (if not set),

Need to do :

`$msg = $event->get('msg');`

`"msg"           => !empty($msg) ? $msg : gT("The file has been successfully uploaded.");`

But OK : let's go.

Shnoulle:
> > A small unit test could be added for testing how the app code reacts to different outputs of the plugin event.
> 
> Can you help on this ?

@gabrieljenik ?

gabrieljenik:
> > > A small unit test could be added for testing how the app code reacts to different outputs of the plugin event.
> > 
> > 
> > Can you help on this ?
> 
> @gabrieljenik ?

ok

gabrieljenik:
> This looks like a great feature, but yeah, a unit test is needed first.

I got thinking about this and...
- If unit test is for the plugin event, not sure how to design the test... Let's say I create a test plugin. After dispatching the event, what would be the expected result... actually, that depends on the plugin. So instead of testing the plugin event I would be testing the test plugin.
- Then unit test shall en about how the controller handle the results of a dispatch event. Them, the best is to wrap that logic into a function or service helper. Then design a test for that.

or not do a unit test :)

c-schmitz:
I see the problem. lets leave it likethat.

c-schmitz (approved):

Shnoulle:
> If unit test is for the plugin event, not sure how to design the test... Let's say I create a test plugin. After dispatching the event, what would be the expected result... actually, that depends on the plugin. So instead of testing the plugin event I would be testing the test plugin.

Right, we test core plugin, not event in reality.

Shnoulle:
https://www.limesurvey.org/manual/BeforeProcessFileUpload

PR #4000, opened 2024-10-21
Fixed issue #19800: Error when selecting single choice answer in secondary language with debug=2

gabrieljenik:


Shnoulle (approved):
Ok, but i think it's betterif we can fix it before. Set '' as default value for `SurveyLanguageSetting->surveyls_url`

Shnoulle:
> Ok, but i think it's betterif we can fix it before. Set '' as default value for `SurveyLanguageSetting->surveyls_url`

Simple part, in https://github.com/LimeSurvey/LimeSurvey/blob/e23c3a2052b3f71fc445d2790b2f6ad63e5df8b6/application/models/SurveyLanguageSetting.php#L52

```
    public $surveyls_language = "";
    public $surveyls_title = "";
    public $surveyls_description = "";
    public $surveyls_welcometext = "";
    public $surveyls_endtext = "";
    public $surveyls_policy_notice = "";
    public $surveyls_url = "";
    public $surveyls_urldescription = "";
```

Maybe some are less needed because it was set by DB, but if we wait string ;: it must be string

c-schmitz (changes requested):
I agree that setting these defaults would be more elegant.

c-schmitz (approved):

PR #4005, opened 2024-10-25
QE0-1153 Open editor

kevin-foster-uk:
- Fix 'open in new editor' URL generation (its not a referer URL)
- Support rendering multiple 'open in new editor' buttons on a single page (target via class instead of id)


twilligls (approved):

c-schmitz:
Obsolete PR

PR #4012, opened 2024-11-03
#19821 - Answer should deleted by delete() method

mbischof:
Fixed issue #19821:

### DIFF
--- a/application/models/Question.php
+++ b/application/models/Question.php
@@ -532,8 +532,9 @@ class Question extends LSActiveRecord
         $answers = Answer::model()->findAll($qidsCriteria);
         if (!empty($answers)) {
             foreach ($answers as $answer) {
+                $answerId = $answer->aid;
                 if ($answer->delete()) {
-                    AnswerL10n::model()->deleteAllByAttributes(['aid' => $answer->aid]);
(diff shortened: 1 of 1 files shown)


application/models/Question.php
```diff
@@ -529,16 +529,13 @@ public function deleteAllAnswers(array $exceptIds = [])
         $ids = array_merge([$this->qid], $this->allSubQuestionIds);
         $qidsCriteria = (new CDbCriteria())->addInCondition('qid', $ids);
         $qidsCriteria->addNotInCondition('aid', $exceptIds);
-        $answerIds = [];
         $answers = Answer::model()->findAll($qidsCriteria);
         if (!empty($answers)) {
             foreach ($answers as $answer) {
-                $answerIds[] = $answer->aid;
+                AnswerL10n::model()->deleteAllByAttributes(['aid' => $answer->aid]);
```
Shnoulle, line 535: Must delete AnswerL10n only after answer was deleted with success. 

If we have an Answer without AnswerL10n : this broke when try to see survey.

c-schmitz: I agree

Shnoulle:
> Must delete AnswerL10n only after answer was deleted with success.

c-schmitz (approved):

Shnoulle (approved):

PR #4014, opened 2024-11-05
QE-1218: Implemented improvements upon the logo

lajosarpad:
Dev:


application/views/admin/super/layout_main.php
```diff
@@ -30,7 +30,7 @@
 
 $containerClass = !Yii::app()->user->isGuest ? 'container-fluid' : 'container-fluid ps-0';
 echo '<!-- Full page, started in SurveyCommonAction::renderWrappedTemplate() -->
-<div class="' . $containerClass . '" id="in_survey_common_action">';
```
kevin-foster-uk, line 33: Are we sure this is not needed any more?

lajosarpad: As per our chat, I brought back the container class and applied CSS in order to avoid the side paddings in this specific case

kevin-foster-uk (approved):

PR #4015, opened 2024-11-05
CT-1044: Prevent negative Uses-Left with Allow-multiple-responses enabled

mohabmes:


c-schmitz (changes requested):
@mohabmes Can you please explain the exact scenario this fix is supposed to fix and how you fixed it?

c-schmitz:
@coderabbitai review

PR #4019, opened 2024-11-07
no border for alert-filled-info

Trischi80:
Dev: alert-filled-info should have no black border (see design)


lajosarpad (changes requested):
This is how it looks alike on ce

image

No top margins or paddings can be seen

lajosarpad (approved):
This works, but we will need to keep in mind that a margin-top is needed for the banner if we are to show it.

PR #4021, opened 2024-11-07
New feature: Plugin event 'beforeSurveyHttpHeaders' allows to modify http headers in surveys

c-schmitz:
This is a small new event to make it possible for plugins to modify the HTTP headers before sending them out.


application/helpers/common_helper.php
```diff
@@ -943,17 +942,26 @@ function returnGlobal($stringname, $bRestrictToString = false)
...
+        }
+        // plugins can modify the
+        $event = new PluginEvent('beforeSurveyHttpHeaders');
+        $event->set('headers', $headers);
+        App()->getPluginManager()->dispatchEvent($event);
+        $headers = $event->get('headers');
+        foreach ($headers as $header) {
+            header($header);
         }
-        header('P3P:CP="IDC DSP COR ADM DEVi TAIi PSA PSD IVAi IVDi CONi HIS OUR IND CNT"'); // this line lets IE7 run LimeSurvey in an iframe
```
gabrieljenik, line 952: What about these headers? Are they not going to be set anymore?
I noticed some of these are kept, but some are not.
I would just keep them. As to avoid un expected problems. If need to be removed, then, maybe can be done in separate ticket and focusing on that

c-schmitz: Support in browser for them has long been discountinued with only the IE supporting them to some extend, but IE is long dead now.

application/helpers/common_helper.php
```diff
@@ -943,17 +942,26 @@ function returnGlobal($stringname, $bRestrictToString = false)
...
-            header('X-Frame-Options: SAMEORIGIN');
+            $headers[] = 'X-Frame-Options: SAMEORIGIN';
+        }
+        // plugins can modify the
+        $event = new PluginEvent('beforeSurveyHttpHeaders');
+        $event->set('headers', $headers);
+        App()->getPluginManager()->dispatchEvent($event);
+        $headers = $event->get('headers');
+        foreach ($headers as $header) {
+            header($header);
```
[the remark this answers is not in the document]
c-schmitz: Thanks, fixed it.

gabrieljenik (approved):
Code structure looks good!

gabrieljenik:
I think the same plugin but for cookies could be great

application/helpers/common_helper.php
```diff
@@ -943,17 +942,28 @@ function returnGlobal($stringname, $bRestrictToString = false)
...
+                    ];
         if (Yii::app()->getConfig('x_frame_options', 'allow') == 'sameorigin') {
-            header('X-Frame-Options: SAMEORIGIN');
+            $headers[] = 'X-Frame-Options: SAMEORIGIN';
+        }
+        // plugins can modify the
+        $event = new PluginEvent('beforeSurveyHttpHeaders');
+        $event->set('headers', $headers);
+        App()->getPluginManager()->dispatchEvent($event);
+        $headers = $event->get('headers', []);
```
Shnoulle, line 961: An idea to allow easier replacement , unsure it's good but i think it can be great.

```
       $headers = [       
            'Cache-Control' => 'Cache-Control: no-store, no-cache, must-revalidate, max-age=0, post-check=0, pre-check=0',
            'Pragma' => 'Pragma: no-cache',
            'Content-Type' => 'Content-Type: text/html; charset=utf-8'
        ];
        if (Yii::app()->getConfig('x_frame_options', 'allow') == 'sameorigin') {
            $headers['X-Frame-Options'] = 'X-Frame-Options: SAMEORIGIN';
        }
```

Then plugins can replace specific core header (or gentle plugin header added too).

Shnoulle: Arg too late ;)

PR #4026, opened 2024-11-12
LA-148: Dashboard minor issues

mohabmes:


lajosarpad (approved):
The changes look good. Since we have three tickets addressed in this PR, I will write a summary for all of them:

1. LA-148: When I hover the survey records in the survey grid, the ... icon has the gray background as expected
2. LA-149: This ticket has been unclear to me, so I was unable to test this one, so if this PR does not solve that ticket, then that can be addressed by a new PR afterwards
3. LA-150: I can confirm that Not Started is among the filters and works as expected

PR #4027, opened 2024-11-13
New feature: API get_quota_properties also returns completeCount

TonisOrmisson:
New Feature/Dev: API get_quota_properties also returns completeCount

Its seems that getting completeCount for a quota is an essential bit of data. 
Also, it seems not reasonable to make a separate API endpoint just for getting the completeCount. I suggest adding the completeCount to the get_quota_properties to include the completeCount as one of the default returned parameter.

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -2659,7 +2659,7 @@ class remotecontrol_handle
      * Get quota attributes (RPC function)
      *
      * Get properties of a quota
-     * All internal properties of a quota are available.
+     * All internal properties and completeCount of a quota are available.
      * @see \Quota for the list of available properties
      *
      * Failure status : Invalid quota ID, Invalid session key, No permission, No valid Data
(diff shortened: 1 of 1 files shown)


Shnoulle (commented):
No code issue

TonisOrmisson:
@Shnoulle I understand your question and I was having the same hesitations. what I added was technically not quotas property if you think about it in terms columns in the db or ORM. But then again neither are the quota_ls fields that are added in the same place similarly. 

Its a bit wider discussion, but if we talk real world Entities, then `completeCount` is a real world property of a quota entity. There is a whole school of developers who clearly distinct Objects and Data Structures and the idea that ORM != Objects and Objects should represent real entities rather than just map what is in the database. In that view having quota->name as a property delivered via API call that is actually fetched via quota->ls->name is a similiar property as `completeCount`?

I would gladly live with having a separate api call but considering real life it seems only logical that completeCount is an essential property of a quota. Also, I kind of have trouble imagining a use case where one would want to get the quota data without seeing the completeCount.

c-schmitz (approved):

Shnoulle:
> @Shnoulle I understand your question and I was having the same hesitations. what I added was technically not quotas property if you think about it in terms columns in the db or ORM. But then again neither are the quota_ls fields that are added in the same place similarly.

Yep, right : https://github.com/LimeSurvey/LimeSurvey/blob/4c4a2dece124e0c20a3f013d422ca360e60f9d68/application/models/Quota.php#L34

And better : if we want to do like Object  language must be send as array : https://github.com/LimeSurvey/LimeSurvey/blob/4c4a2dece124e0c20a3f013d422ca360e60f9d68/application/models/Quota.php#L73

Shnoulle (approved):
The CI pipeline issue came from  https://github.com/LimeSurvey/LimeSurvey/commit/0a4979a3a403aa990e6c271a274fe50791a59de8

PR #4043, opened 2024-11-19
Fixed issue #19694: [security] Update jQuery UI

c-schmitz:
See https://bugs.limesurvey.org/view.php?id=19694


c-schmitz:
@tiborpacalat This would only affect KCFinder, so if you can give it a quick test, would be great.

PR #4046, opened 2024-11-20
New feature MW-1407: Add check for database size before download

marticos:
LimeSurvey includes a built-in feature to download the database as an SQL file. However, for instances with large databases, this process is slow and resource-intensive, often failing due to long execution times and high memory usage.

Solution: This PR introduces a pre-check mechanism to assess the database size before initiating the backup process.

If the database is within the size limit (default: 250 MB), the user can proceed with the download by pressing a button.
If the database exceeds the limit, the user is shown a message advising them to perform a manual backup using the appropriate database client for their instance.
Changes Made:

Added logic to calculate the database size beforehand.
Introduced a new view (dumpdb_view) to display appropriate messages to users.
Configured the size limit via application/config/config-defaults.php in 'maxDatabaseSizeForDump'. 
Testing Instructions:

Check out the branch and navigate to Configuration > Advanced > Backup Entire Database in the admin panel.
Expected Results:
For databases smaller than the configured size limit (default: 250 MB), a message and a green button is displayed, allowing the user to proceed with the download.
For larger databases, a message suggests performing a manual backup.
To simulate the message for large databases, reduce the size limit in config-defaults.php to a value smaller than your current database size.

### DIFF
--- a/application/controllers/admin/dumpdb.php
+++ b/application/controllers/admin/dumpdb.php
@@ -21,6 +21,12 @@
  */
 class Dumpdb extends SurveyCommonAction
 {
+
+    /**
+     * @var array Data used for rendering views
+     */
+    protected array $data = [];
+
     /**
      * Dumpdb constructor.
      * @param $controller
@@ -41,6 +47,8 @@ class Dumpdb extends SurveyCommonAction
             Yii::app()->setFlashMessage(gT('This function cannot be executed because demo mode is active.'), 'error');
             $this->getController()->redirect(Yii::app()->getController()->createUrl("/admin"));
         }
+
+        $this->data = $this->getData();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


c-schmitz (approved):

application/controllers/admin/dumpdb.php
```diff
@@ -69,4 +69,38 @@ private function outputHeaders(string $sFileName)
...
+        // Check if it's a POST request
+        if (!Yii::app()->request->isPostRequest) {
+            throw new CHttpException(405, gT("Invalid action"));
+        }
+
+        // Check if user has necessary permissions
+        if (!Permission::model()->hasGlobalPermission('superadmin', 'read')) {
+            throw new CHttpException(403, gT("You do not have permission to access this page."));
+        }
+
```
marticos, line 98: Need to add another check for the size of the db to prevent an admin to be able to download a db dump making a post request even if the page is requesting to schedule the database.

PR #4051, opened 2024-11-20
Fixed issue #19765: Unable to upload resources via drag'n'drop in KCFinder

gabrieljenik:


Shnoulle (commented):
Double Content-Disposition: in the header ?
No need to remove the previous one ?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

Shnoulle (approved):

PR #4056, opened 2024-11-25
Fixed issue #19661: Data inconsistency in responses after importing LSA files

gabrieljenik:


Shnoulle (approved):

Shnoulle:
If we can put a @\deprecated to insertRecords it can be great too. 
Don't use all of this broken function

Shnoulle (approved):
Better than before !

gabrieljenik:
The automatic tests for this ticket were failing not because we introduced an error but because the fix impacted the import process (and fixed it).

The failing test was trying to import a survey for reviewing stats.
The LSA used in the test has a response with a lot of empty nodes, which technically means that the response is an empty string, but it was being imported as null (not relevant)  missmatch!. Now they are imported as empty, not null. That's why the statistics don't match now.

Seems there was some manual editing on the file because in the other responses it has missing nodes (as if they hadn't been exported because they were null) and in the last one it has all the nodes empty, as if they had data and had been cleaned to simulate that they weren't answered.

We created a new branch based on master with an automatic test to show the issue:
https://github.com/LimeSurvey/LimeSurvey/actions/runs/12263965074/job/34216749065#step:11:146

That test is now green on this PR.

Shnoulle:
Then previous test are broken, right ?

Shnoulle:
> > Then previous test are broken, right ?
> 
> Which previous ones? Got lost.


>  Remote Control Export Statistics Array Questions (ls\tests\RemoteControlExportStatisticsArrayQuestions)
>  ✘ Survey with one question
>    │
>    │ The Count is incorrect for this option.
>    │ Failed asserting that two strings are identical.
>    │ --- Expected
>    │ +++ Actual
>    │ @@ @@
>    │ -'3'
>    │ +'0'
>    │
>    │ /home/runner/work/LimeSurvey/LimeSurvey/tests/unit/helpers/remotecontrol/ExportStatisticsArrayQuestionsTest.php:44
>    │ /home/runner/work/LimeSurvey/LimeSurvey/vendor/bin/phpunit:122
>    │

This one : https://github.com/LimeSurvey/LimeSurvey/blob/master/tests/unit/helpers/remotecontrol/ExportStatisticsArrayQuestionsTest.php ?

Shnoulle:
Issue with 
https://github.com/LimeSurvey/LimeSurvey/blob/f79736c79624884b8f0d77f912c407546f4834b9/tests/unit/helpers/remotecontrol/ExportStatisticsArrayQuestionsTest.php#L78
and 
https://github.com/LimeSurvey/LimeSurvey/blob/f79736c79624884b8f0d77f912c407546f4834b9/tests/unit/helpers/remotecontrol/ExportStatisticsArrayQuestionsTest.php#L83

?

Capture d’écran du 2024-12-12 18-12-02

gabrieljenik:
> **confirmed** Capture d’écran du 2024-12-12 18-15-16
> 
> @gabrieljenik you can fix the test ?

Fixed here: https://github.com/LimeSurvey/LimeSurvey/pull/4108

PR #4064, opened 2024-11-28
Fixed issue #19820: Question Type Bootstrap Dropdown causes survey to jump to middle when starting

gabrieljenik:


themes/question/bootstrap_dropdown/survey/questions/answer/list_dropdown/answer.twig
```diff
@@ -51,15 +51,19 @@
...
+            #}
             {% if width_entry != '' %}
-            selectPicker.on('loaded.bs.select', function () {
-                selectPicker.selectpicker('open');
-                var dropdown = selectPicker.parent();
-                var width = dropdown.find('.dropdown-menu.show').width();
-                dropdown.css('width', width);
-                selectPicker.selectpicker('toggle');
-                $('.bootstrap-select .btn.dropdown-toggle').blur();
+            selectPicker.on('loaded.bs.select.firstload', function () {
```
Shnoulle, line 61: We don't have issue if dropdown start hidden by condition and shown after ?

themes/question/bootstrap_dropdown/survey/questions/answer/list_dropdown/answer.twig
```diff
@@ -51,15 +51,19 @@
...
+                selectPicker.off('loaded.bs.select.firstload');
```
Shnoulle, line 62: Add a data done ? I dislike remove ALL action : this disable js workaround if needed

Shnoulle (commented):
Need to test with condition : start hidden shown after.

Else ; reaslly need the off ?

themes/question/bootstrap_dropdown/survey/questions/answer/list_dropdown/answer.twig
gabrieljenik: Well, not sure... We are using the same event as before (just a custom one), so don't expect issues.
Could be a good testing scenario @tiborpacalat

themes/question/bootstrap_dropdown/survey/questions/answer/list_dropdown/answer.twig
gabrieljenik: This is a custom event, created and used only for this workaround purpose. It will not deactivate anything else.
Adding a data-workaround-done attribute seems too much to me.

Shnoulle: OK great ! I don't know it was a custom event firstload didn't exist in bs.select ?

gabrieljenik: Nope. Created for this matter

Shnoulle (approved):

PR #4066, opened 2024-11-28
Feature/mw 1407 users cannot download db

marticos:
https://limesurvey.atlassian.net/browse/MW-1407

Add settings input to be able to set max size for the db dumps. Please test it on global settings.
Thank you.

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -78,6 +78,7 @@ $config['timeadjust']                = 0; // Number of hours to adjust between y
 $config['maxdumpdbrecords']          = 500; // The maximum number of records that would be ouput in one go during a database backup. Reduce this number if you're getting errors while backing up the entire database.
 $config['maxdbsizeforbackup']        = 0; // The maximum database size in MB that is backed up up by ComfortUpdate - 0 for no limit
 $config['allowexportalldb']          = 0; // Default 0 will only export prefixed tables when doing a database dump. If set to 1 ALL tables in the database will be exported (use carefully)
+$config['maxDatabaseSizeForDump']    = 256; // Maximum database size in megabytes to be able to download without errors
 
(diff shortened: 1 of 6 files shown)


application/config/config-defaults.php
```diff
@@ -78,6 +78,7 @@
 $config['maxdumpdbrecords']          = 500; // The maximum number of records that would be ouput in one go during a database backup. Reduce this number if you're getting errors while backing up the entire database.
 $config['maxdbsizeforbackup']        = 0; // The maximum database size in MB that is backed up up by ComfortUpdate - 0 for no limit
 $config['allowexportalldb']          = 0; // Default 0 will only export prefixed tables when doing a database dump. If set to 1 ALL tables in the database will be exported (use carefully)
+$config['maxDatabaseSizeForDump']    = 256; // Maximum database size in megabytes to be able to download without errors
```
Shnoulle, line 81: Why this value ? It can be totally different according to server. Here it's limesurvey hosting value, right ?

My opinion : no restriction by default, LimeSurvey hosting config can set it at 256

application/controllers/admin/dumpdb.php
```diff
@@ -67,6 +72,48 @@ private function outputHeaders(string $sFileName)
...
+            throw new CHttpException(405, gT("Invalid action"));
+        }
+
+        // Check if user has necessary permissions
+        if (!Permission::model()->hasGlobalPermission('superadmin', 'read')) {
+            throw new CHttpException(403, gT("You do not have permission to access this page."));
+        }
+
+        if ($this->data['downloadable'] === false) {
+            throw new CHttpException(403, gT("The database is too large to be downloaded. Please consider exporting it manually using your database client."));
```
Shnoulle, line 109: If not downloadable : why not save it locally (mybb backup for example save it locally and allow dowload via GUI after)

Shnoulle (commented):
1. Why 256 ?
2. Try alternate solution if not downloadable

PR #4077, opened 2024-12-04
Fixed issue #19838: Plugin upload doesn't allow LSS, LSA, LSG and LSQ files

gabrieljenik:


Shnoulle:
Unsure it need code review ;)

PR #4082, opened 2024-12-05
Fixed issue [security] #19864: Export question and group are not restricted

Shnoulle:
Dev: do not use survey ID in url but the sid from question and group


gabrieljenik (approved):
Code structure looks good?

PR #4085, opened 2024-12-06
Fixed issue [security] #19869: All group deletion accessible by any admin with one survey access

Shnoulle:


application/models/QuestionGroup.php
```diff
@@ -163,14 +163,16 @@ public function insertRecords($data)
...
      * Returns affected rows of question group table (should be 1 or null)
      * @param integer $groupId
-     * @param integer $surveyId
+     * @param integer|null $surveyId deprecated
      * @return int|null
      */
-    public static function deleteWithDependency($groupId, $surveyId)
+    public static function deleteWithDependency($groupId, $surveyId = null)
     {
+        $QuestionGroup ::self::model()->findByPk($groupId);
```
gabrieljenik, line 171: What if the QueestionGroup is not found? Shouldn't we have a small check after query?

Shnoulle: Throw 500 error maybe ?

application/models/QuestionGroup.php
```diff
@@ -163,14 +163,16 @@ public function insertRecords($data)
...
         // Abort if the survey is active
-        $surveyIsActive = Survey::model()->findByPk($surveyId)->active !== 'N';
+        $surveyIsActive = Survey::model()->findByPk($QuestionGroup->sid)->active !== 'N';
         if ($surveyIsActive) {
+            /* TODO : replace by a cleaner way to send error : model muts not use HTML only error system */
             Yii::app()->user->setFlash('error', gT("Can't delete question group when the survey is active"));
```
Shnoulle, line 176: @gabrieljenik do you think it's better to send a 500 error here ? Check of active survey must be done by controller in my opinion. Then if it(s not done : it' a 500

Shnoulle: It's done in LimeSurvey\Models\Services\QuestionGroupService->checkDeletePermission i throw a 500 here

application/models/QuestionGroup.php
Shnoulle: Thorw 500 (but must be a 404 : i report the issue)

Shnoulle:
[19880: PermissionDeniedException and NotFoundException send 500 error](https://bugs.limesurvey.org/view.php?id=19880)

application/models/QuestionGroup.php
```diff
@@ -163,16 +168,20 @@ public function insertRecords($data)
...
+        if (empty($QuestionGroup)) {
+            throw new NotFoundException(gt('Group not found'));
+        }
         // Abort if the survey is active
-        $surveyIsActive = Survey::model()->findByPk($surveyId)->active !== 'N';
+        $surveyIsActive = Survey::model()->findByPk($QuestionGroup->sid)->active !== 'N';
         if ($surveyIsActive) {
-            Yii::app()->user->setFlash('error', gT("Can't delete question group when the survey is active"));
-            return null;
+            throw new PermissionDeniedException(gt("Can't delete question group when the survey is active"));
```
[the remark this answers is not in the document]
gabrieljenik: Is this the right kind of exception to be shown? Maybe BadRequestException.php is better fit?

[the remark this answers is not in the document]
Shnoulle: 1. All LimeSurvey exception throw 500 error :)
2. GUI send a PermissionDeniedException https://github.com/LimeSurvey/LimeSurvey/blob/f10805aee259d9cfba72c35e3c5e6833a714519d/application/models/services/QuestionGroupService.php#L572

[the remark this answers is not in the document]
Shnoulle: But you're right

gabrieljenik (approved):
Code structure looks good.

PR #4087, opened 2024-12-09
Fixed issue [security] #19873: RC : Any user can create or update any participant in CPDB

Shnoulle:
Dev: check permission and continue if no permission

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -3935,7 +3935,7 @@ class remotecontrol_handle
 
             // Participant not found, so we create a new one
             if (!$model) {
-                if (Permission::model()->hasGlobalPermission('participantpanel', 'create')) {
+                if (!Permission::model()->hasGlobalPermission('participantpanel', 'create')) {
                     /* No permission to create : continue */
                     continue;
                 }


gabrieljenik (approved):
Code structure looks good. I am not sure if there is a way to inform about the operations not permitted. If not, people will not know about the error.

Shnoulle:
> Code structure looks good. I am not sure if there is a way to inform about the operations not permitted. If not, people will not know about the error.

There are a result : https://github.com/LimeSurvey/LimeSurvey/blob/47083aad164ecaa5fb85bd5cb0d92e890b034d0b/application/helpers/remotecontrol/remotecontrol_handle.php#L3909

We don't need to send value with issue : you send 10 lines : you get one update and one created : you know there are 8 with issues

PS : i start to send 403 … and understand it's a bad solution ;)

gabrieljenik:
> We don't need to send value with issue : you send 10 lines : you get one update and one created : you know there are 8 with issues

Maybe just a messages item in the reposnse array: "You don't have permission to ..."

Shnoulle:
> > We don't need to send value with issue : you send 10 lines : you get one update and one created : you know there are 8 with issues
> 
> Maybe just a messages item in the reposnse array: "You don't have permission to ..."

I really don't think we need to add it

gabrieljenik:
Maybe we should add some automatic testing for this?

Shnoulle:
> Maybe we should add some automatic testing for this?

On a second time : currently i think we need to fix security :)

I create a mantis for permission check on RC tomorrow.

Shnoulle:
> Maybe we should add some automatic testing for this?

We already have some but the current Remote Control Export Statistics Array Questions (ls\tests\RemoteControlExportStatisticsArrayQuestions) gide it.

There are an error in permission checking here.

PR #4088, opened 2024-12-09
Fixed issue [security] #19870: Any user wit survey create permission can activate tokens on any survey

Shnoulle:
Dev: use test from GUI


gabrieljenik (approved):
Code structure looks good.

PR #4094, opened 2024-12-10
New feature: New setting in Global/user to configure how user names are displayed in the administration.

c-schmitz:
In certain SSO integrations, the provided username can be a cryptic identifier, such as a UUID. To make it easier to identify which user owns a survey or similar item, this patch introduces a configurable global setting. Admins can now choose how user information is displayed from the following options:

- User name (default)
- Full Name
- Full Name (User name)

The new setting is available in Settings/Global/User


gabrieljenik:
Code structure looks good.
Still, why the setting name is "adminNameDisplayType"? I mean, why does it start with "admin"?

application/models/User.php
```diff
@@ -223,34 +223,63 @@ protected function afterDelete()
...
+                return $this->full_name;
+                break;
+            case 'both':
+                return sprintf(gt("%s (%s)"), $this->users_name, $this->full_name);
+                break;
+            case 'username':
+            default:
+                return $this->users_name;
+        }
+        return $this->full_name? $this->full_name : $this->users_name;
```
olleharstedt, line 254: Unreachable code?

application/models/User.php
```diff
@@ -223,34 +223,63 @@ protected function afterDelete()
...
+     *
+     * @return int The number of surveys created by this user.
      */
     public function getSurveysCreated()
     {
         $noofsurveys = Survey::model()->countByAttributes(array("owner_id" => $this->uid));
         return $noofsurveys;
     }
 
+    public function getDisplayName()
```
olleharstedt, line 241: Missing docblock.

application/models/User.php
```diff
@@ -910,7 +939,7 @@ public function getColums()
 
         $cols[] = array(
             "name" => "parentUserName",
-            "header" => gT("Created by"),
+            "header" => gT("Owner"),
```
Shnoulle, line 942: Owner of a user ?

c-schmitz: Good question: What happens if the creator of a user is deleted.

Shnoulle: I know, but owner of user : seems strange, near owner of people

PR #4101, opened 2024-12-11
Fixed issue #19856: Import LDAP - error 500 - ldap_set_option

Shnoulle:
Dev: see https://github.com/adldap/adLDAP/pull/142


gabrieljenik (approved):
Code structure looks ok

Shnoulle:
@tiborpacalat since new php-ldap version set this value under condition : i think we need it.

PR #4109, opened 2024-12-16
DEV: 19801: php8.3 compatibility - Zend (xmlrpc related) - using the ZF1 Future as vendor

gabrieljenik:
Couldn´t add just the lib. Had to update all.
It seems to me that in order for everything not to be updated, the composer.lock has to match what is in vendor. But composer.lock is not in the repo, so ...


Shnoulle (approved):
Great job here !

PR #4114, opened 2024-12-18
Fixed issue #19848: Array questions split on iPad 7 despite its large screen (Vanilla theme)

Ahmedazeem98:


twilligls (approved):
code is fine, but still needs some testing

PR #4117, opened 2024-12-19
QE-1295: Redirecting to editor if it's switch on upon copy or import

lajosarpad:
https://limesurvey.atlassian.net/browse/QE-1295

### DIFF
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -2359,6 +2359,14 @@ class SurveyAdministrationController extends LSBaseController
             }
         }
 
+        if ((App()->getConfig("editorEnabled")) && isset(($aImportResults['newsid']))) {
+            if (!isset($oSurvey)) {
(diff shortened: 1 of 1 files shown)


application/controllers/SurveyAdministrationController.php
```diff
@@ -2360,7 +2360,11 @@ public function actionCopy()
         }
 
         $this->aData = $aData;
-        $this->render('importSurvey_view', $this->aData);
+        if ((App()->getConfig("editorEnabled")) && isset(($aImportResults['newsid']))) {
```
twilligls, line 2363: we also need to check for the surveytheme being "fruity_twentythree" (soon we will have all themes able to open the new editor, but not yet)

lajosarpad: @twilligls applied the change, but unfortunately this will only work for fruitytwentythree, but will not work for themes inherited from it

twilligls: that's fine for now

twilligls (approved):

PR #4118, opened 2024-12-20
Task/mw 1415 progress bar displays 100 value before the end of the survey

marticos:
https://limesurvey.atlassian.net/browse/MW-1415


gabrieljenik (commented):
Like the idea of introducing the new nonRelevantSteps counter.
Code structure looks good. Haven't tested it.
Maybe out of the scope, maybe not, but... assuming this requires lot of testing, could be a good opportunity to introduce a StepCounter class?

That would also create the opportunity to add some php unit automated testing on this area...

tiborpacalat:
Gabriel, I will create an additional ticket for this, and ship this as it is.

application/core/LSETwigViewRenderer.php
```diff
@@ -706,10 +706,10 @@ private function getAdditionalInfos($aData, $oTemplate)
                 foreach ($oTemplate->oOptions as $key => $value) {
                     // TODO: Same issue as commit 2972aea41c51c74db95bfe40c337ae839471152c
                     // Options are not loaded the same way in all places.
-                    if ($value instanceof stdClass) {
```
35grain, line 709: This breaks the use of theme settings when force_xmlsettings_for_survey_rendering setting is enabled.

tiborpacalat: Thanks for reporting, I will contact the dev and get this resolved.

marticos: Hi @35grain,
Unfortunately, I wasn’t able to replicate the issue.
Could you please open a ticket on our bug tracker and include the steps to reproduce it?
https://bugs.limesurvey.org/
Thanks.

35grain: Sure thing: https://bugs.limesurvey.org/view.php?id=20106

PR #4122, opened 2025-01-01
Add LDAP Group list loader #19916

kvolk2git:
New feature #19916:
You can now add survey participants from the LDAP directory groups. To do this, it is enough to specify only the path to the directory containing the groups. All users who are members of the group will be automatically found and added to the list of members. The name of this list of participants for LimeSurvey will be taken from the "info" attribute of the group.

**in AD:**
image
**in LimeSurvey:**
image
image


kvolk2git:
Ok, thanks!

PR #4126, opened 2025-01-02
Fixed issue #19908: Import participants to survey: Can't import CSV without tokens

gabrieljenik:


application/models/Token.php
```diff
@@ -311,7 +311,7 @@ private function generateRandomToken($iTokenLength)
     public static function sanitizeToken($token)
     {
         // According to Yii doc : http://www.yiiframework.com/doc/api/1.1/CSecurityManager#generateRandomString-detail
-        return preg_replace('/[^0-9a-zA-Z_~]/', '', $token);
+        return preg_replace('/[^0-9a-zA-Z_~]/', '', (string) $token);
```
Shnoulle, line 314: I think I prefer 

```
if (empty($token)) {
return "";
}
```

more clear for me

And we muts allow null or string or empty array pbject but no not empty object or array. IMHO

Shnoulle (commented):
See comment

application/models/Token.php
gabrieljenik: > I think I prefer

Will do it

> And we muts allow null or string or empty array pbject but no not empty object or array. IMHO

Can you please clarify / expand? Not following.

I would still leave the casting, just in case.

Shnoulle: Aloow NULL or ""

But still send a 500 if get a [] or Std Class : it must be forbidden (it's an opinion)

> I would still leave the casting, just in case.

I remove it personally (still an opinion) :

Shnoulle (approved):

PR #4135, opened 2025-01-15
added new locale fr-ca

mbischof:
New feature #19891: new locale fr-ca


Shnoulle (approved):
No issue in code,
BUT : unsure on the decision about such situation ping @c-schmitz 
I don't think we want copy of existing language without real update and needed (else : FR_be, FR_ch for starting)

c-schmitz:
Yeah, this is a bit of a egg and hen problem. We will only approve a translation if at least the survey taking part is somewhat translated.

c-schmitz:
If the requirements are not met and nobody does provide translations, we can close this PR for now.

PR #4142, opened 2025-01-21
LA-129: Add 'Last modified' date in Survey list/boxes

mohabmes:


application/models/Survey.php
```diff
@@ -2578,4 +2588,19 @@ private static function shiftedDateTime($datetime)
...
+     * Update last modified date
+     * @param int $sid survey id
+     */
+    public function updateLastModifiedDate($sid)
+    {
+        App()->db->createCommand()->update(
+            '{{surveys}}',
+            array(
+                'lastModified' => date('Y-m-d H:i:s'),
+            ),
```
ptelu, line 2602: I think this should use also the dateshift/shiftedDateTime like the other time based survey columns

ptelu (changes requested):

application/models/Survey.php
ptelu: it seems the only column which does not use the shifted date is the creation date which seems weird, maybe have a quick chat about it with carsten.

mohabmes: The date/time is now always shifted when displayed

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,235 @@
+<?php
+
+namespace LimeSurvey\Helpers\Update;
+
+class Update_632 extends DatabaseUpdateBase
+{
+    public function up()
+    {
+        $this->db->createCommand()->addColumn('{{surveys}}', 'lastModified', 'datetime NULL');
```
lajosarpad: I believe we need to default this to the current moment, which will happen to have the value of the creation initially, or, if the survey already exists, then it will be set to the creation of the field.

MySQL Example: alter table foobar
add column something datetime default current_timestamp;

`datetime` does not exist in PostgreSQL, please use timestamp instead

image

In PostgreSQL you could use timestamp instead, which btw exists both for MySQL and SQL Server, so I would use timestamp instead of datetime to ease the problems. See: https://www.db-fiddle.com/f/fZrXEXvPMsLCeVgCMNGQyd/2

In SQL Server you cannot set a default for timestamp, so for SQL Server you'll need something like

```
create table foobar(a int);

alter table foobar
add something datetime default current_timestamp;

insert into foobar(a) values(2);

select * from foobar;
```

Therefore, the line above needs to be changed to

- use timestamp for PostgreSQL (and possibly for MySQL too)
- use datetime for SQL Server
- default to current_timestamp

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,235 @@
...
+        foreach ($this->getTriggers() as $trigger) {
```
lajosarpad: Good stuff

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,235 @@
...
+            $this->db->createCommand($trigger)->execute();
+        }
+    }
+
+    private function getTriggers(): array
+    {
+        return App()->db->driverName === 'pgsql'
```
lajosarpad: Support for SQL Server is missing. We need to support SQL Server too. Please look into application/helpers/admin/import_helper.php `getTableArchivesAndTimestamps` for reference.

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,235 @@
...
+            ? $this->getPostgresTriggers()
+            : $this->getMySqlTriggers();
+    }
+
+    private function getPostgresTriggers(): array
+    {
+        return [
```
lajosarpad: Even though you have no counterparts for the methods you call here for MySQL nor SQL Server at the moment, it still makes sense to either rename them to clarify they are for Postgres, or have a parameter passed to them specifying it's Postgres (in which case you would need to handle the trigger differences inside the method)

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,235 @@
...
+
+    private function triggerAnswers(): string
+    {
+        return <<<SQL
+CREATE OR REPLACE FUNCTION update_survey_last_modified_from_answers()
+RETURNS TRIGGER AS \$\$
+DECLARE
+    survey_id INT;
+BEGIN
+    SELECT sid INTO survey_id FROM lime_questions WHERE qid = NEW.qid;
```
lajosarpad: We cannot expect the lime_ prefix to be present. When you install CE, you can choose your prefix, or choose not to have a prefix. Therefore you need to either load the prefix, see

```
    $sourceTables = [
        Yii::app()->db->tablePrefix . "survey_" . $sid,
        Yii::app()->db->tablePrefix . "survey_" . $sid,
        Yii::app()->db->tablePrefix . "survey_" . $sid,
        Yii::app()->db->tablePrefix . "old_survey_{$sid}_{$sTimestamp}",
        Yii::app()->db->tablePrefix . "old_survey_{$sid}_{$sTimestamp}",
        Yii::app()->db->tablePrefix . "old_survey_{$sid}_{$sTimestamp}",
    ];
```

Inside getUnchangedColumns in the file of application/helpers/admin/import_helper.php, or use the `{{tablename}}` format if it works for your query. This is true for all usages of the prefix, I shall not repeat this remark at other places to avoid superfluous comments.

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,235 @@
...
+BEFORE UPDATE ON lime_surveys_languagesettings
+FOR EACH ROW
+EXECUTE FUNCTION update_survey_last_modified();
+SQL;
+    }
+
+    private function getMySqlTriggers(): array
+    {
+        return [
+            <<<SQL
```
lajosarpad: It is optional, but it would make sense to separate the triggers into different methods, so if later we want to add a new item or modify an existent item, it will be simple. But, again, this is not of great importance, so you can ignore this remark if you prefer.

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+            $this->triggerQuestions($dbType),
+            $this->triggerSurveys($dbType),
+            $this->triggerLanguageSettings($dbType),
+        ];
+    }
+
+
+    private function triggerAnswers($dbType): string
+    {
+        $prefix = App()->db->tablePrefix;
```
lajosarpad: Instead of repeating the initialization of $prefix in each method, you can do something like

```
protected $prefix = null;

protected function getPrefix()
{
    if ($this->prefix === null) {
        $this->prefix = App()->db->tablePrefix;
    }
    return $this->prefix;
}
```

so if we need to change the way we get the prefix, then we need to do it in a single place.

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+            return <<<SQL
+CREATE TRIGGER answers_last_modified
+ON [{$prefix}answers]
+AFTER UPDATE
+AS
+BEGIN
+    SET NOCOUNT ON;
+    DECLARE @survey_id INT;
+
+    SELECT TOP 1 @survey_id = q.sid
```
lajosarpad: This creates two commands: a select and an update, which are two separate requests to the database. Instead, an update-join would be better, as it would be a single request to the database. For SQL Server see https://www.datacamp.com/tutorial/sql-update-with-join

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+END;
+SQL;
+        } elseif ($dbType == 'pgsql') {
+            return <<<SQL
+CREATE OR REPLACE FUNCTION update_survey_last_modified_from_answers()
+RETURNS TRIGGER AS \$\$
+DECLARE
+    survey_id INT;
+BEGIN
+    SELECT sid INTO survey_id FROM {$prefix}questions WHERE qid = NEW.qid;
```
lajosarpad: Similarly, we don't need two requests to the RDBMS here. We can do an update-join. For PostgreSQL, please refer to https://neon.com/postgresql/postgresql-tutorial/postgresql-update-join

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+
+CREATE TRIGGER answers_last_modified
+BEFORE UPDATE ON {$prefix}answers
+FOR EACH ROW
+EXECUTE FUNCTION update_survey_last_modified_from_answers();
+SQL;
+        }
+
+        return <<<SQL
+CREATE TRIGGER `answers_last_modified` BEFORE UPDATE ON {$prefix}answers
```
lajosarpad: Again, it's better to do update-join for the same reason described earlier. In MySQL, please refer to https://www.geeksforgeeks.org/mysql/mysql-update-join/

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+            return <<<SQL
+            CREATE TRIGGER group_l10ns_last_modified
+ON [{$prefix}group_l10ns]
+AFTER UPDATE
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    UPDATE s
+    SET s.lastModified = GETDATE()
```
lajosarpad: Very good!

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+END;
+SQL;
+        } elseif ($dbType == 'pgsql') {
+            return  <<<SQL
+CREATE OR REPLACE FUNCTION update_survey_last_modified_from_group_l10ns()
+RETURNS TRIGGER AS \$\$
+DECLARE
+    survey_id INT;
+BEGIN
+    SELECT sid INTO survey_id FROM {$prefix}groups WHERE gid = NEW.gid;
```
lajosarpad: Let's change this to update-join.

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+FOR EACH ROW
+EXECUTE FUNCTION update_survey_last_modified_from_group_l10ns();
+SQL;
+        }
+
+        return <<<SQL
+CREATE TRIGGER `group_l10ns_last_modified` BEFORE UPDATE ON {$prefix}group_l10ns
+FOR EACH ROW BEGIN
+    DECLARE survey_id INT;
+    SELECT sid INTO survey_id FROM {$prefix}groups WHERE {$prefix}groups.gid = NEW.gid;
```
lajosarpad: Let's do an update-join instead.

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+
+SQL;
+        } elseif ($dbType == 'pgsql') {
+            return  <<<SQL
+CREATE OR REPLACE FUNCTION update_survey_last_modified_from_question_l10ns()
+RETURNS TRIGGER AS \$\$
+DECLARE
+    survey_id INT;
+BEGIN
+    SELECT sid INTO survey_id FROM {$prefix}questions WHERE qid = NEW.qid;
```
lajosarpad: Let's do update-join instead

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+FOR EACH ROW
+EXECUTE FUNCTION update_survey_last_modified_from_question_l10ns();
+SQL;
+        }
+
+        return  <<<SQL
+CREATE TRIGGER `question_l10ns_last_modified` BEFORE UPDATE ON {$prefix}question_l10ns
+FOR EACH ROW BEGIN
+    DECLARE survey_id INT;
+    SELECT sid INTO survey_id FROM {$prefix}questions WHERE {$prefix}questions.qid = NEW.qid;
```
lajosarpad: Let's do an update-join instead

application/helpers/update/updates/Update_632.php
```diff
@@ -0,0 +1,374 @@
...
+    {
+        $prefix = App()->db->tablePrefix;
+        if ($dbType == 'mssql' || $dbType == 'sqlsrv') {
+            return <<<SQL
+CREATE TRIGGER answers_last_modified
+ON [{$prefix}answers]
+AFTER UPDATE
+AS
+BEGIN
+    SET NOCOUNT ON;
```
lajosarpad: Is there a reason to set NOCOUNT ON?

lajosarpad (approved):

PR #4148, opened 2025-01-23
CT-1121: Creating timings table when activating back a survey

lajosarpad:
Dev:

### DIFF
--- a/application/models/SurveyActivator.php
+++ b/application/models/SurveyActivator.php
@@ -417,7 +417,8 @@ class SurveyActivator
      */
     protected function createTimingsTable()
     {
-        if ($this->survey->isSaveTimings) {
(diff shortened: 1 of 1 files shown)


application/models/SurveyActivator.php
```diff
@@ -417,7 +417,8 @@ protected function createResponseTableKeys()
      */
     protected function createTimingsTable()
     {
-        if ($this->survey->isSaveTimings) {
+        $prow = $this->survey->find('sid = :sid', array(':sid' => $this->survey->sid));
```
c-schmitz, line 420: please add a comment here why you had to implement this construct

c-schmitz (approved):

application/models/SurveyActivator.php
lajosarpad: Added a comment

PR #4165, opened 2025-02-11
Fixed issue #19909: Issues with font handling in the custom configuration of the survey theme - Fonts not loading

gabrieljenik:
- Fix for issue 1 and 3 described in ticket 19909: Fonts not loading

### DIFF
--- a/application/views/themeOptions/options_core.php
+++ b/application/views/themeOptions/options_core.php
@@ -194,7 +194,7 @@ $aOptionAttributes['optionAttributes']['brandlogofile']['dropdownoptions'] = $br
 
                                 <div class="col-12">
                                     <select class="<?= $classValue ?>" data-parent="<?= $attribute['parent'] ?>"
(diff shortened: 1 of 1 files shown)


Shnoulle:
Question : where package is loaded ? In theme or by core ?

Shnoulle (approved):

gabrieljenik:
> Question : where package is loaded ? In theme or by core ?

The PR fixes that the field "packages_to_load" is properly saved.

This is used by TemplateConfiguration::setThisTemplate() to define the dependencies:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/TemplateConfiguration.php#L1211-L1222

And then TemplateConfiguration::createTemplatePackage() uses the dependencies to define and load the theme package:
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/TemplateConfiguration.php#L1721-L1732

PR #4168, opened 2025-02-12
Fixed issue #19959: token will not allow standard uuid format tokens (removes dashes)

TonisOrmisson:
Fixed issue #19959:token will not allow standard uuid format tokens (removes dashes)


gabrieljenik (approved):
Code structure looks good.

Shnoulle (changes requested):
Is it possible to set it in config-default.php ?
Seems easy no ?

TonisOrmisson:
Im having trouble getting why would standard UUID format be a not wanted format for token. Seems like a no-brainer to me tbh. UUID is a widely accepted standard way of representing what a token value essentially is in LS. Why would one not want support that, any specific reason for not accepting that as a default allowed format?

Shnoulle:
> Im having trouble getting why would standard UUID format be a not wanted format for token. Seems like a no-brainer to me tbh. UUID is a widely accepted standard way of representing what a token value essentially is in LS. Why would one not want support that, any specific reason for not accepting that as a default allowed format?

Yes, as the **default** allowed format, but some user can want more extra characters, the adding the regex in config-default.php file allow this user to easily update it.

I don't see real difficulty here.

TonisOrmisson:
ok, I see what you mean, thanks for clarifying. Yes the PRs aim/scope was to fix the fact that a widely accepted token format is not accepted (which is an issue). Having the token fully configurable is a feature and I was not thinking in that context. I am having an actual issue with API-s not being able to submit UUID tokens which creates me real problems at this point and I would like to see that fixed sooner rather than later (via develop branch & new features)

Shnoulle:
I really think we can move it to a config-default with create a dev branch @tiborpacalat ? can you confirm ?

tiborpacalat:
Sure Denis!

application/config/config-defaults.php
```diff
@@ -839,5 +839,8 @@
 // Enable or disable single page application editor
 $config['editorEnabled'] = false;
 
+// Participants token sanitizing rules as regex pattern
+$config['allowedcharacters_token'] = '/[^0-9a-zA-Z_\-~]/';
```
Shnoulle, line 843: Maybe just the name `regexpattern_token` ? but here it's OK :+1:

Shnoulle: `allowedcharacters_pattern_token` ?

PR #4169, opened 2025-02-13
Fixed issue #19962: SMTP session always closed and reopened between emails

c-schmitz:
See description in Mantis.

### DIFF
--- a/application/controllers/admin/Tokens.php
+++ b/application/controllers/admin/Tokens.php
@@ -1688,7 +1688,8 @@ class Tokens extends SurveyCommonAction
 
                     unset($fieldsarray);
                 }
-
+                // Closes a still active SMTP connection if it exists
+                $mail->smtpClose();
                 $aViewUrls = array();
                 $aData['tokenoutput'] = $tokenoutput;
 
--- a/application/core/LimeMailer.php
+++ b/application/core/LimeMailer.php
@@ -223,6 +223,8 @@ class LimeMailer extends PHPMailer
      */
     public function init()
     {
+        // Make sure that any existing SMTP connection is closed
+        $this->smtpClose();
         $this->debug = [];
         $this->ContentType = self::CONTENT_TYPE_PLAINTEXT;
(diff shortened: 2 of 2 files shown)


c-schmitz:
@tiborpacalat Can you please test this PR as soon as possible?

application/controllers/admin/Tokens.php
```diff
@@ -1688,7 +1688,8 @@ public function email(int $iSurveyId)
 
                     unset($fieldsarray);
                 }
-
+                // Closes a still active SMTP connection if it exists
+                $mail->smtpClose();
```
Shnoulle, line 1692: Didn't broke if mailer are not smtp ?

Shnoulle:
Maybe need to reset (just in case) in new self ?

In init ?
https://github.com/LimeSurvey/LimeSurvey/blob/95c7d9e78bd5715116d0bc07dcfb946b14d87763/application/core/LimeMailer.php#L228

Or maybe  need to add a function or parameters to set keepAlive : use it when needed (token email)

Because here : connection was not closed after registering, for example.

application/controllers/admin/Tokens.php
c-schmitz: Shouldn't because the function checks if smtp object exists.

c-schmitz:
> Maybe need to reset (just in case) in new self ?
> 
> In init ?
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/95c7d9e78bd5715116d0bc07dcfb946b14d87763/application/core/LimeMailer.php#L228
> 
> Or maybe need to add a function or parameters to set keepAlive : use it when needed (token email)
> 
> Because here : connection was not closed after registering, for example.

so, like a $mail->smtpClose(); right in the start?

application/controllers/admin/Tokens.php
c-schmitz: Not sure if this is even needed as smtpClose is also done in the PHPMailer desctructor by default.

PR #4172, opened 2025-02-17
Fix issue #19955: Link to PopUp in question type image radio

gabrieljenik:


gabrieljenik:
Please review why those lines, which are here removed, were inserted on first place.
They were set on lime Lime 2.2.5

[Dev: some missing definition for uniq upload directory · LimeSurvey/LimeSurvey@f66c58c](https://github.com/LimeSurvey/LimeSurvey/commit/f66c58cae5918d82fdac187605b8cd6614590807)

Shnoulle:
> Please review why those lines, which are here removed, were inserted on first place. They were set on lime Lime 2.2.5
> 
> [Dev: some missing definition for uniq upload directory · LimeSurvey/LimeSurvey@f66c58c](https://github.com/LimeSurvey/LimeSurvey/commit/f66c58cae5918d82fdac187605b8cd6614590807)

My opinion : must work with `$surveyID = 'uniq';` **if** the feature work

Can be added here https://github.com/LimeSurvey/LimeSurvey/blob/1aa7e652fc6efdcb78d2f48e6ae81f5991a2f0e7/application/core/SurveyCommonAction.php#L177

OR use another parameter

Shnoulle:
Or : did the system work without this link ? Maybe kcfinder look at the good directory ?

gabrieljenik:
>  did the system work without this link ? Maybe kcfinder look at the good directory ?

Not sure what you are saying.
Can you expand a bit more?

There is a Lime configuration that is uniq_upload_dir, which is used so that the KCFinder always points to the same folder ("uniq") instead of one folder for each survey:
image

initKcfinde function checks that config before building the paths/urls.
image

The lines removed didn't only affect KCFinder but the entire logic of the html editor. For example, the survey ID is used to determine if the survey supports emails in HTML format, or for field replacement. "uniq" as surveyId makes all those mechanism to fail. So those lines were removed.

Shnoulle:
> initKcfinde function checks that config before building the paths/urls.

OK, then : fixed :)

Shnoulle (approved):

PR #4175, opened 2025-02-18
Fixed issue #19876: RC : no permission control on get_response_ids

Shnoulle:
Dev:

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -1248,6 +1248,9 @@ class remotecontrol_handle
     public function get_response_ids($sSessionKey, $iSurveyID, $sToken)
     {
         if ($this->_checkSessionKey($sSessionKey)) {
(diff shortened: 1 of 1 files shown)


gabrieljenik (commented):
Should they have token permission read as well?

Shnoulle:
> Should they have token permission read as well?

We don't read token table, only survey table.

With same user (in 5.X) with view/export response: 

- Browse response table : do not see token
- Export : see token but no link with token table

gabrieljenik:
> We don't read token table, only survey table.

Well, but after running the method, you would be able to tell something about a token (if it has responses or not).
That's why I mention it... Still, maybe being too picky...

If you don't see it clear, then, disregard my comment.

gabrieljenik (approved):

Shnoulle:
> > We don't read token table, only survey table.
> 
> Well, but after running the method, you would be able to tell something about a token (if it has responses or not). That's why I mention it... Still, maybe being too picky...
> 
> If you don't see it clear, then, disregard my comment.

Yes, but then must be reported for export too :) .
I understand yourt question, then i look at export part : an user can get the response id with the access code :)

PR #4182, opened 2025-02-21
Fixed issue #19904: Last survey and last question must be in SettingsUser

Shnoulle:
Fixed issue : Last question are not updated


application/helpers/update/updates/Update_628.php
```diff
@@ -0,0 +1,23 @@
...
+    public function up()
+    {
+        /* Delete old uneeded settings */
+        $this->db->createCommand(
+            "DELETE FROM {{settings_global}} WHERE " . $this->db->quoteColumnName('stg_name') . " LIKE 'last_survey_%'"
+        )->execute();
+        $this->db->createCommand(
+            "DELETE FROM {{settings_global}} WHERE " . $this->db->quoteColumnName('stg_name') . " LIKE 'last_question_%'"
+        )->execute();
+        $this->db->createCommand()->update('{{settings_global}}', array('stg_value' => 628), "stg_name='DBVersion'");
```
gabrieljenik, line 21: Is this need?

application/helpers/update/updates/Update_628.php
```diff
@@ -0,0 +1,23 @@
...
+
+use LimeSurvey\Helpers\Update\DatabaseUpdateBase;
+
+class Update_628 extends DatabaseUpdateBase
+{
+    /**
+     * @inheritDoc
+     */
+    public function up()
+    {
```
gabrieljenik, line 13: I was expecting here to migrate current settings in config to user settings, shouldn't we to avoid loss of data? (I know, not so critical, so not sure it is worth it)

application/helpers/update/updates/Update_628.php
Shnoulle: Yes, original issue : i have a config loaded each time with more than 6000 lines like this.

Shnoulle: And : **allways** remove undeeded part from DB are a good practice.

application/helpers/update/updates/Update_628.php
Shnoulle: > I was expecting here to migrate current settings in config to user settings, shouldn't we to avoid loss of data? (I know, not so critical, so not sure it is worth it)

I start with a solution to migrate (only surey since question are not saved) but https://bugs.limesurvey.org/view.php?id=19904#c82083

And the solution must be tested on MS et PG.

application/helpers/update/updates/Update_628.php
gabrieljenik: I meant the setting the version... sorry

Shnoulle: Yes : else it don't happen .

There are no way to delete data one time only except update the database number.

gabrieljenik: Not sure I follow. All other migration scripts don't do that, right?

Shnoulle: > Not sure I follow. All other migration scripts don't do that, right?

Arg, you're right. Misunderstanding the point; System do it automatically.

Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/4182/commits/4433fa0b5c50e8dc811f2883c672a7616b208bd4

Shnoulle:
It's OK no @gabrieljenik ?

gabrieljenik (approved):

gabrieljenik:
Code structure looks good. Haven´t tested it.

PR #4190, opened 2025-03-05
Fixed issue #19994: Unable to get to dashboard in some condition

Shnoulle:
Dev:

### DIFF
--- a/application/core/LSETwigViewRenderer.php
+++ b/application/core/LSETwigViewRenderer.php
@@ -668,7 +668,10 @@ window.addEventListener('message', function(event) {
             if (!empty($aData["sid"]) || LimeExpressionManager::getLEMsurveyId()) {
                 $sid = empty($aData["sid"]) ? LimeExpressionManager::getLEMsurveyId() : $aData["sid"];
                 $language = empty($aData["language"]) ? App()->getLanguage() : $aData["language"];
-                $aData["aSurveyInfo"] = getSurveyInfo($sid, $language);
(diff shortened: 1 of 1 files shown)


gabrieljenik (approved):
COde srtucture looks good. Haven't tested it.

application/core/LSETwigViewRenderer.php
```diff
@@ -668,7 +668,10 @@ private function getAdditionalInfos($aData, $oTemplate)
             if (!empty($aData["sid"]) || LimeExpressionManager::getLEMsurveyId()) {
                 $sid = empty($aData["sid"]) ? LimeExpressionManager::getLEMsurveyId() : $aData["sid"];
                 $language = empty($aData["language"]) ? App()->getLanguage() : $aData["language"];
-                $aData["aSurveyInfo"] = getSurveyInfo($sid, $language);
+                /* Outdated sid in LimeExpressionManager */
```
gabrieljenik, line 671: I think this could be on a block on its own, before line 677.
So you have one block for param initialization (current line 666).
Then a new one for initializing  $aData["aSurveyInfo"]...

But this works ok like this, so no need to reorganize

Shnoulle: > I think this could be on a block on its own, before line 677. So you have one block for param initialization (current line 666). Then a new one for initializing $aData["aSurveyInfo"]...

Yes initialized to an empty array but since we get from function, and function retur false is surveyu don't exist. Then : check if exist to keep the empty array

PR #4191, opened 2025-03-05
Fixed issue #18713: Better focus on radio and checkbox : keyboard navigation broken

Shnoulle:
Dev: template-core part


themes/survey/fruity_twentythree/css/base.css
```diff
@@ -34,3 +34,55 @@
     margin-bottom: 1rem;
 }
 
+/* Fix for kayboard navigation */
```
twilligls, line 37: The changes for fruity_twentythree should be done in the scss files (radio_item.scss, radio_item_colors.scss, checkboxes.scss, checkboxes-colors.scss)
Also in those file whenever possible, variables should be used (#2C3E50 is not part of the design color scheme - something from theme_color_variables.scss should be picked)

We also have automatic building of rtl when using the scss files, so the whole dir-ltr dir-rtl is not necessary anymore. When some element has left: 3px; the gulp process will automaticallly create files for rtl with right: 3px;

The scss files will get processed by using gulp (command: "gulp build_survey_theme_ls6")

twilligls (changes requested):

themes/survey/fruity_twentythree/css/base.css
Shnoulle: > The changes for fruity_twentythree should be done in the scss files (radio_item.scss, radio_item_colors.scss, checkboxes.scss, checkboxes-colors.scss)

Where is this file ?

https://github.com/LimeSurvey/LimeSurvey/tree/master/themes/survey/fruity_twentythree

**I ask for scss or sass files since years now !** Put the scss file before ask toi use it !

Shnoulle: You've got to be kidding me!  https://bugs.limesurvey.org/view.php?id=19996

Shnoulle:
@twilligls : unable to do change requested without the files.

themes/survey/fruity_twentythree/css/base.css
twilligls: look here: assets/survey_themes/fruity_twentythree

Shnoulle:
I search on theme : file related to theme must be in theme …

themes/survey/fruity_twentythree/css/base.css
twilligls: looks very good to me now

twilligls (approved):

PR #4193, opened 2025-03-05
QE-1471: Created a new endpoint for getting archive information that can be imported

lajosarpad:
Dev:


application/helpers/database_helper.php
```diff
@@ -103,14 +103,3 @@ function dbSelectTablesLike($table)
             safeDie("Couldn't create 'select tables like' query for connection type '" . Yii::app()->db->getDriverName() . "'");
     }
 }
-
-/**
```
c-schmitz, line 107: This function is used in alot of places. Why did you move it to the import_helper? You will now have to check all places where it is used and makre sure that import_helper is loaded, instead.

lajosarpad: I was moving another function and by accident moved this one as well. Thanks for pointing it out.

application/libraries/Api/Command/V1/SurveyArchive.php
```diff
@@ -0,0 +1,121 @@
...
+        if (!$survey) {
+            return $this->responseFactory->makeErrorNotFound(
+                (new ResponseDataError(
+                    'SURVEY_NOT_FOUND',
+                    'Survey not found'
+                )
+                )->toArray()
+            );
+        }
+        require_once "application/helpers/admin/import_helper.php";
```
kevin-foster-uk, line 62: This helper should ideally be refactored in to a service and for re-use between the REST API and the core application. I understand we don't have time to do it now. Maybe its worth adding a comment in the code so we are reminded of it?

PR #4196, opened 2025-03-11
Fixed issue #20004: User massive action drop-down list not translated

FGE18:
Fixed issue #20004<Mantis issue number>: User massive action drop-down list not translated


Shnoulle (approved):
:+1:  thanks

PR #4200, opened 2025-03-13
Fixed Issue #19742: Label set export anomaly

HaroWana:
Dev: After exporting a label set, the page would be stuck on the loading screen. I added an focus event listener to the export button handler in order to ensure that the loading screen is hidden.

Fixed issue #19742: Label set export anomaly
Dev: After exporting a label set, the page would be stuck on the loading screen. I added an focus event listener to the export button handler in order to ensure that the loading screen is hidden.

### DIFF
--- a/assets/scripts/admin/labels.js
+++ b/assets/scripts/admin/labels.js
@@ -33,6 +33,12 @@ $(document).on('ready  pjax:scriptcomplete', function(){
         }
     });
 
+    // Use window focus trick to ensure loading spinner is hidden after export
+    $('#exportlabelset').on('submit', function(){
+        $(window).one('focus', function(){ $('#ls-loading').hide(); });
+    });
+
+
     const answersTable = $(".answertable tbody");
     if (answersTable.length && answersTable.children().length == 0) {
         add_label(undefined);


assets/scripts/admin/labels.js
```diff
@@ -33,6 +33,12 @@ $(document).on('ready  pjax:scriptcomplete', function(){
         }
     });
 
+    // Use window focus trick to ensure loading spinner is hidden after export
```
gabrieljenik, line 36: I am not sure about this. With this solution you will get the spinner to hide when someone clicks on the window... It shouldn't be the user who triggers the hiding, but something from the export process. If there is nothing that can be done automated, we could show a message like "the export will start, please close when ready".

Would like others opinion here.

HaroWana:
The CI job failure doesn't seem to be coming from my patch, or am I missing something?

assets/scripts/admin/labels.js
HaroWana: I submitted a new commit which addresses this. In this new commit, the hiding of the loading spinner is triggered by a token set by the backend.

application/controllers/admin/Export.php
```diff
@@ -889,9 +889,23 @@ public function dumplabel()
...
+        if ($token = Yii::app()->request->getPost('export_token')) {
+            Yii::app()->session[$token] = true;
+            session_write_close();
+        }
         Yii::app()->end();
     }
 
+    public function exportstatus($token)
+    {
+        $done = isset(Yii::app()->session[$token]) && Yii::app()->session[$token] === true;
```
Shnoulle: Simple : `$done = boolval(Yii::app()->session[$token])`

And must be set to false after, right ?

application/controllers/admin/Export.php
```diff
@@ -889,9 +889,23 @@ public function dumplabel()
 
         $xml->endElement(); // close columns
         $xml->endDocument();
+        if ($token = Yii::app()->request->getPost('export_token')) {
+            Yii::app()->session[$token] = true;
+            session_write_close();
```
Shnoulle: ? I didn't like to force close session here. Already don in App()->end(), no ?

application/controllers/admin/Export.php
HaroWana: So far the token is not set to false since we generate a unique token for each export. They do accumulate if several label export are executed in one session but I figured 3kB or so per token wouldn't have too big of an impact and having unique token ids would make them safer(?).
I'm not dead-set on this choice, if you think it's bad idea I can change it.

`$done = boolval(Yii::app()->session[$token])` : this is better indeed, I shall apply it. Thank you.

application/controllers/admin/Export.php
HaroWana: It was a remainder from an attempt to solve one of the problems I had while coding this. My bad. I'll remove it.

assets/scripts/admin/labels.js
```diff
@@ -24,15 +24,23 @@ $(document).on('click','[data-action="deletelabelset"]',function(event){
...
-        if ($('#labelsets > option:selected').size()==0) {
-            alert(strSelectLabelset);
-            return false;
-        } else {
-            return true;
-        }
+    $('#exportlabelset').on('submit', function(){
+        $('#ls-loading').show();
+        const token = $(this).find('input[name="export_token"]').val();
+        const pollUrl = 'index.php?r=admin/export&sa=exportstatus&token=' + encodeURIComponent(token);
```
Shnoulle: Broke with url set to PATH, must use https://www.yiiframework.com/doc/api/1.1/CApplication#createUrl-detail

HaroWana:
I deleted the last comment by mistake, my bad. 
The feedback was addressed in my last patch.
Thank you.

assets/scripts/admin/labels.js
```diff
@@ -24,15 +24,24 @@ $(document).on('click','[data-action="deletelabelset"]',function(event){
...
-    $('#btnDumpLabelSets').click(function(){
-        if ($('#labelsets > option:selected').size()==0) {
-            alert(strSelectLabelset);
-            return false;
-        } else {
-            return true;
-        }
+    $('#exportlabelset').on('submit', function(){
+        $('#ls-loading').show();
+        const token = $(this).find('input[name="export_token"]').val();
```
Shnoulle, line 29: If i don't make error : we use more javascript variables https://github.com/LimeSurvey/LimeSurvey/blob/ed055cebb7ffbefee5cf7adf658b4b8011d11c17/assets/scripts/admin/labels.js#L444

Unsure the good way, but i think we must use same

PR #4202, opened 2025-03-14
New feature #20002: Create 2FA enforcement setting for 2FA plugin

Shnoulle:
Dev: attach to beforeAdminMenuRender, call for all GUI.

Dev:


Shnoulle:
I try to create a new event on App()->getCurrentUserId : but no App()->controller at this time, done too soon maybe.
And we muts not enforce for remote_control or one time pass for exemple.

Here : it's call for a lot of admin GUI, except some json ?

beforeControllerAction are not the best too since we have a lot of dfferent controllers in 6.

lajosarpad (approved):

PR #4207, opened 2025-03-20
Fixed issue #14704: DbStorage not found

Shnoulle:
Dev: hack to include application.libraries.PluginManager.Storage in loader
Dev: as know LimeSurvey\\PluginManager\\

Dev:

### DIFF
--- a/application/libraries/PluginManager/PluginManager.php
+++ b/application/libraries/PluginManager/PluginManager.php
@@ -182,15 +182,18 @@ class PluginManager extends \CApplicationComponent
      */
     public function getStore($storageClass)
     {
-        if (
-            !class_exists($storageClass)
-                && class_exists('LimeSurvey\\PluginManager\\' . $storageClass)
-        ) {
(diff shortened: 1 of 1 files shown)


Shnoulle:
Related PR

- [ Fixed issue #14704: DbStorage not found, API major version update #4209 ](https://github.com/LimeSurvey/LimeSurvey/pull/4209)
- [Fixed issue #14704: DbStorage not found : move Storage at good dir](https://github.com/LimeSurvey/LimeSurvey/pull/4208)
- [ Fixed issue #14704: DbStorage not found  : load manually with Yii](https://github.com/LimeSurvey/LimeSurvey/pull/4207)

lajosarpad (changes requested):
We should not suppress the errors, but fix them. We need to check whether the file exists either in the way I suggested, like

```
class_exists($yourpath, false) && class_exists($yourpath)
```

or in some other manner, as long as it fixes the bug, it's safe and it works.

PR #4208, opened 2025-03-20
Fixed issue #14704: DbStorage not found

Shnoulle:
Dev: autoloader didn't look in Storage directory
Dev: then move to PluginManager directory
Dev: check class only one time

PR #4209, opened 2025-03-20
Fixed issue #14704: DbStorage not found, API major version update

Shnoulle:
Dev: can broke existing plugin
Dev: move Storage to Storage directory
Dev: load class only one time

### DIFF
--- a/application/libraries/PluginManager/PluginManager.php
+++ b/application/libraries/PluginManager/PluginManager.php
@@ -182,16 +182,17 @@ class PluginManager extends \CApplicationComponent
      */
     public function getStore($storageClass)
     {
-       if (!@class_exists($storageClass, true)) {
+        if (isset($this->stores[$storageClass])) {
+            return $this->stores[$storageClass];
+        }
+        if (!@class_exists($storageClass, true)) {
             if (@class_exists('LimeSurvey\\PluginManager\\Storage\\' . $storageClass)) {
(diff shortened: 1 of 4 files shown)


Shnoulle:
Related PR

- [ Fixed issue #14704: DbStorage not found, API major version update #4209 ](https://github.com/LimeSurvey/LimeSurvey/pull/4209)
- [Fixed issue #14704: DbStorage not found : move Storage at good dir](https://github.com/LimeSurvey/LimeSurvey/pull/4208)
- [ Fixed issue #14704: DbStorage not found  : load manually with Yii](https://github.com/LimeSurvey/LimeSurvey/pull/4207)

Shnoulle:
> We need to fix the bug instead of suppressing it

The 1st one always return false for core Storage

And here try/cast didn't work

Trying 
```
        try {
           class_exists($storageClass, true); 
        } catch (Exception $e) {
            /* If $storageClass can not be loaded : try with 'LimeSurvey\\PluginManager\\' . $storageClass */
            try {
                class_exists('LimeSurvey\\PluginManager\\' . $storageClass, true);
                /* If no exception : rename $storageClass */
                $storageClass = 'LimeSurvey\\PluginManager\\' . $storageClass;
            } catch (Exception $e) {
                throw new CException('Unable to find ' . $storageClass . ' class');
            }
        }
```
Throw error at class_exists($storageClass, true);  not at the last one

PR #4211, opened 2025-03-21
Fixed issue #19956: Ugly error message after trying to import not compatible responses

Shnoulle:
Dev: import result page only if error

### DIFF
--- a/application/controllers/admin/DataEntry.php
+++ b/application/controllers/admin/DataEntry.php
@@ -356,9 +356,13 @@ class DataEntry extends SurveyCommonAction
                 }
             }
             $imported = 0;
+            $aWarnings = [];
+            $aSuccess = [];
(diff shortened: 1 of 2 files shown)


Shnoulle:
>  I notice DB Transaction is kept :) which I think it is important

What do you mean ? I don't keep it

gabrieljenik:
> > I notice DB Transaction is kept :) which I think it is important
> 
> What do you mean ? I don't keep it

https://github.com/LimeSurvey/LimeSurvey/pull/4211/files#diff-2617f6e873f36ea81e8663bf2c3a722a95fdea02e5e072759e6877da68c2fe15R392

??

Shnoulle:
> ??

Yes, I use a transaction, and if fail i rollback and on success commit. 

I **add** the transaction, it don't exist before :)

PR #4217, opened 2025-03-26
Fixed issue #20030: Import LDAP issue with ldapv3 and ldpaps

Shnoulle:
Dev: ldaps + ldapv3 : $ds are never set
Dev: review whole logic of parameters

### DIFF
--- a/application/helpers/ldap_helper.php
+++ b/application/helpers/ldap_helper.php
@@ -18,39 +18,36 @@
 /*                      */
 /*********************************************/
 
-
+/**
+ * Create the ldap connexion for token management
+ * @param int|null $server_id
+ * @return LDAP\Connection|false
+ */
 function ldap_getCnx($server_id = null)
 {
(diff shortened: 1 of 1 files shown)


application/helpers/ldap_helper.php
```diff
@@ -18,39 +18,36 @@
...
-            }
-
-            if (!$ldap_server[$server_id]['referrals']) {
-                ldap_set_option($ds, LDAP_OPT_REFERRALS, 0);
-            }
-        }
-        return $ds;
+        $ds = ldap_connect("ldap://" . "{$ldap_server[$server_id]['server']}:{$ldap_server[$server_id]['port']}");
+    }
+    if ($ds === false) {
```
gabrieljenik, line 37: Should this be initialized just in case? I don't see in the code right now how can it end being false.

gabrieljenik (approved):
Code structure looks good. Haven't tested it

Shnoulle:
Tested by reporter

PR #4220, opened 2025-03-27
APE-29: added surveytheme connector for react

ptelu:
Dev:


application/models/services/SurveyThemeConfiguration.php
```diff
@@ -0,0 +1,229 @@
...
+    /**
+     * @param int $surveyId
+     * @param array $props properties (in options json string)
+     *
+     * @return void
+     * @throws PermissionDeniedException
+     */
+    public function updateThemeOption($surveyId, $props): void 
+    {
+        if (
```
lajosarpad, line 31: Let's assume that John Smith has global permission for survey updates, but is denied permission to update 123456. In that case if this is survey 123456, then we need to throw the exception.

As it is, your code will only throw an error if both are false. This misses the case when a global setting is overriden by a survey setting.

lajosarpad (changes requested):
We need the permission logic to be correct and to make sure that the automated checks pass

application/models/services/SurveyThemeConfiguration.php
c-schmitz: Global permissions supercede survey permissions, not the other way around.

lajosarpad: @c-schmitz thanks for pointing that out. Then the code is good here. Marking the conversation as resolved

PR #4221, opened 2025-03-27
Fixed issue #19725: when trying to recover the participants it says There are no matching configurations to start restoring the participants table.

gabrieljenik:


Shnoulle (approved):
Ow ! 

Not a typo, usage of bad variable !!!

PR #4231, opened 2025-04-03
Dev #18352: setting up 2FA with YubiKey

gabrieljenik:


application/core/plugins/TwoFactorAdminLogin/TwoFactorAdminLogin.php
```diff
@@ -196,8 +208,12 @@ public function newUserSession()
         $oTFAModel =  TFAUserKey::model()->findByPk($oIdentity->id);
 
         if ($oTFAModel != null) {
+            if (!in_array($oTFAModel->authType, ['totp', 'yubi'])) {
+                $this->setAuthFailure(null, 'Authentication method not supported');
```
Shnoulle, line 212: no $this->gT for translation ?

application/core/plugins/TwoFactorAdminLogin/TwoFactorAdminLogin.php
```diff
@@ -400,22 +416,65 @@ public function directCallConfirmKey($oEvent, $oRequest)
...
-        TFAUserKey::model()->deleteAllByAttributes(['uid' => $uid]);
+        TFAUserKey::model()->deleteAllByAttributes(['uid' => $tfaUserKey['uid']]);
+
+        $oTFAModel = new TFAUserKey();
+        $oTFAModel->setAttributes($tfaUserKey, false);
+        $oTFAModel->firstLogin = 0;
+        if (!$oTFAModel->save()) {
+            return $this->createJSONResponse(false, gT("The two-factor authentication key could not be stored."));
+        }
+        return $this->createJSONResponse(true, "Two-factor method successfully stored", ['reload' => true]);
```
Shnoulle, line 455: $this->gT here too ?

application/core/plugins/TwoFactorAdminLogin/models/TFAUserKey.php
```diff
@@ -18,13 +18,8 @@ class TFAUserKey extends LSActiveRecord
...
      */
     public static $authTypeOptions = [
-        'google' => 'Google Authenticator',
-        'authy' => 'Authy',
-        'yubi' => 'YubiKey',
-        'auplus' => 'Authenticator Plus',
-        'duo' => 'Duo',
-        'hde' => 'HDE OTP',
-        'other' => 'Other solution'
+        'totp' => 'TOTP (Google Authenticator, Authy, etc.)',
```
Shnoulle, line 21: :+1:

Shnoulle (commented):
Just lack of some gT

c-schmitz (changes requested):
Several issues I see here:

- The Yubikey secret is not encrypted in the plugin settings table
- When trying to activate 2FA with Yubikey, just nothing happens and i end up on the same screen  (yes, I have properly configured the Yubikey settings in the plugin)
- The Yubikey settings should be on a separate tab in the plugin settings, imho there should be three tabs: General , TOTP, Yubikey
- The table for the 2FA is named 'twoFactorUsers'  which is a problem because of the uppercasing  - MySQL cannot do uppercase table names, so function tableExists('{{twoFactorUsers') is always false.
- If the table twofactorusers already exists the plugin cannot be activated.
- I recommed to rename the table twofactorusers  to user_mfa_settings, as no users are saved in there.

Shnoulle:
> Several issues I see here:
>     * The table for the 2FA is named 'twoFactorUsers'  which is a problem because of the uppercasing  - MySQL cannot do uppercase table names, so function tableExists('{{twoFactorUsers') is always false.
>     * If the table twofactorusers already exists the plugin cannot be activated.
>     * I recommed to rename the table twofactorusers  to user_mfa_settings, as no users are saved in there.

Seems to be another issue with this plugin 

https://github.com/LimeSurvey/LimeSurvey/blob/41461c7ea57abba4190c01c674db0cdc265be2ae/application/core/plugins/TwoFactorAdminLogin/installer/TFAPluginInstaller.php#L72

c-schmitz:
Sure, but since we touch the table now anyway, we should take the chance to solve this properly.

gabrieljenik:
I agree with those issues and at some point I don't like too much the design fo the 2FA plugin as it is not designed for different 2FA mechanisms neither (it was designed just for OTP) and also, doesn't allow multiple simultaneous for the same user.

Question is, shall we approach all those issues now? After your message, I think so. But wanted to confirm

c-schmitz (approved):

c-schmitz:
@gabrieljenik The DB update is still not working:

if (!tableExists('{{twoFactorUsers}}')) 

will not work on mysql, you will have to also check for lowercase naming of the table

PR #4232, opened 2025-04-04
MW-1526: getStorage fix

lajosarpad:
Dev:


Shnoulle:
1. Must try with a plugin using "MyPlugn\\Storage\\mystorage", since `class_exists($storageClass, false)` do not try to load $storageClass : it was never loaded.
2. $withNamespace lust be tested only if $storageClass don't contain `\` character

PR #4241, opened 2025-04-14
Fixed issue #20058: Can't retreive a file uploaded with RC api

AlainL88:
Fixed issue #20058 (https://bugs.limesurvey.org/view.php?id=20058):
Dev: Alain Lima

### DIFF
--- a/application/helpers/remotecontrol/remotecontrol_handle.php
+++ b/application/helpers/remotecontrol/remotecontrol_handle.php
@@ -3522,7 +3522,7 @@ class remotecontrol_handle
 
         $tempdir = Yii::app()->getConfig("tempdir");
 
-        $sTempUploadDir = $tempdir . '/upload/';
+		$sTempUploadDir = Yii::app()->getConfig('uploaddir') . "/surveys/" . $iSurveyID . "/files/";
         if (!file_exists($sTempUploadDir)) {
             if (!mkdir($sTempUploadDir)) {
                 return array('status' => 'Can not make temporary upload directory');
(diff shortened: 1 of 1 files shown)


application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -3522,7 +3522,7 @@ public function upload_file($sSessionKey, $iSurveyID, $sFieldName, $sFileName, $
 
         $tempdir = Yii::app()->getConfig("tempdir");
 
-        $sTempUploadDir = $tempdir . '/upload/';
+		$sTempUploadDir = Yii::app()->getConfig('uploaddir') . "/surveys/" . $iSurveyID . "/files/";
```
c-schmitz, line 3525: It is not enough to fix the path. Please also rename the variable.

c-schmitz: Also saved files should not start with futmp_ prefix. You will need to fix that further below. Compare how other files are  named in that directory.

c-schmitz (changes requested):

application/helpers/remotecontrol/remotecontrol_handle.php
AlainL88: Thanks @c-schmitz. I just updated the code.

c-schmitz (approved):

AlainL88:
Hi, any news about this fix?
@gabrieljenik
@Shnoulle

Shnoulle:
Code review is already OK :) i think need time for QA eam for testing

PR #4245, opened 2025-04-16
Fixed issue #20061: Update error in PostgreSQL

gabrieljenik:


c-schmitz (approved):

application/helpers/update/updates/Update_629.php
```diff
@@ -21,7 +21,7 @@ public function up(): void
         $this->db->createCommand()->update(
             '{{user_mfa_settings}}',
             ['authType' => 'totp'],
-            "authType != 'yubi'"
+            $this->db->quoteColumnName("authType") . " != 'yubi'"
```
c-schmitz, line 24: Another proof how bad the idea of uppercasing table/field names was.

application/helpers/update/updates/Update_629.php
```diff
@@ -21,7 +21,7 @@ public function up(): void
         $this->db->createCommand()->update(
             '{{user_mfa_settings}}',
             ['authType' => 'totp'],
```
Shnoulle, line 23: You do not need it here ?

Shnoulle (approved):

application/helpers/update/updates/Update_629.php
gabrieljenik: Nope, not needed there. That is solved by the builder.

gabrieljenik: Below is needed as the condition is passed as plain text

PR #4246, opened 2025-04-18
APE-29: theme settings service bugfix

haythemch2:
Dev:

### DIFF
--- a/application/models/TemplateConfiguration.php
+++ b/application/models/TemplateConfiguration.php
@@ -288,25 +288,19 @@ class TemplateConfiguration extends TemplateConfig
     }
 
     /**
-     * Returns a Theme options array based on a surveyID & templateName
+     * Returns a Theme options array based on a surveyID
      *
      * @param integer $iSurveyId
-     * @param string $sTemplateName
      * @param bool $bInherited should inherited theme option values be used?
      * @return array
      */
-    public static function getThemeOptionsFromSurveyId($iSurveyId = 0, $sTemplateName = null, $bInherited = false)
+    public static function getThemeOptionsFromSurveyId($iSurveyId = 0, $bInherited = false)
     {
         $aTemplateConfigurations = array();
         // fetch all themes belonging to $iSurveyId
         $criteria = new CDbCriteria();
         $criteria->addCondition('sid=:sid');
         $criteria->params = array('sid' => $iSurveyId);
-
-        if ($sTemplateName != null) {
-            $criteria->addCondition('template_name = :template_name');
-            $criteria->params['template_name'] = $sTemplateName;
-        }
 
         $oTemplateConfigurations = self::model()->findAll($criteria);
 
--- a/application/models/TemplateConfiguration.php
+++ b/application/models/TemplateConfiguration.php
@@ -1253,7 +1253,7 @@ class TemplateConfiguration extends TemplateConfig
      * Also triggers inheritence checks
      * @return void
      */
-    protected function setOptions()
+    public function setOptions()
     {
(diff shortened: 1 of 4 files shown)


application/models/TemplateConfiguration.php
```diff
@@ -288,26 +288,20 @@ public static function getInstanceFromSurveyId($iSurveyId, $sTemplateName = null
...
-     * Returns a Theme options array based on a surveyID & templateName
+     * Returns a Theme options array based on a surveyID
      *
      * @param integer $iSurveyId
-     * @param string $sTemplateName
      * @param bool $bInherited should inherited theme option values be used?
      * @return array
      */
-    public static function getThemeOptionsFromSurveyId($iSurveyId = 0, $sTemplateName = null, $bInherited = false)
+    public static function getThemeOptionsFromSurveyId($iSurveyId = 0, $bInherited = false)
```
lajosarpad, line 297: Why did you remove the `$sTemplateName` parameter from `getThemeOptionsFromSurveyId`?

application/models/services/SurveyThemeConfiguration.php
```diff
@@ -164,19 +166,24 @@ public function updateCommon(TemplateConfiguration $model, int $sid = null, int
      *
      * @param integer $iSurveyId
      * @param string $sTemplateName
-     * @param bool $bInherited should inherited theme option values be used?
      * @return array
      */
-    public function getSurveyThemeOptions($iSurveyId = 0, $sTemplateName = null, $bInherited = false): array
+    public function getSurveyThemeOptions($iSurveyId = 0, $sTemplateName = null): array
```
lajosarpad, line 171: Why did you remove the `$bInherited` parameter from `getSurveyThemeOptions`?

application/models/services/SurveyThemeConfiguration.php
```diff
@@ -273,7 +281,7 @@ public function getOptionAttributeDataValue($iSurveyId = 0, $sTemplateName = nul
         $options = $dom->getElementsByTagName('option');
 
         foreach ($options as $option) {
-            if ($option->getAttribute('value') === $optionValue) {
+            if ($option->getAttribute('value') === $lowercasedValue) {
```
lajosarpad, line 284: Are you sure every attribute is lower-case? It would be much safer if you converted `$option->getAttribute('value')` to lower-case too.

lajosarpad (changes requested):
Code-wise it looks good, but we will need some adjustments and clarifications, as described in separate comments.

application/models/TemplateConfiguration.php
haythemch2: In previous PR The `$sTemplateName` was added as param & criteria condition , because this method was used by the `SurveyThemeConfiguration` service , but now the service handles the options on its own & not using this method. ( so this is just cleaning up unnecessary previous changes )

application/models/services/SurveyThemeConfiguration.php
haythemch2: The `$bInherited` parameter was added for flexiblity in previous PR because the `getSurveyThemeOptions` was using `getThemeOptionsFromSurveyId` from `TemplateConfiguration` model , but now the method only returns actual values and not any db values (eg: "inherited") for theme options ( because react UI doesn't need the information if a certain option is inherited or not ) .

application/models/services/SurveyThemeConfiguration.php
haythemch2: totally agree , attribute values are now lowercase too

lajosarpad (approved):

PR #4254, opened 2025-04-30
Fixed issue #20041: No warning message if attachment have issue before email set

Shnoulle:
Dev: move checking function to public static function
Dev: use it

### DIFF
--- a/application/controllers/admin/Tokens.php
+++ b/application/controllers/admin/Tokens.php
@@ -3061,7 +3061,7 @@ class Tokens extends SurveyCommonAction
         $bHtml = (getEmailFormat($iSurveyId) == 'html');
         $bEmail = $sSubAction == 'invite';
         $aTokenIds = $this->getTokenIds();
-
+        $aData['warnings'] = [];
         // Fill empty email template by default text
         foreach ($aSurveyLangs as $sSurveyLanguage) {
             $aData['thissurvey'][$sSurveyLanguage] = getSurveyInfo($iSurveyId, $sSurveyLanguage);
@@ -3074,6 +3074,27 @@ class Tokens extends SurveyCommonAction
                         $aData['thissurvey'][$sSurveyLanguage]["email_{$sSubAction}"] = $aDefaultTexts["invitation"];
                     } elseif ($sSubAction == 'remind') {
                         $aData['thissurvey'][$sSurveyLanguage]["email_{$sSubAction}"] = $aDefaultTexts["reminder"];
+                    }
+                }
+            }
+            // Check if all attachment are here : add a warning in case
+            if (!empty($aData['thissurvey'][$sSurveyLanguage]['attachments'])) {
+                $attachments = $aData['thissurvey'][$sSurveyLanguage]['attachments'];
+                if (is_string($attachments)) {
+                    $attachments = unserialize($attachments);
+                }
+                $type = ($sSubAction == 'invite') ? 'invitation' : 'reminder';
+                if (!empty($attachments[$type]) && is_array($attachments[$type])) {
+                    foreach ($attachments[$type] as $attachment) {
+                        if (!LimeMailer::attachmentExist($iSurveyId, $attachment, false)) {
+                            $aData['warnings'][] = sprintf(
+                                gT("There a an issue with attachment with language %s. You can review it on %semail template%s."), 
+                                $sSurveyLanguage,
+                                "<a href='" . App()->createUrl("admin/emailtemplates", ["sa=>index", "surveyid" => $iSurveyId]) ."'>",
+                                "</a>"
+                            );
+                            break;
+                        }
                     }
                 }
             }
--- a/application/core/LimeMailer.php
+++ b/application/core/LimeMailer.php
@@ -933,39 +933,12 @@ class LimeMailer extends PHPMailer
                     LimeExpressionManager::singleton()->loadTokenInformation($this->surveyId, $this->oToken->token);
                 }
                 foreach ($aAttachments[$attachementType] as $aAttachment) {
-                    if ($this->attachementExists($aAttachment) && LimeExpressionManager::ProcessRelevance($aAttachment['relevance'])) {
+                    if (self::attachmentExist($this->surveyId, $aAttachment) && LimeExpressionManager::ProcessRelevance($aAttachment['relevance'])) {
                         $this->addAttachment($aAttachment['url']);
                     }
                 }
             }
         }
-    }
-
-    private function attachementExists($aAttachment)
-    {
-        $throwError = (Yii::app()->getConfig('debug') && Permission::model()->hasSurveyPermission($this->surveyId, 'surveylocale', 'update'));
-
-        $isInSurvey = Yii::app()->is_file(
-            $aAttachment['url'],
-            Yii::app()->getConfig('uploaddir') . DIRECTORY_SEPARATOR . "surveys" . DIRECTORY_SEPARATOR . $this->surveyId,
-            false
-        );
-
-        $isInGlobal = Yii::app()->is_file(
-            $aAttachment['url'],
-            Yii::app()->getConfig('uploaddir') . DIRECTORY_SEPARATOR . "global",
-            false
-        );
-
-        if ($isInSurvey || $isInGlobal) {
-            return true;
-        }
-
-        if ($throwError && !($isInSurvey || $isInGlobal)) {
-            throw new \CException(sprintf(gT("File not found: %s"), $aAttachment['url']));
-        }
-
-        return false;
     }
 
     /**
@@ -1026,7 +999,6 @@ class LimeMailer extends PHPMailer
         return $aOutList;
     }
 
-
     /**
      * @inheritdoc
      * Override to use a better html to text converter (ex. doesn't removes links)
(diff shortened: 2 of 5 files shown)


Shnoulle:
Error was sent only for debug > 1 now. Since we have the warnings.

Shnoulle:
The _Fix debug condition logic_ is part of the original issue :)

gabrieljenik:
> The _Fix debug condition logic_ is part of the original issue :)
> 
> But more : I don't think it must be related to code debug : it's related to mail and survey management. Debug is here for code.
> 
> A solution can be : do not send an email if attachments are not OK : return false when try to sent with error 'Attachment don't exist'

Not sure I follow completely.

https://github.com/LimeSurvey/LimeSurvey/pull/4254/files#diff-2a7c4def4e0208ae2ab6f4156f43ea01aafcca9f507de5f52675bbbc022a0c24R1025
Here, why debug > 1? Could it be debug > 0?

Shnoulle:
> Could it be debug > 0?

See the issue https://bugs.limesurvey.org/view.php?id=20041 (see old title)

and here : https://bugs.limesurvey.org/view.php?id=20041#c82540

- [x] Show warning on email template
- [x] Shown warning when send email (on GUI)
- [x] Send email without error (test with debug=2)
- [x] Send email without error (with debug=0)

gabrieljenik:
> [#4254 (comment)](https://github.com/LimeSurvey/LimeSurvey/pull/4254#issuecomment-2843472418)
> 
> Because it's not related to **code** issue. It's related to Survey content.

If it is not related to code issues, then we shouldn't be querying the debug config parameter, right?
So, why not removing it and showing the error no matter what that debug config is.

> I think we can really remove the error throw:

I like that there is something that alerts about missing attahcments. If not, users may not get aware of the error.

> and decide if we :
> Send email if there are attachment issue
> Or do not send email if there are attachment issue

I believe we shouldn't send an email if there is an attachment issue.
The attachment may be as important as the body.

Shnoulle:
> I like that there is something that alerts about missing attahcments. If not, users may not get aware of the error.

It's fixed now for this part

> I believe we shouldn't send an email if there is an attachment issue.

@c-schmitz @tiborpacalat decisoon ?

I think too we can disable send email if there are issue with attachment, debug set or not.

application/core/LimeMailer.php
```diff
@@ -1039,4 +1011,38 @@ public function html2text($html, $advanced = false)
...
     }
+
+    /**
+     * Checks that the specified attachment exists on one of the allowed paths.
+     * @param integer $surveyId
+     * @param array $attachment data
+     * @param boolean|null $throwError if null use debug and allow edoit survey content
+     * @return bool True if the file exists within the survey or global upload dirs.
+     */
+    public static function attachmentExist($surveyId, $attachment, $throwError = null)
```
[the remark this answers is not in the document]
Shnoulle: ping @c-schmitz here.

1. Gabriel : "LimeMailer is for delivering emails"
2. Denis : "Attachments are totally related to email delivering"

c-schmitz:
> > I like that there is something that alerts about missing attahcments. If not, users may not get aware of the error.
> 
> It's fixed now for this part
> 
> > I believe we shouldn't send an email if there is an attachment issue.
> 
> @c-schmitz @tiborpacalat decisoon ?
> 
> I think too we can disable send email if there are issue with attachment, debug set or not.

Decision: If the attachment is missing, don't send.

application/core/LimeMailer.php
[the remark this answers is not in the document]
c-schmitz: Decision: Put it in Survey (helper) functions.

c-schmitz:
@coderabbitai review

Shnoulle:
Surely incompatible with https://github.com/LimeSurvey/LimeSurvey/pull/4838

application/core/LimeMailer.php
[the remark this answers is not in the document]
Shnoulle: PS : related to SurveyLanguageSettings 

See https://github.com/LimeSurvey/LimeSurvey/pull/4838/changes

https://github.com/LimeSurvey/LimeSurvey/blob/a539d0d03c93987f673a6f19deaab1c5f8b8743f/application/models/SurveyLanguageSetting.php#L379
and
https://github.com/LimeSurvey/LimeSurvey/blob/a539d0d03c93987f673a6f19deaab1c5f8b8743f/application/models/SurveyLanguageSetting.php#L425

Shnoulle:
> Decision: If the attachment is missing, don't send.

We must choose what we commit before needing to redo after https://github.com/LimeSurvey/LimeSurvey/pull/4838
(and code are different)

c-schmitz:
Security fix https://github.com/LimeSurvey/LimeSurvey/pull/4838 first, then we look again at this.

Shnoulle:
Both done + add `bypassattachement` system.

application/core/LimeMailer.php
```diff
@@ -920,23 +928,22 @@ public function doReplacements($string)
...
     public function addAttachementsByType()
     {
         if ($this->_bAttachementTypeDone) {
-            return;
+            return true;
         }
         $this->_bAttachementTypeDone = true;
         if (empty($this->surveyId)) {
-            return;
+            return true;
```
gabrieljenik, line 941: Is this expected? even if the attachment is not done, the function returns true?

application/core/LimeMailer.php
```diff
@@ -611,7 +616,10 @@ public function sendMessage()
             $this->Subject = mb_convert_encoding($this->Subject, $this->CharSet, $this->BodySubjectCharset);
             $this->Body = mb_convert_encoding($this->Body, $this->CharSet, $this->BodySubjectCharset);
         }
-        $this->addAttachementsByType();
+        if (!$this->addAttachementsByType() && !$this->bypassattachement ) {
+            $this->setError(gT('Email was not sent because attachment did not exist.'));
```
gabrieljenik: I would change the message to "Email was not sent. One or more attachments failed."

application/core/LimeMailer.php
```diff
@@ -950,6 +957,11 @@ public function addAttachementsByType()
                 }
             }
         }
+        // If some attachment are not added, disable send
+        if (!$oSurveyLanguageSetting->hasAllAttachments($attachementType)) {
```
gabrieljenik, line 963: This is being checked over SurveyLanguage. I would have thought the validation was to be done over the LimeMailer, analyzing the attachments done on the prior loop (couple of lines up).

application/core/LimeMailer.php
```diff
@@ -1042,4 +1053,38 @@ public function html2text($html, $advanced = false)
...
     }
+
+    /**
+     * Checks that the specified attachment exists on one of the allowed paths.
+     * @param integer $surveyId
+     * @param array $attachment data
+     * @param boolean|null $throwError if null use debug and allow edoit survey content
+     * @return bool True if the file exists within the survey or global upload dirs.
+     */
+    public static function attachmentExist($surveyId, $attachment, $throwError = null)
```
gabrieljenik: I think this is orphan, am I right? Can't find where it used.

application/views/admin/token/invite.php
```diff
@@ -21,63 +21,81 @@
...
+                                    'selectOptions' => [
+                                        '1' => gT('On'),
+                                        '0' => gT('Off'),
+                                    ],
+                                ]); ?>
                             </div>
                         </div>
+                        <div class="col-md-3">
+                            <?php if ($countInvalidAttachments > 0): ?>
+                                <label class='form-label' for='bypassattachement'><?php eT("Bypass attachment control:"); ?></label>
```
gabrieljenik: Not to be picky, but I feel this is not a Bypass but more an "Ignore Attachment Errors". I think it is more clear and could avoid confusions. Just a comment and a point of view. Not a request for change.

application/core/LimeMailer.php
Shnoulle: If there are  no survey IDs (admin global email; email tester etc ...) : there are no attachments. Then yes, it's expected.

We work only on survey template attachments here.

application/core/LimeMailer.php
Shnoulle: No, because we add other existing attachments before the test. Then if you have some attachments and one is lost and force sending, you send the email.

Else : yes, it's for SurveyLanguage, not for all. You can send DE messages (have all attachments) but not FR because of a lack of attachment.

application/core/LimeMailer.php
Shnoulle: Merge issue I think ...

application/views/admin/token/invite.php
Shnoulle: I'm not the best in English :( @c-schmitz ?

application/core/LimeMailer.php
gabrieljenik: What I mean is that the comment says "If some attachment are not added, disable send".
But the method `hasAllAttachments` doesn't look in the email, only in the survey.
Maybe I am missleading because of the comment?

application/views/admin/token/invite.php
c-schmitz: I agree to Gabriels comment.

c-schmitz: "Ignore missing attachments" maybe

application/core/LimeMailer.php
Shnoulle: You're right : it must be something like that. If lack of some attachment, return false.

application/core/LimeMailer.php
Shnoulle: Not really, more "Email was not sent. One or more attachments did not exist." ?

application/core/LimeMailer.php
Shnoulle: No : must deprecate, but leave it for potential plugin.

application/controllers/admin/Tokens.php
```diff
@@ -3172,11 +3174,23 @@ protected function showInviteOrReminderEmailForm(int $iSurveyId, array $aSurveyL
...
+            // Check if all attachment are here : add a warning in case
+            if (!$SurveyLanguageSetting->hasAllAttachments($type)) {
+                $aData['warnings'][] = sprintf(
+                    gT("There is an issue with an attachment for language %s. You can review it in the %semail template%s."),
+                    $sSurveyLanguage,
+                    "<a href='" . App()->createUrl("admin/emailtemplates", ["sa" => "index", "surveyid" => $iSurveyId]) . "'>",
+                    "</a>"
+                );
+                $countInvalidAttachments++;
+            }
```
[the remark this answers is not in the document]
Shnoulle: My opinion : if SurveyLanguageSetting is null, it's better if it break here or before. !

application/core/LimeMailer.php
Shnoulle: Added some more comments

application/views/admin/token/invite.php
Shnoulle: All rename, include parameters and variable

Shnoulle:
> Attachment-check failure can be bypassed on retry due to stale _bAttachementTypeDone state.

1. We really don't want to add an attachment 2 times
2. If an attachment didn't exist at the start of action, it doesn't exist at the end.

c-schmitz (approved):

PR #4256, opened 2025-04-30
Fixed issue #20057: bulk setting the expiration date

gabrieljenik:


Shnoulle (approved):

Shnoulle:
@gabrieljenik you have an idea why jQuery is not used here ?

gabrieljenik:
> @gabrieljenik you have an idea why jQuery is not used here ?

No idea.

PR #4275, opened 2025-05-13
Dev: news feature : send an email when plugin was deactivated

Shnoulle:
Dev:

### DIFF
--- a/application/config/internal.php
+++ b/application/config/internal.php
@@ -189,15 +189,6 @@ $internalConfig = array(
                     'logFile' => 'plugin.log',
                     'categories' => 'plugin.*'  // The category will be the name of the plugin
                 ),
-                // Plugin error happens, plugin was disabled : send email using CLogRoute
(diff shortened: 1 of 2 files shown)


c-schmitz:
There is a PSR linting error in code-check and also the Qodo suggestion up there looks good.

Shnoulle:
> There is a PSR linting error in code-check and also the Qodo suggestion up there looks good.

No need try/catch, right ?

Shnoulle:
> > There is a PSR linting error in code-check and also the Qodo suggestion up there looks good.
> 
> No need try/catch, right ?

And since we already log the primary error : it make a duplicate in the logs

Shnoulle:
About

> Add error handling around the email sending process to prevent the function from
failing if email sending fails. This ensures the plugin deactivation process
completes successfully even if notification fails.

It's tghe reason i deactivate and only after send email.

PR #4284, opened 2025-05-16
Fixed issue #19844 : Menu in fruity_twentythree are unreadable

Shnoulle:
https://www.w3.org/WAI/ARIA/apg/patterns/menu-button

Strangely Up and Down work on BS5 but not here. Since it's optional i leave it.

Not really a menu, more tools


c-schmitz (approved):

tiborpacalat:
This has already been fixed by external team that is working on accessibility before the PR got approved. I just waited to close this because there was a small issue that was also resolved in the meantime.

PR #4285, opened 2025-05-16
Fixed issue #19958: Public registration is not compliant to GDPR

Shnoulle:
Dev: added in theme only


c-schmitz:
I checked  functionality and I see two things.

- Privacy policy should be displayed below the input fields and above the submit button
- There is an issue in the theme settings:

image

c-schmitz:
I am also wondering if this setting should rather be in the participants settings than a theme setting.

Shnoulle:
> Privacy policy should be displayed below the input fields and above the submit button

I really think it's best before adding any information. Participant need the information before put name and email (and others). But no problem : i can move it.


> I am also wondering if this setting should rather be in the participants settings than a theme setting.

I ask https://bugs.limesurvey.org/view.php?id=19958#c82087

Like you want.


> There is an issue in the theme settings:
My opinion : it's more a theme system issue… I can update the core template settings in DB. But there still have issue about theme : you can not add an option in parent theme to get it in child's theme. Really hard to produce new feature.

Shnoulle:
> I am also wondering if this setting should rather be in the participants settings than a theme setting.

What is your decision here ? Theme settings or Survey setting ?

Shnoulle:
https://github.com/LimeSurvey/LimeSurvey/pull/4880

PR #4286, opened 2025-05-19
APE-110: Added hasTokens to the survey detail GET endpoint

lajosarpad:
We expect a hasToken attribute to be present inside the response to the survey details endpoint


haythemch2 (approved):
Looks good !

PR #4290, opened 2025-05-22
Fixed Issue #20120: Advanced ranking height and width are set to 0 wh…

fsay2604:
…en conditional display is applied.

Fixed issue #20120: Advanced ranking height and width are set to 0 when conditional display is applied.

I added a mutation observer on the question to trigger fixChoiceListHeight() when hidden class is removed so the dimensions are recalculated.

### DIFF
--- a/themes/question/ranking_advanced/survey/questions/answer/ranking/assets/scripts/advanced_ranking.js
+++ b/themes/question/ranking_advanced/survey/questions/answer/ranking/assets/scripts/advanced_ranking.js
@@ -228,10 +228,43 @@ var AdvancedRankingQuestion = function (options) {
         });
     };
 
+    /**
+     * Observe visibility changes on the question element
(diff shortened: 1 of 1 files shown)


gabrieljenik:
I amo not sure about this approach. We are creating an observer to watch a change in a question class, whereas the intention behind it is to trigger some code when relevance is changed. I am not a big fan of that implementation assumption

For doing that, I guess the best is to trigger the code on the LEMrel functions, or at least the value for `#releveanceXXX` field.

@Shnoulle knows more about EM

fsay2604:
Jquery.actual() seems to work.
I've implemented the solution with it instead!

Thanks for the review!

Shnoulle:
> Jquery.actual() seems to work. I've implemented the solution with it instead!
> 
> Thanks for the review!

It's still loaded ?

Shnoulle (approved):

PR #4291, opened 2025-05-22
A11y welcome text question fixes

anblikauditteam:

Select element doesn't have an accessible name, ensure select element has an accessible name.
--
The screen reader does not announce the label; it should announce 'Change language' for the combo box.
The screen reader announces the following visual headings as plain text. It should announce the appropriate heading levels as follows:- Sample LimeSurvey EN/FR/DE/IT/RU as <h1>- This survey is anonymous. as <h2>

The progress tooltip 'You have completed 0% of this survey' is not dismissible using the 'Esc' key on the mouse pointer; it should be dismissible using the 'Esc' key.
--
The progress tooltip 'You have completed 0% of this survey' is not accessible, it should be accessible using the keyboard.

The focus indicator is not visible on the button. The focus indicator should be visible on the button.
--
The text 'English-English' gets truncated when text spacing properties are applied. The text should not be truncated when applying text spacing properties.

Dev:


themes/survey/fruity_twentythree/scripts/custom.js
```diff
@@ -6,10 +6,18 @@
...
+	 * @see https://learn.jquery.com/using-jquery-core/document-ready/
+	 */
+	document.body.addEventListener('keydown', function (e) {
+		if (e.key == 'Escape') {
+			console.log('Escape key pressed');
+			const exampleTriggerEl = document.getElementById('progressbar-top');
+			const tooltip = bootstrap.Tooltip.getInstance(exampleTriggerEl);
+			tooltip.hide();
+		}
+	});
```
ptelu: This file should not be used for internal code, this file is used for user based extensions.
In general since all tooltips should have this functionality i would create a folder and module for tooltips in the fruity_twentythree folder and then the module needs to be loaded inside theme_js_modules.js.
The code should be adjusted to use data-bs-toggle="tooltip" as an identifier,

themes/survey/fruity_twentythree/css/custom.css
```diff
@@ -6,3 +6,8 @@
     It will be loaded last, so you can override any other property.
     Also, it will never be updated. So if you inheritate a core template and just add here some CSS, you'll still benefit of all the updates
 */
+
+.form-change-lang #language-changer-select:focus {
+    outline: 2px solid;
+    outline-offset: 2px;
+}
```
ptelu: This needs to be inside the language-changer.scss inside the theme, this file should not be touched. Just checked in with designer, for now the focus state should be
border: 1px solid $active-green;
border-radius: 4px;
padding: 4px;

themes/survey/fruity_twentythree/scripts/theme.js
```diff
@@ -656,7 +656,6 @@ function activateLanguageChanger() {
     $('<input type="hidden" id="onsubmitbuttoninput" name="move" value="changelang" />').appendTo(limesurveyForm);
     limesurveyForm.submit();
   };
```
ptelu: changes to this file should be reverted, they are not permanent, this file will be overwritten by the theme generation. The file that should be changed is old_template_core_pre.js
Also this removing this code breaks the width of the element in certain texts making the carret appear over the text. The code is by design not static, it will adjust the width length of text. The select is not affected an renders in full length.

themes/survey/fruity_twentythree/views/subviews/survey/question_subviews/question_text_container.twig
```diff
@@ -20,6 +20,6 @@
 #}
 
 <!-- Question text -->
-<div class="{{ aSurveyInfo.class.questiontitlecontainer }}  col-12 " {{ aSurveyInfo.attr.questiontitlecontainer }} >
+<div id="question-title-container-{{ aQuestion.SGQ }}" class="{{ aSurveyInfo.class.questiontitlecontainer }}  col-12 " {{ aSurveyInfo.attr.questiontitlecontainer }} >
```
ptelu: would be more consistent to name it question-text-container- instead

themes/survey/fruity_twentythree/views/subviews/survey/group_subviews/group_name.twig
```diff
@@ -20,7 +20,12 @@
...
 {% if aGroup.showgroupinfo == 'N' or aGroup.showgroupinfo == 'B' %}
-    <div class="{{ aSurveyInfo.class.grouptitle }}" {{ aSurveyInfo.attr.grouptitle }}>
+    <h1 class="{{ aSurveyInfo.class.grouptitle }}" {{ aSurveyInfo.attr.grouptitle }}>
         {{ processString(aGroup.name) }}
-    </div>
+    </h1>
 {% endif %}
+
+<script>
+    document.title = "{{ processString(aGroup.name) }} | " + document.title;
```
ptelu: If this is neccassary then i would add this to the fruity_twentythree group folder and add a javascript module/file here that will target the group title as identifier. Then it can be imported into the theme generation with theme_js_modules.js

themes/survey/fruity_twentythree/views/subviews/survey/question_container.twig
```diff
@@ -18,7 +18,10 @@
...
 #}
 <!-- Question {{ aQuestion.qid  }}  -->
-<div {{ aQuestion.attributes }} class="row {{ aQuestion.class }} {{ aQuestion.input_error_class }} {{ aSurveyInfo.class.questioncontainer }}" {{ aSurveyInfo.attr.questioncontainer }}>
+<div role="group" aria-labelledby="question-title-container-{{ aQuestion.SGQ }}" {{ aQuestion.attributes }} class="row {{ aQuestion.class }} {{ aQuestion.input_error_class }} {{ aSurveyInfo.class.questioncontainer }}" {{ aSurveyInfo.attr.questioncontainer }}>
     {{ include('./subviews/survey/question.twig') }}
 </div>
 <!-- End of question  {{ aQuestion.qid  }}  -->
+<script>
+document.querySelector('[aria-labelledby="ls-question-text-{{ aQuestion.SGQ }}"] > table')?.removeAttribute('role');
+</script>
```
ptelu: could you provide some information where this appears exactly?

ptelu (changes requested):

themes/survey/fruity_twentythree/views/subviews/survey/question_container.twig
anblikauditteam: For the questions type "Array Texts" an unnecessary is added which is preventing screenreader to consider it as table so I removed the role.

image

ptelu: okai, in that case you can directly go to the view files of the questiontype itself and remove it from the twig file.
In this case it is located in application/views/survey/questions/answer/arrays/texts/answer.twig:29

themes/survey/fruity_twentythree/views/subviews/survey/group_subviews/group_name.twig
anblikauditteam: This is done!

themes/survey/fruity_twentythree/views/subviews/survey/question_container.twig
anblikauditteam: This is done!

PR #4302, opened 2025-05-28
Fixed issue #20055: Missing save confirmation with validation errors

gabrieljenik:
Seems sometimes `$this->sMove ` may be blank.
Ex: https://github.com/LimeSurvey/LimeSurvey/pull/2889/files#diff-f2c0610f506dc0397d2efd8d8203a78dc44f2825e26eb0ac70efd87205d2e36eR1136

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -910,7 +910,7 @@ class SurveyRuntimeHelper
      */
     private function setPrevStep()
     {
-        if (isset($this->sMove) && $this->sMove != "") {
+        if (isset($this->sMove) && $this->sMove !== "") {
             if (!in_array($this->sMove, array("clearall", "changelang", "saveall", "reload"))) {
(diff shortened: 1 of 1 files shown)


Shnoulle:
But after save : move must be set to save or something like that ?

Strabngley : before we always add a move='default' input …
image


https://github.com/SondagesPro/LimeSurvey-SondagesPro/blob/d0f5b11ff5b72864e7f012ba7abf05085e267f86/themes/survey/vanilla/views/subviews/header/start_form.twig#L71

application/helpers/SurveyRuntimeHelper.php
```diff
@@ -910,7 +910,7 @@ private function checkClearCancel()
      */
     private function setPrevStep()
     {
-        if (isset($this->sMove)) {
+        if (isset($this->sMove) && $this->sMove != "") {
```
Shnoulle, line 913: `!empty($this->sMove)` ?

or `isset($this->sMove) && $this->sMove` since we use !=

Shnoulle (commented):
Just a better test for `$this->sMove != ""`

Shnoulle (approved):

PR #4303, opened 2025-05-29
New feature APE-99: Archive management page

haythemch2:
## Release Notes

* **New Features**
  * Added survey archive management API endpoints for listing, retrieving, and exporting archived survey data
  * Implemented archive aliasing functionality to label and identify archived surveys
  * Added ability to delete specific archived data types with timestamp-based precision
  * Enabled streaming CSV export for archived survey responses and participant tokens
  * Added filtering, sorting, and pagination support for archived data retrieval


application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveDetails.php
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        $searchParams = [
+            'filters' => $request->getData('filters', []),
+            'sort' => $request->getData('sort', []),
+            'page' => (int) $request->getData('page', 1),
+            'pageSize' => (int) $request->getData('pageSize', 10),
+        ];
+
+        switch ($archiveType) {
```
lajosarpad: It is not necessary, but as a general idea, in the future we can apply something like this:

```
$archiveTypeMap = [
    SurveyArchiveService::$Response_archive => 'getResponseArchiveData',
    SurveyArchiveService::$Tokens_archive => 'getTokenArchiveData'
];
```

and

```
if (!$archiveTypeMap[$archiveType]) {
    throw new \InvalidArgumentException("Unsupported archive type");
} else {
    $data = $this->surveyArchiveService($surveyId, $timestamp, $searchParams);
}
```

I think it would look cleaner.

application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveExport.php
```diff
@@ -0,0 +1,122 @@
...
+            return $this->responseFactory->makeErrorNotFound(
+                (new ResponseDataError(
+                    'ARCHIVE_NOT_FOUND',
+                    'Archive not found'
+                )
+                )->toArray()
+            );
+        }
+
+        $typePart = $archiveType === SurveyArchiveService::$Tokens_archive ? 'participants' : 'responses';
```
lajosarpad: It is better to have a mapping, because, if in the future we are to have multiple tables, then ternary will no longer be adequate. This is optional, do it only if you prefer to do some small improvements.

application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveExport.php
```diff
@@ -0,0 +1,122 @@
...
+        $fileName = "survey_{$surveyId}_{$typePart}_{$timestamp}.csv";
```
lajosarpad, line 89: We will need to support non-csv exports too, for the file types we already support as part of lsa export. I do understand that this is a first version, but we need to keep track of that plan for the future as well, please add a note to the ticket

application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveExport.php
```diff
@@ -0,0 +1,122 @@
...
+        $contentType = 'Content-Type: text/csv; charset=utf-8';
+        $streamHandler = function () use ($surveyId, $timestamp, $archiveType): void {
+            if ($archiveType === SurveyArchiveService::$Tokens_archive) {
```
lajosarpad: We can reuse the name-action mapping suggested at the details endpoint to make the source-code clear and easy to maintain

application/models/SurveyDynamicArchive.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+/*
+ * LimeSurvey
+ * Copyright (C) 2013 The LimeSurvey Project Team / Carsten Schmitz
```
lajosarpad: Are you sure about the year and the author?

application/models/TokenDynamicArchive.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+/*
+   * LimeSurvey
+   * Copyright (C) 2013 The LimeSurvey Project Team / Carsten Schmitz
```
lajosarpad: Are you sure about the year and the author?

application/models/services/SurveyArchiveService.php
```diff
@@ -0,0 +1,521 @@
...
+        if (empty($ArchivesToDelete)) {
+            $ArchivesToDelete = [
+                self::$Response_archive,
+                self::$Tokens_archive,
+                self::$Timings_archive,
+            ];
+        }
+        foreach ($ArchivesToDelete as $archiveType) {
+            switch ($archiveType) {
+                case self::$Response_archive:
```
lajosarpad: You can create a mapping, like

```
$permissionMapping = [
    self::$Response_archive => 'responses',
    self::$Tokens_archive => 'tokens',
    self::$Timings_archive => 'timings'
];
```

and then:

```
foreach ($ArchivesToDelete as $archiveType => $name) {
    if ($permissionMapping[$archiveType]) {
        $requirePermissions[$name] = 'delete';
    }
}
```

application/models/services/SurveyArchiveService.php
```diff
@@ -0,0 +1,521 @@
...
+                    $requiredPermissions['timings'] = 'delete';
+                    break;
+            }
+        }
+        foreach ($requiredPermissions as $permName => $permType) {
+            if (!$this->permission->hasSurveyPermission($iSurveyID, $permName, $permType)) {
+                throw new PermissionDeniedException('Permission denied for deleting archive data');
+            }
+        }
+        foreach ($ArchivesToDelete as $archiveType) {
```
lajosarpad, line 198: Again, a mapping instead of a switch-case would be more readable and easier to maintain.

application/models/services/SurveyDeactivate.php
```diff
@@ -274,7 +274,7 @@ protected function handleSurveyTable($iSurveyID, $date, &$aData, $userID, $DBDat
         }
 
         $this->app->db->createCommand()->renameTable($sOldSurveyTableName, $sNewSurveyTableName);
-        $this->archiveTable($iSurveyID, $userID, "old_tokens_{$siddate}", 'response', $DBDate, json_encode(Response::getEncryptedAttributes($iSurveyID)));
+        $this->archiveTable($iSurveyID, $userID, "old_survey_{$siddate}", 'response', $DBDate, json_encode(Response::getEncryptedAttributes($iSurveyID)));
```
lajosarpad: This was a code that was in use earlier. Can you explain why do we change the archiveTable's 3rd parameter? Were all the usages verified and tested? I do remember that in the archives table for some reason we have the tokens name instead of the survey name which this one is probably aiming to fix, but do we need this as part of this epic?

It would be safer to undo this as long as it does not damage your features and create a follow-up ticket. The fourth parameter is specifying that it's a response anyway.

lajosarpad (changes requested):
Please take a look at the notes, some of them are optional. You will also need to fix the merge conflicts

application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveDetails.php
haythemch2: I agree , it now looks way cleaner

application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveExport.php
haythemch2: cleaner always better , done !

application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveExport.php
haythemch2: Added a comment to the ticket !

application/libraries/Api/Command/V1/SurveyArchive/SurveyArchiveExport.php
haythemch2: yes, did something similar here too

application/models/SurveyDynamicArchive.php
haythemch2: now cleaned up

application/models/TokenDynamicArchive.php
haythemch2: now cleaned up

application/models/services/SurveyArchiveService.php
haythemch2: yes cleaner , done !

application/models/services/SurveyArchiveService.php
haythemch2: yes more readable , done here too

application/models/services/SurveyDeactivate.php
haythemch2: I wanted to use the tbl_name column here for archive checking & deletion logic in the new service, But yes i totally get your point ,I reverted this change & noted this to be changed in a seperate ticket.
now the table names is handeled by the archive service !

c-schmitz:
@coderabbitai review

PR #4304, opened 2025-05-30
Fixed issue #20083: Pre-Installation check: Session writable re-check…   … is impossible

gabrieljenik:


application/controllers/InstallerController.php
```diff
@@ -1117,4 +1110,27 @@ private function checkDefaultExtensions()
...
+     */
+    private function checkSessionIsWritable()
+    {
+        // Set a value in the sesssion
+        Yii::app()->session['saveCheck'] = 'save';
+        // Close the session so we can start it again
+        session_write_close();
+        // When calling session_start, if there is a saved session, it will be loaded.
+        // If there is no saved session, it will create an empty one.
+        @session_start();
```
Shnoulle, line 1130: Really unsure it work in all case … 

Really no way to use a readirect or something like this to call php script 2 times ?

Shnoulle (commented):
No way to use a redirect or an ajax call this to call php script 2 times ?

gabrieljenik:
closed in favor of #4315

PR #4316, opened 2025-06-05
Fixed issue #19171: Unable to use dropdown in config.xml with core options

gabrieljenik:


Shnoulle (approved):
I don't really understand, but check if `$attribute['parent']` exist seems a needed part !

And do it for other part too

gabrieljenik:
> check if `$attribute['parent']` exist seems a needed part !

What do you mean? empty() already worarounds in case pattern is not in the attribute array.

Shnoulle:
> > check if `$attribute['parent']` exist seems a needed part !
> 
> What do you mean? empty() already worarounds in case pattern is not in the attribute array.

I just mean do the same for 
https://github.com/LimeSurvey/LimeSurvey/blob/7da2f3ea6aa415268c1b419f5ff12af8b44e91d1/application/views/themeOptions/options_core.php#L247
https://github.com/LimeSurvey/LimeSurvey/blob/7da2f3ea6aa415268c1b419f5ff12af8b44e91d1/application/views/themeOptions/options_core.php#L291

etc … 

Not other things, don't add it if empty.

Shnoulle (approved):

PR #4349, opened 2025-07-02
CT-1247: Changed the return type of getDbTotalSize to int

lajosarpad:
getDbTotalSize previously returned a string, which was compared to maxdbsizeforbackup which is an int, but that's incorrect, as

```
7>"69"
```

is evaluated to false. This PR ensures that we return an int, so at the usage when we do

```
                    if ($dbSize <= $dbChecks->dbSize) {
                        return $this->createDbBackup();
                    } else {
                        $backupDb->result = false;
                        $backupDb->message = 'db_too_big';
                    }
```

we compare them numerically.

### DIFF
--- a/application/models/UpdateForm.php
+++ b/application/models/UpdateForm.php
@@ -645,7 +645,7 @@ class UpdateForm extends CFormModel
 
     /**
      * Return the total size of the current database in MB
-     * @return string
+     * @return int
      */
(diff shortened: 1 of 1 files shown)


c-schmitz:
I agree on the Qodo comment. Please remove the number_format.

PR #4356, opened 2025-07-09
Fixed issue #20161: Stored Cross-Site Scripting (SXSS) in admin portal

gabrieljenik:


application/models/Box.php
```diff
@@ -29,10 +29,12 @@ public function rules()
         return array(
             array('url, title, ico, position, desc, page', 'required'),
             array('url', 'match', 'pattern' => '/(http:\/\/)?[a-zA-Z]([a-zA-Z0-9-_?&"\'=]\/?)*/'),
+            array('url', 'LSYii_Validators', 'isUrl' => true),
             array('position', 'numerical', 'integerOnly' => true),
             array('position', 'unique', 'message' => gT('Position {value} already exists.')),
             array('usergroup', 'numerical', 'integerOnly' => true, 'min' => -3),
             array('ico', 'match', 'pattern' => '/^[A-Za-z0-9_ \-]+$/u','message' => gT('Icon name must be a simple class name (alphanumeric, space, minus and underscore).')),
+            array('title', 'LSYii_Validators'),
```
Shnoulle, line 37: Unsure need to validate here ?

application/models/Box.php
```diff
@@ -29,10 +29,12 @@ public function rules()
         return array(
             array('url, title, ico, position, desc, page', 'required'),
             array('url', 'match', 'pattern' => '/(http:\/\/)?[a-zA-Z]([a-zA-Z0-9-_?&"\'=]\/?)*/'),
+            array('url', 'LSYii_Validators', 'isUrl' => true),
```
Shnoulle, line 32: Works for admin/labels/sa/view ?

Shnoulle:
I think we just need to encode it, since we don't use in HTML **and** encoded : no need to sanitize

gabrieljenik:
Last comments remain?

application/models/Box.php
gabrieljenik: I guess labels have their own set of rules, right?

application/models/Box.php
gabrieljenik: Why wouldn't we need?

gabrieljenik:
> I think we just need to encode it, since we don't use in HTML and encoded : no need to sanitize

Sanitize is always good. What's the harm?

Shnoulle:
> Sanitize is always good.

No ! 

1. Sanitize without informing user is really bad ! You sanitize username, but user save not sanitized username. Whentry to log in : send noty santized : broken.
2. There are hundred of way to sanitize : each need a good reason ! Why not allow only [a-z0-9] then ? Or any other sanitize way. You sanitize for what purpose ?

Shnoulle (approved):

Shnoulle:
@tiborpacalat since we show it encoded : do we really need to use htmlpurifier ?

gabrieljenik:
> 1. Sanitize without informing user is really bad

Not sure what you mean by inform the user... We apply `htmlpurifier` a lot during the code, specially on rules.

> You sanitize username, but user save not sanitized username.

The title is not a username.

I feel more comfortable sanitizing, but if you feel we shouldn't I will just remove it.

Shnoulle:
> Not sure what you mean by inform the user... We apply `htmlpurifier` a lot during the code, specially on rules.

Yes, and it's an UX issue here.

And more : it don't show with HTML editor …

PR #4362, opened 2025-07-11
Fixed issue #20089: HTML shown in notification after creating a user through User management

Shnoulle:
Dev: fix the documented useHtml options
Dev; use it for user managemlent
Dev: purify post value if useHtml is ON

### DIFF
--- a/application/controllers/AjaxAlertController.php
+++ b/application/controllers/AjaxAlertController.php
@@ -25,9 +25,12 @@ class AjaxAlertController extends LSBaseController
     {
         $request = Yii::app()->request;
         $customOptions = $request->getPost('customOptions', []);
-
         $translatedOptions = [];
-        $translatedOptions['text'] = CHtml::encode($request->getPost('message', 'message'));
+        if (empty($customOptions['useHtml'])) {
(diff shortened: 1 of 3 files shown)


application/controllers/AjaxAlertController.php
```diff
@@ -25,9 +25,12 @@ private function translateOptionsForWidget()
     {
         $request = Yii::app()->request;
         $customOptions = $request->getPost('customOptions', []);
-
         $translatedOptions = [];
-        $translatedOptions['text'] = CHtml::encode($request->getPost('message', 'message'));
+        if (empty($customOptions['useHtml'])) {
```
gabrieljenik, line 29: Isn't this the other way around? encode if html is expected?

Shnoulle: > Isn't this the other way around? encode if html is expected?

No , testing : i(s OK else : 

- useHTML is ON : purified (we have HTML inside)
- useHTML is OFF : encode, it's not HML

Seems logical

gabrieljenik: My bad

Shnoulle: No prob

PR #4364, opened 2025-07-15
LE-270: structural embedding

lajosarpad:
Dev:

### DIFF
--- a/application/models/services/embeds/BaseEmbed.php
+++ b/application/models/services/embeds/BaseEmbed.php
@@ -1,0 +1,103 @@
+<?php
+
+namespace LimeSurvey\Models\Services;
+
+abstract class BaseEmbed {
+    protected $width;
+    protected $height;
+    protected $src;
+    protected $structure;
+    public const EMBED_STRUCTURE_STANDARD = "Standard";
+    protected static $supportedEmbeds = null;
+
+    /**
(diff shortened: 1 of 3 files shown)


application/models/services/embeds/BaseEmbed.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Sets the height of the inner content and returns the instance
+     * @param int $height
+     * @return static
+     */
+    public function setHeight(int $height)
+    {
+        $this->height = $height;
+        return $this;
+    }
```
haythemch2, line 55: Thinking about the wrapper extensions later , it might be better to have a single point of control for the size of the survey.
What could be ideal , is to have the iframe defaulting to 100% for both width & height that way it will always fill its container , and let the wrapper be the single source of size & design ( borders , position etc... )
so instead of having setters for width & height , we can have a setWrapperOptions that sets array of options that  would be used in the abstract getWrapper for each embed type class.

PS: This can be adjusted later when implementing the first wrapper

haythemch2 (approved):

application/models/services/embeds/BaseEmbed.php
lajosarpad: I agree with these ideas, but let's do them later so people will have an embedder.

PR #4384, opened 2025-07-30
Fixed issue #20185: Answer Options Not Deleted When Changing Question Type to Multiple Choice

gabrieljenik:


application/models/Question.php
```diff
@@ -1658,6 +1671,25 @@ protected function removeInvalidSubquestions()
...
+    {
+        if ($this->getAllowAnswerOptions()) {
+            return;
+        }
+
+        // Remove answer options if the question's type doesn't allow answer options
+        $answerOptions = Answer::model()->findAll("qid=:qid", array("qid" => $this->qid));
+        if (!empty($answerOptions)) {
+            foreach ($answerOptions as $answerOption) {
+                $answerOption->delete();
```
Shnoulle, line 1688: :+1:

PR #4419, opened 2025-08-26
LE-293: email embed refactor to use clientside for image capture

haythemch2:
Dev:


application/models/services/embeds/EmailEmbed.php
```diff
@@ -29,79 +20,38 @@ protected function getWrapper(string $placeholder = "PLACEHOLDER")
...
-                </div>',
-                $surveyUrl,
-                $imageUrl,
-                $logoUrl
-            );
-        } catch (\Exception $e) {
-            return 'Error capturing survey preview: ' . $e->getMessage();
-        }
+        $filename = "email_embed_" . $surveyId . ".png";
+        $urlPrefix = "{$protocol}://{$host}/";
```
mohabmes, line 31: Maybe using `getSurveyUrl `from the Survey model would be better, as it handles different URL formats `urlFormat`.

PR #4446, opened 2025-09-19
LE-413: Implemented a fix for response importer to support fields with hashtag in their name

lajosarpad:
Dev:

### DIFF
--- a/application/helpers/admin/import_helper.php
+++ b/application/helpers/admin/import_helper.php
@@ -14,6 +14,7 @@
 
 use LimeSurvey\Helpers\questionHelper;
 use LimeSurvey\Models\Services\SurveyAccessModeService;
+use LimeSurvey\Models\Services\LSCDbCommandBuilder;
 
 /**
  * This function imports a LimeSurvey .lsg question group XML file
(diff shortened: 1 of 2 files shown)


lajosarpad:
> The class LSCDbCommandBuilder (rename to LSDbCommandBuilder) should be moved to \application\core\db and should be called by default from the Schema classes already present in that directory by adding the function createCommandBuilder() in the Custom schema classes there.
> 
> That way you don't have to adjust calls in the rest of the code and future code will automatically use the fix.

Thank you for that remark, the changes were applied as planned. Now the automated checks pass and there were some tests I performed which were successful. I asked team members to do further tests which will likely be done tomorrow

application/core/db/MssqlSchema.php
```diff
@@ -2,6 +2,11 @@
 
 class MssqlSchema extends CMssqlSchema
 {
+	protected $_tableNames=array();
```
c-schmitz: are  these still needed?

lajosarpad: @c-schmitz they were removed, auto-checks passed

PR #4454, opened 2025-09-26
Fixed issue #20277: Error when adding survey participant attributes to create

gabrieljenik:


Shnoulle (approved):
Sorry didn't see it ?

PR #4455, opened 2025-09-26
Dev #20273: Research if Limesurvey is using bootstrap-switch library so it can be removed from the code base

gabrieljenik:
Went over and removed it from the code base directly.
All code removed is supposedly not being used.

The package was not removed as it is being used by WhSwitch, which is not used in LS, but could be used by a plugin.


application/controllers/admin/Tokens.php
```diff
@@ -399,9 +399,6 @@ public function browse($iSurveyId, $limit = 50, $start = 0)
             $this->getController()->redirect(array("/admin/tokens/sa/index/surveyid/{$iSurveyId}"));
         }
 
-        // TODO: Why needed?
```
Shnoulle, line 402: LOL

PR #4462, opened 2025-10-02
CT-1301: Ensured that the survey name and the expiry show up dynamically

lajosarpad:
### **User description**
in the survey

Dev:

___

### **PR Type**
Enhancement

___

### **Description**
- Added dynamic survey name placeholder support

- Added survey expiry date placeholder support

- Fixed SID variable to use survey object

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Expression Manager"] --> B["SURVEYNAME Placeholder"]
  A --> C["EXPIRY Placeholder"]
  A --> D["Fixed SID Variable"]
  B --> E["Dynamic Survey Title"]
  C --> F["Dynamic Expiry Date"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>em_manager_helper.php</strong><dd><code>Enhanced placeholder support with dynamic survey data</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

application/helpers/expressions/em_manager_helper.php

<ul><li>Added <code>SURVEYNAME</code> placeholder mapping to survey title<br> <li> Added <code>EXPIRY</code> placeholder mapping to survey expiry date<br> <li> Fixed <code>SID</code> variable to use survey object instead of manager property</ul>

</details>

  </td>
  <td><a href="https://github.com/LimeSurvey/LimeSurvey/pull/4462/files#diff-fa8d35264aa59c85d37ca2d1756472996499f06fb346865594378102f5f7abc0">+13/-1</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

</details>

___

### DIFF
--- a/application/helpers/expressions/em_manager_helper.php
+++ b/application/helpers/expressions/em_manager_helper.php
@@ -4558,7 +4558,7 @@ class LimeExpressionManager
         $LEM->processedRelevance = false;
         $LEM->surveyOptions['hyperlinkSyntaxHighlighting'] = true;    // this will be temporary - should be reset in running survey
         $LEM->qid2exclusiveAuto = [];
-        self::resetTempVars();
+        //self::resetTempVars();
         $surveyinfo = (isset($LEM->sid) ? getSurveyInfo($LEM->sid) : null);
         if (isset($surveyinfo['assessments']) && $surveyinfo['assessments'] == 'Y') {
             $LEM->surveyOptions['assessments'] = true;


Shnoulle:
I think the issue with this fix : it takes only `SURVEYNAME` and `EXPIRY` but not the https://github.com/LimeSurvey/LimeSurvey/blob/d7cc248b5e0cb1abeea49bebab5f559a6cdaf6e5/application/helpers/expressions/em_manager_helper.php#L3367 other template replace KEYWORDS

https://github.com/LimeSurvey/LimeSurvey/blob/d7cc248b5e0cb1abeea49bebab5f559a6cdaf6e5/application/helpers/replacements_helper.php#L293
https://github.com/LimeSurvey/LimeSurvey/blob/d7cc248b5e0cb1abeea49bebab5f559a6cdaf6e5/application/helpers/replacements_helper.php#L321

Maybe an issue with  `$_surveyid = $_SESSION['LEMsid'];`  https://github.com/LimeSurvey/LimeSurvey/blob/d7cc248b5e0cb1abeea49bebab5f559a6cdaf6e5/application/helpers/replacements_helper.php#L86  ?

Shnoulle:
1. `templatereplace` function construct a big array of replacement and move it to `LimeExpressionManager::ProcessString` function
2. `LimeExpressionManager::ProcessString` sedn each element of this array in `LimeExpressionManager::updateReplacementFields`
3. `updateReplacementFields` do what you do for this 2 keywords : https://github.com/LimeSurvey/LimeSurvey/blob/d7cc248b5e0cb1abeea49bebab5f559a6cdaf6e5/application/helpers/expressions/em_manager_helper.php#L7892

The question here is why templatereplace didn't work at 1st call ? But work when force reset in 1st page …

lajosarpad:
> 1. `templatereplace` function construct a big array of replacement and move it to `LimeExpressionManager::ProcessString` function
> 2. `LimeExpressionManager::ProcessString` sedn each element of this array in `LimeExpressionManager::updateReplacementFields`
> 3. `updateReplacementFields` do what you do for this 2 keywords : https://github.com/LimeSurvey/LimeSurvey/blob/d7cc248b5e0cb1abeea49bebab5f559a6cdaf6e5/application/helpers/expressions/em_manager_helper.php#L7892
> 
> The question here is why templatereplace didn't work at 1st call ? But work when force reset in 1st page …

It is called twice if force refreshed, both from `setVariableAndTokenMappingsForExpressionManager`. And it is the second call that successfully hydrates the values. However, the second call's partial stack looks like this:

```
LimeExpressionManager::StartProcessingGroup(-1, false, $surveyid); // start on welcome page
```

```
                $LEM->setVariableAndTokenMappingsForExpressionManager($surveyid, $forceRefresh, $anonymized);
```

and since `$forceRefresh` defaulted to false, it will not get past this:

```
    public function setVariableAndTokenMappingsForExpressionManager($surveyid, $forceRefresh = false, $anonymized = false)
    {
        if (isset($_SESSION['LEMforceRefresh'])) {
            unset($_SESSION['LEMforceRefresh']);
            $forceRefresh = true;
        } elseif ($forceRefresh === false && !empty($this->knownVars) && ((!$this->sPreviewMode) || ($this->sPreviewMode === 'database') || ($this->sPreviewMode === 'logic'))) {
            return false;   // means that those variables have been cached and no changes needed
        }
```

to reach this:

```
        /* Add the core replacement before question code : needed if use it in equation , use SID to never send error */
        /* Added replacement can not be used in condition, only for replacement */
        templatereplace("{SID}");

```

Shnoulle:
> It is called twice if force refreshed, both from `setVariableAndTokenMappingsForExpressionManager`. And it is the second call that successfully hydrates the values. However, the second call's partial stack looks like this:

Yes, I understand this, but the question is "**why the 1st call didn't fill the value**" the second call didn't fille the value because it must be set at this time.

1st call must fill the value : it fills value you add but not when call `templatereplace("{SID}") :` what is the reason here ?

lajosarpad:
> > It is called twice if force refreshed, both from `setVariableAndTokenMappingsForExpressionManager`. And it is the second call that successfully hydrates the values. However, the second call's partial stack looks like this:
> 
> Yes, I understand this, but the question is "**why the 1st call didn't fill the value**" the second call didn't fille the value because it must be set at this time.
> 
> 1st call must fill the value : it fills value you add but not when call `templatereplace("{SID}") :` what is the reason here ?

It's because of this: https://github.com/LimeSurvey/LimeSurvey/blob/8829a0238e1da4b6dafe48eedadde783df27e8a8/application/helpers/expressions/em_manager_helper.php#L4561

Shnoulle:
> It's because of this:
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/8829a0238e1da4b6dafe48eedadde783df27e8a8/application/helpers/expressions/em_manager_helper.php#L4561

Maybe we need a solution to avoid it in some condition ?

lajosarpad:
> > It's because of this:
> > https://github.com/LimeSurvey/LimeSurvey/blob/8829a0238e1da4b6dafe48eedadde783df27e8a8/application/helpers/expressions/em_manager_helper.php#L4561
> 
> Maybe we need a solution to avoid it in some condition ?

The line seems to be unnecessary. If you look at the commit that created it five years ago, it was all about PSR-12 so it seems it was related to styling. However, if we are to remove this line instead of my initial solution, then we need to very thoroughly test it to confirm my hypothesis, as I am not 100% sure I am right here

Shnoulle:
> The line seems to be unnecessary. If you look at the commit that created it five years ago, it was all about PSR-12 so it seems it was related to styling. However, if we are to remove this line instead of my initial solution, then we need to very thoroughly test it to confirm my hypothesis, as I am not 100% sure I am right here

I think it's here to clean up some vars when there are a new page call. Else : you can have some false information saty in tempVars coming from a previous page (EM goes in SESSION).

But here : why we have 2 `StartProcessingPage` calls ? We need only one to show a page. Only the 1st time

Shnoulle:
Original commit for self::resetTempvars

https://github.com/LimeSurvey/LimeSurvey/commit/9c055e54b013833fb905b023787ae93291e5130f

I don't remind why i exatly add it, but since it was in `FinishProcessPublicPage` , i think we can remove it from `StartProcessingPage`

PR #4463, opened 2025-10-02
LE-86: include last modified date in survey list boxes

mohabmes:


c-schmitz (changes requested):
Please check the quodo comments - especially the one regarding the creationdate,.

c-schmitz (approved):

PR #4469, opened 2025-10-07
Fixed issue #20279: Error page didn't get theme option

Shnoulle:
Fixed issue #20261: Create 400 http error code exception for invalid sureyId

### DIFF
--- a/application/core/LSYii_Application.php
+++ b/application/core/LSYii_Application.php
@@ -64,6 +64,9 @@ class LSYii_Application extends CWebApplication
     /* @var integer| null the current userId for all action */
     private $currentUserId;
 
+    /* @var integer|null|false the current survey ID */
+    private static $surveyId;
+
     /**
      *
      * Initiates the application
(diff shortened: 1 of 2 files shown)


c-schmitz:
Looks fine to me - note: not every action has a survey ID, in that case the standard theme should be used.

Shnoulle:
> Looks fine to me - note: not every action has a survey ID, in that case the standard theme should be used.

Yes, it's the case here. See the issue `20279: Error page didn't get theme option` 

But before : if you send =`survey/index&sid=404` : template used was totally invalid (debug set show $sgid as null etc …).

And since i need a solution to get a valid survey ID …

Else : 
> Avoid using static properties for request-scoped data

I don't want to check survey multiple times here. One request only. Maybe we can use non static because `Survey::model()->findByPk($surveyId)` use static variable ?


> Improve caching to handle all states

cache different using $throwerror or not.
removing the static way (since findByPk is already static) : no issue, but we do the same action multiple times. And in majority : survey is set and is valid.

> Use validated integer for consistency

Since its already checked, why not, but needed (we are sure it's a integer or a ctype_digit)

application/core/LSYii_Application.php
```diff
@@ -595,4 +598,45 @@ private function setSessionByDB($aApplicationConfig)
...
+         }
+         $intSurveyId = intval($surveyId);
+         if (strval($intSurveyId) !== strval($surveyId)) {
+             if ($throwError) {
+                 throw new CHttpException(400, gT('Your request is invalid.'));
+             }
+             return false;
+        }
+        $surveyId = intval($surveyId);
+        /* surveyId is set and is an integer */
```
gabrieljenik, line 630: Maybe here add a validation in case intval returns zero or null?

Shnoulle: We have `if (!$surveyId) {` and `if (strval($intSurveyId) !== strval($surveyId)) {`

Sending `"0"` ?

Shnoulle: And we have `$survey = Survey::model()->findByPk($surveyId);` after

gabrieljenik: > And we have `$survey = Survey::model()->findByPk($surveyId);` after

That's exactly why. I mean there is no validation in between the intval and the survey find. Just want to avoid a dump on the find... but honestly not sure if I can think of a situation now... it was more like a reassure validation... like a sanity check.

Please just take this comment as a suggestion, no need to follow it!

Shnoulle: But we have : 
```
        if (!$surveyId) {
             return false;
         }
```

0, "" or null are not tested

PR #4646, opened 2025-12-23
New feature: New audit log plugin setting to disallow non-superadmins to disable AuditLog on individual surveys

seccanj:
Dev: The base version of the AuditLog plugin allows any survey designer to disable the audit log for their survey, which may be undesirable. This new feature adds a plugin configuration parameter that allows platform administrators to decide whether to disable this ability for non superadmin users.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable setting to control whether non-superadmin survey admins may disable per-survey auditing.
  * Survey settings UI now reflects that policy—auditing control is disabled when overrides are not allowed and shows permission hints.

* **Bug Fixes / Enforcement**
  * Prevents unauthorized attempts to disable auditing at display and save time, ensuring policy is consistently enforced.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/application/core/plugins/AuditLog/AuditLog.php
+++ b/application/core/plugins/AuditLog/AuditLog.php
@@ -650,10 +650,10 @@ class AuditLog extends \LimeSurvey\PluginManager\PluginBase
         // Disable the control for non-superadmin users to prevent them from disabling audit
         $oCurrentUser = $this->api->getCurrentUser();
         $allowNonSuperDisable = isset($pluginsettings['AuditLog_AllowNonSuperadminDisable']['current']) && $pluginsettings['AuditLog_AllowNonSuperadminDisable']['current'] == 1;
-        if ($oCurrentUser !== null && !$allowNonSuperDisable && !Permission::model()->hasGlobalPermission('superadmin', 'read', $oCurrentUser->uid)) {
+        if ($oCurrentUser && !$allowNonSuperDisable && !Permission::model()->hasGlobalPermission('superadmin', 'read', $oCurrentUser->uid)) {
             $auditingSetting['htmlOptions'] = array('disabled' => 'disabled');
             $auditingSetting['help'] = gT('Only superadmins can disable the audit log for a survey.');
-        } elseif ($oCurrentUser !== null && $allowNonSuperDisable && !Permission::model()->hasGlobalPermission('superadmin', 'read', $oCurrentUser->uid)) {
+        } elseif ($oCurrentUser && $allowNonSuperDisable && !Permission::model()->hasGlobalPermission('superadmin', 'read', $oCurrentUser->uid)) {
             // If non-superadmins are allowed to disable auditing, provide a hint about permissions
             $auditingSetting['help'] = gT('You can disable the audit log for surveys where you have survey settings update permission.');
         }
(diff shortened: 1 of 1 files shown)


application/core/plugins/AuditLog/AuditLog.php
```diff
@@ -627,19 +634,34 @@ public function beforeSurveySettings()
...
+        // Disable the control for non-superadmin users to prevent them from disabling audit
+        $oCurrentUser = $this->api->getCurrentUser();
+        $allowNonSuperDisable = isset($pluginsettings['AuditLog_AllowNonSuperadminDisable']['current']) && $pluginsettings['AuditLog_AllowNonSuperadminDisable']['current'] == 1;
+        if ($oCurrentUser !== null && !$allowNonSuperDisable && !Permission::model()->hasGlobalPermission('superadmin', 'read', $oCurrentUser->uid)) {
+            $auditingSetting['htmlOptions'] = array('disabled' => 'disabled');
+            $auditingSetting['help'] = gT('Only superadmins can disable the audit log for a survey.');
+        } elseif ($oCurrentUser !== null && $allowNonSuperDisable && !Permission::model()->hasGlobalPermission('superadmin', 'read', $oCurrentUser->uid)) {
+            // If non-superadmins are allowed to disable auditing, provide a hint about permissions
+            $auditingSetting['help'] = gT('You can disable the audit log for surveys where you have survey settings update permission.');
+        }
```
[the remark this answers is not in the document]
seccanj: The beforeSurveySettingsSave method is not called when plugin-only settings are saved. It is called only when core survey settings are saved. Instead, the newSurveySettings method is called when plugin-only settings are saved. I should then implement the authorization checks inside newSurveySettings.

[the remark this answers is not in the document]
seccanj: Authorization check implemented

Shnoulle:
I'm still for a new permission (with default set in admin GUI)

application/core/plugins/AuditLog/AuditLog.php
```diff
@@ -656,28 +634,66 @@ public function beforeSurveySettings()
...
+            'tab' => 'notification', // @todo: Setting no used yet
+            'category' => 'Auditing for person-related data', // @todo: Setting no used yet
+            'label' => 'Audit log for this survey:',
+            'current' => $this->get('auditing', 'Survey', $event->get('survey'))
+        );
+
+        // Disable the control for non-superadmin users to prevent them from disabling audit
+        $oCurrentUser = $this->api->getCurrentUser();
+        $allowNonSuperDisable = isset($pluginsettings['AuditLog_AllowNonSuperadminDisable']['current']) && $pluginsettings['AuditLog_AllowNonSuperadminDisable']['current'] == 1;
+        if ($oCurrentUser && !$allowNonSuperDisable && !Permission::model()->hasGlobalPermission('superadmin', 'read', $oCurrentUser->uid)) {
```
gabrieljenik, line 653: what if $oCurrentUser is null? what is the expected behaviour?

application/core/plugins/AuditLog/AuditLog.php
```diff
@@ -656,28 +634,66 @@ public function beforeSurveySettings()
...
+            // If non-superadmins are allowed to disable auditing, provide a hint about permissions
+            $auditingSetting['help'] = gT('You can disable the audit log for surveys where you have survey settings update permission.');
+        }
+
         $event->set("surveysettings.{$this->id}", array(
             'name' => get_class($this),
             'settings' => array(
-                'auditing' => array(
-                    'type' => 'select',
-                    'options' => array(0 => 'No',
```
gabrieljenik, line 664: Is this expected?

seccanj: The configuration has been moved from inline to the $auditingSetting variable.

application/core/plugins/AuditLog/AuditLog.php
seccanj: Is this a possibility? I guess in this case changing the setting should be disabled.

PR #4651, opened 2026-01-05
New feature #LX-8: Add Quota Exit field to responses table to record the matched quota

gabrieljenik:
## Release Notes

* **New Features**
  * Added a “Save quota exit” setting for surveys, including activation, general settings, inherited group settings, and API access.
  * Matched quota information can now be recorded with survey responses when enabled.
  * Quota exit values are available in data entry, response APIs, and SPSS and STATA exports.
  * Quota names are displayed when matching quota information is available.
  * Quota exit values are preserved when importing archived responses.
  * Existing surveys are updated automatically to support the new setting and response field.


Shnoulle:
An idea about adding new response columns are more like `datstamp` for example.

If a survey is activated, there is no way to update it and set it to No. It can be activated in survey activation or before activation.

Then 

1. No need to check all existing surveys (i have a more than 2 500 activated surveys; the DB is on another server, updating can break due to timing issues).
2. It's clearly a new feature
3. With the new way to get old tables, it takes less than 5 minutes to add to an old survey
4. When no quota: adding a column can break table size.

application/helpers/export_helper.php
```diff
@@ -590,6 +590,10 @@ function SPSSFieldMap($iSurveyID, $prefix = 'V', $sLanguage = '')
                 $fieldtype = 'A';
                 $val_size = 31;
                 break;
+            case 'quota_exit';
+                $fieldtype = 'F';
+                $val_size = 7;
```
Shnoulle, line 597: <s>No translation when exporting as text ? Just 1,2,3,4 ?</s>

Shnoulle: OK, it's the quota ID here ?
Export quota name when export as text ( ID - {name}) ?

application/helpers/update/updates/Update_641.php
```diff
@@ -0,0 +1,32 @@
...
+            $responseTableName = "{{survey_" . $surveyId . "}}";
+            if (!tableExists($responseTableName)) {
+                continue;
+            }
+
+            \Yii::app()->db->createCommand()->addColumn(
+                $responseTableName,
+                'quota_exit',
+                'integer'
+            );
```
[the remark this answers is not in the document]
c-schmitz: I agree here. Some tables might have a problems with updates because they are already on the maximum size.

application/helpers/export_helper.php
c-schmitz: I agree - adding the quota name on export would be great, but lets do that later and lets not store it in the same field.

application/helpers/update/updates/Update_641.php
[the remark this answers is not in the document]
c-schmitz: You might also want to check that the whole survey does not fail if the quota_exist field is still missing,

c-schmitz (changes requested):

application/helpers/export_helper.php
Shnoulle: :+1:

Shnoulle: (Like answers: if you delete an answer, you get no text when you export.)

Shnoulle:
> I agree here. Some tables might have a problems with updates because they are already on the maximum size.

You mean add it as an option? Like datestamp?

application/helpers/update/updates/Update_641.php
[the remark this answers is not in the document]
gabrieljenik: > You might also want to check that the whole survey does not fail if the quota_exist field is still missing,

You mean that when saving responses, we can't expect the field to be there?
So, we will need to verify that before saving?

gabrieljenik:
> 1\. No need to check all existing surveys (i have a more than 2 500 activated surveys; the DB is on another server, updating can break due to timing issues).
It's clearly a new feature
> 2\. It's clearly a new feature

So, on the new codebase, when saving a quota exit, how do we know if the field is there or not? You mentioned an option. That could work. Still, I guess that would always be true for all future surveys, right? How wouldn't want to track the exit quota?

Shnoulle:
> So, on the new codebase, when saving a quota exit, how do we know if the field is there or not? You mentioned an option. That could work. 

`$oSurvey->options->savequotaid === "Y"` or in https://github.com/LimeSurvey/LimeSurvey/blob/31deac7f7e069e102fb1866569bfea1a2652f8ba/application/models/Survey.php#L493 

1. `array('savequotaid', 'in', 'range' => array('Y', 'N', 'I'), 'allowEmpty' => true),`
2. `public function getIsSaveQuotaId()

In SurveyActivator check if `survey->IsSaveQuotaId` etc ...

> Still, I guess that would always be true for all future surveys, right? How wouldn't want to track the exit quota?

Why always be true? There are some surveys that don't need it, and you have a column less.

gabrieljenik:
@c-schmitz Looking forward for your directions in regard to how identify if a quota_id should be saved (option, look for the field, ...)

@Shnoulle 
>> Still, I guess that would always be true for all future surveys, right? How wouldn't want to track the exit quota?
> Why always be true? There are some surveys that don't need it, and you have a column less.

I would say 99% of the people who uses quotas will say yes. One column doesn't make a difference (imo)

Shnoulle:
> I would say 99% of the people who uses quotas will say yes. One column doesn't make a difference (imo)

It can be set by default (inherit) to true. But with a new setting, even the 1% are happy, meaning 100% of user are happy.

**New settings**

- 100% of user are happy
- No issue when update/upgarde with 10 000 active surveys.

**No setting**

- 99% of user are happy, 1% are unhappy
- Potential issue when update/upgrade with 10 000 active surveys.

c-schmitz:
Denis meant that this could be an optional field - if it exists, great, so use it. 
If it does not, because the table could not be updated for some weird reason, handle it and ignore it.

gabrieljenik:
Ready for reviewal!

application/core/plugins/expressionFixedDbVar/expressionFixedDbVar.php
```diff
@@ -82,6 +82,12 @@ class expressionFixedDbVar extends PluginBase
             'default' => '0',
             'column' => 'ipaddr',
         ),
+        'QUOTA_EXIT' => array(
```
Shnoulle, line 85: 😄

application/helpers/common_helper.php
```diff
@@ -1163,6 +1163,13 @@ function getExtendedAnswer($iSurveyID, $sFieldCode, $sValue, $sLanguage, $questi
                 $sValue = convertDateTimeFormat($sValue, "Y-m-d H:i:s", $dateformatdetails['phpdate'] . ' H:i:s');
             }
             break;
+        case 'quota_exit':
```
Shnoulle, line 1194: 👍

application/views/admin/dataentry/caption_view.php
```diff
@@ -72,6 +72,20 @@ function activateSubmit(me)
...
 
+            if ($thissurvey['savequotaexit'] == "Y") //Give quota exit field
+            { ?>
+                <tr>
+                <td valign='top' width='1%'></td>
+                <td valign='top' align='right' width='30%'><strong>
+                <?php echo gT("Quota exit",'html',$sDataEntryLanguage); ?>:</strong></td>
+                <td valign='top'  align='left' style='padding-left: 20px'>
+                <input type='text' name='quota_exit' value='NULL' />
+                </td>
```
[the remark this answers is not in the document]
Shnoulle: I'm OK here : better to show an empty string (save as null value if needed) then NULL

Shnoulle (approved):
Except the null value, seems OK

gabrieljenik:
> **Actionable comments posted: 0**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)

This is an old upgrade script. Doesn't apply.

c-schmitz:
@gabrieljenik Please resolve the conflicts and update the PR.

gabrieljenik:
> @gabrieljenik Please resolve the conflicts and update the PR.

Done.
@tiborpacalat Although quotas doesn't seem to be impacted by the new editor, still, I would keep an eye on it

c-schmitz:
@coderabbitai full review

application/helpers/admin/survey_helper.php
```diff
@@ -39,6 +39,7 @@ function getSurveyDefaultSettings()
     'datestamp'                => 'N',
     'ipaddr'                   => 'N',
     'refurl'                   => 'N',
+    'savequotaexit'            => 'N',
```
Shnoulle, line 42: It's not I here ?

Shnoulle: Same for a lot of other

application/helpers/update/updates/Update_715.php
```diff
@@ -0,0 +1,54 @@
...
+
+            \Yii::app()->db->createCommand()->addColumn(
+                $responseTableName,
+                'quota_exit',
+                'integer'
+            );
+        }
+
+        // Add to surveys table if not exists
+        $surveysTable = $this->db->schema->getTable('{{surveys}}');
```
Shnoulle: My opinoon about upgrade

1. Add savequotaexit with default I to Surveys
2. Add savequotaexit with default I to SurveyGroups
3. Update savequotaexit to N for SurveyGroup=0
4. Update savequotaexit to N for Surveys where active=='Y'

PR #4669, opened 2026-01-15
A11y central participant management four fixes

anblikauditteam:
1.fix Incorrect heading level announced.
2.fix "Save" button is not focusable.
3.verify Focus indicator is not visible.
Dev:

## Summary by CodeRabbit

* **New Features**
  * Improved keyboard accessibility: Save button in modals is now keyboard-focusable.

* **Bug Fixes**
  * Modal headers updated for more consistent and reliable title rendering and styling.


application/views/layouts/partial_modals/modal_header.php
```diff
@@ -3,6 +3,6 @@
 ?>
 
 <div class="modal-header">
-    <h2 class="h1 modal-title"><?php echo $modalTitle; ?></h2>
+    <h2 class="modal-title h1"><?php echo $modalTitle; ?></h1>
```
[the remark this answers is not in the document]
anblikauditteam: fixed

application/views/layouts/partial_modals/modal_footer_cancelsave.php
```diff
@@ -6,7 +6,7 @@
         <?php
         eT("Cancel"); ?>
     </button>
-    <a role="button" class="btn btn-primary btn-ok">
+    <a role="button" class="btn btn-primary btn-ok" tabindex="0">
```
ptelu, line 9: button is alread focusable why do we need a tabindex

ptelu (changes requested):

application/views/layouts/partial_modals/modal_footer_cancelsave.php
anblikauditteam: “We added tabindex="0" because the <a> element does not have an href, so it is not focusable by default. Since it is being used like a button (and not as a link), it wouldn’t receive keyboard focus. Adding tabindex="0" makes it focusable via the Tab key and ensures accessibility for keyboard users.”

PR #4671, opened 2026-01-19
Fixed issue #20404: Exclusive option not placed correct when using ra…

Shnoulle:
…ndomized order

Dev: set as original position

Dev:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of excluded subquestions during randomization: multiple excluded items (e.g., semicolon-separated titles) are now recognized and each restored to its correct position within the group.
  * Rewrote extraction and reinsertion logic to support multiple exclusions while retaining compatibility with single-item exclusions, improving ordering accuracy after randomization.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/application/models/services/QuestionOrderingService/RandomizerHelper.php
+++ b/application/models/services/QuestionOrderingService/RandomizerHelper.php
@@ -89,13 +89,13 @@ class RandomizerHelper
         foreach ($groupedSubquestions as $scaleId => &$scaleArray) {
             $scaleArray = \ls\mersenne\shuffle($scaleArray);
         }
-
         // Reinsert excluded subquestion if needed
         if ($excludedSubquestion !== null) {
             $scaleId = $excludedSubquestion->scale_id;
+
             array_splice(
                 $groupedSubquestions[$scaleId],
-                ($excludedSubquestion->question_order - 1),
+                ($excludedSubquestion->question_order),
                 0,
                 [$excludedSubquestion]
             );
--- a/application/models/services/QuestionOrderingService/RandomizerHelper.php
+++ b/application/models/services/QuestionOrderingService/RandomizerHelper.php
@@ -69,7 +69,7 @@ class RandomizerHelper
         // Check for excluded subquestion before randomization
         /* @param string|null $excludeAllOthers */
         $excludeAllOthers = $question->getQuestionAttribute('exclude_all_others');
-        $excludedSubquestion = null;
+        $excludedSubquestions = [];
 
         if (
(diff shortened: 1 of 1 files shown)


Shnoulle:
I search the code update where the question_order starts at 0 and not at 1.

Shnoulle:
1. I fix the issue with multiple exclusoive oiption
2. I fix issue with different question_order start (or if there are issue in question_order)

c-schmitz (approved):

tiborpacalat:
This can be closed as discussed with Carsten and Denis because this PR https://github.com/LimeSurvey/LimeSurvey/pull/4613 resolves this issue and more.

Shnoulle:
> This can be closed as discussed with Carsten and Denis because this PR #4613 resolves this issue and more.

You test it ? I just test : it's not the case with [limesurvey_survey_MultipleChoiceRandomOrderExcludeAllOther-2.lss](https://bugs.limesurvey.org/file_download.php?file_id=19275&type=bug)

My opinion still

1. Must merge this one in master
2. Must create a totally new PR for develop-minor

Capture d’écran du 2026-04-02 09-34-08
Capture d’écran du 2026-04-02 09-34-37

Shnoulle:
ping @c-schmitz

Shnoulle:
For develop-minor :  add "Exclusive option stay at position" true by default can be great. Then we have : can set Exclusive option manually in the fixed order, and leave some other with random positions OR exclusive oprtion are added automatically.

tiborpacalat:
As I spoke to Denis: I remembered why I waited for the other PR to be merged in first. The exclusive option even if it is "Kept at the position" is moved one answer up. So this still needs to be fixed, but I would say on top of the other solution. I wouldn't introduce any new options. "Keep codes at position" should be used for all answers and exclusive option to keep them at positions while randomizing the options.

If exclusive option issue that I currently in production is not so critical (imo it is not), I would wait until we publish LS7 and then fix it on top of new code.

Shnoulle:
> If exclusive option issue that I currently in production is not so critical (imo it is not), I would wait until we publish LS7 and then fix it on top of new code.

You mean (for develop-minor, I do not speak for master) exclusive options are randomized if they are not in the "Keep codes at position" list?

tiborpacalat:
No, this is not what I said. And no, they are not randomized if they are not in the "Keep codes at position". But we should change it, to be like that and for everything which needs to be at same position use the new setting "Keep codes at position".

Shnoulle:
This commit is NOT for v7. Only for v6.

If another issue in v7 : need a totally different fix

PR #4702, opened 2026-02-05
Fixed issue #20416: PHP8.4 + dbug : unable to upload CSV file

gabrieljenik:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV and TSV import handling for escaped quotes, separators, and special characters.
  * Ensured participant, token, response, and user data imports parse escaped values consistently.
  * Improved delimiter detection for files containing escaped content, reducing parsing errors and preserving imported data more accurately.


Shnoulle (approved):

c-schmitz (changes requested):
THis does not look right.
`You put the default escape character as '\\' , when imho it should be "\\"`

Shnoulle (changes requested):
Arg @c-schmitz have right here !

https://www.php.net/manual/en/function.fgetcsv.php : use `"\\"`

gabrieljenik:
> THis does not look right. `You put the default escape character as '\\' , when imho it should be "\\"`

Nice catch

gabrieljenik:
> THis does not look right. `You put the default escape character as '\\' , when imho it should be "\\"`

'\\' is the same as "\\".
https://3v4l.org/GcWh0

Had my doubts, so I checked here and seems '\\' is ok for specifying a backslash.
https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.single
`To specify a literal backslash, double it (\\)`

In PHP doc the signature says it is using "\\".
https://www.php.net/manual/en/function.fgetcsv.php

Still, on the C implementation, is using '\\'.
https://github.com/php/php-src/blob/php-8.4.17/ext/standard/file.c#L1674

You let me know if you prefer to keep on going with the swap from '\\' to "\\"

c-schmitz:
Sorry, I don't understand. What are you saying? 
Yes, you can use `'/'`, but currently you are using `'//'`

`It does not matter whether  you use "//" or '/' , just '//' is wrong.`

gabrieljenik:
> Sorry, I don't understand. What are you saying?

I am sorry, the editor escaped my backslash. Should be...

> THis does not look right. You put the default escape character as '\\' , when imho it should be "\\"

`'\\'` is the same as `"\\"`.
https://3v4l.org/GcWh0

Shnoulle:
> In PHP doc the signature says it is using `"\\"`, https://www.php.net/manual/en/function.fgetcsv.php

I'm for following exactly the PHP manual. Less questions for the next developer.

c-schmitz:
I see. Did not know that in single quote strings `\\` and ` \'`  are still escaped.
I agree with Denis here to use double quotes as in the manual.

gabrieljenik:
Ok, let me add something else.

PHPDoc says to use double quotes.

Still, on the C implementation, is using single quotes.
https://github.com/php/php-src/blob/php-8.4.17/ext/standard/file.c#L1674

And the VSCode extension uses single quotes as well
image

Having said that, let me know if you prefer to use single quotes or double.

Thanks

Shnoulle:
> > In PHP doc the signature says it is using `"\\"`, https://www.php.net/manual/en/function.fgetcsv.php
> 
> I'm for following exactly the PHP manual. Less questions for the next developer.

Sorry for this : [fputcsv detail on php.net](https://www.php.net/manual/fr/function.fputcsv.php) use , `"\t",  "\"", "\\"` ,  `"\""` for enclosure.

@c-schmitz did we need to move to  `"\""` in place of `'"'` ?

Shnoulle:
> @c-schmitz did we need to move to `"\""` in place of `'"'` ?

ping @c-schmitz then we can remove the discussion ;)

c-schmitz:
@copilot resolve the merge conflicts in this pull request

c-schmitz (approved):

Shnoulle (approved):

PR #4732, opened 2026-03-02
Fixed issue #20372: Incorrect compatibility set when installing question theme

c-schmitz:
See https://bugs.limesurvey.org/view.php?id=20372

Fix updates only compatibility tag if not existing and fixes the duplicate entry problem.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved legacy theme XML conversion by handling custom attributes more safely, preventing duplicate attribute blocks during tag renaming.
  * Updated compatibility version handling to be additive and duplicate-safe, adding version **5.0** only when the compatibility node/version is missing.
  * Ensured the theme conversion used during the 4.25 update follows the improved legacy conversion behavior.


application/models/QuestionTheme.php
```diff
@@ -894,26 +894,34 @@ public static function convertLS3toLS5($sXMLDirectoryPath)
...
-        ) {
-            $oThemeConfig->compatibility->version = '5.0';
+        // set compatibility version only if missing or below 5.0
+        $shouldUpdateCompatibility = false;
+        if (isset($oThemeConfig->compatibility)) {
+            $existingVersions = [];
+            foreach ($oThemeConfig->compatibility->version as $ver) {
+                $existingVersions[] = (string)$ver;
+            }
+            // If no version or any version < 5.0, update
```
Shnoulle: No version mean 3.X ? I'm unsure we can set oit to 6.

See https://github.com/tpartner/ 

https://github.com/tpartner/LimeSurvey-Card-Sorting-3.x VS https://github.com/tpartner/LimeSurvey-Card-Sorting-5x-6x

application/models/QuestionTheme.php
```diff
@@ -894,26 +894,34 @@ public static function convertLS3toLS5($sXMLDirectoryPath)
...
+                    if (version_compare($ver, '5.0', '<')) {
+                        $shouldUpdateCompatibility = true;
+                        break;
+                    }
+                }
+            }
+            if ($shouldUpdateCompatibility) {
+                // Remove all <version> children
+                unset($oThemeConfig->compatibility->version);
+                $oThemeConfig->compatibility->addChild('version', '5.0');
```
Shnoulle: If we care about 5.X version : why leave it ? Just remove ?

Shnoulle (commented):
V5 !== V6

application/models/QuestionTheme.php
c-schmitz: This function is about converting question thesem from 3 to 5. 
THe version does not have to necessarily follow the main version,  5er themes are still compatible in v6.

application/models/QuestionTheme.php
c-schmitz: see comment above.

application/models/QuestionTheme.php
Shnoulle: My opinion: 3.X and 6.X theme versions can be incompatible.

We move from `<custom_attributes>` to `<attributes>` : https://gitlab.com/SondagesPro/QuestionTheme/arrayTextArea/-/commit/62382bb300446003107ff49d6e23e25e420e8b92

[5.X accept both](https://gitlab.com/SondagesPro/QuestionTheme/MapIgn/-/commit/67cbccd6cf4862937e2573badd1435f93a3b4475), 6.X didn't allow `custom_attributes` 

Then : you are not sure the theme are 6.X compatible.

application/models/QuestionTheme.php
c-schmitz: I changed it so the existing behaviour is preserved.
It only fixes the duplicate tags that were added in error previously.

c-schmitz:
@coderabbitai full review

Shnoulle (approved):
Fix is OK.
But I'm still upset about force not compatible QuestionTheme.
We move from BS3 to BS5 : QuestionTheme can be broken in BS5 but work with BS3

application/models/QuestionTheme.php
```diff
@@ -898,24 +904,18 @@ public static function convertLS3toLS5($sXMLDirectoryPath)
...
-            $compatibility = $oThemeConfig->addChild('compatibility');
-            $compatibility->addChild('version');
-            $oThemeConfig->compatibility->version = '5.0';
+        // Ensure '5.0' is listed in <compatibility>, adding it if absent.
+        // Never create a duplicate <compatibility> element.
+        $aExistingVersions = [];
+        foreach ($oThemeConfig->compatibility->version as $oVersion) {
+            $aExistingVersions[] = (string) $oVersion;
+        }
+        if (!in_array('5.0', $aExistingVersions, true)) {
```
Shnoulle: I use <compatibility>5</compatibility> Why use 5.0

And maybe we need to check 5.0, 5 etc ...

Shnoulle:
We are in 7.X now, this is still needed ? We don't need to update a `<compatibilty>7</compatibilty>` question theme

c-schmitz:
It would still mangle older themes that are practically still compatible.
You just add the compatiblity tag and the rest is done by the automatic conversion.
You can test with the attached file from https://bugs.limesurvey.org/view.php?id=20372

Shnoulle (commented):
I update my comment : if a QuetsionTheme have 
```
<compatibility>
<version>6</version>
<version>7</version>
</compatibility>
```

We do not need to add `<version>5.0</version>`

Shnoulle:
>  You can test with the attached file from https://bugs.limesurvey.org/view.php?id=20372

I have (an it's great) : _Le thème de la question n'est pas compatible avec votre version de LimeSurvey._

But this a 7 compatible survey : `<version>5.0</version>` is added and not needed

(apiVersion is done to allow multiple compatible system apiVersion:3 OK for 3,4 and 5. It was the original ID, but broken now).

c-schmitz:
I understand. So you would rather see that the update is more generic, converts/removes old attributes and that's it?

c-schmitz:
Anyway, it fixes the bug and does not introduce new ones, so I will leave it at this.

Shnoulle:
> Anyway, it fixes the bug and does not introduce new ones, so I will leave it at this.

I add 2 plugin zip file

> I understand. So you would rather see that the update is more generic, converts/removes old attributes and that's it?

No, I'm for "don't set it as compatible; maybe it's not the case".

c-schmitz:
I will check again.

c-schmitz:
Ok, now compatiiblity tag is only set if none exists. Otherwise it is ignored..

c-schmitz:
@tiborpacalat Should be fine now. Please re-test.

application/models/QuestionTheme.php
```diff
@@ -898,24 +910,9 @@ public static function convertLS3toLS5($sXMLDirectoryPath)
...
-        } elseif (
-            $oThemeConfig->compatibility->version
-            && count($oThemeConfig->compatibility->version) === 1
-        ) {
-            $oThemeConfig->compatibility->version = '5.0';
-        } else {
-            $compatibility = $oThemeConfig->addChild('compatibility');
-            $compatibility->addChild('version');
-            $oThemeConfig->compatibility->version = '5.0';
+        // Only add a <compatibility> block if none exists at all.
```
Shnoulle, line 913: 👍

Shnoulle (approved):

PR #4771, opened 2026-03-13
Fixed issue #20457: tableExists helper function returns outdated result due to cache.

35grain:
Fixed issue #20457: https://bugs.limesurvey.org/view.php?id=20457

The fix makes use of Yii::app()->db->schema->getTable(), where the second parameter is a flag for whether to clear schema cache for the given table. https://www.yiiframework.com/doc/api/1.1/CDbSchema#getTable-detail

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database table existence checks for more consistent detection.
  * Fixed repeated survey deactivations so each action receives a fresh, unique archive timestamp, including actions performed within the same second.
  * Improved archive naming reliability to prevent collisions during rapid successive deactivations.
  * Improved response recovery by using the correct archived table settings.
  * Improved deactivation handling by clearing cached dates and safely handling missing user identifiers.


Shnoulle:
This disables all DB cache. 

Unsure, maybe it's the best solution. 

Else : we refresh ALL table cache each time , maybe just force reset after activation, table creation.

Something like `tableExists($sTableName, $reset = true)` ?

Ping @c-schmitz

c-schmitz (approved):

c-schmitz (changes requested):
I agree to Denis. Probably it would be best to catch where the table is created and invalidate the cache there.

Shnoulle:
> I agree to Denis. Probably it would be best to catch where the table is created and invalidate the cache there.

I think too using `App()->db->schema->getTable()` in place of `App()->db->schema->getTableNames()` is really better.

With `getTable` check only one table, with `getTableNames` : list all tables.

Maybe we need to check if getTable 10 times without cache is better than  getTableNames with cache and reset cache one time.

35grain:
Yeah, great points and it seems that in Token::createTable, for example, the schema is actually already refreshed via getTable, but somehow still causes the described issue.
https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Token.php#L228

Yii v2 has refreshTableSchema() which would be really helpful here: https://www.yiiframework.com/doc/api/2.0/yii-db-schema#refreshTableSchema()-detail.

Shnoulle:
> Yeah, great points and it seems that in Token::createTable, for example, the schema is actually already refreshed via getTable, but somehow still causes the described issue. https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/Token.php#L228

Seems great to have this solution everywhere :)

35grain:
So I decided to fiddle around and found something interesting.. 
Yii::app()->db->schema->getTableNames() returns an expired value whereas Yii::app()->db->schema->getTable($sTableName) gives the correct result. So I will simply refactor to not use the refresh parameter in tableExists.

Shnoulle:
> So I decided to fiddle around and found something interesting.. Yii::app()->db->schema->getTableNames() returns an expired value whereas Yii::app()->db->schema->getTable($sTableName) gives the correct result. So I will simply refactor to not use the refresh parameter in tableExists.

I think can happen when 

- table didin't exist
- code check if table exist
- code create table but do not refresh globally
- code check if table exist

I think getTable cache return NULL in this case, then : check if exist.

Shnoulle (approved):

c-schmitz:
Since this seems to break some tests, I think something is still wrong here.

35grain:
Highly doubt it.. perhaps the tests are wrong as they worked with cached data before?

Shnoulle:
> Highly doubt it.. perhaps the tests are wrong as they worked with cached data before?

Maybe but must be fixed :)

> Base table or view already exists: 1050 Table 'lime_old_survey_969899_20260318083415' already exists

seems relevant

tiborpacalat:
@35grain I see that there is no update regarding required test fixes. Any plans to still tackle this?

35grain:
The initial patch changed tableExists() to use getTable() !== null instead of in_array(..., getTableNames()), eliminating a stale schema cache. This has exposed several pre-existing bugs in the deactivation/activation restore flow that had been masked by deactivate() exiting early when tableExists() falsely returned false.

Bugs fixed:

- archiveTable() model reuse (SurveyDeactivate.php): The injected ArchivedTableSettings model instance was reused across calls. After the first save(), Yii sets isNewRecord = false on the instance, causing all subsequent saves to silently run UPDATE WHERE id IS NULL (0 rows affected) instead of INSERT. Fixed by always instantiating a fresh model via new $modelClass().

- str_replace bug in recoverSurveyResponses (import_helper.php): The tbl_name used to look up archived encryption metadata was passed through str_replace('old_survey', 'old_tokens', ...), converting the response archive name into a token archive name. The lookup always returned null, so encryption state was never applied during restore. Removed the erroneous str_replace.

- Siddate collision (SurveyDeactivate.php): getSiddate() cached the timestamp for the object's lifetime, so reusing the same SurveyDeactivate instance for a second deactivation reused the first deactivation's timestamp. Switched to microsecond precision (YmdHisu) and reset the cache at the start of each deactivate() call.

- Null user_id (SurveyDeactivate.php): App()->user->getId() returns null when no user is authenticated (e.g. in tests). ArchivedTableSettings requires user_id to be a non-null integer, so saves failed silently. Added ?? 0 fallback.

- Manual DROP TABLE in test (ImportResponsesTest.php): The test was manually dropping the survey table between the second deactivation and activation. Now that deactivation runs fully and renames the table, this DROP failed. Removed it.

application/helpers/common_helper.php
```diff
@@ -2876,7 +2876,7 @@ function isCaptchaEnabled($screen, $captchamode = '')
 function tableExists($sTableName)
 {
     $sTableName = Yii::app()->db->tablePrefix . str_replace(array('{', '}'), array('', ''), $sTableName);
-    return in_array($sTableName, Yii::app()->db->schema->getTableNames());
+    return Yii::app()->db->schema->getTable($sTableName) !== null;
```
c-schmitz, line 2879: @Shnoulle This might be a huge problem. Afaik this is much slower than the previous solution.

c-schmitz: Why not clear the schema cache once after deactivation instead of slowing down the whole application.

Shnoulle: > @Shnoulle This might be a huge problem. Afaik this is much slower than the previous solution.

No, you ask for one table only, before you ask for ALL tables.

I think it can improve speed if you have a lot of table.

Shnoulle: PS : https://www.yiiframework.com/doc/api/1.1/CDbSchema#getTable-detail getTable use Caching system too.

c-schmitz: I know from experience that getTableNames is much faster, because it only gets the names, while getTable($sTableName)  gets the full schema.
If you don't believe me, feel free to run performance tests.

c-schmitz: Check this codebase's Yii1 framework code.

getTableNames() ([CDbSchema.php:136-141](vscode-webview://0scvk8pnboip8il8jmvlrhac02cg760jk28qjeci22rvm4fhfgkf/vendor/yiisoft/yii/framework/db/schema/CDbSchema.php#L136-L141)) calls findTableNames(), which for MySQL ([CMysqlSchema.php:258-266](vscode-webview://0scvk8pnboip8il8jmvlrhac02cg760jk28qjeci22rvm4fhfgkf/vendor/yiisoft/yii/framework/db/schema/mysql/CMysqlSchema.php#L258-L266)) just runs:


SHOW TABLES
One cheap query, cached per-schema in $_tableNames.

getTable($sTableName) ([CDbSchema.php:72](vscode-webview://0scvk8pnboip8il8jmvlrhac02cg760jk28qjeci22rvm4fhfgkf/vendor/yiisoft/yii/framework/db/schema/CDbSchema.php#L72)) calls loadTable(), which for MySQL ([CMysqlSchema.php:123-135](vscode-webview://0scvk8pnboip8il8jmvlrhac02cg760jk28qjeci22rvm4fhfgkf/vendor/yiisoft/yii/framework/db/schema/mysql/CMysqlSchema.php#L123-L135)) runs, per table:

SHOW FULL COLUMNS FROM <table> (findColumns, line 165)
SHOW CREATE TABLE <table> (findConstraints, line 231, to regex-parse foreign keys)
So getTable() does 2 extra queries per table and builds full CMysqlTableSchema/CMysqlColumnSchema objects for every column, versus getTableNames()'s single lightweight SHOW TABLES call. If you only need table names (e.g. checking existence or listing tables), getTableNames() is indeed significantly faster — the difference grows with the number of tables checked, since getTable() cost is per-call while getTableNames()'s single result can be reused/filtered in PHP.

Shnoulle: > getTableNames is much faster, because it only gets the names, while getTable($sTableName) gets the full schema.

Oh ! Right :+1: 

Thehn : need to keep current fix but move to a better fetTbaleName system

Seems each getTablesNames are DB specific

https://www.yiiframework.com/doc/api/1.1/CDbSchema#findTableNames-detail

Maybe we need to implement it for the 3 DB ? https://github.com/LimeSurvey/LimeSurvey/tree/master/application/core/db

35grain: > Why not clear the schema cache once after deactivation instead of slowing down the whole application.

I think deactivation wouldn't be the correct place since the original issue lies within the activation flow. Besides, this wouldn't change the fact that tableExists behaves unexpectedly. So if schema refresh is the path to take, then I guess we would need to use this basically everywhere with subsequent calls to tableExists. As I noted ealrier, Yii 2 has a function for refreshing only the schema for a specific table, which would be definitely the best solution.

I'm definitely open to improvement ideas, but as I also described on Mantis, the original issue APPEARS to be fixed in v7, but the underlying issues are still there.

PR #4779, opened 2026-03-16
Fixed issue #20447: [security] User can set any existing theme when saving survey

gabrieljenik:
## Summary by CodeRabbit

* **Bug Fixes**
  * Template selection now respects permission rules: "inherit" remains always allowed.
  * Existing surveys without permission preserve their stored template (or fallback to "inherit") instead of forcing the global default.
  * New surveys without permission default to "inherit".
  * Template names are validated and returned consistently after these checks.

### DIFF
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -599,15 +599,13 @@ class Survey extends LSActiveRecord implements PermissionInterface
      */
     public function filterTemplateSave($sTemplateName)
     {
+        // Make sure the user has permission to use the template.
         // 'inherit' is always permitted — the dropdown shows it without any permission check
         if ($sTemplateName !== 'inherit' && !Permission::model()->hasTemplatePermission($sTemplateName)) {
-            // Reset to default only if different from actual value
+            // For new records we can just set it to inherit, for existing ones we keep the old value.
             if (!$this->isNewRecord) {
                 $oSurvey = self::model()->findByPkNoCache($this->sid);
-                if ($oSurvey->template != $sTemplateName) {
-                    // No need to test !is_null($oSurvey)
-                    $sTemplateName = getGlobalSetting('defaulttheme');
-                }
+                $sTemplateName = $oSurvey->template ?? 'inherit';
             } else {
                 $sTemplateName = 'inherit';
             }
--- a/application/models/Survey.php
+++ b/application/models/Survey.php
@@ -603,8 +603,7 @@ class Survey extends LSActiveRecord implements PermissionInterface
         if ($sTemplateName !== 'inherit' && !Permission::model()->hasTemplatePermission($sTemplateName)) {
             // Reset to default only if different from actual value
             if (!$this->isNewRecord) {
-                $this->resetCache();
-                $oSurvey = self::model()->findByPk($this->sid);
+                $oSurvey = self::model()->findByPkNoCache($this->sid);
                 if ($oSurvey->template != $sTemplateName) {
                     // No need to test !is_null($oSurvey)
                     $sTemplateName = getGlobalSetting('defaulttheme');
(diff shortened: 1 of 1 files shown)


gabrieljenik:
> One minor concern: Line 607 accesses $oSurvey->template without null checking.

It is an impossible situation. I wouldn't spend time on it.

application/models/Survey.php
```diff
@@ -599,10 +599,11 @@ public function afterFindSurvey()
      */
     public function filterTemplateSave($sTemplateName)
     {
-        if (!Permission::model()->hasTemplatePermission($sTemplateName)) {
+        // 'inherit' is always permitted — the dropdown shows it without any permission check
+        if ($sTemplateName !== 'inherit' && !Permission::model()->hasTemplatePermission($sTemplateName)) {
             // Reset to default only if different from actual value
             if (!$this->isNewRecord) {
-                $oSurvey = self::model()->findByPk($this->sid);
+                $oSurvey = self::model()->findByPkNoCache($this->sid);
```
Shnoulle, line 607: Yes, why not? In my plugins I use findByAttribute(['sid'=>$sid]).

New function is great

application/models/Survey.php
```diff
@@ -599,10 +599,11 @@ public function afterFindSurvey()
...
                 if ($oSurvey->template != $sTemplateName) {
                     // No need to test !is_null($oSurvey)
                     $sTemplateName = getGlobalSetting('defaulttheme');
```
Shnoulle: It's an issue here: must reset to `$oSurvey->template` : no update must be done.

`$sTemplateName = $oSurvey->template;`

Shnoulle (commented):
I think current reset is bad. Must be a _do not update current one_

application/models/Survey.php
gabrieljenik: I am not sure to change that. 
First of all is not in the exact scope of this ticket. 
Then, I guess there is a reason to set it that way. And there could be unthought collateral damage if changing it.

Ex: in the case you suggest you supposing that the user has permission for the prior saved value ... 

I guess there could be others. Sorry, not sure, not so an obvious change... Risk reward for doing it feels like more risk than benefits...

Shnoulle: > Ex: in the case you suggest you supposing that the user has permission for the prior saved value ...

If user didn't have read access to current survey theme set : then :

1. Theme is selected in GUI
2. When save theme are not updated : no check of access

It's the desired behavior ! Updating to default it's not the desired behaviour !

Remind : set tpo a bad theme are done **hacking** HTML , it's not a easy way. Without hack HTML, woithout any access to theme : theme not updated.

Shnoulle (changes requested):

Shnoulle:
No reason to reset, if bad theme is set : don't update it.
Simple

tiborpacalat:
@gabrieljenik can you please resolve conflicts here and make requested changes?

Shnoulle:
> @gabrieljenik can you please resolve conflicts here and make requested changes?

See the discussion here: https://github.com/LimeSurvey/LimeSurvey/pull/4779#discussion_r2951515167

In my opinion: the template must be checked ONLY if it's different.

Then Superadmin can set a specific template for a survey even if the user didn't have accesss for all surveys.

c-schmitz:
Please use the following commit message on sqash commit:

Fixed issue #20447: [security] Missing permission enforcement when saving a survey's theme. Severity: Medium (CVSS 4.3, AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:L/A:N). An authenticated user with permission to edit their own survey's general settings, but no permission for a specific theme, could still apply that theme to their survey, Affects versions since 2015 (2.0x) through 7.1.2. Fixed in 7.1.3

Shnoulle:
I still believe that this commit should not be accepted as is. It removes the superadmin's ability to assign specific templates to certain surveys even if the owner does not have access to them.

PR #4790, opened 2026-03-19
Fix issue #20466: Issue 20337 fix causes tablesorter dependency error in Survey permissions view

35grain:
Fixed issue #20466: https://bugs.limesurvey.org/view.php?id=20466

This is quite a hacky solution but I don't really see a better way to go about it at the moment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed grid action controls so dropdowns and related UI initialize reliably even when lists start empty, ensuring AJAX-loaded rows retain full action support.
  * Adjusted table-sorting script inclusion timing to improve page render order and stability of permission and table interfaces.

### DIFF
--- a/application/controllers/SurveyPermissionsController.php
+++ b/application/controllers/SurveyPermissionsController.php
@@ -65,7 +65,13 @@ class SurveyPermissionsController extends LSBaseController
         App()->getClientScript()->registerPackage('usermanagement');
         App()->getClientScript()->registerPackage('select2-bootstrap');
 
-        // Register client scripts for grid action dropdown in case no users are added yet.
+        // Register client scripts for grid action dropdown
+        // This is usually done inside the grid widget.
+        // Still, in case no users are added yet, the grid widget doesn't render any row 
+        // and doesn't include this widget.
+        // As this page reloads through ajax when adding a new row, we need the GridActionsWidget to be already loaded
+        // In the future, the Grid widget could have a callback to allow to register widgets, even when no rows.
+        // But for now, this workaround is ok
         App()->getController()->widget('ext.admin.grid.GridActionsWidget.GridActionsWidget', [], true);
         App()->getClientScript()->registerScriptFile(App()->getConfig('adminscripts') . 'surveypermissions.js', CClientScript::POS_END);
         $oSurvey = Survey::model()->findByPk($surveyid);
--- a/application/controllers/SurveyPermissionsController.php
+++ b/application/controllers/SurveyPermissionsController.php
@@ -67,7 +67,7 @@ class SurveyPermissionsController extends LSBaseController
 
         // Register client scripts for grid action dropdown
         // This is usually done inside the grid widget.
-        // Still, in case no users are added yet, the grid widget doesn't render any row 
+        // Still, in case no users are added yet, the grid widget doesn't render any row
         // and doesn't include this widget.
         // As this page reloads through ajax when adding a new row, we need the GridActionsWidget to be already loaded
         // In the future, the Grid widget could have a callback to allow to register widgets, even when no rows.
--- a/application/controllers/SurveyPermissionsController.php
+++ b/application/controllers/SurveyPermissionsController.php
@@ -61,9 +61,7 @@ class SurveyPermissionsController extends LSBaseController
             Yii::app()->user->setFlash('error', gT("No permission or survey does not exist."));
             $this->redirect(Yii::app()->request->urlReferrer);
         }
-        App()->getClientScript()->registerPackageScriptOnPosition('jquery-tablesorter', CClientScript::POS_END);
-        // Keep jQuery in head to not break other scripts in head.
-        App()->getClientScript()->registerPackageScriptOnPosition('jquery', CClientScript::POS_HEAD);
+        App()->getClientScript()->registerPackage('jquery-tablesorter');
         App()->getClientScript()->registerPackage('usermanagement');
         App()->getClientScript()->registerPackage('select2-bootstrap');
 
--- a/application/controllers/SurveyPermissionsController.php
+++ b/application/controllers/SurveyPermissionsController.php
@@ -61,11 +61,14 @@ class SurveyPermissionsController extends LSBaseController
             Yii::app()->user->setFlash('error', gT("No permission or survey does not exist."));
             $this->redirect(Yii::app()->request->urlReferrer);
         }
-        App()->getClientScript()->registerPackage('jquery-tablesorter');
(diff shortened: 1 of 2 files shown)


application/controllers/SurveyPermissionsController.php
```diff
@@ -61,11 +61,14 @@ public function actionIndex($surveyid)
             Yii::app()->user->setFlash('error', gT("No permission or survey does not exist."));
             $this->redirect(Yii::app()->request->urlReferrer);
         }
-        App()->getClientScript()->registerPackage('jquery-tablesorter');
+        App()->getClientScript()->registerPackageScriptOnPosition('jquery-tablesorter', CClientScript::POS_END);
```
gabrieljenik: This changes seems to be the only one needed. Why the rest? Is that part of the specific fix ?

35grain: I actually described two issues in the bug tracker: https://bugs.limesurvey.org/view.php?id=20466.

https://github.com/LimeSurvey/LimeSurvey/pull/4790/changes#diff-fc510db939b5afc725e1e6f3d5ce9ac31c56bb3b978865297e34d621c38b2691R65-R66 (lines 65-66) is required as jQuery is a dependency of jquery-tablesorter and moving it to POS_END also moves jQuery. This simply ensures that jQuery stays in the head section so that other scripts do not break (and yes, they did break without this line). 

https://github.com/LimeSurvey/LimeSurvey/pull/4790/changes#diff-fc510db939b5afc725e1e6f3d5ce9ac31c56bb3b978865297e34d621c38b2691R70-R71 (lines 70-71) is for fixing the second issue regarding action_dropdown.js not being loaded if no users are listed in the Survey permissions table.

gabrieljenik: > is required as jQuery is a dependency of jquery-tablesorter and moving it to POS_END also moves jQuery.

If jquery is a dependency of jquery-tablesorter, we shouldn't load it manually. That's not consistent with the rest of the code. Also, we have the config/packages.php for that, as well as the application\config\vendor.php.

I suggest setting the position of the jquery-tablesorter to begin.
> 'position' => CClientScript::POS_BEGIN,

That way jquery will be loaded at the beggining.

---

As per the action_dropdown.js, if it is not being loaded when no users are listed, we should identify where that is happening and correct it there. Have you located why action_dropdown.js is not being loaded?

35grain: > As per the action_dropdown.js, if it is not being loaded when no users are listed, we should identify where that is happening and correct it there. Have you located why action_dropdown.js is not being loaded?

Because GridActionsWidget doesn't exist on the page, therefore the scripts are not registered on load. If the widget is created via `$.fn.yiiGridView.update('gridPanel', {});`, the scripts aren't loaded dynamically. I unfortunately do not have the time to study the ins and outs of LimeSurvey, so please do suggest a cleaner way if one exists.

gabrieljenik:
OK, will review the GridActionsWidget part of the issue.

application/controllers/SurveyPermissionsController.php
gabrieljenik: After analyzing, the workaround is OK until we have a better system for loading widgets which could by called by widgets.

application/controllers/SurveyPermissionsController.php
```diff
@@ -65,7 +65,14 @@ public function actionIndex($surveyid)
...
 
-
+        // Register client scripts for grid action dropdown
+        // This is usually done inside the grid widget.
+        // Still, in case no users are added yet, the grid widget doesn't render any row
+        // and doesn't include this widget.
+        // As this page reloads through ajax when adding a new row, we need the GridActionsWidget to be already loaded
+        // In the future, the Grid widget could have a callback to allow to register widgets, even when no rows.
+        // But for now, this workaround is ok
+        App()->getController()->widget('ext.admin.grid.GridActionsWidget.GridActionsWidget', [], true);
```
Shnoulle, line 75: We register the script ?

It's more a workaround than a solution. No way to fix it with a _stable_ solution ?

application/controllers/SurveyPermissionsController.php
```diff
@@ -65,7 +65,14 @@ public function actionIndex($surveyid)
         App()->getClientScript()->registerPackage('usermanagement');
         App()->getClientScript()->registerPackage('select2-bootstrap');
 
-
+        // Register client scripts for grid action dropdown
+        // This is usually done inside the grid widget.
+        // Still, in case no users are added yet, the grid widget doesn't render any row
+        // and doesn't include this widget.
+        // As this page reloads through ajax when adding a new row, we need the GridActionsWidget to be already loaded
+        // In the future, the Grid widget could have a callback to allow to register widgets, even when no rows.
```
Shnoulle, line 73: Maybe it's the best solution here ? Fix the widget ?

Shnoulle (commented):
I think the best solution is to fix the widget and register the script if there are no rows.

application/controllers/SurveyPermissionsController.php
Shnoulle: > After analyzing, the workaround is OK until we have a better system for loading widgets which could by called by widgets.

I really think the **needed** solution is to fix the widget.

application/controllers/SurveyPermissionsController.php
gabrieljenik: It is a workarond in deed.
The dependency is as follows:
- GridWidget renders. 
- If rows to be rendered, GridWidget  Invokes Model.
- Model includes GridActionsWidget
So the dependency in between the 2 widgets is not fixed for all.

The stable solution IMHO could be:
- Alt A:
  + Add a callback "postRender" (or similar), as to include the widget all time. The callback could call `App()->getController()->widget('ext.admin.grid.GridActionsWidget.GridActionsWidget', [], true);`
- Alt B:
  + Review if there is already a standard event we can use to register the GridActionsWidget widget. 
- Alt C:
  + Add a specific callback for the case where no rows are found.
- Alt D:
  + Add a call to the Model in case there are no rows to be rendered. The model could register the GridActionsWidget. I don't like this as the Model is not to be aware that the GridActionsWidget needs to be included anyway. That requirement appears because the grid is updated through ajax in the CRUD.

But all these looks more like a feature/enhancement than a bugfix

Shnoulle: > Model includes GridActionsWidget

Models include views :(

Alt C seems great.

Maybe we can merge it like that and create a better fix ?

Shnoulle (approved):
TODO : create a better fix for future

PR #4797, opened 2026-03-20
Fixed issue #20468: Condition not properly behaving when respondents enters 'false' as response

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed expression evaluation to properly preserve non-numeric string values in their original form, rather than applying unnecessary normalization rules.


Shnoulle (commented):
But : no issue happened with this situation?

Shnoulle (approved):
Testing must do a lot of Expression test !

PR #4799, opened 2026-03-20
Fixed issue #20463: Quick add labels in label set doesn't generate label codes correctly

c-schmitz:
## Summary by CodeRabbit

* **New Features**
  * Quick-add labels now auto-generates missing codes and assigns sensible sequential defaults.

* **Bug Fixes**
  * Prevents duplicate codes when adding labels in bulk.
  * Skips empty or whitespace-only lines during bulk import to avoid unintended entries.

* **Refactor**
  * Internal cleanup of label editing logic with no user-facing changes.


lajosarpad (changes requested):
The changes make sense, however, if we take a look at the definition of `isNumeric` in assets/scripts/admin/questionEditor.js then we see this:

```
  /**
   * @param {any} mixedVar
   * @return {boolean}
   */
  function isNumeric(mixedVar /*: mixed */) {
    return (typeof (mixedVar) === 'number' || typeof (mixedVar) === 'string') && mixedVar !== '' &&
      (typeof mixedVar === 'number' && !isNaN(mixedVar));
  }

```

Translated into plain words:

- mixedVar is a number or a string
- mixedVar is not empty
- mixedVar is a number and not not a number

This will fail for cases when we pass a number as a string, such as "1". So, if we call `getNextCode`, which is

```
  /**
   * @param {string} sSourceCode
   * @return {string}
   * @todo Used in label sets? But not in question editor?
   * @todo Remove
   */
  function getNextCode(sSourceCode) {  // jshint ignore: line
    const sourcecode = sSourceCode;
    let i = 1;
    let found = true;
    let foundnumber = -1;
    const sclength = sourcecode.length;
    while (i <= sclength && found === true) {
      found = isNumeric(sourcecode.substr(sclength - i, i));
      if (found) {
        foundnumber = parseInt(sourcecode.substr(sclength - i, i));
        i++;
      }
    }
    if (foundnumber === -1) {
      return sourcecode;
    }

    foundnumber++;
    const foundnumberString = foundnumber.toString();
    const result = sourcecode.substr(0, sclength - foundnumberString.length) + foundnumberString;
    return (result);
  }
```

and therefore trying to get the number digit by digit, then it will find that from the default "L001" the last digit is "1" and will pass that to `isNumeric` which, in turn return false. It is a logical error to expect the parameter to be of number type and not not a number, because in an earlier criteria we allowed a string, with the intent of treating numeric strings, such as "1" as a number and return true. The change I propose is

```
  /**
   * @param {any} mixedVar
   * @return {boolean}
   */
  function isNumeric(mixedVar /*: mixed */) {
    return (typeof (mixedVar) === 'number' || typeof (mixedVar) === 'string') && mixedVar !== '' &&
      (typeof mixedVar === 'number' || !isNaN(mixedVar));
  }
```

which in plain words would mean:

- mixedVar is a number or a string
- mixedVar is not empty
- mixedVar is either a number, or at least not not a number

If I pass an input of "L001" to getNextCode after such a change, then I get "L002", as expected.

c-schmitz:
@lajosarpad The isNumeric in questionEditor.js is not used here. label.js uses its own isNumeric function which does not have the issue you pointed out.

lajosarpad (approved):

lajosarpad:
> @lajosarpad The isNumeric in questionEditor.js is not used here. label.js uses its own isNumeric function which does not have the issue you pointed out.

Understood. Then, as long as it was tested, it looks good. However, mssql pipeline is failing at the time of this writing

PR #4803, opened 2026-03-20
New feature #18275: Long and huge free text question properly handle long input

c-schmitz:
Long free text field size now defaults to 100kb, Huge free text field size now defaults to 1MB.
These limits are the same across all database types

## Summary by CodeRabbit

* **New Features**
  * Enforced standardized character limits for Long Free Text (T) and Huge Free Text (U), with clearer validation feedback when limits are exceeded (while keeping the submitted value).
  * Updated Long/Huge Free Text textareas to use consistent `maxlength` values based on effective defaults when `maximum_chars` is empty, with a shared global cap (100KB for T / 1MB for U, capped at 10MB), and matching styling.
  * Refreshed `maximum_chars` help text for both question types to document defaults and maximums.
* **Chores**
  * Bumped the database schema version and added a MySQL/MariaDB migration to adjust response column types for T/U answers.


application/views/survey/questions/answer/longfreetext/config.xml
```diff
@@ -588,8 +588,8 @@
             <category>Input</category>
             <sortorder>100</sortorder>
             <inputtype>integer</inputtype>
-            <default></default>
-            <help>Maximum characters allowed</help>
+            <default>102400</default>
+            <help>Maximum characters allowed (default: 102400 = 100KB, max: 10485760 = 10MB)</help>
```
Shnoulle: I like to still have an empty value here, and if empty ( === '' or is_null) get the fixed default one.

application/helpers/update/updates/Update_702.php
```diff
@@ -0,0 +1,77 @@
...
+class Update_702 extends DatabaseUpdateBase
+{
+    /**
+     * Alter TEXT columns to MEDIUMTEXT for Long free text (T) and Huge free text (U)
+     * question types in existing MySQL/MariaDB response tables.
+     *
+     * @return void
+     */
+    public function up()
+    {
```
Shnoulle, line 13: Can break easily with BIG DB (more than 3000 surveys)

maybe only for new survey table ? Or on demand ?

application/views/survey/questions/answer/hugefreetext/config.xml
```diff
@@ -588,8 +588,8 @@
             <category>Input</category>
             <sortorder>100</sortorder>
             <inputtype>integer</inputtype>
-            <default></default>
-            <help>Maximum characters allowed</help>
+            <default>1048576</default>
```
Shnoulle: Can be great to leave `<default></default>` and show the real defauilt in helps only

Shnoulle (commented):
Afraid of big DB (more than 3000 surveys)

And default set as ""

application/helpers/expressions/em_manager_helper.php
```diff
@@ -9878,7 +9878,10 @@ public static function convertNonLatinNumerics($str, $lang)
      * @param boolean $set : update the invalid string or not. Used for #14649 (invalid default value)
      * @throw Exception
      *
-     * @return boolean true : if question is OK to be put in session, false if must be set to null
+     * @return boolean true if value is valid and should be kept in session;
+     *                  false if value must be set to null (e.g. invalid choice).
+     *                  Note: for T/U text length violations, returns true (value preserved)
```
Shnoulle, line 9983: I think we have same process for number: if not a valid DECIMAL : set an invlid string but return false

Shnoulle: Confirm : https://github.com/LimeSurvey/LimeSurvey/blob/00741c9b1bddd56279ccec529314849739205e22/application/helpers/expressions/em_manager_helper.php#L9986

application/helpers/expressions/em_manager_helper.php
```diff
@@ -10028,6 +10031,22 @@ private static function checkValidityAnswer($type, $value, $sgq, $qinfo, $set =
                 break;
             case 'U': // Huge text
```
Shnoulle, line 10132: Must add a control too for

- Array of text : https://github.com/LimeSurvey/LimeSurvey/blob/00741c9b1bddd56279ccec529314849739205e22/application/helpers/expressions/em_manager_helper.php#L9939
- Equation : https://github.com/LimeSurvey/LimeSurvey/blob/00741c9b1bddd56279ccec529314849739205e22/application/helpers/expressions/em_manager_helper.php#L9964

application/helpers/update/updates/Update_702.php
Shnoulle: @c-schmitz some users already have issues with the big DB upgrading to 7 due to the number of active surveys. I think it can be OK to force it via CLI for major releases. But here, it's a minor release :(

c-schmitz: It is failsafe though. Even if the Db udpate should fail here it can be run again without throwing errors. Fields that were already converted will not throw an error.

Shnoulle: >  Even if the Db udpate should fail here it can be run again without throwing errors

I don't understand here, 

1. DBversion stay 702 
2. Still seacrh for all responses\_% table
3. No test if need to be done

Then if it's break after 1000 responses table, you launch again and break after v1000etc ..

Maybe add a "response table done" in session, and cotinue if it's already done ?

PR #4842, opened 2026-04-02
Bug/tb 82 anonymize responses dates

Trischi80:
Fixed issue #T82: anonymize responses should have higher priority then datestamp. Submit date and date last action and start date should be anonymized if the setting is set to on.

## Summary by CodeRabbit

* **Bug Fixes**
  * When surveys are configured as anonymized, timestamp fields now display a constant default value instead of actual submission times.


Trischi80:
old version...

PR #4844, opened 2026-04-02
LE-662: change script fixes

lajosarpad:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved field-name resolution across multiple question types (multiple-choice variants, ranking, semicolon/colon arrays and timing fields) to reduce mis-mapped responses.
  * Corrected survey field-mapping logic to avoid missed or incorrect mappings during migrations.
  * Prevented erroneous saves during relevance/text migration when source data is incomplete.

* **Improvements**
  * Strengthened cache availability checks.
  * Safer theme validation to avoid unintended redirects or messages.


application/helpers/update/updates/Update_700.php
```diff
@@ -1082,6 +1083,9 @@ public function up()
                 $postinsert = "SET IDENTITY_INSERT {$scripts[$TABLE_NAME]['new_name']} OFF;";
             }
             $scripts[$TABLE_NAME]['handled'] = true;
+            if (!isset($scripts[$TABLE_NAME]['new-name'])) {
+                continue;
+            }
```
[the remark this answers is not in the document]
lajosarpad: Fixed

application/helpers/common_helper.php
```diff
@@ -1468,8 +1488,43 @@ function getFieldName(string $tableName, string $fieldName, array $questions, in
...
+                        }
+                        $diff = 0;
+                        if ($minSortOrder === 0) {
+                            $diff = -1;
+                        } else if ($minSortOrder > 1) {
+                            $diff = $minSortOrder;
+                        }
+                        foreach ($answers as $answer) {
+                            if (($rankingSuffix == $answer->code) || ((intval($iRankingSuffix) > 0) && ($rankingSuffix + $diff == $answer->sortorder))) {
+                                $newFieldName = "Q{$qid}_{$prefix}{$answer->aid}";
```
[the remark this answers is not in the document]
lajosarpad: That scenario is a broken scenario and therefore the behavior is correct, because if there is no valid new fieldname, then the original fieldname is a valid fallback value.

application/helpers/update/updates/Update_700.php
```diff
@@ -1116,6 +1117,7 @@ public function up()
                 if (strpos($TABLE_NAME, "old") !== false) {
                     continue;
                 } else {
+                    echo $scripts[$TABLE_NAME]['CREATE'] . "\n\n";
```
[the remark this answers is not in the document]
lajosarpad: We needed to have these echoes due to test env limitations, but they were cleaned up since then

application/helpers/update/updates/Update_700.php
```diff
@@ -1204,7 +1206,9 @@ public function up()
                             $this->fixText($entity, $ef['fields'], $additionalNames)
                         ];
                         if ($save[0] || $save[1] || $save[2]) {
-                            $entity->save();
+                            if (!(in_array('relevance', $ef['fields']) && $entity->qid && (!$entity->survey))) {
+                                $entity->save();
+                            }
```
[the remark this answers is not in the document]
lajosarpad: This is a guard against broken database structures. If the question has some changes we were to do, but has no survey reference, then the structure of the question is broken and we need to skip it. We don't need to update question fields where the structure is broken because then we may presume something incorrect and do more damage with the fix than without it.

Collecting all inconsistencies everywhere is not very useful, because we don't intend to fix all inconsistencies of all users, only those they ask us to fix. So if they already have inconsistencies, then it's best to avoid changing the field value and let the owner of the survey fix the inconsistency and the value, or if they need help they can reach out to our support team and provide the context that's necessary to properly understand the intent behind the broken structure.

application/helpers/common_helper.php
```diff
@@ -1317,31 +1317,40 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
...
                 }
             }
             $questionIndex++;
         }
+        $questions = [$rootQuestion];
+        foreach ($rawQuestions as $rawQuestion) {
+            if ($rawQuestion->parent_qid == $rootQuestion->qid) {
+                $questions[] = $rawQuestion;
+            }
+        }
```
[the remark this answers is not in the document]
lajosarpad: This is a deprecated concern

application/helpers/common_helper.php
```diff
@@ -1317,31 +1317,40 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
...
         $qid = $rootQuestion->qid;
+        if ($fieldName === "260105X3X56") {
+            echo var_dump([$rootQuestion->qid, $rootQuestion->type]);
+        }
```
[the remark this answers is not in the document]
lajosarpad: Debug code was necessary due to test environment limitations, but were removed since then.

application/helpers/common_helper.php
```diff
@@ -1361,7 +1370,12 @@ function getFieldName(string $tableName, string $fieldName, array $questions, in
                     }
                 }
                 $hashTags = explode("#", $fieldName);
-                if ($currentQuestion === null) {
+                if (($rootQuestion->type === \Question::QT_M_MULTIPLE_CHOICE) && ($length < strlen($hashTags[0]))) {
+                    $currentQuestion = \Question::model()->find("parent_qid = {$qid} and title = :title", [
+                        ":title" => substr($hashTags[0], $length)
+                    ]);
+                }
```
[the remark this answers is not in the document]
lajosarpad: No, this is the actual qid, loaded directly from the question which is guaranteed to be a number

application/helpers/common_helper.php
```diff
@@ -1292,160 +1292,221 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
...
+                            }
+                            $diff = 0;
+                            if ($minSortOrder === 0) {
+                                $diff = -1;
+                            } else if ($minSortOrder > 1) {
+                                $diff = $minSortOrder;
+                            }
+                            foreach ($answers as $answer) {
+                                if (($rankingSuffix == $answer->code) || ((intval($iRankingSuffix) > 0) && ($rankingSuffix + $diff == $answer->sortorder))) {
+                                    $newFieldName = "Q{$qid}_{$prefix}{$answer->aid}";
```
karcstefan: break here if match found?

application/helpers/common_helper.php
```diff
@@ -1292,160 +1292,221 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
...
+                                }
+                            }
+                            if (!$newFieldName) {
+                                $newFieldName = $fieldName;
+                            }
+                        }
+                    } catch (\Exception $ex) {
+                        if (strpos($tableName, 'old') === false) {
+                            throw $ex; //Ignore inconsistencies in archive rankings
```
karcstefan: Trivial: Not sure if comment matches intention, as in throwing exception again is not "ignoring".

application/helpers/common_helper.php
```diff
@@ -1292,160 +1292,221 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
...
+                        $newFieldName .= "_Ccomment";
+                    }
+                    break;
+                case \Question::QT_R_RANKING:
+                    $prefix = ((strpos($tableName, "timing") !== false) ? "C" : "R");
+                    $index = substr($fieldName, strlen("{$sid}X{$gid}X{$qid}"));
+                    try {
+                        $rankingSuffix = substr($fieldName, strlen("{$sid}X{$gid}X{$qid}"));
+                        $iRankingSuffix = intval($rankingSuffix);
+                        if ((intval($iRankingSuffix) > 0) && isset($questions[0]->answers[($iRankingSuffix - 1)])) {
```
karcstefan: Trivial: intval already used on L1470, can be removed

application/helpers/common_helper.php
```diff
@@ -1292,160 +1292,221 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
...
+                    } elseif (strpos($fieldName, "filecount") !== false) {
+                        $newFieldName .= "_Cfilecount";
+                    }
+                    if ($isComment) {
+                        $newFieldName .= "_Ccomment";
+                    }
+                    break;
+                case \Question::QT_R_RANKING:
+                    $prefix = ((strpos($tableName, "timing") !== false) ? "C" : "R");
+                    $index = substr($fieldName, strlen("{$sid}X{$gid}X{$qid}"));
```
karcstefan, line 1463: Trivial: not used?

application/helpers/common_helper.php
```diff
@@ -1292,160 +1292,221 @@ function createCompleteSGQA($iSurveyID, $aFilters, $sLanguage)
...
-                        $currentQuestion = $question;
-                    } elseif (in_array($code, ["other", "comment", "othercomment", $question->title . "other", $question->title . "comment", $question->title . "othercomment"])) {
-                        $currentQuestion = $question;
-                        $commentText = $code;
-                        if (strpos($code, $question->title) === 0) {
-                            $commentText = substr($code, strlen($question->title));
-                        } else {
-                            $excludeSubquestion = true;
+                    $hashTags = explode("#", $fieldName);
+                    if (($rootQuestion->type === \Question::QT_M_MULTIPLE_CHOICE) && ($length < strlen($hashTags[0]))) {
```
karcstefan: Do we check here whether `$currentQuestion` is not set or should it overwrite `$currentQuestion` regardess if condition matched?

application/helpers/common_helper.php
lajosarpad: Good catch! We can return here the result rather than loop for further candidates

application/helpers/common_helper.php
lajosarpad: The comment was correct, but at the wrong place. Moved it to the right place

application/helpers/common_helper.php
lajosarpad: Indeed, removed it. Thanks!

application/helpers/common_helper.php
lajosarpad: Both $prefix and $index are used. $prefix determined whether the subquestion needs to start with a C or an R, depending on whether it's a timings table, whilst $index determines the answer index which comes after the root question form.

application/helpers/common_helper.php
lajosarpad: Added a check. It makes more sense that way. Thank you!

karcstefan (approved):

c-schmitz (changes requested):
Please review the 'Duplicate comments' by RabbitAI since they raise valid points, imho.

lajosarpad:
> Please review the 'Duplicate comments' by RabbitAI since they raise valid points, imho.

The first comment refers to lines 1347-1356. It is a fallback logic for the case when the passed related subquestions don't have the title specified.

For example the field is 1X2X3abc

and then we don't find the subquestion whose title is abc. In this case we attempt to load it from the database in case it exists and was just not passed to the method among the related questions.

If it fails to find the question even in the database, so there was no question among the related questions with the title we expected and we did not find it in the database either, then we cannot do anything better with the field than to append _Cabc . Because if we converted it into 1X2X3 like CoreRabbit suggested and we had another field by the name of 1X2X3def whose title is not found either, then that would be converted into 1X2X3 too, causing a duplicated field error.

Logically, if the question ends with a title and the title does not exist for the subquestions, then we have an inconsistency in the database.

As about the second comment, with $questions being persisted across iterations, CodeRabbit is right and a fix was implemented and pushed.

karcstefan (changes requested):
There are failing tests here and merge conflicts

PR #4857, opened 2026-04-08
Fixed issue #20480: Proper font support for non-latin chars in PDF export

mohabmes:
## Summary by CodeRabbit

* **Bug Fixes**
  * Switched default PDF fonts to DejaVu Sans for page and data text, improving font consistency across generated documents (including headers/footers).
  * Removed several alternate PDF font mappings for specific languages, simplifying fallback behavior for those locales.


c-schmitz:
After reviewing the customer case I changed my mind and I think it is better to change the default PDF font.

c-schmitz (approved):

PR #4860, opened 2026-04-09
TB-84: handle question import

Ahmedazeem98:
## Summary by CodeRabbit

* **Bug Fixes**
  * Legacy ranking answers are now converted into ranking subquestions with corresponding translations during import. The import process will skip those legacy answers and their localization entries to prevent them from being imported as regular answers and avoid duplicate localization records.


application/helpers/admin/import_helper.php
```diff
@@ -972,15 +977,40 @@ function XMLImportQuestion($sFullFilePath, $iNewSID, $iNewGID, $options = array(
...
+
+            $oldQid = (int) $insertdata['qid'];
+            if (!isset($aQIDReplacements[$oldQid])) {
+                continue;
             }
-            if (!isset($aQIDReplacements[(int) $insertdata['qid']])) {
+
+            $newParentQid = $aQIDReplacements[$oldQid];
+
+            if (isset($questionTypeMap[$newParentQid]) && $questionTypeMap[$newParentQid] == 'R') {
```
lajosarpad: We need to handle both the case when we have a legacy file to import which has answers that we are to convert subquestions and the case when it already contains subquestions.

A rule-of-thumb to be applied here is that whenever an answer is to be inserted for a ranking question, we need to create an equivalent subquestion instead.

We need to use Question::QT_R_RANKING instead of the hard-coded 'R'

application/helpers/admin/import_helper.php
```diff
@@ -972,15 +977,40 @@ function XMLImportQuestion($sFullFilePath, $iNewSID, $iNewGID, $options = array(
...
+                $subQuestionData = [
+                    'sid' => $iNewSID,
+                    'gid' => $iNewGID,
+                    'parent_qid' => $newParentQid,
+                    'type' => 'R',
```
lajosarpad: We need to use the Question::QT_R_RANKING instead of hard-coding it here, because if the value changes from 'R' to anything else, then this code will be broken if we hard-code R

application/helpers/admin/import_helper.php
```diff
@@ -972,15 +977,40 @@ function XMLImportQuestion($sFullFilePath, $iNewSID, $iNewGID, $options = array(
...
+                    'title' => $insertdata['code'],
```
lajosarpad: Other subquestions handle translink fields:

```
                if ($options['translinkfields']) {
                    $insertdata['question'] = translateLinks('survey', $iOldSID, $iNewSID, $insertdata['question']);
                    $insertdata['help'] = translateLinks('survey', $iOldSID, $iNewSID, $insertdata['help']);
                }
```

application/helpers/expressions/em_manager_helper.php
```diff
@@ -3602,6 +3602,26 @@ public function setVariableAndTokenMappingsForExpressionManager($surveyid, $forc
...
+                    $varName = $fielddata['title'] . '_' . $fielddata['aid'];
+                    $question = $fielddata['subquestion'];
+                    // In M and P , we use $question (sub question) for shown. With other : we show to the user 'other_replace_text' if it's set. see #13505
+                    if ($other == "Y") {
+                        if (isset($qattr[$questionNum]['other_replace_text']) && trim((string) $qattr[$questionNum]['other_replace_text']) != '') {
+                            $question = trim((string) $qattr[$questionNum]['other_replace_text']);
+                        } else {
+                            $question = $this->gT('Other:');
+                        }
+                    }
```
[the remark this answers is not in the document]
Ahmedazeem98: not part of this PR changes

application/helpers/admin/import_helper.php
```diff
@@ -4873,3 +4930,131 @@ function savePendingInsertansUpdates($pendingInsertansUpdates)
...
+            continue;
+        }
+
+        // Use the old answer ID as the placeholder qid so the subquestions
+        // import loop can track it in $aQIDReplacements and assign a real qid.
+        $subQuestionData = [
+            'sid'            => $iNewSID,
+            'gid'            => $iGID,
+            'parent_qid'     => $iOldParentQID, // old qid — subquestions loop remaps this
+            'type'           => Question::QT_R_RANKING,
```
lajosarpad: The main question has a ranking type, subquestions need to be texts, let's change this type to `QT_T_LONG_FREE_TEXT`.

lajosarpad (changes requested):

application/helpers/admin/import_helper.php
Ahmedazeem98: Updated

application/helpers/admin/import_helper.php
Ahmedazeem98: should be resolved

application/helpers/admin/import_helper.php
Ahmedazeem98: should be resolved

application/helpers/admin/import_helper.php
Ahmedazeem98: should be resolved

PR #4868, opened 2026-04-10
New feature #20482: Add one click unsubscribe functionality

mfavetti:
https://bugs.limesurvey.org/view.php?id=20482

New feature #20482: Add one click unsubscribe functionality

## Summary by CodeRabbit

* **New Features**
  * Added RFC 8058 one‑click unsubscribe support and included List‑Unsubscribe headers in outgoing emails.

* **Improvements**
  * Unified and more reliable opt‑out/opt‑in state handling across flows.
  * Stricter request validation, added redirect guards, and explicit error handling when persistence fails.

* **Chores**
  * Normalized routing and CSRF configuration to enable the one‑click flow from email clients.

### DIFF
--- a/application/config/internal.php
+++ b/application/config/internal.php
@@ -161,6 +161,7 @@ $internalConfig = array(
                 'rest',
                 'admin/remotecontrol',
                 'plugins/unsecure',
+                'optout/tokens',
             ),
             'csrfCookie' => array(
                 'secure' => ((!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off') || (isset($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] == 443)),
--- a/application/config/routes.php
+++ b/application/config/routes.php
@@ -1,4 +1,6 @@
-<?php  if (!defined('BASEPATH')) {
+<?php
+
+if (!defined('BASEPATH')) {
     exit('No direct script access allowed');
 }
 /*
@@ -20,7 +22,7 @@ $route['<_sid:\d+>/lang-<_lang:\w+[-\w]+>/tk-<_token:\w+>/*'] = "survey/index/si
 $route['<_sid:\d+>/lang-<_lang:\w+[-\w]+>/*'] = "survey/index/sid/<_sid>/lang/<_lang>";
 $route['<_sid:\d+>/tk-<_token:\w+>/*'] = "survey/index/sid/<_sid>/token/<_token>";
 $route['<_sid:\d+>/*'] = "survey/index/sid/<_sid>";
-$route['<sid:\d+>'] = array('survey/index', 'matchValue'=>true);
+$route['<sid:\d+>'] = array('survey/index', 'matchValue' => true);
 
 //Admin Routes
 $route['admin/index'] = "admin"; // this can not be changed to "dashboard/view" as long as the AdminController exists, will break "path" urlFormat
@@ -57,11 +59,11 @@ $route['rest/<_api_version:\w+>/<_entity>/<_id>/<_basetable>'] = array(
 
 //optout - optin
 $route['optout/<_sid:\d+>/(:any)/(:any)'] = "optout/index/<_sid>/$2/$3";
-$route['optout/tokens/<surveyid:\d+>'] = array('optout/tokens', 'matchValue'=>true);
-$route['optout/participants/<surveyid:\d+>'] = array('optout/participants', 'matchValue'=>true);
-$route['optin/tokens/<surveyid:\d+>'] = array('optin/tokens', 'matchValue'=>true);
-$route['optin/participants/<surveyid:\d+>'] = array('optin/participants', 'matchValue'=>true);
-$route['statistics_user/<surveyid:\d+>'] = array('StatisticsUser/action', 'matchValue'=>true);
+$route['optout/tokens/<surveyid:\d+>'] = array('optout/tokens', 'matchValue' => true);
+$route['optout/participants/<surveyid:\d+>'] = array('optout/participants', 'matchValue' => true);
+$route['optin/tokens/<surveyid:\d+>'] = array('optin/tokens', 'matchValue' => true);
+$route['optin/participants/<surveyid:\d+>'] = array('optin/participants', 'matchValue' => true);
+$route['statistics_user/<surveyid:\d+>'] = array('StatisticsUser/action', 'matchValue' => true);
 $route['statistics_user/action'] = 'StatisticsUser/action';
 
 //$route['<_controller:\w+>/<_action:\w+>'] = '<_controller>/<_action>';
--- a/application/controllers/OptoutController.php
+++ b/application/controllers/OptoutController.php
@@ -13,6 +13,42 @@
  *
  */
 
+/*
+    there are a few different flows through this controller
+
+    GET optout/actiontokens -> GET optout/removetokens
+
+    user vists optouturl and for older themes there is a button
+    that actually opts out via a GET request when clicked. the
+    problem with this is that some mail filters crawl the optout url
+    and perform that GET request, causing participants to be opted out
+
+    GET optout/actiontokens -> POST optout/removetoken
+
+    user vists optouturl and for newer themes there is a form
+    which POSTs to an endpoint intead. the automated filters will
+    not perform the POST request. this prevents the above problem.
+
+    POST optout/actiontokens
+
+    this is one click unsubscribe, a special post request is sent
+    to the list unsubscribe url (the same url a user would visit
+    to manually opt out, first flow above) when mail clients have
+    this available, if a user reports a message as spam, the client
+    can then offer to unsubscribe the user instead.
+
+    GET optout/participants -> GET optout/removetokens
+
+    similar to the first flow above, but opts out the participant
+    from the central participant database as well as the survey
+    and all other surveys if they request it
+
+    GET optout/participants -> POST optout/removetokens
+
+    similar to the second flow above, solves the get vs post request
+    issue. similar functionality to the previous flow.
+*/
+
 /**
  * optout
  *
@@ -34,6 +70,19 @@ class OptoutController extends LSYii_Controller
      */
     public function actiontokens()
     {
+        // handle a one click unsubscribe post request
+        // this is distinct from the "secure POST link"
+        // mentioned above which is a different endpoint
+        // per rfc8058
+        // "The target of
+        // the POST action is the same as the one in the GET action for a manual
+        // unsubscription, so this is intended to allow the same server code to
+        // handle both."
+        if(Yii::app()->request->isPostRequest) {
+            $this->oneClickUnsubscribe();
+            return;
+        }
+
         $iSurveyID     = Yii::app()->request->getQuery('surveyid');
         $sLanguageCode = Yii::app()->request->getQuery('langcode');
         $sToken        = Token::sanitizeToken(Yii::app()->request->getQuery('token'));
@@ -65,7 +114,7 @@ class OptoutController extends LSYii_Controller
         if (!isset($oToken)) {
             $sMessage = gT('You are not a participant of this survey.');
         } else {
-            if (substr((string) $oToken->emailstatus, 0, strlen('OptOut')) == 'OptOut') {
+            if ($oToken->optOutStatus) {
                 $sMessage = gT('You have already been removed from this survey.');
             } else {
                 $sMessage = gT('Please confirm that you want to opt out of this survey by clicking the button below.') . '<br>' . gT("After confirmation you won't receive any invitations or reminders for this survey anymore.");
@@ -75,6 +124,32 @@ class OptoutController extends LSYii_Controller
             $tokenAttributes = $oToken->getAttributes();
         }
         $this->renderHtml($sMessage, $oSurvey, $link, $tokenAttributes, [], $postLink ?? '');
+    }
+
+    /**
+     * Handle post request for one click unsubcribe
+     * See: https://datatracker.ietf.org/doc/html/rfc8058
+     *
+     * Note the following explicit requirement:
+     * The mail sender MUST NOT return an HTTPS redirect, since redirected
+     * POST actions have historically not worked reliably, and many browsers
+     * have turned redirected HTTP POSTs into GETs.
+     */
+    public function oneClickUnsubscribe()
+    {
+        // per rfc8058
+        // "A mail receiver can do a one-click unsubscription by performing an
+        // HTTPS POST to the HTTPS URI in the List-Unsubscribe header.  It sends
+        // the key/value pair in the List-Unsubscribe-Post header as the request
+        // body."
+        // and
+        // "The List-Unsubscribe-Post header MUST contain the single
+        // key/value pair "List-Unsubscribe=One-Click"."
+        if (Yii::app()->request->getPost('List-Unsubscribe') !== "One-Click") {
+            throw new CHttpException(400, gT('Invalid request.'));
+        }
+
+        $this->handleOptout();
     }
 
     /**
@@ -119,8 +194,6 @@ class OptoutController extends LSYii_Controller
         if (!isset($token)) {
             $message = gT('You are not a participant of this survey.');
         } else {
-            $optedOutFromSurvey = substr((string) $token->emailstatus, 0, strlen('OptOut')) == 'OptOut';
-
             $blacklistHandler = new LimeSurvey\Models\Services\ParticipantBlacklistHandler();
             $participant = $blacklistHandler->getCentralParticipantFromToken($token);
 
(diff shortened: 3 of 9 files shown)


application/models/Token.php
```diff
@@ -420,7 +420,7 @@ public function rules()
             array('lastname', 'filter', 'filter' => array(self::class, 'sanitizeAttribute')),
             array('language', 'LSYii_Validators', 'isLanguage' => true),
             array('language','in','range' => array_merge(array($this->survey->language), explode(' ', $this->survey->additional_languages)),'allowEmpty' => true,'message' => gT('Language code is invalid in this survey')),
-            array(implode(',', $this->tableSchema->columnNames), 'safe'),
+            array(implode(',', $this->tableSchema->columnNames ?? []), 'safe'),
```
mfavetti, line 423: this was to fix a php warning of null being passed to implode

application/controllers/OptoutController.php
```diff
@@ -72,6 +72,90 @@ public function actiontokens()
...
+            throw new CHttpException(404);
+        }
+
+        if(!$oSurvey->hasTokensTable) {
+            throw new CHttpException(400);
+        }
+
+        $oToken = Token::model($oSurvey->sid)->findByAttributes([
+            'token' => Token::sanitizeToken($token),
+        ]);
```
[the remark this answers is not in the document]
mfavetti: see line 138

application/controllers/OptoutController.php
```diff
@@ -175,8 +258,8 @@ public function actionremovetokens()
         if (!isset($token)) {
             $message = gT('You are not a participant in this survey.');
         } else {
-            if (substr((string) $token->emailstatus, 0, strlen('OptOut')) !== 'OptOut') {
-                $token->emailstatus = 'OptOut';
+            if (!$token->optOutStatus) {
+                $oToken->optOut();
                 $token->save();
```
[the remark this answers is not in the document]
mfavetti: whoops, will fix

application/models/services/ParticipantBlacklistHandler.php
```diff
@@ -121,7 +121,7 @@ private function optoutParticipantFromAllSurveys($participant)
             if ($survey->hasTokensTable) {
                 $token = \Token::model($survey->sid)->findByAttributes(['participant_id' => $participant->participant_id], "emailstatus <> 'OptOut'");
                 if (!empty($token)) {
-                    $token->emailstatus = 'OptOut';
+                    $token->optOut();
                     $token->save();
```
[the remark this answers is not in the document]
mfavetti: yeah I noticed this could potentially be an issue in a couple queries. here and application/controllers/admin/Token.php:L3084

change from 
emailstatus <> 'OptOut'
to
emailstatus not like 'OptOut%'
?

application/controllers/OptoutController.php
```diff
@@ -72,6 +72,90 @@ public function actiontokens()
...
+        // opt out the token if not already
+
+        // provided token is already opted out
+        if($oToken->optOutStatus) {
+            return;
+        }
+
+        // opt out the token
+        $oToken->optOut();
+        $oToken->save();
```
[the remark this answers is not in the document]
mfavetti: hmm this case isn't handled by any of the optout/optin endpoints

application/controllers/OptoutController.php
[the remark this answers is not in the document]
mfavetti: hmm need to test this

the other endpoints would suffer from the same problem if this actually happens as they just use the sanitized value directly with no check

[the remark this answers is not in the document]
mfavetti: yeah this is a problem if there are any participants with a blank token

application/models/services/ParticipantBlacklistHandler.php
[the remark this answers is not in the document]
mfavetti: okay I made changes, review this again

application/controllers/OptoutController.php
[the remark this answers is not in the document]
mfavetti: okay I made changes, review this again

application/controllers/OptoutController.php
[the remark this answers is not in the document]
mfavetti: okay I made changes, review this again

application/controllers/OptoutController.php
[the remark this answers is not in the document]
mfavetti: okay I made changes, review this again

application/models/services/ParticipantBlacklistHandler.php
```diff
@@ -119,9 +119,9 @@ private function optoutParticipantFromAllSurveys($participant)
         $optedoutSurveyIds = [];
         foreach ($surveys as $survey) {
             if ($survey->hasTokensTable) {
-                $token = \Token::model($survey->sid)->findByAttributes(['participant_id' => $participant->participant_id], "emailstatus <> 'OptOut'");
+                $token = \Token::model($survey->sid)->findByAttributes(['participant_id' => $participant->participant_id], "emailstatus NOT LIKE 'OptOut%'");
                 if (!empty($token)) {
-                    $token->emailstatus = 'OptOut';
+                    $token->optOut();
                     $token->save();
```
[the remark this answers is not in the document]
mfavetti: yeah inconsistent, but one is opting out the participant from all surveys they are not already opted out (emailstatus doesn't start with OptOut) and the other is getting participants to email.

for the second one there is a switch labeled "Bypass participants with failing email addresses:" if its Off, it will only email emailstatus OK, if its On, it will email any emailstatus other than starting with OptOut (bounced, OK, NULL, etc)

c-schmitz:
This probably duplicates work from #4831

mfavetti:
@c-schmitz should I redo this on top of your branch then? or wait until yours is merged and then redo it?

c-schmitz:
I'd say rather wait until mine is merged.

application/config/routes.php
mfavetti, line 1: these are changes from one of the static analysis tools

application/controllers/admin/Tokens.php
```diff
@@ -3082,7 +3082,7 @@ protected function getSQLemailstatuscondition()
         if ($request->getPost('bypassbademails') == '1') {
             return "emailstatus = 'OK'";
         } else {
-            return "emailstatus <> 'OptOut' OR emailstatus IS NULL";
+            return "emailstatus NOT LIKE 'OptOut%' OR emailstatus IS NULL";
```
Shnoulle, line 3085: Work on all DB ?

application/controllers/OptoutController.php
```diff
@@ -34,6 +70,19 @@ class OptoutController extends LSYii_Controller
...
     {
+        // handle a one click unsubscribe post request
+        // this is distinct from the "secure POST link"
+        // mentioned above which is a different endpoint
+        // per rfc8058
+        // "The target of
+        // the POST action is the same as the one in the GET action for a manual
+        // unsubscription, so this is intended to allow the same server code to
+        // handle both."
+        if(Yii::app()->request->isPostRequest) {
```
Shnoulle, line 81: See https://github.com/LimeSurvey/LimeSurvey/commit/8bf8045c13839d1349caa47532626c79ebf010de#diff-19b3b994376243672d485bdbb3675f7acb67b4cab421b7311a5109a896c0aac9R29-R73

<s>We already moved to an opt-out with POST value. Need to use this one, no ?</s>

**False and bad idea**

application/config/internal.php
```diff
@@ -161,6 +161,7 @@
                 'rest',
                 'admin/remotecontrol',
                 'plugins/unsecure',
+                'optout/tokens',
```
Shnoulle, line 164: No ! Better to use anorthe url !! `'optout/oneclick',`

application/controllers/OptoutController.php
Shnoulle: And better to use another action

application/core/LimeMailer.php
```diff
@@ -343,6 +343,23 @@ public function setToken($token)
             $this->addAddress($sEmailaddress, $oToken->firstname . " " . $oToken->lastname);
         }
         $this->addCustomHeader("X-tokenid", $oToken->token);
+
+        // add list unsubscribe header
+        // see: https://datatracker.ietf.org/doc/html/rfc2369
+        // the url is the same as OPTOUTURL replacement field
+        $oneClickURL = App()->getController()->createAbsoluteUrl(
```
Shnoulle, line 350: My opinion here : 

We need https://datatracker.ietf.org/doc/html/rfc8058#section-4 valid DKIM
Add some https://datatracker.ietf.org/doc/html/rfc8058#section-6 extra _opaque identifier_

Maybe we need 2 options :  

- Onclick with extra POST action : `Yii::app()->createUrl('optout/token', array('surveyid' => $iSurveyID, 'langcode' => $sBaseLanguage, 'token' => $sToken));`
- By option (Survey ?) `Yii::app()->createUrl('optout/oneclick', array('surveyid' => $iSurveyID, 'langcode' => $sBaseLanguage, 'token' => $sToken));`

Shnoulle (changes requested):
The url **must** be different, and you must update to the last version of LimeSurvey.

Other are discussion.

application/core/LimeMailer.php
Shnoulle: Seems we can have limit n the Opt-OUT in days

application/controllers/OptoutController.php
mfavetti: i did wait until @c-schmitz merged his changes

however these endpoints are two different functionalities. carsten's is meant to prevent the spam filtering from inadvertently unsubscribing tokens by crawling the unsubscibe link and making the get request that was previously used. so its a form that makes a post request rather than a button that makes a get request.

this PR adds support for one click unsubscribe, which allows email clients to unsubscribe users with a single post request (used to unsubscribe instead of reporting spam)

the relevant spec, rfc8058 states:

"The target of the POST action is the same as the one in the GET action for a manual unsubscription, so this is intended to allow the same server code to handle both."

https://datatracker.ietf.org/doc/html/rfc8058#section-3.2

this maintains backwards compatibility with https://datatracker.ietf.org/doc/html/rfc2369, the spec that describes the List-Unsubscribe header

so basically it needs to be the same endpoint as the OPTOUTURL placeholder uses

application/config/internal.php
mfavetti: needs to be the same endpoint as a get unsubscribe request, per the spec

application/core/LimeMailer.php
mfavetti: yes, DKIM is required per spec, but that is up to the configuration of the mail server used to send it. i see in limemailer that there is already support to sign via dkim, but it could be signed downstream also

the token is the opaque identifier, otherwise we would need to create some new table of identifiers that maps to a particular sid/token and I think using the token meets the requirements of the spec

application/controllers/admin/Tokens.php
mfavetti: just checked for mysql/mariadb/mssql/postgres and I believe they all do

application/config/internal.php
mfavetti: perhaps we should change the action of @c-schmitz new form to be the one-click endpoint rather than adding a separate one.

then we would want CORS for the manual way, but no CORS for the one click

application/controllers/OptoutController.php
mfavetti: Hmm, the rfc actually talks about this

"This document describes a method for signaling a one-click function
   for the List-Unsubscribe email header field.  The need for this
   arises out of the actuality that mail software sometimes fetches URLs
   in mail header fields, and thereby accidentally triggers
   unsubscriptions in the case of the List-Unsubscribe header field."

so the implementation is the same for both things

i'm thinking we move the form action to the endpoint I've created and handle cors vs no-cors there

mfavetti: actually no

different issues actually because this is specifically for fetching the URL in the list-unsubscribe header, which wasn't even present before

so can keep both endpoints

application/config/internal.php
mfavetti: nm this last bit, see above

mfavetti:
> The url **must** be different, and you must update to the last version of LimeSurvey.
> 
> Other are discussion.

see my other response, spec says the url for get/post unsubscribe is the same

not sure what you mean on update to latest version of limesurvey. the PR is in line with develop-minor (or it was a couple hours ago) don't why the "has conflicts" tag wasn't automatically removed. says no conflicts at the bottom anyways

application/core/LimeMailer.php
mfavetti: > Seems we can have limit n the Opt-OUT in days

are you suggesting some time limit on allowing the opt out? based on how long ago the invitation/reminder was sent?

mfavetti: >Maybe we need 2 options :
>
>Onclick with extra POST action : Yii::app()->createUrl('optout/token', array('surveyid' => $iSurveyID, 'langcode' => >$sBaseLanguage, 'token' => $sToken));
>By option (Survey ?) Yii::app()->createUrl('optout/oneclick', array('surveyid' => $iSurveyID, 'langcode' => >$sBaseLanguage, 'token' => $sToken));

so a survey setting to enable the one click headers being added? then if you don't have dkim you disable? is that what you mean here?

c-schmitz:
There is still a codesniff error in the CI.

mfavetti:
> There is still a codesniff error in the CI.

okay will fix tmrw

application/controllers/OptoutController.php
Shnoulle: Yes, i review : we need a different endpoint for oneclick url : mandatory !

application/core/LimeMailer.php
Shnoulle: > so a survey setting to enable the one click headers being added? then if you don't have dkim you disable? is that what you mean here?

Not exactly : 

1. Allways add the optout url , by default to optout/token
2. By option : set it to one click to optout/onclick with post header

Shnoulle:
> see my other response, spec says the url for get/post unsubscribe is the same

No, you can have some in HTML content different than the header.

1. In HTML content : always use optout/token
2. In Header allow by survey option
     - optout/token or
     - optout/oneclick

then only optout/oneclick need to be out of CRSF.


Always add `List-Unsubscribe` , `List-Unsubscribe` url and `List-Unsubscribe-Post` depending on some condition (can be in global mail settings).

application/controllers/OptoutController.php
mfavetti: can you explain why?

i will change, i just don't understand

Shnoulle:
To be more clear

Always add `List-Unsubscribe` , `List-Unsubscribe` url and `List-Unsubscribe-Post` depending on some condition (can be in global mail settings: seems the **best** in fact).

application/controllers/OptoutController.php
Shnoulle: One specific action needs a specific URL.

Here : you check if it's a POST request and go to another action. Then it's clearly another action.

mfavetti: I think the RFC says the same url should be a manual unsubscription when it's a GET request and an automatic unsubscription when it's a POST request

i looked at routing to a different controller action depending on the verb, but it only works for Yii path routing, not Yii get routing

what if I add a new route/action which shows the new post form carsten made for GET and does the one click unsubscribe for POST (exempt from CORS) and leave the current endpoints alone? does that work for you?

i do like the idea of a global mail setting for the header so users can toggle based on if it would be signed by dkim

Shnoulle: > I think the RFC says the same url should be a manual unsubscription when it's a GET request and an automatic unsubscription when it's a POST request

I don't find this in RFC ? 

And in this case : i prefer to reroute to /token action if it's not a post request. 

> i do like the idea of a global mail setting for the header so users can toggle based on if it would be signed by dkim

yes, no need survey option, just global email , true by default for adding List-Unsubscribe and List-Unsubscribe-Post , unsure we need 3 options

Shnoulle:
PS : https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/OptoutController.php#L152 optout/removetokens do action without POST ;)

Shnoulle:
@mfavetti maybe the solution is adding some global settings is the best .

Then we can easily have 2 options 

1. Add  List-Unsubscribe with
    - Yes
    - Yes with List-Unsubscribe-POST
    - No 
2. Keep [legacy optout](https://github.com/LimeSurvey/LimeSurvey/blob/76f08dffbb2ed42bc154642002b11972f8575834/application/controllers/OptoutController.php#L148) 


Else : maybe remove actionremovetoken from CRSF **but** control CRSF manually if List-Unsubscribe is not One-Click

application/models/Token.php
Shnoulle: But if $this->tableSchema->columnNames is null : there are no token table ? Must broke before, no ?

PR #4870, opened 2026-04-13
Dev Task: Remove obsolete ajax mode

c-schmitz:
## Summary by CodeRabbit

* **Refactor**
  * Simplified theme configuration update process by removing automatic setting modifications prior to applying user-provided settings.
  * Removed ajax mode defaults from theme configurations.
  * Removed related helper methods and code paths.

* **Tests**
  * Updated test suite to remove ajax mode–specific test cases and align coverage with simplified configuration handling.


Shnoulle:
That terrible code caused a lot of trouble!

application/models/TemplateConfig.php
```diff
@@ -1068,6 +1068,11 @@ public static function convertOptionsToJson($options)
     {
         $optionsArray = [];
         foreach ($options as $option => $optionValue) {
+            // Skip deprecated ajaxmode option from old theme configs.
+            // Old themes may still have this in their config.xml, but it's no longer used.
+            if ($option === 'ajaxmode') {
```
Shnoulle, line 1073: Do we really need to remove ?

application/helpers/update/updates/Update_703.php
```diff
@@ -0,0 +1,25 @@
...
+use TemplateConfiguration;
+
+class Update_703 extends DatabaseUpdateBase
+{
+    public function up()
+    {
+        // Remove deprecated ajaxmode from all template configurations
+        // (whether from old migrations or fresh installs)
+        $themes = TemplateConfiguration::model()->findAll();
+        foreach ($themes as $theme) {
```
Shnoulle, line 14: Do we really need to remove it ?
Just don't use it

Shnoulle (commented):
Can we just not use it anymore ? 
No need to update ?

It disappear each time a theme option was saved.

c-schmitz:
@copilot resolve the merge conflicts in this pull request

c-schmitz:
> Can we just not use it anymore ? No need to update ?
> 
> It disappear each time a theme option was saved.

Yes, removed the unnecessary db update.

application/models/TemplateConfig.php
Shnoulle: @c-schmitz you test remove this part ?

I mean aSurvey->options['anything'] can still be here , just `anything` is not used.
And next time theme option was saved : it was deleted.

Shnoulle (commented):
I still think we don't have to remove ajaxmode from TemplateConfiguration

1. Not used, then don't really care
2. Removed automatically when savec Theme option

The interest: We really remove ajaxmode and don't think in 2 years _what is this option  and why we removed it ?_

Shnoulle (approved):

PR #4879, opened 2026-04-14
Dev: fix failing ranking test

karcstefan:
Dev: fix failing ranking test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ranking-question field mapping and subquestion identifier handling to ensure accurate display and processing.
  * Refined ranking-question count/validation so subquestions are counted and tracked more precisely.

* **Tests**
  * Enabled and updated ranking-question functional tests to run to completion and more reliably verify subquestion behavior.


Ahmedazeem98 (changes requested):

application/helpers/expressions/em_manager_helper.php
```diff
@@ -6019,12 +6019,12 @@ public function _ValidateQuestion($questionSeq, $force = false)
...
             /* $sgqa and subQrelInfo are not the same information */
-            if ($qInfo['type'] == \Question::QT_R_RANKING) {
+            if ($qInfo['type'] == 'R') {
                 /** @var integer counter to have current rank number (subquestion) */
                 $iCountRank = 0;
 
-                /** @var integer Get total of answers (all potential answers) * */
-                $answersCount = \Question::model()->count('parent_qid = :parent_qid', [':parent_qid' => $qid]);
+                /** @var integer Get total of subquestions (all potential answers) * */
+                $answersCount = \Question::model()->count('parent_qid = :qid', [':qid' => $qid]);
```
lajosarpad, line 6027: This is a correct fix

application/helpers/expressions/em_manager_helper.php
```diff
@@ -3656,7 +3656,7 @@ public function setVariableAndTokenMappingsForExpressionManager($surveyid, $forc
                             $question = $this->gT('Other:');
                         }
                     }
-                    $sqsuffix = '_' . $fielddata['qid'];
+                    $sqsuffix = '_' . $fielddata['aid'];
```
lajosarpad, line 3659: I'm not sure about this fix, it is the title of a subquestion as per createFieldMap, but for ranking advanced it was not working. We have an alternative here https://github.com/LimeSurvey/LimeSurvey/pull/4878/changes that uses ranking index for the relevance script

application/helpers/common_helper.php
```diff
@@ -1971,6 +1971,7 @@ function createFieldMap($survey, $style = 'short', $force_refresh = false, $ques
                     "sqid" => $abrow['qid'],
                     "aid" => $abrow['title'],
                     "suffix" => '_S' . ($aQIDReplacements[$abrow['qid']] ?? $abrow['qid']),
+                    'csuffix' => $i,
```
lajosarpad, line 1974: I'm not sure about this fix, if csuffix is condition suffix, then the ranking index makes sense, but I'm unsure what the issue was and what this fixed. But it does not seem to cause any problem and it may be beneficial

tests/functional/frontend/RankingArrayFilterMaxColumnTest.php
```diff
@@ -82,8 +80,8 @@ public function testRanking()
             sleep(1);
 
             $subQuestions = $this->getSubquestions($survey->groups[1]->questions[0]->qid);
-            /** @var string List item id to first subquestion. */
-            $sgqa1 = $this->getItemListId($survey) . '_S' . reset($subQuestions)->qid;
+            /** @var string List item id to first subquestion (title "1"). */
+            $sgqa1 = $this->getItemListId($survey) . '_S' . $subQuestions['1']->qid;
```
lajosarpad, line 84: This change suggests that you likely did not get the subquestions via getSubquestions in the right order. It is because it's agnostic to the order of questions, like:

```
    protected function getSubquestions(int $qid)
    {
        $rawQuestions = \Question::model()->findAll("parent_qid = :qid", [":qid" => $qid]);
        $questions = [];
        foreach ($rawQuestions as $rawQuestion) {
            $questions[$rawQuestion->title] = $rawQuestion;
        }
        return $questions;
    }

```

and perhaps enforcing the order would fix the issue with the order:

```
    protected function getSubquestions(int $qid)
    {
        $rawQuestions = \Question::model()->findAll([condition => "parent_qid = :qid", 'params' => [":qid" => $qid], 'order' => 'question_order']);
        $questions = [];
        foreach ($rawQuestions as $rawQuestion) {
            $questions[$rawQuestion->title] = $rawQuestion;
        }
        return $questions;
    }
```

(untested)

PR #4900, opened 2026-04-21
Fix issue #20494: Conditional Remote Code Execution via Question Theme Path Traversal and Custom Twig Extension Injection

gabrieljenik:
## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin and question-theme uploads now validate names from config files and abort installation with an error when names are invalid, preventing unsafe or unintended directories.
* **New Features**
  * Added a reusable path-component validation utility enforcing character rules and a 150-character limit for installer names.
* **Tests**
  * Added unit tests covering plugin name, theme name, and path-component validation (allowed and disallowed cases).

### DIFF
--- a/application/helpers/sanitize_helper.php
+++ b/application/helpers/sanitize_helper.php
@@ -586,7 +586,9 @@ function sanitize_alphanumeric($value)
  */
 function validate_path_component($string)
 {
-    $string = (string) $string;
+    if (!is_string($string)) {
+        return false;
+    }
 
     return !(
         $string === ''
--- a/application/helpers/sanitize_helper.php
+++ b/application/helpers/sanitize_helper.php
@@ -586,9 +586,11 @@ function sanitize_alphanumeric($value)
  */
 function validate_path_component($string)
 {
+    $string = (string) $string;
+
     return !(
         $string === ''
-        || strpos((string) $string, '.') === 0
+        || strpos($string, '.') === 0
         || preg_match('/[\\\\\/]/', $string) === 1
         || preg_match('/[\x00-\x1F\x7F]/', $string) === 1
     );
--- a/application/helpers/sanitize_helper.php
+++ b/application/helpers/sanitize_helper.php
@@ -577,9 +577,9 @@ function sanitize_alphanumeric($value)
 /**
  * Validate that a value is safe to use as a single filesystem path component.
  *
- * This rejects empty values, path separators, "." / ".." and ASCII control
- * characters so callers can safely append the value to a trusted base path
- * without worrying about directory traversal.
+ * This rejects empty values, leading-dot names, path separators and ASCII
+ * control characters so callers can safely append the value to a trusted
+ * base path without worrying about directory traversal or hidden directories.
  *
  * @param string $string
  * @return bool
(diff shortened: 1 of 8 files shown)


application/helpers/sanitize_helper.php
```diff
@@ -573,3 +573,25 @@ function sanitize_alphanumeric($value)
...
+ *
+ * This rejects empty values, leading-dot names, path separators and ASCII
+ * control characters so callers can safely append the value to a trusted
+ * base path without worrying about directory traversal or hidden directories.
+ *
+ * @param string $string
+ * @return bool
+ */
+function validate_path_component($string)
+{
```
Shnoulle, line 588: Why not use https://github.com/LimeSurvey/LimeSurvey/blob/ba5bbf0a588901f40c5096bf5246dc1643e5937e/application/helpers/sanitize_helper.php#L223 ?

application/helpers/sanitize_helper.php
```diff
@@ -573,3 +573,25 @@ function sanitize_alphanumeric($value)
...
+    $string = (string) $string;
```
Shnoulle: Do not cast if it's not a string directly; must return false.

application/helpers/sanitize_helper.php
```diff
@@ -573,3 +573,25 @@ function sanitize_alphanumeric($value)
...
+
+    return !(
+        $string === ''
+        || strpos($string, '.') === 0
+        || preg_match('/[\\\\\/]/', $string) === 1
+        || preg_match('/[\x00-\x1F\x7F]/', $string) === 1
+    );
```
Shnoulle, line 598: We need more : https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html#filename-safety

_Implement a maximum length_  needed for example : https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

The question is more, what restriction must do ?

Shnoulle (commented):
I think we must restrict more in filename :)

Else all other part are OK !

Shnoulle:
Config name must not be like directory name? Or zip name when upload ?

Maybe we can just : check all directory name in theme directory, and if config->name not in the array : it's invalid

application/helpers/sanitize_helper.php
gabrieljenik: We must be carefull here in terms of restrictions. We don't want to make current installations to fail.
Ideally, would only allow letters and numbers.

Shnoulle: > We must be carefull here in terms of restrictions. We don't want to make current installations to fail. Ideally, would only allow letters and numbers.

Yes, it's too bad we didn't set restrictions from the start ;)

Does it work if the name is different than the directory name ?
https://github.com/LimeSurvey/LimeSurvey/pull/4900#issuecomment-4294363539

With a plugin, the name must be equal to the directory name (and PHP filename).

Shnoulle: > Does it work if the name is different than the directory name ?

Yes ..

application/libraries/PluginManager/PluginManager.php
```diff
@@ -424,10 +427,26 @@ public function getPluginFolder(\ExtensionConfig $config, $pluginType)
...
+    /**
+     * Validate that a plugin name can safely serve as folder, file and class name.
+     * Plugin names are used as a flat class identifier throughout the plugin manager.
+     *
+     * @param string $pluginName
+     * @return bool
+     */
+    public function validatePluginName($pluginName)
+    {
+        return preg_match('/^[A-Za-z_][A-Za-z0-9_]*$/', (string) $pluginName) === 1;
```
Shnoulle, line 447: Then you choose really safer :)

Shnoulle (approved):

PR #4910, opened 2026-04-24
Fixed issue [security] #20495: Unrestricted getConfig() access in Twig sandbox

Shnoulle:
Dev: need template edit access

## Summary by CodeRabbit

* **New Features**
  * Added two admin configuration options to control which template configuration keys can be accessed: an optional allowlist and an extra denylist.
  * Runtime enforcement now blocks retrieval of restricted or sensitive settings by combining a built-in denylist with the configurable denylist and optional allowlist for finer-grained exposure control.

### DIFF
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -884,5 +884,14 @@ $config['allow_unserialize_attributedescriptions'] = false;
 // @see https://www.php.net/unserialize
 $config['allow_unserialize_attachments'] = false;
 
+// Allow to use a allow list for getConfig function in twig
+// used only if it's an array, else use a fixed denylist and the extra denylist in config
+// @see https://bugs.limesurvey.org/view.php?id=20495
+$config['twig_getConfig_allowlist'] = null;
+
+// Allow to add more global settings to a deny for getConfig function in twig
+// @see https://bugs.limesurvey.org/view.php?id=20495
+$config['twig_getConfig_extradenylist'] = [];
+
 return $config;
 //settings deleted
--- a/application/config/config-defaults.php
+++ b/application/config/config-defaults.php
@@ -886,10 +886,14 @@ $config['allow_unserialize_attachments'] = false;
 
 // Allow to use a allow list for getConfig function in twig
 // used only if it's an array, else use a fixed denylist and the extra denylist in config
+// If you want to allow 'securesetting', set this to ['securesetting'].
+// All settings except 'securesetting' will be denied. 'securesetting' will be allowed.
 // @see https://bugs.limesurvey.org/view.php?id=20495
 $config['twig_getConfig_allowlist'] = null;
 
 // Allow to add more global settings to a deny for getConfig function in twig
+// LimeSurvye already disallow most unsecure settings.
+// If you want to disallow 'unsecuresetting', set this to ['unsecuresetting']
 // @see https://bugs.limesurvey.org/view.php?id=20495
 $config['twig_getConfig_extradenylist'] = [];
 
--- a/application/core/LS_Twig_Extension.php
+++ b/application/core/LS_Twig_Extension.php
@@ -660,7 +660,7 @@ class LS_Twig_Extension extends AbstractExtension
      */
     public static function getConfig($name)
     {
-        /* if allowlist is an array, use it */
+        /* if allowlist is an array, use it but only for deny; not for allowing */
         $allowlist = App()->getConfig('twig_getConfig_allowlist');
         if (is_array($allowlist) && !in_array($name, $allowlist, true)) {
             return false;
@@ -701,13 +701,13 @@ class LS_Twig_Extension extends AbstractExtension
             'ssl_disable_alert',
             'ssl_emergency_override',
             'registrationEmailDelay',
-             /* autentication related */
+             /* authentication related */
             'auth_webserver',
             'auth_webserver_user_map',
             'auth_webserver_autocreate_user',
             'auth_webserver_autocreate_profile',
             'auth_webserver_autocreate_permissions',
-            /* connexion related */
+            /* connection related */
             'proxy_host_name',
             'proxy_host_port',
             'reverseProxyIpHeader',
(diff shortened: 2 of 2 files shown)


application/core/LS_Twig_Extension.php
```diff
@@ -652,9 +652,81 @@ public static function lightencss($cssColor, $grade = 10, $alpha = 1)
...
+            'corequestionthemerootdir',
+            'styledir',
+            'questiontypedir',
+            'userthemerootdir',
+            'usertwigextensionrootdir',
+            'customquestionthemedir',
+            'userquestionthemerootdir',
+            'userfontsrootdir',
+            'lsadminmodulesrootdir',
+        ];
```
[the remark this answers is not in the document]
Shnoulle: @gabrieljenik or @c-schmitz thought here ?

For debug : easy to check, just go to a 404, debugsql is more easy, shown at each page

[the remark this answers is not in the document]
c-schmitz: Just add them all, I think.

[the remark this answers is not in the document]
Shnoulle: > Just add them all, I think.

We can not remove it, hard coded :). I leave debug and debugsql

application/core/LS_Twig_Extension.php
```diff
@@ -652,9 +652,104 @@ public static function lightencss($cssColor, $grade = 10, $alpha = 1)
...
+            'proxy_host_name',
+            'proxy_host_port',
+            'reverseProxyIpHeader',
+            'reverseProxyIpAddresses',
+            /* Setting keys and password */
+            'googleMapsAPIKey',
+            'ipInfoDbAPIKey',
+            'googleanalyticsapikey',
+            'googletranslateapikey',
+            'GeoNamesUsername',
```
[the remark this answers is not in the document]
Shnoulle: Config came from DB too, are in views : https://github.com/LimeSurvey/LimeSurvey/tree/master/application/views/admin/globalsettings

application/config/config-defaults.php
```diff
@@ -884,5 +884,14 @@
 // @see https://www.php.net/unserialize
 $config['allow_unserialize_attachments'] = false;
 
+// Allow to use a allow list for getConfig function in twig
+// used only if it's an array, else use a fixed denylist and the extra denylist in config
+// @see https://bugs.limesurvey.org/view.php?id=20495
```
gabrieljenik, line 891: I would add a bit more of documentation, like an example.
If not, later, for understanding these settings, we need to go to the source code.
In this case, this is simple, but never hurts.

application/config/config-defaults.php
```diff
@@ -884,5 +884,14 @@
...
+$config['twig_getConfig_allowlist'] = null;
+
+// Allow to add more global settings to a deny for getConfig function in twig
+// @see https://bugs.limesurvey.org/view.php?id=20495
```
gabrieljenik, line 897: I would add a bit more of documentation, like an example.
If not, later, for understanding these settings, we need to go to the source code.
In this case, this is simple, but never hurts.

application/config/config-defaults.php
Shnoulle: It's already for theme developer, using {{ getConfig('setting') }}

You mean something like this : 

`// If you want to disallow all settings except 'securesetting', set this to ['securesetting']`

application/config/config-defaults.php
Shnoulle: `// If you want to disallow 'unsecuresetting', set this to ['unsecuresetting']`

application/core/LS_Twig_Extension.php
```diff
@@ -652,11 +652,113 @@ public static function lightencss($cssColor, $grade = 10, $alpha = 1)
...
+     * @param string $name
+     * @return mixed
+     */
+    public static function getConfig($name)
     {
-        return Yii::app()->getConfig($item);
+        /* if allowlist is an array, use it but only for deny; not for allowing */
+        $allowlist = App()->getConfig('twig_getConfig_allowlist');
+        if (is_array($allowlist) && !in_array($name, $allowlist, true)) {
+            return false;
```
Shnoulle, line 666: Unsure on the way here,

Maybe just `return !in_array($name, $allowlist, true);` then developer can allow denyed by LimeSurvey code settings?

application/config/config-defaults.php
gabrieljenik: // If you want to allow 'securesetting', set this to ['securesetting'].
// All settings except 'securesetting' will be denied. 'securesetting' will be allowed.

application/core/LS_Twig_Extension.php
gabrieljenik: > then developer can allow denyed by LimeSurvey code settings
Not sure I follow.

What I am thinking now is that the norm should be to only allow by a few things, by exception.
By having all conditions to exit with false, although now should work, you are not following that norm, as at the end, you are returning the value.

I would make the allowlist to be reviewed at the bottom:
```
if (is_array($allowlist) && in_array($name, $allowlist, true)) {
            return App()->getConfig($name);;
}

// If got to here, return false.
```

That way, you are returning false all time, except a few occasions.
I feel it is less error prone, a bit more secure...

Shnoulle: Here : you mean more: use only a core allow list ?

Then : 

1. If it's core allow list : return
2. If it's in a user allow list (in config.php) : return
3. return false

Core allow list : adminmail, adminname, sitename and maybe some other
User allow list : can choose specific one only if needed

I create another PR with such solution ?

Shnoulle: @gabrieljenik : use deny by default solution in https://github.com/LimeSurvey/LimeSurvey/pull/4917

Shnoulle:
Alternative solution: deny by default: https://github.com/LimeSurvey/LimeSurvey/pull/4917

gabrieljenik:
> Alternative solution: deny by default: #4917

I think the deny alternative is simpler, right?

c-schmitz:
At this point deny by default seems simpler.

Shnoulle:
> At this point deny by default seems simpler.

My original idea was not to block certain uses of getConfig.
Note: This will also apply to Twig on the admin side.

PR #4911, opened 2026-04-27
LE-732: statistics performance

haythemch2:
# Thank you for contributing to LimeSurvey!

To make our work easier, please make sure to follow the instructions below.

## Guidelines

- A pull request to LimeSurvey can either be a **bug fix**, a **new feature**, or an **internal development fix** (refactoring etc).
- For bug fixes and new features, you must include the number to the Mantis issue from [bugs.limesurvey.org](https://bugs.limesurvey.org). If no issue exists yet, please create it.
- Make sure to write down exactly how to reproduce a bug.
- For smaller internal changes, a Mantis issue is not necessary, but bigger refactoring tasks should always be discussed in a Mantis issue before implementation.

## Branch policy

- **Fixed issues** should always go to the **master** branch, UNLESS they fix an issue in a yet unreleased feature in the develop branch.
- **New features** should always go to the **major-develop** or ***minor-develop** branches. For more information about release schedule and code requirements, please see the following manual pages:
  - [LimeSurvey roadmap - Current release schedule](https://manual.limesurvey.org/LimeSurvey_roadmap#Current_release_schedule)
  - [How to contribute new features](https://manual.limesurvey.org/How_to_contribute_new_features)

## PR type

> Keep one of the below lines and delete the others.

Fixed issue #<Mantis issue number>: <Description>
New feature #<Mantis issue number>: <Description>
Dev: <Description for a change that's neither a feature nor a bug>

## Summary by CodeRabbit

* **Refactor**
  * Batch response-count queries and centralize filtered query construction into single aggregated SQL operations.
  * Cache per-survey total response counts to avoid repeated calculations.
  * Build chart legends and data in bulk, using batched counts for subquestions and options.
  * Standardize handling of answered, “other”/comment, and no-answer counts for more consistent charts.

### DIFF
--- a/application/models/services/SurveyStatistics/Charts/Questions/Processors/AbstractQuestionProcessor.php
+++ b/application/models/services/SurveyStatistics/Charts/Questions/Processors/AbstractQuestionProcessor.php
@@ -286,6 +286,7 @@ abstract class AbstractQuestionProcessor
             }
 
             if ($addNoAnswer) {
+                $legend[] = 'NoAnswer';
                 $items[] = [
                     'key' => 'NoAnswer',
                     'title' => 'No Answer',
--- a/application/models/services/SurveyStatistics/Charts/Questions/Processors/AbstractQuestionProcessor.php
+++ b/application/models/services/SurveyStatistics/Charts/Questions/Processors/AbstractQuestionProcessor.php
@@ -58,12 +58,6 @@ abstract class AbstractQuestionProcessor
     private $maxId = null;
 
     /**
-     * Cache for getCountsByColumn() results, keyed by fieldName.
-     * @var array<string, array<string, int>>
-     */
-    private array $columnCountsCache = [];
-
-    /**
      * Build the identifier for current question.
      *
      * Ex: Qqid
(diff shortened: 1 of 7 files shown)


application/models/services/SurveyStatistics/Charts/Questions/Processors/AbstractQuestionProcessor.php
```diff
@@ -160,63 +165,138 @@ public function getAggregateResponses($fields, $params)
...
+                    'value' => (int)($row[$aliasMap[$field]['blank']] ?? 0),
+                ];
+            }
+
+            $output[$field] = [$legend, $items];
         }
 
-        return [$legend, $items];
+        return $output;
     }
```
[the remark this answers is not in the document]
haythemch2: fixed

application/models/services/SurveyStatistics/Charts/Questions/Processors/AbstractQuestionProcessor.php
```diff
@@ -117,38 +117,43 @@ public function setMaxId(?int $maxId): AbstractQuestionProcessor
...
+    }
 
-        $this->applyFilters($criteria);
+    /**
+     * Returns the count of null/empty responses for a column.
+     *
+     * @param string $fieldName Column name in the response table
+     * @return int
+     */
+    protected function countFieldNullResponses(string $fieldName): int
```
lajosarpad: Let's change this to

```
protected function countFieldNullResponses(int $sid, string $fieldName): int
{
    return AbstractQuestionProcessor::getTotalCount($sid) - $this->countFieldResponses($fieldName);
}
```

Advantage: We won't need to separately compute the null values. In order to use the full potential of this idea, we'll need to

- make sure `getTotalCount` is implemented in a manner that it computes `select count(*) from {{responses_123456}}`
- make sure `getTotalCount` caches the result and returns the cached value if it exists
- make sure `countFieldResponses` also caches the result after first computation

Doing

```
select count(*)
from yourtable
```

exactly once and then subtract a certain number already computed from it is quicker than doing

```
select case when ... end
from your table
```

per field, because:

- field count could be much larger than 1 and counting for nulls after knowing how many non-nulls you have is getting through the same records again to figure out a metric you can already infer from the results you already had
- doing a case-when is slower per record than not doing it, so when it's not necessary, it's better avoided

application/models/services/SurveyStatistics/Charts/Questions/Processors/AbstractQuestionProcessor.php
```diff
@@ -160,63 +165,139 @@ public function getAggregateResponses($fields, $params)
...
-        if ($this->minId !== null) {
-            $criteria->compare('id', '>=' . (int)$this->minId);
+        $db = $this->getDb();
+        ['table' => $table, 'where' => $where, 'params' => $params] = $this->buildFilteredQuery();
+        $selects = [];
+
+        foreach ($fieldNames as $i => $field) {
+            $col = $db->quoteColumnName($field);
+            $alias = $db->quoteColumnName('_ctn' . $i);
+            $selects[] = "SUM(CASE WHEN $col IS NOT NULL AND $col <> '' THEN 1 ELSE 0 END) AS $alias";
```
lajosarpad: The

```
SUM(CASE WHEN $col IS NOT NULL AND $col <> '' THEN 1 ELSE 0 END)
```

is used twice, so if a change is needed, chances are, we'll fix one and forget the other. Let's create a parameterized function that you could use like

```
$selects[] = $this->getColSumClause($col, $alias);
```

lajosarpad (changes requested):
Looks pretty good, almost ready to be merged, but we'll need to do some adjustments.

lajosarpad (approved):

PR #4919, opened 2026-04-28
Fixed issue [security] #20505: Stored XSS in Limesurvey via Global Settings (reported by masquerad3r )

Shnoulle:
Dev: hard filter all load of googleMapsAPIKey
Dev: hard filter on save and show alert and log
Dev: see https://docs.cloud.google.com/docs/authentication/api-keys#components

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation/sanitization of Google API keys to remove invalid characters, reducing configuration errors and preventing broken map integrations.
  * Admin save now shows a visible warning when a submitted API key is altered by sanitization and records a server-side warning log.

* **Chores**
  * Unified API key handling across the app and added centralized sanitization for consistent, safer use of Google Maps-related keys.

### DIFF
--- a/application/controllers/admin/globalsettings.php
+++ b/application/controllers/admin/globalsettings.php
@@ -438,9 +438,19 @@ class GlobalSettings extends SurveyCommonAction
         SettingGlobal::setSetting('bPdfQuestionBold', sanitize_int(Yii::app()->getRequest()->getPost('bPdfQuestionBold')));
         SettingGlobal::setSetting('bPdfQuestionBorder', sanitize_int(Yii::app()->getRequest()->getPost('bPdfQuestionBorder')));
         SettingGlobal::setSetting('bPdfResponseBorder', sanitize_int(Yii::app()->getRequest()->getPost('bPdfResponseBorder')));
-        SettingGlobal::setSetting('googleMapsAPIKey', Yii::app()->getRequest()->getPost('googleMapsAPIKey'));
-        SettingGlobal::setSetting('googleanalyticsapikey', Yii::app()->getRequest()->getPost('googleanalyticsapikey'));
-        SettingGlobal::setSetting('googletranslateapikey', Yii::app()->getRequest()->getPost('googletranslateapikey'));
+        /* @see https://docs.cloud.google.com/docs/authentication/api-keys#components */
+        $googleapikeys = ['googleMapsAPIKey', 'googleanalyticsapikey', 'googletranslateapikey'];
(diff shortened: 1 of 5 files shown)


application/controllers/admin/globalsettings.php
```diff
@@ -438,9 +438,19 @@ private function saveSettings()
...
+        /* @see https://docs.cloud.google.com/docs/authentication/api-keys#components */
+        $googleapikeys = ['googleMapsAPIKey', 'googleanalyticsapikey', 'googletranslateapikey'];
+        foreach ($googleapikeys as $googleapikey) {
+            $value = trim(App()->getRequest()->getPost($googleapikey));
+            $fixedValue = preg_replace('/[^A-Za-z0-9_-]/', '', $fixedValue);
+            if ($value !== $fixedValue) {
+                /* Add an alert to the user */
+                Yii::app()->setFlashMessage(sprintf(gT("Invalue value set for %s, reset to %s"), $value, $fixedValue), 'warning');
+                /* Add a warning for log for server admin */
+                Yii::log("Invalid value %s set for %s in global settings", 'warning', 'application.controller.admin.globalsettings');
```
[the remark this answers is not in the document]
Shnoulle: In flash message: user entered value , no issue to show it fixed.

application/helpers/sanitize_helper.php
```diff
@@ -499,6 +499,16 @@ function sanitize_languagecodeS($codestringtosanitize)
...
 
+/**
+/* Sanitize a google api key
+ * @see https://docs.cloud.google.com/docs/authentication/api-keys#components
+ * @param string|null
+ * @return string
+ **/
+function sanitize_googleapikey($string)
+{
+    return preg_replace('/[^A-Za-z0-9_-]/', '', trim(strval($string)));
```
gabrieljenik, line 510: Could be too much, but maybe add a max length validation? generouse one, but not huge...

gabrieljenik (approved):
Code structure looks good. Haven't yested

application/helpers/sanitize_helper.php
Shnoulle: Seems to be 39 characters, but there is no clear source for this. Yes, we can limit, but no real reason.

gabrieljenik: The reason would be to limit the edge for attacks, not only by filtering characters, but also max lenght. If we set to 100 as max, I think would be plenty for api keys and not so much for a XSS attack. But I think this is more like a general comment about these kind of validations.

PR #4940, opened 2026-05-04
Fixed issue #20499: Updating question code didn't update condition

raymondoyondi:
Fixed issue #20499: Updating question code didn't update condition

**Summary**

- Fixed question save logic to detect when the question code (title) changes and then refresh relevance for all dependent questions that use conditions based on that question (conditions.cqid = current qid). This ensures condition rendering reflects the updated question code instead of the stale one. 
- Kept the change scoped to QuestionService::updateQuestionData() so it runs automatically on normal question edits and does not alter unrelated condition flows. 

**Testing**

- ✅ php -l application/models/services/QuestionAggregateService/QuestionService.php

## Summary by CodeRabbit

* **Improvements**
  * Enhanced question editing with better handling of question properties and dependent relationships.
  * Improved data integrity when modifying question types or titles.
  * Refined cleanup of question answers and associated conditions during edits.

### DIFF
--- a/application/models/services/QuestionAggregateService/QuestionService.php
+++ b/application/models/services/QuestionAggregateService/QuestionService.php
@@ -284,6 +284,7 @@ class QuestionService
         }
 
         $originalRelevance = $question->relevance;
+        $originalTitle = $question->title;
 
         if ($question->type !== ($data['type'] ?? $question->type)) {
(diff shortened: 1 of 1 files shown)


Shnoulle (approved):
👍

PR #4941, opened 2026-05-04
Fixed issue #08030: Prefilled values for not relevant questions lost

Shnoulle:
Fixed issue #08030: Prefilled values for not relevant questions lost
Dev: do it a second time in _ValidateQuestion

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid prefilled/starting values for certain question types are now removed, and starting values are restored, validated, and persisted when session entries are missing. Session initialization now prefers starting values over defaults.
* **Tests**
  * Added a functional test and accompanying survey fixture to verify prefill/starting-value behavior across pages.


Shnoulle:
Issue with ubuntu server

About _Restore startingValues through the same normalization path as StartSurvey()._ unsure here, maybe best is to filter startingValues before then. Do not enter bad staringValuies directly. 
No issue to set it in session, user just sees something invalid, but like he tries to enter with the URL.

gabrieljenik (approved):
Code structure looks good. Code inside EM is hard to follow, making reviewal harder. This needs carefull testing @tiborpacalat . If there is already some automatic tests, I would suggest to add some cases for the new situation.

Shnoulle:
> Code inside EM is hard to follow, making reviewal harder. 

It's updated at the same way than default value here :)

> If there is already some automatic tests, I would suggest to add some cases for the new situation.

Yes, good idea !

Shnoulle:
I merge it when test is OK

Shnoulle:
Revert the last one , do it just for test

Shnoulle:
@gabrieljenik i don't understand : why new test didn't happen ? can you check ?

PR #4946, opened 2026-05-04
Fixed issue #20318: Automatic load url didn't work with question code is URL

raymondoyondi:
Fixed issue #20318: Automatic load url didn't work with question code is URL

**Summary**

- Fixed the automatic end-URL redirection bug when an equation/question code is named URL by preventing the standard replacement token URL from overriding Expression Manager resolution in runtime string processing. This ensures redirects can resolve to the equation value as expected. 
- Added an inline comment explaining the collision and why URL must be removed from the standard replacement set before calling ProcessStepString(). 

**Testing**

- ✅ php -l application/helpers/SurveyRuntimeHelper.php
- ✅ git add application/helpers/SurveyRuntimeHelper.php && git commit -m "Fix autoredirect when expression code is URL"

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the standard URL placeholder from overriding a question/equation named "URL", restoring expected redirect behavior.
  * Removed the automatic provision of the standard {URL} template field to avoid conflicts with similarly named items.

### DIFF
--- a/application/helpers/SurveyRuntimeHelper.php
+++ b/application/helpers/SurveyRuntimeHelper.php
@@ -1408,6 +1408,10 @@ class SurveyRuntimeHelper
         if ((strpos((string) $sProcessedString, "{") !== false)) {
             // process string anyway so that it can be pretty-printed
             $aStandardsReplacementFields = getStandardsReplacementFields($this->aSurveyInfo);
+            // Keep the expression variable `URL` usable in end-URL equations.
+            // Otherwise the standard replacement field `URL` (template link placeholder)
+            // overrides question/equation code named URL and prevents expected redirects.
+            unset($aStandardsReplacementFields['URL']);
             $sProcessedString = LimeExpressionManager::ProcessStepString($sString, $aStandardsReplacementFields, $iRecursionLevel, $static);
         }
         return $sProcessedString;
--- a/application/helpers/replacements_helper.php
+++ b/application/helpers/replacements_helper.php
@@ -440,7 +440,6 @@ function getStandardsReplacementFields($thissurvey)
     $coreReplacements['SURVEYLANGUAGE'] = $surveylanguage = App()->language;
     $coreReplacements['SURVEYNAME'] = ($thissurvey['name'] ?? Yii::app()->getConfig('sitename'));
     $coreReplacements['SURVEYRESOURCESURL'] = (isset($thissurvey['sid']) ? Yii::app()->getConfig("uploadurl") . '/surveys/' . $thissurvey['sid'] . '/' : '');
-    $coreReplacements['URL'] = $_linkreplace;
     $coreReplacements['WELCOME'] = ($thissurvey['welcome'] ?? '');
     $coreReplacements['CLOSE_TRANSLATION'] = gT('Close');
     $coreReplacements['ASSESSMENT_CURRENT_TOTAL'] = $_assessment_current_total;


application/helpers/SurveyRuntimeHelper.php
```diff
@@ -1408,6 +1408,10 @@ private function processString($sString, $iRecursionLevel = 1, $static = false)
         if ((strpos((string) $sProcessedString, "{") !== false)) {
             // process string anyway so that it can be pretty-printed
             $aStandardsReplacementFields = getStandardsReplacementFields($this->aSurveyInfo);
+            // Keep the expression variable `URL` usable in end-URL equations.
+            // Otherwise the standard replacement field `URL` (template link placeholder)
+            // overrides question/equation code named URL and prevents expected redirects.
+            unset($aStandardsReplacementFields['URL']);
```
Shnoulle, line 1414: I think we have 2 solutions

1. Leave URL in aStandardsReplacementFields everywhere
2. Disable URL for question code

My opinion : `$aStandardsReplacementFields['URL']` (and a lot of others) came from the old template system and are not used since 3.0.0.
The solution is to track down current usage and remove it if not used.

Shnoulle (changes requested):
If an expression works in one place, it should work the same way elsewhere. There shouldn't be any special exceptions like this one.

Shnoulle (approved):

application/helpers/SurveyRuntimeHelper.php
Shnoulle, line 1414: The deletion doesn't seem to come from `getStandardsReplacementFields`

Shnoulle:
@gabrieljenik the issue still persist, but here we remove URL from standard replacement field; I think it's a great idea.

But I think the **a lot of standard replacement field** must be removed! Then maybe for major-develop ? https://github.com/LimeSurvey/LimeSurvey/blob/6ecc85e4b86ae2a2eefd47487ae6909e9718bea9/application/helpers/replacements_helper.php#L416

I mean : except SID, GID, ? 
And move needed one in ExpressionManager

gabrieljenik:
> @gabrieljenik the issue still persist, but here we remove URL from standard replacement field; I think it's a great idea.
 > But I think the **a lot of standard replacement field** must be removed! Then maybe for major-develop ?

I am not a fan of removing URL field or replacing old expressions.
Sounds like trouble and error prone.

I think the same problem may arise if there is a question named WELCOME or SURVEYDESCRIPTION.
Those are reserved words. So, removing the field doesn't sound like a complete solution.

I like your idea of not allowing questions titled URL.
> If used, i think we must add it in https://github.com/LimeSurvey/LimeSurvey/blob/71721f13d752a9d250f8004b694b0eaaa13db254/application/models/Question.php#L265

Having said that, I am having trouble finding when the URL field is being used.
Imagine we don't have a question named URL, when is the standard URL replacement-field used?

Shnoulle:
> > @gabrieljenik the issue still persist, but here we remove URL from standard replacement field; I think it's a great idea.
> > But I think the **a lot of standard replacement field** must be removed! Then maybe for major-develop ?
> 
> I am not a fan of removing URL field or replacing old expressions. Sounds like trouble and error prone.
> 
> I think the same problem may arise if there is a question named WELCOME or SURVEYDESCRIPTION. Those are reserved words. So, removing the field doesn't sound like a complete solution.

Yes, it's the reason why I think if we remove it, we need to remove it in develop-major.

On second thought, I've already used `SURVEYDESCRIPTION` in question text. (and before languagechanger).

> 
> I like your idea of not allowing questions titled URL.
> 
> > If used, i think we must add it in https://github.com/LimeSurvey/LimeSurvey/blob/71721f13d752a9d250f8004b694b0eaaa13db254/application/models/Question.php#L265
> 
> Having said that, I am having trouble finding when the URL field is being used. Imagine we don't have a question named URL, when is the standard URL replacement-field used?

You can use it in question text, group text etc ... **but**  here  : URL is always empty , set as empty here : https://github.com/LimeSurvey/LimeSurvey/blob/6ecc85e4b86ae2a2eefd47487ae6909e9718bea9/application/helpers/replacements_helper.php#L235  and set as URL here : https://github.com/LimeSurvey/LimeSurvey/blob/6ecc85e4b86ae2a2eefd47487ae6909e9718bea9/application/helpers/replacements_helper.php#L325

Then , to fix the issue here : Remove deprecated https://github.com/LimeSurvey/LimeSurvey/blob/6ecc85e4b86ae2a2eefd47487ae6909e9718bea9/application/helpers/replacements_helper.php#L325 (no real change in cuirrent survey)

And create another mantis + PR for others replacements .

1. Check if it's OK, return something (i think NAVIGATOR , MOVEPREVBUTTON etc .;  are brken)
2. If it's still OK : add in Question model ruels.

gabrieljenik:
> URL is always empty , set as empty here

There is no other place where the URL expression is used?
So, this that we see now is just a leftover from something deprecated?

> Then , to fix the issue here : Remove deprecated

If you remove it and someone uses the expression in his surveys, even if empty, then removing it will make the expression to fail.

I think the costs and risks of removing are much more than the benefits.
Yes, this sounds legacy, but I wouldn´t remove it yet.
If this was to be removed, we need to:
- Do it on major upgrade
- Create some kind of validation or transformation script in the upgrade for current surveys

Shnoulle:
>  So, this that we see now is just a leftover from something deprecated?

We really must remove deprecated parts like that ! Else we have more and more issues due to deprecated code (and potential security issue)

> 
> If this was to be removed, we need to:
> 
>- Do it on major upgrade

Yes, OK :)

>- Create some kind of validation or transformation script in the upgrade for current surveys

I really don't think it's currently used a lot.

PR #4953, opened 2026-05-05
Fixed issue #20477: Type C description matches rendered Yes/Uncertain/No order

Calacly:
## PR type

Fixed issue #20477: Type C description matches rendered Yes/Uncertain/No order

## Notes for review

- Includes `Update_650` so existing installs receive the corrected `lime_question_themes` text, not only fresh installs.
- The `gT()` source change introduces a new msgid; translation catalogs will need to be refreshed via translate.limesurvey.org once merged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected array question option order from "Yes/No/Uncertain" to "Yes/Uncertain/No" in survey themes and templates.

* **Chores**
  * Added database migration to restore missing primary keys on legacy question tables.
  * Added migration to update stored question theme titles/descriptions to match the corrected order.
  * Bumped database version number to 708.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LimeSurvey/LimeSurvey/pull/4953?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Calacly:
Thanks for the catch on the SPSS export inconsistency. CodeRabbit suggested two paths:

(A) Re-encode `case 'C'` so codes match the new label (Y=1, U=2, N=3) and update value labels accordingly.
(B) Revert the label change in `export_helper.php` so it matches the existing codes.

I went with (B): reverted both `export_helper.php` edits (the line 208 case-comment and the line 479 SPSS label) so that file is back to its pre-PR state. After the revert, label/codes/value-labels in `export_helper.php` all agree on `Yes/No/Uncertain` order.

Reasoning for (B) over (A): (A) changes numeric codes that downstream SPSS analysis scripts have been written against (1=Yes, 2=No, 3=Uncertain), which would silently break existing analyses and make new exports incomparable to historical ones. Bug #20477 is about the question-type description shown in admin, not the SPSS encoding — so I'd prefer to leave the encoding question as a separate ticket if it should be revisited.

The picker-description fix (`QuestionType.php`, theme `config.xml`, twig comment, etc.) is unchanged and still does what #20477 asks.

Force-pushed to update the PR. Happy to take it the other direction if you'd prefer (A).

Shnoulle (approved):

Shnoulle (changes requested):
Need a new DB update https://github.com/Calacly/LimeSurvey/blob/614011f78b295ecce2ff89f8b150044990b32f01/application/helpers/update/updates/Update_651.php

Calacly:
@Shnoulle Re-added the `question_themes` migration as `Update_652` and bumped `dbversionnumber` to 652. Ready for another look — thanks!

Shnoulle (approved):

Calacly:
@Shnoulle Rebased on 7.0 — migration is now `Update_707` / dbversion 707. Happy to re-bump if it drifts before merge.

Shnoulle:
> @Shnoulle Rebased on 7.0

Why? I assign the bug to @tiborpacalat for testing. Nobody asked for it for development version?

tiborpacalat:
What is the problem here? We had long weekend, so I just saw this today.

Calacly:
@Shnoulle My mistake — I rebased onto 7.0 without asking, and I shouldn't have. When master absorbed the 7.0 merge the branch went stale, and I treated that as a routine rebase rather than thinking about where this should land.

If you'd like it back on 6.x for @tiborpacalat to test, I can revert to the pre-rebase state (`Update_652` / dbversion 652). If you'd prefer 7.0, it's already there. Either way — just let me know.

c-schmitz:
I think it is better if it goes to version 7. It is non-critical and with the release of 7.0, the 6.x version from now on will only receive critical bugfixes.

Calacly:
Thanks @c-schmitz — understood, leaving it on 7.0.

c-schmitz:
Since there is already a dbversion 707 you will have to move  to 708.

Calacly:
Done — moved to `Update_708` / dbversion 708 (head `cb889a4d`).

PR #4965, opened 2026-05-08
Dev Fix unit tests and CI to work with active debug mode

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved null-safety and defensive checks, safer option/field handling, guarded calculations to avoid divide-by-zero, and more robust inheritance fallback with cycle detection.
* **Tests**
  * Stabilized tests: suppressed libxml warnings, refined assertions and annotations, added default test-password fallback, added test group annotation, and updated test bootstrap to honor a debug env var and set request server fields.
* **Documentation**
  * Expanded inline docs and PHPDoc across multiple components.
* **Chores**
  * CI: added an extra unit-test run honoring the debug-level env var with conditional exclusion.


application/helpers/expressions/em_core_helper.php
```diff
@@ -2918,9 +2918,9 @@ function exprmgr_sumifop($args)
 function exprmgr_checkdate($month, $day, $year)
 {
     if (
-        (!ctype_digit($month) && !is_int($month))
-        || (!ctype_digit($day) && !is_int($day))
-        || (!ctype_digit($year) && !is_int($year))
+        (!ctype_digit((string)$month) && !is_int($month))
+        || (!ctype_digit((string)$day) && !is_int($day))
+        || (!ctype_digit((string)$year) && !is_int($year))
```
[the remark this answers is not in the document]
c-schmitz: ctype_digit((string)true) returns true for "1", but checkdate(1, 1, 1) is a valid date (Jan 1, year 1). A boolean true being treated as 1 is harmless here since checkdate handles it fine, and this function is only called from LimeExpressionManager with values parsed from survey responses (always strings or ints). Adding a is_bool guard would be defensive against a scenario that doesn't occur in production.

application/models/TemplateConfiguration.php
```diff
@@ -1467,22 +1470,25 @@ public function __get($name)
 
         if (in_array($name, $aAttributesThatCanBeInherited) && $this->bUseMagicInherit) {
             // Full inheritance of the whole field
-            $sAttribute = parent::__get($name);
+            $sAttribute = $this->getAttribute($name);
```
c-schmitz, line 1476: The changes in this function are the only ones I am worried about. 
Thus Template inheritance needs to be checked/tested thoroughly.
This might also solve a long standing bug where errors are shown before a TemplateConfiguration is saved, once.

c-schmitz: See https://bugs.limesurvey.org/view.php?id=20513

tiborpacalat: I have tested following and it looks good:
1. Survey-level theme inheriting from global theme
2. Create a child theme and verify inheritance
3. Inheritance chain (grandchild)
4. Global theme options vs. survey-level overrides
5. The issue from the bug report 20513

Shnoulle: @tiborpacalat you check SurveysGroup inheritance too ?

tests/unit/models/QuestionTest.php
```diff
@@ -29,22 +29,6 @@ public function testGetQuestionThemeOnNewQuestion()
         $this->assertEquals("browserdetect", $questionTheme->name);
     }
 
-    /**
-     * Test getting the theme from a saved question
-     */
-    public function testGetQuestionThemeOnSavedQuestion()
```
Shnoulle, line 35: Removed ?

PR #4968, opened 2026-05-11
Fixed issue #20503: Using quick add for multilingual label sets fails on save

gabrieljenik:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed label creation to ensure newly added labels maintain proper linkage across all supported language tabs, improving data consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LimeSurvey/LimeSurvey/pull/4968)


assets/scripts/admin/labels.js
```diff
@@ -198,9 +198,11 @@ function quickaddfunction() {
     retrieveRowHtml(datas)
         .then((result) => {
             $.each(parsedRows, (index, parsedRow) => {
+                // One randomid per label, shared across all language tabs so the save
+                // serializer can correlate title_en_XXXX with title_es_XXXX, etc.
```
Shnoulle, line 202: Comment seems valid (and seems better to use same randomid)

Why it don't work ?

HTML code for each row came from https://github.com/LimeSurvey/LimeSurvey/blob/master/application/views/admin/labels/labelRow.twig done via https://github.com/LimeSurvey/LimeSurvey/blob/01a00c5212ba3eecb75ce7436cd5e36443685170/application/controllers/admin/Labels.php#L734 

All seems OK here :(

Shnoulle (commented):
Did you understand why the comment is invalid?

gabrieljenik:
> Did you understand why the comment is invalid?

What do you mean? Where does it says the comment is invalid?

Shnoulle:
> > Did you understand why the comment is invalid?
> 
> What do you mean? Where does it says the comment is invalid?

You move the random ID inside the language loop. The comment say _serializer can correlate title_en_XXXX with title_es_XXXX, etc._ and it seems valid looking at code. 
I don't understand why we need it to move inside loop. And more : with existing row : we have `title_en_123` and `title_fr_123` . I try to understand why i break with new line.

assets/scripts/admin/labels.js
gabrieljenik: I am not sure I understand your question. JS Random ID is used for new rows. I guess the twig is used for existing rows. Can you expand on your question please?

Shnoulle: Because the ID includes language, the generated randomID goes to {{ rowId }}

Then: row_{{ language }}_{{ rowId }} do `row_fr_randomid` and `row_en_randomid` (current code)

No reason to need  `row_fr_randomid1` and `row_en_randomid2` (code of this commit)

Shnoulle: > I guess the twig is used for existing rows.

No, see https://github.com/LimeSurvey/LimeSurvey/blob/2cb407a2c0633b6604d68e4f693a43aa62fae1e7/assets/scripts/admin/labels.js#L308

`data-url="/master/index.php?r=admin/labels/sa/getLabelRowForAllLanguages"` 

https://github.com/LimeSurvey/LimeSurvey/blob/7048683dbe5a44b74406de94f6d993f338c2238c/application/controllers/admin/Labels.php#L686

twig generation : https://github.com/LimeSurvey/LimeSurvey/blob/7048683dbe5a44b74406de94f6d993f338c2238c/application/controllers/admin/Labels.php#L726

gabrieljenik: > No reason to need row_fr_randomid1 and row_en_randomid2 (code of this commit)

No, the commit changes that. It makes the newid to be the same across languages for the same row / label.
Isn't that the expected behaviour? 

Can you please explain again the issue that you see?

Shnoulle (approved):

assets/scripts/admin/labels.js
Shnoulle: > No, the commit changes that. It makes the newid to be the same across languages for the same row / label.

I'm such an idiot, I read it backwards.... sorry !

gabrieljenik: I am glad we got to a shared conclusion... I was driving my self crazy think what you were trying to say :)

Shnoulle:
> > Did you understand why the comment is invalid?
> 
> What do you mean? Where does it says the comment is invalid?

Yes , sorry, sorry ....

PR #4994, opened 2026-05-21
Bug/AT-1977 missing survey languages

Trischi80:
Dev: In the new editor the languages in general settings where missing for users that are not the survey owner. It was just a wrong comparisson between int and string on backend side.

## Summary by CodeRabbit

* **Bug Fixes**
  * Request user IDs are now consistently normalized to integers during processing to prevent permission-check mismatches.
  * Theme option conversion now excludes the ajaxmode setting from JSON output so it won't appear in exported theme data.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LimeSurvey/LimeSurvey/pull/4994?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


application/libraries/Api/Command/V1/UserDetail.php
karcstefan: psalm complains about another cast to int here, seems reasonable to remove the cast here as it's already been done at L45

karcstefan (changes requested):

application/libraries/Api/Command/V1/UserDetail.php
Trischi80: Done :). Good one, thank you.

karcstefan (approved):

PR #5005, opened 2026-05-26
Fixed issue #20138:  ComfortUpdate does not sufficiently check if existing files can be modified

c-schmitz:
Bug report: https://bugs.limesurvey.org/view.php?id=20138

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update validation to ensure the updater can actually perform required modifications, not just that targets are writable.
  * Targets that are writable but not modifiable by the process are now treated as read-only to avoid failed updates.
  * Enhanced ownership/modifiability checks when system info is available; falls back to permissive checks on unsupported platforms.
  * Suppressed noisy errors during update archive extraction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LimeSurvey/LimeSurvey/pull/5005?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/application/models/UpdateForm.php
+++ b/application/models/UpdateForm.php
@@ -197,7 +197,7 @@ class UpdateForm extends CFormModel
             if (file_exists($lsRootPath . $check)) {
                 if (!is_writable($lsRootPath . $check)) {
                     $readOnly[] = $lsRootPath . $check;
-                } elseif (!$this->isFileOwnedByCurrentProcess($lsRootPath . $check)) {
+                } elseif (!$this->isFileModifiable($lsRootPath . $check)) {
                     // File is writable but not owned by current process
                     // This prevents updates from failing mid-process due to ownership issues
                     // Fixes: Bug #20138
@@ -792,12 +792,12 @@ class UpdateForm extends CFormModel
                 }
             }
 
-            // Check not only if directory is writable, but also if it's owned by current process
-            // This ensures the user running the update owns the files
+            // Check not only if directory is writable, but also if it can be modified
+            // This ensures the current process can modify files (by ownership, root, or permissions)
             // Fixes: Bug #20138 - ComfortUpdate fails when files owned by different user
             if ($is_writable && file_exists($searchpath)) {
-                $is_owned = $this->isFileOwnedByCurrentProcess($searchpath);
-                if (!$is_owned) {
+                $is_modifiable = $this->isFileModifiable($searchpath);
+                if (!$is_modifiable) {
                     $is_writable = false;
                 }
             }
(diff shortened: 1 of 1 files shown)


application/models/UpdateForm.php
```diff
@@ -797,11 +812,64 @@ private function getReadOnlyCheckedFile($file)
...
+            $currentUid = @posix_geteuid();
+            $fileUid = @fileowner($path);
+
+            // If we successfully got both UIDs, check if we can modify:
+            // 1. We own the file (UID match), OR
+            // 2. We're root (UID 0) - root can modify any file, OR
+            // 3. File is writable (permissions allow modification)
+            if ($currentUid !== false && $fileUid !== false) {
+                return ($currentUid === $fileUid) || ($currentUid === 0) || is_writable($path);
+            }
```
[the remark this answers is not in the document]
c-schmitz: But that is fine - isn't it. We just want to make sure that the file is writable at all for the executing process, no matter who owns it.

twilligls (approved):

application/models/UpdateForm.php
```diff
@@ -797,17 +812,70 @@ private function getReadOnlyCheckedFile($file)
...
+     * @return bool True if current process can modify the file, false otherwise
+     */
+    private function isFileModifiable($path)
+    {
+        if (!file_exists($path)) {
+            return false;
+        }
+
+        // Check if POSIX functions are available (available on Linux/Unix, not on Windows)
+        if (function_exists('posix_geteuid') && function_exists('fileowner')) {
```
karcstefan, line 853: These checks are already covered by `is_writtable`.
`is_writtable` wraps around an UNIX `acccess(2)` which resolves all of these checks + GID check.
Trace goes `is_writtable` -> `php_stat(filename, FS_IS_W)` -> VCWD_ACCESS (wrapper of access)
https://github.com/php/php-src/blob/master/ext/standard/filestat.c#L778

application/models/UpdateForm.php
```diff
@@ -797,17 +812,70 @@ private function getReadOnlyCheckedFile($file)
...
+            if ($currentUid !== false && $fileUid !== false) {
+                return ($currentUid === $fileUid) || ($currentUid === 0) || is_writable($path);
+            }
+        }
+
+        // POSIX not available or unable to determine ownership
+        // On Windows or systems without POSIX, we can't directly check ownership
+        // but is_writable() should handle basic permission checks
+        // Return true to allow the update to proceed (permission already verified by caller)
+        return true;
```
karcstefan: should return `is_writtable`

application/models/UpdateForm.php
```diff
@@ -797,17 +812,70 @@ private function getReadOnlyCheckedFile($file)
...
+        if (function_exists('posix_geteuid') && function_exists('fileowner')) {
+            $currentUid = @posix_geteuid();
+            $fileUid = @fileowner($path);
+
+            // If we successfully got both UIDs, check if we can modify:
+            // 1. We own the file (UID match), OR
+            // 2. We're root (UID 0) - root can modify any file, OR
+            // 3. File is writable (permissions allow modification)
+            if ($currentUid !== false && $fileUid !== false) {
+                return ($currentUid === $fileUid) || ($currentUid === 0) || is_writable($path);
```
karcstefan: is_writtable should be removed from here and moved on return

karcstefan (changes requested):

PR #5023, opened 2026-06-02
Fixed issue #20544: Read-only survey user can delete conditions

Shnoulle:
Dev: check permission on question->sid and thirw 403
Dev: add test on existing parameters and throw 400

## Summary by CodeRabbit

* **Bug Fixes**
  * Stronger permission checks for condition management — invalid or unauthorized requests now return proper HTTP errors (400/403/404).
  * Resets and other mutating actions are blocked unless the user has update rights.

* **Improvements**
  * Question and survey context are validated before actions, preventing operations on missing or mismatched questions.
  * Permission checks applied consistently across condition forms, partial views, navigator/URL generation, and client-side matching logic.


Shnoulle:
Think we need an LSYii_Application->getQuestionId too.
I work on it

application/controllers/admin/ConditionsAction.php
```diff
@@ -130,18 +130,14 @@ public function index($subaction, $iSurveyID = null, $gid = null, $qid = null)
...
-        if (!isset($iSurveyID)) {
-            $iSurveyID = returnGlobal('sid');
-        }
-        if (!isset($qid)) {
-            $qid = returnGlobal('qid');
-        }
+        $iSurveyID = LSYii_Application::getSurveyId();
+        $qid = LSYii_Application::getQuestionId();
+
         if (!isset($gid)) {
```
gabrieljenik: Isn't there an LSYii_Application::getGroupid()? Just to keep it consistent

application/core/LSYii_Application.php
```diff
@@ -639,4 +640,48 @@ public static function getSurveyId($throwError = true)
...
+        $intQuestionId = intval($questionId);
+        if (strval($intQuestionId) !== strval($questionId)) {
+            if ($throwError) {
+                throw new CHttpException(400, gT('Your request is invalid.'));
+            }
+            return false;
+        }
+        $questionId = $intQuestionId;
+        /* questionId is set and is an integer */
+        $question = Question::model()->findByPk($questionId);
```
gabrieljenik, line 705: Are question models being cached (as survey model)? I am just worried about performance on this when it gets used more widespreaded. At least caching if the question is found or not in the DB (instead of caching the question model)

Shnoulle: Not cached: `static` integer variable

1. Parameters cannot update during the same action.
2. Avoid multiple `Question::model()->findByPk($questionId);`
3. Static variable is only an integer: no memory issue

application/controllers/admin/ConditionsAction.php
Shnoulle: Yes, easy to do :)

Shnoulle: I wait for response here https://github.com/LimeSurvey/LimeSurvey/pull/5023#discussion_r3350362679 :)

application/core/LSYii_Application.php
gabrieljenik: > Not cached: static integer variable

Agree. I was being too generic.

gabrieljenik: I think that an automated test for this would be important as it may get very widespreaded used

Shnoulle: For LSYii_Application::getQuestionId(); function only ?

Else, another idea here : previous function get qid from public function, maybe we need to allow `LSYii_Application::getQuestionId($qid = null)` and then validating $qid if it's set ?

gabrieljenik: Not sure I follow, but go for it! :)

Shnoulle: I "deprecate" the params

application/controllers/admin/ConditionsAction.php
```diff
@@ -625,7 +630,9 @@ protected function getSurveyIsAnonymized()
     public function getCopyForm(int $qid, int $gid, array $conditionsList, array $pquestions): string
     {
         App()->getClientScript()->registerScriptFile(App()->getConfig('adminscripts') . 'checkgroup.js', LSYii_ClientScript::POS_BEGIN);
-
+        $qid = LSYii_Application::getQuestionId();
```
gabrieljenik, line 632: Shouldn't this be changed from the application\controllers\admin\ConditionsAction.php index method?

application/core/LSYii_Application.php
```diff
@@ -621,27 +620,109 @@ public static function getSurveyId($throwError = true)
...
+        self::$surveyId = $surveyId;
+        return self::$surveyId;
+    }
+
+    /**
+     * Get survey survey id by param
+     * @param boolean $throwError Whether to throw an error
+     * @return false|integer
+     */
+    public static function getGroupId($throwError = true)
```
gabrieljenik, line 658: Don't want to nag you, but is there a unit test for this? (and / or for the others)?

application/controllers/admin/ConditionsAction.php
```diff
@@ -543,49 +537,60 @@ protected function getConditionText(int $qid)
...
         /** @var string $p_cid */
-        /** @var string $qid */
-        /** @var string $gid */
+        /** @var integer $qid */
+        /** @var integer $gid */
+        /** @var integer $sid */
         /** @var string $p_newscenarionum */
         /** @var string $p_scenario */
         extract($args);
+        $qid = LSYii_Application::getQuestionId();
```
gabrieljenik: Is this needed? Shoudln't qid be coming from $args?

Shnoulle: args didn't test if qid and sid are OK, it's **THE** issue here !

application/core/LSYii_Application.php
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/pull/5023/changes#diff-5e3de44919fb0c5009ae3e860bd1444c5f31d10b096272592909b9f672d2bd7dR72 

But we don't have currently something using getGroupId , then no test.

You ask to add it .... i don't add it because there are no usage

application/controllers/admin/ConditionsAction.php
Shnoulle: https://github.com/LimeSurvey/LimeSurvey/blob/53c4220ee903c356c0d2f034e9c36a498e4b14f0/application/controllers/admin/ConditionsAction.php#L630 is a public function, the can be call directly (and call directly currently)

No index function call here.

application/controllers/admin/ConditionsAction.php
Shnoulle: But : protected function, can leave arg if you want

gabrieljenik: > args didn't test if qid and sid are OK, it's THE issue here !

So shouldn't we correct that ? I mean have args to have good valid values.

application/core/LSYii_Application.php
gabrieljenik: I meant a unit test on the individual function.

application/controllers/admin/ConditionsAction.php
Shnoulle: > So shouldn't we correct that ? I mean have args to have good valid values.

What ? You mean to rewrite all of this system without reason? The important part ois check request value .....

I reset to args, since it's not a public function : no need to check it again.

application/core/LSYii_Application.php
Shnoulle: It's currently used only for condition  and the current test on condition do the job.

We need to test if parameters are OK , no more. Current test do the job for qid. I can add a gid (with a good qid and sid) if you want

No need to test single function since functinal/backend test it.

Shnoulle: > I meant a unit test on the individual function.

Current test check the 2 individual function. No need Unit test

gabrieljenik (approved):
Haven't tested it

tiborpacalat:
Unit tests are failing, Denis can you please take a look at this?

Shnoulle:
> Unit tests are failing, Denis can you please take a look at this?

Yes, it's new Uite test. I search and improve :)
Surely add some hack too ;)

Shnoulle:
@tiborpacalat fixed, I remove unrelated test and fix related test.

I think I create a DEV issue about not related test :)

PR #5030, opened 2026-06-03
Fixed issue #20548

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Admin creation and password-reset emails now only use trusted bases for absolute links and fail safely when a secure link cannot be produced, improving resilience and security.
* **Tests**
  * Added unit tests validating secure absolute URL generation, host header protections, scheme/port/path handling, fallback behavior, and failure cases.
* **Chores**
  * Added a configurable whitelist for trusted hostnames to mitigate host-header injection.

### DIFF
--- a/application/models/services/PasswordManagement.php
+++ b/application/models/services/PasswordManagement.php
@@ -31,37 +31,58 @@ class PasswordManagement
     }
 
     /**
-     * Creates an absolute URL for use in emails without relying on the HTTP Host header.
+     * Creates an absolute URL for use in emails without relying on an unvalidated HTTP Host header.
      * This prevents Host Header Injection attacks where an attacker could manipulate
      * the Host header to generate password reset links pointing to a malicious domain.
      *
-     * Uses the configured 'publicurl' if it contains a scheme and host,
-     * otherwise falls back to SERVER_NAME.
+     * Priority:
+     * 1. If 'allowedHosts' is configured and the request host is whitelisted → use it
+     *    (supports domain aliasing where multiple domains serve the same instance)
+     * 2. If 'publicurl' is an absolute URL → use it as trusted base
+     * 3. Otherwise → return null (insecure configuration, refuse to generate URL)
      *
      * @param string $route the URL route
      * @param array $params additional GET parameters
-     * @return string the constructed absolute URL
+     * @return string|null the constructed absolute URL, or null if no trusted base URL can be determined
      */
-    private static function createSecureAbsoluteUrl(string $route, array $params = []): string
+    private static function createSecureAbsoluteUrl(string $route, array $params = []): ?string
     {
-        $publicUrl = \Yii::app()->getConfig('publicurl');
-        $parsedPublicUrl = parse_url((string) $publicUrl);
+        $baseUrl = null;
 
-        if (isset($parsedPublicUrl['scheme']) && isset($parsedPublicUrl['host'])) {
-            // publicurl is absolute - use it as trusted base
-            $baseUrl = rtrim($publicUrl, '/');
-        } else {
-            // Construct base URL from SERVER_NAME (controlled by server config, not client)
+        // Option 1: Validate request host against allowedHosts whitelist
+        // This preserves domain aliasing (multiple domains pointing to same instance)
+        $allowedHosts = \Yii::app()->getConfig('allowedHosts');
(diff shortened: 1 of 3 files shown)


application/models/services/PasswordManagement.php
```diff
@@ -30,6 +30,71 @@ public function __construct(\User $user)
...
+     * 1. If 'allowedHosts' is configured and the request host is whitelisted → use it
+     *    (supports domain aliasing where multiple domains serve the same instance)
+     * 2. If 'publicurl' is an absolute URL → use it as trusted base
+     * 3. Otherwise → return null (insecure configuration, refuse to generate URL)
+     *
+     * @param string $route the URL route
+     * @param array $params additional GET parameters
+     * @return string|null the constructed absolute URL, or null if no trusted base URL can be determined
+     */
+    private static function createSecureAbsoluteUrl(string $route, array $params = []): ?string
```
gabrieljenik: I would name it a bit different to avoid confusions with HTTPs Urls.
What about createValidatedAbsoluteUrl ?

gabrieljenik:
I think some automated tests for the new function would be good.

application/models/services/PasswordManagement.php
```diff
@@ -30,6 +30,71 @@ public function __construct(\User $user)
...
+    {
+        $baseUrl = null;
+
+        // Option 1: Validate request host against allowedHosts whitelist
+        // This preserves domain aliasing (multiple domains pointing to same instance)
+        $allowedHosts = \Yii::app()->getConfig('allowedHosts');
+        if (!empty($allowedHosts) && is_array($allowedHosts)) {
+            $isSecure = !empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off';
```
gabrieljenik, line 57: Why don't we call createAbsoluteUrl() here?

application/models/services/PasswordManagement.php
```diff
@@ -30,6 +30,71 @@ public function __construct(\User $user)
...
+            if ($hostOnly && in_array($hostOnly, $allowedHosts)) {
+                // Only scheme + host (+ port). Do NOT append getBaseUrl() here
+                // because createUrl() already includes the base path.
+                $baseUrl = $scheme . '://' . $httpHost;
+            }
+        }
+
+        // Option 2: Fall back to absolute publicurl
+        if ($baseUrl === null) {
+            $publicUrl = \Yii::app()->getConfig('publicurl');
```
gabrieljenik, line 71: Maybe try to call/reuse createPublicUrl?

c-schmitz: Not possible without reintroducing the original problem

gabrieljenik: Not so sure about it, as we would only be calling it when publicurl is defined.
But I just gave it a glance. Probably I am missing something.

gabrieljenik:
Code structure looks good.

Honestly I don't like to have 2 functions to create URLs and both with different behaviour. 
Those set of function were tricky and created some headaches.
I know tests were created for the new one... but you know... unexpected things happen :)

I prefer to have the new function to rely on the old one (which was already used and tested and validated by users).

Quick question: if the new function, instead of trying to create a url, just calls the old url function and then validates it starts with an allowed host (or public url config)? I think that would be a bit simpler and closer to the intention. We don't want to rewrite the logic for URLs creation, we just want them to be REAL valid in some situations.

gabrieljenik (approved):
Not my favorite approach, but still code structure looks good.
haven't tested it.

c-schmitz:
> Quick question: if the new function, instead of trying to create a url, just calls the old url function and then validates it starts with an allowed host (or public url config)? I think that would be a bit simpler and closer to the intention. We don't want to rewrite the logic for URLs creation, we just want them to be REAL valid in some situations.

Yeah, I am also not convinced. If you go down that road you also have to be always defining what situations are these. Wouldn't it be much better if all URLs were validated and could not be spoofed?
Because right now ngxinx seems to be the only webserver providing a valid server_name in *most* situtations. 
Allowed_hosts would imho be the correct setting to define all valid hostnames.

gabrieljenik:
> Wouldn't it be much better if all URLs were validated and could not be spoofed?

Well yes, but many people may not have the allowedHosts set.
We could require that during a major upgrade

c-schmitz:
Closing issue for a better approach.

PR #5032, opened 2026-06-04
Fixed issue #20548: [security] Host Header Injection in Password Reset

c-schmitz:
## Summary by CodeRabbit

* **New Features**
  * Auto-creates an allowed-hosts config on first successful admin login; UI shows configured hosts or warns if none.

* **Security Improvements**
  * Validates hosts for generated absolute URLs; email and survey links are suppressed when host is not allowed.
  * Centralized allowlist now enforces request host checks across web and console environments.

* **Reliability**
  * Email/password workflows now fail safely when validated admin/reset URLs cannot be generated.

* **Tests**
  * Added unit tests covering allowlist behavior and validated URL creation.

* **Chores**
  * VCS ignores the allowed-hosts config file.


application/controllers/admin/Authentication.php
```diff
@@ -175,6 +175,13 @@ public static function prepareLogin()
                 $event->set('identity', $identity);
                 App()->getPluginManager()->dispatchEvent($event);
 
+                // If allowed_hosts.php does not exist, write the current host as valid
+                $allowedHosts = App()->loadAllowedHosts();
+                if (empty($allowedHosts)) {
+                    $currentHost = App()->request->getServerName();
+                    App()->writeAllowedHosts([$currentHost]);
+                }
```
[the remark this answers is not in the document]
c-schmitz: I'd expect that if an admin manages to successfully authenticate that it is a trustworthy source. You have to start somewhere.

Shnoulle:
We already have config settings and a function: https://github.com/LimeSurvey/LimeSurvey/blob/124f91dca0dbe11f70b6f2fcce96b14b6bae5b32/application/core/LSHttpRequest.php#L306

I think allowedHosts must be only for admin with SSH/FTP access, not by DB and clearly not by GUI.

Question: original issue work if allowedHosts is set ?

c-schmitz:
> We already have config settings and a function:
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/124f91dca0dbe11f70b6f2fcce96b14b6bae5b32/application/core/LSHttpRequest.php#L306
> 
> I think allowedHosts must be only for admin with SSH/FTP access, not by DB and clearly not by GUI.
> 
> Question: original issue work if allowedHosts is set ?

Good point.
I would merge that setting. It was also not documented.
The new solution is file-based and cannot be edited by GUI.

Shnoulle:
> Good point. I would merge that setting. It was also not documented. 

Yes, my fault here

> The new solution is file-based and cannot be edited by GUI.

I already have a config.php file reading another file (hide DB password to all user on a windows IIS.)

application/core/Traits/LSApplicationTrait.php
```diff
@@ -75,4 +75,155 @@ public function getPublicBaseUrl($absolute = false)
...
+            ) {
+                $sanitized[] = $host;
+            }
+        }
+        if (empty($sanitized)) {
+            return false;
+        }
+
+        $filePath = Yii::app()->getBasePath() . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'allowed_hosts.php';
+        $content = "<?php\n"
```
Shnoulle: I really think it's best to have a copy-sample file to copy and update .

Easier for us to update

application/core/Traits/LSApplicationTrait.php
```diff
@@ -75,4 +75,155 @@ public function getPublicBaseUrl($absolute = false)
...
+        $sanitized = [];
+        foreach ($hosts as $host) {
+            if (!is_string($host)) {
+                continue;
+            }
+            $host = trim($host);
+            // Must match a valid hostname (RFC 952/1123) or IP address
+            if (
+                filter_var($host, FILTER_VALIDATE_IP) !== false
+                || preg_match('/^([a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?\.)*[a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?$/', $host)
```
Shnoulle: Why not [[FILTER_VALIDATE_URL](https://www.php.net/manual/fr/filter.constants.php#constant.filter-validate-url)](https://www.php.net/manual/fr/filter.constants.php#constant.filter-validate-url) [RFC 2396](https://datatracker.ietf.org/doc/html/rfc2396). and check scheme

Or  [FILTER_VALIDATE_DOMAIN](https://www.php.net/manual/fr/filter.constants.php#constant.filter-validate-domain) ? [RFC 952](https://datatracker.ietf.org/doc/html/rfc952), [» RFC 1034](https://datatracker.ietf.org/doc/html/rfc1034), [» RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035), [» RFC 1123](https://datatracker.ietf.org/doc/html/rfc1034), [» RFC 2732](https://datatracker.ietf.org/doc/html/rfc1034), et [» RFC 2181](https://datatracker.ietf.org/doc/html/rfc2181).

PHP dev make work for us.

application/core/Traits/LSApplicationTrait.php
```diff
@@ -75,4 +75,155 @@ public function getPublicBaseUrl($absolute = false)
...
+     */
+    public function loadAllowedHosts()
+    {
+        $filePath = Yii::app()->getBasePath() . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'allowed_hosts.php';
+        if (file_exists($filePath)) {
+            $hosts = require($filePath);
+            if (is_array($hosts) && !empty($hosts)) {
+                return $hosts;
+            }
+        }
```
Shnoulle, line 166: Need to add `config('allowedHosts')` if it's an array, no?

Shnoulle (commented):
Comment + adding `allowedHosts` (i update manual tomorrow)

PS : I think a great update was to set allowedHosts when installing. I can create the feature request and the PR after (next week)

c-schmitz:
> I already have a config.php file reading another file (hide DB password to all user on a windows IIS.)

Having it as a separate file allows it to be automatically created on update.

application/core/Traits/LSApplicationTrait.php
c-schmitz: I don't agreee, not easier then it being automatically created.

c-schmitz:
> Comment + adding `allowedHosts` (i update manual tomorrow)
> 
> PS : I think a great update was to set allowedHosts when installing. I can create the feature request and the PR after (next week)

no, please don't. This patch obsoletes the old config setting.

application/core/Traits/LSApplicationTrait.php
c-schmitz: Since it does not exist anymore (and never really officially existed) I won't add it.

Shnoulle: > Since it does not exist anymore (and never really officially existed) I won't add it.

I really prefer only one file for config settings. 
I already moved the security part in config.php

Shnoulle:
> > I already have a config.php file reading another file (hide DB password to all user on a windows IIS.)
> 
> Having it as a separate file allows it to be automatically created on update.

Both is always good. You can do it like the security.php file : 

```
$config = array();
$config['encryptionnonce'] = 'XXXXXXXXXXXXX';
$config['encryptionsecretboxkey'] = 'XXXXXXXXXXXXXXXX';
return $config;
```

Then allow both easily

application/core/Traits/LSApplicationTrait.php
Shnoulle: I mean, if we want to update to add new settings ourselves.
Look at the last config.php generated with , at start of a line, a single empty line .

```
		//'publicurl' => 'https://www.example.org/limesurvey',

		// Mysql database engine (INNODB|MYISAM):
		 'mysqlEngine' => 'MYISAM'

,		// Update default LimeSurvey config here
	)
```

application/controllers/admin/Authentication.php
```diff
@@ -175,6 +175,13 @@ public static function prepareLogin()
                 $event->set('identity', $identity);
                 App()->getPluginManager()->dispatchEvent($event);
 
+                // If allowed_hosts.php does not exist, write the current host as valid
+                $allowedHosts = App()->loadAllowedHosts();
+                if (empty($allowedHosts)) {
```
gabrieljenik, line 180: Should this happen only for admin logins I think, right?

application/core/Traits/LSApplicationTrait.php
```diff
@@ -75,4 +75,155 @@ public function getPublicBaseUrl($absolute = false)
...
+        $publicUrl = Yii::app()->getConfig('publicurl');
+        if (!empty($publicUrl)) {
+            $parsed = parse_url($publicUrl);
+            if (isset($parsed['host']) && strtolower($parsed['host']) === $host) {
+                return true;
+            }
+        }
+
+        foreach ($allowedHosts as $allowed) {
+            if (strtolower(trim($allowed)) === $host) {
```
gabrieljenik: I think it does, but asking just in case.
This function will handle gracefully "/", "http://", "https:" prefixes which may appear in the config file (if manually entered)? 

I know sounds silly, but it could happen maybe... and we are already considering spaces and case

application/core/LimeMailer.php
```diff
@@ -866,22 +866,47 @@ public function getTokenReplacements()
                 $aTokenReplacements[strtoupper((string) $attribute)] = $value;
             }
         }
-        /* Set the minimal url and add it to Placeholders */
-        $aTokenReplacements["OPTOUTURL"] = App()->getController()
-            ->createAbsoluteUrl("/optout/tokens", array("surveyid" => $this->surveyId, "token" => $token,"langcode" => $language));
+        /* Set the minimal url and add it to Placeholders - use validated URLs to prevent host header injection */
+        $aTokenReplacements["OPTOUTURL"] = App()
+            ->createValidatedAbsoluteUrl("/optout/tokens", array("surveyid" => $this->surveyId, "token" => $token,"langcode" => $language));
+        if ($aTokenReplacements["OPTOUTURL"] === false) {
```
gabrieljenik, line 872: Is this OK? If there is no validated url, we trust in the host header?

gabrieljenik:
I like this approach!

application/controllers/admin/Authentication.php
c-schmitz: yes, only on successful admin login.

application/core/LimeMailer.php
c-schmitz: That should  basically never happen, it is just a fallback in case the allowed_hosts file does not exist, yet.

application/core/Traits/LSApplicationTrait.php
c-schmitz: well, the comment in the file clearly states that domains need to be entered without protocol and port, even gives examples. We can't heal stupidity.

application/core/LimeMailer.php
Shnoulle: For information : issue 20548 can not happen if you set `allowedHosts` 

There are no changes here: _the check only runs when file was updated_ (config.phpo or allowed_hosts.php) was not updated by default or when installing.

> LimeSurvey overrides getHostInfo() and calls checkIsAllowedHost(), which is meant to reject untrusted hosts. The check only runs when the optional allowedHosts configuration value is a non-empty array, and that value is not defined in the shipped configuration. The check is a no-op out of the box:

from reported issue.

c-schmitz:
> Both is always good. You can do it like the security.php file :

Good point. Changed it so it is filling the config variable.

c-schmitz (commented):
I think I addressed all points now.

application/controllers/admin/Authentication.php
Shnoulle: Currently happen at each call : https://github.com/LimeSurvey/LimeSurvey/blob/15da6388b3d74dcac85c3f05a5aabe4daa99975f/application/core/LSHttpRequest.php#L296 

Send `"Host: evil.attacker.example:1337"` send a 400 error.

Shnoulle: Her, we create it with authentication: then if there are different URLs for admin and public, break the public page.
If there are simply different URLs , domain namle : broke other access.

Can broke installation when update.

Shnoulle:
> > Both is always good. You can do it like the security.php file :
> 
> Good point. Changed it so it is filling the config variable.

Thank you !

We must send a warning after/before creation of the file. This can break some instance (i have some where the domain name for admin and public are not the same : /admin/ disable for public domain or use SERVER authentication on specific domain, for example).

application/core/LSYii_Application.php
```diff
@@ -165,6 +165,12 @@ public function setConfigs()
                 $this->config = array_merge($this->config, $securityConfig);
             }
         }
+        if (file_exists($configdir . '/allowed_hosts.php')) {
+            $allowedHostsConfig = require($configdir . '/allowed_hosts.php');
+            if (is_array($allowedHostsConfig)) {
+                $this->config = array_merge($this->config, $allowedHostsConfig);
+            }
+        }
```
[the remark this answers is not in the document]
Shnoulle: @c-schmitz seems is right here : https://github.com/LimeSurvey/LimeSurvey/blob/c4a4b3a3d2e1205409de9b10e55f312d7b7ca791/application/config/config-defaults.php#L738

I never found a good way to set this dir (no way toi set via ENV or anything else), but i think it was used in limesurvey hosting ?

Shnoulle (approved):
Think it's OK. Just upset about some user updating and don't have server access.

c-schmitz:
THere are also no conflicts with v6 branch, so after merge this can be cherrypicked to v6 branch without problems.

Shnoulle:
I make some test, idea

1. With `publicurl` set different than the 1st login URL, both hosts must be accepted, and `publicurl` must be used for the email message
2. With some `allowedHosts` set: after the 1st login, all hosts must be accepted, and an email must be sent with all hosts. 
3. Without both: only one address accepted and email sent with this address.


Other solution ?

Shnoulle:
> I make some test, idea

Test OK :+1:

application/core/ConsoleApplication.php
```diff
@@ -76,6 +76,12 @@ public function __construct($aApplicationConfig = null)
                 $lsConfig = array_merge($lsConfig, $securityConfig);
             }
         }
+        if (file_exists(__DIR__ . '/../config/allowed_hosts.php')) {
+            $allowedHostsConfig = require(__DIR__ . '/../config/allowed_hosts.php');
```
gabrieljenik, line 80: require or include?

application/controllers/admin/Authentication.php
gabrieljenik: > yes, only on successful admin login.

I might be miss reading, but isn't this happening for all successfull logins, no matter if they are admins or not?

application/core/ConsoleApplication.php
c-schmitz: Does not really matter since it is since it is guarded by file_exists

PR #5053, opened 2026-06-12
Dev Remove unmaintained composer packages - Statistics export now as xlsx instead of xls

c-schmitz:
## Summary by CodeRabbit

* **Improvements**
  * Statistics and user statistics exports now produce modern XLSX files (improved compatibility and download behavior).
  * XLSX exports use a new writer with better worksheet/format handling, preserving bold and numeric formats.

* **Removed**
  * Console database seeder command and related helper removed.

* **Chores**
  * Project ignore rules updated to exclude a specific vendor path.


c-schmitz:
The new Spreadsheet library is already used on Cloud, btw, so this is a good thing.

karcstefan (approved):

PR #5063, opened 2026-06-17
#AT-1459: Fix bug where groups get unfolded when we add a new one

ramishenouda:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of question group management in the survey editor when adding, duplicating, or rendering question groups.

* **New Features**
  * Added a shared question-group creator to standardize default values and initialization for new question groups.

* **Refactor**
  * Updated key handling for question group rendering to use deterministic identifiers rather than non-deterministic values.

* **Tests**
  * Added unit tests covering question-group creation and duplication behavior.


karcstefan (changes requested):
CodeRabbits remarks seem legit to me.

karcstefan (approved):

PR #5075, opened 2026-06-22
Fixed issue #20532: STARTPOLICYLINK and ENDPOLICYLINK return errors in the survey logic view

gabrieljenik:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of specific policy-related placeholders (STARTPOLICYLINK and ENDPOLICYLINK) in expression evaluation to prevent them from being incorrectly flagged as undefined variables or displayed with error highlighting.

* **Tests**
  * Added comprehensive test coverage to verify that policy-related placeholders are properly recognized and handled without raising validation errors.

### DIFF
--- a/application/helpers/expressions/em_core_helper.php
+++ b/application/helpers/expressions/em_core_helper.php
@@ -703,6 +703,10 @@ class ExpressionManager
                         }
                         $this->RDP_StackPush($result);
                         return true;
+                    } elseif ($this->RDP_isKnownNonEmPlaceholder($token[0])) {
+                        // Pass through non-EM placeholders like STARTPOLICYLINK
+                        $this->RDP_StackPush(array('{' . $token[0] . '}', $token[1], 'WORD'));
+                        return true;
                     } else {
                         $this->RDP_AddError(self::gT("Undefined variable"), $token);
                         return false;
@@ -1563,10 +1567,7 @@ class ExpressionManager
                         $stringParts[] = $token[0];
                         $stringParts[] = "</span>";
                     } else {
-                        if (!$this->RDP_isValidVariable($token[0])) {
-                            $class = 'em-var-error';
-                            $displayName = $token[0];
-                        } else {
+                        if ($this->RDP_isValidVariable($token[0])) {
                             $jsName = $this->GetVarAttribute($token[0], 'jsName', '');
                             $code = $this->GetVarAttribute($token[0], 'code', '');
                             $question = $this->GetVarAttribute($token[0], 'question', '');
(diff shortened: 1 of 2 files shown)


Shnoulle (commented):
Here we add it always, but it's not the case . We need to add it only when it's really added 

https://github.com/LimeSurvey/LimeSurvey/blob/ccc07bf9ae26313f3e2203eb5f7fe4ac8ed52d2c/application/helpers/expressions/em_manager_helper.php#L9312 

If you add it in the question text, it's not replaced for example : it's still an issue
```
$LEM->ProcessString($aSurveyInfo['surveyls_policy_notice'], 0, ['STARTPOLICYLINK'=>gT("Start"), 'ENDPOLICYLINK'], => gT("End")]); 
$sPrint = viewHelper::purified(viewHelper::filterScript($LEM->GetLastPrettyPrintExpression()));
$LEM->setTempVars( ['STARTPOLICYLINK'=>null, 'ENDPOLICYLINK'] => null]);
```

gabrieljenik:
> Here we add it always, but it's not the case . We need to add it only when it's really added

Can you expand a bit on this. I am not following.

Shnoulle:
> > Here we add it always, but it's not the case . We need to add it only when it's really added
> 
> Can you expand a bit on this. I am not following.

If you put STARTPOLICYLINK in the question text, it's an error.

You can only put it in surveyls_policy_notice not elsewhere.

gabrieljenik:
> > > Here we add it always, but it's not the case . We need to add it only when it's really added
> > 
> > 
> > Can you expand a bit on this. I am not following.
> 
> If you put STARTPOLICYLINK in the question text, it's an error.
> 
> You can only put it in surveyls_policy_notice not elsewhere.

If I understand correctly:
- The current state of the PR allows STARTPOLICYLINK to be used anywhere in the EM and not fail.
- It should fail everywhere and just don't throw error when used on surveyls_policy_notice 
Is this what you are saying?

Shnoulle:
> It should fail everywhere and just don't throw error when used on surveyls_policy_notice

Yes

And see the proposal code

```
$LEM->ProcessString($aSurveyInfo['surveyls_policy_notice'], 0, ['STARTPOLICYLINK'=>gT("Start"), 'ENDPOLICYLINK'], => gT("End")]); 
$sPrint = viewHelper::purified(viewHelper::filterScript($LEM->GetLastPrettyPrintExpression()));
$LEM->setTempVars( ['STARTPOLICYLINK'=>null, 'ENDPOLICYLINK'] => null]);
```

Just update Start and End for the explanation here

gabrieljenik:
Closed in favor of new approach https://github.com/LimeSurvey/LimeSurvey/pull/5133

PR #5105, opened 2026-07-01
LE-799: Fixed issues at SQL Server

lajosarpad:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retrieval of survey question statistics for localized surveys, ensuring question and answer data are correctly matched by language.
  * Refined survey/question filtering to better align results with the selected survey and eliminate unintended exclusions, improving reliability when loading statistics.

### DIFF
--- a/application/models/services/SurveyStatistics/Charts/Questions/QuestionStatistics.php
+++ b/application/models/services/SurveyStatistics/Charts/Questions/QuestionStatistics.php
@@ -143,12 +143,18 @@ class QuestionStatistics implements StatisticsChartInterface
         $command = Yii::app()->db->createCommand()
             ->select($select)
             ->from('{{questions}} q')
-            ->leftJoin('{{question_l10ns}} ql', "q.qid = ql.qid AND ql.language = " . Yii::app()->db->quoteValue($this->language))
+            ->leftJoin('{{question_l10ns}} ql', "q.qid = ql.qid AND ql.language = :language1")
             ->leftJoin('{{answers}} a', 'q.qid = a.qid')
-            ->leftJoin('{{answer_l10ns}} al', "a.aid = al.aid AND al.language = " . Yii::app()->db->quoteValue($this->language))
+            ->leftJoin('{{answer_l10ns}} al', "a.aid = al.aid AND al.language = :language2")
             ->leftJoin('{{question_attributes}} qa', 'q.qid = qa.qid')
-            ->where("q.sid = {$this->surveyId}")
+            ->where("q.sid = :sid")
             ->order('q.parent_qid ASC, q.scale_id ASC, q.question_order ASC, q.title ASC, a.sortorder ASC, a.code ASC');
+
+        $command->params = [
+            ':language1' => $this->language,
+            ':language2' => $this->language,
+            ':sid' => $this->surveyId
+        ];
 
         return $command;
     }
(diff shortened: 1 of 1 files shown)


application/models/services/SurveyStatistics/Charts/Questions/QuestionStatistics.php
```diff
@@ -143,18 +143,13 @@ private function buildBaseQuery(): CDbCommand
...
             ->select($select)
             ->from('{{questions}} q')
-            ->leftJoin('{{question_l10ns}} ql', 'q.qid = ql.qid AND ql.language = :language')
+            ->leftJoin('{{question_l10ns}} ql', "q.qid = ql.qid AND ql.language = " . Yii::app()->db->quoteValue($this->language))
             ->leftJoin('{{answers}} a', 'q.qid = a.qid')
-            ->leftJoin('{{answer_l10ns}} al', 'a.aid = al.aid AND al.language = :language')
+            ->leftJoin('{{answer_l10ns}} al', "a.aid = al.aid AND al.language = " . Yii::app()->db->quoteValue($this->language))
             ->leftJoin('{{question_attributes}} qa', 'q.qid = qa.qid')
-            ->where('q.sid = :sid AND (q.parent_qid = 0 OR q.parent_qid IN (SELECT qid FROM {{questions}} WHERE sid = :sid))')
+            ->where("q.sid = {$this->surveyId}")
```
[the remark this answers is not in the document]
lajosarpad: That's an inconsistency that should not happen. If we were to handle this problem, then we could as well handle questions where parent_qid matches their own qid, or in other words, rootless circular references and other kinds of inconsistencies. Instead, simplicity is to be preferred unless there is a reproducible real problem that needs to be handled. If orphan subquestions may be in the questions table, then that's a problem with question consistency that needs to be fixed instead of accomodating use-cases to the problem.

Ahmedazeem98 (approved):

application/models/services/SurveyStatistics/Charts/Questions/QuestionStatistics.php
```diff
@@ -143,18 +143,13 @@ private function buildBaseQuery(): CDbCommand
         $command = Yii::app()->db->createCommand()
             ->select($select)
             ->from('{{questions}} q')
-            ->leftJoin('{{question_l10ns}} ql', 'q.qid = ql.qid AND ql.language = :language')
+            ->leftJoin('{{question_l10ns}} ql', "q.qid = ql.qid AND ql.language = " . Yii::app()->db->quoteValue($this->language))
             ->leftJoin('{{answers}} a', 'q.qid = a.qid')
-            ->leftJoin('{{answer_l10ns}} al', 'a.aid = al.aid AND al.language = :language')
+            ->leftJoin('{{answer_l10ns}} al', "a.aid = al.aid AND al.language = " . Yii::app()->db->quoteValue($this->language))
```
c-schmitz: This is not  the correct way to fix this. Instead use for each a  differently named parameter, like :language1, :language2.
That will correctly work in MSSQL.

c-schmitz (changes requested):

application/models/services/SurveyStatistics/Charts/Questions/QuestionStatistics.php
lajosarpad: Applied the change

PR #5109, opened 2026-07-02
LE-:817: Update ranking questions to use type R for subquestions

Ahmedazeem98:
1. Update suq-questions of ranking questions to have type of R instead of T.
2. When we create new subquestions of ranking the type needs to be R instead of T.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ranking subquestions now inherit and persist the correct type from their parent ranking question when created or updated.
  * XML-based imports now ensure imported subquestion records store the same type as their already-imported parent ranking question.
  * Existing ranking subquestion records are automatically corrected to keep types consistent across supported database engines.


application/models/services/QuestionAggregateService/SubQuestionsService.php
```diff
@@ -179,6 +179,11 @@ private function storeSubquestion(
         $subquestion->sid = $question->sid;
         $subquestion->gid = $question->gid;
         $subquestion->parent_qid = $question->qid;
+
+        if ($question->type === Question::QT_R_RANKING) {
+            $subquestion->type = Question::QT_R_RANKING;
```
Shnoulle, line 184: Why needed only for ranking and not others ?

lajosarpad: @c-schmitz can answer this question

lajosarpad (approved):
The code looks good, thanks for it Ahmed! We applied some changes during our meeting, but we'll need this to be tested in

- MySQL (Tibor)
- PostgreSQL (Tibor)
- SQL Server (me)

and if it passes all the tests then we can merge it

PR #5117, opened 2026-07-06
Fixed issue #20574: Unable to update with not compatible theme

Shnoulle:
Dev: Do not redirect in case of issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the theme uninstall flow to detect unsuccessful removals and treat them as failures instead of continuing as if they succeeded.
  * Incompatible themes now show clearer error messaging when removal fails, rather than proceeding to the success redirect flow.
  * Recursive theme cleanup now stops on the first failure and reports an overall success/failure outcome.
  * Compatibility checks now return the actual computed result to keep subsequent decisions consistent.

### DIFF
--- a/application/models/TemplateConfig.php
+++ b/application/models/TemplateConfig.php
@@ -863,7 +863,7 @@ class TemplateConfig extends CActiveRecord
     /**
      * Uninstalls the selected surveytheme and deletes database entry and configuration
      * @param string $templatename Name of Template
-     * @return bool|int
+     * @return bool|int, if false Template are not uninstalled, integr mean the number of TemplateConfiguration uninstalled
      * @throws CDbException
      */
     public static function uninstall($templatename)
@@ -876,9 +876,12 @@ class TemplateConfig extends CActiveRecord
                         'template_name=:templateName',
                         [':templateName' => $templatename]
                     );
-                    /* Add the already deleted by findByAttributes */
-                    $count++;
-                    return $count;
+                    if ($count) {
+                        /* Number of TemplateConfiguration uninstalled */
+                        return $count;
+                    }
+                    /* No TemplateConfiguration, only Template */
+                    return true;
                 }
             }
         }
--- a/application/models/TemplateConfig.php
+++ b/application/models/TemplateConfig.php
@@ -956,14 +956,13 @@ class TemplateConfig extends CActiveRecord
         }
         // add more tests here
 
-        // all checks succeeded, continue loading the theme
-        return $isCompatible;
+        // all checks succeeded, continue loading the theme if it compatible
+        return boolval($isCompatible);
     }
 
     /**
      * Checks if theme is compatible with the current limesurvey version
      * @param $themePath
-     * @param bool $redirect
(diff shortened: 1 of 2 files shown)


application/models/TemplateConfig.php
```diff
@@ -885,18 +885,20 @@ public static function uninstall($templatename)
...
             foreach ($extendedTemplates as $extendedTemplate) {
-                self::uninstallThemesRecursive($extendedTemplate->name);
+                if (!self::uninstallThemesRecursive($extendedTemplate->name)) {
+                    return false;
+                }
             }
         }
-        self::uninstall($templateName);
+        return boolval(self::uninstall($templateName));
     }
```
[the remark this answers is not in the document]
Shnoulle: Here : deleting by name with installed theme. **Must** return one line minimum

application/models/TemplateConfig.php
```diff
@@ -863,7 +863,7 @@ public function __toString()
     /**
      * Uninstalls the selected surveytheme and deletes database entry and configuration
      * @param string $templatename Name of Template
-     * @return bool|int
+     * @return bool|int, if false Template are not uninstalled, integr mean the number of TemplateConfiguration uninstalled
      * @throws CDbException
```
gabrieljenik, line 871: I would change this to be a bit more clear (in these days where AI read the comments :) )
> If true, No TemplateConfiguration were uninstalled, only the Template was uninstalled. If false Template was not uninstalled. If integer, it means the Template was unsintalled and tells how many TemplateConfigurations were uninstalled.

application/models/TemplateConfig.php
```diff
@@ -885,18 +891,20 @@ public static function uninstall($templatename)
...
         if (!empty($extendedTemplates)) {
             foreach ($extendedTemplates as $extendedTemplate) {
-                self::uninstallThemesRecursive($extendedTemplate->name);
+                if (!self::uninstallThemesRecursive($extendedTemplate->name)) {
+                    return false;
+                }
             }
         }
-        self::uninstall($templateName);
+        return boolval(self::uninstall($templateName));
```
gabrieljenik, line 911: If unsinstall returns zero, how would this be evaluated?

Shnoulle: Checked here : https://github.com/LimeSurvey/LimeSurvey/pull/5117/changes#diff-508c0106b60d82f7d3a33272d04c1272ade4e7cc8eeb81d9f3a5e6ecc0ac1570R879

No TemplaceConfiguration deleted (0) but Template deleted : return true.

gabrieljenik (approved):

application/models/TemplateConfig.php
Shnoulle: Return value can be : 
    - true : Template was uninstalled, No TemplateConfiguration were uninstalled
    - integer : Template was unsintalled and tells how many TemplateConfigurations were uninstalled
    - false : Template was not uninstalled : lack of permission Not template of this name, issue when delete.

PR #5127, opened 2026-07-10
LE-778: Ranking json fixes

Ahmedazeem98:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how selected answers are displayed for additional list-style question types.
  * “Other” now consistently uses the translated label.
  * Ranking questions now show localized “Rank N: …” entries based on the selected subquestions.
  * Language responses now display readable language names instead of raw codes.
* **Chores**
  * Refined sub-question metadata generation during XML export (no user-visible behavior change).

### DIFF
--- a/application/helpers/common_helper.php
+++ b/application/helpers/common_helper.php
@@ -1048,7 +1048,11 @@ function getExtendedAnswer($iSurveyID, $sFieldCode, $sValue, $sLanguage, $questi
             case Question::QT_L_LIST:
             case Question::QT_EXCLAMATION_LIST_DROPDOWN:
             case Question::QT_O_LIST_WITH_COMMENT:
-            case Question::QT_I_LANGUAGE:
+                $this_answer = Answer::model()->getAnswerFromCode($fields['qid'], $sValue, $sLanguage);
+                if ($sValue == "-oth-") {
+                    $this_answer = gT("Other", null, $sLanguage);
+                }
+                break;
             case Question::QT_R_RANKING:
                 $items = json_decode($sValue, true);
                 if (is_array($items)) {
@@ -1163,8 +1167,13 @@ function getExtendedAnswer($iSurveyID, $sFieldCode, $sValue, $sLanguage, $questi
                     }
                 }
                 break;
+            case Question::QT_I_LANGUAGE:
+                if (!empty($sValue)) {
+                    $this_answer = getLanguageNameFromCode($sValue, false);
+                }
+                break;
             default:
-                ;
+                break;
         } // switch
     }
     switch ($sFieldCode) {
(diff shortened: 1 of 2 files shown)


application/helpers/common_helper.php
```diff
@@ -1163,8 +1167,13 @@ function getExtendedAnswer($iSurveyID, $sFieldCode, $sValue, $sLanguage, $questi
                     }
                 }
                 break;
+            case Question::QT_I_LANGUAGE:
```
lajosarpad: Why was QT_I_LANGUAGE separated from QT_L_LIST & co? What was the problem and why is this a solution to it?

application/helpers/common_helper.php
lajosarpad: Let's load the ranking subquestions at once rather than querying for them in a loop, if you have 60 subquestions, then just after is_array turned out to be true you could use a single query, like

```
$subquestions = Question::model()->findAll([
    'condition' => 'parent_qid = :qid',
    'order' => 'question_order',
    'params' => [':qid' => $fields['qid']]
]);
```

then map the question titles to questions and you will be able to check whether the code you have exists there.

lajosarpad (changes requested):
Please explain the need to change on the language types and please apply the suggested improvement on subquestion finding.

application/helpers/common_helper.php
Ahmedazeem98: Because it is not an answer. As shown below, it is incorrectly resolved as an answer. In this case, the result is displayed as **en**
 ```
case Question::QT_L_LIST:
            case Question::QT_EXCLAMATION_LIST_DROPDOWN:
            case Question::QT_O_LIST_WITH_COMMENT:
            case Question::QT_I_LANGUAGE:
                $this_answer = Answer::model()->getAnswerFromCode($fields['qid'], $sValue, $sLanguage);
                if ($sValue == "-oth-") {
                    $this_answer = gT("Other", null, $sLanguage);
                }
                break;
```

After the changes the result is **English [en]**

application/helpers/common_helper.php
Ahmedazeem98: updated

application/helpers/common_helper.php
lajosarpad: We'll need @tiborpacalat to test and confirm this, I lack the context of this feature and to me it's unclear why **en** was bad and **English [en]** is good. Can you look into it @tiborpacalat ?

application/helpers/common_helper.php
lajosarpad: Indeed, this looks great!

application/helpers/common_helper.php
Ahmedazeem98: The main issue is that an unnecessary search is performed because the value cannot be found in the answers table.
Question::QT_I_LANGUAGE does not belong to the List question family.

Ahmedazeem98: I reverted this change. We can address it in a separate ticket if needed.

lajosarpad (approved):

PR #5146, opened 2026-07-17
Fixed issue #20606: Response table schema is incorrect when activatin…

haythemch2:
…g a survey after previewing

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed survey activation so outdated preview or session response data is cleared before creating the response table.
  * Ensured newly activated surveys start with clean response data.

### DIFF
--- a/application/models/SurveyActivator.php
+++ b/application/models/SurveyActivator.php
@@ -66,6 +66,11 @@ class SurveyActivator
             return ['error' => 'plugin'];
         }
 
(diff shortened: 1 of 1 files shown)


Ahmedazeem98 (approved):
I would suggest to do =>
`Yii::app()->session->remove('responses_' . $this->survey->sid);`
Better to use the framework session, and avoids the redundant variable.

PR #5147, opened 2026-07-17
Fixed issue #20606: Response table schema is incorrect when activatin…

haythemch2:
…g a survey after previewing

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed survey activation to ignore stale preview or session data.
  * Ensured activation tables are created using current survey responses.


Ahmedazeem98 (approved):
I would suggest to do =>
`Yii::app()->session->remove('responses_' . $this->survey->sid);`
Better to use the framework session, and avoids the redundant variable.

PR #5159, opened 2026-07-22
Fixed issue #20392: invalid URL placeholders when using hostInfo or baseUrl and publicurl

Shnoulle:
Dev: set hostInfo and baseurl to needed way if publicurl is set
Dev: reset after get the url

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved public URL generation when the configured “public URL” base differs from the current application base, with correct host and prefix handling across both script-name and non-script-name routing modes.
* **Tests**
  * Expanded coverage for public URL creation to validate behavior across multiple URL manager settings (path vs. get formats, and showScriptName true/false).
* **Chores**
  * Reformatted the phpunit configuration opening tag for consistency (no functional changes).

### DIFF
--- a/application/core/Traits/LSApplicationTrait.php
+++ b/application/core/Traits/LSApplicationTrait.php
@@ -47,11 +47,27 @@ trait LSApplicationTrait
     public function createPublicUrl($route, $params = array(), $schema = '', $ampersand = '&')
     {
         $sPublicUrl = $this->getPublicBaseUrl(true);
-        $sActualBaseUrl = $this->getBaseUrl(true);
-        if ($sPublicUrl !== $sActualBaseUrl) {
+        $sActualAbsoluteBaseUrl = $this->getBaseUrl(true);
+        if ($sPublicUrl !== $sActualAbsoluteBaseUrl) {
+            /* @var string keep current baseUrl */
+            $sActualBaseUrl = $this->getBaseUrl(false);
+            /* @var string keep current hostInfo */
+            $sActualhostInfo = $this->getRequest()->getHostInfo();
+            /* Set hostInfo to empty and baseUrl according to showScriptName (needed) */
+            if ($this->getUrlManager()->showScriptName) {
+                $this->getUrlManager()->setBaseUrl("/index.php");
+            } else {
+                $this->getUrlManager()->setBaseUrl("");
+            }
+            $this->getRequest()->setHostInfo("");
+            /* @var string the url */
             $url = $this->createAbsoluteUrl($route, $params, $schema, $ampersand);
-            if (substr((string)$url, 0, strlen((string)$sActualBaseUrl)) == $sActualBaseUrl) {
-                $url = substr((string)$url, strlen((string)$sActualBaseUrl));
+            /* Reset baseUrl and hostInfo to previous one */
+            $this->getUrlManager()->setBaseUrl($sActualBaseUrl);
+            $this->getRequest()->setHostInfo($sActualhostInfo);
+            /* Replace Yii public uirl by publicuirl set in config */
+            if (substr((string)$url, 0, strlen((string)$sActualAbsoluteBaseUrl)) == $sActualAbsoluteBaseUrl) {
+                $url = substr((string)$url, strlen((string)$sActualAbsoluteBaseUrl));
             }
             return trim((string)$sPublicUrl, "/") . $url;
         } else {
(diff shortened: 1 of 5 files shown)


Shnoulle:
Shit : 
```
        Yii::app()->getUrlManager()->setBaseUrl("/master");
        Yii::app()->setConfig('publicurl', 'http://www.example.com/limesurvey');
        tracevar([
            Yii::app()->createPublicUrl('controller/action'),
            Yii::app()->createUrl('controller/action')
        ]);
```
result with `'showScriptName' => false,`

```
array
(
    0 => 'http://www.example.com/limesurvey/controller/action'
    1 => '/controller/action'
) 
```

result with `'showScriptName' => true,`

```
array
(
    0 => 'http://www.example.com/limesurvey/index.php/controller/action'
    1 => '/index.php/controller/action'
)
```

Shnoulle:
There are an issue in the order : 'http://www.example.com/index.php/controller/action' is the cirrent value returend : it's OK .
 'http://www.example.com/controller/action'  it's the value return by `Yii::app()->createUrl('controller/action')` this one is invalid !

I review the test and the issue about Yii::app()->createUrl

Shnoulle:
OK, Actual is OK, Expected is not OK : use fixed string for exapected

Shnoulle:
Shit !!

```
t$ DOMAIN=lstest.local ./vendor/bin/phpunit  ./tests/unit/LSYiiApplicationTest.php
Error Reporting and Debug: No
Set $forceDebug=true in tests/bootstrap.php to enable more logging.

PHPUnit 9.6.34 by Sebastian Bergmann and contributors.

.........                                                           9 / 9 (100%)

Time: 00:00.121, Memory: 22.00 MB

OK (9 tests, 12 assertions)

Generating code coverage report in Clover XML format ... ^[[A^[[A^[[B^[[Bdone [00:03.409]

```

application/core/Traits/LSApplicationTrait.php
```diff
@@ -47,11 +47,27 @@ public function getCurrentUserId()
     public function createPublicUrl($route, $params = array(), $schema = '', $ampersand = '&')
     {
         $sPublicUrl = $this->getPublicBaseUrl(true);
-        $sActualBaseUrl = $this->getBaseUrl(true);
-        if ($sPublicUrl !== $sActualBaseUrl) {
+        $sActualAbsoluteBaseUrl = $this->getBaseUrl(true);
+        if ($sPublicUrl !== $sActualAbsoluteBaseUrl) {
```
gabrieljenik, line 55: I think I have been through this function more than couple of times. Do you mind adding a comment on why do we need this if? what is the intention in case the public url doesn't match the baseurl?

gabrieljenik:
@coderabbitai: is there any test scenarion that is missing for the method  createPublicUrl()?

gabrieljenik:
@Shnoulle What do you think about https://github.com/LimeSurvey/LimeSurvey/pull/5159#issuecomment-5096198751 ? I think that is a non-existant scenario, right?

application/core/Traits/LSApplicationTrait.php
Shnoulle: I don't understand: here, I just update the `$sActualBaseUrl` to `$sActualAbsoluteBaseUrl` variable name.

Here we compare `config['publicurl']` to the _real_ URL: is it the comment you want?

Shnoulle:
> @Shnoulle What do you think about [#5159 (comment)](https://github.com/LimeSurvey/LimeSurvey/pull/5159#issuecomment-5096198751) ? I think that is a non-existant scenario, right?

I think the other one does the job, but yes, it can be clearer.

gabrieljenik:
@coderabbitai: regarding this comment (https://github.com/LimeSurvey/LimeSurvey/pull/5159#discussion_r3667434993), why does it "... It temporarily removes the host and sets the base path to ..." ?

gabrieljenik:
@Shnoulle 

> reuse Yii’s routing logic to generate the route/script fragment independently of the current deployment host and base directory, then safely prepend publicurl.

This is a comment I would add to the method as to make it more clear.

gabrieljenik (approved):
Code structure looks good.
Automated tests added.

Shnoulle:
> @Shnoulle
> 
> > reuse Yii’s routing logic to generate the route/script fragment independently of the current deployment host and base directory, then safely prepend publicurl.
> 
> This is a comment I would add to the method as to make it more clear.

You know you already work on this function?

I really think 

```
     * Creates an absolute URL based on the given controller and action information.
     * Check publicurl from config to get the final url
```

Is clear enough !

Shnoulle:
> @Shnoulle
> 
> > reuse Yii’s routing logic to generate the route/script fragment independently of the current deployment host and base directory, then safely prepend publicurl.
> 
> This is a comment I would add to the method as to make it more clear.

https://github.com/LimeSurvey/LimeSurvey/commit/ad076b19382b806bf530a8ee5e29b95be3d9a07d

PR #5160, opened 2026-07-22
Feature/AT-2179 selectall floating menu

Trischi80:
New feature #AT-2179: Additional button "Select all" in flaoting menu. Should check all checkboxes in gridviews over pagination

## Summary by CodeRabbit

- **New Features**
  - Added a “Select all” option to the survey list’s floating actions to select all matching surveys across pagination.
  - “Select all” respects the current grid filters and updates selection state, counts, and visible row checkboxes.
  - Added a new AJAX-backed capability to retrieve all matching survey identifiers for bulk selection.
- **Bug Fixes**
  - Fixed the redirect after deleting a URL parameter in survey administration by correcting an extra trailing slash.

### DIFF
--- a/application/controllers/SurveyAdministrationController.php
+++ b/application/controllers/SurveyAdministrationController.php
@@ -52,6 +52,7 @@ class SurveyAdministrationController extends LSBaseController
                     'getCurrentEditorValues',
                     'getDataSecTextSettings',
                     'getDateFormatOptions',
+                    'getAllSurveyIds',
                     'updateAccessMode',
                     ''
                 ],
@@ -313,6 +314,41 @@ class SurveyAdministrationController extends LSBaseController
                 'caption'     => gT('Selected surveys'),
             )
         );
+    }
+
+    /**
+     * Returns all survey IDs matching the current filter as a JSON array.
+     * Used by the "Select all" button in the floating actions bar to select
+     * all surveys across pagination.
+     *
+     * @return void
+     */
+    public function actionGetAllSurveyIds()
+    {
+        if (!Yii::app()->request->isAjaxRequest) {
+            throw new CHttpException(400, 'Invalid request');
+        }
+        if (!Permission::model()->hasGlobalPermission('surveys', 'read')) {
+            throw new CHttpException(403, 'Forbidden');
+        }
+
+        $model = new Survey('search');
+        $filters = Yii::app()->request->getParam('Survey', []);
+        if (is_array($filters)) {
+            $model->setAttributes($filters, false);
+        }
+
+        // Get all matching surveys without pagination limit
+        $dataProvider = $model->search(['pageSize' => 100000]);
+        $surveys = $dataProvider->getData();
+
+        $ids = array_map(static function ($survey) {
+            return (int) $survey->sid;
+        }, $surveys);
+
+        header('Content-Type: application/json');
+        echo json_encode(array_values($ids));
+        Yii::app()->end();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


application/extensions/admin/grid/FloatingActionsWidget/views/floating_bar.php
```diff
@@ -21,6 +23,20 @@ class="floating-actions-bar"
...
+        <!-- Select all button -->
+        <button
+            type="button"
+            class="floating-actions-btn floating-actions-select-all"
+            title="<?= gT('Select all') ?>"
+        >
+            <i class="ri-check-line"></i>
+            <?= gT('Select all') ?>
+        </button>
+    <?php endif; ?>
```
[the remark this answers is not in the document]
Trischi80: No, it was a design decision and should stay like it is.

application/controllers/SurveyAdministrationController.php
```diff
@@ -315,6 +316,60 @@ public function actionRenderItemsSelected()
...
+        $dataProvider = $model->search();
+        $dataProvider->setPagination(false);
+        $surveys = $dataProvider->getData();
+
+        $ids = array_map(static function ($survey) {
+            return (int) $survey->sid;
+        }, $surveys);
+
+        header('Content-Type: application/json');
+        echo json_encode(array_values($ids));
```
karcstefan, line 369: should call `$this->renderJSON()` from LSBaseController

application/controllers/SurveyAdministrationController.php
```diff
@@ -315,6 +316,60 @@ public function actionRenderItemsSelected()
...
+        if (!empty($activeFilter)) {
+            $model->active = $activeFilter;
+        }
+
+        // Apply survey-group filter
+        $gsid = Yii::app()->request->getParam('gsid', '');
+        if (!empty($gsid)) {
+            $model->gsid = (int) $gsid;
+        }
+
```
karcstefan, line 355: Are there any other filters, like text searched to be applied here?

application/controllers/SurveyAdministrationController.php
```diff
@@ -315,6 +316,60 @@ public function actionRenderItemsSelected()
...
+        // Fetch every matching ID in a single unbounded query.
+        // getData() only returns one page, so a large fixed pageSize can still miss
+        // rows on larger installations.  Disabling pagination removes the LIMIT/OFFSET
+        // entirely so all matching survey IDs are always returned.
+        $dataProvider = $model->search();
+        $dataProvider->setPagination(false);
+        $surveys = $dataProvider->getData();
```
karcstefan, line 362: We can rethink of this approach later, but bulk returning and applying actions this way can be a bit tricky, think of 20 000 entries or more and peforming check/uncheck

karcstefan (changes requested):

application/controllers/SurveyAdministrationController.php
Trischi80: There are 3 filters. The text search, status and group. All of them should work, when getting the survey ids for checked rows.

application/controllers/SurveyAdministrationController.php
Trischi80: The function renderJSON() does not work in this case for the feature "Select all".

application/controllers/SurveyAdministrationController.php
Trischi80: Yes, it would be interesting to check this one for performance with a lot of datasets.

karcstefan (approved):

PR #5161, opened 2026-07-22
Fixed issue #20596: url params (from panel integration) were not copi…

Trischi80:
Fixed issue #20596: url params were not copied in the new copy survey process

## Summary by CodeRabbit

* **New Features**
  * When copying a survey, panel integration URL parameters are now retained on the new survey.
  * URL parameters are automatically re-linked to the corresponding copied questions and subquestions (when available).
  * Parameters that can’t be mapped to valid copied questions/subquestions are skipped to prevent incorrect links.

### DIFF
--- a/application/models/services/CopySurvey.php
+++ b/application/models/services/CopySurvey.php
@@ -19,6 +19,7 @@ use QuestionL10n;
 use Survey;
 use Permission;
 use SurveyLanguageSetting;
+use SurveyURLParameter;
 use Template;
 use Yii;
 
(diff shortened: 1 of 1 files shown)


application/models/services/CopySurvey.php
```diff
@@ -601,4 +609,55 @@ private function copyDefaultAnswers($mappingQuestionIds, $mappedSuquestionIds)
...
+    {
+        $sourceParams = SurveyURLParameter::model()->findAllByAttributes(['sid' => $this->sourceSurvey->sid]);
+        $cntErrors = 0;
+
+        foreach ($sourceParams as $sourceParam) {
+            // targetqid must resolve to a copied question; skip if it can't be mapped
+            // (e.g. the question was excluded or the source data is inconsistent).
+            $newTargetQid = null;
+            if (!empty($sourceParam->targetqid)) {
+                if (!isset($mappingQuestionIds[$sourceParam->targetqid])) {
```
SaraDimoska, line 635: If a parameter can't be mapped it's skipped silently, and the user still gets a "Success" screen with the parameter missing. Maybe we can add '$cntErrors++' before the 'continue;' (here and in the targetsqid block below) so the existing warning is shown and the user knows a parameter wasn't copied. 
Just a suggestion :)

PR #5185, opened 2026-07-31
Fixed issue #20624: Error when updating to 7.0.7 due to missing DB version check

c-schmitz:
## Summary by CodeRabbit

* **New Features**
  * Added database compatibility checks for MariaDB, MySQL, PostgreSQL, and Microsoft SQL Server.
  * Installer validation now alerts users when the connected database version is unsupported.
  * Compatibility checks account for recognized server version formats and database-specific requirements.

* **Bug Fixes**
  * Database upgrades now stop safely on unsupported server versions before migrations begin.
  * Upgrade attempts provide an error message when compatibility requirements are not met.
  * Failures while detecting database version information are handled safely and logged.


gabrieljenik (approved):
Haven't tested it. I think I would have used something like inheritance for the DBVersionHelper, but this is so little that doesn't matter.

PR #5196, opened 2026-08-03
Fixed issue #20629: Loading fonts locally instead from cdn

karcstefan:
## Summary by CodeRabbit

* **Improved Experience**
  * Added bundled IBM Plex Sans fonts, including italic and multiple weight options, for more consistent typography.
  * Improved loading and display of editor styles, controls, forms, navigation, maps, and responsive layouts.
  * Updated editor assets to support the latest interface build and more reliable styling resource loading.

* **Accessibility**
  * Preserved focus states and other accessibility-focused interface styling.


c-schmitz (changes requested):
This patch does not work if LimeSurvey is installed in a subfolder.

haythemch2 (dismissed):

karcstefan:
@c-schmitz good catch, and very tricky issue too.
Trying out different approaches but the simplest one wins imo, relative URL in scss and let the loader copy over the fonts . Downside is there will be duplicate fonts but it's bulletproof for all cases.

c-schmitz (approved):

PR #5197, opened 2026-08-04
New feature: Statistics v2 for the react Question Editor

karcstefan:
## Summary by CodeRabbit

* **New Features**
  * Added paginated, language-aware survey statistics and response APIs.
  * Added searchable response and comment views with highlighted terms and “Load more” support.
  * Added advanced filters for questions, participants, dates, languages, completion status, uploads, and numeric ranges.
  * Added interactive chart, table, grid, ranking, comment, and numerical-data views.
  * Added percentage/count switching, localized content, and comment drill-downs.
  * Added resizable tables, collapsible menus, improved dropdowns, and reusable search controls.

* **Bug Fixes**
  * Improved validation for filters, sorting, selected fields, and search terms.
  * Improved handling of missing answers and localized labels.


editor/src/components/UIComponents/LSTable/LSTable.js
ramishenouda48, line 1: Since we already use `@tanstack/react-table` for the Responses table, would it make sense to use it here as well?

We are currently implementing sorting, row selection, and especially column resizing manually. The resizing logic in particular requires document-level mouse/touch listeners, cleanup, and changes to `document.body` styles.

Using TanStack here could reduce the amount of custom table logic we need to maintain and would also keep our table implementations more consistent across the codebase. Since TanStack is headless, we can still keep the existing Bootstrap markup, styling, and LimeSurvey-specific UI.

Unless there is a specific reason for keeping this table independent from TanStack, I think reusing the approach we already have in the Responses table could be worth considering.

ramishenouda (approved):

PR #2353, opened 2022-04-14
Fixed issue #18026: Session expire message when submit button is clicked twice

gabrieljenik:

PR #2337, opened 2022-04-08
Fixed issue #18026: Session expire message when submit button is clicked twice

gabrieljenik:

PR #5213, opened 2026-08-06
Fix issue #18026: Make the final survey submit idempotent

1000i100:
## Summary

- make the final survey submit idempotent server side: a replayed final submit displays the completed page again, instead of reporting an expired session
- skip on a replay everything that must happen only once: answer processing, confirmation and notification emails, `afterSurveyComplete`
- keep a client side guard as a second line of defence: form level lock, plus a fallback for browsers without `SubmitEvent.submitter`
- add functional browser regression tests

> **Note** — this PR changed approach after field testing. It started as a client side lock only; the first two commits are that lock, the third one is the actual fix. See "Why the client side guard is not enough" below.

## Context and root cause

Follow-up to [LimeSurvey bug #18026](https://bugs.limesurvey.org/view.php?id=18026), which reports an expired session page after the final submit button is clicked twice.

The earlier fixes in #2337 (master) and #2353 (3.x-LTS) disable the submit buttons when the first final `submit` event is handled. That prevents later clicks, but it does not cancel a second `submit` event the browser has already queued, so both requests proceed.

Guarding the click is treating a symptom. The defect is that **the final submit is not idempotent**, in `SurveyIndex`:

```php
private function didSessionTimeout($surveyid)
{
    return (!isset($_SESSION['responses_' . $surveyid]['step']) && null !== App()->request->getPost('thisstep'));
}
```

This predicate conflates two different states: a session that never started or timed out, and a session that finished normally and was cleaned up. `SurveyRuntimeHelper::moveSubmitIfNeeded()` calls `killSurveySession()` right after sending the completed page, so any replayed POST lands in the first branch and is answered with *"We are sorry but your session has expired."*

At that moment the response is fully recorded: `submitdate` set, token consumed, participant counted as *Completed*. **The participant is told everything is lost while the server counted a success.**

A click is only one way to replay that request. The others are not reachable from JavaScript at all:

- double click while the request is in flight
- browser "Try again" prompt after a back navigation (Firefox shows *"The document expired"* with a button that resends the POST)
- reload of the result page
- restored tab

## Why the client side guard is not enough

Testing on a production-like install (LimeSurvey 7.0.11, shared hosting, SMTP relay) gave these figures:

| | server time |
|---|---|
| the 9 first "Next" | **160 – 249 ms** |
| the last "Next" | **4 496 – 5 948 ms** |
| the last "Next", with `emailmethod = mail()` | **1 484 ms** |

Roughly 3.4 s of SMTP conversation (participant confirmation and admin notification) and 1.3 s of completion work. During those seconds the page is unchanged and the button still looks active. The observed patience threshold before a participant clicks again is about 1.5 s — a factor of three, not a margin.

The reclick is not a user mistake, it is a correct inference from a false signal: a button that stays active after a click says "nothing was taken into account".

## The fix

`SurveyRuntimeHelper` sets the finished session aside before killing it. `SurveyIndex` puts it back when a final submit is replayed, so the completed page is rendered again.

On a replay, the one time work is skipped:

- the answers are not processed again (`GetLastMoveResult` instead of `NavigateForwards`)
- `submittokens()` and `sendSubmitNotifications()` are not called again
- `afterSurveyComplete` is **not** dispatched again, so plugins cannot act twice

That last point is worth flagging for review. Today a replay never reaches this code, so plugins are never called twice; making the render replayable without that guard would have introduced a double dispatch, and a plugin that charges or notifies would act twice. The price is that the plugin content block is missing from the replayed page, which is the right side of that trade.

A session that is really gone still reports an expired session: the finished session is only restored when this very browser session completed the survey, and starting a new run (`newtest=Y`) drops it.

**Known cost**: the survey session is now released on the next entry rather than at the end of the run. For a participant who finishes and leaves, it lives until session GC instead of being dropped immediately.

## Measurements

Local install, verified before and after each change.

| scenario | before | after |
|---|---|---|
| **two simultaneous final POSTs** (double click) | one of them → expired session | both → completed page |
| same final POST replayed 5 times | expired session | completed page every time |
| notification emails sent (1 submit + 4 replays) | — | **1** |
| responses recorded | — | **1** |
| survey session actually dropped server side | expired session | expired session (unchanged) |
| next / previous / submit / new run | ok | ok |
| final submit without `SubmitEvent.submitter` (Safari < 15.4) | second submit not blocked | second submit blocked |

## Client side commits

`SubmitEvent.submitter` is unavailable on Safari below 15.4, whose version is tied to the OS on macOS and iOS, so those users cannot always update. The handler returned immediately there, leaving no protection at all. The fallback remembers the clicked submit button and forgets it on the next tick, so a stale button is never used for a later submit, and script triggered submits keep their current behaviour: they send no button along with the request, so no submitter is inferred for them.

With the server fix in place, this guard is no longer load bearing — it just avoids a pointless request and the seconds of waiting that come with it.

## Deliberately not in this PR

- **the final request blocking on SMTP** (4.9 s against 160–249 ms for every other page). Moving the emails after the page is rendered would bring the last "Next" down to ~1.5 s. Separate concern, separate change.
- **an in-page waiting indication** (disabled button, "sending…" label). Worth doing, but it is ergonomics, not correctness. Notably, the browser *did* signal the wait — stop button, tab spinner, status bar — and the reclick happened anyway; nobody watches the status bar while answering a questionnaire.

## Validation

- reproduced before the fix and verified after, on a local LimeSurvey install with Chromium 151, both over plain HTTP and through WebDriver
- the two simultaneous POSTs scenario reproduces the field failure exactly: red before, green after
- `php -l` on the three changed PHP files
- `node --check` on the three changed JavaScript files

The PHPUnit WebDriver suite itself was not run: this checkout has no PHP binary nor the development dependencies. The logic of each added test was executed against a live install through WebDriver, red before the fix and green after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate final survey submissions, including implicit submissions.
  * Improved submit-button detection and value handling across supported browsers, including older Safari versions.
  * Ignored submissions that were already canceled or prevented.
  * Preserved final submission safeguards by disabling controls and retaining the selected submit value.
  * Replayed final submissions now display the completed survey page without reprocessing responses or triggering completion actions.

* **Tests**
  * Added coverage for repeated final submissions and completed-page display.

### DIFF
--- a/assets/packages/limesurvey/survey.js
+++ b/assets/packages/limesurvey/survey.js
@@ -483,16 +483,28 @@ $(document).on('submit', function (e) {
     if (e.target.id != 'limesurvey') {
         return;
     }
+    var $form = $(e.target);
     // We only care about the final submit, not normal forward/backward navigation.
     var submitter = e.originalEvent ? $(e.originalEvent.submitter) : null;
     if (!submitter || submitter.attr('value') != 'movesubmit') {
         return;
     }
+    if (e.isDefaultPrevented()) {
+        return;
+    }
+    // Disabling the button is not sufficient when the browser has already queued
+    // more than one submit event. Explicitly cancel every repeated final submit.
+    if ($form.data('final-submit-active')) {
+        e.preventDefault();
+        e.stopImmediatePropagation();
+        return false;
+    }
+    $form.data('final-submit-active', true);
     // Still, we disable all submit buttons to make sure the "back" button is not
     // pressed while submitting.
-    $('#limesurvey button[type="submit"]').prop('disabled', true);
+    $form.find('button[type="submit"]').prop('disabled', true);
 
     // We also add a hidden input with the button's value, because it's not included
     // in the request when the button is disabled.
-    $('#limesurvey').append('<input id="onsubmitbuttoninput" name=\'' + submitter.attr('name') + '\' value=\'' + submitter.attr('value') + '\' type=\'hidden\' />');
+    $form.append('<input id="onsubmitbuttoninput" name=\'' + submitter.attr('name') + '\' value=\'' + submitter.attr('value') + '\' type=\'hidden\' />');
 });
--- a/assets/survey_themes/fruity_twentythree/core/old_template_core_pre.js
+++ b/assets/survey_themes/fruity_twentythree/core/old_template_core_pre.js
@@ -460,18 +460,30 @@ $(document).on('submit', function (e) {
     if (e.target.id != 'limesurvey') {
         return;
     }
+    var $form = $(e.target);
     // We only care about the final submit, not normal forward/backward navigation.
     var submitter = e.originalEvent ? $(e.originalEvent.submitter) : null;
     if (!submitter || submitter.attr('value') != 'movesubmit') {
         return;
     }
+    if (e.isDefaultPrevented()) {
+        return;
+    }
+    // Disabling the button is not sufficient when the browser has already queued
+    // more than one submit event. Explicitly cancel every repeated final submit.
+    if ($form.data('final-submit-active')) {
+        e.preventDefault();
+        e.stopImmediatePropagation();
+        return false;
+    }
+    $form.data('final-submit-active', true);
     // Still, we disable all submit buttons to make sure the "back" button is not
     // pressed while submitting.
-    $('#limesurvey button[type="submit"]').prop('disabled', true);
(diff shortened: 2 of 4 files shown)


1000i100:
@c-schmitz Hello, do you plan to merge this ?
I need this to have confidence using limesurvey in my work, so i would like to know if i need to use my fork directly, or if you will merge and i can continue with confort update mecanism.

PR #5224, opened 2026-08-11
Fixed issue #19622: Mail not possible if name contains a < char

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of email addresses that include display names containing `<` characters.
  * Correctly preserves names when parsing sender and recipient details.
  * Supports explicitly empty sender names when an address begins with `<`.

### DIFF
--- a/application/core/LimeMailer.php
+++ b/application/core/LimeMailer.php
@@ -398,16 +398,19 @@ class LimeMailer extends PHPMailer
     public function setFrom($from, $fromname = null, $auto = true)
     {
         $fromemail = $from;
-        if (strpos($from, '<') !== false) {
+        /* Email is always the last «<email>» part: use the last angle brackets so a name containing '<' is kept intact */
+        $ltPosition = strrpos($from, '<');
+        if ($ltPosition !== false) {
             if (is_null($fromname)) {
-                if (strpos($from, '<')) {
-                    $fromname = trim(substr($from, 0, strpos($from, '<') - 1));
+                if ($ltPosition > 0) {
+                    $fromname = trim(substr($from, 0, $ltPosition));
                 } else {
                     /* Allow to force empty name in token email */
                     $fromname = '';
                 }
             }
-            $fromemail = substr($from, strpos($from, '<') + 1, strpos($from, '>') - 1 - strpos($from, '<'));
+            $gtPosition = strrpos($from, '>');
+            $fromemail = substr($from, $ltPosition + 1, $gtPosition - 1 - $ltPosition);
         }
         if (is_null($fromname)) {
             $fromname = $this->FromName;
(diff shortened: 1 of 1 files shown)


application/core/LimeMailer.php
```diff
@@ -398,16 +398,19 @@ public function setTypeWithRaw($emailType, $language = null)
     public function setFrom($from, $fromname = null, $auto = true)
     {
         $fromemail = $from;
-        if (strpos($from, '<') !== false) {
+        /* Email is always the last «<email>» part: use the last angle brackets so a name containing '<' is kept intact */
```
Shnoulle, line 401: No way to use a regexp ? 

```
if (preg_match('/^.*<([^<>]*@[^<>]*)>$/', $string, $matches)) {
    $email = $matches[1];// Check it after
}
```

application/core/LimeMailer.php
```diff
@@ -435,10 +438,13 @@ public function setTo($addressTo, $name = '')
     public function addAddress($addressTo, $name = '')
     {
         $address = $addressTo;
-        if (strpos($address, '<')) {
-            $address = substr($addressTo, strpos($addressTo, '<') + 1, strpos($addressTo, '>') - 1 - strpos($addressTo, '<'));
+        /* Email is always the last «<email>» part: use the last angle brackets so a name containing '<' is kept intact */
```
Shnoulle, line 441: 2 time same code : need a new function.

Shnoulle (commented):
For preg_match: it's a why not
For a new function: I think it's needed

application/core/LimeMailer.php
```diff
@@ -435,10 +438,13 @@ public function setTo($addressTo, $name = '')
...
+        $ltPosition = strrpos($addressTo, '<');
+        if ($ltPosition !== false) {
+            $gtPosition = strrpos($addressTo, '>');
+            $address = substr($addressTo, $ltPosition + 1, $gtPosition - 1 - $ltPosition);
             if (empty($name)) {
-                $name = trim(substr($addressTo, 0, strpos($addressTo, '<') - 1));
+                $name = trim(substr($addressTo, 0, $ltPosition));
```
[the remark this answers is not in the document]
Shnoulle: true

application/core/LimeMailer.php
```diff
@@ -398,16 +398,19 @@ public function setTypeWithRaw($emailType, $language = null)
...
+                if ($ltPosition > 0) {
+                    $fromname = trim(substr($from, 0, $ltPosition));
                 } else {
                     /* Allow to force empty name in token email */
                     $fromname = '';
                 }
             }
-            $fromemail = substr($from, strpos($from, '<') + 1, strpos($from, '>') - 1 - strpos($from, '<'));
+            $gtPosition = strrpos($from, '>');
+            $fromemail = substr($from, $ltPosition + 1, $gtPosition - 1 - $ltPosition);
```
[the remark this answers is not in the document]
Shnoulle: Seem true (regexp_match fix the issue)

gabrieljenik:
Why not just simplify and trim < from the "from" fields?

Shnoulle:
> Why not just simplify and trim < from the "from" fields?

For `firstname < lastname <firstname@example.org>` ?

gabrieljenik:
> > Why not just simplify and trim < from the "from" fields?
> 
> For `firstname < lastname <firstname@example.org>` ?

Let me rephrase it.
- Extract the email, keep the full name.
- If there is a `<` on the full name, drop the full name, just keep the email.

Shnoulle:
> Let me rephrase it.
> 
>     * Extract the email, keep the full name.
> 
>     * If there is a `<` on the full name, drop the full name, just keep the email.

But : the purpose of the Pull request is to manage email set like this `firstname < lastname <firstname@example.org>`

No restriction on name, why add a restriction on name ? Why not `firstname lastname <firstname@example.org> <noreply@example.org>` for example

But I think a regex can be a better solution.

application/core/LimeMailer.php
Shnoulle:
```
$pattern = '/^(.*?)\s*<\s*([^<>@\s]+@[^<>@\s]+)\s*>$/';

if (preg_match($pattern, $value, $matches)) {
    $name = trim($matches[1]);
    $email = trim($matches[2]);
}
```
To have both

PR #5225, opened 2026-08-12
Fixed issue #20648: [security] SQL injection in statistics

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved survey statistics filtering to recognize only valid response fields.
  * Prevented invalid or unexpected filter values from affecting statistics queries.
  * Ensured filters are processed independently, avoiding carryover between selections.
  * Improved reliability and consistency when viewing filtered survey results.
  * Enhanced handling of filters across standard, numeric, text, date, file-upload, and multiple-choice responses.
  * Improved consistency between administrative and user-facing statistics filtering.

### DIFF
--- a/application/helpers/userstatistics_helper.php
+++ b/application/helpers/userstatistics_helper.php
@@ -331,6 +331,16 @@ function buildSelects($allfields, $surveyid, $language)
         $validColumns = array_map('strtolower', $responseSchema->getColumnNames());
     }
 
+    // Quotes a response-table column for a filter condition and rejects any
+    // identifier that is not a real column, so quoteColumnName() (which does not
+    // escape identifier quotes) cannot be abused for SQL injection (bug #20648).
+    $quoteColumn = function ($column) use ($validColumns) {
+        if (!in_array(strtolower((string) $column), $validColumns, true)) {
+            throw new InvalidArgumentException('Statistics filter references an unknown column.');
+        }
+        return Yii::app()->db->quoteColumnName($column);
+    };
+
     // creates array of post variable names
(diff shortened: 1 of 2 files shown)


application/helpers/userstatistics_helper.php
```diff
@@ -366,7 +378,8 @@ function buildSelects($allfields, $surveyid, $language)
             if (
                 $pv != "sid" && $pv != "display" && $firstletter != "M" && $firstletter != "P" && $firstletter != "T" &&
                     $firstletter != "Q" && $firstletter != "D" && $firstletter != "N" && $firstletter != "K" && $firstletter != "|" &&
-                    $pv != "summary" && substr($pv, 0, 2) != "id" && substr($pv, 0, 9) != "datestamp"
+                    $pv != "summary" && substr($pv, 0, 2) != "id" && substr($pv, 0, 9) != "datestamp" &&
+                    in_array(strtolower($pv), $validColumns, true)
```
ptelu: shouldnt this protect all following if coditions as well and be check before all the conditions so it does not need to be repeated in every if else block? coderabbit was already mentioning this as well but only for N K else block

ptelu (changes requested):

application/helpers/userstatistics_helper.php
c-schmitz: Problem is that each condition uses individual POST names making it difficult to detect valid field names with a blanket validation.
Therefore I rather replaced the quoteColumn function with one  that does the check.

ptelu (approved):

PR #5236, opened 2026-08-14
AT-2107: new feature survey project title

ramishenouda:
## Summary by CodeRabbit

* **New Features**
  * Added support for survey internal titles across survey details, editing, validation, API responses, and listings.
  * Added an internal-title badge and editing form with save, cancel, validation, permission, and error handling.
  * Added improved keyboard interaction and input handling in the editor.

* **Bug Fixes**
  * Survey selector visibility now responds correctly to configuration and focus state.

* **Maintenance**
  * Added a database update to support internal titles and refreshed editor build assets.


mbischof:
I think it would be worth reconsidering the name `projecttitle` before this becomes part of the database/API contract.

From my understanding, this field does not represent the title of a "project" as a domain entity. There is no Project entity or Survey → Project relationship involved here. Instead, the field is the **internal, language-independent title of a survey**, mainly used to identify it in the administration interface.

The existing data model already expresses this distinction quite well:

* `survey_languagesettings.surveyls_title` is the localized/public survey title.
* The new field in `surveys` is language-independent and intended for internal identification.

For that reason, I think a name such as `internal_title` would describe the actual semantics much more precisely:

```text
surveys.internal_title
    "ACME Employee Survey Q3 2026"

survey_languagesettings.surveyls_title
    de: "Mitarbeiterbefragung 2026"
    en: "Employee Survey 2026"
```

`projecttitle` suggests a different domain model to developers: that there is (or will be) a Project entity with a title and that the survey somehow belongs to that project.

This could become particularly confusing if LimeSurvey ever introduces real project/grouping functionality:

```php
$survey->projecttitle       // actually the internal survey title
$survey->project_id
$survey->project->title     // actual project title
```

The original feature request also describes the concept as an "internal title", which seems to match the actual purpose very well.

Since this PR introduces a new persistent field and API property, now would be the least expensive time to choose the precise terminology. Once integrations or plugins start depending on `projecttitle`, changing it later would be considerably harder.

So I would suggest using `internal_title` (and `internalTitle` where camelCase is appropriate) instead of `projecttitle`.

ramishenouda48:
> I think it would be worth reconsidering the name `projecttitle` before this becomes part of the database/API contract.
> 
> From my understanding, this field does not represent the title of a "project" as a domain entity. There is no Project entity or Survey → Project relationship involved here. Instead, the field is the **internal, language-independent title of a survey**, mainly used to identify it in the administration interface.
> 
> The existing data model already expresses this distinction quite well:
> 
> * `survey_languagesettings.surveyls_title` is the localized/public survey title.
> * The new field in `surveys` is language-independent and intended for internal identification.
> 
> For that reason, I think a name such as `internal_title` would describe the actual semantics much more precisely:
> 
> ```
> surveys.internal_title
>     "ACME Employee Survey Q3 2026"
> 
> survey_languagesettings.surveyls_title
>     de: "Mitarbeiterbefragung 2026"
>     en: "Employee Survey 2026"
> ```
> 
> `projecttitle` suggests a different domain model to developers: that there is (or will be) a Project entity with a title and that the survey somehow belongs to that project.
> 
> This could become particularly confusing if LimeSurvey ever introduces real project/grouping functionality:
> 
> ```
> $survey->projecttitle       // actually the internal survey title
> $survey->project_id
> $survey->project->title     // actual project title
> ```
> 
> The original feature request also describes the concept as an "internal title", which seems to match the actual purpose very well.
> 
> Since this PR introduces a new persistent field and API property, now would be the least expensive time to choose the precise terminology. Once integrations or plugins start depending on `projecttitle`, changing it later would be considerably harder.
> 
> So I would suggest using `internal_title` (and `internalTitle` where camelCase is appropriate) instead of `projecttitle`.

Thank you for the feedback. I found it valid and have applied it by renaming the project title to “Internal.” Regarding the use of more precise terminology, I would prefer to keep it as it is so that everything matches consistently in the database.

c-schmitz:
Why don't we name it survey_code?
Like question code, answer code, question group code, etc.

ramishenouda48:
> Why don't we name it survey_code? Like question code, answer code, question group code, etc.

Technically, the naming is consistent, but from a user perspective the term “code” may be misleading. Users usually interpret a code as an alphanumeric identifier or something functional, not as a project title. So the concern is that users may not realize that the value they enter there is actually the name that will later appear in the survey list.

application/models/Survey.php
```diff
@@ -558,6 +558,8 @@ public function rules()
             array('showqnumcode', 'in', 'range' => array('B', 'N', 'C', 'X', 'I'), 'allowEmpty' => true),
             array('format', 'in', 'range' => array('G', 'S', 'A', 'I'), 'allowEmpty' => true),
             array('googleanalyticsstyle', 'numerical', 'integerOnly' => true, 'min' => '0', 'max' => '3', 'allowEmpty' => true),
+            array('internaltitle', 'LSYii_FilterValidator', 'filter' => 'trim', 'skipOnEmpty' => true),
```
Shnoulle: I think we can restrict it to text only. No need for HTML here.

With LSYii_FilterValidator : we allow superadmins to add HTML.

Shnoulle:
> > Why don't we name it survey_code? Like question code, answer code, question group code, etc.
> 
> Technically, the naming is consistent, but from a user perspective the term “code” may be misleading. Users usually interpret a code as an alphanumeric identifier or something functional, not as a project title. So the concern is that users may not realize that the value they enter there is actually the name that will later appear in the survey list.

A code : [A-Z0-9], a title allow space, dot, underscore etc ...

Here: need to use it in survey list ? Right ? 

Adding getInternatlTitle Survey model function with

1. internaltitle if exist
2. Survey default language title if exist (or current title)

PR #5240, opened 2026-08-18
New feature : Replace question default-value switches with editor act…

adilmoumni:
## What was changed

- Replace the “Save as default values” switch with a button.
- Show “Clear default values” only when defaults exist.
- Add expandable hints matching the Question Settings 2.0 design.
- Save all current question attributes.
- Include the question type in save and clear operations.
- Prefill new questions of the same type with saved defaults.
- Apply saved defaults in the backend when creating questions.
- Keep existing questions unchanged when saving or clearing defaults.

## Testing

- Verified saving and overwriting defaults.
- Verified clearing defaults.
- Verified new questions receive saved defaults.
- Verified existing questions are unaffected.
- Added focused frontend and backend tests.
- React build, ESLint, Prettier, and PHP syntax checks pass.

## Summary by CodeRabbit

* **New Features**
  * Added controls to save or clear default question settings, with expandable guidance.
  * New questions now automatically use configured defaults for their question type.
  * Default settings can be applied while preserving explicitly entered values.

* **Bug Fixes**
  * Improved handling of saved defaults, including invalid or missing configuration data.
  * Restricted settings actions when editing is unavailable or a question is not yet saved.


application/models/services/QuestionAggregateService/QuestionService.php
```diff
@@ -398,6 +404,68 @@ private function saveDefaults($data)
...
+        return $defaults;
+    }
+
+    /**
+     * Return the current user's saved attribute defaults indexed by question type.
+     */
+    public function getDefaultAttributeValuesByQuestionType(): array
+    {
+        $defaultsByQuestionType = [];
+        foreach (array_keys(\QuestionType::modelsAttributes()) as $questionType) {
```
SaraDimoska: No need to loop over the modelsAttributes() list. That's one query per question type on every survey load. A single query filtering on stg_name would fetch them all at once.

application/models/services/QuestionAggregateService/QuestionService.php
```diff
@@ -398,6 +404,68 @@ private function saveDefaults($data)
...
+        }
+        return $attributes;
+    }
+
+    /**
+     * Return the current user's saved advanced attribute defaults for a
+     * question type, flattened to the API's attribute => language => value
+     * representation.
+     */
+    public function getDefaultAttributeValues(string $questionType): array
```
SaraDimoska, line 430: It looks like this method does the same thing as QuestionAttributeHelper::getUserDefaultsForQuestionType(). Same key, same decoding, same output.

You could reuse the existing one so the two can't drift apart.

SaraDimoska (changes requested):

editor/src/components/QuestionSettings/Setting.js
```diff
@@ -186,6 +191,11 @@ export const Setting = ({
         const attributeProps = {
           ...attribute.props,
           ...(options ? { options } : {}),
+          ...(attribute.action
+            ? {
+                hasDefaultAttributeValues: question.hasDefaultAttributeValues,
```
SaraDimoska: This flag only gets set when the question first loads, so if you switch the type later nothing updates, the Clear button just keeps showing (or hiding) for the old type until you reload. 

We already got survey.questionTypeDefaultAttributeValues in the payload, so you could just read it off the current type instead and it will always be right.

adilmoumni: Good catch, thanks. I’ve updated this so **hasDefaultAttributeValues.** 
Switching the question type now immediately updates the Clear button visibility without requiring a reload.

application/models/services/QuestionAggregateService/QuestionService.php
adilmoumni: I’ve removed the duplicate decoding logic

editor/src/components/QuestionSettings/Setting.js
SaraDimoska: Amazing! :D
Also, now that the check comes from React, the $question['hasDefaultAttributeValues'] in transformQuestions() isn't read anywhere anymore. Could be removed from the API response.

application/models/services/QuestionAggregateService/QuestionService.php
adilmoumni: t. I replaced the per-question-type lookups with a single user-settings query filtered

SaraDimoska (approved):
Approving 👍 

Non-blocking, but nicer to do now than clean up later: the per-question hasDefaultAttributeValues isn't read anywhere now, so it'd be good to drop it from the API response before it ships.

adilmoumni:
@SaraDimoska  Done. I removed the unused per-question hasDefaultAttributeValues field from the API response and cleaned up the front code

PR #5252, opened 2026-08-24
Fixed issue #20519: When importing a question group (or old survey) codes are badly replaced

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed survey imports so expression-code replacements are limited to newly imported groups when appropriate.
  * Prevented unintended updates to expressions, conditions, relevance settings, localized content, default values, and group descriptions in unrelated groups.
  * Improved matching accuracy by enforcing identifier boundaries during replacements.
  * Preserved survey end messages when replacements are restricted to selected groups.
  * Restored survey-wide replacements for imports that require updates across the entire survey.


Shnoulle (approved):
The regexp 😮 seem great ! 👍

PR #5262, opened 2026-08-28
TB-36: export tracking

razvanpandor:
## Summary by CodeRabbit

* **New Features**
  * Added response export tracking for supported export types and application versions.
  * Tracking information is recorded when available without interrupting the export process.

* **Bug Fixes**
  * Improved resilience by safely handling tracking configuration, database, and event-processing errors while allowing exports to complete normally.


karcstefan (changes requested):
Keep only the plugin event here

PR #5281, opened 2026-09-08
Fixed issue #20674: Invalid html lang attribute set for some translation languages

c-schmitz:
## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected HTML language tags across survey, print, maintenance, error, and question editor preview pages.
  - Language tags now use valid standardized values, including proper formatting for regional, aliased, and informal language variants.
  - Improved language metadata consistency for printable surveys, survey lists, and administrative pages.
  - Updated survey links to expose the appropriate language tag for improved browser and accessibility support.


twilligls (changes requested):
Could/Should we also apply this fix to
application/views/admin/super/header.php
and
application/views/layouts/header.php ?

Then the admin area would also benefit from it.

c-schmitz:
> Could/Should we also apply this fix to application/views/admin/super/header.php and application/views/layouts/header.php ?
> 
> Then the admin area would also benefit from it.

Done :-)

twilligls (approved):

PR #5303, opened 2026-09-14
Fixed issue #15803: No warning shown when comparing numeric result with string result

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric comparisons involving function results now correctly distinguish numeric values from non-numeric text.
  * Non-numeric text results trigger an appropriate warning, while numeric text and numeric results compare without warnings.

* **Tests**
  * Added coverage for numeric and non-numeric function results in comparisons.


Shnoulle:
Didn't check more currently, but the result returned during _Logic validation_ can be different from the function itself, no ?

I think we must add a new part in the [function definitiion](https://www.limesurvey.org/manual/ExpressionManagerStart#Function_definition) .
But it's complex to add it with the current definition (_Number of arguments_ just at the end).

tests/unit/helpers/ExpressionWarningTest.php
```diff
@@ -22,4 +22,43 @@ public function testBasic()
...
+
+    /**
+     * Functions whose result is a numeric-looking string (e.g. str_pad() used to
+     * zero-pad a number) must not trigger a false-positive type-mismatch warning.
+     * @see https://bugs.limesurvey.org/view.php?id=15803
+     */
+    public function testNoWarningWhenStringFunctionResultIsNumeric()
+    {
+        $em = new \ExpressionManager();
+        $res = $em->RDP_Evaluate('3 lt str_pad(3, 5, "0")');
```
Shnoulle, line 48: The issue is more `3 lt str_pad(3, 5, QCODE.NAOK) `for example. Where QCODE is a numerci question or a text question.

c-schmitz: Yeah, this is just a function test.

c-schmitz:
> Didn't check more currently, but the result returned during _Logic validation_ can be different from the function itself, no ?
> 
> I think we must add a new part in the [function definitiion](https://www.limesurvey.org/manual/ExpressionManagerStart#Function_definition) . But it's complex to add it with the current definition (_Number of arguments_ just at the end).

Yeah, this patch does not change logic, it only makes the warnings visible that were previously not working at all.

gabrieljenik (commented):
I like it. I think it is better for @Shnoulle to see it, considering it is EM.
I think this is not backward compatible, so, thi should be released on a major release, right?

Shnoulle:
> I like it. I think it is better for @Shnoulle to see it, considering it is EM. I think this is not backward compatible, so, thi should be released on a major release, right?

Why not _backward compatible_ ? There are no update of logic, just show the warning or not.

gabrieljenik:
> > I like it. I think it is better for @Shnoulle to see it, considering it is EM. I think this is not backward compatible, so, thi should be released on a major release, right?
> 
> Why not _backward compatible_ ? There are no update of logic, just show the warning or not.

Shows warning or throws errors? If just warns, you are right.

PR #5308, opened 2026-09-15
Fixed issue #19747: Theme configuration hierarchy loading is not correct

c-schmitz:
## Summary
- getInstanceFromSurveyGroup() and getInstanceFromSurveyId() mutated the generic template instance returned by getInstanceFromTemplateName() in place instead of cloning it
- That generic instance is cached by template name and returned by reference, so the mutation corrupted the cached base template configuration for all subsequent lookups of that template
- Clone the instance before mutating it in both methods so the cache stays intact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where creating survey-group or survey-specific template configurations could unintentionally modify or reuse cached generic configuration values.
  - New configurations now resolve inherited settings in their own context while preserving the original generic template configuration.
  - Improved consistency when initializing and saving configurations for individual survey groups or surveys.


gabrieljenik:
@c-schmitz Have you seen the comment by CodeRabbit? https://github.com/LimeSurvey/LimeSurvey/pull/5308/changes#r4017734026

Maybe the TemplateConfiguration should implement the clone method?

c-schmitz:
Fixed the CodeRabbit AI issue 

> Maybe the TemplateConfiguration should implement the clone method?

I checked but would not make sense.

gabrieljenik (approved):

PR #5317, opened 2026-09-17
 New feature #LX-8: Add Quota Exit field to responses table to record the matched quota

c-schmitz:
## Summary by CodeRabbit

- **New Features**
  - Added an optional “Save quota exit” setting for surveys and survey groups.
  - Matched quota IDs can now be saved with responses and displayed during data entry and response listings.
  - Added quota-exit details to survey responses, imports, exports, and API data.
  - Added inherited settings and activation controls for configuring this behavior.

- **Bug Fixes**
  - Improved quota-exit handling during response recovery and validation errors.
  - Corrected numeric formatting for quota-exit values in statistical exports.
  - Updated survey storage to support the new setting and response information.


c-schmitz:
@Shnoulle Created a clean patch since the other PR had too many comments. If you could  check and give your thumbs up, would be nice.

application/helpers/remotecontrol/remotecontrol_handle.php
```diff
@@ -454,13 +454,13 @@ public function get_survey_properties($sSessionKey, $iSurveyID, $aSurveySettings
...
-     * Properties available are restricted
-     * * Always
+     * Some properties may not be modified depending on the survey's state
+     * * Always restricted
      *     * sid
      *     * active
      *     * language
      *     * additional_languages
-     * * If survey is active
+     * * Restricte if survey is active
```
Shnoulle, line 463: Must not add quotaexit in the comment?

application/helpers/update/updates/Update_715.php
```diff
@@ -0,0 +1,54 @@
...
+    public function up()
+    {
+        // Get all active survey IDs (includes expired surveys)
+        $command = \Yii::app()->db->createCommand();
+        $activeSurveyIds = $command->select('sid')
+            ->from('{{surveys}}')
+            ->where("active='Y'")
+            ->queryColumn();
+
+        foreach ($activeSurveyIds as $surveyId) {
```
Shnoulle, line 19: Not needed since it's an optionnal set to I by defulat. See next comment

application/helpers/update/updates/Update_715.php
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            \Yii::app()->db->createCommand()->addColumn(
+                $responseTableName,
+                'quota_exit',
+                'integer'
+            );
+        }
+
+        // Add to surveys table if not exists
```
Shnoulle, line 37: My opinion

1. Add to survey as I for default
2. Add to survey groups as I for default
3. Set as N (if not already exist) for survey group 0 (i do it in a previous update)
4. Set as N to Surveys WHERE active = 'Y'

Shnoulle: See the comment on the previous PR: https://github.com/LimeSurvey/LimeSurvey/pull/4651#discussion_r4035878417

Shnoulle (changes requested):
The update is not done for optional settings here

application/helpers/update/updates/Update_715.php
c-schmitz: Sure, but probably people don't want to deactivate and reactivate all their existing survey to activate the feature.

application/helpers/remotecontrol/remotecontrol_handle.php
c-schmitz: The quotaexit field is always created. Why restrict it?

Shnoulle (commented):
Must add it in https://github.com/LimeSurvey/LimeSurvey/blob/f57d0e4018a264aee40288dc157ba9c2f0496b95/application/models/services/SurveyAggregateService/GeneralSettings.php#L285 
with ['default' => 'I'],

application/helpers/update/updates/Update_715.php
c-schmitz: Update_715's existing code already does the right thing: addColumn(..., DEFAULT 'N') backfills all existing rows (including gsid=0) with the concrete value 'N', then update(..., ['savequotaexit' => 'I'], 'gsid<>0') immediately overrides every row except gsid=0 to 'I'. So migrated non-global rows (including gsid=1) already end up 'I' — the premise that they don't is false.

Changing the column's schema-level DEFAULT from 'N' to 'I'  would actually introduce a real bug: it wouldn't change what the migrated rows end up with (that's driven by the explicit UPDATE, not the DEFAULT), but it would make the true global row (gsid=0) default to 'I' — "inherit" from nothing — which is invalid.

application/helpers/update/updates/Update_715.php
Shnoulle: > Sure, but probably people don't want to deactivate and reactivate all their existing survey to activate the feature.

But you potentially broke some upgrade : https://github.com/LimeSurvey/LimeSurvey/pull/4651#discussion_r4035878409 

Set it to I, and N for activated survey : no issue

c-schmitz:
> Must add it in
> 
> https://github.com/LimeSurvey/LimeSurvey/blob/f57d0e4018a264aee40288dc157ba9c2f0496b95/application/models/services/SurveyAggregateService/GeneralSettings.php#L285
> 
> with ['default' => 'I'],

@Shnoulle The file you are referring here is outdated. Please check the current version.

application/helpers/update/updates/Update_715.php
c-schmitz: Well, we had bigger upgrades ;). 
I added a fix so that update will continue even if an addColumn fails.

Shnoulle:
> > Must add it in
> > https://github.com/LimeSurvey/LimeSurvey/blob/f57d0e4018a264aee40288dc157ba9c2f0496b95/application/models/services/SurveyAggregateService/GeneralSettings.php#L285
> > 
> > with ['default' => 'I'],
> 
> @Shnoulle The file you are referring here is outdated. Please check the current version.

What ? It still in develop_minor and i have issue if i don't add if in encryotion systeml (don't remind when)

And it's still there  [   private function getFields(Survey $survey)](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/services/SurveyAggregateService/GeneralSettings.php#L211) without any deprecated indication.

application/helpers/update/updates/Update_715.php
Shnoulle: I see for SurveyGroup, but not for Survey currently.

savequotaexit didn't exist before this update then cabn be set to I for all surveys.

Shnoulle: Since you try to add column to all tables : you can set it to Y when update Survey column maybe ?

Shnoulle: With default to Y for active survey

1. Create Surveys->savequotaexit = I
2. Create SurveyGroups->savequotaexit = I, set SurveyGroups->savequotaexit = Y for SurveyGroup 0
3. If the loop : select Surveys->active = Y and Surveys->savequotaexit = 'I' response table only
4. In the response table update: set savequotaexit='Y' if it's OK to N if not in Surveys table.

c-schmitz:
> What ? It still in develop_minor and i have issue if i don't add if in encryotion systeml (don't remind when)
> 
> And it's still there [ private function getFields(Survey $survey)](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/services/SurveyAggregateService/GeneralSettings.php#L211) without any deprecated indication.

I mean the link to that file is outdated. It shows contents from the previous PR, not the current one.

Shnoulle:
> > What ? It still in develop_minor and i have issue if i don't add if in encryotion systeml (don't remind when)
> > And it's still there [ private function getFields(Survey $survey)](https://github.com/LimeSurvey/LimeSurvey/blob/master/application/models/services/SurveyAggregateService/GeneralSettings.php#L211) without any deprecated indication.
> 
> I mean the link to that file is outdated. It shows contents from the previous PR, not the current one.

Ok, then in master ; https://github.com/LimeSurvey/LimeSurvey/blob/5766a01d6d1fce48a51aeef04f67a0474bca1aa3/application/models/services/SurveyAggregateService/GeneralSettings.php#L211

application/helpers/remotecontrol/remotecontrol_handle.php
tiborpacalat: Also typo "Restricte" -> "Restricted"

application/helpers/update/updates/Update_715.php
```diff
@@ -0,0 +1,59 @@
...
+            ->where("active='Y'")
+            ->queryColumn();
+
+        foreach ($activeSurveyIds as $surveyId) {
+            $responseTableName = "{{responses_" . $surveyId . "}}";
+            if (!tableExists($responseTableName)) {
+                continue;
+            }
+
+            \Yii::app()->db->createCommand()->addColumn(
```
karcstefan: We should check here whether this column exists, if for some reason this update fails and needs to be re-run, will exit out due to duplicate column error

application/helpers/update/updatedb_helper.php
```diff
@@ -1704,6 +1704,9 @@ function createSurveysGroupSettingsTable(CDbConnection $oDB)
     /* Added in 712 update */
     unset($attributes['preselectnoanswer']);
 
+    /* Added in 709 update */
```
karcstefan: Small typo - "Added in 715 update"

application/helpers/common_helper.php
```diff
@@ -1657,6 +1664,14 @@ function createFieldMap($survey, $style = 'short', $force_refresh = false, $ques
         }
     }
 
+    // Add 'quota_exit' to fieldmap.
+    $fieldmap["quota_exit"] = array("fieldname" => "quota_exit", 'type' => "quota_exit", 'sid' => $surveyid, "gid" => "", "qid" => "", "aid" => "");
```
karcstefan, line 1670: Should we add a check here before adding it to the fieldmap like ipaddr and refurl?
`if($prow['savequotaexit'] == "Y")`

application/helpers/common_helper.php
```diff
@@ -1191,6 +1191,13 @@ function getExtendedAnswer($iSurveyID, $sFieldCode, $sValue, $sLanguage, $questi
                 $sValue = convertDateTimeFormat($sValue, "Y-m-d H:i:s", $dateformatdetails['phpdate'] . ' H:i:s');
             }
             break;
+        case 'quota_exit':
+            // Try to get quota name
+            $quota = Quota::model()->findByAttributes(['sid' => $iSurveyID, 'id' => $sValue]);
```
karcstefan: Check for empty/null values here to avoid additional query per call

application/helpers/common_helper.php
c-schmitz: No, the field should always exist, so the feature can be activated/deactivated at any time.

application/helpers/update/updates/Update_715.php
c-schmitz: Added a try/except protection. Checking for the column is too slow.

karcstefan (approved):

Shnoulle:
I'm still upset about the table size limit where this extra column can break activation.

But decision here ....

c-schmitz:
> I'm still upset about the table size limit where this extra column can break activation.

It is very unlikely as table size limit (at least for MySQL myisam) has been increased by 10 - 50%  with the new response table format in DBversion 700.

Shnoulle:
> > I'm still upset about the table size limit where this extra column can break activation.
> 
> It is very unlikely as table size limit (at least for MySQL myisam) has been increased by 10 - 50% with the new response table format in DBversion 700.

Maybe we don't need survey option then ? If we add it to al surveys, I don't find why someone need to disable it ?

(Sorry for the primary opinion here)

PR #5322, opened 2026-09-19
Fixed issue #20702: Unclear error when importing CPDB participants without permission

c-schmitz:
## Summary
- When importing a CSV into the Central Participant Database (CPDB), a user without Update permission on an existing participant would either silently have that record skipped/overwritten, or see the generic "duplicate entry, no need to create a new participant" warning — with no indication that the real reason was a permissions issue.
- `uploadCSV()` now checks the current user's global `participantpanel` update permission and ownership/share on the specific existing participant before overwriting it, matching the permission pattern already used elsewhere in the model.
- Records that were skipped for this reason are now reported separately in the upload summary with a clear "already exist but you do not have permission to update them" message, instead of being lumped into the generic duplicate message.

Fixes https://bugs.limesurvey.org/view.php?id=20702

## Test plan
- [ ] As superadmin, import a CSV creating participants with known `participant_id`s.
- [ ] Create a user with only Create + Import permission on the CPDB (no Update).
- [ ] Log in as that user and import a second CSV with the same `participant_id`s but changed fields, with "Overwrite" checked.
- [ ] Verify the participant data is *not* overwritten and the summary shows the new "no permission to update" warning instead of the generic duplicate message.
- [ ] Repeat as a user who owns those participants (or as superadmin) and confirm overwrite still works as before.

## Summary by CodeRabbit

* **New Features**
  * Participant CSV uploads now check update permissions and ownership before overwriting existing participants.
  * Upload summaries list existing participants that could not be updated due to insufficient permissions.

* **Bug Fixes**
  * Prevented unauthorized overwrites of existing participant records during CSV uploads.


application/controllers/admin/ParticipantsAction.php
```diff
@@ -1042,10 +1049,15 @@ public function uploadCSV()
                     ];
                 }
                 $existingParticipant = Participant::model()->findByAttributes($duplicateCriteriaAttributes);
+                $noPermissionForExisting = false;
                 if (!empty($existingParticipant)) {
                     $thisduplicate = 1;
                     $dupcount++;
-                    if ($overwrite == "true") {
+                    $canUpdateExisting = $hasUpdatePermission && Participant::model()->isOwner($existingParticipant->participant_id);
```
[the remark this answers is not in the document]
c-schmitz: I think the chance of this is very low, and row level checking would be too slow.

application/controllers/admin/ParticipantsAction.php
```diff
@@ -1042,10 +1049,15 @@ public function uploadCSV()
...
+                    if ($overwrite == "true" && !$canUpdateExisting) {
+                        // The user requested an overwrite but has no update permission on this particular participant
```
gabrieljenik, line 1058: what if $canUpdateExisting is false because $hasUpdatePermission  is false?
The comment "// The user requested an overwrite but has no update permission on this particular participant" may be missleading.

application/views/admin/participants/uploadSummary_view.php
```diff
@@ -50,6 +51,14 @@
             }
             $warningSummary .= "</ul></div></li>";
         }
+        if (count($noPermissionList) > 0) {
+            $warningSummary .= "<li>" . sprintf(gT("%s participants already exist but you do not have permission to update them."), count($noPermissionList));
```
gabrieljenik, line 55: This message could be missleading as well. They could lack permission to update any participant.

Shnoulle: Then : he add no permission to update this participants.

Maybe it's whole particpants, but it's not tyhe issue here.

PR #5323, opened 2026-09-20
Dev PHP 8.5 fixes, minimum coverage and CI compatibility

c-schmitz:
This PR gets our CI testing on PHP 8.5 (in addition to 8.1), and fixes everything that broke along the way - outdated dependencies, deprecated code, a few real bugs the stricter checks caught, and a code-coverage check that turned out to have been silently broken. 

Also added some missing tests to make that coverage check meaningful again. Also cleaned up some CI warnings and gave the main workflow a clearer name.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission checks for users with differing effective and session identities.
  * Corrected survey and survey-group settings lookups for null or non-integer identifiers.
  * Fixed quota-exit export field mapping and improved CAPTCHA reliability.
  * Prevented incorrect caching for null survey identifiers.

* **Tests**
  * Added extensive survey statistics and chart-processing coverage.
  * Improved test initialization and validation.

* **Chores**
  * Updated PHP tooling, static analysis, CI workflows, and dependency exclusions.


.github/workflows/main.yml
```diff
@@ -166,7 +166,7 @@ jobs:
           XDEBUG_MODE: coverage
         run: php vendor/bin/phpunit --testdox --stop-on-failure --fail-on-skipped ${{ matrix.db-type != 'mysql' && '--exclude-group mysql' || '' }} tests/unit
       - name: Upload cov.xml
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
```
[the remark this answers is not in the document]
Ahmedazeem98: @c-schmitz 
Should we take this into account?

 uses `actions/upload-artifact@v7`  is valid and will work, but it is a mutable tag, meaning the tag could be moved later and the workflow would run different code without a YAML change. Pinning it to the full commit SHA makes the executed code immutable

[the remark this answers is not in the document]
c-schmitz: ok, fixed

application/helpers/update/updates/Update_700.php
```diff
@@ -2010,7 +2011,7 @@ public function up()
         $responsesTables = $this->db->createCommand($this->scriptMapping['responses'])->queryAll();
         foreach ($responsesTables as $responsesTable) {
             if (((strpos($responsesTable['old_name'], 'old_') === false) && (strpos($responsesTable['old_name'], 'timing') === false))) {
-                $parts = explode('_', $responsesTable['old_name']);
+                explode('_', $responsesTable['old_name']);
```
Ahmedazeem98, line 2014: Since `$parts` is unused, this entire `if` block can be removed. The `explode()` result is neither assigned nor used, so the operation has no effect.

application/helpers/update/updatedb_helper.php
```diff
@@ -2698,7 +2695,7 @@ function upgradeSurveys145()
         $aDefaultTexts = templateDefaultTexts($sLanguage, 'unescaped');
         unset($sLanguage);
         $aDefaultTexts['admin_detailed_notification'] = $aDefaultTexts['admin_detailed_notification'] . $aDefaultTexts['admin_detailed_notification_css'];
-        $sSurveyUpdateQuery = "update {{surveys_languagesettings}} set
+        "update {{surveys_languagesettings}} set
```
Ahmedazeem98, line 2698: Should remove the variable's value as well not only `$sSurveyUpdateQuery = `

`"update {{surveys_languagesettings}} set
        email_admin_responses_subj=" . $aDefaultTexts['admin_detailed_notification_subject'] . ",
        email_admin_responses=" . $aDefaultTexts['admin_detailed_notification'] . ",
        email_admin_notification_subj=" . $aDefaultTexts['admin_notification_subject'] . ",
        email_admin_notification=" . $aDefaultTexts['admin_notification'] . "
        where surveyls_survey_id=" . $aSurveyRow['surveyls_survey_id'];`

The snippt above should be removed  because the real update is already done immediately below with Yii’s query builder

application/helpers/update/updatedb_helper.php
```diff
@@ -3337,7 +3334,7 @@ function runAddPrimaryKeyonAnswersTable400(&$oDB)
         $oDB->createCommand()->renameTable('{{answers}}', 'answertemp');
         $oDB->createCommand()->createIndex('answer_idx_10', 'answertemp', ['qid', 'code', 'scale_id']);
 
-        $dataReader = $oDB->createCommand("SELECT qid, code, scale_id FROM answertemp group by qid, code, scale_id")->query();
+        $oDB->createCommand("SELECT qid, code, scale_id FROM answertemp group by qid, code, scale_id")->query();
```
Ahmedazeem98, line 3337: the redundant first query should be removed at all.

The query before `createTable()` **at line 3337** had no side effects and its result was not consumed. The later identical query is the one actually used by the `while (($row = $dataReader->read()) !== false)` loop, so it needs to stay (**line 3350**).

So the whole `$oDB->createCommand("SELECT qid, code, scale_id FROM answertemp group by qid, code, scale_id")->query();` **at line 3337** should be removed

Ahmedazeem98 (changes requested):

c-schmitz:
@coderabbitai full review

Ahmedazeem98 (approved):

PR #5326, opened 2026-09-21
Fixed issue #19095: [security] Session does not expire on password change

c-schmitz:
Fixed issue #19095: [security] Insufficient session invalidation vulnerability in user password change. Severity: Medium (CVSS 5.4, AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:N). An attacker who has obtained a victim's active session (e.g. via a shared device or a previously hijacked session) could retain access to that session even after the victim changes their password to lock them out. Affects all versions up to 7.1.2.

## Summary
- Password changes (self-service, forgot-password reset, or an admin resetting another user's password) only updated the password hash — any other already-authenticated session for that user kept working indefinitely.
- Since sessions aren't stored keyed by `user_id`, there's no way to look up and kill a user's other sessions directly. Instead, a per-user `session_token` (new nullable `users.session_token` column, DB v715) is rotated on every password change and cached in `$_SESSION` at login.
- `LSApplicationTrait::getCurrentUserId()` (the single per-request validity check already used for expired/deactivated accounts) now also compares the DB token against the session-cached one; a mismatch forces that session's existing logout path.
- The session that performs its own password change (via personal settings) refreshes its own cached token immediately, so it stays logged in.

Severity: Medium (CVSS 5.4, `AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:N`). Affects all versions up to 7.1.2. Fixed in 7.1.3.

## Test plan
- [ ] Log in as the same user in two browsers, change the password in one, confirm the other is logged out on its next request
- [ ] Change password via forgot-password email link, confirm any previously active session for that user is invalidated
- [ ] Admin resets another user's password via User Management, confirm that user's existing session is invalidated
- [ ] Fresh login on an unmigrated/upgraded install still works (lazy token generation)
- [ ] Run `application/helpers/update/updates/Update_715.php` against an existing DB and confirm `users.session_token` is added without error

## Summary by CodeRabbit

* **Security**
  * Password changes now invalidate authenticated sessions on other devices.
  * Session credentials are refreshed after updating personal settings, keeping the current session active.
  * One-time-password sign-ins are more reliably associated with the authenticated account.

* **Bug Fixes**
  * Improved session validation helps prevent outdated credentials from being used.

* **Maintenance**
  * Database updates add support for session credential tracking.
  * Password reset and administrator password changes now apply session updates consistently.


c-schmitz:
Commit message should be:

Fixed issue #19095: [security] Insufficient session invalidation vulnerability in user password change. Severity: Medium (CVSS 5.4, AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:N). An attacker who has obtained a victim's active session (e.g. via a shared device or a previously hijacked session) could retain access to that session even after the victim changes their password to lock them out. Affects all versions up to 7.1.2.

application/models/User.php
```diff
@@ -346,24 +346,44 @@ public static function findByUsername($sUserName)
...
      *
+     * Also rotates the user's session token, so any other already
+     * authenticated session for this user (which still caches the previous
+     * token) gets logged out on its next request. See
+     * LSApplicationTrait::getCurrentUserId().
+     *
      * @param string $sPassword The clear text password
      * @return \User
      */
     public function setPassword($sPassword, $save = false)
```
karcstefan, line 369: REST api token and JSON-RPC tokens are never invalided.. maybe an addition of
`Session::model()->deleteAllByAttributes(['data' => $this->users_name]);` here?

karcstefan (changes requested):
We need to catch every reset password so that the session token is also rotated.

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/UserManagementController.php#L1614
`$oUser->setPassword((string) $aUser['password'])`

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/commands/ResetPasswordCommand.php#L31
`$oUser->save(true, ['password', 'session_token'])`

https://github.com/LimeSurvey/LimeSurvey/blob/master/application/controllers/UserManagementController.php#L952

application/models/User.php
c-schmitz: Good catch, fixed in 36c90a0667: `setPassword()` now also calls `Session::model()->deleteAllByAttributes(['data' => $this->users_name])` to revoke outstanding RemoteControl (JSON-RPC/REST) tokens.

karcstefan (approved):

PR #5327, opened 2026-09-21
Fixed issue #20699: [security] Broken access control in participant sharing

c-schmitz:
Fixed issue #20699: [security] Broken access control vulnerability in participant sharing. Severity: High (CVSS 7.1, AV:N/AC:L/PR:L/UI:N/S:U/C:L/I:H/A:N). An authenticated user holding only participant-panel read permission and a read-only share of a participant could grant itself, or any other account of its choosing, edit rights on that participant, allowing it to modify and redistribute data it had no relationship to. Affects versions 3.17.4 through 7.1.2. Reported by: Ada Logics.

## Summary
- The participant-sharing endpoint (issue #20699) computed its update-permission flag using the wrong permission entity, so it never matched a real permission for ordinary accounts.
- The model-level authorization check that was meant to gate creating/updating a share was dead code (`!$permission` on an array that is never empty), so the endpoint enforced no authorization at all beyond refusing to share a participant with its own owner.
- As a result, a user holding only read-only participant-panel access and a read-only share of a participant could grant itself (or any other account of its choosing) edit rights on that participant, then use those rights to edit or redistribute data it had no relationship to.
- Fixed by correcting the permission check and replacing the dead condition with an explicit check: only the participant's owner, a superadmin, a user with participant-panel update rights, or a user who already holds an editable share of that participant may create or update a share.

Severity: High (CVSS 7.1, AV:N/AC:L/PR:L/UI:N/S:U/C:L/I:H/A:N). Affects versions 3.17.4 through 7.1.2, fixed in 7.1.3.

Reported by: Ada Logics.

## Test plan
- [x] Verify a superadmin can still share/edit participants and grant shares as before.
- [x] Verify a user with participant-panel update permission can still create/update shares.
- [x] Verify a read-only share holder can no longer flip `can_edit` on their own share.
- [x] Verify a read-only share holder can no longer create a share (with or without edit rights) for another account.
- [x] Verify a user who already holds an editable share can *not* share the participant onward.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected participant-sharing access checks to use the appropriate participant-panel update permission.
  - Strengthened authorization so sharing is allowed only for participant owners, super administrators, or users with participant-panel update permission.
  - Prevented unauthorized share creation, updates, editability changes, and deletion of selected shares.
  - Improved responses for unauthorized requests by returning appropriate permission or failure results.


c-schmitz:
The proper commit message will be:

Fixed issue #20699: [security] Broken access control vulnerability in participant sharing. Severity: High (CVSS 7.1, AV:N/AC:L/PR:L/UI:N/S:U/C:L/I:H/A:N). An authenticated user holding only participant-panel read permission and a read-only share of a participant could grant itself, or any other account of its choosing, edit rights on that participant, allowing it to modify and redistribute data it had no relationship to. Affects versions 3.17.4 through 7.1.2. Reported by: Ada Logics.

haythemch2 (approved):

PR #5328, opened 2026-09-21
Fixed issue #20700: [security] Missing permission and POST checks on Expression Manager maintenance actions

c-schmitz:
## Summary
- Expression Manager admin actions (`admin/expressions`) dispatched sub-actions based on the `sa` query parameter with almost no permission check — only two names (`survey_logic_file`, `navigation_test`) were gated, and only when a survey id was present.
- Several unlisted sub-actions read or rewrite `relevance` for every survey in the database with no survey scoping at all: `upgrade_conditions2relevance`, `revert_upgrade_conditions2relevance`, `upgrade_relevance_location`, and `conditions2relevance` (read-only preview).
- An authenticated user with no rights over any survey could reach these via plain GET requests and rewrite/reset another user's survey branching logic installation-wide.

## Fix
- `Expressions::index()` now requires global superadmin permission for the installation-wide sub-actions listed above.
- The three that actually write to the database additionally require a POST request (`requirePostRequest()`), closing the CSRF/bookmarked-link angle.
- Updated the two GET links to the bulk-convert action (`_configuration_menu.php`, `test_view.php`) into POST forms so the legitimate feature keeps working for superadmins.

## Test plan
- [ ] As superadmin, confirm "Bulk convert conditions to ExpressionScript" still works from both the mega-menu and the Expression Manager test suite page
- [ ] As a non-superadmin user, confirm `admin/expressions/sa/upgrade_conditions2relevance` (and the other three gated sub-actions) now return "Access denied"
- [ ] Confirm a GET request to `admin/expressions/sa/upgrade_conditions2relevance` as superadmin now returns 405 instead of executing

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved access controls for installation-wide administration actions.
  - Restricted database-changing actions to POST requests.
  - Updated bulk conditions-to-relevance actions to submit securely through forms instead of links.


c-schmitz:
Squash-merge commit message:

Fixed issue #20700: [security] Missing permission and POST checks on Expression Manager maintenance actions. Severity: Medium (CVSS 6.5, AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:H/A:N). An authenticated user with no permissions over other users' surveys could rewrite or reset the branching/skip logic of every survey in the installation. Affects all versions up to 7.1.2. Reported by: Ada Logics.

karcstefan (approved):

PR #5335, opened 2026-09-21
Fixed issue #20713: Ranking question answers exported as codes instead of full text

c-schmitz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ranking-question exports now expand responses into rank-position columns based on the highest rank used.
  * Full-answer exports display translated subquestion text in rank order, while other export formats preserve raw answer codes.
  * Empty, null, malformed, unused, and unmatched values are handled safely without losing available content.

* **Tests**
  * Added regression coverage across supported export paths for dynamic ranking columns, translated text, fallback handling, and blank padding.


application/helpers/admin/export/SurveyObj.php
```diff
@@ -117,7 +114,13 @@ public function getFullAnswer($fieldName, $answerCode, Translator $translator, $
                 break;
 
             case Question::QT_R_RANKING:   // Ranking TYPE
-                $fullAnswer = $answer;
+                // Ranking options are stored as subquestions, not as rows in the answers table,
```
Shnoulle: No need json_decode/json_encode ?

c-schmitz: Yeah, stupid me. Should have seen that.

application/helpers/admin/export/SurveyObj.php
```diff
@@ -117,7 +114,19 @@ public function getFullAnswer($fieldName, $answerCode, Translator $translator, $
...
+                // Ranking options are subquestions, not rows in the answers table
+                // (hence not resolved by the generic $answers lookup above), and by
+                // this point $answerCode is expected to be a single subquestion
+                // code for one rank position: the export layer decodes the
+                // question's raw JSON-array storage into one column per rank
+                // before this is called. Non-string values (including one that
+                // slipped through as invalid/malformed JSON) are returned
+                // unchanged rather than passed to the title lookup.
+                if (!is_string($answerCode) || $answerCode === "") {
+                    $fullAnswer = $answerCode;
```
gabrieljenik, line 126: Should we cast it to string here?

